PR #5, opened 2017-11-23
Add message detail and warn criteria

aidantwoods:
* Make the message less scary in favour of providing actionable detail
* Only warn if breach was verified (I think this makes sense)

### DIFF
--- a/client-addon/src/bootstrap.js
+++ b/client-addon/src/bootstrap.js
@@ -20,11 +20,7 @@ function initSiteList() {
       let sites = JSON.parse(xhr.responseText);
       siteSet = new Set(sites.map(site => site.Domain));
       breachData = sites.reduce((lookup, site) => {
-        lookup[site.Domain] = {
-          breachDate: site.BreachDate,
-          verified: site.IsVerified,
-          dataClasses: site.DataClasses,
-        }
+        lookup[site.Domain] = site
         return lookup
       }, {} )
       siteSet.add("haveibeenpwned.com");
@@ -69,7 +65,8 @@ function warnIfNeeded(browser, host) {
     host = host.substring(4);
   }
 
-  if (warnedHostSet.has(host) || !siteSet.has(host) || !breachData[host].verified) {
+  let data = breachData[host]
+  if (warnedHostSet.has(host) || !data || !data.IsVerified) {
     return;
   }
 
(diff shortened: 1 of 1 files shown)


aidantwoods:
E.g.
screen shot 2017-11-23 at 02 22 48

---

Please throw any and all code structure/style changes at me if desired – I'm not a JS dev ;p

client-addon/src/bootstrap.js
```diff
@@ -54,13 +62,14 @@ function stopObserving() {
 
 var siteSet = new Set();
 var warnedHostSet = new Set();
+var breachData = {}
```
nhnt11: Instead of having a new object, we can change siteSet to be a Map (breachData is a good name; let's keep it). Then, whenever we need to get the data for a host, we can do (for example)

```
let data = breachData.get(host);
if (!warnedHostSet.has(host) || !data || !data.verified) {
  return;
}
// Etc
```

Apart from that, here's a nit: please terminate statements with a semicolon. :)

client-addon/src/bootstrap.js
```diff
@@ -108,18 +117,24 @@ function warnIfNeeded(browser, host) {
...
-        doc.defaultView.alert("Breach alerts prototype: user successfully registered with server.");
-      }
-    };
-    let addUserURL = Services.prefs.getCharPref("extensions.breachalerts.serverURL", kDefaultServerURL) + "/user/add";
-    xhr.open("POST", addUserURL, true);
-    xhr.setRequestHeader("Content-Type", "application/json;charset=UTF-8");
-    xhr.send(JSON.stringify({ email: aText }));
-  }, "Enter email address and press enter");
+  showPopupNotification(
+    "\"" + host + "\" suffered a data breach on " + breachData[host].breachDate
```
nhnt11: For readability, let's construct this string, store it in a variable, and pass it to showPopupNotification instead of inlining it.

client-addon/src/bootstrap.js
aidantwoods: > Apart from that, here's a nit: please terminate statements with a semicolon. :)

Hah 😉 I probably should have noticed you were doing that in surrounding code – I never know what people prefer in languages where it is optional ;p

aidantwoods:
Let me know whether to keep 3dfedc5, I find it easier to write about the conditions which you want to be true than false (avoids need of parens due to and/or precedence here too – but the `else` might be bit hidden at the end so 🤷‍♂️)

groovecoder:
The add-on code has been removed from this repo. Closing this PR.

aidantwoods:
Shall I resubmit this to the repo to which the client code has moved to? (if there is one?)

PR #8, opened 2017-11-23
Lint the code using ESLint

### LINKED ISSUE #36: Add ESLint rule to prevent usage of process.env.foo
Environment variable usage should be done via AppConstants.

pdehaan:
Fixes #36


server/package.json
```diff
@@ -1,4 +1,5 @@
 {
+  "private": true,
```
pdehaan: Prevents an accidental push to npm, although that probably isn't a concern since there is no `name` or `version` field or any other required fields in the package.json file. ¯\\\_(ツ)\_/¯

.eslintrc.js
```diff
@@ -0,0 +1,19 @@
...
+  globals: {
+    Components: false,
+    Services: false
+  },
+  rules: {
+    'eqeqeq': 'warn',
+    'no-var': 'warn',
+    'prefer-const': 'warn'
+  }
+};
```
pdehaan: The one "gotcha" with these files is that we don't have a root level package.json file, so we aren't installing ESLint locally, and assume that the user would have it installed globally (assuming one wanted to run `$ eslint .`)

pdehaan:
Also, if you're planning on merging this back into **moz-central**, we may also want to add the [**eslint-plugin-mozilla**](https://www.npmjs.com/package/eslint-plugin-mozilla) plugin.

pdehaan:
I committed the package-lock.json file, but that should be easy enough to remove if we don't want it. Or I think we can add an ".npmrc" `package-lock=false` value, if we want to disable that file (per https://codeburst.io/disabling-package-lock-json-6be662f5b97d).

pdehaan:
But the TLDR is that I mostly just ran `eslint . --fix`, hand tweaked a few errors/warnings, sorted the package.json, converted a few places to arrow functions and template strings, since we were already using ES6 elsewhere. ¯\\\_(ツ)\_/¯

groovecoder:
@pdehaan - the README update PR is merged. Want to try to update this one?

pdehaan:
@groovecoder I'm rebased and ready to land, unless you're saying you want me to edit the README doc with some linting steps.

nhnt11:
Awesome. Before we review/merge this: is there a standard set of ESLint rules you're using? Do we have a code-style document somewhere?

pdehaan:
> Awesome. Before we review/merge this: is there a standard set of ESLint rules you're using? Do we have a code-style document somewhere?

I'm just using the ESLint recommended rules, and a few random custom ones. AFAIK, Mozilla doesn't have a global standard that we use (unless you're an add-on, in which case we prefer to extend `eslint-plugin-mozilla`). Some projects just use custom rules, or eslint:recommended, or AirBnb. Just as long as we have _some_ linting, I'm happy. We can tweak and optimize rules later, if needed.

pdehaan:
Alright. Rebased and ready for some 👀.

.eslintrc.js
```diff
@@ -0,0 +1,21 @@
...
+    "eslint:recommended",
+  ],
+  root: true,
+  rules: {
+    "comma-dangle": "off",
+    "eqeqeq": "warn",
+    "no-console": "warn",
+    "no-unused-vars": ["error", {vars: "all", args: "none", ignoreRestSiblings: false}],
+    "no-var": "error",
+    "prefer-const": "error",
```
nhnt11: Let's not do prefer-const please. I really hate seeing `for (const foo of bar)` for example. My brain is used to consts being literals for stuff like like cookie lifetime, etc.

.eslintrc.js
```diff
@@ -0,0 +1,19 @@
...
+    webextensions: true
+  },
+  globals: {
+    Components: false,
+    Services: false
+  },
+  rules: {
+    'eqeqeq': 'warn',
+    'no-var': 'warn',
+    'prefer-const': 'warn'
```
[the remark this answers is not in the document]
pdehaan: Downgraded from an `"error"` to a `"warn"`ing.
We could also do `// eslint-disable-next-line prefer-const` for the `for..in` and `for..of` loops to disable it in specific cases.

pdehaan:
@nhnt11 The `prefer-const` rule is now turned off, and I added the `no-process-env` rule while I was at it.

routes/user.js
```diff
@@ -40,37 +40,38 @@ router.post("/add", function(req, res) {
...
   res.json({ email: req.body.email, info: "removed user" });
 });
 
-router.post("/reset", function(req, res) {
+router.post("/reset", (req, res) => {
   gEmails.clear();
   res.json({ info: "user list cleared" });
 });
 
 if (process.env.DEBUG_ALLOW_USER_LIST) {
```
nhnt11: Do we need to whitelist this for eslint?

email-utils.js
```diff
@@ -16,34 +16,34 @@ const EmailUtils = {
   init() {
     // Allow a debug mode that will send JSON back to the client instead of sending emails.
     if (process.env.DEBUG_DUMMY_SMTP) {
```
nhnt11: And this

routes/user.js
pdehaan: Yeah, I can add `// eslint-disable-next-line no-process-env`, but wasn't sure if we were keeping those flags, or adding them to AppConstants, per https://github.com/mozilla/blurts-server/issues/36#issuecomment-364069848.

pdehaan: Done!

PR #13, opened 2018-01-24
Initial wip

nhnt11:


public/main.js
```diff
@@ -0,0 +1,55 @@
+function doXHR(aURL, aBodyObj, aAlertText, aDebug=true) {
+  return new Promise((resolve) => {
+    let xhr = new XMLHttpRequest();
```
pdehaan: Why XMLHttpRequest and not the promise-based [`fetch()` API](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API/Using_Fetch#Uploading_JSON_data)?

nhnt11: Filed issue#18

server.js
```diff
@@ -3,22 +3,90 @@
...
 
+// The SMTP transport object. This is initialized to a nodemailer transport
+// object while reading SMTP credentials, or to a dummy function in debug mode.
+var gTransporter;
+
+app.use(sessions({
+  cookieName: "session",
+  secret: process.env.COOKIE_SECRET,
+  duration: 15 * 60 * 1000, // 15 minutes
+  activeDuration: 5 * 60 * 100, // 5 minutes
```
pdehaan: Should this be `* 100` or `* 1000`?

server.js
```diff
@@ -31,12 +100,17 @@ app.post("/user/reset", function(req, res) {
...
+app.post("/user/list", function(req, res) {
+  res.json({ emails: Array.from(gEmails) });
+});
 
 app.post("/user/breached", function(req, res) {
   let emails = req.body.emails;
   let response = [];
 
+  // Send notification email to the intersection of the set of
+  // emails in the requet and the set of registered emails.
```
pdehaan: typo: "request"

groovecoder:
This is looking good enough to me to merge into master. Then I can branch off of master to work an another issue.

server.js
```diff
@@ -31,12 +100,17 @@ app.post("/user/reset", function(req, res) {
   res.json({ info: "user list cleared" });
 });
 
-let gTransporter;
+// This exists only right now for development purposes.
```
pdehaan: nit: Not sure if we want to add a "// TODO" or some follow up bug to remove this route in the future. Having this endpoint sneak onto production makes me a bit uneasy.

server.js
```diff
@@ -57,8 +131,67 @@ app.post("/user/breached", function(req, res) {
...
+
+app.get('/oauth/redirect', function (req, res) {
+  if (!req.session.state) {
+    res.send("Who are you?");
+    return;
+  }
+  FxAOAuth.code.getToken(req.originalUrl, { state: req.session.state })
+    .catch(function (err) {
+      res.send(err);
+    })
```
pdehaan: I think I may need somebody explain this bit to me. IIUC, the `.catch()` statement in the middle of this promise chain will swallow any error that may occur during `FxAOAuth.code.getToken(...)`, but then the `.catch()` will send an error to the user and then continue onto the `.then(function(user) {...})` code path (where `user` will probably be `null|undefined`).

Should the `.catch()` be at the end (after the `.then()`)? Do we need to add more error handling in the case of `popsicle.get(...).then(data => {})` generates an error?

server.js
nhnt11: This will not sneak into production. As soon as we add persistent storage we won't need this, and removing this API will be part of the process of removing the gEmails Set.

server.js
nhnt11: Yeah, this is wonky. The catch you're bothered by (and rightly so) was an artifact of me trying to debug why getToken wasn't working (see https://github.com/mozilla/fxa-oauth-server/issues/513). I'll move the catch to after the then() block, but I'll also file an issue about adding real error logging.

server.js
```diff
@@ -57,8 +131,67 @@ app.post("/user/breached", function(req, res) {
...
+});
+
+app.get('/oauth/redirect', function (req, res) {
+  if (!req.session.state) {
+    res.send("Who are you?");
+    return;
+  }
+  FxAOAuth.code.getToken(req.originalUrl, { state: req.session.state })
+    .then(function (user) {
+      popsicle.get({
```
pdehaan: The `.catch()` is now in the better place, but I think this still has a weird side effect of any errors generated during the `popsicle.get(...).then(...)` won't throw up to the `.catch()` statement. Not sure if the right answer is to do a `return popsicle.get()` or how best to flatten the promises (or rethrow, or try and use `async`/`await`.

nhnt11: FWIW, I'm investigating using async/await. I want to refactor this stuff in #16.

PR #29, opened 2018-02-07
Modularize code.

nhnt11:
Modularized all the things. Let me know what you think. I'll leave inline comments to try and clearly point out the changes I've made.


app-constants.js
```diff
@@ -0,0 +1,33 @@
...
+  "OAUTH_PROFILE_URI",
+  "OAUTH_CLIENT_ID",
+  "OAUTH_CLIENT_SECRET",
+];
+
+var AppConstants = {
+  init() {
+    for (let v of kEnvironmentVariables) {
+      if (process.env[v] === undefined) {
+        throw "Required environment variable was not set: " + v;
```
pdehaan: nit: We should throw an error:

```js
throw new Error(`Required environment variable was not set: ${v}`);
```

Also, should `DUMMY_DEBUG_SMTP` be required, or can we safely just assume the default is `false` unless specified?

email-utils.js
```diff
@@ -0,0 +1,83 @@
...
+        from: "\"Firefox Breach Alerts\" <" + kSMTPUsername + ">",
+        to: aRecipient,
+        subject: aSubject,
+        text: aBody,
+      };
+
+      gTransporter.sendMail(mailOptions, (error, info) => {
+        if (error) {
+          console.log("error! " + error);
+          reject("WHAT");
```
pdehaan: I'm not sure what `error` is in this `sendMail()` context, but should we just be rejecting with the `error` object?

```js
reject(error);
```

app-constants.js
```diff
@@ -0,0 +1,33 @@
...
+  "SMTP_USERNAME",
+  "SMTP_PASSWORD",
+  "OAUTH_AUTHORIZATION_URI",
+  "OAUTH_TOKEN_URI",
+  "OAUTH_PROFILE_URI",
+  "OAUTH_CLIENT_ID",
+  "OAUTH_CLIENT_SECRET",
+];
+
+var AppConstants = {
```
pdehaan: nit: we should probably use `const`

email-utils.js
```diff
@@ -0,0 +1,83 @@
+"use strict";
+
+const AppConstants = require("./app-constants");
+
+const nodemailer = require("nodemailer");
+
+// This is set later when reading SMTP credentials from the environment.
+// This exists as a variable so we can use it in the from header of emails.
+var kSMTPUsername;
```
pdehaan: `let`?

email-utils.js
nhnt11: lol, this wasn't meant to get into the PR. I'll fix it. Thanks!

email-utils.js
```diff
@@ -0,0 +1,83 @@
...
+
+// The SMTP transport object. This is initialized to a nodemailer transport
+// object while reading SMTP credentials, or to a dummy function in debug mode.
+var gTransporter;
```
pdehaan: `let`

email-utils.js
```diff
@@ -0,0 +1,83 @@
...
+
+var EmailUtils = {
```
pdehaan: `const`

app-constants.js
nhnt11: Yeah I think we can make `DUMMY_DEBUG_SMTP` a special case actually. I expect to remove it at some point.

email-utils.js
```diff
@@ -0,0 +1,83 @@
...
+    console.log("Attempting to get SMTP credentials from environment...");
+    kSMTPUsername = AppConstants.SMTP_USERNAME;
+    let password = AppConstants.SMTP_PASSWORD;
+    let host = AppConstants.SMTP_HOST;
+    let port = AppConstants.SMTP_PORT;
+    if (!(kSMTPUsername && password && host && port)) {
+      return Promise.reject("SMTP credentials could not be read from the environment");
+    }
+
+    // TODO: stop hardcoding this stuff.
```
pdehaan: I think we can remove this comment, since we're grabbing these values from env vars now. ❤️

routes/oauth.js
```diff
@@ -0,0 +1,65 @@
...
+});
+
+router.get("/redirect", function (req, res) {
+  if (!req.session.state) {
+    res.send("Who are you?");
+    return;
+  }
+  FxAOAuth.code.getToken(req.originalUrl, { state: req.session.state })
+    .then(function (user) {
+      popsicle.get({
```
pdehaan: I think this should be `return popsicle.get({...})` so any errors that happen in this nested promise chain will bubble up to the `.catch((err) => {...});` below.

routes/user.js
```diff
@@ -0,0 +1,96 @@
...
+const gEmails = require("../subscribers");
+
+// We send verification emails to addresses that want to subscribe.
+// These addresses are temporarily stored here, mapped to the unique
+// string token that is used for verification.
+var gUnverifiedEmails = new Map();
+
+router.post("/add", function(req, res) {
+  // TODO: use a hash of the email address instead of a random string.
+  let state = crypto.randomBytes(40).toString("hex");
```
pdehaan: This is also duplicated in https://github.com/mozilla/blurts-server/pull/29/files#diff-4e57e667bb9833a8936a9e4a87bb539eR34

```diff
router.get("/init", function(req, res) {
  // Set a random state string in a cookie so that we can verify
  // the user when they're redirected back to us after auth.
+ let state = crypto.randomBytes(40).toString("hex");
  let uri = FxAOAuth.code.getUri({state});
  req.session.state = state;
  res.redirect(uri);
});
```

Not sure if they're related, and we should extract into some helper function, or if it's simply two completely separate places we need random strings.

pdehaan:
:100: Overall, I think this is a great refactor! Code is much easier to read and understand.

Since this will conflict greatly with the ESLint PR, feel free to merge this first after @groovecoder's review and I can rebase my ESLint branch onto this PR after it lands.

email-utils.js
groovecoder: Can remove this `TODO` now that we're not hard-coding anymore! 🎉

routes/oauth.js
```diff
@@ -0,0 +1,65 @@
...
+  redirectUri: AppConstants.SERVER_URL + "/oauth/redirect",
+  scopes: ["profile:email"],
+});
+
+router.get("/init", function(req, res) {
+  // Set a random state string in a cookie so that we can verify
+  // the user when they're redirected back to us after auth.
+  let state = crypto.randomBytes(40).toString("hex");
+  let uri = FxAOAuth.code.getUri({state});
+  req.session.state = state;
```
groovecoder: We should make sure this session state cookie is set Secure and HTTP-only.

Also, does the server side of this state only exist in the node process memory? If so, that won't work when we scale this out to multiple processes/servers.

routes/user.js
```diff
@@ -0,0 +1,96 @@
...
+const EmailUtils = require("../email-utils");
+const gEmails = require("../subscribers");
+
+// We send verification emails to addresses that want to subscribe.
+// These addresses are temporarily stored here, mapped to the unique
+// string token that is used for verification.
+var gUnverifiedEmails = new Map();
+
+router.post("/add", function(req, res) {
+  // TODO: use a hash of the email address instead of a random string.
```
groovecoder: Specifically, HIBP email hashsets are SHA1 hashes.

groovecoder: Wait, if this is the capability token that will authorize/confirm that the user owns the email address, it needs to be a salted hash of the email, or keep it as a random string.

Otherwise any attacker can put an email address in and then visit the `/verify?state={email-hash}` link to verify the email address.

routes/user.js
```diff
@@ -0,0 +1,96 @@
...
+  gEmails.delete(req.body.email);
+  res.json({ email: req.body.email, info: "removed user" });
+});
+
+router.post("/reset", function(req, res) {
+  gEmails.clear();
+  res.json({ info: "user list cleared" });
+});
+
+// This exists only right now for development purposes.
```
groovecoder: We should gate this code on a `process.env` check for something like `DEBUG` to help ensure we don't accidentally ship things like this into production.

routes/user.js
nhnt11: Right, I definitely did not mean a simple unsalted hash. I just don't want the verification token to have ONLY a random component, for philosophical reasons (clashes are unimaginably unlikely, yet possible) as well as paranoia about people finding entropy-related exploits.

routes/oauth.js
nhnt11: Yeah, now that this refactor is done I'm going to start working on storage; this state as well as the email verification token should use a database.

routes/user.js
nhnt11: Filed #33

routes/oauth.js
nhnt11: Filed #32

PR #38, opened 2018-02-08
start get_hashsets.js

groovecoder:


get_hashsets.js
```diff
@@ -0,0 +1,47 @@
+"use strict";
+
+require("dotenv").load();
+
+const fs = require("fs");
+const path = require("path");
+const request = require("request");
+
+const HIBP_API_ROOT = process.env.HIBP_API_ROOT;
+const HIBP_API_TOKEN = process.env.HIBP_API_TOKEN || '';
```
pdehaan: Shouldnt these be imported from "./app-constants.js"?
https://github.com/mozilla/blurts-server/blob/c94d8403a234c3714d46c826fd1b02304ce3e31b/email-utils.js#L3

get_hashsets.js
```diff
@@ -0,0 +1,47 @@
+"use strict";
+
+require("dotenv").load();
+
+const fs = require("fs");
+const path = require("path");
```
pdehaan: Is this used?

get_hashsets.js
```diff
@@ -0,0 +1,47 @@
...
+const request = require("request");
+
+const HIBP_API_ROOT = process.env.HIBP_API_ROOT;
+const HIBP_API_TOKEN = process.env.HIBP_API_TOKEN || '';
+const HIBP_AUTH = 'Bearer ' + new Buffer(HIBP_API_TOKEN);
+
+
+function getBreachHashset(breach) {
+  if (breach.IsActive && breach.IsVerified && breach.DataClasses.indexOf("Email addresses") > -1) {
```
pdehaan: If `breach.DataClasses` is an array, you can use [`Array.prototype.includes()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/includes) instead of checking `.indexOf() > -1`.

get_hashsets.js
```diff
@@ -0,0 +1,47 @@
...
+    console.log("Active, verified breach with email addresses: " + breach.Name);
+    const url = HIBP_API_ROOT + '/enterprisesubscriber/hashset/' + breach.Name;
+    console.log("url: " + url);
+
+    const headers = {
+      'User-Agent': 'request',
```
pdehaan: Instead of generic "request" User Agent, do you want to do use the package.json `name` and `version`?
For example:

```js
const pkg = require("./package.json");
// ...
const headers = {
  'User-Agent': `${pkg.name}@${pkg.version}`, // "blurts-server@1.0.0"
  'Authorization': HIBP_AUTH
};
```

groovecoder:
Good suggestions @pdehaan. I cleaned them up after a rebase so this should be ready for review and merge by @nhnt11 when he wakes up. :)

get_hashsets.js
```diff
@@ -0,0 +1,47 @@
...
+
+
+function getBreachHashset(breach) {
+  if (breach.IsActive && breach.IsVerified && breach.DataClasses.includes("Email addresses")) {
+    console.log("Active, verified breach with email addresses: " + breach.Name);
+    const url = AppConstants.HIBP_API_ROOT + '/enterprisesubscriber/hashset/' + breach.Name;
+    console.log("url: " + url);
+
+    const headers = {
+      'User-Agent': `${pkg.name}@{pkg.version}`,
```
pdehaan: nit: I think we need `${pkg.version}` (with the leading `$`).

get_hashsets.js
```diff
@@ -0,0 +1,47 @@
...
+  let breachesJSON = JSON.parse(body);
+  for (let breach of breachesJSON) {
+    getBreachHashset(breach);
+  }
+}
+
+const breachesRequestObject = {
+  url: AppConstants.HIBP_API_ROOT + '/breaches',
+  headers: {
+    'User-Agent': 'request'
```
pdehaan: Not sure if we want to change the User-Agent here to match the value from L21 above or not. ¯\\\_(ツ)\_/¯

get_hashsets.js
```diff
@@ -0,0 +1,47 @@
...
+    const headers = {
+      'User-Agent': `${pkg.name}@${pkg.version}`,
+      'Authorization': HIBP_AUTH
+    }
+    const hashsetRequestObject = {
+      url: url,
+      headers: headers
+    };
+
+    request(hashsetRequestObject).pipe(fs.createWriteStream('breach_hashsets' + path.sep + breach.Name + '.zip'));
```
pdehaan: Not sure which is more readable (`path.sep` vs `path.join()`), so I'll let you two debate it:

```js
fs.createWriteStream('breach_hashsets' + path.sep + breach.Name + '.zip')
```

vs

```js
fs.createWriteStream(path.join('breach_hashsets', `${breach.Name}.zip`)
```

get_hashsets.js
```diff
@@ -0,0 +1,47 @@
...
+}
+
+const breachesRequestObject = {
+  url: AppConstants.HIBP_API_ROOT + '/breaches',
+  headers: {
+    'User-Agent': 'request'
+  }
+}
+
+request(breachesRequestObject, handleBreachesResponse);
```
pdehaan: Actually, I'm not 100% sure how I'd use this file (especially since I'm not privileged enough to have an `HIBP_API_TOKEN`). If we're just calling this file directly from the CLI (a la `$ node get_hashsets`), we should document that in the README.md file, and/or optionally make it get called automatically via `npm start` or some other method so we know the expected hash sets are there before starting the server.

package.json
```diff
@@ -13,7 +13,8 @@
     "dotenv": "^5.0.0",
     "express": "^4.16.2",
     "nodemailer": "^4.4.0",
-    "popsicle": "^9.2.0"
+    "popsicle": "^9.2.0",
```
nhnt11: NB: Let's add an eslint rule to enforce trailing commas in order to preserve blame. I'll file an issue soon.

tests/fixtures/make_sha1_hashes.js
```diff
@@ -0,0 +1,14 @@
+const crypto = require('crypto');
```
nhnt11: "use strict"

get_hashsets.js
nhnt11: path.join() hands down. More than readability, let's not assume that the join() operation is a simple string concatenation. join() provides platform abstraction for us and also normalizes the resulting path string - let's use it.

get_hashsets.js
nhnt11: I agree! Let's match it. Let's define as a const, or even add it to AppConstants. :)

package.json
pdehaan: Well _actually_... No trailing commas here, because it's a JSON file and I'm sure we'd break some stuff with invalid JSON markup in a package.json, but I see your point (re: #40).

nhnt11: Oh no, you're right! Thanks! *shakes fist at the JSON gods*

tests/fixtures/make_sha1_hashes.js
nhnt11: We can add an eslint rule for this as well!

package.json
pdehaan: But _actually_, now that my eyes have seen this, I'll disagree with the use of _both_ **popsicle** _and_ **request** modules, since they both seem to roughly do the same thing. We should probably use one or the other (or [**isomorphic-fetch**](https://www.npmjs.com/package/isomorphic-fetch) or [**got**](https://www.npmjs.com/package/got) or ...)

nhnt11: Yeah. I will file an issue to get rid of our popsicle usage.

get_hashsets.js
pdehaan: +1 to just a vanilla ```const USER_AGENT = `${pkg.name}@${pkg.version}`;``` at the top of this file. Extracting out into AppConstants seems a bit overkill for this usage.

get_hashsets.js
```diff
@@ -0,0 +1,47 @@
...
+function getBreachHashset(breach) {
+  if (breach.IsActive && breach.IsVerified && breach.DataClasses.includes("Email addresses")) {
+    console.log("Active, verified breach with email addresses: " + breach.Name);
+    const url = AppConstants.HIBP_API_ROOT + '/enterprisesubscriber/hashset/' + breach.Name;
+    console.log("url: " + url);
+
+    const headers = {
+      'User-Agent': `${pkg.name}@${pkg.version}`,
+      'Authorization': HIBP_AUTH
+    }
```
pdehaan: nit: missing semicolon

get_hashsets.js
```diff
@@ -0,0 +1,47 @@
...
+    getBreachHashset(breach);
+  }
+}
+
+const breachesRequestObject = {
+  url: AppConstants.HIBP_API_ROOT + '/breaches',
+  headers: {
+    'User-Agent': 'request'
+  }
+}
```
pdehaan: nit: missing semicolon

tests/fixtures/make_sha1_hashes.js
pdehaan: Ref: #41

get_hashsets.js
nhnt11: Beautiful, another eslint opportunity has been revealed

package.json
pdehaan: Ref: #42

get_hashsets.js
nhnt11: Wait, the `semi` rule should be taking care of this, never mind.

get_hashsets.js
```diff
@@ -0,0 +1,47 @@
...
+
+
+function getBreachHashset(breach) {
+  if (breach.IsActive && breach.IsVerified && breach.DataClasses.includes("Email addresses")) {
+    console.log("Active, verified breach with email addresses: " + breach.Name);
+    const url = AppConstants.HIBP_API_ROOT + '/enterprisesubscriber/hashset/' + breach.Name;
+    console.log("url: " + url);
+
+    const headers = {
+      'User-Agent': `${pkg.name}@${pkg.version}`,
```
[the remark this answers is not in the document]
pdehaan: Ref: #43

get_hashsets.js
```diff
@@ -0,0 +1,47 @@
...
+
+const AppConstants = require("./app-constants").init();
+
+const fs = require("fs");
+const path = require("path");
+const request = require("request");
+
+const pkg = require("./package.json");
+
+const HIBP_AUTH = 'Bearer ' + new Buffer(AppConstants.HIBP_API_TOKEN);
```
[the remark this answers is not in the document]
pdehaan: Only because I love you both:

> **/Users/pdehaan/dev/github/mozilla/blurts-server/get_hashsets.js**
>  11:31   **error**    `new Buffer()` was deprecated since [Node] v6. Use `Buffer.alloc()` or `Buffer.from()` (use 'https://www.npmjs.com/package/safe-buffer' for '<4.5.0') instead  node/no-deprecated-api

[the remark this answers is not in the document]
nhnt11: Lol, nice, guess I was wrong about "less important". 👍🏼

get_hashsets.js
pdehaan: ESLint _was_ catching this. The opportunity for CI has been revealed (and there should already be an open PR for it).

pdehaan:
@groovecoder We landed some new-and-improved ESLint rules in #44, which should auto-fix most/all of the formatting issues mentioned above with a quick `npm run lint -- --fix` (after a rebasin').

🍻

.gitignore
```diff
@@ -1,2 +1,5 @@
 node_modules
 .env
+.DS_Store
+README.html
```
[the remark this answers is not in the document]
groovecoder: convenience for when I check my `README.md` updates & rendering

get_hashsets.js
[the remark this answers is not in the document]
groovecoder: `Buffer` is only left-over from when I was base64-encoding the token. It's not needed anymore.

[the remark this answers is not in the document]
groovecoder: Changed to template string with substitution like the rest of the constants.

groovecoder:
Addressed all comments except I'm keeping the `const` declarations since those identifiers are not meant to be re-assigned in their scope. If we need to re-assign them in the future we can change them to `let` no problem.

.gitignore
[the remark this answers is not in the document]
pdehaan: Cool, I was reading about this the other day: https://help.github.com/articles/ignoring-files/#explicit-repository-excludes

Looks like you can possibly just add your <kbd>README.html</kbd> to your own local <kbd>.git/info/exclude</kbd> blocklist since it is a testing artifact on your local machine only.

get_hashsets.js
```diff
@@ -0,0 +1,52 @@
...
+    const headers = {
+      "User-Agent": USER_AGENT,
+      "Authorization": HIBP_AUTH
+    };
+    const hashsetRequestObject = {
+      url: url,
+      headers: headers
+    };
+
+    request(hashsetRequestObject).pipe(fs.createWriteStream(path.join("breach_hashsets", `${breach.Name}.zip`)));
```
pdehaan: nit: this `"breach_hashsets"` in the path should now use the `BREACH_HASHSET_DIR` const for improved greatness.

Which also makes me question whether that should be added to the app-constants.js file, in the case where we may need to access that directory in other places in the future when we want to _consume_ those files. But I'm also fine with leaving that tweak until the time we need to specify the path twice in two different files.

.gitignore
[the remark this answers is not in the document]
nhnt11: Yes, I was going to suggest that as well: please include these in your global ignore list.

get_hashsets.js
```diff
@@ -0,0 +1,52 @@
...
+
+const pkg = require("./package.json");
+
+const HIBP_AUTH = `Bearer ${AppConstants.HIBP_API_TOKEN}`;
+const USER_AGENT = `${pkg.name}@${pkg.version}`;
+const BREACH_HASHSET_DIR = "breach_hashsets";
+
+
+function getBreachHashset(breach) {
+  if (breach.IsActive && breach.IsVerified && breach.DataClasses.includes("Email addresses")) {
```
nhnt11: Not urgent, but a comment documenting the structure of the `breach` object would be nice!

get_hashsets.js
```diff
@@ -0,0 +1,52 @@
...
+    console.log("Active, verified breach with email addresses: " + breach.Name);
+    const url = AppConstants.HIBP_API_ROOT + "/enterprisesubscriber/hashset/" + breach.Name;
+    console.log("url: " + url);
+
+    const headers = {
+      "User-Agent": USER_AGENT,
+      "Authorization": HIBP_AUTH
```
nhnt11: nit: Trailing comma

get_hashsets.js
```diff
@@ -0,0 +1,52 @@
...
+    };
+    const hashsetRequestObject = {
+      url: url,
+      headers: headers
```
nhnt11: nit: Trailing comma

get_hashsets.js
```diff
@@ -0,0 +1,52 @@
...
+
+const HIBP_AUTH = `Bearer ${AppConstants.HIBP_API_TOKEN}`;
+const USER_AGENT = `${pkg.name}@${pkg.version}`;
+const BREACH_HASHSET_DIR = "breach_hashsets";
+
+
+function getBreachHashset(breach) {
+  if (breach.IsActive && breach.IsVerified && breach.DataClasses.includes("Email addresses")) {
+    console.log("Active, verified breach with email addresses: " + breach.Name);
+    const url = AppConstants.HIBP_API_ROOT + "/enterprisesubscriber/hashset/" + breach.Name;
```
nhnt11: WARNING: begin personal opinion rant that is not important for making progress on this repo:

I personally hold the opinion that const provides no real value and only adds decision pressure while writing code. It's also (imo) confusing, since it only prevents reassignment - which means you need to be very aware of exactly what operators count as reassignment in JS to really understand the contract that const enforces. A great post about this: http://web.archive.org/web/20160315114714/http://blog.getify.com/constantly-confusing-const/. But if you're comfortable with consts, I won't complain further.

get_hashsets.js
```diff
@@ -0,0 +1,52 @@
...
+  let breachesJSON = JSON.parse(body);
+  for (let breach of breachesJSON) {
+    getBreachHashset(breach);
+  }
+}
+
+const breachesRequestObject = {
+  url: AppConstants.HIBP_API_ROOT + "/breaches",
+  headers: {
+    "User-Agent": USER_AGENT
```
nhnt11: nit: Trailing comma

get_hashsets.js
groovecoder: `const` is at least [a signal-of-intent by the developer that identifier won't/shouldn't be re-assigned](https://medium.com/javascript-scene/javascript-es6-var-let-or-const-ba58b8dcde75). I see that only makes the value immutable when assigning primitives, but meh - Kyle's post reads more like one JS expert's opinion, not a technical principle. 

I'm okay either way - whatever we want, let's just make it part of an `eslint` config so we're consistent across the code-base.

groovecoder:
I didn't change the `const` declarations yet. I don't have a strong opinion except that whatever we pick should be able to be enforced across the code-base (ideally via `eslint`) so I'd rather file an issue for that.

Can we merge this without it? I've had to rebase & resolve `package-lock.json` and other conflicts 3 times now. 😢 (Also is there a way to make `package-lock.json` easier?)

pdehaan:
> Can we merge this without it?

I've cheekily submitted #46 ("`var` vs `let` vs `const`, oh my!") which we can use to further discuss the rationale for using `let` vs `const`, which I believe is sufficient to unblock this PR.

Re: "Is there a way to make `package-lock.json` easier?",
What's the problem with it currently? Is it purely a merge conflict hell? Or you just don't like the file in general?
If it's merge conflict issues, I usually just end up rage typing `rm -rf node_modules package-lock.json` and then doing a fresh `npm install` and letting that be the source of new truth (assuming I've rebased against latest master).
If you just hate the file in general, we can probably just add [something to a local ".npmrc" file](https://codeburst.io/disabling-package-lock-json-6be662f5b97d) which will disable the lock file.

PR #39, opened 2018-02-09
 fix #24: add Travis and Coveralls 

### LINKED ISSUE #24: Add Circle-CI or Travis-CI
We probably also want to add Circle-CI or Travis-CI git hooks so we can run linting/tests/[**nsp**](http://npm.im/nsp) on a per-PR basis.

groovecoder:
This sets up a lot of tooling, but I found this all to be a very nice testing stack on my [`blok`](https://github.com/mozilla/blok) repository.

* `.travis.yml` to make Travis run builds
* [`tape`](https://www.npmjs.com/package/tape) as the test harness
* [`faucet`](https://www.npmjs.com/package/faucet) to pretty-print the test output
* [`nyc`](https://www.npmjs.com/package/nyc) command-line interface for [istanbul](https://www.npmjs.com/package/istanbul) JS code coverage
* [`coveralls`](https://www.npmjs.com/package/coveralls) to publish the test code coverage to coveralls.io

It adds a `test` script to `package.json` that combines all of the above in a way that produces ...

* [Travis CI report](https://travis-ci.org/mozilla/blurts-server/builds/339294486?utm_source=github_status&utm_medium=notification) 
* [Coveralls code coverage report](https://coveralls.io/builds/15446378)

... for every pull request & commit.


.travis.yml
```diff
@@ -0,0 +1,3 @@
+language: node_js
+node_js:
+    - "9.5"
```
pdehaan: The README says we require Node 8+. Should we be building on Node 8 here, or update the README to match this value?

package.json
```diff
@@ -16,18 +16,26 @@
     "popsicle": "^9.2.0"
   },
   "devDependencies": {
-    "eslint": "^4.17.0"
+    "coveralls": "^3.0.0",
+    "eslint": "^4.17.0",
+    "faucet": "0.0.1",
```
pdehaan: set this to `^0.0.1` (semver range)?

package.json
```diff
@@ -16,18 +16,26 @@
...
   "private": true,
   "repository": {
     "type": "git",
     "url": "git+https://github.com/mozilla/blurts-server.git"
   },
   "scripts": {
     "lint": "eslint .",
-    "start": "node server.js"
+    "start": "node server.js",
+    "test": "eslint . && nyc tape tests/**/*.js | faucet && nyc report --reporter=text-lcov | coveralls"
```
pdehaan: Instead of calling ESLint here, we should instead call `npm run lint` so we don't need to update linting configs in 2 places.
Also, not sure if we want to scrap linting here and just use a `"pretest": "npm run lint"` hook.

tests/test-subscribers.js
```diff
@@ -0,0 +1,8 @@
+const test = require("tape");
+
+const gEmails = require("../subscribers.js");
```
pdehaan: nit: we should drop the ".js" extension here, for consistency.

package.json
groovecoder: When I used `npm run lint` here, it messed up Travis. `npm run lint` on Travis made `eslint` scan `node_modules/*` too, which threw all kinds of errors. I searched around and found a few bug reports, but didn't find a fix.

.travis.yml
groovecoder: I'll change it to test on both "node" (9.x) and "lts" (8.x) to run on the major versions. I don't think we've written enough code to know if we're going to use a library or something that only works in one version or the other.

I'm personally using 9.3 so I'll have the README reflect 9.x until we run into something that tells us the exact version we need to stay on.

package.json
groovecoder: Huh. This is how `npm install` added it. I'll fix it.

groovecoder:
I'm still not using `npm run lint` inside of `test` because it breaks Travis CI. And fixing Travis CI was turning into a bigger hassle than I wanted. I filed [a tooling bug for this](https://github.com/mozilla/blurts-server/issues/45).

groovecoder:
Rebased and fixed the new `eslint` rules.

package.json
pdehaan: > Huh. This is how npm install added it. I'll fix it.

I have $1 internet buck saying your **~/.npmrc** file has `save-exact=true`, which would drop any semver ranges.

package.json
pdehaan: This ~~interests~~ fascinates me, and I wonder if it's related to you adding a **.npmignore** file. Although I can't find any docs stating that if ESLint cannot find an **.eslintignore** file, it will revert back to using **.npmignore** _then_ **.gitignore**.

We _could_ add our own **.eslintignore** file with node_modules in it, but that should be the default behavior, which is why this is so bizarre. But I'm fine to land this whenever and fix that later. I guess the quick test would maybe be to try adding "node_modules" to the **.npmignore** file.

pdehaan:
Anyways... I love this. I don't have commit/merge access, but I think we should land this ASAP so we know the code will lint everything on each PR and we can tweak it later if needed.

package.json
groovecoder: You owe me an internet buck ...

```
groovetop:blurts-server lcrouch$ cat ~/.npmrc
//registry.npmjs.org/:_authToken = [redacted]
groovetop:blurts-server lcrouch$
```

pdehaan: Double or nothing on `$ npm config list`?
Or possibly `npm config ls -l | grep save` may show if you've overridden your `save-prefix` from `^`.

.travis.yml
```diff
@@ -0,0 +1,4 @@
+language: node_js
+node_js:
+    - "node"
```
nhnt11: nit: 2-space indent?

PR #47, opened 2018-02-13
Re-enable prefer-const and a few minor tweaks

pdehaan:


get_hashsets.js
```diff
@@ -9,7 +9,7 @@ const request = require("request");
 const pkg = require("./package.json");
 
 const HIBP_AUTH = `Bearer ${AppConstants.HIBP_API_TOKEN}`;
-const USER_AGENT = `${pkg.name}@${pkg.version}`;
+const HIBP_USER_AGENT = `${pkg.name}/${pkg.version}`;
```
pdehaan: Renamed for consistency (since it is HIPB-specific), and tweaked to use a `/` separator instead of `@` (since a backslash seems to be more of a standard).

get_hashsets.js
```diff
@@ -22,38 +22,53 @@ function getBreachHashset(breach) {
...
-  headers: {
-    "User-Agent": USER_AGENT,
-  },
-};
+  try {
+    const breachesJSON = JSON.parse(body);
+
+    if (!fs.existsSync(BREACH_HASHSET_DIR)) {
+      fs.mkdirSync(BREACH_HASHSET_DIR);
+    }
```
pdehaan: Moved most of the logic out of the root level and into the functions, closer to where they were used. Also only create the hashset directory if we didn't get an error and the JSON response parses. I figure a bunch of this may change slightly as we switch to **got** anyways.

tests/fixtures/make_sha1_hashes.js
```diff
@@ -3,14 +3,14 @@
 const crypto = require("crypto");
 const stdin = process.openStdin();
 
+const PROMPT = "\nEnter an email address to get the SHA1 hash as it would appear in a HIBP hashset file:";
```
pdehaan: Moved the message into a separate variable since it was duplicated in 2 places.
Not sure if this is used directly by [future] tests (since it is IO blocking), or if we should make a bin/ folder at the root level with all our scripts (like this one and /get_hashsets.js).

Issue #30, opened 2018-02-07
Resolve code style inconsistencies


nhnt11:
I want to do an audit of all the code to clean up stuff like:
- Package imports should be ordered nicely
- Use async/await instead of Promises
- Make variable naming consistent (especially for function arguments - prefix with "a" e.g. aArg1, aArg2)
- Sensible whitespace

groovecoder:
I think we're covering most of this in our `eslint` usage?

pdehaan:
I think most of it is good in ESLint, but not sure if there are any rules/plugins we can use to detect Promise usage...

Here's a really crude list of a few places we're [probably/maybe] using Promises (if that's a problem):

```sh
$ git rev-parse --short HEAD # ffa1706
$ git grep "\.then(" | grep -v "/vendor/"

public/js/test.js:    .then(function() {
scripts/get-hashsets.js:models.sequelize.sync().then(()=>{
server.js:EmailUtils.init().then(() => {
tests/test-user.js:  user.validate().then(user => {
tests/test-user.js:  user.validate().then(user=>{
tests/test-user.js:  user = models.User.findOne({ where: { email: VALID_EMAIL } }).then(user=>{
```

pdehaan:
Another random thing I found the other day was that sometimes we use `e` or `err` or `error` when referring to errors. I couldn't find any ESLint rule to enforce that though. But my quick manual audit came up with the following:

```sh
$ git grep -n "(e)" | grep -v "/vendor/"

db/utils.js:24:    } catch(e) {
db/utils.js:25:      console.error(e);
public/js/test.js:54:function enableBtnIfEmailValid(e) {
routes/user.js:43:  } catch (e) {
routes/user.js:44:    console.error(e);
scripts/notify-subscribers-of-breach.js:34:    } catch (e) {
scripts/notify-subscribers-of-breach.js:35:      console.log(e);
```

```sh
$ git grep -n "(err)" | grep -v "/vendor/"

routes/oauth.js:65:  } catch (err) {
routes/oauth.js:66:    console.error(err);
routes/oauth.js:67:    res.send(err);
```

```sh
$ git grep -n "(error)" | grep -v "/vendor/"

email-utils.js:62:        if (error) {
email-utils.js:63:          console.error(error);
email-utils.js:65:          reject(error);
email-utils.js:88:        if (error) {
email-utils.js:89:          reject(error);
hibp.js:39:    } catch (error) {
hibp.js:40:      console.error(error);
[...]

PR #48, opened 2018-02-13
Start database interface

nhnt11:
This implements a database interface (for Postgres) that is up to parity with the previous Set implementation. It's still rough around the edges but it works (I quickly tested out all the features in test.html).

I want to do a self-review of this and iron out all the wrinkles, but also wanted to get this PR up asap so that y'all can take a look.

You'll need to run a Postgres server to test this out, and set env vars to tell the db code how to takl to it (see changes to .env-dist).

There's a pretty basic db-scripts.js that has a function to reset the users table as well as a quick smoke test (that currently relies on console.logs, and doesn't do any REAL testing - I'm working on it).


subscribers.js
```diff
@@ -1,7 +1,39 @@
...
-const gEmails = new Set();
+const Subscribers = {
+  async addUser(aEmail) {
+    if (await this.getUser(aEmail)) {
+      console.log("already added");
+      return;
+    }
+
+    await db.doQuery(
+      `INSERT INTO users ( email ) VALUES ( '${aEmail}' );`);
```
nhnt11: We'll want to use prepared statements. I'm going to look into this.

subscribers.js
```diff
@@ -1,7 +1,39 @@
...
 
+const db = require("./db");
+
 // Set of all registered emails.
 // TODO: Implement a real persistent storage solution.
-const gEmails = new Set();
+const Subscribers = {
+  async addUser(aEmail) {
+    if (await this.getUser(aEmail)) {
+      console.log("already added");
```
nhnt11: We'll want to handle this better than console.log.

subscribers.js
```diff
@@ -1,7 +1,39 @@
...
+    }
+
+    await db.doQuery(
+      `INSERT INTO users ( email ) VALUES ( '${aEmail}' );`);
+    console.log("added!");
+  },
+  async deleteUser(aEmail) {
+    let user =  await this.getUser(aEmail);
+    if (!user) {
+      console.log("user doesn't exist");
```
nhnt11: Handle this better

subscribers.js
```diff
@@ -1,7 +1,39 @@
...
+    }
+
+     await db.doQuery(
+      `DELETE FROM users WHERE id = '${user.id}';`);
+     console.log("deleted!");
+  },
+  async getUser(aEmail) {
+    let rows = await db.doQuery(
+      `SELECT * FROM users WHERE email = '${aEmail}';`);
+    return rows[0];
```
nhnt11: Maybe actually check if this exists, and also check number of rows? Currently leaves it to the caller to check whether the user actually existed. We can do this better.

routes/oauth.js
```diff
@@ -54,8 +54,9 @@ router.get("/redirect", (req, res) => {
         },
       }).then((data) => {
         const email = JSON.parse(data.body).email;
-        gEmails.add(email);
-        res.send(`Registered ${email} for breach alerts. You may now close this window/tab.`);
+        Subscribers.addUser(email).then(() => {
```
nhnt11: Need to file an issue to replace then() with async/await as much as possible. This particular case is getting very nested and definitely needs to be refactored to chain the thens instead of nesting.

db.js
```diff
@@ -0,0 +1,26 @@
...
+      database: AppConstants.DATABASE_NAME,
+      host: AppConstants.DATABASE_HOST,
+      port: AppConstants.DATABASE_PORT,
+    });
+    let result;
+    try {
+      await client.connect();
+      result = await client.query(aQueryStr);
+    } catch(e) {
+      console.log(e);
```
nhnt11: Better error handling needed here.

db-scripts.js
```diff
@@ -0,0 +1,34 @@
...
+    await client.connect();
+    await client.query("DROP TABLE users;");
+    await client.query("CREATE TABLE users ( id serial PRIMARY KEY, email varchar(320) );");
+    await client.end();
+  } catch (e) {
+    console.log(e);
+  }
+}
+
+async function smokeTest() {
```
nhnt11: Make this a real test, instead of relying on console.logs.

db-scripts.js
```diff
@@ -0,0 +1,34 @@
...
+async function setup() {
+  const client = new pg.Client({
+    database: AppConstants.DATABASE_NAME,
+    host: AppConstants.DATABASE_HOST,
+    port: AppConstants.DATABASE_PORT,
+  });
+
+  try {
+    await client.connect();
+    await client.query("DROP TABLE users;");
```
nhnt11: This isn't meant to go to production, or maybe ever get committed into the repository at all, but if we do, we should make it far more sophisticated and maybe make a backup of the table here.

db-scripts.js
```diff
@@ -0,0 +1,34 @@
...
+    await client.query("CREATE TABLE users ( id serial PRIMARY KEY, email varchar(320) );");
```
pdehaan: Is this going to be a plaintext email, or are we going to SHA1 it or something?

nhnt11:
Still working on this PR, needs a final bit of cleanup around testing and some of the abstraction.

routes/oauth.js
pdehaan: I believe this was filed as part of #30; _"Resolve code style inconsistencies"_.

pdehaan: And on that note, it may be nice to add some additional error checking here when we refactor in case of an error. Currently I'd say let's add a `.catch(err => {...})` to handle promise chain rejections/errors, but I guess post `async..await` we'd just use good old `try..catch` block to handle the unexpected.

subscribers.js
```diff
@@ -1,7 +1,95 @@
...
+  max: 10,
+};
+
+const dbPool = new pg.Pool(dbConfig);
+
+function dbq(aQuery) {
+  return dbPool.query(aQuery);
+}
+
+function _ret(aSuccess, aError, aOther) {
```
pdehaan: Not sure if we want to possibly use default arguments here.
For example, `function _ret(aSuccess, aError=null, aOther={})`. That way we wouldn't need `|| null`. ¯\\\_(ツ)\_/¯

subscribers.js
```diff
@@ -1,7 +1,95 @@
...
+
+  async getUser(aEmail) {
+    try {
+      const rows = (await dbq({
+        text: "SELECT * FROM users WHERE email = $1;",
+        values: [ aEmail ],
+      })).rows;
+
+      if (!rows.length) {
+        return _ret(false);
```
pdehaan: This feels like it should be more verbose on why it was an error.
Perhaps something like:

```js
return _ret(false, "Entry was not found");
// or
return _ret(false, new Error("Entry was not found"));
```

subscribers.js
```diff
@@ -1,7 +1,95 @@
...
+      }
+
+      const { email } = rows[0];
+      if (!email) {
+        // This shouldn't happen, but... better safe than sorry.
+        return _ret(false, "Entry was found but with no value");
```
pdehaan: In the `try..catch` blocks (see below L79, etc), we do `return _ret(false, e)` to return an error. Should this `aError` string be wrapped in an Error object too?
For example:
```js
return _ret(false, new Error("Entry was found but with no value"));
```

Not sure if I've grokked enough of the API code yet to see how we're handling the errors in Express yet to see how it'd handle strings vs objects or if these error messages aren't user facing yet.

tests/test-subscribers.js
```diff
@@ -1,10 +1,71 @@
...
+const AppConstants = require("../app-constants");
+const Subscribers = require("../subscribers");
+const DBUtils = require("../db-utils");
 
-test("gEmails is an object", (t) => {
-  t.plan(1);
-  t.equal(typeof(gEmails), "object");
+// eslint-disable-next-line no-process-env
+if (!process.env.TESTING_ENVIRONMENT) {
+  console.log("Attempting to run database tests without TESTING_ENVIRONMENT set, exiting.");
```
pdehaan: Why the explicit env var check here?
Does this mean `npm run test` will fail unless we add `TESTING_ENVIRONMENT=1 npm test`?

subscribers.js
```diff
@@ -1,7 +1,95 @@
...
+    const DUPLICATE_ERROR = 23505;
+    try {
+      await dbq({
+        text: "INSERT INTO users ( email ) VALUES ( $1 );",
+        values: [ aEmail ],
+      });
+
+      return _ret(true);
+    } catch (e) {
+      if (e.code == DUPLICATE_ERROR) {
```
pdehaan: Should be `===` (and we should verify that ESLint also catches strict eq).

routes/user.js
```diff
@@ -43,30 +44,34 @@ router.post("/add", (req, res) => {
 router.get("/verify", (req, res) => {
   const email = gUnverifiedEmails.get(req.query.state);
```
pdehaan: Should the `gUnverifiedEmails` Map be stored in Postgres as well?

```js
const gUnverifiedEmails = new Map();
```

tests/test-subscribers.js
```diff
@@ -1,10 +1,71 @@
 "use strict";
 
 const test = require("tape");
+const pg = require("pg");
 
-const gEmails = require("../subscribers");
+const AppConstants = require("../app-constants");
```
pdehaan: Nit: this failed CI due to these 2 lines (L4 and L6):

```
/home/travis/build/mozilla/blurts-server/tests/test-subscribers.js
   4:7  error    'pg' is assigned a value but never used            no-unused-vars
   6:7  error    'AppConstants' is assigned a value but never used  no-unused-vars
```

.travis.yml
```diff
@@ -2,3 +2,10 @@ language: node_js
 node_js:
     - "node"
     - "lts/*"
+services:
+  - postgresql
+env:
+  - TESTING_ENVIRONMENT=1
```
pdehaan: Travis-CI has a default `CI=true` environment variable: https://docs.travis-ci.com/user/environment-variables/#Default-Environment-Variables (and so does [Circle-CI](https://circleci.com/docs/2.0/env-vars/#build-details)), if we don't want to create/manage another env variable. Or, we could set `process.env.NODE_ENV=test` or something.

app-constants.js
```diff
@@ -23,16 +23,13 @@ const kEnvironmentVariables = [
...
-    for (const v of kEnvironmentVariables) {
-      if (process.env[v] === undefined) {
-        throw new Error(`Required environment variable was not set: ${v}`);
-      }
-      this[v] = process.env[v];
-    }
-    return this;
-  },
-};
+const AppConstants = { };
```
pdehaan: Should this be a `Map()` instead, and/or, should we use [`Object.freeze()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/freeze) (when exporting) to make the object immutable-_ish_?

.env-dist
```diff
@@ -20,3 +20,9 @@ HIBP_API_TOKEN=""
 
 AWS_ACCESS_KEY_ID=""
 AWS_SECRET_ACCESS_KEY=""
+
+PGUSER=postgres
+PGPASSWORD=
```
pdehaan: i assume there is no real difference between `PGPASSWORD=` and `PGPASSWORD=""`?

subscribers.js
```diff
@@ -1,7 +1,90 @@
...
+}
+
+const Subscribers = {
+  async addUser(aEmail) {
+    // Error code for attempting to add a duplicate entry on a UNIQUE key.
+    const DUPLICATE_ERROR = 23505;
+    try {
+      await dbq({
+        text: "INSERT INTO users ( email ) VALUES ( $1 );",
+        values: [ aEmail ],
```
pdehaan: Actually, looking at the db creation code, it feels like we may want to limit the email length here possibly (although I'm not 100% certain what happens when you try and write 4000 characters into a 320 character field in Postgres):

```js
await client.query("CREATE TABLE users ( email varchar(320) UNIQUE );");
```

... Or if we need to take unicode character length into consideration if people have email addresses w/ emojis or other valid glyphs.

groovecoder:
Travis is green!! good work!

routes/oauth.js
```diff
@@ -52,13 +52,13 @@ router.get("/redirect", (req, res) => {
...
-        const email = JSON.parse(data.body).email;
-        gEmails.add(email);
-        res.send(`Registered ${email} for breach alerts. You may now close this window/tab.`);
       });
-    })
-    .catch((err) => {
+    }).then((data) => {
+      return Subscribers.addUser(JSON.parse(data.body).email);
+    }).then(() => {
+      res.send(`Registered ${email} for breach alerts. You may now close this window/tab.`);
```
pdehaan: I'm a bit lost here... I'm not sure what, if anything, the `${email}` refers to here. Previously it was a stored copy of `JSON.parse(data.body).email`, but that doesn't seem to be getting stored or passed in the promise chain, that I can immediately see.

pdehaan: And from my limited knowledge of `Subscribers.addUser()`, it should return an empty object on success, so we can't rely on returning the `.addUser()` response. We may need to cache the `JSON.parse(data.body).email` value and return that in the promise chain.

app-constants.js
nhnt11: Filed #58

routes/user.js
nhnt11: Filed #59

nhnt11:
I'm going to squash this PR into one commit when we merge it, the commit history has gotten messy due to frustratedly debugging CI failure and overall it's not a severely complex patch if you look at it file by file.

routes/user.js
```diff
@@ -43,30 +44,34 @@ router.post("/add", (req, res) => {
...
   if (email) {
-    gEmails.add(email);
-    gUnverifiedEmails.delete(req.query.state);
-    res.json({ email, info: `Successfully added ${email}`});
+    Subscribers.addUser(email).then(() => {
+      gUnverifiedEmails.delete(req.query.state);
```
nhnt11: Hmm, we should delete the email from gUnverifiedEmails even if addUser fails (force the user to re-do the transaction).

If addUser fails, something is weird with the database and I'd rather not keep user emails around indefinitely in a table whose entries are supposed to be ephemeral.

routes/user.js
```diff
@@ -43,30 +44,34 @@ router.post("/add", (req, res) => {
...
+      res.json({ email, info: `Successfully added ${email}`});
+    });
```
nhnt11: This needs a catch. In its current form, if addUser fails, the HTTP response is never sent.

routes/user.js
```diff
@@ -43,30 +44,34 @@ router.post("/add", (req, res) => {
...
   // TODO: Needs better error message.
   res.json({ info: "Who are you?" });
 });
 
 router.post("/remove", (req, res) => {
-  gEmails.delete(req.body.email);
-  res.json({ email: req.body.email, info: "removed user" });
+  Subscribers.deleteUser(req.body.email).then(() => {
+    res.json({ email: req.body.email, info: "removed user" });
+  });
```
nhnt11: Needs a catch.

routes/user.js
```diff
@@ -43,30 +44,34 @@ router.post("/add", (req, res) => {
...
 });
 
 router.post("/reset", (req, res) => {
-  gEmails.clear();
-  res.json({ info: "user list cleared" });
+  Subscribers.clearAllUsers().then(() => {
+    res.json({ info: "user list cleared" });
+  });
```
nhnt11: Needs a catch.

routes/user.js
```diff
@@ -90,7 +99,7 @@ router.post("/breached", (req, res) => {
             response.push({ email, error });
           });
       });
-    }
+    });
```
nhnt11: Needs a catch with a console.log, at least.

db-utils.js
```diff
@@ -0,0 +1,20 @@
...
+const pg = require("pg");
+
+// Use PG* env vars to configure client.
+const client = new pg.Client();
+
+const DBUtils = {
+  async setupDatabase() {
+    try {
+      await client.connect();
+      await client.query("DROP TABLE IF EXISTS users;");
```
pdehaan: Should we add any magical checks to make sure this isn't run in `process.env.NODE_ENV=production`?
I tried doing a quick Ctrl+F on this page and only saw the one reference in tests/test-subscribers.js, but something about `DROP TABLE` makes me super paranoid that something gets nuked during when deploying or scaling.

subscribers.js
```diff
@@ -1,7 +1,89 @@
 "use strict";
 
-// Set of all registered emails.
-// TODO: Implement a real persistent storage solution.
-const gEmails = new Set();
+const pg = require("pg");
 
-module.exports = gEmails;
+const dbConfig = {
+  max: 10,
```
pdehaan: Is this something that OPs or whoever may want to tweak?

subscribers.js
```diff
@@ -1,7 +1,89 @@
...
+        return _ret(new Error("Entry was found but with no email value"));
+      }
+
+      return _ret(null, { email });
+    } catch(e) {
+      return _ret(e);
+    }
+  },
+
+  async clearAllUsers() {
```
pdehaan: Does anybody call this API endpoint?

I see we removed `clearUserList()` and `dumpUserList()`, but am not convinced this is even remotely related to those public/js/test.js methods:

```diff
-// eslint-disable-next-line no-unused-vars	
-function clearUserList() {	
-  doXHR("/user/reset", {});	
-}	
-	
-// eslint-disable-next-line no-unused-vars	
-function dumpUserList() {	
-  doXHR("/user/list", {});	
-}
```

subscribers.js
nhnt11: Hmm, this would be nice, yes, but I'm going to procrastinate on it because:
- This is not immediately important.
- Getting numerical values from the environment means using parseint, and that seems like a prime surface area for bugs and failures. 
- We should probably first improve AppConstants to also know what type each value is supposed to be and automate the parseints.

nhnt11:
I think this is nearly ready, shall I merge? @groovecoder @pdehaan

nhnt11:
Yeah, this patch is getting big and I'm fairly confident in it, let's merge.

subscribers.js
pdehaan: Cool. Filed as #62 for tracking.

PR #52, opened 2018-02-16
Rename make_sha1_hashes.js to hyphens for consistency

pdehaan:
Not sure if this should live in `./tests/fixtures/*` or `./scripts/make-sha1-hashes.js`.

@groovecoder: Thoughts?


groovecoder:
No strong preference on it ... yet.

Issue #31, opened 2018-02-07
Implement Intro page from mock-ups


groovecoder:
Breach Alert UX on Drive: https://drive.google.com/file/d/1AuhQCFHOsNduMF_Sne32mCMfMeaTFuVJ/view?usp=sharing

Including:

* [x] Input field for the user to input their email address
* [x] Button to search breaches for their email address
* [x] Button to subscribe to email notifications if their email address is included in a future breach

groovecoder:
[BreachAlert_Webpage_UX.pdf](https://github.com/mozilla/blurts-server/files/1741801/BreachAlert_Webpage_UX.pdf)

pdehaan:
Ref: #53 (which added wireframe UI)

groovecoder:
Done.

PR #53, opened 2018-02-23
for #31: start html templates with hbs & foundation

groovecoder:
Need first feedback on using [`hbs`](http://github.com/donpark/hbs) for HTML templating and using [Foundation](https://foundation.zurb.com/sites/docs/) for style framework.

Screenshot:
image


public/css/foundation.min.css
```diff
@@ -0,0 +1 @@
...x;margin-bottom:1px;font-size:.9rem;-ms-flex:0 0 auto;flex:0 0 auto}.button-group .button:last-child{margin-right:0}.button-group.tiny .button{font-size:.6rem}.button-group.small .button{font-size:.75rem}.button-group.large .button{font-size:1.25rem}.button-group.expanded .button{-ms-flex:1 1 0px;flex:1 1 0px}.button-group.primary .button{background-color:#1779ba;color:#fefefe}.button-group.primary .button:focus,.button-group.primary .button:hover{background-color:#126195;color:#fefefe}.button-group.secondary .button{background-color:#767676;color:#fefefe}.button-group.secondary .button:focus,.button-group.secondary .button:hover{background-color:#5e5e5e;color:#fefefe}.button-group.success .button{background-color:#3adb76;color:#0a0a0a}.button-group.success .button:focus,.button-group.success .button:hover{background-color:#22bb5b;color:#0a0a0a}.button-group.warning .button{background-color:#ffae00;color:#0a0a0a}.button-group.warning .button:focus,.button-group.warning .button:hover{background-color:#cc8b00;color:#0a0a0a}.button-group.alert .button{background-color:#cc4b37;color:#fefefe}.button-group.alert .button:focus,.button-group.alert .button:hover{background-color:#a53b2a;col
```
pdehaan: Instead of bundling vendor code into app, can we install **foundation** via npm and copy over the relevant files from ./node_modules/?

views/home.hbs
```diff
@@ -0,0 +1,66 @@
+<!DOCTYPE html>
+<html lang="en">
+    <head>
+        <meta charset="utf-8">
+        <meta name="viewport" content="width=device-width, initial-scale=1.0">
+        <title>Firefox Breach Alerts</title>
+        <link rel="stylesheet" href="/css/foundation.min.css">
+        <script type="text/javascript" src="/js/app.js" async></script>
+        <script type="text/javascript" src="/js/vendor/jquery.js" defer></script>
+        <script type="text/javascript" src="/js/vendor/foundation.min.js" defer></script>
```
pdehaan: (a) Should these `<script ...>` tags be at the bottom of `<body>` so they aren't blocking?
(b) Ridiculous pre-optimization question, but should we be including Twitter/OpenGraph `<meta>` tags (tl;dr: should i file a separate bug to add those)?

views/home.hbs
```diff
@@ -0,0 +1,66 @@
...
+            <h1><a href="/">Firefox Breach Alert</a></h1>
+          </div>
+          <div class="top-bar-right">
+            <ul class="menu">
+              <li><a href="/">About</a></li>
+              <li><a href="#tips">Tips</a></li>
+              <li><a class="button" data-open="subscribe-modal">Subscribe</a></li>
+            </ul>
+          </div>
+        </div>
```
pdehaan: Not sure how semantic we want to be w/ `<header>` and `<nav>` elements here.

public/css/foundation.min.css
pdehaan: Same w/ jquery.js (which doesn't seem to be minimized).

pdehaan:
I'm unclear on <kbd>public/main.js → public/js/app.js</kbd>, and if we still need that file. IIRC, it was just a debug page that displayed some responses to the DOM for debugging, and probably should have been yanked before going into production.

views/home.hbs
```diff
@@ -0,0 +1,66 @@
+<!DOCTYPE html>
+<html lang="en">
+    <head>
+        <meta charset="utf-8">
+        <meta name="viewport" content="width=device-width, initial-scale=1.0">
+        <title>Firefox Breach Alerts</title>
+        <link rel="stylesheet" href="/css/foundation.min.css">
```
pdehaan: Should we put the CSS into /css/vendor/ directory to match the JS code?
For example:
- public/css/app.css
- public/css/vendor/foundation.min.css

views/home.hbs
```diff
@@ -0,0 +1,66 @@
...
+
+    <body>
+        <div class="top-bar">
+          <div class="top-bar-left">
+            <h1><a href="/">Firefox Breach Alert</a></h1>
+          </div>
+          <div class="top-bar-right">
+            <ul class="menu">
+              <li><a href="/">About</a></li>
+              <li><a href="#tips">Tips</a></li>
```
pdehaan: nit: there is no DOM element with an `id` of "tips" (or anchor name). But again, I'm happy to file as a separate bug after this lands (since it's something I'll probably end up forgetting to check in the future unless I remind myself).

views/home.hbs
groovecoder: (a) `async` and `defer` attributes on script tags are the new hotness to gain the performance of async & deferred script loading.

(b) yeah, follow-up.

groovecoder:
I'm keeping the `test.html` page and it needs functions in `public/js/app.js`. Not calling this anything near "production" yet.

pdehaan:
> I'm keeping the `test.html` page and it needs functions in `public/js/app.js`.

OK, how about renaming `public/js/app.js` to `public/js/test.js` since it's only used by `test.html`?

views/home.hbs
pdehaan: Filed; #54

views/home.hbs
```diff
@@ -0,0 +1,65 @@
+<!DOCTYPE html>
+<html lang="en">
+    <head>
```
nhnt11: nit: the indent is inconsistent in this file: it starts with 4 spaces here but later uses 2 spaces in the body.

groovecoder: Updated so the entire doc is indented consistently.

Issue #56, opened 2018-02-26
Investigate adding automated HTML standards-compliance test


nhnt11:
Just an idea, not sure how doable this is or how it'd play with a templating system, but let's see.

pdehaan:
Not sure what you had in mind for linters, so I'll randomly throw some modules/results in this comment.
I haven't researched HTML linters in a while (but occasionally manually check via https://html5.validator.nu/ and https://validator.w3.org/.

Looks like Mozilla has an old version of a linter at https://github.com/mozilla/html5-lint which I can research and see if it's still valid.

---

I found [**html-minifier-lint**](https://github.com/kangax/html-lint), which pointed out a couple nits:

```sh
$ npx html-minifier-lint views/home.hbs

npx: installed 16 in 6.445s
 - Found event attribute (onclick) on <a> element.
 - Found presentational <hr> element
 - Found presentational <hr> element
```

---

[**html5-validator**](https://www.npmjs.com/package/html5-validator) says:

```sh
$ npx html5-validator views/home.hbs

  ⚠ 8:5 The “type” attribute is unnecessary for JavaScript resources.
  css">\n    <script type="text/javascript" src="/js/vendor/jquery.js" defer></scri

  ⚠ 9:5 The “type” attribute is unnecessary for JavaScript resources.
  ript>\n    <script type="text/javascript" src="/js/vendor/foundation.min.js" defer></scri

✖ You've got 0 errors, and 2 warnings.
```

---

http://htmllint.github.io/demo.html says:

```
line: 12 col: 16, ids and classes must match the format: underscore
line: 13 col: 18, ids and classes must match the format: underscore
line: 16 col: 18, ids and classes must match the format: underscore
line: 20 col: 32, attribute names must match the format: lowercase
line: 24 col: 16, ids and classes must match the format: underscore
line: 25 col: 18, ids and classes must match the format: underscore
line: 30 col: 22, ids and classes must match the format: underscore
line: 31 col: 26, ids and classes must match the format: underscore
line: 32 col: 24, ids and classes must match the format: underscore
line: 38 col: 18, ids and classes must match the format: underscore
line: 42 col: 28, the "data-reveal" attribute is not double quoted
line: 42 col: 5, Attribute values cannot be empty.
line: 42 col: 28, ids and classes must match the format: underscore
line: 42 col: 28, attribute names must match the format: lowercase
line: 45 col: 25, ids and classes must match the format: underscore
[...]

PR #57, opened 2018-02-27
Add HTMLLint

pdehaan:
Ref: #56

### DIFF
--- a/views/home.hbs
+++ b/views/home.hbs
@@ -48,7 +48,7 @@
     <div class="reveal" id="subscribe-modal" data-reveal>
       <h3>Subscribe to Breach Alerts</h3>
       <p>Explanation of what the user will get if they subscribe to breach alerts.</p>
-      <a href="#" class="button sign-in-btn" onClick="doOauth();">Subscribe with Firefox Accounts</a>
+      <a href="#" class="button sign-in-btn" onclick="doOauth();">Subscribe with Firefox Accounts</a>
       <div class="grid-x">
         <div class="cell small-5"><hr></div>
         <div class="cell small-2">or</div>
--- a/.htmllintrc
+++ b/.htmllintrc
@@ -2,6 +2,7 @@
   // names of npm modules to load into htmllint
   "plugins": [],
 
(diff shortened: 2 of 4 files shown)


views/home.hbs
```diff
@@ -48,7 +48,7 @@
     <div class="reveal" id="subscribe-modal" data-reveal>
       <h3>Subscribe to Breach Alerts</h3>
       <p>Explanation of what the user will get if they subscribe to breach alerts.</p>
-      <a href="#" class="button sign-in-btn" onClick="doOauth();">Subscribe with Firefox Accounts</a>
+      <a href="#" class="button sign-in-btn" onclick="doOauth();">Subscribe with Firefox Accounts</a>
```
groovecoder: Is there a way to configure HTMLLint to disallow `on*` attributes? These are security no-no.

pdehaan: As close as I can tell, we'd need to use the [`attr-bans`](https://github.com/htmllint/htmllint/wiki/Options#attr-bans) rule, and explicitly specify any `on*` attribute we wanted to disallow. Not great, but I don't see any other easy config or RegExp in the docs [yet].

pdehaan:
> We should add this to `.travis.yml` if we're going to use it.

❤️ No need. It's a `lint:*` task, which gets run automatically for each PR.
https://travis-ci.org/mozilla/blurts-server/jobs/346641658#L1668-L1672

views/home.hbs
pdehaan: Oops, missed this...
Added `"onclick"` to the `attr-bans` rule... which now makes the CI fail.
Ideas on how to tweak the breakage? Or are we cool w/ merging a PR that fails the CI and we can submit a follow-up issue?

groovecoder: Let's not merge a CI fail. Are you comfortable moving the script behavior into a separate `.js` file yourself, or prefer someone else do it?

groovecoder:
Superseded by https://github.com/mozilla/blurts-server/pull/66.

PR #64, opened 2018-03-06
Use database for email verification temp storage

### LINKED ISSUE #59: Use database for keeping track of emails in the process of being verified
Will be needed when we scale.

nhnt11:
Fixes #59

### DIFF
--- a/db-utils.js
+++ b/db-utils.js
@@ -1,21 +1,45 @@
 "use strict";
 
+require("dotenv").load();
+
 const pg = require("pg");
 
-// Use PG* env vars to configure client.
-const client = new pg.Client();
+function assertTestingEnvironment() {
+  // eslint-disable-next-line no-process-env
+  if (!process.env.TESTING_ENVIRONMENT) {
+    throw new Error("Attempting to run database setup without TESTING_ENVIRONMENT set, exiting.");
+  }
+}
 
 const DBUtils = {
-  async setupDatabase() {
-    // eslint-disable-next-line no-process-env
-    if (!process.env.TESTING_ENVIRONMENT) {
-      throw new Error("Attempting to run database setup without TESTING_ENVIRONMENT set, exiting.");
-    }
+  async setupUsersTable() {
+    assertTestingEnvironment();
 
+    // Use PG* env vars to configure client.
+    const client = new pg.Client();
     try {
       await client.connect();
       await client.query("DROP TABLE IF EXISTS users;");
-      await client.query("CREATE TABLE users ( email varchar(320) UNIQUE );");
+      await client.query("CREATE TABLE users ( email VARCHAR(320) UNIQUE );");
+    } finally {
+      await client.end();
+    }
+  },
+
+  async setupTempUsersTable() {
+    assertTestingEnvironment();
+
+    // Use PG* env vars to configure client.
+    const client = new pg.Client();
+    try {
+      await client.connect();
+      await client.query("DROP TABLE IF EXISTS users_temp;");
+      await client.query("CREATE TABLE users_temp ( \
+                            email VARCHAR(320) UNIQUE, \
+                            token VARCHAR(80), \
+                            time_added TIMESTAMP DEFAULT NOW() \
+                          );");
+
     } finally {
       await client.end();
     }
--- a/routes/user.js
+++ b/routes/user.js
@@ -2,58 +2,72 @@
 
 const AppConstants = require("../app-constants");
 
-const crypto = require("crypto");
 const express = require("express");
 const router = express.Router();
 
 const EmailUtils = require("../email-utils");
 const Subscribers = require("../subscribers");
 
-// We send verification emails to addresses that want to subscribe.
-// These addresses are temporarily stored here, mapped to the unique
-// string token that is used for verification.
-const gUnverifiedEmails = new Map();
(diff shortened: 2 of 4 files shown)


nhnt11:
Fixes #59

routes/user.js
```diff
@@ -2,58 +2,72 @@
...
-        info: "sent verification link",
-        // Send the would-be link back to the client in dummy mode.
-        // eslint-disable-next-line no-process-env
-        link: process.env.DEBUG_DUMMY_SMTP ? url : undefined,
-      });
-    })
-    .catch(error => {
-      res.json({ email, info: "failed to send verification link", error });
-      throw error;
+  const url = `${AppConstants.SERVER_URL}/user/verify?state=${state}&email=${email}`;
```
pdehaan: Sorry, i don't know the context of `state` or `email`, but do they need to be URI encoded?

routes/user.js
```diff
@@ -2,58 +2,72 @@
...
+
+  try {
+    const { error } = await Subscribers.addTempUser(email, state);
+
+    if (error) {
+      res.json({ email, error: new Error("Database error.") });
```
pdehaan: Not sure how "user facing" these errors may be, but are we sure that dumping an Error() object won't leak any useful information (like server file paths or stack traces)? Or should we just return a string, a la:

```js
res.json({ email, error: "Database error" });
```

routes/user.js
```diff
@@ -2,58 +2,72 @@
...
+    res.json({
+      email,
+      info: "sent verification link",
+      // Send the would-be link back to the client in dummy mode.
+      // eslint-disable-next-line no-process-env
+      link: process.env.DEBUG_DUMMY_SMTP ? url : undefined,
     });
+  } catch (error) {
+    console.log(error);
+    res.json({ email, info: "failed to send verification link", error });
```
pdehaan: Same as comment above, but should we be returning Error _objects_ to the user, or just the error messages?

```js
res.json({ email, info: "Failed to send verification link", error: error.message });
```

pdehaan:
> Fixes #59

Unrelated to anything specific, but... this comment needs to be in your _original_ commit message (you can edit it above) if you want GitHub to auto-close the associated issue.

Sorry, I don't seem to have permissions in this repo, otherwise I could edit the comments directly.

routes/user.js
nhnt11: Good point thanks

routes/user.js
nhnt11: Yeah, I thought about this but promptly forgot. A string is probably better here. In any case this should be replaced by an error code (#65)

subscribers.js
```diff
@@ -73,6 +74,74 @@ const Subscribers = {
...
+      return _ret(e);
+    }
+  },
+
+  async getTempUser(aEmail) {
+    try {
+      const rows = (await dbq({
+        text: "SELECT * FROM users_temp WHERE email = $1;",
+        values: [ aEmail ],
+      })).rows;
```
pdehaan: Per https://github.com/mozilla/blurts-server/pull/70#discussion_r173312299

> Not sure if we could just destructure `const {rows}` here, and save some brackets and `.rows` property. Maybe.
>
> ```js
> const {rows} = await dbq({
>   text: "SELECT * FROM users_temp WHERE email = $1;",
>   values: [ aEmail ],
> });
> ```

routes/user.js
```diff
@@ -2,58 +2,72 @@
...
+    if (token !== req.query.state) {
+      res.json({ error: new Error("Token and email don't match."), email });
+      return;
+    }
+
+    await Subscribers.addUser(email);
+    res.json({ email, info: `Successfully added ${email}` });
+  } catch (error) {
+    console.log(error);
+    res.json({ info: "Failed to add email to database.", error });
```
pdehaan: Should this also include the `email` property?

db-utils.js
```diff
@@ -1,21 +1,45 @@
...
+    // Use PG* env vars to configure client.
+    const client = new pg.Client();
+    try {
+      await client.connect();
+      await client.query("DROP TABLE IF EXISTS users_temp;");
+      await client.query("CREATE TABLE users_temp ( \
+                            email VARCHAR(320) UNIQUE, \
+                            token VARCHAR(80), \
+                            time_added TIMESTAMP DEFAULT NOW() \
+                          );");
```
pdehaan: I think if you use template strings with a backtick (``` ` ```), you can avoid doing the multiline backslash trick.

```js
      await client.query(`CREATE TABLE users_temp (
                            email VARCHAR(320) UNIQUE,
                            token VARCHAR(80),
                            time_added TIMESTAMP DEFAULT NOW()
                          );`);
```

nhnt11: Fixed in #70, let's continue the conversation there?

routes/user.js
nhnt11: Getting rid of returning the email in the response in #70.

subscribers.js
nhnt11: Fixed this in a commit on #70.

routes/user.js
nhnt11: #70 fixes this.

nhnt11:
Let's close this and just do it in #70

PR #66, opened 2018-03-08
Updated html with htmllint

groovecoder:


groovecoder:
Supersedes https://github.com/mozilla/blurts-server/pull/57 with a fix for the `htmllint` violation and a rebase off of the latest `master` branch.

public/js/test.js
```diff
@@ -50,3 +50,7 @@ function simulateBreach() {
 function doOauth() {
   window.open("/oauth/init");
 }
+
+$(document).foundation();
+
+document.querySelector("#subscribe-fxa-btn").addEventListener("click", doOauth);
```
pdehaan: Sorry, I'm not 100% on how the `<script defer>` attribute works. Do we know that the DOM will have the `#subscribe-fxa-btn` rendered by the time the deferred `<script>` tag is eventually loaded? Or do we need to throw this in a jQuery `$(document).ready()` type thing?

views/home.hbs
```diff
@@ -7,6 +7,7 @@
     <link rel="stylesheet" href="/css/vendor/foundation.min.css">
     <script type="text/javascript" src="/js/vendor/jquery.js" defer></script>
     <script type="text/javascript" src="/js/vendor/foundation.min.js" defer></script>
+    <script type="text/javascript" src="/js/test.js" defer></script>
```
pdehaan: If we have 3 `<script src="..." defer>` tags in the `<head>`, are we guaranteed they will load in a predictable order? Or is it possible that "/js/test.js" can load before "/js/vendor/jquery.js" and calling `$(document).foundation()` in "/js/test.js" could fail (since neither jQuery or Foundation are loaded)?

public/js/test.js
pdehaan: Nit: should this be jQuery-ified? A la `$("#subscribe-fxa-btn").click(doAuth);`, or something.

.htmllintrc
```diff
@@ -0,0 +1,12 @@
+{
+  // names of npm modules to load into htmllint
+  "plugins": [],
+
+  "attr-bans": ["align", "background", "bgcolor", "border", "frameborder", "longdesc", "marginwidth", "marginheight", "onclick", "scrolling", "style", "width"],
+  "attr-name-style": "dash",
```
groovecoder: Ideally, we could use something like `"attr-name-style": "^(?!on).*"` to disallow *ALL* `on*` attributes, but I'm not clever enough to figure out how `htmllint` wants that defined.

public/js/test.js
groovecoder: In local testing, yes: with the `defer` attribute, the `#subscribe-fxa-btn` element is available by the time `test.js` accesses it. Without the `defer` attribute, it's broken.

I want to use `document.querySelector` in our own code so that Foundation is the only code that depends on jQuery. So if we drop Foundation we can also drop jQuery.

views/home.hbs
groovecoder: [Per MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/script), yes: "Scripts with the defer attribute will execute in the order in which they appear in the document."

.htmllintrc
pdehaan: Yeah, unfortunately I don't see an `attr-bans-regex` option in the docs, and looking at the [`attr-bans` rule](https://github.com/htmllint/htmllint/blob/21c62bd8321b289b9d92d00265cdb48034d1133c/lib/rules/attr-bans.js), it looks like it's just doing a simple `.forEach()` loop, which doesn't support regexp.

I can try filing an upstream issue to see if they'd consider adding support for regular expression style syntax, or if it's currently possible but not obviously documented.
Currently I think the only way to do it would be to list every "on*" attribute individually.

pdehaan: Upstream `attr-bans-regex` request at https://github.com/htmllint/htmllint/issues/236 🐟

groovecoder:
Merging my own PR before too much bit-rot.

PR #67, opened 2018-03-08
Use sequelize for hashset script

groovecoder:
I'm not committed to using Sequelize, so this PR is just to show how it's used as a comparison to the more manual database code currently in `db-utils.js` and `subscribers.js`. In particular `subscribers.js` is already pointed in this "model" direction anyway: i.e., providing connection plumbing and helper methods to get, add, and delete user records in the database table.

I'll make inline comments on the main "magical" bits ...

### DIFF
--- a/.env-dist
+++ b/.env-dist
@@ -26,3 +26,5 @@ PGPASSWORD=""
 PGDATABASE="blurts"
 PGHOST="localhost"
 PGPORT=5432
+
+DATABASE_URL="postgres://postgres@localhost:5432/blurts"
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,1 +1,2 @@
 public/js/vendor/*.js
+db/migrations
--- a/.sequelizerc
+++ b/.sequelizerc
@@ -1,0 +1,7 @@
+const path = require("path");
+
+module.exports = {
+  "config": path.resolve("db", "config", "config.json"),
+  "migrations-path": path.resolve("db", "migrations"),
+  "models-path": path.resolve("db", "models"),
+}
--- a/db/models/index.js
+++ b/db/models/index.js
@@ -1,0 +1,38 @@
+"use strict";
+
+const fs        = require("fs");
+const path      = require("path");
+const Sequelize = require("sequelize");
+const basename  = path.basename(__filename);
+const env       = process.env.NODE_ENV || "development"; // eslint-disable-line no-process-env
+const config    = require(__dirname + "/../config/config.json")[env];
+const db        = {};
+
+let sequelize;
+
+if (config.use_env_variable) {
+  sequelize = new Sequelize(process.env[config.use_env_variable], config); // eslint-disable-line no-process-env
+} else {
+  sequelize = new Sequelize(config.database, config.username, config.password, config);
+}
+
+fs
+  .readdirSync(__dirname)
+  .filter(file => {
+    return (file.indexOf(".") !== 0) && (file !== basename) && (file.slice(-3) === ".js");
+  })
+  .forEach(file => {
+    const model = sequelize["import"](path.join(__dirname, file));
+    db[model.name] = model;
+  });
+
+Object.keys(db).forEach(modelName => {
+  if (db[modelName].associate) {
+    db[modelName].associate(db);
+  }
+});
+
+db.sequelize = sequelize;
+db.Sequelize = Sequelize;
+
+module.exports = db;
--- a/db/models/user.js
+++ b/db/models/user.js
@@ -1,0 +1,11 @@
+"use strict";
+module.exports = (sequelize, DataTypes) => {
+  const User = sequelize.define("User", {
+    email: DataTypes.STRING,
+    sha1: DataTypes.STRING,
+  }, {});
+  User.associate = function(models) {
+    User.hasMany(models.BreachedUser);
+  };
+  return User;
+};
--- a/app-constants.js
+++ b/app-constants.js
@@ -21,6 +21,7 @@ const kEnvironmentVariables = [
   "HIBP_API_TOKEN",
   "AWS_ACCESS_KEY_ID",
   "AWS_SECRET_ACCESS_KEY",
+  "DATABASE_URL",
 ];
 
 const AppConstants = { };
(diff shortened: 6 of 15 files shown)


.eslintignore
```diff
@@ -1 +1,2 @@
 public/js/vendor/*.js
+db/migrations
```
groovecoder: I'm ignoring the migration code that is automatically generated by sequelize.

.sequelizerc
```diff
@@ -0,0 +1,7 @@
+const path = require("path");
+
+module.exports = {
+  "config": path.resolve("db", "config", "config.json"),
+  "migrations-path": path.resolve("db", "migrations"),
+  "models-path": path.resolve("db", "models"),
+}
```
groovecoder: Only need this because I want to keep [storing the db code in the `db/` directory while running `sequelize` command-line scripts from the project directory](http://docs.sequelizejs.com/manual/tutorial/migrations.html#the-sequelizerc-file).

.env-dist
```diff
@@ -26,3 +26,5 @@ PGPASSWORD=""
 PGDATABASE="blurts"
 PGHOST="localhost"
 PGPORT=5432
+
+DATABASE_URL="postgres://postgres@localhost:5432/blurts"
```
groovecoder: I like this single `DATABASE_URL` config, even if we don't use sequelize models.

db/models/index.js
```diff
@@ -0,0 +1,38 @@
...
+Object.keys(db).forEach(modelName => {
+  if (db[modelName].associate) {
+    db[modelName].associate(db);
+  }
+});
+
+db.sequelize = sequelize;
+db.Sequelize = Sequelize;
+
+module.exports = db;
```
groovecoder: This module just scans the other files in the `db/models/` directory and loads them all up at once. So you only need to add a model file to `db/models/` and then you can start using it like this:
```
const models = require("../db/models");

models.Breach.findOrCreate(...);
```

db/models/user.js
```diff
@@ -0,0 +1,11 @@
+"use strict";
+module.exports = (sequelize, DataTypes) => {
+  const User = sequelize.define("User", {
+    email: DataTypes.STRING,
+    sha1: DataTypes.STRING,
```
groovecoder: I added this `sha1` field for users to hold a pre-computed SHA1 hash for faster lookups.

db/models/index.js
```diff
@@ -0,0 +1,38 @@
...
+const path      = require("path");
+const Sequelize = require("sequelize");
+const basename  = path.basename(__filename);
+const env       = process.env.NODE_ENV || "development"; // eslint-disable-line no-process-env
+const config    = require(__dirname + "/../config/config.json")[env];
+const db        = {};
+
+let sequelize;
+
+if (config.use_env_constiable) {
```
pdehaan: While struggling to figure out if "constiable" is a word, I think this is the unholy union of a search-replace `var` with `const`, and this should be `config.use_env_variable`.

db/models/index.js
```diff
@@ -0,0 +1,38 @@
...
+  sequelize = new Sequelize(process.env[config.use_env_variable], config); // eslint-disable-line no-process-env
+} else {
+  sequelize = new Sequelize(config.database, config.username, config.password, config);
+}
+
+fs
+  .readdirSync(__dirname)
+  .filter(file => {
+    return (file.indexOf(".") !== 0) && (file !== basename) && (file.slice(-3) === ".js");
```
pdehaan: You could possibly use [`String.prototype.endsWith()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/endsWith) API:

```js
file.endsWith(".js")
```

Not sure if you could replace `file.indexOf(".") !== 0` with [`!file.startsWith(".")`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/startsWith), or if my logic/understanding is horribly wrong.

.env-dist
pdehaan: Should we just kill L24-28?

groovecoder: When we remove the "raw" `pg` code we can remove this at the same time.

db/models/index.js
```diff
@@ -0,0 +1,38 @@
...
+const Sequelize = require("sequelize");
+const basename  = path.basename(__filename);
+const env       = process.env.NODE_ENV || "development"; // eslint-disable-line no-process-env
+const config    = require(__dirname + "/../config/config.json")[env];
+const db        = {};
+
+let sequelize;
+
+if (config.use_env_constiable) {
+  sequelize = new Sequelize(process.env[config.use_env_variable], config); // eslint-disable-line no-process-env
```
pdehaan: Instead of `process.env`, not sure if we could use `AppConstants[config.use_env_variable]` wrapper. Or just `AppConstants.DATABASE_URL` directly, since all 3 envs seem to use the same key.

db/models/index.js
```diff
@@ -0,0 +1,38 @@
...
+} else {
+  sequelize = new Sequelize(config.database, config.username, config.password, config);
```
pdehaan: I'm not sure if this would work since I don't see a `database`, `username`, or `password` field in [./db/config/config.json](https://github.com/mozilla/blurts-server/pull/67/files#diff-3050f5967be4a98d0035b27aa4050e97).

db/models/index.js
```diff
@@ -0,0 +1,38 @@
...
+}
+
+fs
+  .readdirSync(__dirname)
```
pdehaan: Not sure if this would be easier if we moved this file into the parent directory, ./db/models.js, and then the models/* directory was just the model schemas. Then you could use something like [**glob**](http://npm.im/glob) to do:

```js
glob.sync("./models/*.js")
  .forEach(file => {
    const model = sequelize.import(path.join(__dirname, file));
    db[model.name] = model;
  });
```

groovecoder:
I hesitate to change (more about) the way `db/models/index.js` works since it was auto-generated by Sequelize. If Sequelize changes some way of loading model modules in the future, I'd like to stay as close to their implementation as possible so we can stay on their updates.

pdehaan:
> I hesitate to change (more about) the way db/models/index.js works since it was auto-generated by Sequelize.

If it's auto generated, do you want to add that file to .eslintignore too?

groovecoder:
I merged https://github.com/mozilla/blurts-server/pull/71 which included this.

PR #70, opened 2018-03-08
Use database for email verification, use real response codes, some refactoring.

### LINKED ISSUE #59: Use database for keeping track of emails in the process of being verified

### LINKED ISSUE #65: Define API response codes and use them; get rid of strings in response JSON

### LINKED ISSUE #68: Remove /user/breached API
We should just make this a script. It does not need to have a HTTP API.

### LINKED ISSUE #69: Use async callbacks for /user and /oauth API endpoints

nhnt11:
Fixes #59 
Fixes #68 
Fixes #69
Fixes #65

### DIFF
--- a/db-utils.js
+++ b/db-utils.js
@@ -1,21 +1,45 @@
 "use strict";
 
+require("dotenv").load();
+
 const pg = require("pg");
 
-// Use PG* env vars to configure client.
-const client = new pg.Client();
+function assertTestingEnvironment() {
+  // eslint-disable-next-line no-process-env
+  if (!process.env.TESTING_ENVIRONMENT) {
+    throw new Error("Attempting to run database setup without TESTING_ENVIRONMENT set, exiting.");
+  }
+}
 
 const DBUtils = {
-  async setupDatabase() {
-    // eslint-disable-next-line no-process-env
-    if (!process.env.TESTING_ENVIRONMENT) {
-      throw new Error("Attempting to run database setup without TESTING_ENVIRONMENT set, exiting.");
-    }
+  async setupUsersTable() {
+    assertTestingEnvironment();
 
+    // Use PG* env vars to configure client.
+    const client = new pg.Client();
     try {
       await client.connect();
       await client.query("DROP TABLE IF EXISTS users;");
-      await client.query("CREATE TABLE users ( email varchar(320) UNIQUE );");
+      await client.query("CREATE TABLE users ( email VARCHAR(320) UNIQUE );");
+    } finally {
+      await client.end();
+    }
+  },
+
+  async setupTempUsersTable() {
+    assertTestingEnvironment();
+
+    // Use PG* env vars to configure client.
+    const client = new pg.Client();
+    try {
+      await client.connect();
+      await client.query("DROP TABLE IF EXISTS users_temp;");
+      await client.query("CREATE TABLE users_temp ( \
+                            email VARCHAR(320) UNIQUE, \
+                            token VARCHAR(80), \
+                            time_added TIMESTAMP DEFAULT NOW() \
+                          );");
+
     } finally {
       await client.end();
     }
--- a/db-utils.js
+++ b/db-utils.js
@@ -34,11 +34,11 @@ const DBUtils = {
     try {
       await client.connect();
       await client.query("DROP TABLE IF EXISTS users_temp;");
-      await client.query("CREATE TABLE users_temp ( \
-                            email VARCHAR(320) UNIQUE, \
-                            token VARCHAR(80), \
-                            time_added TIMESTAMP DEFAULT NOW() \
-                          );");
+      await client.query(`CREATE TABLE users_temp (
+                            email VARCHAR(320) UNIQUE,
+                            token VARCHAR(80),
+                            time_added TIMESTAMP DEFAULT NOW()
+                          );`);
 
     } finally {
       await client.end();
--- a/routes/user.js
+++ b/routes/user.js
@@ -15,12 +15,12 @@ const ResponseCodes = Object.freeze({
   TokenMismatch: 102,
 });
 
-function resDatabaseError(res) {
+function resDatabaseError(res, error) {
   console.log(error);
   res.status(500).json({
     code: 500,
     error_code: ResponseCodes.InternalError,
-    info: "Database error."
+    info: "Database error.",
   });
 }
 
@@ -30,7 +30,11 @@ router.post("/add", async (req, res) => {
   const email = req.body.email;
 
   if (!email)  {
-    resDatabaseError(res);
+    res.status(400).json({
+      code: 400,
+      error_code: ResponseCodes.EmailNotProvided,
+      info: "Request did not include email address.",
+    });
     return;
   }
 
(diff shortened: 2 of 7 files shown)


nhnt11:
This is based on top of #64

db-utils.js
```diff
@@ -1,21 +1,45 @@
...
+    // Use PG* env vars to configure client.
+    const client = new pg.Client();
+    try {
+      await client.connect();
+      await client.query("DROP TABLE IF EXISTS users_temp;");
+      await client.query("CREATE TABLE users_temp ( \
+                            email VARCHAR(320) UNIQUE, \
+                            token VARCHAR(80), \
+                            time_added TIMESTAMP DEFAULT NOW() \
+                          );");
```
pdehaan: I think if you use template strings with a backtick (``` ` ```), you can avoid doing the multiline backslash trick.

```js
      await client.query(`CREATE TABLE users_temp (
                            email VARCHAR(320) UNIQUE,
                            token VARCHAR(80),
                            time_added TIMESTAMP DEFAULT NOW()
                          );`);
```

subscribers.js
```diff
@@ -73,6 +74,74 @@ const Subscribers = {
...
+    const DUPLICATE_ERROR = "23505";
+    try {
+      await dbq({
+        text: "INSERT INTO users_temp ( email, token ) VALUES ( $1, $2 );",
+        values: [ aEmail, aToken ],
+      });
+
+      return _ret();
+    } catch (e) {
+      if (e.code === DUPLICATE_ERROR) {
```
pdehaan: Not sure how to test this, but `DUPLICATE_ERROR` above (L84) looks like a String, `const DUPLICATE_ERROR = "23505";`. Is `e.code` similarly a string version of a number? Since this is using `===` I don't think it will auto-convert to a number and this may not do what we expect.

subscribers.js
```diff
@@ -73,6 +74,74 @@ const Subscribers = {
...
+      });
+      return _ret();
+    } catch(e) {
+      return _ret(e);
+    }
+  },
+
+  async getTempUser(aEmail) {
+    try {
+      const rows = (await dbq({
```
pdehaan: Not sure if we could just destructure `const {rows}` here, and save some brackets and `.rows` property. Maybe.

```js
const {rows} = await dbq({
  text: "SELECT * FROM users_temp WHERE email = $1;",
  values: [ aEmail ],
});
```

subscribers.js
nhnt11: Yup, e.code is a string. Sigh. I'm actually glad I caught this, DUPLICATE_ERROR used to be a number so this check would never evaluate to true.

subscribers.js
nhnt11: Nice

routes/user.js
```diff
@@ -2,100 +2,123 @@
...
+  InternalError: 999,
+  EmailNotProvided: 100,
+  EmailNotFound: 101,
+  TokenMismatch: 102,
+});
+
+function resDatabaseError(res) {
+  console.log(error);
+  res.status(500).json({
+    code: 500,
```
pdehaan: I'll argue that returning `code` in the JSON response may be all that useful since it just seems to echo the server response code. Plus, using `fetch()` or `XMLHttpRequest()` or `request()`/`got()` should give the user that information if they need it. And having `code` and `error_code` in the response may be confusing (not that I expect people to hit this API directly, I guess -- so maybe all of this is acedemic).

@groovecoder Thoughts?

routes/user.js
```diff
@@ -2,100 +2,123 @@
...
-router.post("/remove", (req, res) => {
+router.post("/remove", async (req, res) => {
   const email = req.body.email;
-  Subscribers.deleteUser(email).then(() => {
+  try {
+    await Subscribers.deleteUser(email);
     res.json({ email, info: "removed user" });
-  }).catch((error) => {
+  } catch (error) {
     res.json({ email, info: "Failed to remove user.", error });
```
pdehaan: Should this set a custom `.status()`?

nhnt11: I'm going to fix up the /remove API in a future commit. It needs email verification of some sort too.

routes/user.js
nhnt11: Hmm, I copied this from the fxa-oauth-server API [1]
I figured it's just for consumers' convenience, I don't care about it, I'm okay with not including the HTTP code.

[1] https://github.com/mozilla/fxa-oauth-server/blob/master/docs/api.md

subscribers.js
pdehaan: Migrated comment to https://github.com/mozilla/blurts-server/pull/64/files#r173320471

routes/user.js
```diff
@@ -2,100 +2,122 @@
...
+  EmailNotProvided: 100,
+  EmailNotFound: 101,
+  TokenMismatch: 102,
+});
+
+function resDatabaseError(res, error) {
+  console.log(error);
+  res.status(500).json({
+    error_code: ResponseCodes.InternalError,
+    info: "Database error.",
```
pdehaan: Since we're now passing the `error` argument, would there be any value in trying to say what the database error was?
For example:
```js
info: `Database error: ${error.message}`,
```

nhnt11: I left this out kind of intentionally. We're logging the error server-side, I don't see any reason to give clients too much information :P 

With the info strings I'm aiming to give clients enough to get back to us with a hint of what error they're seeing.

routes/user.js
```diff
@@ -2,100 +2,122 @@
...
 const router = express.Router();
 
 const EmailUtils = require("../email-utils");
 const Subscribers = require("../subscribers");
 
-// We send verification emails to addresses that want to subscribe.
-// These addresses are temporarily stored here, mapped to the unique
-// string token that is used for verification.
-const gUnverifiedEmails = new Map();
+const ResponseCodes = Object.freeze({
```
groovecoder: Is there a doc or some other link that describes where these error codes came from? Can we include it in a comment here?

routes/user.js
```diff
@@ -2,100 +2,122 @@
...
+      error_code: ResponseCodes.EmailNotFound,
+      info: "Email not marked for verification.",
     });
     return;
   }
-  // TODO: Needs better error message.
-  res.json({ info: "Who are you?" });
+
+  // Invalidate the entry even if the token doesn't match.
+  ({ error } = await Subscribers.deleteTempUser(email));
```
groovecoder: Does this open a DOS attack against email addresses? I.e., an attacker could repeatedly send requests to `/verify?email=victim@example.com` so their temp user record would be deleted before they could ever verify it?

subscribers.js
groovecoder: If/when we convert Subscriber DB code to use sequelize, we may get duplicate error checking "for free".

routes/user.js
pdehaan: Do we plan on having some cron script which will auto-delete temp users after X hours/days?

groovecoder: I would prefer to have that, yes.

nhnt11: Nice :D

routes/user.js
nhnt11: Yeah, I'm going to update api.md

PR #71, opened 2018-03-12
Sequelize all the things

groovecoder:
This builds on top of #67 and #70 and is mainly meant to show how the rest of the code will start to look when we use Sequelize everywhere.

Pay most attention to the last https://github.com/mozilla/blurts-server/pull/71/commits/51b3d567d442e22a43dc3aefd8e2cf9803ffe8cc commit to see the main refactorings:

* Moves user email validation "up" into the model object using Sequelize built-in validation code
* Moves the verification token into a field on the User model/table with default value generation, so we don't need an extra table
* Removes almost all of the DB "plumbing" code from `routes/users.js` since that is handled by Sequelize

If this looks good, I'll build on top of this to convert all the code to Sequelize. (Including tests)


routes/oauth.js
```diff
@@ -37,31 +37,30 @@ router.get("/init", (req, res) => {
...
+    const user = await FxAOAuth.code.getToken(req.originalUrl, { state: req.session.state });
+    const data = await popsicle.get({
+      method: "get",
+      url: FxAOAuthUtils.profileUri,
+      body: "",
+      headers: {
+        Authorization: `Bearer ${user.accessToken}`,
+      },
+    });
+    const email = JSON.parse(data.body).email;
```
pdehaan: Random thought, but should we be trying to validate an email address is valid-ish before trying to create a User object (via something like [**hapijs/isemail**](https://github.com/hapijs/isemail))? Or, is that automatically done by sequelize around https://github.com/mozilla/blurts-server/pull/71/files#diff-ea1d787ec7c82cea4820cbcab09ccee5R10?

routes/user.js
```diff
@@ -6,95 +6,63 @@ const crypto = require("crypto");
...
+const ResponseCodes = Object.freeze({
+  InternalError: 999,
+  EmailNotProvided: 100,
+  EmailNotFound: 101,
+  TokenMismatch: 102,
+});
+
+router.post("/add", async (req, res) => {
+  const user = await models.User.create({ email: req.body.email });
+  const url = `${AppConstants.SERVER_URL}/user/verify?state=${user.verificationToken}&email=${user.email}`;
```
pdehaan: Should we be URI-encoding the `user.verificationToken` and `user.email` here since they are query params?

routes/user.js
```diff
@@ -6,95 +6,63 @@ const crypto = require("crypto");
...
-router.get("/verify", (req, res) => {
-  const email = gUnverifiedEmails.get(req.query.state);
-  if (email) {
-    gUnverifiedEmails.delete(req.query.state);
-    Subscribers.addUser(email).then(() => {
-      res.json({ email, info: `Successfully added ${email}`});
-    }).catch((error) => {
-      res.json({ info: "Failed to add email to database.", error });
+router.get("/verify", async (req, res) => {
+  // eslint-disable-next-line prefer-const
```
pdehaan: Not sure this comment is valid, since you are in fact using a `const`.

routes/oauth.js
groovecoder: It's automatically done by Sequelize.

routes/user.js
groovecoder: They are hex digest values, which should be safe for uris.

routes/user.js
```diff
@@ -6,95 +6,63 @@ const crypto = require("crypto");
...
+  const user = await models.User.findOne({ where: { email: req.query.email, verificationToken: req.query.state } });
+  if (user === null) {
+    res.status(400).json({
+      error_code: ResponseCodes.EmailNotFound,
+      info: "Email not found or verification token does not match.",
     });
     return;
+  } else {
```
pdehaan: nit: Probably don't need this `else` line, since the `if` statement above already `return`ed.

tests/test-user.js
```diff
@@ -0,0 +1,33 @@
+"use strict";
+
+require("dotenv").load();
```
pdehaan: Do we use any ENV vars here that we need loaded via `dotenv`? Or does ../db/models require something?

tests/test-user.js
```diff
@@ -0,0 +1,33 @@
...
+  t.equal(user.verificationToken.length, 80, "verificationToken is expected length");
+});
+
+test("Test building user validates email", t => {
+  t.plan(2);
+  const valid_email = "test@test.com";
+  const invalid_email = "asdfghjkl";
+
+  let user = models.User.build({ email: valid_email });
+  user.validate().then(user=>{
```
pdehaan: Not sure if it's worth using `async`/`await` in the tests to be consistent with the other code. I don't feel strongly either way.

tests/test-user.js
groovecoder: No, we don't need `dotenv` here. In fact, I'm starting to dislike `dotenv` and `AppConstants` in general. Because other libraries (e.g., Sequelize) are written to use actual `process.env` variables - not use the `dotenv` magic.

I may file an issue to remove `dotenv`, and perhaps refactor `AppConstants` too.

tests/test-user.js
```diff
@@ -0,0 +1,46 @@
...
+  t.equal(user.sha1, undefined, "sha1 starts undefined");
+  user.save();
+
+  user = models.User.findOne({ where: { email: VALID_EMAIL } }).then(user=>{
+    const savedSha1 = user.saveSha1();
+    t.notEqual(savedSha1, undefined, "sha1 is no longer undefined");
+    t.equal(savedSha1.length, 40, "sha1 is expected hash length");
+  });
+});
+*/
```
pdehaan: Why commented out, and should we add a TODO or issue to uncomment these in the future?

tests/test-user.js
pdehaan: I'm torn. I like [**node-convict**](https://github.com/mozilla/node-convict) for managing configs and flags, but it can get a bit verbose. But I'd be curious to see your proposal for replacing .env files and keeping the 20+ env var configs sane.

Another big plus for **convict** is that you could have separate config files for "development", "production", and "testing", if we wanted to store different values per environment (although it may not be wise checking those in to source control). But still having all my local API keys and flags set in a local "./configs/development.json" would be useful).

groovecoder: From what I see, I also like `node-convict` more than `dotenv`. It seems more sane to put 20+ configs in the single `config.js` module with default values than to declare and define them across `.env-dist`, `.env`, and `app-constants.js`.

groovecoder: FWIW, I have only ever used `.env` files along with [`autoenv`](https://github.com/kennethreitz/autoenv) for a local/development convenience. I've never used `.env` files for staging or production app servers. So, my local `.env` looks like:
```
export DATABASE_URL=<local-dev-value>
export DEBUG_DUMMY_SMTP=1
...
```
And `autoenv` simply runs those `export` statements when I `cd` into the working directory.

The app code itself *always* pulls from environment variables or default values - never depends on `.env` files.

tests/test-user.js
groovecoder: Filed https://github.com/mozilla/blurts-server/issues/72

Issue #34, opened 2018-02-07
Implement Breach Search Results Page


groovecoder:
Includes:

* [x] List of breaches in which the email address was included
* [x] [CTA ](https://docs.google.com/document/d/1xknOqyDzCPqatXltkxsV9iC-9sf75b0uJN2OcjYHIaI/edit)

Tasks:

* [ ] [Implement FetchHashsets to download email hashsets of HIBP breaches](https://github.com/mozilla/blurts-server/issues/35)
* [x] Search the email hashsets for the email address entered by the user
* [x] Display breach information for each of the breaches found

groovecoder:
Done.

PR #75, opened 2018-04-05
for #31 #34 views for home, scan, and confirm

groovecoder:
See https://drive.google.com/file/d/1AuhQCFHOsNduMF_Sne32mCMfMeaTFuVJ/view for mock-ups.

### DIFF
--- a/db/models/subscriber.js
+++ b/db/models/subscriber.js
@@ -4,6 +4,8 @@ const crypto = require("crypto");
 
 module.exports = (sequelize, DataTypes) => {
 
+  const EmailHash = sequelize.import("./emailhash");
+
   const Subscriber = sequelize.define("Subscriber", {
     email: {
       type: DataTypes.STRING,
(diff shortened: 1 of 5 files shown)


db/models/subscriber.js
```diff
@@ -4,6 +4,8 @@ const crypto = require("crypto");
 
 module.exports = (sequelize, DataTypes) => {
 
+  const EmailHash = sequelize.import("./emailhash");
```
groovecoder: Have to import here to use `EmailHash` class methods.

routes/oauth.js
```diff
@@ -58,10 +58,14 @@ router.get("/redirect", jsonParser, async (req, res) => {
...
-    const user = await models.User.create({ email: email });
+    const user = await models.Subscriber.create({ email: email });
     user.saveSha1();
 
-    res.send(`Registered ${email} for breach alerts. You may now close this window/tab.`);
+    res.render("confirm", {
+      title: "Firefox Breach Alerts: Subscribed",
+      email: email,
+      user: user,
+    });
```
groovecoder: The page now renders a template instead of showing a message. We should probably change the URL from `/redirect` to `/confirm` ?

views/home.hbs
```diff
@@ -45,6 +73,6 @@
       </div>
       <form action="/subscribe" method="post">
         <input type="email" name="email">
-        <input type="submit" class="button" value="Subscribe with Email Account">
+        <button type="submit" class="button" disabled>Subscribe with Email Account</button>
```
groovecoder: `button` so I can use `disabled` attribute.

routes/oauth.js
nhnt11: That does make sense, but it also makes sense to keep it "redirect" to stay idiomatic with the OAuth API. I'll let you make the call on this.

groovecoder: In my understanding of OAuth the "redirect" is from the auth provider - Fx Accounts - to the relying party - Breach Alerts. So when I think "redirect URL" I think "to what Breach Alerts URL should FxA redirect when auth is confirmed?"

In that understanding/thinking, a URL like `/confirm` or `/auth-confirmed` makes idiomatic sense too. It would also make the URL match the contents of the page.

To update it, we'll need to change the registered redirect URL in the FxA dashboard. So, I'll file a follow-up issue to do this later.

PR #80, opened 2018-04-11
fix #77: enable subscribe-email-input and button

### LINKED ISSUE #77: Enable subscribe-by-email flow
Move/copy the subscribe-by-email functionality from test.html into the subscribe modal alongside Firefox Accounts.

groovecoder:
This actually enables the "Subscribe with Email Account" button, and puts the user onto a verification page when they click it.

Note: you need to unset `DEBUG_DUMMY_SMTP` and working `SMTP_*` environment variables for the email verification to actually send. https://fx-breach-alerts.herokuapp.com/ is set up with a demo account of Mailgun that is only authorized to send emails to lcrouch at mozilla dot com and nhnt11 at gmail dot com.

### DIFF
--- a/public/js/test.js
+++ b/public/js/test.js
@@ -45,6 +45,22 @@ function doOauth() {
   window.open("/oauth/init");
 }
 
+function isValidEmail(val) {
+  // https://stackoverflow.com/a/46181
+  const re = /^(([^<>()[\]\\.,;:\s@"]+(\.[^<>()[\]\\.,;:\s@"]+)*)|(".+"))@((\[[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\])|(([a-zA-Z\-0-9]+\.)+[a-zA-Z]{2,}))$/;
+  return re.test(String(val).toLowerCase());
+}
+
+function enableBtnIfEmailValid(e) {
+  const emailBtn = document.getElementById("subscribe-email-btn");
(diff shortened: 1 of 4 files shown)


public/js/test.js
```diff
@@ -45,6 +45,22 @@ function doOauth() {
...
 }
 
+function isValidEmail(val) {
+  // https://stackoverflow.com/a/46181
+  const re = /^(([^<>()[\]\\.,;:\s@"]+(\.[^<>()[\]\\.,;:\s@"]+)*)|(".+"))@((\[[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\])|(([a-zA-Z\-0-9]+\.)+[a-zA-Z]{2,}))$/;
+  return re.test(String(val).toLowerCase());
+}
+
+function enableBtnIfEmailValid(e) {
+  const emailBtn = document.querySelector("#subscribe-email-btn");
```
nhnt11: getElementById is three orders of magnitude faster than querySelector: https://jsperf.com/nhnt11-queryselector-vs-getelementbyid

Since this is being called continuously with input, we should use getElementById. I'd even say we could just cache emailBtn outside the function to avoid accessing the DOM every time.

groovecoder: updated to `getElementById`

groovecoder:
Rebased on master after the merge from #81.

PR #81, opened 2018-04-12
Make things slightly prettier

nhnt11:


public/css/app.css
```diff
@@ -0,0 +1,72 @@
+@import url('https://fonts.googleapis.com/css?family=Fira+Sans:400,600,700');
```
groovecoder: Could we potentially copy and host this resource ourselves?

views/layouts/default.hbs
```diff
@@ -5,29 +5,28 @@
...
-        <input type="submit" class="button" value="Subscribe with Email Account">
-      </form>
+      <div class="top-bar-right">
+        <ul class="menu">
+          <li><a href="/">About</a></li>
+          <li><a href="#tips">Tips</a></li>
+          <li><a class="button" data-open="subscribe-modal">Subscribe</a></li>
+        </ul>
+      </div>
+    </div>
```
groovecoder: Thanks for moving this up into `default.hbs`

views/home.hbs
```diff
@@ -1,78 +1,119 @@
...
+    </p>
+  </div>
+  <div class="cell small-5 padding">
+    <img alt="placeholder" class="thumbnail" src="https://placehold.it/550x310">
+  </div>
+
+  <div class="cell small-1"></div>
+
+
+<div class="reveal grid-x" id="subscribe-modal" data-reveal>
```
groovecoder: We need to move this to `default.hbs` too, otherwise the 'Subscribe' button on `/scan` doesn't work.

public/css/app.css
pdehaan: Not sure if it'd give us all the weights we want, but it looks like Mozilla does have some sort of Fira CDN:

Per https://github.com/mozilla/Fira#usage,

> **Usage**
> Use this font on your website!
>
> ```html
> <link rel="stylesheet" href="https://code.cdn.mozilla.net/fonts/fira.css">
> ```

PR #82, opened 2018-04-13
Make add and confirm pretty; some style refactoring

nhnt11:


groovecoder:
Looks good. I noticed that our `/verify` url is still an ugly JSON response ... should we make it a template render similar to `/confirm` ?

nhnt11:
Yeah, makes sense to make a template for /verify. I can do it.

Issue #79, opened 2018-04-09
Script to notify subscribers of new breaches


groovecoder:
When a new breach lands in the database, send an email notification to all subscribers who are included in the breach.

groovecoder:
This is done. We will make improvements with follow-up issues.

PR #94, opened 2018-04-23
 for #79: use nodemailer-express-handlebars for emails 

groovecoder:
Still `TODO`:
* [x] Figure out why the script hangs and doesn't return
* [ ] Style the HTML emails
* [ ] Inline the Foundation CSS used in the emails
* [x] Mark the subscriber/email_hash as being notified after we send them an email, to prevent double-sending

To try it out:
```
node tests/fixtures/make-breach-with-emails.js some-other@email.com
```
should add `some-other@email.com` as an email in all the breaches.

```
node scripts/notify-subscribers-of-breach.js "Test Breach 1"
```
should `console.log` all of the subscribers involved in `Test Breach 1`


scripts/notify-subscribers-of-breach.js
```diff
@@ -0,0 +1,33 @@
...
+      console.log(e);
+    }
+  }
+  return;
+}
+
+const breachName = process.argv[2];
+(async () => {
+  await notifySubscribersOfNewBreach(breachName);
+  console.log("hello");
```
pdehaan: 👋 Hello!

scripts/notify-subscribers-of-breach.js
```diff
@@ -0,0 +1,33 @@
...
+    } catch (e) {
+      console.log(e);
+    }
+  }
+  return;
+}
+
+const breachName = process.argv[2];
+(async () => {
+  await notifySubscribersOfNewBreach(breachName);
```
pdehaan: Should we do a null check on `process.argv[2]` somewhere around here?

db/utils.js
```diff
@@ -84,6 +84,18 @@ const DBUtils = {
     });
   },
 
+  async getSubscribersForBreach(breach) {
+    const breachedSubscribers = await breach
+      .$relatedQuery("email_hashes")
+      .whereNotNull("email");
+
+    if (breachedSubscribers.length) {
```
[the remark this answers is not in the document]
groovecoder: Huh, I guess so. I thought I had it that way at first but it caused an error in `notify-subscribers-of-breach.js`. But it works now.

db/utils.js
```diff
@@ -126,6 +138,10 @@ const DBUtils = {
   getBreachesForEmail(email) {
     return this.getBreachesForHash(getSha1(email));
   },
+
+  async getBreachByName(breachName) {
+    return (await Breach.query().where("name", breachName))[0];
```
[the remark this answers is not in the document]
groovecoder: Looks like Objections `QuerySet` objects handle this for us with some kind of fancy getter that returns `undefined` instead of throwing an error?
```
> breaches
[  ]
> breaches[0]
undefined
```

groovecoder:
I think I hit the main review comments, but I got stuck trying to figure out how to mark the `breached_hashes` record `notified=true` when sending the email. @nhnt11 - can you help me figure out how to set the extra data field on that `ManyToMany` relationship table?

I put my start of the code in https://github.com/mozilla/blurts-server/pull/94/commits/11e486745221d9ed6da09fe15ef7efc0550564c9

email-utils.js
```diff
@@ -3,6 +3,18 @@
...
+const hbs = require("nodemailer-express-handlebars");
+
+const hbsOptions = {
+  viewEngine: {
+    extname: ".hbs",
+    layoutsDir: __dirname + "/views/layouts/",
+    defaultLayout: "email",
+    partialsDir: __dirname + "/views/partials/",
+  },
+  viewPath: __dirname + "/views/email/",
```
pdehaan: Not sure how relevant this is, but this may need a `path.join()` type thing on these `__dirname` references to make it more cross-OS compatible.

routes/user.js
```diff
@@ -30,8 +30,12 @@ router.post("/add", urlEncodedParser, async (req, res) => {
   console.log(url); // Temporary for debugging.
 
   try {
-    await EmailUtils.sendEmail(email, "Firefox Breach Alert",
-      `Visit this link to subscribe: ${url}`);
+    await EmailUtils.sendEmail(
+      email,
+      "Verify your email address to subscribe to Firefox Monitor.",
+      "email_verify",
+      { email: email, url: url}
```
pdehaan: Not sure if we want some ESLint rule for this, but we can use shorthand notation here:

```js
{ email, url }
```

Ref: https://eslint.org/docs/rules/object-shorthand

scripts/notify-subscribers-of-breach.js
```diff
@@ -0,0 +1,46 @@
...
+        .$relatedQuery("breaches")
+        .where("breach_id", "=",  breach.id)
+        .patch({notified: true});
+      console.log("DONE setting notified=true");
+      // TODO: mark the subscriber as already notified for this breach
+    } catch (e) {
+      console.log(e);
+    }
+  }
+  return;
```
pdehaan: Not sure this does anything... ¯\\_(ツ)\_\/¯

scripts/notify-subscribers-of-breach.js
```diff
@@ -0,0 +1,46 @@
...
+}
+
+notifySubscribersOfNewBreach(breachName).then(()=>{return;});
```
pdehaan: What does the `.then(()=>return;});` do here?
It seems like it'd just ... return an undefined in the promise chain, but there isn't anything else in the chain that I see.

scripts/notify-subscribers-of-breach.js
```diff
@@ -0,0 +1,46 @@
...
+  console.log(`Found ${breachedSubscribers.length} un-notified subscribers in the breach`);
+  for (const subscriber of breachedSubscribers) {
+    console.log(`Sending email to subscriber ID: ${subscriber.id} ...`);
+    const subject = `${subscriber.email} was involved in the ${breach.name} breach.`;
+    try {
+      await EmailUtils.sendEmail(
+        subscriber.email,
+        subject,
+        "breach_notification",
+        { email: subscriber.email, breach: breach }
```
pdehaan: Should this pass the entire `breach` object, or just the `breach.name` property?

I think breach_notification.hbs only has the following:

```
Uh-oh, {{ email }} was involved in {{ breach.name }} data breach.
```

views/email/breach_notification.hbs
```diff
@@ -0,0 +1,2 @@
+{{!< email }}
+Uh-oh, {{ email }} was involved in {{ breach.name }} data breach.
```
pdehaan: Should this be wrapped in a `<p>...</p>` tag, like the strings in email_verify.hbs below?

routes/user.js
groovecoder: Hmm ... I think I like `object-shorthand` when all of the object properties' values have the same name, but not when there's a mix of same-name values and others. So I'll fix this one, but won't add the rule to `.eslintrc`.

scripts/notify-subscribers-of-breach.js
groovecoder: Yeah, this is just me trying to figure out how to get the script to return. Currently, when it runs, it does everything but just hangs at the end. The process never returns. :/

scripts/notify-subscribers-of-breach.js
groovecoder: this is just me trying to figure out how to get the script to return

scripts/notify-subscribers-of-breach.js
groovecoder: This is a habit of mine from Django templates ... I find that returning an entire object to templates tends to be more future-proof. So if we need to change the template to access other object properties or methods in the future, we don't have to change the corresponding route.

views/email/breach_notification.hbs
groovecoder: We need to do an entire round of UI design & implementation on these emails. These `.hbs` templates are just the start.

groovecoder:
Okay, I introduced a new `BreachedHash` model to set its `notified=true` when we send the breach notification.

groovecoder:
I also added a `process.exit` after the `await notifySubscribers...` call to ensure the script returns.

As far as I can tell, this PR is ready to merge for testing on stage. We will need to `knex migrate:rollback` and `knex migrate:latest` again to reset the stage DB.

db/utils.js
```diff
@@ -123,8 +122,11 @@ const DBUtils = {
     return await this.addBreachedHash(breachName, getSha1(email));
   },
 
-  async setBreachedHashNotified(breach, subscriber) {
-    return await BreachedHash.query().patch({notified: true}).where("breach_id", "=", breach.id).andWhere("sha1_id", "=", subscriber.id);
+  async setBreachedHashNotified(breach, email) {
+    return await breach
+      .$relatedQuery("email_hashes")
+      .where("sha1", getSha1(email))
+      .patch({ notified: true });
```
groovecoder: This is how I first tried to do this, but it doesn't actually work. ...
```
blurts=# select * from breached_hashes;
 id | sha1_id | breach_id | notified
----+---------+-----------+----------
  1 |       1 |         1 | f
  2 |       3 |         1 | f
  3 |       2 |         1 | f
  4 |       3 |         2 | f
  5 |       2 |         2 | f
  6 |       4 |         2 | f
  7 |       4 |         3 | f
  9 |       1 |         3 | f
  8 |       2 |         3 | f
(9 rows)
```
Then I ran it: 
```
groovetop:blurts-server lcrouch$ node scripts/notify-subscribers-of-breach.js AllMusic
Attempting to get SMTP credentials from environment...
Found breach:  Breach {
  id: 3,
  name: 'AllMusic',
  meta:
   { date: '6 December 2015',
     dataClasses: 'Email addresses, IP addresses, Passwords, Usernames, Website activity',
     acCount: '1,436,486' } }
Found 1 un-notified subscribers in the breach
Sending email to subscriber ID: 2 ...
DONE sending email.
Setting notified=true ...
DONE setting notified=true
```
But the database table still has `false`/`f` in the `notified` column:
```
blurts=# select * from breached_hashes;
 id | sha1_id | breach_id | notified
----+---------+-----------+----------
  1 |       1 |         1 | f
  2 |       3 |         1 | f
  3 |       2 |         1 | f
  4 |       3 |         2 | f
  5 |       2 |         2 | f
  6 |       4 |         2 | f
  7 |       4 |         3 | f
  9 |       1 |         3 | f
  8 |       2 |         3 | f
(9 rows)
```

That's why I resorted to using an explicit `BreachedHash` model. It did the trick.

db/utils.js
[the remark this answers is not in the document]
nhnt11: Ah right, we are accessing an element that may not exist but we're not doing anything with it so it doesn't need a null check here - that's up to the consumer of this function. 👍

PR #96, opened 2018-04-24
New Scan results view

nhnt11:


db/migrations/initial_schema.js
```diff
@@ -10,7 +10,7 @@ exports.up = knex => {
     .createTable("breaches", table => {
       table.increments("id").primary();
       table.string("name").unique();
-      table.string("meta");
+      table.json("meta");
```
groovecoder: Does this need to happen in a new/different migration? Or did you mean for us to destroy & re-create the initial schema on the staging site with this?

Issue #101, opened 2018-05-07
Change /scan code to use HIBP hash range query


groovecoder:
To avoid downloading entire breach hashsets, when a user scans for their email address:

1. [x] Check session cache for previously-fetched results
2. [x] Send the request along to HIBP email hash range query endpoint (Exact URL TBD)
3. [x] Cache results into a session cache to improve next scan performance
4. [x] Extract the user's results (if any) from the full response
5. [x] Display the user's results (if any) on the /scan page

groovecoder:
Until the exact HIBP API is known, we can stand up a mock API URL to be replaced with the real URL as soon as it's ready.

PR #104, opened 2018-05-08
for #101: start /hibp/.../range/ stub api and HIBP module to scan with hibp api

groovecoder:


routes/home.js
```diff
@@ -27,8 +27,16 @@ router.get("/monitor", urlEncodedParser, handleIndexRoute);
 router.post("/scan", urlEncodedParser, async (req, res) => {
   const email = req.body.email;
   let foundBreaches;
+
   if (email) {
-    foundBreaches = await DBUtils.getBreachesForEmail(email);
+    if (req.session.scanResults && req.session.scanResults[email]) {
+      foundBreaches = req.session.scanResults[email];
+    } else {
+      req.session.scanResults = {};
```
pdehaan: I'm a bit confused.

L32 says, if you have a `req.session.scanResults` object, and an `email` key is found, then do _X_.
Else, create an empty `req.session.scanResults` object.

Is it possible I'd have a `req.session.scanResults` object, but no match for that `email` key? It almost looks like if `req.session.scanResults` is truthy, but `req.session.scanResults[email]` is falsey, you'd erase the `req.session.scanResults` object.

Is it possible that you could have results for multiple email addresses in the `req.session.scanResults` object? Or is this strictly just caching results?

groovecoder: I updated this code to not clear out the whole `req.session.scanResults` object when the submitted `email` isn't found.

We may have results for multiple email addresses, if a user submits multiple email addresses.

db/utils.js
```diff
@@ -49,6 +49,20 @@ const DBUtils = {
     }
   },
 
+  async _getSha1EntriesFromPrefixAndDo(sha1Prefix, aFoundCallback, aNotFoundCallback) {
+    console.log("sha1Prefix: ", sha1Prefix);
+    const existingEntries = await EmailHash.query().where("sha1", "like", sha1Prefix + "%").eager("breaches");
```
pdehaan: `LIKE` queries kind of scare me...
What's the chance/risk that somebody could somehow pass a short `sha1Prefix` (like "a") and somehow get a _lot_ of results?

db/utils.js
```diff
@@ -144,6 +168,11 @@ const DBUtils = {
   async getBreachByName(breachName) {
     return (await Breach.query().where("name", breachName))[0];
   },
+
+  async getBreachesByNames(breachNames) {
+    return await Breach.query().where("name", "in", breachNames).debug(true);
```
pdehaan: What does `.debug(true)` do?

db/utils.js
groovecoder: Good catch! Although these are just stub API endpoints never intended for production, I'll go ahead and check that the passed value is a full 6-character prefix.

groovecoder:
Okay, I think this is actually ready for review & merge.

db/utils.js
```diff
@@ -49,6 +49,24 @@ const DBUtils = {
     }
   },
 
+  async _getSha1EntriesFromPrefixAndDo(sha1Prefix, aFoundCallback, aNotFoundCallback) {
+    // Only accept 6-character hash prefixes so requests:
+    // 1. can't get more than the intended hash range results FROM us
+    // 2. can't reveal a more specific hash query TO us
+    if (sha1Prefix.length !== 6) {
```
pdehaan: Now that `6` is defined in 2 places (in 2 different files), not sure if we should give it a `const` in AppConstants, then tweak this to something like:

```js
if (sha1Prefix.length !== AppConstants.SHA_PREFIX_LENGTH) {
  // ...
}
```

hibp.js
```diff
@@ -0,0 +1,50 @@
...
+      // Parse response body, format:
+      // {breachedAccount1sha1}:{breach1Name}[,{breach2Name},...]
+      // {breachedAccount2sha1}:{breach3Name}[,{breach4Name},...]
+      for (const breachedAccount of response.body.split("\n")) {
+        if (!breachedAccount) {
+          break;
+        }
+        const split = breachedAccount.split(":");
+        const breachedSha1 = split[0];
+        const breaches = split[1].split(",");
```
pdehaan: You could _almost maybe kinda_ do an array destructure here, but not sure it really improves on readability or brevity:

```js
const [breachSha1, breachStr] = breachedAccount.split(":");
const breaches = breachStr.split(",");
if (sha1 === breachedSha1) {
  foundBreaches = await DBUtils.getBreachesByNames(breaches);
}
```

db/utils.js
nhnt11: It might also be worth considering putting all the HIBP stuff in one file (a utils file?), and require that wherever it's necessary. No strong feelings about this though.

Edit: never mind, things got a bit clearer as I read more of the code in this PR

db/utils.js
```diff
@@ -49,6 +49,24 @@ const DBUtils = {
...
+      return await aNotFoundCallback();
+    }
+    const existingEntries = await EmailHash.query().where("sha1", "like", sha1Prefix + "%").eager("breaches");
+
+    if (existingEntries.length && aFoundCallback) {
+      return await aFoundCallback(existingEntries);
+    }
+
+    if (aNotFoundCallback) {
```
nhnt11: This will be called even if we got existing entries, if the caller of the function didn't supply `aFoundCallback`. We need to either return early before this check or add a `!existingEntries.length` to the condition - I prefer the former.

db/utils.js
```diff
@@ -134,6 +152,16 @@ const DBUtils = {
       return await aEntry
         .$relatedQuery("breaches")
         .orderBy("name");
+    }, async () => {
+      return [];
+    });
+  },
+
+  async getBreachesForHashPrefix(sha1Prefix) {
+    return await this._getSha1EntriesFromPrefixAndDo(sha1Prefix, async aEntries => {
```
nhnt11: Is this the only time we do this? The naming of _getSha1EntriesFromPrefixAndDo is confusing to me based on what it does. It seems like what it's really doing is getting *breaches*? and the two callbacks do the same thing - we can just:

a) get rid of this function,
b) rename _getSha1EntriesFromPrefixAndDo to getBreachesForHashPrefix,
c) make it just return either the entries or the empty array

and things would still work the same, no? Let me know if I'm missing something :)

hibp.js
```diff
@@ -0,0 +1,50 @@
...
+    console.log(`Fetching ${url}...`);
+
+    try {
+      const response = await got(url, {headers});
+      // Parse response body, format:
+      // {breachedAccount1sha1}:{breach1Name}[,{breach2Name},...]
+      // {breachedAccount2sha1}:{breach3Name}[,{breach4Name},...]
+      for (const breachedAccount of response.body.split("\n")) {
+        if (!breachedAccount) {
+          break;
```
nhnt11: Do we want to break here or continue? Not sure what this check is for, actually, when will it be falsey?

hibp.js
nhnt11: I like it! However, we should not be using split, because I wouldn't put faith in none of the breaches having a colon in their name. We should explicitly get the index of the first colon and make two string slices using it.

groovecoder: Ah, that's true. I'll update.

groovecoder: I went ahead with `split(":", 2)` to stop it splitting on the first `:` character.

hibp.js
groovecoder: I saw it falsey after the last line of the response. A newline at the end of the response makes an extra last element of the array which was empty, which was throwing an error when trying to `split` on it.

db/utils.js
groovecoder: Yeah these functions confused me. (I based this function on the existing `_getSha1EntryAndDo` function) I started to re-write it but quickly gave up since this prefix-only version of the function is only used by the hibp stub API, which we should be able to throw away.

db/utils.js
groovecoder: I chose the latter approach so I didn't accidentally change any caller's behavior by assuming when I should return early.

I'd honestly like to get rid of these `get...AndDo` functions completely. I find `andDo(... [callback, ...])` functions hard to fit into my brain. 😐

groovecoder:
Updated with other fixes, but I still have those 2 questions/comments.

hibp.js
pdehaan: If that's the case, shouldn't we be calling `.trim()` on the `response.body`?

groovecoder: Oh yeah ... great point.

hibp.js
```diff
@@ -29,15 +29,11 @@ const HIBP = {
...
       // {breachedAccount2sha1}:{breach3Name}[,{breach4Name},...]
-      for (const breachedAccount of response.body.split("\n")) {
-        if (!breachedAccount) {
-          break;
-        }
-        const split = breachedAccount.split(":");
-        const breachedSha1 = split[0];
-        const breaches = split[1].split(",");
+      for (const breachedAccount of response.body.trim().split("\n")) {
+        const [breachedSha1, breachNamesStr] = breachedAccount.split(":", 2);
```
nhnt11: This still doesn't work, for example, if breachedAccount is "a:b:c:d:e", split(":", 2) will return ["a", "b"] and omit the rest. What we want is ["a", "b:c:d:e"]

hibp.js
```diff
@@ -0,0 +1,48 @@
...
+    try {
+      const response = await got(url, {headers});
+      // Parse response body, format:
+      // {breachedAccount1sha1}:{breach1Name}[,{breach2Name},...]
+      // {breachedAccount2sha1}:{breach3Name}[,{breach4Name},...]
+      for (const breachedAccount of response.body.trim().split("\n")) {
+        const sliceIndex = breachedAccount.indexOf(":");
+        const breachedSha1 = breachedAccount.slice(0, sliceIndex);
+        const breachNamesStr = breachedAccount.slice(sliceIndex + 1);
+        const breachNames = breachNamesStr.split(",");
```
pdehaan: Not sure if this is a rare case where RegEx would have solved all the world's problems...

```js
for (const breachedAccount of response.body.trim().split("\n")) {
  const [, breachedSha1, rest] = breachedAccount.match(/^(.*?):(.*)$/);
  const breachNames = rest.split(",");

  console.log(breachedSha1, `${JSON.stringify(breachNames)}`);
}
```

PR #112, opened 2018-05-22
update hibp to use real range API

groovecoder:
* Adds `HIBP_STAGE` env vars
* Adds handlebars template helpers to format the real HIBP breach metadata
* Updates `hibp` module to call the real stage endpoint for email hash range queries
* Removes `get-hashsets` script as we won't use it anymore

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,7 +8,6 @@
   },
   "dependencies": {
     "arg": "^2.0.0",
-    "aws-sdk": "^2.241.1",
     "body-parser": "^1.18.3",
     "client-oauth2": "^4.2.1",
     "client-sessions": "^0.8.0",
@@ -21,9 +20,7 @@
     "nodemailer": "^4.6.4",
     "nodemailer-express-handlebars": "^3.0.0",
     "objection": "^1.1.9",
-    "pg": "^7.4.3",
-    "popsicle": "^9.2.0",
-    "request": "^2.86.0"
+    "pg": "^7.4.3"
   },
   "devDependencies": {
     "coveralls": "^3.0.1",
--- a/routes/oauth.js
+++ b/routes/oauth.js
@@ -5,8 +5,8 @@ const AppConstants = require("../app-constants");
 const ClientOAuth2 = require("client-oauth2");
 const crypto = require("crypto");
 const express = require("express");
+const got = require("got");
 const bodyParser = require("body-parser");
-const popsicle = require("popsicle");
 
 const DBUtils = require("../db/utils");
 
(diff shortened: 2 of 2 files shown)


hbs-helpers.js
```diff
@@ -0,0 +1,26 @@
+"use strict";
+
+const HBSHelpers = {
+  init(hbs) {
+
+    hbs.registerHelper("prettyDate", function(date, options) {
```
pdehaan: Me thinks `options` is unused.
This also feels a bit too locale specific. What about something like `toLocaleDateString()`?

```js
return new Date(date).toLocaleDateString('en-US', {year: 'numeric', month: 'long', day: 'numeric'});
```

hibp.js
```diff
@@ -17,25 +17,23 @@ const HIBP = {
 
     const sha1 = getSha1(email);
     const sha1Prefix = sha1.slice(0, 6);
-    const url = `${AppConstants.HIBP_STUB_API_ROOT}/breachedaccount/range/${sha1Prefix}`;
+    const url = `${AppConstants.HIBP_STAGE_API_ROOT}/range/${sha1Prefix}?code=${AppConstants.HIBP_STAGE_API_TOKEN}`;
```
pdehaan: I'm guessing the `HIBP_STAGE_API_TOKEN` is just a basic alpha-numeric string and doesn't need to be special URL-encoded?

groovecoder:
Addressed comments, and added 0d99beb7bfaf1943ebe5aad8ed91ecfcdb8694e3 to replace the last use of `popsicle` with `got` so we're using `got` everywhere. (I removed the last use of `request` when I removed the `get-hashsets` script.)

package.json
```diff
@@ -22,8 +21,7 @@
     "nodemailer-express-handlebars": "^3.0.0",
     "objection": "^1.1.9",
     "pg": "^7.4.3",
-    "popsicle": "^9.2.0",
-    "request": "^2.86.0"
+    "popsicle": "^9.2.0"
```
nhnt11: Can we get rid of popsicle from package.json now?

views/scan.hbs
```diff
@@ -63,10 +63,11 @@
...
           <p>
-            <span class="bold">Breach date:</span> {{ meta.date }}<br>
-            <span class="bold">Compromised data:</span> {{ meta.dataClasses }}<br>
-            <span class="bold">Compromised accounts:</span> {{ meta.acCount }}<br>
-            {{ meta.description }}
+            <span class="bold">Breach date:</span> {{prettyDate meta.BreachDate }}<br>
+            <span class="bold">Compromised data:</span> {{breachDataClasses meta.DataClasses }}<br>
+            <span class="bold">Compromised accounts:</span> {{localeString meta.PwnCount }}<br>
+            {{! TODO: make sure meta.Description is a safe string so we can use triple-{{{ }}
+            {{{ meta.Description }}}
```
pdehaan: Should we use something like [**dompurify**](https://github.com/cure53/DOMPurify) to sanitize the `meta.Description`?

package.json
groovecoder: oh, yeah sorry I meant to do that.

views/scan.hbs
groovecoder: We should. Should it be part of this PR or a new issue?

pdehaan: I'll file a new issue, but we should probably address before release.

groovecoder:
Merging. I'll fix up stage server when it deploys. It will need new required env vars.

PR #121, opened 2018-05-29
fix #21: serve Dockerflow responses

### LINKED ISSUE #21: Add Dockerflow standard endpoints
I.e., [`/app/version.json` object](https://github.com/mozilla-services/Dockerflow/blob/master/docs/version_object.md) at `/__version__`, `/__heartbeat__`, and `/__lbheartbeat__` as described here:

https://github.com/mozilla-services/Dockerflow#containerized-app-requirements

groovecoder:
Using @pdehaan's suggestions in #21.

### DIFF
--- a/package.json
+++ b/package.json
@@ -14,6 +14,7 @@
     "dotenv": "^5.0.1",
     "express": "^4.16.2",
     "express-hbs": "^1.0.4",
+    "git-rev-sync": "^1.12.0",
     "got": "^8.3.1",
     "hbs": "^4.0.1",
     "knex": "^0.14.6",
@@ -36,7 +37,7 @@
   "engines": {
     "node": ">=8"
   },
-  "homepage": "https://github.com/mozilla/blurts-server#readme",
+  "homepage": "https://github.com/mozilla/blurts-server",
   "license": "ISC",
   "main": "server.js",
   "nyc": {
--- a/routes/dockerflow.js
+++ b/routes/dockerflow.js
@@ -1,0 +1,41 @@
+"use strict";
+
+const express = require("express");
+const fs = require("fs");
+const path = require("path");
+const {version, homepage} = require("../package.json");
+
+
+const router = express.Router();
+
+let commit;
+try {
+  commit = require("git-rev-sync").short();
+} catch (err) {
+  console.error("Error in git-rev-sync: ", err);
+}
+
+const versionJsonPath = path.join(__dirname, "..", "public", "version.json");
+const versionJson = {
+  commit,
+  source: homepage,
+  version,
+};
+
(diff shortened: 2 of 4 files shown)


groovecoder:
@Micheletto - do our stage & prod app environments allow filesystem access? So we can serve our `__version__` endpoint by creating & reading the local `version.json` file?

Also, do our stage & prod app environments have `.git` in their working directory? So we can use `git-rev-sync` module to load the git commit data?

package.json
```diff
@@ -14,6 +14,7 @@
     "dotenv": "^5.0.1",
     "express": "^4.16.2",
     "express-hbs": "^1.0.4",
+    "git-rev-sync": "^1.12.0",
```
pdehaan: I think we're missing updates to the package-lock.json file if we're adding a module, no?

routes/dockerflow.js
```diff
@@ -0,0 +1,41 @@
...
+const {version, homepage} = require("../package.json");
+
+
+const router = express.Router();
+
+let commit;
+try {
+  commit = require("git-rev-sync").short();
+} catch (err) {
+  console.error("Error in git-rev-sync: ", err);
```
pdehaan: I think everything looks 👌!
If for some reason **git-rev-sync** fails, the `commit` will be undefined/null and be omitted from the version.json file.

Micheletto:
> @Micheletto - do our stage & prod app environments allow filesystem access? 
Yes.

> So we can serve our __version__ endpoint by creating & reading the local version.json file?
Yes.

> Also, do our stage & prod app environments have .git in their working directory? 
At build time, and if you are packaging it up in the docker container.

groovecoder:
@Micheletto - Okay, I'll leave this as it is for now.

Do I understand correctly that something like Circle CI will build the image that will be deployed? So as long as we don't put `.git` into our `.dockerignore`, it will build with `.git` in it?

package.json
groovecoder: Added `package-lock.json` to this PR.

Micheletto:
> Do I understand correctly that something like Circle CI will build the image that will be deployed?

Yes.  And publish it to Docker hub.  

> So as long as we don't put .git into our .dockerignore, it will build with .git in it?

I believe you also need to make sure you're copying it into /app.

groovecoder:
Makes sense. [I'm currently doing `COPY . /app`](https://github.com/mozilla/blurts-server/pull/122/files#diff-3254677a7917c6c01f55212f86c57fbf) so it should get the `.git` directory too.

I could use some help debugging a permission denied error though:

https://github.com/mozilla/blurts-server/pull/122#discussion_r191555052

PR #123, opened 2018-05-29
wip: Template updates

lesleyjanenorton:
First pass on reworking templates to match the mock-ups. Still need to clean up the css, rework the arrow graphic, get the 'sign up for alerts' button functionality going again, and add media query break-point styling.


public/css/app.css
```diff
@@ -1,5 +1,372 @@
...
+.demi {
+  font-weight:400;
+}
+
+.medium {
+  font-weight:500;
+}
+
+.bold {
+  font-weight: 700;
```
pdehaan: Should this be "600"?
I only see us importing the 300/400/500/600 font weights:

```css
@import url('https://fonts.googleapis.com/css?family=Fira+Sans:300,400,500,600');
```

views/layouts/default.hbs
```diff
@@ -1,77 +1,36 @@
 <!DOCTYPE html>
 <html lang="en">
   <head>
+    
     <meta charset="utf-8">
     <meta name="viewport" content="width=device-width, initial-scale=1.0">
     <title>{{ title }}</title>
+    <link href="https://fonts.googleapis.com/css?family=Fira+Sans:300,400,500,600" rel="stylesheet">
```
pdehaan: Does this need to be specified in the HTML _and_ in the CSS?

```css
@import url('https://fonts.googleapis.com/css?family=Fira+Sans:300,400,500,600');
```

&mdash; per https://github.com/mozilla/blurts-server/pull/123/files?short_path=e0d7738#diff-f2f4863f79a70d8706d0df15af6012d0R1

views/monitor.hbs
```diff
@@ -1,4 +1,25 @@
 {{!< default }}
+<div id="banner-info" class="section-container">
+  <div id="banner-sections-wrapper">
+    <div id="banner-left">
+      <div class="banner-content">
+        {{> protect_yourself }}
+        </div>
+    </div><!--closes banner-left-->
+    <div id="banner-right">
+      <img id="arrow-divider" src="img/banner_arrow.svg" alt="content divider" />
```
pdehaan: We could probably use an empty `alt=""` here, since this alt text doesn't add much (and may confuse screen readers).

views/partials/protect_yourself.hbs
```diff
@@ -0,0 +1,13 @@
...
+    <form action="/scan" method="post">
+    <div class="input-group">
+        <input class="input-group-field email-to-hash" type="email" name="email" placeholder="Enter Email Address">
+        <div class="input-group-button">
+            <img src="img/Search.svg" alt="search icon" />
+            <input id="submit-email" type="submit" class="button" value="Search">
+        </div>
+    </div>
+    <span>Your email will not be stored. Find out more in our <a href="
+    https://www.mozilla.org/en-US/privacy/firefox/" alt="Mozilla Firefox privacy policy">Privacy Policy.</a></span>
```
pdehaan: nit: we can probably remove the `/en-US/` path in the URL and let the mozilla.org site redirect to the best available locale.

views/partials/what_to_do.hbs
```diff
@@ -0,0 +1,18 @@
+<ul id="what-to-do">
```
pdehaan: Should this be an `<ol>` if we're adding numbers to each item anyways?

views/scan.hbs
```diff
@@ -1,4 +1,37 @@
...
+<div class="content-wrapper">
+  <div class="column-container-twothirds">{{> compromised_accounts }}</div>
+  <div class="column-container-third">{{> what_to_do }}</div>
+{{ else }}
+  {{> tips }}
+{{/if}}
+
+
+
+{{!-- {{!< default }}
```
pdehaan: Should we just remove this code instead of commenting it out?
(same comment applies to above sections w/ commented out code)

views/scan.hbs
```diff
@@ -1,4 +1,37 @@
...
+        {{> sign_up }}
+      </div>
+    </div><!--close banner-right -->
+  </div>
+</div>
+
+{{#if foundBreaches }}
+<div class="content-wrapper">
+  <div class="column-container-twothirds">{{> compromised_accounts }}</div>
+  <div class="column-container-third">{{> what_to_do }}</div>
```
pdehaan: Are we missing a closing `</div>` here?

views/partials/tips.hbs
```diff
@@ -1,4 +1,175 @@
...
+  <h5>The Key to Account Security</h5>
+  <p>Your passwords protect all of the personal information that resides in your online accounts, from bank balances and credit card numbers to your home address and private messages. Passwords are so valuable that thousands of passwords are stolen every day and accounts are traded on the black market.  Take these six steps to protect yours.</p>
+  </div>
+</div>
+<div id="tips-container" class="section-container">
+    <div class="tips-wrapper">
+      <div class="tip-container">
+          <div class="tip">
+              <div class="svg-background">
+                <img src="img/Password_Tip_1@2x.png" alt="Account Security Tip #1">
```
pdehaan: Should we delete the .svg files now? I think if we're switching to PNG images, they're potentially orphaned:

```diff
- <img src="img/Tip1.svg">
+ <img src="img/Password_Tip_1@2x.png" alt="Account Security Tip #1">
```

views/partials/tips.hbs
```diff
@@ -1,4 +1,175 @@
...
+              <p>You can't prevent a data breach, but you can limit your exposure by always using different passwords for different websites.</p>
+          </div><!--closes tip-->
+      </div><!--closes tip-container 1-->
+
+      <!-- tip # 2 -->
+
+      <div class="tip-container">
+          <div class="tip">
+              <div class="svg-background">
+                <img src="img/Password_Tip_2@2x.png" alt="Account Security Tip #2">
```
pdehaan: I'll let @nhnt11 or @groovecoder make an official call, but should these `alt=""` tags say this ^^^, or something like:

```html
alt="Create strong passwords"
```
... where that text is the `div.tip-header` text below (minus HTML formatting tags).

views/partials/tips.hbs
```diff
@@ -1,4 +1,175 @@
...
+      <div class="tip-container">
+          <div class="tip">
+              <div class="svg-background">
+                <img src="img/Password_Tip_4@2x.png" alt="Account Security Tip #4">
+                <span class="tip-numeral medium">4</span>
+              </div>
+              <div class="tip-header">
+                  <h3>Use a password<br>manager</h3>
+              </div>
+              <p>Password manager like 1Password, LastPass, or Dashlane can generate strong passwords for you, remember them for you, and fill them into websites so you don't have to type them in.</p>
```
pdehaan: Should we link the password managers to their respective sites?

views/partials/tips.hbs
groovecoder: That sounds like a good tweak to change the `alt` content to match the `div.tip-header` content.

views/partials/tips.hbs
groovecoder: @pdolanjski can tell us for sure, but I say no - don't link to the external sites. These are good products, but we shouldn't link off to external sites unless/until we've done some more thorough review of their products. A link could be interpreted as an endorsement.

views/partials/tips.hbs
```diff
@@ -1,4 +1,175 @@
...
+              <div class="svg-background">
+                <img src="img/Password_Tip_1@2x.png" alt="Account Security Tip #1">
+                <span class="tip-numeral medium">1</span>
+              </div>
+              <div class="tip-header">
+                  <h3>Use a different password<br>for every account</h3>
+              </div>
+              <p>You can't prevent a data breach, but you can limit your exposure by always using different passwords for different websites.</p>
+          </div><!--closes tip-->
+      </div><!--closes tip-container 1-->
```
pdehaan: This is a lot of duplicated code, but I'm not sure if somehow converting the tips into a JSON object and trying to do loops and templates makes it worse and more convoluted.

```js
{
  "tips": [
    {
      "numeral": 1,
      "image_src": "img/Password_Tip_1@2x.png",
      "image_alt": "Use a different password for every account",
      "header": "<h3>Use a different password<br/>for every account</h3>",
      "description": "<p>You can't prevent a data breach, but you can limit your exposure by always using different passwords for different websites.</p>"
    },
   {"numeral": 2, ...},
   {"numeral": 3, ...},
   {"numeral": 4, ...},
   {"numeral": 5, ...},
   {"numeral": 6, ...}
  ]
}
```

We probably don't even need `"numeral"` and can just use the array index. And since the `"image_src"` has the index in the filename, we can probably lose that field too (as long as the HBS loop has access to it's index).

Even the `"image_alt"` could probably be generated by stripping out any HTML tags from the `"header"` field (using DOMPurify or a RegEx or whatever)...

pdehaan:
Wow, I think this is amazing!
By moving out the logic into separate templates, this is _much_ easier to read and follow.

views/partials/tips.hbs
pdehaan: Mentioning them is an endorsement/recommendation here. The link is just a courtesy at this point.
But yeah, I guess they could just right click the name and select the "Search Google for 1Password" text, then click in the Google search results to get to the site. A 2-step detour (versus just `<a href="https://1password.com" target="_blank" rel="nofollow noopener noreferrer">...</a>` &mdash; or whatever our magic incantation would be).

views/partials/tips.hbs
lesleyjanenorton: I also considered this and I'm not sure which is the better option! It would definitely be cleaner and less 'repeated code stenchy' to bring them in as a JSON object. I will have to explore the looping in Handlebars - I'm brand new to .hbs and am still finding my way around! I can't imagine that you can't access the index (but you never know).

views/partials/what_to_do.hbs
lesleyjanenorton: Oh it's definitely an '<ol>'. What a great, great catch. I'm embarrassed and now thinking back to other times I have misused list tags.

lesleyjanenorton: whoops, it didn't appreciate my quotation marks evidently...There should be an <ol> there.

lesleyjanenorton: `<ol>` (I had to have Luke help me).

lesleyjanenorton:
This is very 'work-in-progressy.' Will be creating the false door div dynamically later and need to make more changes to reflect the previous code review comments.

views/partials/tips.hbs
pdehaan: There is an HBS looping example at https://github.com/mozilla/blurts-server/blob/d963172fdc5fe3b5dc5d2a02977e5c109715d573/views/scan.hbs#L58-L74

views/partials/tips.hbs
```diff
@@ -1,4 +1,175 @@
...
+              <p>We'll let you know if your account information is compromised in a data breach or exposed to hackers in some other way.</p>
+          </div><!--closes tip-->
+      </div><!--closes tip-container 6-->
+  </div>
+  </div>
+
+
+
+
+  {{!-- <div class="cell small-1"></div>
```
pdehaan: ✂️

views/monitor.hbs
```diff
@@ -1,4 +1,24 @@
...
+        {{> sign_up }}
+      </div>
+    </div><!--close banner-right -->
+  </div>
+</div>
+  {{> tips }}
+
+
+
+{{!-- {{!< default }}
```
pdehaan: ✂️

views/scan.hbs
```diff
@@ -1,4 +1,38 @@
...
+{{ else }}
+  {{> tips }}
+{{/if}}
+</div>
+
+
+
+{{!-- {{!< default }}
```
pdehaan: ✂️

views/scan.hbs
```diff
@@ -1,4 +1,38 @@
...
+</div>
+
+{{#if foundBreaches }}
+<div class="content-wrapper">
+  <div class="column-container-twothirds">{{> compromised_accounts }}</div>
+  <div class="column-container-third">{{> what_to_do }}</div>
+{{ else }}
+  {{> tips }}
+{{/if}}
+</div>
```
pdehaan: Something feels off here...
I _think_ either the opening `div.content-wrapper` needs to be outside the `{{#if foundBreaches }}`, or this closing `</div>` needs to be within the `{{#if}}`.

views/layouts/default.hbs
lesleyjanenorton: So I'm wondering if the font import needs to be there at all. If the plan is still to test this initially on Windows only desktops, then I think (but need to revisit the Photon documentation) the font specified for Windows users is "Segoe UI" which is potentially a system font that doesn't need to be imported. So maybe we can do away with this entirely.

views/partials/tips.hbs
nhnt11: I think there's a problem with the copy here - it should be "Password manager**s**" (plural), no?

lesleyjanenorton:
pushing updates to pull some weeds, definitely not a finished product!

public/js/test.js
```diff
@@ -41,26 +41,47 @@ function removeUser() {
...
+//     emailBtn.disabled = false;
+//   } else {
+//     emailBtn.disabled = true;
+//   }
+// }
+
+
+
+function showFalseDoor(){
+  const falseDoorBlurb = "<div class='section-container'><h4>Thank you for trying Firefox Monitor</h4><p>FireFox Monitor is a concept we are testing. We hope to provide the service to everyone soon.</p><p>Stay up-to-date with Firefox Monitor and other new features when you sign up for the <a href='https://www.mozilla.org/en-US/newsletter/firefox/'>Firefox newsletter.</a></p><button class='button' id='close-false-door'>Close</button></div>";
```
pdehaan: nit: remove "/en-US/" from the URL and let mozilla.org handle the magic.

views/layouts/default.hbs
pdehaan: Also, should we link out a 3rd party google site, or should we use the Fira CDN?
https://github.com/mozilla/Fira#usage
I don't have a lot of knowledge here to know if our https://code.cdn.mozilla.net/fonts/fira.css CDN is the same as the Google version and has all the same weights. Although I guess I could try loading both fonts and see what the network responses are...

lesleyjanenorton:
Work in progress push - still cleaning, and working out the header sign up button.

.vscode/settings.json
```diff
@@ -0,0 +1,4 @@
+{
+    "editor.tabSize":2,
+    "files.insertFinalNewline": true
+}
```
pdehaan: Me thinks you may want to add ".vscode" to the .gitignore, or possibly your global .gitignore file. (https://help.github.com/articles/ignoring-files/)

groovecoder:
The circleci failure is my fault. I sent https://github.com/mozilla/blurts-server/pull/130 which I hope fixes it up.

public/img/Tip1.svg
```diff
@@ -1,86 +0,0 @@
...
-</g>
-<g>
-</g>
-<g>
-</g>
-<g>
-</g>
-<g>
-</g>
-</svg>
```
pdehaan: Out of curiosity, why the switch from SVG to PNG? Was it purely a file size thing?

views/partials/banner_breach.hbs
```diff
@@ -0,0 +1,14 @@
+<div class="account">
+  <div id="breach-title-wrapper">
+    <div class="image-wrap">
+      <img alt="placeholder"src="img/{{ breach.name }}.png">
```
pdehaan: nit: Should we use `alt="{{ breach.name }}"` instead of `alt="placeholder"`? (or even an empty `alt=""` since the logo alt doesn't add much in the way of a11y in this case.

views/partials/compromised_accounts.hbs
```diff
@@ -0,0 +1,17 @@
...
+      </div>
+      <div class="breach-info-wrap">
+        <h3 class="demi">{{ name }}</h3>
+        <span class="breach-info"><span class="medium">Breach date:</span> {{prettyDate meta.BreachDate }}</span>
+        <span class="breach-info"><span class="medium">Compromised data:</span> {{breachDataClasses meta.DataClasses }}</span>
+        <span class="breach-info"><span class="medium">Compromised accounts:</span> {{localeString meta.PwnCount }}</span>
+      </div>
+    </div>
+  {{/each}}
+  <span id="source-info">Breach data provided by <span id="pwned">';--</span><span class="medium"> Have I been Pwned?</span></span>
```
pdehaan: Should this be a link to their site?
Also, not sure if there is a way to hide the `';--` bit from the screenreader, but that's probably a separate a11y issue/question for the future.

views/scan.hbs
```diff
@@ -1,97 +1,32 @@
...
-    <div class="cell"></div>
+{{#if foundBreaches }}
+  <div class="section-container">
+    <div id="compromised-accounts-wrapper">
+    <div class="column-container-twothirds">{{> compromised_accounts }}</div>
+    <div class="column-container-third">{{> what_to_do }}</div>
+  </div>
+  {{ else }}
+    {{> tips }}
   </div>
```
pdehaan: I think this needs to be moved up into the `{{#if}}` block above, as it looks like it has unbalanced `<div>..</div>` tags.

views/partials/compromised_accounts.hbs
lesleyjanenorton: I think I asked about this and the answer was no linking, but I will double check. I'm not aware of any screenreader-only invisibility cloaks but I'll check around. Could always sub with an icon.

public/img/Tip1.svg
lesleyjanenorton: Not sure on this. The previous SVGs didn't match up with the icons used in the spec, but I'm not sure what brought about the PNG decision.

lesleyjanenorton:
corrections and updates to templates

groovecoder:
I can rebase this on fixed circle CI and resolve he conflicts.

routes/oauth.js
```diff
@@ -42,7 +42,6 @@ router.get("/init", jsonParser, (req, res) => {
 
 router.get("/confirmed", jsonParser, async (req, res) => {
   if (!req.session.state) {
-    // TODO: Needs better error message
```
pdehaan: nit: we may want to add this TODO back, or something.
The current error message is near useless and could probably stand a bit of improved copy. (although I'm happy to submit a new issue for that as well, and not block this PR)

views/layouts/default.hbs
```diff
@@ -4,75 +4,32 @@
...
-        </a>
-        <ul class="menu">
-          <li><a href="#start">Start</a></li>
-          <li><a href="#tips">Tips</a></li>
-        </ul>
-      </div>
-      <div class="top-bar-right">
-        <a class="button" data-open="subscribe-modal">Subscribe</a>
+  <body>
+    <header class ="top-bar">
```
pdehaan: nit: remove space after the `class`.

views/partials/protect_yourself.hbs
```diff
@@ -0,0 +1,20 @@
...
+<form action="/scan" method="post">
+    <div class="input-group">
+        <input class="input-group-field email-to-hash" type="email" name="email" placeholder="Enter Email Address">
+        <div class="input-group-button">
+            <img src="img/Search.svg" alt="search icon" />
+            <input id="submit-email" type="submit" class="button" value="Search">
+        </div>
+    </div>
+    <span>Your email will not be stored. Find out more in our <a href="
+    https://www.mozilla.org/privacy/firefox/" alt="Mozilla Firefox privacy policy">Privacy Policy.</a></span>
```
pdehaan: I _think_ we can lose the `alt=""` here, since I'm not sure that is supported on an `<a>` tag.
Or not sure if we meant `title=""` or `aria-label=""` (or whatever).

groovecoder:
Moving this PR to https://github.com/mozilla/blurts-server/pull/133 which has conflicts resolved.

PR #139, opened 2018-06-01
fix #138: show real breach data

### LINKED ISSUE #138: Implement real db query for /?breach=<domain.com>
STR:

1. Go to https://fx-breach-alerts.herokuapp.com/?breach=adobe.com

Expected results:

Page should show the breach data for adobe.com

Actual results:

Page shows the breach data for AllMusic.com - our test breach.

groovecoder:


lesleyjanenorton:
Hey when I tried to start this I got a No

routes/home.js
```diff
@@ -2,18 +2,18 @@
...
+const DBUtils = require("../db/utils");
+
 const router = express.Router();
 
-function handleIndexRoute(req, res) {
+async function handleIndexRoute(req, res) {
+  let breach = null;
+  if (req.query.breach) {
+    breach = await DBUtils.getBreachByName(req.query.breach);
+  }
```
nhnt11: The template assumes that a breach object exists, so if we don't set a value for breach here, how about responding with an error instead of render'ing anything?

lesleyjanenorton:
looks good!

routes/home.js
groovecoder: Since this is the `home` route, we shoudn't error when the optional `breach` parameter fails. But I wrapped the template in a `{{#if breach }}` check instead.

Issue #55, opened 2018-02-26
Security Checklist


jvehent:
Risk Management
---------------
* [ ] The service must have performed a Rapid Risk Assessment and have a Risk Record bug
* [ ] Public staging and production endpoints must be added to the [security baseline](https://github.com/mozilla-services/foxsec/blob/master/services/README.md)

Infrastructure
--------------

* [ ] Access and application logs must be archived for a minimum of 90 days
* [ ] Use [Modern](https://wiki.mozilla.org/Security/Server_Side_TLS#Modern_compatibility) or [Intermediate](https://wiki.mozilla.org/Security/Server_Side_TLS#Intermediate_compatibility) TLS
* [ ] Set HSTS to 31536000 (1 year)
  * `strict-transport-security: max-age=31536000`
* [ ] Set HPKP to 5184000 (60 days)
  * `Public-Key-Pins: max-age=5184000; pin-sha256="WoiWRyIOVNa9ihaBciRSC7XHjliYS9VwUGOIud4PB18="; pin-sha256="r/mIkG3eEpVdm+u/ko/cwxzOMo1bk4TyHIlByibiA5E="; pin-sha256="YLh1dUR9y6Kja30RrAn7JKnbQG/uEtLMkBgFF2Fuihg="; pin-sha256="sRHdihwgkaib1P1gxX8HFszlD+7/gTfNvuAybgLPNis=";`
    * Start with max-age set to 5 minutes (`max-age=300`) and increase progressively
    * The first two pins are for Digicert EV and DV roots, the last two are for Let's Encrypt X3 and X4 intermediates (LE is only used for backup)
  * [ ] If the service is not hosted under `services.mozilla.com`, it must be manually added to [Firefox's preloaded pins](https://dxr.mozilla.org/mozilla-central/source/security/manager/tools/PreloadedHPKPins.json#184).
* If service has an admin panels, it must:
  * [ ] only be available behind Mozilla VPN (which provides MFA)
  * [ ] require Auth0 authentication

Development
-----------
* [ ] Sign all release tags, and ideally commits as well
  * Developers should [configure git to sign all tags](http://micropipes.com/blog//2016/08/31/signing-your-commits-on-github-with-a-gpg-key/) and upload their PGP fingerprint to https://login.mozilla.com
[...]

Issue #144, opened 2018-06-01
Audit handlebars for possibly unused files


pdehaan:
Not sure if this file is still used. It feels a bit lorem-ipsum-y...

https://github.com/mozilla/blurts-server/blob/master/views/home.hbs

Looking at <kbd>$ git grep "res.render("</kbd>, it looks like we render the following templates:

```sh
routes/hibp-stubs.js:    res.render("range", {foundEntries});
routes/home.js:  res.render("monitor", {
routes/oauth.js:    res.render("confirm", {
routes/scan.js:  res.render("scan", {
routes/user.js:    res.render("add", {
routes/user.js:  res.render("confirm", {
```

It also looks like I can figure out which templates use layouts using <kbd>$ git grep '{{!<'</kbd>:

```sh
views/add.hbs:{{!< default }}
views/confirm.hbs:{{!< default }}
views/email/breach_notification.hbs:{{!< email }}
views/email/email_verify.hbs:{{!< email }}
views/home.hbs:{{!< default }}
views/monitor.hbs:{{!< default }}
views/partials/tips.hbs:{{!< default }}
views/scan.hbs:{{!< default }}
```

... and a list of partials via <kbd>$ git grep '{{>'</kbd>:

```sh
views/layouts/default.hbs:            {{> sign_up_header }}
views/monitor.hbs:                {{> banner_breach }}
views/monitor.hbs:                {{> protect_yourself }}
views/monitor.hbs:                {{> protect_yourself }}
views/monitor.hbs:                {{> sign_up }}
views/monitor.hbs:  {{> tips }}
views/scan.hbs:          {{> oh_no }}
views/scan.hbs:          {{> no_breaches }}
views/scan.hbs:        {{> sign_up }}
views/scan.hbs:    <div class="column-container-twothirds">{{> compromised_accounts }}</div>
views/scan.hbs:    <div class="column-container-third">{{> what_to_do }}</div>
views/scan.hbs:    {{> tips }}
```

pdehaan:
```sh
$ git ls-files views | grep ".hbs$"

views/add.hbs
views/confirm.hbs
views/email/breach_notification.hbs
views/email/email_verify.hbs
views/home.hbs
views/layouts/default.hbs
views/layouts/email.hbs
views/monitor.hbs
views/partials/banner_breach.hbs
views/partials/compromised_accounts.hbs
views/partials/no_breaches.hbs
views/partials/oh_no.hbs
views/partials/protect_yourself.hbs
[...]

PR #146, opened 2018-06-01
fix #143: hash email client-side before scanning 

### LINKED ISSUE #143: Hash email address on the client
For the shield study, we don't need to receive the user's plaintext email. Let's hash the email address on the client before sending it to the server.

groovecoder:
Also for #55: disable un-used routes and urls


public/js/monitor.js
```diff
@@ -75,9 +55,24 @@ function showFalseDoor(){
...
+}
 
+async function hashEmailAndSend(emailFormSubmitEvent) {
+  emailFormSubmitEvent.preventDefault();
+  const emailForm = emailFormSubmitEvent.target;
+  for (const emailInput of emailForm.querySelectorAll("input[type=email]")) {
+    emailInput.value = await sha1(emailInput.value);
+  }
+  emailForm.submit();
+}
```
nhnt11: This will change the value of the textbox to the hash, no? Instead, how about this:

- Change the name/id of the current email input box (or just get rid of the name/id)
- Add a hidden input with that name/id (name="email" I think)
- In this function, set the value of the hidden input to the hash, and then submit the form. This way the user won't see a hash in their UI.

public/js/monitor.js
```diff
@@ -25,42 +26,21 @@ function doXHR(aURL, aBodyObj, aAlertText, aDebug=true) {
...
-//   const emailBtn = document.getElementById("subscribe-email-btn");
-//   if (isValidEmail(e.target.value)) {
-//     emailBtn.disabled = false;
-//   } else {
-//     emailBtn.disabled = true;
-//   }
-// }
-
-
+*/
```
pdehaan: Why not just ✂️ the code instead of commenting it out?
Or are we planning on resurrecting it shortly?

public/js/monitor.js
groovecoder: Good call.

public/js/monitor.js
groovecoder: We will likely bring it back right after the shield study runs.

public/js/monitor.js
```diff
@@ -75,9 +55,24 @@ function showFalseDoor(){
 
 }
 
+async function sha1(message) {
+  const msgBuffer = new TextEncoder("utf-8").encode(message);
+  const hashBuffer = await crypto.subtle.digest("SHA-1", msgBuffer);
+  const hashArray = Array.from(new Uint8Array(hashBuffer));
+  const hashHex = hashArray.map(b => ("00" + b.toString(16)).slice(-2)).join("");
```
pdehaan: Could we use something like [`String#padStart()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/padStart) here?

Not sure if that'd reduce it to something like this, or if there is something else I'm not  understanding:
```js
const hashHex = hashArray.map(b => b.toString(16).padStart(2, "0")).join("");
```

routes/hibp-stubs.js
```diff
@@ -1,24 +0,0 @@
...
-
-router.get("/api/v3/breachedaccount/range/:hashPrefix", urlEncodedParser, async (req, res) => {
-  const hashPrefix = req.params.hashPrefix;
-
-  const foundEntries = await DBUtils.getBreachesForHashPrefix(hashPrefix);
-
-  if (!foundEntries.length) {
-    res.status(404).send("Not Found");
-  } else {
-    res.render("range", {foundEntries});
```
pdehaan: Is this file coming back? Or more to the point, does anything else use the views/range.hbs file, or is that an orphan?

public/js/monitor.js
pdehaan: I'd almost argue we should have some "shield-study" branch where we'll deploy from, and then we can keep "master" branch less commenty.

routes/hibp-stubs.js
groovecoder: Removed.

views/home.hbs
```diff
@@ -13,9 +13,10 @@
       Lorem ipsum dolor sit amet, consectetur adipiscing elit. Donec volutpat hendrerit !nibus.
     </p>
     <br>
-    <form action="/scan" method="post">
+    <form action="/scan" class="email-scan" method="post">
```
pdehaan: Per #144, are we even using this file for anything? I was struggling to find a reference to this template.

views/partials/protect_yourself.hbs
```diff
@@ -5,9 +5,10 @@
 {{else}}
   <p class="demi">The first step to keeping your online accounts safe is knowing what you’re up against. Enter your email to find out if your accounts have been compromised.</p>
 {{/if}}
-<form action="/scan" method="post">
+<form action="/scan" class="email-scan" method="post">
     <div class="input-group">
         <input class="input-group-field email-to-hash" type="email" name="email" placeholder="Enter Email Address">
```
nhnt11: We should get rid of the name attribute on this one, otherwise its value will be sent when the form is submitted.

views/partials/what_to_do.hbs
```diff
@@ -16,10 +16,11 @@
     <span class="bold">Subscribe</span> for alerts from Firefox Monitor to learn sooner about your compromised accounts.
   </li>
   <li>
-    <form action="/scan" method="post">
+    <form action="/scan" class="email-scan" method="post">
       <label class="medium">Scan another email address</label>
       <div class="input-group">
         <input class="input-group-field email-to-hash" type="email" name="email" placeholder="Enter Email">
+        <input type="hidden" name="emailHash">
```
pdehaan: Semi-related, but I submitted #147 to discuss why the form is duplicated in 2-3 places... and if it's possible to convert to a DRYer partial.

views/partials/protect_yourself.hbs
pdehaan: ☝️ great catch!
Also, we'd need to make that change in the other 2 email form templates (home.hbs and partials/what_to_do.hbs).

groovecoder: Updated to remove the `name` attribute from the forms.

public/js/monitor.js
groovecoder: I couldn't make this work quickly ... keeping the way I have it.

groovecoder:
Made all the updates needed. Code is still commented. If @nhnt11 and @Micheletto want to make a `shield-study` branch for Circle CI deployment until we go live, we could delete the code comments from that branch and keep them in `master`

PR #149, opened 2018-06-04
with spinner + new scan form partial

lesleyjanenorton:
- (#137) loader graphic appears in form button after a valid email address has been submitted/while data is loading
- (#147) consolidates email-scan form into partial, tweaks to css for different button styles in "what to do OL"

PR #151, opened 2018-06-06
Merged spinner

groovecoder:
Updated https://github.com/mozilla/blurts-server/pull/149 with merge of latest master.


views/partials/compromised_accounts.hbs
```diff
@@ -3,7 +3,7 @@
   {{#each foundBreaches }}
     <div class="account">
       <div class="image-wrap">
-        <img alt="{{ name }}" src="img/{{ name }}.png">
+        <img alt="{{ name }}" src="img/logos/{{ name }}.png">
```
pdehaan: I can file this as a separate issue if we don't want to block on this right meow, but...

This won't work for roughly 50% of our logos since we use a mix of SVG and PNG files.

@groovecoder I noticed in the HIBP /api/v2/breaches endpoint, that they expose a "LogoType" attribute of `"png"` or `"svg"`. Is that available in our breaches table DB meta somewhere?

```js
LogoType: "png"
```

views/partials/scan_form.hbs
```diff
@@ -0,0 +1,13 @@
...
+          <input class="input-group-field email-to-hash" type="email" name="email" placeholder="Enter Email Address">
+          <input type="hidden" name="emailHash">
+          <div class="input-group-button">
+              <img src="img/Search.svg" alt="search icon" />
+              <img class="loader" src="img/loader.gif" alt="loading data" />
+              <input id="submit-email" type="submit" class="button" value="Search">
+          </div>
+      </div>
+      <span>Your email will not be stored. Find out more in our <a href="
+      https://www.mozilla.org/privacy/firefox/" alt="Mozilla Firefox privacy policy">Privacy Policy.</a></span>
```
pdehaan: nit: remove `alt="..."` or change it to some `aria-label="..."` or something if we're doing this fo a11y reasons.

views/partials/compromised_accounts.hbs
groovecoder: Yup, I think that png/svg data should be in our DB meta field.

PR #154, opened 2018-06-07
Website logos fix

lesleyjanenorton:
fixes website logos, updates 'false-door' copy, and prunes  'spinner' branch.


scripts/load-breaches.js
```diff
@@ -33,7 +33,7 @@ async function handleBreachesResponse(response) {
 
     for (const breach of breachesJSON) {
       // purify the description going into the DB
-      breach.Description = DOMPurify.sanitize(breach.Description);
+      breach.Description = DOMPurify.sanitize(breach.Description, {ALLOWED_TAGS: []});
```
pdehaan: What is this change doing?
My gut is telling me it'd strip all the HTML tags in the description.

views/layouts/default.hbs
```diff
@@ -5,11 +5,8 @@
     <meta name="viewport" content="width=device-width, initial-scale=1.0">
     <title>{{ title }}</title>
     <link href="https://fonts.googleapis.com/css?family=Fira+Sans:300,400,500,600" rel="stylesheet" />
-
     <link rel="stylesheet" href="/css/vendor/foundation.min.css">
     <link rel="stylesheet" href="/css/app.css">
-    <script type="text/javascript" src="/js/vendor/jquery.js" defer></script>
-    <script type="text/javascript" src="/js/vendor/foundation.min.js" defer></script>
```
pdehaan: Are we getting rid of these for good, or just temporarily?
If the former, should we nuke the [/public/js/vendor](https://github.com/mozilla/blurts-server/tree/master/public/js/vendor) directory?

views/partials/what_to_do.hbs
```diff
@@ -1,4 +1,5 @@
 <ol id="what-to-do">
+  <li>What to Do</li>
```
pdehaan: Should we give this an `id=""` or `class=""` and try and avoid `#what-to-do li:first-of-type {}` in the CSS, or are there subtle differences?

scripts/load-breaches.js
lesleyjanenorton: Your gut is correct.

views/layouts/default.hbs
lesleyjanenorton: They can go as far I'm concerned. I have been nervous about pruning ancestral code/imports/templates but will go through and clean house today. Might be able to get rid of the foundation css too!

views/partials/what_to_do.hbs
lesleyjanenorton: Yeah, or maybe even taken out of the list and set as a sub head? There should be an `<lh>` tag.

pdehaan: Interesting, I’ve never even heard of the ‘<lh>’ tag.

lesleyjanenorton: It was almost a thing like 10 years ago and then disappeared. I don't know why it was never made official.

lesleyjanenorton: Also, in response to another review comment about the status of the images... after MUCH trying and failing to batch them into smaller, more reasonably sized assets (that still looked decent), I finally just copied and pasted from the add-on. Once I'm not so mired in google analytics grief I'll probably start trying to figure out a better way to approach them. Any thoughts/ideas are totally appreciated!

.env-dist
```diff
@@ -18,7 +18,7 @@ OAUTH_TOKEN_URI="https://oauth-stable.dev.lcip.org/v1/token"
 
 HIBP_API_ROOT="https://stage.haveibeenpwned.com/api/v2/"
 HIBP_API_TOKEN=""
-HIBP_STAGE_API_ROOT="https://stage-api.haveibeenpwned.com"
+HIBP_STAGE_API_ROOT="https://api.haveibeenpwned.com"
```
nhnt11: let's keep this stage-api in .env-dist; we should use the HIBP_API_ROOT for production and we can fix this up in another PR

lesleyjanenorton: agreed!

scripts/load-breaches.js
pdehaan: But what about all the breaches that have links in the description, a la https://fx-breach-alerts.herokuapp.com/?breach=Adobe

.env-dist
pdehaan: I'm confused... It almost looks reversed.
The `HIBP_API_ROOT` references stage.haveibeenowned.com and now the `HIBP_STAGE_API_ROOT` is pointing at api.haveibeenpwned.com, which looks more production-y.

scripts/load-breaches.js
nhnt11: This change was done intentionally to make sure we don't have links in the description. The alternative would have been to de-link-ify stuff from the client side. FWIW, the text content of the links are preserved after sanitize() :)

pdehaan: OK... but now I'm wondering why I'm still seeing links on https://fx-breach-alerts.herokuapp.com/?breach=Adobe because that means either the Heroku site isn't deploying properly, or something is cached, or the code isn't doing what we're expecting.

pdehaan: Also, this will make some breaches a bit awkward. Like in the case of [7k7k Breach](https://fx-breach-alerts.herokuapp.com/?breach=7k7k):

> In approximately 2011, it's alleged that the Chinese gaming site known as [7k7k](http://www.7k7k.com/) suffered a data breach that impacted 9.1 million subscribers. Whilst there is evidence that the data is legitimate, due to the difficulty of emphatically verifying the Chinese breach it has been flagged as "unverified". The data in the breach contains usernames, email addresses and plain text passwords. [Read more about Chinese data breaches in Have I been pwned.](https://www.troyhunt.com/handling-chinese-data-breaches-in-have-i-been-pwned/)

Note that the last line will theoretically not have the actual text linked, so we'll have some confusing messaging saying "Read more about..." where the user cannot, in fact, read more.

PR #160, opened 2018-06-07
Don't stash scan results in session cookie

nhnt11:


groovecoder:
Dang, was this causing a bug/breakage?

nhnt11:
@groovecoder Yeah, scan results were easily exceeding max cookie size resulting in angry browsers. We should probably store the actual results in local storage and then set a timestamp in the cookie that the server can check and tell the client if its cache is stale. To be fair though, the original purpose of caching was to reduce load on the server (and the HIBP endpoint) right? We have rate limiting set up on our deployment so I guess this isn't a huge deal.

groovecoder:
Reduce load on the server, and improve performance on the client. Not a big deal either way though.

PR #163, opened 2018-06-07
Update test-user.js

pdehaan:
Add bug number to TODO.


pdehaan:
Self merging, because I'm an animal.

PR #171, opened 2018-06-08
Optimize images using svgo and imagemin-cli

pdehaan:


lesleyjanenorton:
this is fantastic! and sorry for the delay here- my github notifications have been going to my personal email address (which I haven't been checking as religiously) instead of my mozilla account. updating now.

Issue #165, opened 2018-06-07
"Uncaught ReferenceError: $ is not defined" at monitor.js:71


pdehaan:
### Steps to reproduce:

1. Go to https://fx-breach-alerts.herokuapp.com/
1. Open devtools console.

Looks like we commented out the jquery.js and foundation.min.js `<script>` tags in 3615cc272e44f0fb57d054ac057edae4f152053a (and deleted them in https://github.com/mozilla/blurts-server/pull/154/files#diff-2e80970a1383333b75cabe1a8ab2729b):

https://github.com/mozilla/blurts-server/blob/3615cc272e44f0fb57d054ac057edae4f152053a/views/layouts/default.hbs#L10-L11

But looks like a jQuery reference in https://github.com/mozilla/blurts-server/blob/95d499881819c3e7b55e3ca1d692b6326c1a0175/public/js/monitor.js#L71

pdehaan:
I'm not seeing this on latest Heroku deploy, and I believe it was addressed by #175. Closing.

PR #175, opened 2018-06-08
Sign up button fix + Errors Fix + Alignment Fix

lesleyjanenorton:
quick fixes : 

-updates false door blurb
-fixes sign up button triggering 
-removes jquery/foundation reference error #165

### DIFF
--- a/public/css/app.css
+++ b/public/css/app.css
@@ -622,17 +622,13 @@ span.breach-info {
   display: none !important;
 }
 
-#source-info {
+.source-info {
   margin-top: calc(15px + 0.5vh + 0.5vw);
   font-size: 15px;
 }
 
-#pwned {
-  display: inline-block;
-  border: 1px solid white;
-  border-radius: 5px;
-  padding: 0 0.45rem 0.1rem 0.45rem;
-  margin: 0 0.5rem;
+.account .source-info {
+  font-size: 10px;
 }
 
 /* sign up button in header on {{breach}} */
--- a/public/css/app.css
+++ b/public/css/app.css
@@ -527,6 +527,8 @@ form span {
   border-radius: 9%;
   overflow: hidden;
   max-width: 60px;
+  width: 60px;
+  min-width: 60px;
 }
 
 span.breach-info {
--- a/public/js/monitor.js
+++ b/public/js/monitor.js
@@ -68,7 +68,7 @@ if(document.querySelector(".email-scan")){
   document.querySelector(".email-to-hash").addEventListener("input", enableBtnIfEmailValid);
 }
 
-$(document).foundation();
+
 
 document.querySelector("#sign-up").addEventListener("click", showFalseDoor);
 // document.querySelector("#subscribe-fxa-btn").addEventListener("click", doOauth);
--- a/public/js/monitor.js
+++ b/public/js/monitor.js
@@ -1,5 +1,6 @@
 /* eslint-env browser */
 
+
 "use strict";
 
 function isValidEmail(val) {
@@ -9,7 +10,7 @@ function isValidEmail(val) {
 }
 
 function enableBtnIfEmailValid(e) {
-  const emailBtn = document.getElementById("submit-button");
+  const emailBtn = document.getElementById("submit-email");
   if (isValidEmail(e.target.value)) {
     emailBtn.disabled = false;
   } else {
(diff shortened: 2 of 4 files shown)


public/css/app.css
```diff
@@ -527,6 +527,8 @@ form span {
   border-radius: 9%;
   overflow: hidden;
   max-width: 60px;
+  width: 60px;
+  min-width: 60px;
```
pdehaan: I'm not a CSS expert, but do we still need `min-width` and `max-width` if we're specifying the `width` attribute?

views/partials/compromised_accounts.hbs
```diff
@@ -13,5 +13,5 @@
       </div>
     </div>
   {{/each}}
-  <span id="source-info">Breach data provided by <span id="pwned">';--</span><span class="medium"> Have I been Pwned?</span></span>
+  <span class="source-info">Breach data provided by<span class="medium"> Have I been Pwned?</span></span>
```
pdehaan: Looking at these side-by-side, is it OK that one of these HIPB citations is in a `<span class="demi" />` and the other is in a `<span class="medium" />`?

views/partials/banner_breach.hbs
```diff
@@ -12,5 +12,6 @@
   <span class="breach-info"><span class="medium">Compromised data:</span> {{breachDataClasses breach.meta.DataClasses }}</span>
   <span class="breach-info"><span class="medium">Compromised accounts:</span> {{localeString breach.meta.PwnCount }}</span>
   <span class="breach-info">{{{ breach.meta.Description }}}</span>
+  <span class="source-info">Breach data provided by<span class="demi"> Have I been Pwned?</span></span>
```
pdehaan: I think I mentioned this elsewhere, but should we add a link to the HIPB website here?

public/js/monitor.js
```diff
@@ -31,7 +32,7 @@ function displayLoader(){
 }
 
 function showFalseDoor(){
-  const falseDoorBlurb = "<div class='section-container'><h4>Thank you for trying Firefox Monitor</h4><p>FireFox Monitor is a concept we are testing. We hope to provide the service to everyone soon.</p><p>Stay up-to-date with Firefox Monitor and other new features when you sign up for the <a href='https://www.mozilla.org/newsletter/firefox/'>Firefox newsletter.</a></p><button class='button' id='close-false-door'>Close</button></div>";
+  const falseDoorBlurb = "<div class='section-container'><h4>Thank you for trying Firefox Monitor</h4><p>Firefox Monitor is a concept we are testing. During this test, we are not storing email addresses. This means that while we will use your email to give you real results about data breaches, we will not keep your email to alert you in case of future breaches</p><p>We hope to provide this service soon, but in the meantime, you can stay up-to-date on Firefox Monitor and other new features when you sign up for the <a href='https://www.mozilla.org/newsletter/firefox/'>Firefox newsletter.</a></p><button class='button' id='close-false-door'>Close</button></div>";
```
pdehaan: Unrelated to your change, but more of a personal curiosity... Why dynamically create the DIV and append it to DOM and have to juggle event listeners and query selectors using JavaScripten, versus just putting this directly into the HTML to begin with and toggle it's visibility by appending/removing a CSS class?

public/js/monitor.js
```diff
@@ -68,8 +69,4 @@ if(document.querySelector(".email-scan")){
   document.querySelector(".email-to-hash").addEventListener("input", enableBtnIfEmailValid);
 }
 
-$(document).foundation();
```
pdehaan: If we're now going jQuery free, I think you can remove the jquery env from the .eslintrc config file:

https://github.com/mozilla/blurts-server/blob/349e39226e78f22d8761c5b6ae5e1a42ae11bb95/.eslintrc.js#L5

pdehaan: Oh, plus maybe nuke the [public/js/vendor](https://github.com/mozilla/blurts-server/tree/master/public/js/vendor) directory since we now have orphaned files.

pdehaan:
Merging and I will address the jQuery and vendor code in a separate PR.

public/css/app.css
lesleyjanenorton: you wouldn't think so and I'm not sure what was happening here but when width was not set explicitly the icons were totally distorted.

views/partials/compromised_accounts.hbs
lesleyjanenorton: Total oversight ! I'll fix this in next branch.

views/partials/banner_breach.hbs
lesleyjanenorton: I asked about this and was told that the agreement was to not link. I do think a link here would be helpful and make sense however.

public/js/monitor.js
pdehaan: PR #176 submitted.

public/css/app.css
pdehaan: CSS, am i right... ¯\\\_(ツ)\_/¯

PR #176, opened 2018-06-08
Remove vendor/js code and jQuery from ESLint config and tweak ESLint config

pdehaan:
Yay deleting 10.2k lines!


.eslintrc.js
```diff
@@ -1,14 +1,25 @@
...
     "eslint:recommended",
     "plugin:node/recommended",
   ],
   overrides: [
+    {
+      files: [
+        'public/js/**/*.js'
+      ],
+      env: {
+        browser: true,
```
pdehaan: Everything in public/js runs in a browser context, so moving config here so we don't need `/* eslint-env browser */` in any .js files.

scripts/add-breached-emails.js
```diff
@@ -16,8 +16,6 @@ if (args["--help"]) {
   console.log("Usage: node add-breached-emails.js [--extraTestEmail=<...>]");
   console.log("Adds test[1-3]@test.com emails to LinkedIn, Adobe, and AllMusic breaches.");
   console.log("--extraTestEmail also adds the supplied test email address and includes it in the LinkedIn, Adobe, and AllMusic breaches.");
-  // We can `process.exit()` here since it's a CLI script.
-  // eslint-disable-next-line no-process-exit
```
pdehaan: This rule is already disabled for any scripts/*.js files via the .eslintrc config file. Removing the noise/funk to keep things a bit cleaner.

lesleyjanenorton:
NICE!!!

PR #177, opened 2018-06-08
Make email address field required on scan form

### LINKED ISSUE #114: Email address is not required when scanning from the homepage
~~(Not sure what the deployed Git SHA version is, because it doesn't look like we have a [/\_\_version\_\_](https://fx-breach-alerts.herokuapp.com/__version__) endpoint on Heroku -- Ah, this is already filed as #21)~~

### Steps to reproduce:

1. Go to https://fx-breach-alerts.herokuapp.com/
1. Click the [Scan] button without entering an email address.

### Actual results:

> **So far, so good.**
> does not appear in our list of compromised accounts. That’s good news. But it doesn’t mean you’re safe - data breaches can happen at any time in a number of ways, so take action to protect your passwords and sign up for alerts from Firefox Monitor.

### Expected results:

Make sure there is at least 1 character is entered (and input is a valid email address) before the user can submit the form.
If you type "aa" in the field, you will get a tooltip saying:

firefox_breach_alerts__scan_results

pdehaan:
Fixes #114

### DIFF
--- a/views/partials/scan_form.hbs
+++ b/views/partials/scan_form.hbs
@@ -1,6 +1,6 @@
   <form action="/scan" class="email-scan" method="post" data-no-csrf>
       <div class="input-group">
-          <input class="input-group-field email-to-hash" type="email" name="email" placeholder="Enter Email Address">
+          <input class="input-group-field email-to-hash" type="email" name="email" placeholder="Enter Email Address" required="required">
           <input type="hidden" name="emailHash">
           <div class="input-group-button">
               <img src="img/Search.svg" alt="search icon" />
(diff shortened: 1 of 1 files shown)


views/partials/scan_form.hbs
```diff
@@ -1,14 +1,12 @@
...
           <input type="hidden" name="emailHash">
           <div class="input-group-button">
               <img src="img/Search.svg" alt="search icon" />
               <img class="loader" src="img/loader.gif" alt="loading data" />
               <input id="submit-email" type="submit" class="button" value="Search">
           </div>
       </div>
-      <span>Your email will not be stored. Find out more in our <a href="
-      https://www.mozilla.org/privacy/firefox/" alt="Mozilla Firefox privacy policy">Privacy Policy.</a></span>
+      <span>Your email will not be stored. Find out more in our <a href="https://www.mozilla.org/privacy/firefox/" aria-label="Mozilla Firefox privacy policy">Privacy Policy.</a></span>
```
pdehaan: Changed `alt=".."` to `aria-label=".."`.

views/partials/scan_form.hbs
```diff
@@ -1,14 +1,12 @@
   <form action="/scan" class="email-scan" method="post" data-no-csrf>
       <div class="input-group">
-          <input class="input-group-field email-to-hash" type="email" name="email" placeholder="Enter Email Address">
+          <input class="input-group-field email-to-hash" type="email" name="email" placeholder="Enter Email Address" required="required">
```
pdehaan: Not pretty, but that's browser validation for you...

firefox_monitor_required

PR #181, opened 2018-06-08
Removes Foundation CSS & Vendor File + Nit

lesleyjanenorton:
- removes Foundation, updates app.css to accomodate anything that was still relying on foundation.
- Fixes hibp nit
- fixes sign false door window positioning


pdehaan:
Probably remove this, if we’re dropping foundation.css
https://github.com/mozilla/blurts-server/blob/4b66776478e3e1f0817e47fccbf01f1e131b36a5/.stylelintrc#L3

PR #184, opened 2018-06-09
Routing changes

### LINKED ISSUE #183: /scan endpoint doesn't handle GETs at all
This is bad because if a user copy-pastes the URL to share with a friend, and they paste it as-is into their browser, they will see a scary error message "Cannot GET /scans".

We should
- Redirect to / when a GET /scan is attempted
- Redirect to / when a POST request is attempted on /scan without an emailHash param.

nhnt11:
Redirect all unhandled GET paths to index; also do so for POSTs to /scan that don't supply an emailHash or supply the empty string's sha1.

Fixes #183.

### DIFF
--- a/routes/home.js
+++ b/routes/home.js
@@ -6,7 +6,7 @@ const DBUtils = require("../db/utils");
 
 const router = express.Router();
 
-async function handleIndexRoute(req, res) {
+router.get("/", async (req, res) => {
   let breach = null;
   if (req.query.breach) {
     breach = await DBUtils.getBreachByName(req.query.breach);
@@ -15,9 +15,10 @@ async function handleIndexRoute(req, res) {
     title: "Firefox Monitor",
     breach: breach,
   });
-}
+});
 
-router.get("/", handleIndexRoute);
-router.get("/monitor", handleIndexRoute);
+router.use(async (req, res) => {
+  res.status(404).render("404");
+});
 
 module.exports = router;
--- a/routes/scan.js
+++ b/routes/scan.js
@@ -3,30 +3,33 @@
 const express = require("express");
 const bodyParser = require("body-parser");
 
-const HIBP = require("../hibp");
+const sha1 = require("../sha1-utils");
 
+const HIBP = require("../hibp");
 
 const router = express.Router();
 const urlEncodedParser = bodyParser.urlencoded({ extended: false });
 
-
+const blankStringSha1 = sha1("");
 
 router.post("/", urlEncodedParser, async (req, res) => {
   const emailHash = req.body.emailHash;
-  let foundBreaches = [];
 
-  if (!req.session.scanResults) {
-    req.session.scanResults = {};
+  if (!emailHash || emailHash === blankStringSha1) {
+    res.redirect("/");
+    return;
   }
 
-  if (emailHash) {
-    foundBreaches = await HIBP.getBreachesForEmail(emailHash);
-  }
+  const foundBreaches = await HIBP.getBreachesForEmail(emailHash);
 
   res.render("scan", {
     title: "Firefox Breach Alerts: Scan Results",
-    foundBreaches: foundBreaches,
+    foundBreaches,
   });
+});
+
+router.get("/", async (req, res) => {
+  res.redirect("/");
 });
 
 module.exports = router;
--- a/server.js
+++ b/server.js
@@ -67,11 +67,11 @@ app.use(sessions({
   },
 }));
 
-app.use("/", HomeRoutes);
 app.use("/", DockerflowRoutes);
 app.use("/scan", ScanRoutes);
 // app.use("/oauth", OAuthRoutes);
 // app.use("/user", UserRoutes);
+app.use("/", HomeRoutes);
 
 // EmailUtils.init().then(() => {
   const listener = app.listen(AppConstants.PORT, () => {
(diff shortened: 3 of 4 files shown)


lesleyjanenorton:
nice!

routes/home.js
```diff
@@ -15,9 +15,10 @@ async function handleIndexRoute(req, res) {
...
     breach: breach,
   });
-}
+});
 
-router.get("/", handleIndexRoute);
-router.get("/monitor", handleIndexRoute);
+router.get("*", (req, res) => {
+  res.redirect("/");
+});
```
pdehaan: A bit curious how this works. It seems like you just may have hijacked the 404 route, so now every broken image and favicon may try loading the homepage and returning a 200 status code. Maybe. 

I really need to try and fix up my dev envy during the all hands so I can try some of this locally again. But I’m almost curious enough to let y’all merge it, and we can test it out on heroku.

routes/scan.js
```diff
@@ -3,25 +3,30 @@
...
   let foundBreaches = [];
 
   if (!req.session.scanResults) {
     req.session.scanResults = {};
   }
 
-  if (emailHash) {
-    foundBreaches = await HIBP.getBreachesForEmail(emailHash);
-  }
+  foundBreaches = await HIBP.getBreachesForEmail(emailHash);
```
pdehaan: Should we just do a `const foundBreaches = ...` here and nuke L23 above? Not sure why we’d set to empty array and then overwrite here.

lesleyjanenorton:
Hey guys! What is the verdict on the router.get("*",... ?

pdehaan:
I still think it’s a risk to assume each 404 route should redirect to the home page, versus just explicitly watching the various GET routes and adding a new one for GET /scan which redirects to the homepage.
Or even having a regular 404 page which has some copy and a link to the homepage so the 404 will show up in our traffic logs in case it is a missing logo/asset.

routes/home.js
```diff
@@ -15,9 +15,10 @@ async function handleIndexRoute(req, res) {
     title: "Firefox Monitor",
     breach: breach,
   });
-}
+});
 
-router.get("/", handleIndexRoute);
-router.get("/monitor", handleIndexRoute);
```
pdehaan: Was anything using this now absent “/monitor” endpoint?
Also, did you still want to add the GET /scans route redirecting to the index route so users won’t get 404s on refreshes or redirects?

views/404.hbs
```diff
@@ -0,0 +1,11 @@
+{{!< default }}
+<div id="banner-info" class="section-container">
+    <div id="banner-sections-wrapper">
+        <div id="banner-left">
+            <div class="banner-content">
+              <h1>404: Page not found.</h1>
```
pdehaan: Good enough(tm) for shield, but we may want to enlist Michelle or somebody for some better long-term copy, and possibly a link back to homepage or embedded search/subscribe capabilities so this feels less dead-endy.

nhnt11: That's why I thought we could redirect to index for 404s :P 
I'm not a fan of the idea of putting "real" UI in a 404 page because of people possibly copying the link to share it.
Let me tinker a bit more with this before we merge.

routes/home.js
nhnt11: It's a legacy endpoint, I took the opportunity to remove it.

lesleyjanenorton:
Hey @nhnt11 on the subject of routing - I am running the addon (popup variants) in nightly and in at least one of the variants (screen-capped below) the 'Go to Firefox Monitor' is sending me to "/?breach=[examplewebsite.com]"  instead of "/?breach=[ExampleWebsiteName]". 

For instance, clicking 'Go to Firefox Monitor' from https://www.gawker.com lands me here:

```https://fx-breach-alerts.herokuapp.com/?breach=gawker.com ```

instead of here:

```https://fx-breach-alerts.herokuapp.com/?breach=Gawker```

Right now the server is set up to query the breaches by name, do we want to try and change that to query by something like meta.Domain? Or am I looking at the wrong/an old thing? Let me know how you want to proceed! 

doorhanger1

**Comment by lesleyjanenorton** [comment]
I'm late to the party like always.

File: server.js [hunk 5]
```diff
@@ -67,11 +67,11 @@ app.use(sessions({
   },
 }));
 
-app.use("/", HomeRoutes);
 app.use("/", DockerflowRoutes);
 app.use("/scan", ScanRoutes);
 // app.use("/oauth", OAuthRoutes);
 // app.use("/user", UserRoutes);
+app.use("/", HomeRoutes);
```
pdehaan: Personal curiosity, why did this line move? Was there a bug in the previous ordering?

nhnt11: I added app.use("*", ... ) in the home routes. This needs to be added last, otherwise it will hijack everything else. Hence the move.

nhnt11:
Quick recap: after the latest edition of this patch:

1. GET /scan redirects to index
2. Requests to unhandled endpoints are 404'd

pdehaan:
Sorry, I'm confused.... I thought you modified the [/scripts/make-sha1-hashes.js](https://github.com/mozilla/blurts-server/blob/master/scripts/make-sha1-hashes.js) file in an earlier commit, but now I can't find it...

pdehaan:
I'm seeing #201 as fallout from this, but I'm not convinced that it should hold up this PR, or is caused by this PR, or hasn't been there all along and I'm just terrible at my job (or all 3).

:+1: R+=me

PR #185, opened 2018-06-11
Google Analytics 

lesleyjanenorton:
hey guys, apologies in advance because I know it's All Hands and this is a bunch to look over!  

Bigger Things: 
- Added 'tracking-projections.js' that checks for user privacy settings and installs google analytics if tracking protection is not enabled.
- Added google analytics event handlers in 'monitor.js' (these could/maybe should be moved into their own file?).
- Modified helmet settings to accommodate google analytics.

Little(r) Things:
- Condensed Sign up button partials (button and text are no longer repeated).
- Minified Mozilla logo
- Moved false door html into it's own partial and out of monitor.js


public/js/monitor.js
```diff
@@ -1,45 +1,91 @@
...
+  const eventCategory = "Sign Ups";
+  const eventAction = "Clicked Sign Up";
+  if(document.getElementById("header-button")){
+    sendEvent(eventCategory, eventAction, "Sign up button clicked from featured breach.");
+  }
+  else if (document.getElementById("what-to-do")) {
+    sendEvent(eventCategory, eventAction, "Sign up button clicked after scanning email with breaches.");
+  }
+  else {
+    const headlineText = document.getElementsByTagName("h2")[0].innerHTML;
```
pdehaan: This feels scary. Do you just want to give that “h2” a unique id or class instead?

public/js/monitor.js
```diff
@@ -1,45 +1,91 @@
...
+    if(headlineText.includes("good")){
```
pdehaan: This seems scary too. We’d break the code if the wording changes, or we localize anything.

public/js/monitor.js
```diff
@@ -1,45 +1,91 @@
...
 }
 
-function removeLoader(){
-  if(document.getElementsByClassName("input-group-button")[0].classList.contains("loading-data")){
-    document.getElementsByClassName("input-group-button")[0].classList.remove("loading-data");
+function handleEmailScans(){
+  const eventCategory = "Scans";
+  const eventAction = "Scanned Email Address";
+  if(document.getElementById("header-button")){
+    sendEvent("User submitted email from a featured breach.");
```
pdehaan: This looks funny. I think it’s missing ‘eventCategory’ and ‘eventAction’, no?

public/js/tracking_protection.js
```diff
@@ -0,0 +1,48 @@
...
+		userPrivacySettings="Unspecified";
+	} 
+	else if(isIE && platform&&anomalousWinVersions.indexOf(platform.toString()) !== -1) {
+		userPrivacySettings="Unspecified";
+	}
+	else {
+		userPrivacySettings = {
+		0 : "Disabled",
+		1 : "Enabled",
+		}[userPrivacySettings] || "Unspecified";}
```
pdehaan: I had to read this a few times to try and understand the temp object and array index code here. Not sure if there is a clearer way to refactor this for legibility (since userPrivacySettings can be an int or string or already Unspecified.

pdehaan: Actually, why bother with Unspecified at all since we only ever check if it equals Enabled below?

pdehaan: Ah, missed the sneaky trailing `}`, which I believe closes the else statement. We may want to try running this through a formatter or something to ensure consistent indenting, etc. ... or leave it for now, and I can try running codebase through prettier.

lesleyjanenorton: I thought my powers of copy/paste/rename vars were stronger than this but clearly some things got lost in transit from Schalk Neethling's most excellent dnt-helper repo to my branch. I would in any case like to go through this later and make it more "regular human readable" in the hopes that passive javascripters might be less intimated/more inclined to use it.

PR #189, opened 2018-06-13
False door as partial & Condensed Sign Up Partials

lesleyjanenorton:
- Getting false door blurb out of monitor.js and into its very own partial.
- Condensing sign up button partials into one.

Part 2 of "Breaking up Metrics PR" to follow...


groovecoder:
Looks like this is in https://github.com/mozilla/blurts-server/pull/190 too, and I like the look of that one.

PR #190, opened 2018-06-13
Analytics + Tracking Protection Protector

lesleyjanenorton:
Google Analytics events on Sign up button clicks, email scans, and pageviews + 'tracking-projections.js' that checks for user privacy settings and installs google analytics if tracking protection is not enabled. Code from previous PR with tweaks.


public/js/monitor.js
```diff
@@ -1,4 +1,56 @@
...
+function sendEvent(category, action, label){
+  if(ga){
+  ga("send", "event", category, action, label);
+  }
+}
+
+function handleSignUpClicks(string){
+  const eventCategory = "Sign Ups";
+  const eventAction = "Clicked Sign Up";
+  if(window.location.href.indexOf("breach") > -1){
```
pdehaan: What is this meaning to do?
Are you trying to check for `?breach=Adobe` style things in the URL? If so, should we use something more specific, like `window.location.search.indexOf("breach=")`?

```js
console.log(window.location.href); // "https://fx-breach-alerts.herokuapp.com/?breach=Adobe"
console.log(window.location.search); // "?breach=Adobe"
```

Or even something similar to this oddity (per [`URLSearchParams#has()`](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams/has)):

```js
if (new URLSearchParams(window.location.search.substring(1)).has("breach")) {...}
```

public/js/monitor.js
```diff
@@ -1,4 +1,56 @@
...
+    sendEvent(eventCategory, eventAction, "Sign up button clicked from featured breach.");
+  }
+  else if (document.getElementById("what-to-do")) {
+    sendEvent(eventCategory, eventAction, "Sign up button clicked after scanning email with breaches.");
+  }
+  else {
+    if(window.location.href.indexOf("scan") > -1){
```
pdehaan: Same here. I think this scares me.

If it works, it's by chance and magic.
Consider https://fx-breach-alerts.herokuapp.com/?breach=Adobe
```js
console.log(window.location.href.indexOf("scan")); // -1
```

But if a new breach for the imaginary "Adobescan" site appears tomorrow, we may get some unexpected results (or not, I'm just being delightfully hypothetical):

https://fx-breach-alerts.herokuapp.com/?breach=Adobescan

```js
console.log(window.location.href.indexOf("scan")); // 52
```

It'd also break if we ever added some "/scanner" URL or something else vague.

I think this is fine today, because I didn't see any breaches with "scan" in the name in the current breaches.json:

```sh
$ cat ../blurts-addon/src/breaches.json | grep scan # returns zero results
```

public/js/monitor.js
```diff
@@ -1,4 +1,56 @@
...
+    else {
+      sendEvent(eventCategory, eventAction, "Sign up button clicked from landing page.");
+    }
+  }
+}
+
+function handleEmailScans(){
+  const eventCategory = "Scans";
+  const eventAction = "Scanned Email Address";
+  if(window.location.href.indexOf("breach") > -1){
```
pdehaan: See https://github.com/mozilla/blurts-server/pull/190/files#r194950688 comment above.

public/js/monitor.js
```diff
@@ -1,4 +1,56 @@
...
+  }
+  else {
+    sendEvent(eventCategory, eventAction, "User submitted email from homepage.");
+  }
+}
+
+function handlePageViews(){
+  const eventCategory = "Scan Results";
+  const eventAction = "Viewed Scan Results";
+  if(window.location.href.indexOf("scan") > -1){
```
pdehaan: See https://github.com/mozilla/blurts-server/pull/190/files#r194951250 comment above.

Also, since we're duplicatin' the `?breach=` and `/scan` logic in 2 places, not sure if it's worthy of a helper function, a la `function isScanPage() {...}` and `function isBreachPage() {...}`. Maybe not, I'm just sayin'.

public/js/tracking_protection.js
```diff
@@ -0,0 +1,46 @@
+/* eslint-env browser */
+"use strict";
+
+/* 
+	Respect user privacy settings and honor user privacy choices.
+	From Schalk Neethling at https://github.com/schalkneethling/dnt-helper */
```
pdehaan: Random thought, please feel free to hurl profanity at me in any channel you want, but...
Would it make more sense to include this dnt-helper 3rd party code via npm instead of copy/pasta?

Well, technically Schalk Neethling hasn't published to npm, but did generously put a package.json file in the repo, so we should be able to include this via GitHub using:

```sh
$ npm i schalkneethling/dnt-helper -S
```

Then we'd just have to copy the file from ./node_modules/ using something adjacent to this:

```js
$ cp ./node_modules/dnt-helper/js/dnt-helper.js public/js/vendor
```
We could even wrap that in some "postinstall" npm script, which copies that one vendor file into the correct location. Or, maybe this is well overkill and we revisit this after Shield launch.

public/js/monitor.js
lesleyjanenorton: You're right. Going to put IDs on the resulting h2s and check for those instead. Still cheesy, but feels less perilous for the meantime. Less magic, more muggle.

public/js/monitor.js
lesleyjanenorton: I thought about this and if my recollection serves it done got convoluted on account of the event labeling. Will reinvestigate.

public/js/tracking_protection.js
lesleyjanenorton: No I like this. It feels bonafide.

public/js/monitor.js
```diff
@@ -1,4 +1,56 @@
 "use strict";
+/* global ga */
+
+function sendEvent(category, action, label){
+  if(ga){
+  ga("send", "event", category, action, label);
```
pdehaan: nit: indenting looks off here.

pdehaan:
This PR has 2 merge conflicts, but it looks like they're from the other PR. Not sure if you want me to rebase those changes into this branch and resolve those conflicts, or if we should try and remove that single commit.

server.js
```diff
@@ -40,8 +40,8 @@ app.use(helmet.contentSecurityPolicy({
     defaultSrc: ["'none'"],
     fontSrc: ["https://code.cdn.mozilla.net/fonts/"],
     frameAncestors: ["'none'"],
-    imgSrc: ["'self'"],
-    scriptSrc: ["'self'"],
+    imgSrc: ["'self'", "https://www.google-analytics.com", "www.google-analytics.com"],
```
groovecoder: Do we need both `https://www...` and `www...`? I'd rather remove the protocol-relative `www...` so our CSP enforces that everything is fetched over `https`

lesleyjanenorton: Doesn't look like it! Removing..;.

groovecoder:
Conflicts are resolved and this looks good ...

groovecoder:
The GA & DNT work for me. This looks good to me.

public/js/tracking_protection.js
lesleyjanenorton: Going to merge as is for now and add this as an issue.

PR #206, opened 2018-06-19
Circle Appeasement

lesleyjanenorton:
So basically, Docker has no Git (npm ERR! code ENOGIT === Ain'tNoGit). The options are to imbue Docker with Git (which has its own set of potential issues) or to move dnt-helper.js into a vendor folder and be done with it (for now). Less elegant, but less grief for the time-being.


pdehaan:
I think we want to kill this line and commit the dnt script too, no?
https://github.com/mozilla/blurts-server/blob/03068c1887a589abd24dec1f5013e791edb20ea4/.gitignore#L9

pdehaan:
Merging... Did some public/js/vendor juggling in #211; _"Add back public/js/vendor folder"_.

Issue #150, opened 2018-06-05
Remove the 'AllMusic' test fixture breach?


pdehaan:
Randomly spotted, but something feels weird to me...

- https://fx-breach-alerts.herokuapp.com/?breach=AllMusic has some details, and a logo.
- https://mozilla.invisionapp.com/share/KEIV66OTS23#/screens/297625875 lists a description (but is just a Photoshop mockup, so not 100% reliable as a data source -- but seems odd that the description looks so legit)
- https://haveibeenpwned.com/PwnedWebsites doesn't seem to any mention of "AllMusic" or "All Music", so I can't even find the breach on HIBP site to see what description they have (if any).

![firefox_monitor_-_allmusic](https://user-images.githubusercontent.com/557895/40991505-d4296d1c-68a8-11e8-9e1a-ba81e7669848.png)

lesleyjanenorton:
I will investigate this. I thought the AllMusic Breach was an invented breach, created for test purposes only but it definitely doesn't need to be popping up at fx-breach-arlerts/herokuapp if that is the case.

pdehaan:
I don't see it on https://github.com/mozilla/blurts-addon/blob/54b42fc67071b023a1af808a4a32d76293a302e0/src/breaches.json either.

Yeah, this feels sketchy. If it _is_ a dummy breach, we definitely do NOT want to use a real company name and logo, because that's wrong and rude (even if it is difficult to randomly find).

pdehaan:
But I think you're right, https://github.com/mozilla/blurts-server/blob/da9a81d48a2696ce1c7c721f67a4380ee85632fb/scripts/load-breaches.js#L22-L28

I think we should definitely change the name/flag/logo and rename it something less legit sounding like "Firefox Monitor Test Breach", and make sure it doesn't appear on production.

lesleyjanenorton:
We'll get rid of it!

PR #209, opened 2018-06-19
getting breaches by domain

### LINKED ISSUE #215: The "Sign up for alerts" pop-up from the "Firefox Monitor" website cannot be dismissed by any other methods except the "Close" button
Via @Softvision-MariusComan in https://github.com/mozilla/blurts-addon/issues/53

---

**[Affected versions]:**
- Firefox 60.0.2

**[Affected Platforms]:**
- All Windows

**[Prerequisites]:**
- Have a new Firefox profile with the latest version of the "Firefox Monitor" add-on (0.01 custom built on 20.06.2018) installed.

**[Steps to reproduce]:**
1. Open the browser with the profile from prerequisites.
2. Navigate to the "https://fx-breach-alerts.herokuapp.com/" page.
3. Click the "Sign up for alerts" button.
4. Press the "Esc/Enter" key or click outside the pop-up area.
5. Observe the behavior.

**[Expected result]:**
- The pop-up is dismissed.

**[Actual result]:**
- Nothing happens.

**[Notes]:**
- Attached a screen recording of the issue: 
rec of the issue

lesleyjanenorton:
WIP / Needs cleaning - What do we think of getting the breaches by domain ("/?breach=siteDomain.com") instead of by Name? This will set us up for better future handling of sites with multiple breaches and help get the branch_variants PR one step closer to merged (site.Title just needs to be site.Domain in Add-on). We can figure out what to do with the additional breaches later. 

- The 'Example Breach' I had created to replace the All Music breach is commented out right now because it started giving grief after the database reorganization. Will return to it after knocking out some Add-on things. 

Fixes #215


README.md
```diff
@@ -52,7 +52,7 @@ for more background, objectives, key use cases.
     npm start
     ```
 
-2. Visit the `test.html` page at [localhost:6060/test.html](http://localhost:6060/test.html)
+2. Visit the `test.html` page at [localhost:6060] (http://localhost:6060)
```
pdehaan: I'd just remove the v confusing `test.html` reference.

db/migrations/initial_schema.js
```diff
@@ -11,7 +11,8 @@ exports.up = knex => {
     })
     .createTable("breaches", table => {
       table.increments("id").primary();
-      table.string("name").unique();
+      table.string("name");
+      table.string("domain");
```
pdehaan: I don't know enough about Knex migrations to know if we can add columns here, or if it needs to be in some migration file.

README.md
lesleyjanenorton: hahaha yeah I am inclined to agree

scripts/load-breaches.js
```diff
@@ -41,14 +41,17 @@ async function handleBreachesResponse(response) {
...
 (async () => {
-  if (args["--createAMBreach"]) {
-    await DBUtils.createBreach("AllMusic", {
-      Name: "AllMusic",
-      BreachDate: "2015-012-06",
-      DataClasses: ["Email addresses", "IP addresses", "Passwords", "Usernames", "Website activity"],
-      PwnCount: 1436486,
-    });
-  }
+  // if (args["--createAMBreach"]) {
```
pdehaan: nit: change this to `--createExampleBreach` to match L22 above.

lesleyjanenorton: Well well well. This explains why I was having such a bad time trying to get the Example Breach to load.

db/migrations/initial_schema.js
lesleyjanenorton: I had to do the psql dropdb/createdb song and dance and then it worked.

pdehaan: I'm also fine w/ ignoring this and submitting a follow up issue.

scripts/load-breaches.js
```diff
@@ -41,17 +41,17 @@ async function handleBreachesResponse(response) {
...
-  //     LogoType: "svg",
-  //     Description: "This is example breach data. Users arrive on this page by clicking the 'Go to Firefox Monitor' button from the Firefox Monitor Add-on (https://github.com/mozilla/blurts-addon) after visiting a site with known data breaches. ",
-  //   });
-  // }
+//   if (args["--createExample"]) {
+//     console.log('here');
+//     await DBUtils.createBreach("Example", {
+//       Name: "Example Breach Name",
+//       Domain: "examplebreachURL",
+//       BreachDate: "2018-012-06",
```
pdehaan: nit: i know this is commented out, but we should probably fix the date to "2018-12-06" (not "012")

scripts/load-breaches.js
```diff
@@ -11,19 +11,19 @@ const pkg = require("../package.json");
...
+// });
 
-if (args["--help"]) {
-  console.log("Usage: node load-breaaches.js [--createAMBreach]");
-  console.log("--createAMBreach creates the 'AllMusic' test fixture breach.");
-  process.exit();
-}
+// if (args["--help"]) {
+//   console.log("\n\n  Usage: node load-breaches.js [--createExample]");
+//   console.log("  --createExampleBreach creates an example breach.");
```
pdehaan: nit: I think this should be `--createExample` and not `--createExampleBreach`

Also, I think there are a couple more hidden references to AllMusic in **/scripts/add-breached-emails.js**:

1. https://github.com/mozilla/blurts-server/blob/fe08c44b25d13ac34841bd29a977860e6a249c37/scripts/add-breached-emails.js#L17-L18
1. https://github.com/mozilla/blurts-server/blob/fe08c44b25d13ac34841bd29a977860e6a249c37/scripts/add-breached-emails.js#L31-L34

lesleyjanenorton:
updated for issues (53, 54, 16) raised in the add-on repo
fixes #150 ,
fixes #212

lesleyjanenorton:
circle hate again?
hey circle, it's hip to be square.

pdehaan:
> circle hate again?

Error seems to be _"Tag 8.11.2-alpine not found in repository docker.io/library/node"_.

Not sure if we can just get away w/ bumping it to 8.11.3 (since I see that tag in https://store.docker.com/images/node/plans/e81cd2fb-4415-49b7-80ff-5ec7ac39431f?tab=tags).

public/js/monitor.js
```diff
@@ -57,59 +57,79 @@ function isValidEmail(val) {
...
 
 function enableBtnIfEmailValid(e) {
+  removeInvalidMessage();
   const emailBtn = document.getElementById("submit-email");
   if (isValidEmail(e.target.value)) {
       emailBtn.disabled = false;
       removeInvalidMessage();
   } else {
       emailBtn.disabled = true;
+      if(e.keyCode === 13) {
```
pdehaan: Per https://github.com/mozilla/blurts-addon/pull/57/files#r197748253

> `keyCode` is deprecated, let's do `e.code !== "Enter"`

public/js/monitor.js
```diff
@@ -57,59 +57,79 @@ function isValidEmail(val) {
...
       emailBtn.disabled = false;
       removeInvalidMessage();
   } else {
       enableBtnIfEmailValid(e);
   }
 }
 
-function showFalseDoor(){
+function closeByEscape(e) {
+  if(e.keyCode === 27 || e.keyCode === 13) {
```
pdehaan: Per https://github.com/mozilla/blurts-server/pull/209/files#r197845073 above

> `keyCode` is deprecated, let's do `e.code === "Escape" || e.code === "Enter"`

db/migrations/initial_schema.js
nhnt11: I'm fine with this (i.e. dropping the db and resetting) for now since we don't have any data we care about yet

public/js/monitor.js
lesleyjanenorton: so this is weird. e.code works just fine with "Escape" but does absolutely nothing with "Enter" ?

db/migrations/initial_schema.js
Micheletto: > Sure, works for me, as long as it will work in stage/prod without needing too much OPs work.

That's fine for now.  In future we won't want to do anything that way, and also we are getting pretty close to the wire so please let me know when there are configuration and ops related tasks to run any given version of the code we want released.

public/js/monitor.js
```diff
@@ -57,59 +57,78 @@ function isValidEmail(val) {
...
+  if(e.target.value === "") {
       emailBtn.disabled = false;
       removeInvalidMessage();
   } else {
       enableBtnIfEmailValid(e);
   }
 }
 
-function showFalseDoor(){
+function closeByEscape(e) {
```
nhnt11: Seems like we're only using this function once. How about using a named inline function like this:

```
window.addEventListener("keydown", function listener(e) {
  if (e.code !== "Enter" && e.code !== "Escape") {
    return;
  }
  addClass("#false-door", "hidden");
  window.removeEventListener("keydown", listener);
}
```

public/js/monitor.js
```diff
@@ -57,59 +57,78 @@ function isValidEmail(val) {
...
+  if(e.code === "Enter" || e.code === "Escape") {
+    addClass("#false-door", "hidden");
+    window.removeEventListener("keydown", (e) => {
```
nhnt11: This doesn't work - it tries to remove a NEW function (that happens to call closeByEscape within it). It should be `window.removeEventListener("keydown", closeByEscape)` but even that wouldn't work, because the event listener that was added was an anonymous function that calls closeByEscape within it. This combination of add/remove would work though:

```
window.addEventListener("keydown", closeByEscape);
window.removeEventListener("keydown", closeByEscape);
```

public/js/monitor.js
```diff
@@ -57,59 +57,78 @@ function isValidEmail(val) {
...
   handleEvents("SignUp");
   removeClass("#false-door", "hidden");
-  document.getElementById("close-false-door").onclick = function (){
-    addClass("#false-door", "hidden");
-  };
+  const falseDoor = document.getElementById("false-door");
+  window.addEventListener("keydown", (e) => {
+    closeByEscape(e);
+  });
+  falseDoor.addEventListener("click", (e) => {
```
nhnt11: The event listener never gets removed :)

We probably want to copy what we do with keydown here.

db/utils.js
```diff
@@ -16,18 +16,18 @@ Model.knex(knex);
...
-        .insert({ name, meta });
+        .insert({ name, domain, meta });
     } catch(e) {
       console.error(e);
-      if (e.code && e.code === "23505") {
-        // Duplicate error, silently log.
-        console.error(`Duplicate breach: ${name}`);
-        return;
-      }
+      // if (e.code && e.code === "23505") {
```
nhnt11: Why did you comment this out?

lesleyjanenorton:
updated!

db/utils.js
```diff
@@ -198,6 +198,10 @@ const DBUtils = {
     return await Breach.query().where("name", "in", breachNames);
   },
 
+  async getBreachesByDomain(breachDomain){
```
nhnt11: nit: space before `{`

public/js/monitor.js
```diff
@@ -57,59 +57,78 @@ function isValidEmail(val) {
...
 
 function enableBtnIfEmailValid(e) {
+  removeInvalidMessage();
   const emailBtn = document.getElementById("submit-email");
   if (isValidEmail(e.target.value)) {
       emailBtn.disabled = false;
       removeInvalidMessage();
   } else {
       emailBtn.disabled = true;
+      if(e.code === "Enter") {
```
nhnt11: nit: space after `if` and before `(` (this goes elsewhere in this patch as well)

scripts/load-breaches.js
```diff
@@ -1,6 +1,6 @@
 "use strict";
 
-const arg = require("arg");
+// const arg = require("arg");
```
nhnt11: Why are we removing the argument support? :)

pdehaan:
Is this still a thing, or is this PR replaced by #216?

lesleyjanenorton:
Closing this PR but leaving the branch in case we want to query by domain at some point in the future.

PR #219, opened 2018-06-27
Add in some missing pwned logos

nhnt11:
Also ran svgo on all svg logos in a separate commit.


pdehaan:
:+1: LGTM, but I don't see those breaches on the dev or stage server.
But these are bonus images, so lets merge away while we sync up the DB.

PR #221, opened 2018-06-28
toLowerCase the email before hashing

nhnt11:
Emails are not case sensitive so the service shouldn't be either.

### DIFF
--- a/public/js/monitor.js
+++ b/public/js/monitor.js
@@ -143,7 +143,7 @@ async function sha1(message) {
   const hashBuffer = await crypto.subtle.digest("SHA-1", msgBuffer);
   const hashArray = Array.from(new Uint8Array(hashBuffer));
   const hashHex = hashArray.map(b => ("00" + b.toString(16)).slice(-2)).join("");
-  return hashHex;
+  return hashHex.toUpperCase();
 }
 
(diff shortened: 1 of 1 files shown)


pdehaan:
It _looks_ logical...
For sanity checking sake, here's our current usage of `.toLowerCase()` and `.toUpperCase()` in this repo:

```sh
$ git grep "to.*erCase()"

hibp.js:    const sha1Prefix = sha1.slice(0, 6).toUpperCase();
hibp.js:        if (sha1.toUpperCase() === sha1Prefix + breachedAccount.hashSuffix) {
public/js/monitor.js:  return re.test(String(val).toLowerCase());
scripts/make-sha1-hashes.js:  shasum.update(trimmedString.toLowerCase());
```

lesleyjanenorton:
Looks good to me but are we sure it's necessary? I tried a few emails with and without the toLowerCase and all came back with the expected results but that could also be because they're cached (reminiscing about the .toUpperCase bug).

nhnt11:
I made it toUpperCase the hash before returning as well.

Let's be consistent with this. FWIW, current production deployment shows no results for scanning my email address in upper case, which is wrong.

lesleyjanenorton:
Great catch @nhnt11 . Merging.

PR #228, opened 2018-07-04
branding updates

lesleyjanenorton:
References to HIBP without the "?" and with links to the HIBP website.

### DIFF
--- a/views/partials/banner_breach.hbs
+++ b/views/partials/banner_breach.hbs
@@ -13,6 +13,6 @@
     <span class="breach-info"><span class="medium">Compromised Data</span> {{breachDataClasses breach.meta.DataClasses }}</span>
     <span class="breach-info"><span class="medium">Compromised Accounts</span> {{localeString breach.meta.PwnCount }}</span>
     <span class="breach-info">{{{ breach.meta.Description }}}</span>
-    <span class="source-info">Breach data provided by<span class="demi"> Have I Been Pwned?</span></span>
+    <span class="source-info">Breach data provided by<a class="demi" href="https://www.haveibeenpwned.com" alt="Have I Been Pwned"> Have I Been Pwned</a></span>
   </div>
 {{/if}}
--- a/views/partials/banner_breach.hbs
+++ b/views/partials/banner_breach.hbs
@@ -13,6 +13,6 @@
     <span class="breach-info"><span class="medium">Compromised Data</span> {{breachDataClasses breach.meta.DataClasses }}</span>
     <span class="breach-info"><span class="medium">Compromised Accounts</span> {{localeString breach.meta.PwnCount }}</span>
     <span class="breach-info">{{{ breach.meta.Description }}}</span>
-    <span class="source-info">Breach data provided by<a class="demi" href="https://www.haveibeenpwned.com" alt="Have I Been Pwned"> Have I Been Pwned</a></span>
+    {{> hipb_attribution }}
   </div>
 {{/if}}
--- a/views/partials/compromised_accounts.hbs
+++ b/views/partials/compromised_accounts.hbs
@@ -14,5 +14,5 @@
     </div>
   {{/each}}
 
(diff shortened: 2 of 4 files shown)


lesleyjanenorton:
Unsure on why this is failing on CircleCI... it looks like a lint issue but when I run lint locally I'm not turning anything up?

views/partials/banner_breach.hbs
```diff
@@ -13,6 +13,6 @@
     <span class="breach-info"><span class="medium">Compromised Data</span> {{breachDataClasses breach.meta.DataClasses }}</span>
     <span class="breach-info"><span class="medium">Compromised Accounts</span> {{localeString breach.meta.PwnCount }}</span>
     <span class="breach-info">{{{ breach.meta.Description }}}</span>
-    <span class="source-info">Breach data provided by<span class="demi"> Have I Been Pwned?</span></span>
+    <span class="source-info">Breach data provided by<a class="demi" href="https://www.haveibeenpwned.com" alt="Have I Been Pwned"> Have I Been Pwned</a></span>
```
pdehaan: `<a>` tags don't support `alt=""`. You could use `aria-label` if you want to set something special for a11y, but since the text and the `alt` are the same, you can probably just remove it here (and in the diff below).

pdehaan: Not technically a problem, per se, and barely even a nit, but...

It feels odd that we have a leading space at the beginning of the `<a>` tag's contents. Since we have a `text-decoration: none` in the CSS, it's a bit moot, but if we eventually refactor that to add default underlines or whatever, the link will have a leading space. Same with the link below. It almost may be nice to refactor this out into some "attibution" partial [eventually], as its fairly duplicated code (apart from the `<a class="demi">` versus `<a class="demi compromised">`).

views/partials/compromised_accounts.hbs
```diff
@@ -14,5 +14,5 @@
     </div>
   {{/each}}
 
-  <span class="source-info">Breach data provided by <span class="demi">Have I Been Pwned?</span></span>
+  <span class="source-info">Breach data provided by <a class="demi compromised" href="https://www.haveibeenpwned.com" alt="Have I Been Pwned">Have I Been Pwned</a></span>
```
pdehaan: For reasons currently unknown, i can't test this path. When I search a compromised email on localhost, it isn't returning any results (even if I run the <kbd>$ node scripts/add-breached-emails.js</kbd> script).

views/partials/banner_breach.hbs
```diff
@@ -13,6 +13,6 @@
     <span class="breach-info"><span class="medium">Compromised Data</span> {{breachDataClasses breach.meta.DataClasses }}</span>
     <span class="breach-info"><span class="medium">Compromised Accounts</span> {{localeString breach.meta.PwnCount }}</span>
     <span class="breach-info">{{{ breach.meta.Description }}}</span>
-    <span class="source-info">Breach data provided by<span class="demi"> Have I Been Pwned?</span></span>
```
pdehaan: Final nonsense random thought... Should we add a `target="_blank" rel="noopener noreferrer"` to open this in a new tab (so we don't lose our current page/results)?

lesleyjanenorton:
updated with HIBP attribution as partial

views/partials/hipb_attribution.hbs
```diff
@@ -0,0 +1 @@
+ <span class="source-info">Breach data provided by<a href="https://www.haveibeenpwned.com" target="_blank">Have I Been Pwned</a></span>
```
pdehaan: we should add the `rel="noopener noreferrer"` to the `<a target="_blank">` tag to avoid leaking referrer information.

Plus, I think we're missing a space between the "by" and the `<a>` tag.

lesleyjanenorton:
Fixed. And back in Travis' good graces evidently?

PR #235, opened 2018-07-11
Ui updates

lesleyjanenorton:
-Filters out unverified breaches that are classified as spam, retired, and/or sensitive. 
-Disables input while email scan form is processing.

### DIFF
--- a/hibp.js
+++ b/hibp.js
@@ -12,8 +12,8 @@ const HIBP_USER_AGENT = `${pkg.name}/${pkg.version}`;
 
 const HIBP = {
   async getBreachesForEmail(sha1) {
+    const filteredBreaches = [];
     let foundBreaches = [];
-
     const sha1Prefix = sha1.slice(0, 6).toUpperCase();
     const url = `${AppConstants.HIBP_STAGE_API_ROOT}/breachedaccount/range/${sha1Prefix}?code=${encodeURIComponent(AppConstants.HIBP_STAGE_API_TOKEN)}`;
     const headers = {
@@ -37,7 +37,12 @@ const HIBP = {
     } catch (error) {
       console.error(error);
     }
-    return foundBreaches;
+    for (const breach of foundBreaches) {
+      if ( breach.meta.IsVerified && !breach.meta.IsRetired && !breach.meta.IsSensitive && !breach.meta.IsSpamList ) {
+        filteredBreaches.push(breach);
+      }
+    }
+    return filteredBreaches;
   },
 };
 
--- a/hibp.js
+++ b/hibp.js
@@ -12,7 +12,6 @@ const HIBP_USER_AGENT = `${pkg.name}/${pkg.version}`;
 
 const HIBP = {
   async getBreachesForEmail(sha1) {
-    const filteredBreaches = [];
     let foundBreaches = [];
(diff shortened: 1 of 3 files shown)


public/js/monitor.js
```diff
@@ -158,6 +158,9 @@ async function hashEmailAndSend(emailFormSubmitEvent) {
   for (const emailInput of emailForm.querySelectorAll("input[type=email]")) {
     emailForm.querySelector("input[name=emailHash]").value = await sha1(emailInput.value);
     emailInput.value = "";
+    emailInput.disabled = true;
+    emailInput.placeholder="";
```
pdehaan: minorest of nits, but spaces around the `=` would make it easier to read (and more consistent).

hibp.js
```diff
@@ -37,7 +37,12 @@ const HIBP = {
     } catch (error) {
       console.error(error);
     }
-    return foundBreaches;
+    for (const breach of foundBreaches) {
+      if ( breach.meta.IsVerified && !breach.meta.IsRetired && !breach.meta.IsSensitive && !breach.meta.IsSpamList ) {
+        filteredBreaches.push(breach);
+      }
+    }
+    return filteredBreaches;
```
pdehaan: Silly comment, but not sure if it's easier to create a second array versus just doing something like this pseudo-shenanigans:

```js
return foundBreaches.filter(({meta}) => meta.IsVerified && !meta.IsRetired && !meta.IsSensitive && !meta.IsSpamList);
```

nhnt11: Why silly? I was going to suggest this :)

hibp.js
```diff
@@ -37,7 +37,12 @@ const HIBP = {
     } catch (error) {
       console.error(error);
     }
-    return foundBreaches;
+    for (const breach of foundBreaches) {
+      if ( breach.meta.IsVerified && !breach.meta.IsRetired && !breach.meta.IsSensitive && !breach.meta.IsSpamList ) {
```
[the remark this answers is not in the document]
nhnt11: Great insight, thanks! I've reached out to Troy for his comment on this.

public/js/monitor.js
lesleyjanenorton: hahaha i'll fix it. 'minorest of nits' might be my new favorite phrase.

hibp.js
lesleyjanenorton: way better! not silly at all.

pdehaan:
@nhnt11 Did we get any response from Troy as to what results we should show (re: `.IsVerified` and `.IsSpamList`)?

nhnt11:
Update: I proposed that we exclude breaches that are (spam lists || unverified || sensitive || retired); Troy agreed.

groovecoder:
Looks good. I updated some of my breaches to be spam, unverified, sensitive, and/or retired and the filter worked correctly on all cases.

lesleyjanenorton:
I'm still confused but I think the decision to filter out unverified breaches was partly in response to user testing videos where users were totally confused by the unverified breaches. In a few cases, users remarked that they'd never had an account on 'Anti Public Combo List' (or whatever the unverified breach was) and then started questioning all of the other results. If a decision is made to list and flag them later we'll probably need to do something similar to HIBP and provide additional information about what 'unverified' means in this context as well.

PR #240, opened 2018-07-20
more front end clean up

lesleyjanenorton:
more front end clean up // prep for forthcoming ui specs.

### DIFF
--- a/public/css/app.css
+++ b/public/css/app.css
@@ -4,7 +4,7 @@ body {
   margin: 0 0 0 0;
   padding: 0 0 0 0;
   color: #fff;
-  font-family: 'Fira Sans', sans-serif;
+  font-family: 'Fira Sans', helvetica, arial, sans-serif;
   font-weight: 300;
   box-sizing: border-box;
   font-size: 100%;
@@ -29,6 +29,7 @@ body {
   -webkit-font-smoothing: antialiased;
   font-size: 17px;
   position: relative;
+  overflow: scroll;
 }
 
 body,
@@ -108,7 +109,7 @@ input {
   font-weight: 400;
   line-height: 1.5;
   color: #0a0a0a;
-  transition: box-shadow 0.5s, border-color 0.25s ease-in-out;
+  transition: box-shadow 0.2s, border-color 0.2s ease-in-out;
   -webkit-appearance: none;
   -moz-appearance: none;
   padding: 0.5rem;
@@ -132,7 +133,7 @@ input {
   border-right-style: solid;
   border-right-color: transparent;
   border-radius: 3px;
-  transition: all 0.25s ease-out;
+  transition: all 0.2s ease-out;
   font-size: 0.9rem;
   line-height: 1;
   text-align: center;
@@ -140,6 +141,14 @@ input {
   background-color: #0060df;
   color: #fefefe;
   width: 100%;
+}
+
+img {
+  user-select: none;
+  -webkit-user-select: none;
+  display: inline-block;
+  vertical-align: middle;
+  -ms-interpolation-mode: bicubic;
 }
 
 input::placeholder {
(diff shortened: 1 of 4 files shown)


public/css/app.css
```diff
@@ -651,14 +651,19 @@ form span {
   justify-content: center;
   align-items: center;
   background-color: rgba(0, 0, 0, 0.38);
+  transition: all 0.25s ease-out;
```
pdehaan: microscopic nit: but my eyes are noting that almost all other transitions seem to be 0.2s and this is an outlier at 0.25s.
That's probably fine and all, just noting the difference.

```sh
$ git grep -n "transition: "

public/css/app.css:87:  transition: all cubic-bezier(0.07, 0.95, 0, 1);
public/css/app.css:91:  transition: all cubic-bezier(0.07, 0.95, 0, 1);
public/css/app.css:112:  transition: box-shadow 0.2s, border-color 0.2s ease-in-out;
public/css/app.css:136:  transition: all 0.2s ease-out;
public/css/app.css:449:  transition: background-color 0.2s ease-out, color 0.2s ease-out;
public/css/app.css:519:  transition: all 0.2s ease-out;
public/css/app.css:575:  transition: all 0.2s ease-out;
public/css/app.css:580:  transition: all 0.2s ease-out;
public/css/app.css:654:  transition: all 0.25s ease-out;
public/css/app.css:666:  transition: all 0.2s ease-out;
public/css/app.css:685:  transition: all 0.2s ease;
public/css/app.css:937:  transition: all 0.2s ease-out;
```

public/js/monitor.js
```diff
@@ -114,25 +114,9 @@ function enableBtnIfEmailValid(e) {
...
-
-function showMessageIfDisabled() {
-  const emailBtns = document.querySelectorAll(".submit-email");
-  for (const emailBtn of emailBtns) {
-    if (emailBtn.disabled) {
-      addClass(".email-form", "invalid");
-    }
-  }
+function showMessageIfDisabled(e) {
+  const thisForm = e.target.parentElement.parentElement;
```
pdehaan: This feels risky. Like, it only works because the current target element is wrapped in a `<div>` or something, and if we modify nesting slightly, this will subtly break.
Not sure if there is an easier way to get a parent form wrapper or not.

lesleyjanenorton:
better finding of parental forms...

public/css/app.css
```diff
@@ -474,50 +477,40 @@ form span {
...
   width: 25px;
   height: auto;
   visibility: visible;
 }
 
-.input-group-button.loading-data .button {
-  color: rgba(255, 255, 255, 0);
+.form-group.loading-data .input-group-button,
+.form-group.loading-data .button {
+  color: rgba(255, 255, 255, 0) !important;
```
pdehaan: Actually, I'm a bit confused here, now that I read this again. Above on L480, we define the `.form-group.loading-data .button {}` with a color of `rgba(#000, 0) !important` but here we're overriding the same selector with a value of `rgba(#fff, 0) !important`. Since the alpha values are 0 in both cases, not sure it matters, but we could possibly drop `.form-group.loading-data .button` from one group or the other.

```diff
+.form-group.loading-data .button, ... {
+  color: rgba(0, 0, 0, 0) !important;
 }
```

lesleyjanenorton:
updated per review comments

public/js/monitor.js
```diff
@@ -114,24 +114,20 @@ function enableBtnIfEmailValid(e) {
...
-function enableBtnIfInputEmpty(e) {
-  const thisForm = e.target.form;
-  if(thisForm.querySelector(".submit-email")) {
-    if (!e.target.value) {
-      thisForm.classList.remove("invalid");
-
-    } else {
-      enableBtnIfEmailValid(e);
-    }
+function findParent(element, tag) {
```
groovecoder: 😻 this function

PR #244, opened 2018-08-01
Load breaches on startup 242

groovecoder:


hibp.js
```diff
@@ -70,13 +69,13 @@ const HIBP = {
       // ]
       for (const breachedAccount of response.body) {
         if (sha1.toUpperCase() === sha1Prefix + breachedAccount.hashSuffix) {
-          foundBreaches = await DBUtils.getBreachesByNames(breachedAccount.websites);
+          foundBreaches = allBreaches.filter(breach => breachedAccount.websites.includes(breach.Name));
```
nhnt11: I just realized this loop doesn't break for some reason. While we're here, can you refactor this to something like:
```
if (the hash doesn't match) {
  continue;
}
foundBreaches = [...]
```

groovecoder:
Still need to decide how to “refresh” the list of breaches in the app process memory. When it was in the database, we were going to run the `scripts/load-breaches.js` script on a cron job.

Now that it's in process memory, we will either need to:

1. Restart the server
2. Use a thread (!) to refresh it
3. Something else?

nhnt11:
Another thing that worries me is how this will scale: depending on how often or not we refresh breaches, if we're running multiple server processes they may go out of sync.

hibp.js
```diff
@@ -1,50 +1,81 @@
...
+    return await got(url, reqOptions);
+  },
+
+  async loadBreachesIntoApp(app) {
+    console.log("Loading breaches from HIBP into app.locals");
+    try {
+      const breachesResponse = await this.req("/breaches");
+      const breaches = [];
+
+      for (const breachIndex in breachesResponse.body) {
```
pdehaan: Should this be `for (const breach of breachesResponse.body) {...}`?
Then we shouldn't need the intermediate `breachIndex` and `breachesResponse.body[breachIndex]` stuffs.

hibp.js
groovecoder: Interesting. What's the advantage to inverting the conditional logic? Optimization?

hibp.js
groovecoder: Weird. I thought I started with that but the `got` return value wasn't iterable. I just tried it with `of` and it seems to work fine now. 🤷‍♂️

hibp.js
groovecoder: I thought about this some more, and I'm keeping the conditional logic like it is, so that I can `break` when I find the match.

That should achieve the same optimization of the loop.

If there's another strong reason to invert the conditional logic, let me know here and/or file an issue for follow|fix-up.

PR #248, opened 2018-08-02
fix #247: refactor db/utils to db/db and just knex

### LINKED ISSUE #247: Refactor db/utils into db/models/subscriber
Now that we’re down to a single DB table, I want to refactor the DB layer code. Most of `db/utils.js` is gone now, and the remaining functions are for the `email_hashes` table that we’re still using. 

Refactor `db/utils.js` and `db/models/emailhash.js` into a new `db/models/subscriber.js` model & table, clean up the knex migrations to match.

groovecoder:


db/db.js
```diff
@@ -44,40 +40,24 @@ const DBUtils = {
...
+        return res[0];
       }
 
       return aEntry;
     }, async () => {
-      return await EmailHash
-        .query()
-        .insert({ sha1, email, verified });
+      const res = await knex("subscribers")
+        .insert({ sha1: sha1, email: email, verified: verified })
```
pdehaan: Not sure if it is any easier just to use the shorthand syntax of `.insert({ sha1, email, verified })` since the keys and values are the "same".

There's probably an ESLint rule for that, if you want me to research it and address in a separate PR.

db/migrations/initial_schema.js
```diff
@@ -13,5 +13,5 @@ exports.up = knex => {
 
 exports.down = knex => {
   return knex.schema
-    .dropTableIfExists("email_hashes");
+    .dropTableIfExists("subscribers");
```
pdehaan: I honestly don't know anything about knex and schemas, but is this all cool to do in the "initial_schema", or should we be doing patch files?

Not sure how much this may mess up @Micheletto when he goes to deploy to stage/prod and table names may change and we may not be dropping ye olde "email_hashes" table, etc.

routes/hibp.js
```diff
@@ -6,17 +6,18 @@ const bodyParser = require("body-parser");
...
   try {
     const hashes = req.body.hashSuffixes.map(suffix=>req.body.hashPrefix + suffix);
     console.log("hashes: ", hashes);
 
-    const subscribers = await DBUtils.getSubscribersByHashes(hashes);
+    const subscribers = await DB.getSubscribersByHashes(hashes);
     console.log("subscribers: ", subscribers);
 
-    const breach = DBUtils.getBreachByName(req.body.breachName);
+    const reqBreachName = req.body.breachName;
```
pdehaan: Do we need a `.toLowerCase()` here as well? (I see a similar pattern below in routes/home.js:10-11)

scripts/notify-subscribers-of-breach.js
```diff
@@ -5,7 +5,7 @@ if (!breachName) {
   throw("Usage: node notify-subscribers-of-breach.js \"breach-name\"");
 }
 
-const DBUtils = require("../db/utils");
+const DBUtils = require("../db/db");
```
pdehaan: Should we bother renaming this to `DB` as well, to be consistent with the other files.
I can't imagine caring much either way, it's just that now I've seen it, so I feel legally obliged to say it.

db/migrations/initial_schema.js
groovecoder: In fact, I want to make this db-altering change now, because neither stage nor prod have a database yet.

I'll manually delete the schema and run knex initial migration on heroku to test it all out before we try anything on stage/prod.

db/db.js
groovecoder: Thought I had that before and it wasn't working. But it works now, so yup - added commit to use shorthand syntax.

routes/hibp.js
groovecoder: Good catch!

db/db.js
```diff
@@ -4,35 +4,31 @@
...
-  async addUnverifiedEmailHash(email) {
-    return await EmailHash.query().insert(
-      {email: email, sha1: getSha1(email), verification_token: uuidv4(), verified: false}
-    );
+const DB = {
+  async addSubscriberUnverifiedEmailHash(email) {
+    const res = await knex("subscribers").insert(
+      { email: email, sha1: getSha1(email), verification_token: uuidv4(), verified: false }
+    ).returning("verification_token");
+    return res[0];
```
groovecoder: Note: `knex` queries return arrays, so have to return the `0`th item.

package.json
```diff
@@ -24,7 +24,6 @@
     "knex": "^0.14.6",
     "nodemailer": "^4.6.4",
     "nodemailer-express-handlebars": "^3.0.0",
-    "objection": "^1.1.9",
```
groovecoder: image

lesleyjanenorton: Leave now and never come back!

groovecoder:
Good point. No change needed to the `knex` migration script line on the `README` ... but there's some other stuff in that `README` that can be updated now. Will file a follow-up issue.

db/db.js
nhnt11: Suggestion: while we're refactoring, let's proactively add a stub function for encryption (that just returns the email as-is for now) and use it everywhere. Later when we add encryption, we can just add it in one place and not have to refactor everything again.

scripts/notify-subscribers-of-breach.js
nhnt11: +1, let's call this DB now.

db/db.js
groovecoder: Can we do in a follow-up? I'm picturing this as part of moving `sha1-utils.js` into a new `crypto.js` module which will pull the private key in only when needed.

groovecoder:
Renamed to `DB`. I'd like to file a follow-up issue for the encrypted email stub function.

pdehaan:
@nhnt11 Mind re-reviewing this and giving it an 👍👎.

PR #250, opened 2018-08-03
fix #249: update README to match smaller DB

### LINKED ISSUE #249: Update README
Now that https://github.com/mozilla/blurts-server/pull/244 is merged, we can update the README to match.

groovecoder:
and drop other things we don't need like AWS creds

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,20 +1,15 @@
-# Breach Alerts
+# Firefox Monitor Server
 
 ## Summary
 
-Firefox Breach Alerts notifies users when their credentials have possibly been leaked or stolen in a data breach. Powered by [haveibeenpwned.com](https://haveibeenpwned.com/).
+Firefox Monitor notifies users when their credentials have been compromised in a data breach.
 
-Communicates with the [blurts-addon](https://github.com/mozilla/blurts-addon) client-side add-on for Firefox Monitor.
+This code is for the monitor.firefox.com service & website.
 
-## Context
+Breach data is powered by [haveibeenpwned.com](https://haveibeenpwned.com/).
 
 See the [Have I Been Pwned about page](https://haveibeenpwned.com/About) for
 the "what" and "why" of data breach alerts.
-
-This project explores how Firefox - as the user agent - can support this. 
-
-See [the Breach Alert Product Brief](https://docs.google.com/document/d/1GTS0HIihfTErA7P19HPYfvHCA3v9g67B_Cf2bpmE0Bw/edit)
-for more background, objectives, key use cases.
 
 ## Development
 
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -1,21 +1,16 @@
-# Breach Alerts
+# blurts-server
```
pdehaan: Odd change... Bonus points if we can just change this to "Firefox Monitor Server" or something readable, since "blurt" is confusing.

README.md
```diff
@@ -1,21 +1,16 @@
...
 
 ## Summary
 
-Firefox Breach Alerts notifies users when their credentials have possibly been leaked or stolen in a data breach. Powered by [haveibeenpwned.com](https://haveibeenpwned.com/).
+Firefox Monitor notifies users when their credentials have been compromised in a data breach.
 
-Communicates with the [blurts-addon](https://github.com/mozilla/blurts-addon) client-side add-on for Firefox Monitor.
+This code is for the monitor.firefox.com service & website.
```
pdehaan: Why remove the link to the add-on?

README.md
```diff
@@ -1,21 +1,16 @@
...
 
-## Context
+Brech data is powered by [haveibeenpwned.com](https://haveibeenpwned.com/).
```
pdehaan: typo: "Breach"

README.md
```diff
@@ -54,19 +49,28 @@ for more background, objectives, key use cases.
...
+   ```
 
-* `node scripts/load-breaches.js` loads breaches from HIBP into the database
-* `node scripts/add-breached-emails.js` adds some breached email records for
-  `test[1-3]@test.com` users
+3. Run the migrations:
 
-* `node scripts/add-breached-emails.js --help` for usage help.
+   ```
+   ./node_modules/.bin/knex --knexfile db/knexfile.js migrate:latest
```
pdehaan: nit: Not sure if it's worth adding these to an npm script to avoid having to use the `./node_modules/bin/` prefix, and we could just run <kbd>$ npm run db:migrate</kbd>.
Where package.json has the following script:

```json
"db:migrate": "knex --knexfile db/knexfile.js migrate:latest"
```

README.md
```diff
@@ -102,23 +109,6 @@ OAUTH_TOKEN_URI="https://oauth-stable.dev.lcip.org/v1/token"
 
 #### Breach Hashsets
```
pdehaan: Kill this orphaned header?

README.md
groovecoder: It doesn't actually communicate with the add-on at all. The add-on just sends users to the site.

pdehaan: OK, makes sense. I think I just liked the add-on repo link because it was the only place that I recall that mentioned the add-on portion of the project.
But maybe that will be moot if the add-on code gets moved into tree or not.

PR #252, opened 2018-08-06
WIP: start tests

groovecoder:


groovecoder:
Updated with db fixture plumbing and first test for a db function.

package.json
```diff
@@ -31,14 +31,14 @@
     "coveralls": "^3.0.1",
     "eslint": "^4.18.1",
     "eslint-plugin-node": "^6.0.1",
-    "faucet": "^0.0.1",
+    "faucet": "0.0.1",
```
pdehaan: Did we mean to pin this to a specific version?

db/seeds/test_subscribers.js
```diff
@@ -0,0 +1,29 @@
+"use strict";
+
+exports.seed = function(knex, Promise) {
```
pdehaan: Not sure if `Promise` is used or if we can remove it. It may be confusing naming given the native `Promise` object/class.

db/seeds/test_subscribers.js
```diff
@@ -0,0 +1,29 @@
...
+  // Deletes ALL existing entries
+  return knex("subscribers").del()
+    .then(function () {
+      // Inserts seed entries
+      return knex("subscribers").insert([
+        {
+          sha1: "e76f6c9da0e88b888a98f896b25ee4dcbf769579",
```
pdehaan: Do we want to hardcode the SHA1 of the emails, or just use the `getSha1()` method to keep everything in sync??

db/knexfile.js
```diff
@@ -2,11 +2,24 @@
...
+      max: 10,
+    },
+  },
+  tests: {
+    client: "postgresql",
+    connection: TEST_DATABASE_URL,
+    pool: {
+      min: 2,
+      max: 10,
+    },
```
pdehaan: Do we need a `production` key here as well? Or does OPs need to do anything special to deploy now that we're using an additional level of config nesting here?

tests/test-db.js
```diff
@@ -0,0 +1,20 @@
...
+const knex = Knex(knexConfig["tests"]);
+*/
+
+
+test("getSubscribersByHashes returns array of subscriber objects", async t => {
+  // test@example.com
+  const testHashes = ["b84c4c03b2af4050ac2d3b105e58adf83fa5af05"];
+  const subscribers = await DB.getSubscribersByHashes(testHashes);
+  console.log("subscribers: ", subscribers);
+  t.end();
```
pdehaan: I assume this is still "WIP" since we aren't asserting anything in this test.

tests/test-sha1.js
```diff
@@ -0,0 +1,19 @@
...
+    if (parseInt(character, 16).toString(16) != character.toLowerCase()) {
+      return false;
+    }
+  }
+  return true;
+}
+
+test("getSha1 returns hex digest", function (t) {
+  t.plan(1);
+  t.ok(isHexString(getSha1("test@test.com")));
```
pdehaan: Do we want to assert the returned value from `getSha1()` here matches an expected value for "test@test.com" (versus just checking each character is a valid hex string? Or are we doing that elsewhere?

tests/test-sha1.js
```diff
@@ -0,0 +1,19 @@
...
+function isHexString(hashDigest) {
+  for (const character of hashDigest) {
+    if (parseInt(character, 16).toString(16) != character.toLowerCase()) {
+      return false;
+    }
+  }
+  return true;
+}
+
+test("getSha1 returns hex digest", function (t) {
```
pdehaan: I notice on L14 of the previous file we were using arrow functions... Not sure if we have a preference or if anything matters. Just thought I'd point it out. I can't imagine there are pros/cons to either approach, apart from consistency.

tests/test-db.js
```diff
@@ -0,0 +1,20 @@
...
+/*
+const knexConfig = require("../db/knexfile");
+
+const knex = Knex(knexConfig["tests"]);
+*/
+
+
+test("getSubscribersByHashes returns array of subscriber objects", async t => {
+  // test@example.com
+  const testHashes = ["b84c4c03b2af4050ac2d3b105e58adf83fa5af05"];
```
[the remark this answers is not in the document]
groovecoder: Yup, I have an update with all the right values.

package.json
groovecoder: Not sure why npm changed that.

db/seeds/test_subscribers.js
groovecoder: Good catch. This was auto-generated by knex. Removing the `Promise` param.

db/seeds/test_subscribers.js
groovecoder: Updated.

db/knexfile.js
groovecoder: Yes, I'll plan to add a production section too.

groovecoder:
Argh! I've tried both `tape-async` and `tape-promise` but I can't get an `async` `tape` test to return. @pdehaan do you know any other projects that use `tape`? Or if they use a different TAP-compatible test library that supports async?

tests/test-db.js
```diff
@@ -0,0 +1,25 @@
...
+test("getSubscribersByHashes accepts hashes and only returns verified subscribers", async t => {
+  const testHashes = [
+    "firefoxaccount@test.com",
+    "unverifiedemail@test.com",
+    "verifiedemail@test.com",
+  ].map(email => getSha1(email));
+  const subscribers = await DB.getSubscribersByHashes(testHashes);
+  for (const subscriber of subscribers) {
+    t.ok(subscriber.verified);
+  }
```
pdehaan: Does this need a `t.end()` to match the test file below?
Similarly, do we need any `t.plan(3)` bits, like we had in some previous iterations?

groovecoder: I've tried every combination, and also wrapping the `await` line in `try...catch` too. :(

pdehaan: Turns out the solution seems to be calling `knex.destroy()` during test suite teardown to kill the DB pool, or possibly disable connection pooling during <kbd>$ npm run test</kbd>.

Ref: https://knexjs.org/#Installation-pooling

groovecoder:
Thanks to @pdehaan and @tatumcreative the tests now run AND complete.

I covered all the "public" `DB` methods with at least 1 test. So this PR now increases our test coverage from 0% to 20%!

groovecoder:
Still TODO and/or follow-up issues?

* [ ] Stub/proxy out the HIBP call in `_verifySubscriber`
* [ ] Cover moar code in moar tests!!

db/knexfile.js
```diff
@@ -2,11 +2,21 @@
...
 
-module.exports = {
+
+// For runtime, use DATABASE_URL
+const RUNTIME_CONFIG = {
   client: "postgresql",
   connection: AppConstants.DATABASE_URL,
-  pool: {
-    min: 2,
-    max: 10,
```
pdehaan: Did you want to disable connection pooling for _all_ environments, or just for the "tests" env?

db/DB.js
```diff
@@ -8,22 +10,29 @@ const knexConfig = require("./knexfile");
 const HIBP = require("../hibp");
 const getSha1 = require("../sha1-utils");
 
-const knex = Knex(knexConfig);
+const knex = Knex(knexConfig[AppConstants.NODE_ENV]);
```
pdehaan: Do we need to add more safety around this? Like, if `process.env.NODE_ENV` is set to "production" (instead of "prod"), the `knexConfig[AppConstants.NODE_ENV]` would presumably be `undefined`, but I don't know what that would do if we pass `undefined` to the `Knex()` constructor.
Wasn't sure if storing `knexConfig[AppConstants.NODE_ENV]` in a temporary variable and checking for truthiness (else failing hard) would be worth the extra hassle or not.

db/DB.js
```diff
@@ -8,22 +10,29 @@ const knexConfig = require("./knexfile");
...
 
 
 const DB = {
+  async getSubscriberByEmailAndToken(email, token) {
+    const res = await knex("subscribers")
+      .where("verification_token", "=", token)
+      .andWhere("email", "=", email);
+
+    return res[0];
```
pdehaan: Sorry, was `res[0]` a thing? Or did you have a TODO there before to use `.first()` API?

https://knexjs.org/#Builder-first

package.json
```diff
@@ -38,7 +39,9 @@
     "nyc": "^11.8.0",
     "stylelint": "^9.2.0",
     "stylelint-config-standard": "^18.2.0",
-    "tape": "^4.9.0"
+    "tape": "^4.9.1",
+    "tape-async": "^2.3.0",
+    "tape-promise": "^3.0.0"
```
pdehaan: Do we need **BOTH** `tape-async` and `tape-promise`? I think the latter is currently unused.

tests/test-db.js
```diff
@@ -0,0 +1,73 @@
...
+const DB = require("../db/DB");
+const getSha1 = require("../sha1-utils");
+
+test("getSubscriberByEmailAndToken accepts email and token and returns subscriber", async t => {
+  const testEmail = "unverifiedemail@test.com";
+  const testToken = "0e2cb147-2041-4e5b-8ca9-494e773b2cf0";
+  const subscriber = await DB.getSubscriberByEmailAndToken(testEmail, testToken);
+
+  t.ok(subscriber.email === testEmail);
+  t.ok(subscriber.verification_token === testToken);
```
pdehaan: I'm not a **tape** expert, but do we want `t.ok()` or something like [`t.deepEqual()`](https://github.com/substack/tape#tdeepequalactual-expected-msg) here?

tests/test-db.js
```diff
@@ -0,0 +1,73 @@
...
+  t.ok(verifiedSubscriber.verified);
+  t.ok(verifiedSubscriber.sha1 === getSha1(testEmail));
+});
+
+test("removeSubscriber accepts email and removes the email address", async t => {
+  const testEmail = "removingFirefoxAccount@test.com";
+
+  const verifiedSubscriber = await DB.addSubscriber(testEmail);
+  const removedSubscriber = await DB.removeSubscriber(verifiedSubscriber.email);
+  console.log("removedSubscriber: ", removedSubscriber);
```
pdehaan: Nit: remove `console.log()`

tests/test-sha1.js
```diff
@@ -0,0 +1,25 @@
...
+test("getSha1 returns hex digest", t => {
+  t.ok(isHexString(getSha1("test@test.com")));
+  t.end();
+});
+
+test("tape-async example", async t => {
+  t.ok(isHexString(getSha1("test@test.com")));
+  const a = await Promise.resolve(42);
+  t.equal(a, 42);
+});
```
pdehaan: Nit: nuke this test

tests/test-db.js
```diff
@@ -0,0 +1,73 @@
...
+
+  t.ok(removedSubscriber.email === null);
+});
+
+test("teardown", async t => {
+  DB.destroyConnection();
+});
```
pdehaan: :+1:

db/knexfile.js
groovecoder: These are the `knex` defaults anyway, so I thought I'd just remove them.

db/DB.js
groovecoder: `first` only works on `select` queries/functions. Doesn't work on these `update` and `insert`

package.json
groovecoder: Good catch.

tests/test-db.js
```diff
@@ -0,0 +1,73 @@
...
+  const subscribers = await DB.getSubscribersByHashes(testHashes);
+  for (const subscriber of subscribers) {
+    t.ok(subscriber.verified);
+  }
+});
+
+test("addSubscriberUnverifiedEmailHash accepts email and returns unverified subscriber with sha1 hash and verification token", async t => {
+  const testEmail = "test@test.com";
+  // https://stackoverflow.com/a/13653180
+  const uuidRE = /^[0-9a-f]{8}-[0-9a-f]{4}-[1-5][0-9a-f]{3}-[89ab][0-9a-f]{3}-[0-9a-f]{12}$/i;
```
pdehaan: I think this is cool, although _mildly_ confused by that `-[1-5][0-9af]{3}-` bit in the middle there.

If I'm reading these links below correctly, it seems like that `[1-5]` may imply the version, and we could possibly just hardcode it as "4" since we're using UUIDv4, per https://github.com/mozilla/blurts-server/blob/069ea0bb3fa95347c46848d9be760716438fa82e/db/DB.js#L4

1. https://github.com/chriso/validator.js/blob/master/lib/isUUID.js#L16
1. https://gitlab.com/unwttng/uuid-v4-regex/blob/master/regex.js
1. https://github.com/mixer/uuid-validate/blob/master/index.js

tests/test-db.js
groovecoder: I tend to avoid assertions on full object structures from experiences where it makes tests brittle. 😐 I'm fine using strict equality comparisons For Now™.

package.json
```diff
@@ -28,6 +28,7 @@
     "uuid": "^3.2.1"
   },
   "devDependencies": {
+    "blue-tape": "^1.0.0",
```
pdehaan: Is this used, or just leftover from testing?

db/DB.js
groovecoder: It results in a knex connection error which kills the server or tests process pretty early. Which seems like as good a time as any for an error to throw.

tests/test-db.js
pdehaan: Sorry, I didn't mean **full structure** comparisons, I just meant something like:

```js
t.deepEqual(subscriber.email, testEmail);
t.deepEqual(subscriber.verification_token, testToken);
```

Not really a big deal either way, I just usually find the debugging a bit better w/ `.deepEqual()` vs `.ok()`.

Using `.ok()` with `===` just tells me an expected/actual of true/false, which can sometimes be trickier to debug:

```sh
// t.ok(subscribers.length === 3);

# getSubscribersByHashes accepts hashes and only returns verified subscribers
not ok 1 should be truthy
  ---
    operator: ok
    expected: true
    actual:   false
    at: testp (/Users/pdehaan/dev/github/mozilla/blurts-server/tests/test-async.js:34:7)
    stack: |-
      Error: should be truthy
          at Test.assert [as _assert] (/Users/pdehaan/dev/github/mozilla/blurts-server/node_modules/tape/lib/test.js:224:54)
          at Test.bound [as _assert] (/Users/pdehaan/dev/github/mozilla/blurts-server/node_modules/tape/lib/test.js:76:32)
          at Test.assert (/Users/pdehaan/dev/github/mozilla/blurts-server/node_modules/tape/lib/test.js:342:10)
          at Test.bound [as ok] (/Users/pdehaan/dev/github/mozilla/blurts-server/node_modules/tape/lib/test.js:76:32)
          at testp (/Users/pdehaan/dev/github/mozilla/blurts-server/tests/test-async.js:34:7)
          at <anonymous>
```

... versus `.deepEqual()` which gives a more obvious expected/actual, which makes it easier to see exactly why it failed instead of just a Boolean value:

```sh
// t.deepEqual(subscribers.length, 3);

# getSubscribersByHashes accepts hashes and only returns verified subscribers
not ok 1 should be equal
  ---
    operator: equal
    expected: 3
    actual:   2
    at: testp (/Users/pdehaan/dev/github/mozilla/blurts-server/tests/test-async.js:34:7)
    stack: |-
      Error: should be equal
          at Test.assert [as _assert] (/Users/pdehaan/dev/github/mozilla/blurts-server/node_modules/tape/lib/test.js:224:54)
          at Test.bound [as _assert] (/Users/pdehaan/dev/github/mozilla/blurts-server/node_modules/tape/lib/test.js:76:32)
          at Test.equal (/Users/pdehaan/dev/github/mozilla/blurts-server/node_modules/tape/lib/test.js:384:10)
          at Test.bound [as equal] (/Users/pdehaan/dev/github/mozilla/blurts-server/node_modules/tape/lib/test.js:76:32)
          at testp (/Users/pdehaan/dev/github/mozilla/blurts-server/tests/test-async.js:34:7)
          at <anonymous>
```

But overall, I love everything. Let's just land these tests and iterate from there (assuming @nhnt11 and @lesleyjanenorton are kosher with this PR).

tests/test-db.js
```diff
@@ -0,0 +1,72 @@
...
+  const subscribers = await DB.getSubscribersByHashes(testHashes);
+  for (const subscriber of subscribers) {
+    t.ok(subscriber.verified);
+  }
+});
+
+test("addSubscriberUnverifiedEmailHash accepts email and returns unverified subscriber with sha1 hash and verification token", async t => {
+  const testEmail = "test@test.com";
+  // https://stackoverflow.com/a/13653180
+  const uuidRE = /^[0-9a-f]{8}-[0-9a-f]{4}-4[0-9a-f]{3}-[89ab][0-9a-f]{3}-[0-9a-f]{12}$/i;
```
pdehaan: Do we do UUID `verification_token` validation anywhere else in the server (or _should_ we be validating it elsewhere)?
Should this RegExp (or a helper function) be added to the ../db/DB.js file as some `isValidVerificationToken(token) {...}` method?

tests/test-db.js
groovecoder: Ah ... yeah that's a nicer output. Good call.

tests/test-db.js
groovecoder: Hmm ... it would be belt-and-suspenders validation, since we're checking the token equality in the database already. We can make it a follow-up.

Issue #251, opened 2018-08-06
Use SES for emails


groovecoder:
On local machines and the heroku dev instance, we are using mailgun for emails.

For staging & prod we will need to use SES. To use SES, we will need to:

1. [x] Change the SMTP config to SES (See [bug 1480435](https://bugzilla.mozilla.org/show_bug.cgi?id=1480435))
2. [x] Specify a configuration set with an SMTP header: `X-SES-CONFIGURATION-SET: <ConfigSet>` (See [docs](https://docs.aws.amazon.com/ses/latest/DeveloperGuide/using-configuration-sets-in-email.html))
3. [x] Add a `/ses/notification` endpoint to handle email bounce notifications and delete those users from our subscribers
4. [ ] Add authentication (HTTP basic auth or digital signature verification) to `/ses/notification`

groovecoder:
@Micheletto - please let me know what SES SMTP values we can try ASAP.
```
SMTP_HOST=""
SMTP_PORT=
SMTP_USERNAME=""
SMTP_PASSWORD=""
```

pdehaan:
> Add a `/ses-bounce` endpoint to handle email bounce notifications and delete those users from our subscribers

How many bounced emails would we allow before removing a user from the subscribers table? 1? 4?

groovecoder:
I'm leaving this issue open because I still need to add authentication to the endpoint. Either HTTP basic auth (has been problematic) or digital signature verification. I'll update the description.

PR #256, opened 2018-08-10
for #251: SES needs a specific From: address

groovecoder:
for #251: fix verification token for email template

 for #251: add /ses/notification endpoint


routes/user.js
```diff
@@ -22,9 +22,10 @@ const urlEncodedParser = bodyParser.urlencoded({ extended: false });
 
 router.post("/add", urlEncodedParser, async (req, res) => {
   const email = req.body.email;
-  const verificationToken = await DB.addSubscriberUnverifiedEmailHash(email);
+  const unverifiedSubscriber = await DB.addSubscriberUnverifiedEmailHash(email);
+  const verificationToken = unverifiedSubscriber.verification_token;
 
-  const url = `${AppConstants.SERVER_URL}/user/verify?token=${encodeURIComponent(verificationToken)}&email=${encodeURIComponent(email)}`;
+  const url = `${AppConstants.SERVER_URL}/user/verify?token=${encodeURIComponent(verificationToken)}`;
```
groovecoder: I decided to drop the extra use of the plaintext email in the verification link to prevent leakage via any `Referer` values.

email-utils.js
```diff
@@ -76,8 +76,9 @@ const EmailUtils = {
 
     return new Promise((resolve, reject) => {
       gTransporter.use("compile", hbs(hbsOptions));
+      const emailFrom = AppConstants.EMAIL_FROM || `"Firefox Monitor" < ${kSMTPUsername}`;
```
pdehaan: This feels suspicious to me now.
Before we had `<${kSMTPUsername}>`, but now we have an extra space after the `<` and it looks like the `>` is missing.

routes/ses.js
```diff
@@ -0,0 +1,24 @@
+"use strict";
+
+const express = require("express");
+const bodyParser = require("body-parser");
+
+const router = express.Router();
+const jsonParser = bodyParser.json();
+
+
+router.post("/notification", jsonParser, async (req, res) => {
```
pdehaan: Does this need to be an `async` handler?

routes/ses.js
```diff
@@ -0,0 +1,24 @@
...
+  try {
+    console.log("ses-notification request body: ", req.body);
+
+    res.status(200).json(
+      {status: "OK"}
+    );
+  } catch (e) {
+    res.status(500).json(
+      {info: "Internal error."}
+    );
```
pdehaan: Do we have some server logging enabled somewhere that will tell us if we ever have cases of 500s? Curious how the `.catch()` would ever be reached even, since the method doesn't really _do_ anything or process the `req.body` or whatever.

db/DB.js
```diff
@@ -30,7 +29,7 @@ const DB = {
   },
 
   async verifyEmailHash(token, email) {
```
pdehaan: Drop the `email` attribute now?

email-utils.js
groovecoder: Good catch.

routes/ses.js
groovecoder: It will be. It will make `await` calls to async `DB` functions for updating the `subscribers` table on bounce & complaint notifications.

routes/ses.js
groovecoder: We have heroku & papertrail logging on heroku. I'm sure @Micheletto and Ops team have their own response logging too.

This is still WIP so no functional code in here yet. Just making sure we're getting notifications from SES.

PR #261, opened 2018-08-15
handle ALL permanent bounces from SES

groovecoder:


routes/ses.js
```diff
@@ -1,14 +1,32 @@
...
+const express = require("express");
 
+const AppConstants = require("../app-constants");
 const {notification} = require("../controllers/ses");
 
 
 const router = express.Router();
 const textParser = bodyParser.text();
 
+function check (user, pass) {
```
pdehaan: Nit: I'd probably suggest a more descriptive name. `checkAuth()` or something maybe.

tests/test-ses.js
```diff
@@ -23,24 +29,27 @@ test("setup", async t => {
...
-  const resp = httpMocks.createResponse();
-
-  await ses.notification(req, resp);
-  t.deepEqual(resp.statusCode, 200);
-
-  subscribers = await DB.getSubscribersByHashes([getSha1(testEmail)]);
-  t.deepEqual(subscribers.length, 0);
+  const bounceSubTypes = ["General", "NoEmail", "Suppressed"];
+  for (let i=0; i < bounceSubTypes.length; i++) {
+    const testEmail = "bounce@simulator.amazonses.com";
```
pdehaan: super nit: Not sure if you want to move any of these static values outside of the loop.
Like this line, and possibly the 2 calls to `[getSha1(testEmail)]` below, which would be getting called 6x with the same value. Although I don't know how "expensive" a call to `getSha1()` is, so maybe not a huge deal to shave only a few ms off of the test run, but moving it out of the loop into `const testHashes = [getSha1(testEmail)];` would reduce it from 6 calls to 1.

```js
const bounceSubTypes = ["General", "NoEmail", "Suppressed"];
const testEmail = "bounce@simulator.amazonses.com";
const testHashes = [getSha1(testEmail)];

for (let i = 0; i < bounceSubTypes.length; i++) {
  await DB.addSubscriber(testEmail);
  let subscribers = await DB.getSubscribersByHashes(testHashes);
  t.deepEqual(subscribers.length, 1);
  // ...
  subscribers = await DB.getSubscribersByHashes(testHashes);
  t.deepEqual(subscribers.length, 0);
}
```

groovecoder:
Re-based and addressed review comments. Will merge and push to heroku. Then I'm going to start a new branch on this and repeat the routes+controllers+tests architecture for the remaining routes.

PR #262, opened 2018-08-17
fix #254: mock hibp for tests

### LINKED ISSUE #254: tests: stub out calls to HIBP API
In `test-db.js`, the `verifyEmailHash` and `addSubscriber` tests make live calls to the HIBP API. We should stub out those calls to improve the performance and isolation of the DB tests.

groovecoder:
This also switches to `jest` for testing, which takes the place of: `tape`, `tape-async`, `nyc`


tests/.eslintrc
```diff
@@ -0,0 +1,2 @@
+{ "rules": { "no-undef": "off" } }
```
groovecoder: `jest` injects a bunch of globals, which I don't love. Is there an `eslintrc` rule directive to allow just a certain set of globals?

groovecoder: Ooo, I found `eslint` has a `jest` environment already. Updated.

tests/.eslintrc
```diff
@@ -0,0 +1,2 @@
+{ "env": { "jest": true } }
```
pdehaan: Rather than a nested .eslint file, I'd probably suggest adding this to the `overrides` section in our root .eslintrc.js file:

https://github.com/mozilla/blurts-server/blob/03fdd03dabf7cf37a6f60c1e328f5efb761c0cea/.eslintrc.js#L10-L32

Probably something adjacent to:
```js
{
  "files": [
    "tests/**/*.js",
  ],
  "env": {
    "jest": true
  }
}
```

groovecoder: Moved into our root `.eslintrc.js`

Issue #255, opened 2018-08-09
add UUIDv4 validation to /user/verify route


groovecoder:
For belt+suspenders validation, we can validate that the verification token passed to /user/verify is UUIDv4 before checking the database.

pdehaan:
Per https://fx-breach-alerts.herokuapp.com/user/verify?token=576a72dc-4727-4f97-8df1-11fef

Passing an invalid UUID v4 token seems to trigger an "Application error" on the [/user/verify?token=](https://fx-breach-alerts.herokuapp.com/user/verify?token=576a72dc-4727-4f97-8df1-11fef) route (at least on Heroku).

![invalid-token](https://user-images.githubusercontent.com/557895/43984518-e69872ea-9cb5-11e8-84fa-9a9aa23c9dca.png)

pdehaan:
Other imaginary examples... I can do a POST to /user/add without a `req.body.email`, and then I just get weird, unhelpful error messages in the front end. It's a fairly bogus situation (since I can only POST using curl or httpie), but still...

```bash
$ curl -XPOST https://fx-breach-alerts.herokuapp.com/user/add

    ...
    <h2>Error</h2>
    <p> Cannot read property &#x27;normalize&#x27; of undefined </p>
```

Not 100% certain what/where that error is coming from yet. I can't find any references to `"normalize"` in our repo, outside of package names in package-lock.json.

pdehaan:
https://fx-breach-alerts.herokuapp.com/user/verify

> #### Error
> Undefined binding(s) detected when compiling SELECT query: select * from "subscribers" where "verification_token" = ?

groovecoder:
Good catch. Adding this to launch milestone.

pdehaan:
I didn't go through all the GET/POST endpoints, but it may be worth it to make sure we aren't leaking any server info and have reasonable user-facing error messages.

After a bit of digging, I _think_ this is a maybe/hopefully accurate route map:

| METHOD | ROUTE | QUERY/BODY
|:-------|:------|:-----------
| GET  | "/" | `req.query.breach`❓ 
| GET  | "/\_\_version\_\_" | n/a
| GET  | "/\_\_heartbeat\_\_" | n/a
| GET  | "/\_\_lbheartbeat\_\_" | n/a
| POST | "/hibp/notify" | `req.body.breachName` `req.body.hashPrefix`
| GET  | "/hibp/breaches" | `req.headers["if-modified-since"]`
[...]

PR #263, opened 2018-08-18
Moar tests

groovecoder:
* Adds a `resetDB` module for every test suite that touches the database, and uses `--runInBand` so tests aren't simultaneously changing each others' fixture data.
* Refactors `routes/user` to `controllers/user` so it can be tested.


tests/resetDB.js
```diff
@@ -0,0 +1,30 @@
...
+  DB.createConnection();
+});
+
+
+beforeEach(async () => {
+  const knex = Knex(knexConfig[AppConstants.NODE_ENV]);
+  await knex("subscribers").truncate();
+  await seed(knex);
+  knex.destroy();
+});
```
pdehaan: Should we have some `DB.reset()` method, or would that be too risky?

tests/resetDB.js
```diff
@@ -0,0 +1,30 @@
...
+const {seed} = require("../db/seeds/test_subscribers");
+
+
+// (Re-)create DB connection at the beginning of each test suite
+beforeAll(async () => {
+  DB.createConnection();
+});
+
+
+beforeEach(async () => {
```
pdehaan: Nit: add a code comment above saving we’re nuking your `subscribers` table before each test?

tests/ses.test.js
```diff
@@ -6,6 +6,9 @@ const DB = require("../db/DB");
 const getSha1 = require("../sha1-utils");
 const ses = require("../controllers/ses");
 
+require("./resetDB");
```
pdehaan: What happens if I add an `afterAll()` in this file now? Would it conflict with the one injected via resetDB.js?

tests/resetDB.js
```diff
@@ -0,0 +1,30 @@
...
+  const knex = Knex(knexConfig[AppConstants.NODE_ENV]);
```
pdehaan: Unrelated, but the more I see this line, the more I notice we always call it the same. What if ‘knexConfig’ just inspected the current `NODE_ENV` and only returned that config? Then we could just simplify to `Knex(knexConfig)` and not have to pass NODE_ENV each time.

tests/resetDB.js
```diff
@@ -0,0 +1,30 @@
...
+const Knex = require("knex");
+
+const AppConstants = require("../app-constants");
+const DB = require("../db/DB");
+const knexConfig = require("../db/knexfile");
+const {seed} = require("../db/seeds/test_subscribers");
+
+
+// (Re-)create DB connection at the beginning of each test suite
+beforeAll(async () => {
```
pdehaan: Not sure this needs to be async

tests/resetDB.js
groovecoder: Yeah, I thought it would be too risky.

tests/ses.test.js
groovecoder: Hmm ... yes it would. I'll have to think about that.

pdehaan: We could make ./resetDB.js super boring, and just have it export the `beforeAll()`, `beforeEach()`, and `afterAll()` functions.

Then in each test file, we do something similar to `beforeAll(resetDB.beforeAll)`.
If you want a custom `beforeEach()`, you just call the `resetDB.beforeEach();` and then call your custom code:

```js
beforeEach(async () => {
  await resetDB.beforeEach();
  customStuffIsFun();
});
```

A bit more verbose and boilerplate, but possibly more footgun protection and less side-effect-y.

groovecoder: I'd like to think about the boilerplate-y code vs. foot-gun-y/side-effect-y code more before I change this. It may be like the `knexConfig` - the more we see it used/un-used the clearer it will be which approach is best.

controllers/home.js
```diff
@@ -0,0 +1,25 @@
...
+
+function home(req, res) {
+  let breach = null;
+  if (req.query.breach) {
+    const reqBreachName = req.query.breach.toLowerCase();
+    breach = req.app.locals.breaches.filter(breach => breach.Name.toLowerCase() === reqBreachName)[0];
+  }
+  res.render("monitor", {
+    title: "Firefox Monitor",
+    breach: breach,
```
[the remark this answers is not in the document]
groovecoder: As a style preference, I only use object short-hand properties if the whole object uses it. I'm not sure why though?

controllers/home.js
```diff
@@ -0,0 +1,25 @@
...
+  });
+}
+
+
+function notFound(req, res) {
+  res.status(404);
+  res.render("404");
```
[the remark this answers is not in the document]
groovecoder: I had that, but it was harder to test.

controllers/user.js
```diff
@@ -0,0 +1,49 @@
...
+  );
+
+  res.send({
+    title: "Firefox Monitor : Confirm Email",
+  });
+}
+
+
+async function verify(req, res) {
+  const verifiedEmailHash = await DB.verifyEmailHash(req.query.token);
```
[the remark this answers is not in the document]
groovecoder: File a follow-up?

package.json
```diff
@@ -76,7 +76,7 @@
     "start": "node server.js",
     "test:db:migrate": "knex migrate:latest --knexfile db/knexfile.js --env tests",
     "test:db:seed": "knex seed:run --knexfile db/knexfile.js --env tests",
```
groovecoder: May not need this script now that `resetDB` is doing the seed.

controllers/user.js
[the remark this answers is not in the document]
pdehaan: Actually, I think that's already covered by ye olde belt-n-suspenders bug at #255.

controllers/hibp.js
```diff
@@ -0,0 +1,25 @@
...
+const DB = require("../db/DB");
+
+
+async function notify (req, res) {
+  const hashes = req.body.hashSuffixes.map(suffix=>req.body.hashPrefix + suffix);
+
+  const subscribers = await DB.getSubscribersByHashes(hashes);
+
+  const reqBreachName = req.body.breachName.toLowerCase();
+  const breach = req.app.locals.breaches.filter(breach => breach.Name.toLowerCase() === reqBreachName)[0];
```
pdehaan: Same as comment below, we _may_ want to use `.find()` here: https://github.com/mozilla/blurts-server/pull/263/files#r211107305

Also, because we're duplicating this code in 2 places in at least 2 files, we may want to have some shared `getFilterByName(name) {...}` convenience method.

controllers/oauth.js
```diff
@@ -0,0 +1,65 @@
...
+// This object exists instead of inlining the env vars to make it easy
+// to abstract fetching API endpoints from the OAuth server (instead
+// of specifying them in the environment) in the future.
+const FxAOAuthUtils = {
+  get authorizationUri() { return AppConstants.OAUTH_AUTHORIZATION_URI; },
+  get tokenUri() { return AppConstants.OAUTH_TOKEN_URI; },
+  get profileUri() { return AppConstants.OAUTH_PROFILE_URI; },
+};
+
+const FxAOAuth = new ClientOAuth2({
```
pdehaan: Small nit, but should this be called `FxAOAuthClient` ("Utils" vs "Client"), plus then below it may be a bit clearer as `function init(req, res, client = FxAOAuthClient) {...}`.

controllers/scan.js
```diff
@@ -0,0 +1,57 @@
...
+
+  if (req.body.featuredBreach) {
+    let featuredBreach = req.body.featuredBreach;
+    if(foundBreaches.filter(breach => breach.Name === featuredBreach).length > 0) {
+      featuredBreach = foundBreaches.filter(breach => breach.Name === featuredBreach);
+      if (foundBreaches.length > 1) {
+        foundBreaches.splice(foundBreaches.findIndex(breach => breach.Name === req.body.featuredBreach),1);
+      }
+      if (foundBreaches.length === 1) {
+        foundBreaches = true;
```
pdehaan: I'm a bit confused here... It seems like `foundBreaches` is either an Array (L15), an Object (L28), or possibly a Boolean (L31). Is that correct?

controllers/user.js
```diff
@@ -0,0 +1,49 @@
...
+
+  res.render("confirm", {
+    title: "Firefox Breach Alerts: Subscribed",
```
pdehaan: Nit: this should possibly be `title: "Firefox Monitor : Subscribed"` for consistency.

tests/oauth.test.js
```diff
@@ -0,0 +1,68 @@
...
+  const mockRequest = { session: { state: { } }, originalUrl: "" };
+  const mockResponse = { render: jest.fn() };
+  const mockFxAClient = { code : { getToken: jest.fn().mockReturnValueOnce({ accessToken: "testToken"}) } };
+  got.mockResolvedValue({ body: `{"email": "${testFxAEmail}"}` });
+
+  await confirmed(mockRequest, mockResponse, mockFxAClient);
+
+  const mockFxACallArgs = mockFxAClient.code.getToken.mock.calls[0];
+  expect(mockFxACallArgs[0]).toBe(mockRequest.originalUrl);
+  expect(mockFxACallArgs[1]).toEqual({state: mockRequest.session.state});
```
pdehaan: This is a lot of array access operators. Not sure if additional destructuring would help, or just pollute here. But then again, it's "only" tests, so maybe I'm overthinking.

```js
const [originalUrl, state] = mockFxAClient.code.getToken.mock.calls[0];
expect(originalUrl).toBe(mockRequest.originalUrl);
expect(state).toEqual({state: mockRequest.session.state});

const {verified, email} = (await DB.getSubscribersByHashes([getSha1(testFxAEmail)]))[0];
expect(verified).toBeTruthy();
expect(email).toBe(testFxAEmail);

// etc.
```

controllers/scan.js
```diff
@@ -0,0 +1,57 @@
...
+      foundBreaches,
+      featuredBreach,
+    });
+  }
+
+  else {
+    res.render("scan", {
+      title: "Firefox Monitor : Scan Results",
+      foundBreaches,
+    });
```
pdehaan: Possibly out of scope for the PR, but I think we discussed briefly in a different thread...
Since L34-38 and L42-45 are so similar, should we combine into one block and just pass a possibly falsy `featuredBreach` to the HBS template?

package.json
pdehaan: I think the only difference is that the resetDB deletes the seeded data after the tests run (I think).

controllers/dockerflow.js
```diff
@@ -0,0 +1,38 @@
...
+
+
+function heartbeat (req, res) {
+  return res.send("OK");
+}
+
+module.exports = {
+  vers,
+  heartbeat,
+}
```
pdehaan:
```
/app/controllers/dockerflow.js
  38:2  error  Missing semicolon  semi
```

tests/dockerflow.test.js
```diff
@@ -0,0 +1,24 @@
...
+
+
+test("GET __version__ calls sendFile", () => {
+  const mockRequest = {};
+  const mockResponse = { sendFile: jest.fn() };
+
+  vers(mockRequest, mockResponse);
+
+  const mockSendFileCallArgs = mockResponse.sendFile.mock.calls[0];
+  expect(mockSendFileCallArgs[0]).toContain("version.json")
```
pdehaan:
```
/app/tests/dockerflow.test.js
  13:60  error  Missing semicolon  semi
```

tests/dockerflow.test.js
```diff
@@ -0,0 +1,24 @@
...
+});
+
+test("GET __heartbeat__ calls send OK", () => {
+  const mockRequest = {};
+  const mockResponse = { send: jest.fn() };
+
+  heartbeat(mockRequest, mockResponse);
+
+  const mockSendCallArgs = mockResponse.send.mock.calls[0];
+  expect(mockSendCallArgs[0]).toBe("OK")
```
pdehaan:
```
/app/tests/dockerflow.test.js
  23:41  error  Missing semicolon  semi
```

tests/oauth.test.js
groovecoder: I like the array de-structuring syntax, but in combination with the jest `mock.calls` it removes the semantics that `calls[0]` is the call.

Can you nest array de-structuring? So ...

```
const [firstCall [originalUrl, state] ] = mockFxAClient.code.getToken.mock;
```

That would be sweet.

controllers/scan.js
groovecoder: Yup, planning to address that after merging this tests branch and the UI branch.

PR #268, opened 2018-08-21
Merged tests and visuals

lesleyjanenorton:


controllers/home.js
```diff
@@ -1,15 +1,19 @@
 "use strict";
 
+const TIPS = require("../tips");
+
 
 function home(req, res) {
-  let breach = null;
+  const passwordTips = TIPS.getPasswordTips();
```
groovecoder: Could get rid of the `getPasswordTips` call by doing:

`tips.js`:
```js
const TIPS = [
  {
    "image": ...
  },
  ...
]

module.exports = TIPS
```

And then in `controllers/home.js`:
```js
  res.render("monitor", {
    title: "Firefox Monitor",
    featuredBreach: featuredBreach,
    passwordTips: TIPS, 
  });
```

views/layouts/default.hbs
```diff
@@ -22,31 +21,43 @@
...
-          {{> sign_up_button }}
-        {{/if}}
+      </div>
+      <div class="half">
+        {{> social_media_sharing_buttons}}
       </div>
     </header>
+    
       {{{ body }}}
-
```
groovecoder: this line.

views/layouts/default.hbs
```diff
@@ -22,31 +21,43 @@
...
+          <li><a id="GitHub-logo" href="https://github.com/mozilla/blurts-server/"><img src="/img/github.svg" alt="Github Logo" /></a></li>
+        </ul>
+        <ul class="fourth">
+          <li>Legal</li>
+          <li><a href="https://www.mozilla.org/about/legal/">Terms</a></li>
+          <li><a href="https://www.mozilla.org/privacy/websites/">Privacy</a></li>
+        </ul>
+        <ul class="fourth">
+          <li>Firefox</li>
+          <li><a href="https://www.mozilla.org/firefox/new/">Download Firefox</a></li>
```
groovecoder: We will want to add `utm_*` campaign params to these links.

groovecoder: Eventually.

views/partials/compromised_accounts.hbs
```diff
@@ -1,23 +1,36 @@
...
-{{else}}
-  <h3>Your Compromised Accounts</h3>
-{{/if}}
+
+{{#if foundBreaches}}
+  <div class="compromised-accounts">
+    <div class="section-wrapper">
+      <h3 class="whole section-headline">
+        {{#if userAccountCompromised }}
+          Other Compromised Accounts Found>
```
groovecoder: Stray `>` character?

public/img/password_tips/password-tip-1.svg
```diff
@@ -0,0 +1,35 @@
...
+                                </g>
+                            </g>
+                        </g>
+                    </g>
+                </g>
+            </g>
+        </g>
+    </g>
+</svg>
```
pdehaan: We may be able to shave a few KB by running these through SVGO.

public/img/what_to_do/what-to-do-4.svg
```diff
@@ -0,0 +1,24 @@
...
+                <g transform="translate(722.000000, 98.000000)">
+                    <circle id="Oval" stroke="url(#linearGradient-1)" stroke-width="2" cx="20" cy="20" r="20"></circle>
+                    <text id="1" font-family="FiraSans-Medium, Fira Sans" font-size="20" font-weight="400" fill="#0A84FF">
+                        <tspan x="15" y="26">4</tspan>
+                    </text>
+                </g>
+            </g>
+        </g>
+    </g>
+</svg>
```
pdehaan: All 4 of these SVGs are the same, except for the `<tspan x="15" y="26">4</tspan>` and `<title>4</title>` tags. Not sure if there is a way to pass a variable into an SVG, or set via content via CSS or something.

tips.js
```diff
@@ -0,0 +1,49 @@
...
+      {
+        "image" : "password-tip-6.svg",
+        "alt": "Sign up for alerts from Firefox Monitor.",
+        "headline": "Sign Up for Alerts From<br>Firefox Monitor",
+        "summary": "We'll let you know if your account information is compromised in a data breach or exposed to hackers in some other way.",
+      },
+    ];
+    return allTips;
+  },
+    };
```
pdehaan: Indenting looks "funny" here.

views/layouts/default.hbs
```diff
@@ -1,18 +1,17 @@
 <!DOCTYPE html>
 <html lang="en">
   <head>
+    <link rel="stylesheet" href="https://code.cdn.mozilla.net/fonts/fira.css">
+    <link rel="stylesheet" href="/css/app.css">
     <meta charset="utf-8" />
```
pdehaan: nit: `<meta charset="utf-8" />` should come first.

views/layouts/default.hbs
lesleyjanenorton: It never gets old

views/partials/compromised_accounts.hbs
lesleyjanenorton: Homeless as all get out.

views/layouts/default.hbs
```diff
@@ -1,18 +1,17 @@
...
     <meta name="viewport" content="width=device-width, initial-scale=1.0, viewport-fit=cover" />
     <meta name="description" content="Firefox Monitor notifies users when their credentials have possibly been leaked or stolen in a data breach. Powered by haveibeenpwned.com." />
     <meta property="og:title" content="Firefox Monitor" />
-    <meta name="twitter:title" content="Firefox Monitor" />
     <meta property="og:description" content="Protect yourself from hackers." />
-    <meta name="twitter:description" content="Protect yourself from hackers." />
-    <meta property="og:url" content="https://monitor.firefox.com" />
-    <meta property="og:image" content="https://monitor.firefox.com/img/firefox-share.png" />
+    <meta property="og:url" content="https://fx-breach-alerts.herokuapp.com/" />
+    <meta property="og:image" content="/img/firefox-share.png" />
```
pdehaan: I _think_ these need to be absolute URLs for when Twitter/Facebook scrapes the content and tries to embed it in a Card.

It'd be nice if we could specify some base URL via an env variable (like the `og:url`) and just use that as the prefix.

views/layouts/default.hbs
```diff
@@ -22,31 +21,43 @@
...
+          <li>Legal</li>
+          <li><a href="https://www.mozilla.org/about/legal/">Terms</a></li>
+          <li><a href="https://www.mozilla.org/privacy/websites/">Privacy</a></li>
+        </ul>
+        <ul class="fourth">
+          <li>Firefox</li>
+          <li><a href="https://www.mozilla.org/firefox/new/">Download Firefox</a></li>
+          <li><a href="https://www.mozilla.org/firefox/mobile/">Download Firefox Mobile</a></li>
+          <li><a href="https://www.mozilla.org/firefox/features/">Features</a></li>
+          <li><a href="https://www.mozilla.org/en-US/firefox/channel/desktop/">Beta, Nightly, Developer Edition</a></li>
```
pdehaan: nit: remove the "/en-US/" for better auto-detecting locale support.

views/monitor.hbs
```diff
@@ -2,25 +2,23 @@
...
   <div class="section-wrapper">
     <div class="half">
-      <div class="banner-content">
-        {{#if breach }}
-          {{> banner_breach }}
-        {{ else }}
-          {{> protect_yourself }}
+      <h2 class="section-headline">Protect yourself from hackers.<br />Start here.</h2>
+        {{#if featuredBreach }}
+          <p class="demi">See if your account was exposed in the <span class="bold">{{ featuredBreach.Name }}</span> breach or other data breaches.</p>
```
pdehaan: Should this be `featuredBreach.Title` (instead of `.Name`)?

views/layouts/default.hbs
```diff
@@ -1,18 +1,17 @@
...
   <head>
+    <link rel="stylesheet" href="https://code.cdn.mozilla.net/fonts/fira.css">
+    <link rel="stylesheet" href="/css/app.css">
     <meta charset="utf-8" />
     <meta name="viewport" content="width=device-width, initial-scale=1.0, viewport-fit=cover" />
     <meta name="description" content="Firefox Monitor notifies users when their credentials have possibly been leaked or stolen in a data breach. Powered by haveibeenpwned.com." />
     <meta property="og:title" content="Firefox Monitor" />
-    <meta name="twitter:title" content="Firefox Monitor" />
     <meta property="og:description" content="Protect yourself from hackers." />
-    <meta name="twitter:description" content="Protect yourself from hackers." />
```
pdehaan: WHy did we lose the `twitter:description`?

views/partials/compromised_accounts.hbs
```diff
@@ -1,23 +1,36 @@
...
-{{else}}
-  <h3>Your Compromised Accounts</h3>
-{{/if}}
+
+{{#if foundBreaches}}
+  <div class="compromised-accounts">
+    <div class="section-wrapper">
+      <h3 class="whole section-headline">
+        {{#if userAccountCompromised }}
+          Other Compromised Accounts Founds
```
pdehaan: "Founds" sounds exotic. Probably should be singular.

views/partials/featured_breach.hbs
```diff
@@ -0,0 +1,15 @@
+{{#if featuredBreach }}
+  <div class="account">
+    <div id="breach-title-wrapper">
+      <div class="image-wrap">
+        <img alt="{{ featuredBreach.Name }}" src="img/logos/{{ featuredBreach.Name }}.{{ featuredBreach.LogoType }}">
+      </div>
+      <span class="breach-headline">{{ featuredBreach.Name }} Breach</span>
```
pdehaan: Should this be `.Name` or `.Title`?

Actually, looking at ./views/partials/breach_listing.hbs (from a few files up), it seems like we do have a few inconsistencies betweenst that partial and this partial:

https://github.com/mozilla/blurts-server/blob/ea4947b4f4a0a5237f1e43267659a5565fad8227/views/partials/breach_listing.hbs#L1-L11

Note that we have stuff like `<span class="medium">Breach date:</span>` vs `<span class="bold">Breach Date</span>`:

```html
<span class="breach-info"><span class="medium">Breach date:</span> {{prettyDate BreachDate }}</span>
```

vs 

```html
<span class="breach-info"><span class="bold">Breach Date</span> {{prettyDate featuredBreach.BreachDate }}</span>
```

controllers/home.js
```diff
@@ -13,7 +12,7 @@ function home(req, res) {
   res.render("monitor", {
     title: "Firefox Monitor",
     featuredBreach: featuredBreach,
-    passwordTips: passwordTips,
+    passwordTips: TIPS,
```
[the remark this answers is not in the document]
lesleyjanenorton: yeah that would probably help things

views/partials/scan_results_blurb.hbs
```diff
@@ -0,0 +1,7 @@
+{{#if foundBreaches }}
+    <p>Your email address appears in our list of compromised accounts. Keep scrolling to see the publicly available breached accounts associated with the email you scanned and subscribe to Firefox Monitor for a full report including sensitive breaches, tips on protecting your accounts, and alerts if your account is compromised in the future.   </p>
```
pdehaan: Super-ultra bonus nit: remove the empty spaces before the closing `</p>` tag.

views/partials/sign_up_banner.hbs
```diff
@@ -0,0 +1,14 @@
+<div id="sign-up-banner">
+  <div class="section-wrapper">
+    <div class="fourth subscribe-icon-wrapper">
+      <img src="img/subscribe-icon.svg" />
+    </div>
+    <div class="half">
+      <span class="medium">Subscribe to Firefox Monitor</span>
+      <p class="demi">Subscribe to Firefox Monitor for a full report on your compromised accounts (including sensitive sites and spam lists).</p>
```
pdehaan: Do we include sensitive sites and spam lists? Or not on the site and only if you subscribe?

views/partials/social_media_sharing_buttons.hbs
```diff
@@ -1,13 +1,12 @@
...
-  </li>
-  <li>
-   <img src="/img/social_media_icons/facebook.svg" />
+    <img class="facebook-button" alt="Share Firefox Monitor on Facebook" src="/img/social_media_icons/social-facebook.svg" />
+    Share
   </li>
   <li>
-    <a href="mailto:?&subject=Firefox%20Monitor%20%3A%20Protect%20Your%20Data&body=Firefox%20Monitor%20%3A%20Get%20Michelle%20to%20whip%20up%20a%20blurb%20about%20Firefox%20Monitor%20that%20goes%20here%3F%20%0A%0Ahttps%3A%2F%2Fmonitor.firefox.com/" target="_blank" rel="noopener noreferrer" aria-label="Email"><img src="/img/social_media_icons/mail.svg" /></a>
+    <a class="twitter-share-button" href="https://twitter.com/intent/tweet?text=Find%20out%20if%20your%20email%20addresses%20have%20been%20hacked%20at%20https://monitor.firefox.com.">
+      <img alt="Share Firefox Monitor on Twitter" src="/img/social_media_icons/social-twitter.svg" />
```
pdehaan: Why does the Facebook image have a `class="facebook-button"` but the Twitter image doesnt need a custom class?

views/partials/compromised_accounts.hbs
```diff
@@ -1,23 +1,36 @@
...
+          {{#each_from_to foundBreaches 4 foundBreaches.length }}
+            <div class="half listings">
+              {{> breach_listing }}
+            </div>
+          {{/each_from_to}}
+        </div>
+          <button id="show-additional-breaches" class="button transparent-button">Show Additional Breaches</button>
+        {{> hibp_attribution }}
+      {{/if_conditional}}
+    </div><!--coses .section-wrapper-->
```
pdehaan: typo "closes"

views/partials/sign_up_banner.hbs
groovecoder: We don't show the sensitive breaches. We can include them in the "Welcome to Firefox Monitor" email.

views/layouts/default.hbs
groovecoder: We can do this (and the upcoming `STATIC_URL_HOST`) like so:

`server.js`:
```js
app.locals.SERVER_URL = AppConstants.SERVER_URL;
```

And then in `default.hbs`:
```html
<meta content="{{ SERVER_URL }}" />
```

public/img/what_to_do/what-to-do-4.svg
lesleyjanenorton: We could! So we could do to the 'What To Do' steps what was done to the 'Password Advice' Tips and pass a variable to one cleaned up inline SVG. I am not aware of a way to style SVGs that are embedded with  tags from the CSS though.

views/partials/compromised_accounts.hbs
lesleyjanenorton: I sware I can spel.

pdehaan:
I'll propose we should probably never use `{{ Name }}` in any user facing place for the .HBS templates and should only use `{{ Title }}`.

I think we're already doing this in most places already, so if that sounds reasonable, it should be an easy fix:

Current status: <kbd>$ git grep "Name }}"</kbd>:

```
views/monitor.hbs:          <p class="demi">See if your account was exposed in the <span class="bold">{{ featuredBreach.Name }}</span> breach or other data breaches.</p>
views/partials/breach_listing.hbs:        <img alt="{{ Name }}" src="img/logos/{{ Name }}.{{ LogoType }}" />
views/partials/featured_breach.hbs:        <img alt="{{ featuredBreach.Name }}" src="img/logos/{{ featuredBreach.Name }}.{{ featuredBreach.LogoType }}">
views/partials/featured_breach.hbs:      <span class="breach-headline">{{ featuredBreach.Name }} Breach</span>
views/partials/scan_form.hbs:      <input id="featured" type="hidden" name="featuredBreach" value="{{ featuredBreach.Name }}"/>
```

- views/monitor.hbs &mdash; should be changed to `{{ featuredBreach.Title }}`
- views/partials/breach_listing.hbs &mdash; `alt=""` tag should be changed to `alt="{{ Title }}"`.
- views/partials/featured_breach.hbs &mdash; `alt` tag should be changed to `alt="{{ featuredBreach.Title }}"`. The `<span>` inner text should be changed to `{{ featuredBreach.Title }}</span>` (removing the "Breach" text, for consistency).
- views/partials/scan_form.hbs &mdash; This seems fair, given that it is a hidden form field and not user facing in any way.

controllers/scan.js
```diff
@@ -2,16 +2,18 @@
 
 const sha1 = require("../sha1-utils");
 const HIBP = require("../hibp");
+const TIPS = require("../tips");
```
pdehaan: Why not just import this as `passwordTips`? Bonus points, you could then use object shorthand syntax below (if you wanted).

controllers/home.js
```diff
@@ -1,15 +1,18 @@
...
+const TIPS = require("../tips");
+
 
 function home(req, res) {
-  let breach = null;
+  let featuredBreach = null;
   if (req.query.breach) {
     const reqBreachName = req.query.breach.toLowerCase();
-    breach = req.app.locals.breaches.filter(breach => breach.Name.toLowerCase() === reqBreachName)[0];
+    featuredBreach = req.app.locals.breaches.filter(breach => breach.Name.toLowerCase() === reqBreachName)[0];
```
pdehaan: Should this use `.find()` a la https://github.com/mozilla/blurts-server/blob/5cb8a1cc8467dd45b0201c05b2729184fcfad2cf/controllers/hibp.js#L12

controllers/scan.js
```diff
@@ -21,27 +23,33 @@ async function post (req, res) {
   });
 
   if (req.body.featuredBreach) {
-    let featuredBreach = req.body.featuredBreach;
-    if(foundBreaches.filter(breach => breach.Name === featuredBreach).length > 0) {
-      featuredBreach = foundBreaches.filter(breach => breach.Name === featuredBreach);
+    featuredBreach = req.app.locals.breaches.filter(breach => breach.Name.toLowerCase() === req.body.featuredBreach.toLowerCase())[0];
```
pdehaan: Same as above? Should we use `.find()`?

https://github.com/mozilla/blurts-server/blob/5cb8a1cc8467dd45b0201c05b2729184fcfad2cf/controllers/hibp.js#L12

pdehaan: I mean... if we're using this pattern in at least 3-4 places, we should probably consider turning it into a helper somewhere. Like `req.app.locals.getBreach(req.body.featuredBreach)`

public/img/what_to_do/what-to-do-4.svg
groovecoder: I pulled this off in 88cc010e17ba3b56383274446dd5a8093802cd4c by creating a handlebars partial - `what_to_do_svg.hbs` - that accepts a `number` argument and renders the SVG inline with the appropriate number.

Goodbye 4 duplicate images!

views/partials/sign_up_banner.hbs
pdehaan: OK, so we'd include past sensitive breaches in the welcome email, and then email them in near-real-time on future sensitive breaches (but still never show them on the site)?

public/img/what_to_do/what-to-do-4.svg
pdehaan: Currently lost in the abyss of https://www.w3.org/TR/SVGParamPrimer/ and trying to figure out if that's a real thing.

Not to say your solution isn't beautiful (if not SVG verbose in the HTML markup), but something like this seems fancy: `<object type="image/svg+xml" data="what_to_do.svg?number=1" />`

Of course, with great fanciness probably comes limited browser support.

groovecoder: It looks like `<object data="what_to_do.svg?number=1" />` would trigger the same HTTP requests that `` tags do. The nice thing about the inline SVG approach is it cuts out the HTTP requests, and the "bloat" of the SVG-in-HTML is hidden by the handlebars partials.

views/partials/social_media_sharing_buttons.hbs
lesleyjanenorton: Because of it's shape and proximity to the Twitter logo. It required especial margining to appear evenly spaced from the 'Share.'

pdehaan: So are you saying the design will break if we add another logo or reorder the logos?

Should we just add some margining to the "Share" text instead?

controllers/scan.js
```diff
@@ -2,46 +2,57 @@
 
 const sha1 = require("../sha1-utils");
 const HIBP = require("../hibp");
+const TIPS = require("../tips");
 
 
 async function post (req, res) {
   const emailHash = req.body.emailHash;
+  const whatToDo = "4";
```
pdehaan: What does this do? I see we seem to be passing it to "scan.hbs" (L46 and L55 below), but I'm not seeing how/where it's _consumed_, and why it's a numbery string (or stringy number, if you prefer).

views/partials/social_media_sharing_buttons.hbs
lesleyjanenorton: No it should look totally fine to add icons + text to either side. The images in that ul are set to margin-right: 8px, with a special case being made for the facebook logo since it's much narrower. We could instead add the spacing to the words ('Share', 'Tweet') but we'd have to wrap them in a <span> tag or something first since they're in <li> or <li><a> tags alongside the icons.

controllers/scan.js
lesleyjanenorton: hahaha it's a relic from when you first mentioned passing variables to svgs. removing!

views/partials/featured_breach.hbs
```diff
@@ -2,13 +2,13 @@
   <div class="account">
     <div id="breach-title-wrapper">
       <div class="image-wrap">
-        <img alt="{{ featuredBreach.Name }}" src="img/logos/{{ featuredBreach.Name }}.{{ featuredBreach.LogoType }}">
+        <img alt="{{ featuredBreach.Title }}" src="img/logos/{{ featuredBreach.Title }}.{{ featuredBreach.LogoType }}">
```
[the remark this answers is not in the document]
lesleyjanenorton: yes and yes and yes.

views/partials/what_to_do_svg.hbs
```diff
@@ -0,0 +1,19 @@
+<svg role="img" width="42px" height="42px" viewBox="0 0 42 42" aria-labelledby="title" version="1.1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink">
+  <title>{{tip_title}}</title>
```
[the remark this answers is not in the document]
lesleyjanenorton: By some miracle or magic it passes itself from one template to the next.

controllers/scan.js
pdehaan: Actually, this is what I was thinking, possibly in HIBP.js:

```js
app.locals.getBreach = (name) => breaches.find(breach => breach.Name.toLowerCase() === name.toLowerCase());
```

Then to consume it, it'd just be the `req.app.locals.getBreach(req.body.featuredBreach);` (or whatever).
But now that I type that out, I'm wondering if a lookup is the best solution, or if `app.locals.breaches` should be a `Map()` so we could just do `app.locals.breaches.has(req.body.featuredBreach)` or `app.locals.breaches.get(reqBreachName)` instead of having to loop.

```sh
$ git grep "app.locals.breaches."

controllers/hibp.js:  const breach = req.app.locals.breaches.find(breach => breach.Name.toLowerCase() === reqBreachName);
controllers/home.js:    featuredBreach = req.app.locals.breaches.filter(breach => breach.Name.toLowerCase() === reqBreachName)[0];
controllers/scan.js:  let foundBreaches = await HIBP.getBreachesForEmail(emailHash, req.app.locals.breaches);
controllers/scan.js:    featuredBreach = req.app.locals.breaches.find(breach => breach.Name.toLowerCase() === req.body.featuredBreach.toLowerCase());
hibp.js:      app.locals.breaches = breaches;
```

Or maybe that's just getting silly.

views/partials/sign_up_banner.hbs
groovecoder: Correct.

views/partials/what_to_do_svg.hbs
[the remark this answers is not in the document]
groovecoder: Partials inherit the context from which they are included. When this is included in the `{{#each passwordTips }}` block of `tips.hbs`, it inherits the individual tip context, so it can render its properties.

groovecoder:
Let's file a follow-up for using `Title` everywhere.

PR #272, opened 2018-08-22
add tests for email-utils

groovecoder:
Also consolidates email settings into a single SMTP_URL


email-utils.js
```diff
@@ -18,65 +18,37 @@ const hbsOptions = {
...
 const EmailUtils = {
-  init() {
-    // Allow a debug mode that will send JSON back to the client instead of sending emails.
-    // eslint-disable-next-line no-process-env
-    if (process.env.DEBUG_DUMMY_SMTP) {
-      console.info("Running in dummy SMTP mode, /user/breached will send a JSON response instead of sending emails.");
+  async init(smtpUrl = AppConstants.SMTP_URL) {
+
+    // Allow a debug mode that will log JSON instead of sending emails.
+    if (smtpUrl === "") {
```
pdehaan: ANy benefits to checking an empty string vs just falsy value, a la `if (!smtpUrl) {..}`?

tests/email.test.js
```diff
@@ -0,0 +1,41 @@
...
+const EmailUtils = require("../email-utils");
+
+
+jest.mock("nodemailer");
+
+
+test("EmailUtils.init with empty host doesnt invoke nodemailer", () => {
+  nodemailer.createTransport = jest.fn();
+
+  EmailUtils.init("");
```
pdehaan: `EmailUtils.init()` is async above, should we do an `await` here (and make parent func async as well)?

```js
async init(smtpUrl = AppConstants.SMTP_URL) {..}
```

Same w/ the 2 other calls below. Doesn't look like `EmailUtils.sendEmail()` is async, so I think that one is fine as-is.

groovecoder: `init` is a bit strange and could use a refactor I think. It returns `Promise.resolve()` values, so even though it's marked as `async` so it can `await` the `gTransporter.verify()`, it returns final values.

groovecoder:
Merging for the +7.5% test coverage. Still not super happy with `EmailUtils.init` but it's working right now and I need to move on to more stuff.

Issue #107, opened 2018-05-10
Add script to remove old unverified emails


groovecoder:
We need a script that will delete old unverified emails. Which means we will also want to add a `created` timestamp on `email_hashes`.

PR #291, opened 2018-08-25
for #107: script to delete unverified subscribers

groovecoder:


pdehaan:
How does this work? Are we calling it via cron job?
I’m a bit confused about timing. If we run this daily at midnight to purge unverified, it feels like we need more checks. Say I create an account at 11:53pm, and am waiting for a verification email, or get sidetracked. 7 minutes later at midnight, the cron job runs, deleting my 7 minute old unverified account before I had a chance to verify. Not sure what our messaging is on trying to verify a deleted account, but I’m guessing we’d just consider it an invalid token and balk.

Not sure if we need some more foot gun protection where we say, delete all unverified accounts where the creation date was over 24h ago. That way, users always have at least a full day to verify their account before we dump it in the bin.

scripts/delete-unverified-subscribers.js
```diff
@@ -0,0 +1,14 @@
...
+
+
+async function deleteUnverifiedSubscribers() {
+  await DB.deleteUnverifiedSubscribers();
+}
+
+
+(async () => {
+  await deleteUnverifiedSubscribers();
+  process.exit();
```
pdehaan: Does this need a `try..catch` in case something goes wrong? I’m fearing the `process.exit()` never executing if the delete errors and having a stalled child process.

scripts/delete-unverified-subscribers.js
```diff
@@ -0,0 +1,14 @@
...
+const DB = require("../db/DB");
+
+
+async function deleteUnverifiedSubscribers() {
+  await DB.deleteUnverifiedSubscribers();
+}
+
+
+(async () => {
+  await deleteUnverifiedSubscribers();
```
pdehaan: Why not just call the `DB.deleteUnverifiedSubscribers()` directly instead of the proxy func?

groovecoder:
Ah yes, this also needs an update to add `createdAt` and `modifiedAt` columns to the `subscribers` table. Then the logic here will change to `where({ verified: false, {createdAt, "<=", Date.now() - 24*60*60*1000}})`

pdehaan:
I found https://github.com/TryGhost/knex-migrator while trying to see if ESLint had any knex specific configs/plugins. I didn’t see Postgres as a supported DB, but wasn’t sure if it wasn’t supported, or not tested.

.env-dist
```diff
@@ -6,6 +6,7 @@ PORT=6060
 COOKIE_SECRET=3895d33b5f9730f5eb2a2067fe0a690e298f55f5e382c032fd3656863412
 
 DATABASE_URL="postgres://postgres@localhost:5432/blurts"
+DELETE_UNVERIFIED_SUBSCRIBERS_TIMER=86400
```
pdehaan: May want a comment here saying this is in seconds.

groovecoder:
I fixed the knex migration - we are supposed to have a timestamp on our `initial_schema.js` migration so `knex` knows the order of the migrations.

PR #297, opened 2018-08-27
fix #251: verify signatures of SNS messages

### LINKED ISSUE #251: Use SES for emails

groovecoder:


middleware.js
```diff
@@ -1,25 +1,5 @@
 "use strict";
 
-const auth = require("basic-auth");
-
-const AppConstants = require("./app-constants");
-
-
-function checkSESAuth (user, pass) {
-  return user === AppConstants.SES_USERNAME && pass === AppConstants.SES_PASSWORD;
```
pdehaan: Are we using `SES_USERNAME` or `SES_PASSWORD` still?


```sh
$ git grep SES_USERNAME
  # .env-dist:SES_USERNAME=""
  # app-constants.js:  "SES_USERNAME",
  # middleware.js:  return user === AppConstants.SES_USERNAME && pass === AppConstants.SES_PASSWORD;

$ git grep SES_PASSWORD
  # .env-dist:SES_PASSWORD=""
  # app-constants.js:  "SES_PASSWORD",
  # middleware.js:  return user === AppConstants.SES_USERNAME && pass === AppConstants.SES_PASSWORD;
```

groovecoder: Good catch! We can remove them now.

package.json
```diff
@@ -38,6 +39,7 @@
     "jest": "^23.5.0",
     "jest-tap-reporter": "^1.9.0",
     "node-mocks-http": "^1.7.0",
+    "nodemon": "^1.18.3",
```
pdehaan: Are we using **nodemon** somewhere?
I didn't see it in any npm scripts or used anywhere specifically.

groovecoder: I started using it locally. This PR includes a new `nodemonConfig` section in `package.json` so we can use `nodemon server.js` to automatically refresh the `server.js` process when files are changed. I'll add it to the README>

groovecoder:
Added nodemon to README. Going to merge this before it bit-rots.

PR #304, opened 2018-08-29
start fx newsletter basket code

groovecoder:


groovecoder:
Dang, this old node module has an insecure dependency chain. I'll see if I can cut out the node module and just `got` directly.

pdehaan:
> Dang, this old node module has an insecure dependency chain.

Submitted upstream https://github.com/mozilla/node-basket/pull/8 to refresh the module versions.

.env-dist
```diff
@@ -22,6 +22,11 @@ SES_CONFIG_SET=""
 # 1: only log messages coming back from SES
 SES_NOTIFICATION_LOG_ONLY=
 
+# Firefox newsletter
+BASKET_URL=https://basket.allizom.org
+BASKET_API_KEY=
+BASKET_NEWLSETTER=mozilla-and-you
```
pdehaan: typo on "NEWLSETTER".

db/DB.js
```diff
@@ -98,14 +99,31 @@ const DB = {
...
       .update({
         verified: true,
         updated_at: knex.fn.now(),
       })
       .returning("*");
+
+    // If the user opted in, send newsletter subscription request to basket
+    if (emailHash.fx_newsletter) {
+      const b = new Basket({BASKET_URL: AppConstants.BASKET_URL, API_KEY: AppConstants.BASKET_API_KEY});
+      // Duplicative calling code; see https://github.com/mozilla/node-basket/issues/4
```
pdehaan: nit: Can you add a noisy "TODO:" prefix on that comment, so we can maybe remember to unduplicative that once the upstream magic is merged and published.

.env-dist
pdehaan: Also, is this the correct newsletter, or do we have some specific Monitor one we're creating?

db/migrations/20180829161115_add_fx_newsletter_column.js
```diff
@@ -0,0 +1,13 @@
+"use strict";
+
+exports.up = function(knex, Promise) {
+  return knex.schema.table("subscribers", table => {
+    table.boolean("fx_newsletter").defaultTo(false);
+  });
+};
+
+exports.down = function(knex, Promise) {
```
pdehaan: Not sure what's up with the odd little unused `Promise` argument in these, but I'll assume it's just a **knex** oddity.

db/DB.js
```diff
@@ -36,9 +37,9 @@ const DB = {
       .where("email", "=", email);
   },
 
-  async addSubscriberUnverifiedEmailHash(email) {
+  async addSubscriberUnverifiedEmailHash(email, fxNewsletter = false) {
     const res = await knex("subscribers").insert(
-      { email: email, sha1: getSha1(email), verification_token: uuidv4(), verified: false }
+      { email: email, sha1: getSha1(email), verification_token: uuidv4(), verified: false, fx_newsletter: fxNewsletter }
```
pdehaan: nit: this line may be starting to be grown up enough to warrant multiple lines for readability.

pdehaan:
Re: breaking CI,

**nsp** has a `--warn-only` flag, which will just return a status code of 0, regardless of sketchy deps. We can use that to unblock us until we can hopefully get the upstream PRs landed.

---

**Semi-related:** Also, I should check our bugs, but my beloved **nsp** is no more ([as of Sept 30ish](https://blog.npmjs.org/post/175511531085/the-node-security-platform-service-is-shutting)). It is being replaced with the <kbd>$ npm audit</kbd> command. So we should probably proactively yank it from my cold dead hands and replace it with the newer, less fun, npm version.

PR #312, opened 2018-09-01
Adds thank you message, radio button pushability, and better tabbing …

### LINKED ISSUE #332: "Thank you for your feedback" message is not displayed after the user unsubscribes
[Affected versions]:
- Firefox 62.0
- Firefox 63.0a1

[Affected Platforms]:
- Windows 10 x64
- macOS 10.13
- Ubuntu 16.04 x64

[Prerequisites]:
You have to be subscribed to Firefox Monitor and have the Unsubscribe page loaded. 

[Steps to reproduce]:
1. Click on the "Unsubscribe" button. 

[Expected result]:
- The user is redirected to "Unsubscribe Confirmation Survey" and then a "Thank you for your feedback" message is displayed.
- https://mozilla.invisionapp.com/share/CBNKNFNJ8E6#/screens/317396218

[Actual result]:
- The user is redirected to the Firefox Monitor home page.

[Regression]:

lesleyjanenorton:
…to unsubscribe form.

Fixes #332

### DIFF
--- a/public/js/monitor.js
+++ b/public/js/monitor.js
@@ -158,7 +158,7 @@ function openModalWindow() {
 function checkBoxStates(checkBoxEvent) {
   checkBoxEvent.preventDefault();
   let checkBox;
-  // user tabs (keyCode === 16) or tabs BACK (keyCode === 16) to checkbox element
+  // user tabs (keyCode === 9) or tabs BACK (keyCode === 16) to checkbox element
   if (checkBoxEvent.keyCode === 9 || checkBoxEvent.keyCode === 16) {
     checkBox = checkBoxEvent.target;
     checkBox.focused = true;
@@ -193,7 +193,7 @@ async function hashEmailAndSend(emailFormSubmitEvent) {
     emailForm.querySelector("input[name=emailHash]").value = await sha1(emailInput.value);
     emailInput.value = "";
   }
-  emailForm.submit();
+  // emailForm.submit();
 }
 
 const addUser = (formEvent) => {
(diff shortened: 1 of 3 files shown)


public/js/monitor.js
```diff
@@ -193,7 +193,7 @@ async function hashEmailAndSend(emailFormSubmitEvent) {
     emailForm.querySelector("input[name=emailHash]").value = await sha1(emailInput.value);
     emailInput.value = "";
   }
-  emailForm.submit();
+  // emailForm.submit();
```
groovecoder: Do we need this line at all?

public/js/monitor.js
```diff
@@ -226,29 +226,53 @@ const postData = (url, data = {}) => {
...
+  additionalBreaches.classList.toggle("show-breaches");
+  //setting height this way enables transition easing... setting the new height to "auto" 
+  if (additionalBreaches.classList.contains("show-breaches")) {
+    additionalBreaches.style.height = additionalBreaches.scrollHeight + "px";
+  }
+}
+
+const handleRadioButtons = function(form) {
+  const inputFields = form.querySelectorAll(".radio-button-group, .button");
+  inputFields[0].querySelector("input").checked = true;
```
groovecoder: Do we need a choice to be pre-checked? It may lead users to over-select this choice.

lesleyjanenorton: This was in accordance with a Photon radio-button stipulation that one should always be checked, but I wondered about over-selection/lazy-selection also.

public/js/monitor.js
lesleyjanenorton: Yes. Particularly. Uncommenting!

public/js/monitor.js
```diff
@@ -226,29 +226,53 @@ const postData = (url, data = {}) => {
...
+  for (let x = 0; x < inputFields.length ; x++) { 
+    const input = inputFields[x];
+    input.addEventListener("focus", (e) => {
+      input.addEventListener("keydown", (e) => {
```
pdehaan: Do we need to remove the `keydown` listener anywhere?
Like, if we focus the input 3 times, will it add the `keydown` event listener 3 times?

public/js/monitor.js
groovecoder: Let's remove it ... that may apply more/better in Firefox UI where we have a good idea of sensible defaults. Here we'd like to keep it open-ended.

lesleyjanenorton:
Added keyCode commenting, removed default selected radio button, and added error message on submissions with no option selected.

PR #328, opened 2018-09-05
for #55: add SameSite flag to cookies

groovecoder:


pdehaan:
I think this looks reasonable, per https://www.owasp.org/index.php/SameSite.

I think it's an r+ unless you want to run it past secops or @moz-jvehent or somebody.

Although #55 says this, so maybe "lax" is a preapproved value:

> Session cookies must have HttpOnly and Secure flags set and the SameSite attribute set to 'strict' or 'lax' (which allows external regular links to login).

PR #352, opened 2018-09-05
fix #346 fix #279: strip tokens and hashes from ga pageview

### LINKED ISSUE #279: Remove utm_* url params after sending pageview to GA
When https://github.com/mozilla/blurts-addon/issues/96 lands, we will want to remove the `utm_*` url params *after we send the ping to Google Analytics*.

See MDN code for example:

https://github.com/mozilla/kuma/blob/7296bee281dbfee76fd675f0ca47d60598d2fcc5/jinja2/includes/google_analytics.html#L55-L65

### LINKED ISSUE #346: Strip token and hash url params before sending pageview to GA
Like https://github.com/mozilla/blurts-server/issues/279. We don't want to leak these values out to GA.

groovecoder:
remove utm_* url params after sending pageview to ga


groovecoder:
I'm not super happy with this approach ... but I'm not sure I can think of a better approach.

public/js/tracking_protection.js
```diff
@@ -27,13 +27,23 @@ function browserName() {
...
-  
-  if (browserName() === "firefox") {
-    if (document.getElementById("download-firefox")) {
-    document.getElementById("download-firefox").classList.add("hide");
-    }
+
+
+function removeUtms() {
+  const win = window;
+  const loc = window.location;
```
pdehaan: nit: Should this be `win.location`? A bit pointless though, I s'pose.

public/js/tracking_protection.js
```diff
@@ -27,13 +27,23 @@ function browserName() {
...
+  if (loc.href.indexOf("utm") !== -1 && win.history.replaceState) {
```
pdehaan: Do we need the entire `href`? Or can we limit it to `loc.href.search` query string? Should we check for just "utm" or "utm_"? I guess my fear is that we'd have some breach name which just happens to coincidentally have "utm" string somewhere in the Name and we accidentally get into this flow.

Also, we could possibly use something like [`String#includes()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/includes), a la:

```js
if (loc.search.includes("utm_") && win.history.replaceState) {..}
```

Not sure if it'd be worth it to try parsing the entire search query into an object and verifying that the query parameter name starts with some `/^utm_/i` regexp.

public/js/tracking_protection.js
```diff
@@ -44,5 +54,18 @@ if (!_dntEnabled()) {
...
 	ga("set", "anonymizeIp", true);
-	ga("send", "pageview");
+
+
+  // Strip url params if they contain token or hash
+  const loc = document.location;
+  let pageValue = loc.pathname + loc.search;
+  if (document.location.search &&
+      (document.location.search.indexOf("token") !== -1 ||
+       document.location.search.indexOf("hash") !== -1)
```
pdehaan: Same here, we could possibly use [`String#includes()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/includes) which returns a Boolean (and use our handy `loc` variable from L60):

```js
if (loc.search && (loc.search.includes("token") || loc.search.includes("hash"))) {..}
```

public/js/tracking_protection.js
```diff
@@ -44,5 +54,18 @@ if (!_dntEnabled()) {
...
 	ga("create", "UA-77033033-16");
 	ga("set", "anonymizeIp", true);
-	ga("send", "pageview");
+
+
+  // Strip url params if they contain token or hash
+  const loc = document.location;
+  let pageValue = loc.pathname + loc.search;
+  if (document.location.search &&
+      (document.location.search.indexOf("token") !== -1 ||
```
pdehaan: Sorry, I'm not 100% certain what we're trying to solve, but this feels almost too greedy.
What would happen if we had a breach Name like `/?breach=Bitcointokens`?

We don't seem to have any matches currently, but it could happen at any random time based on the breach Name that Troy gives a future breach.

```js
const got = require("got");

async function main(prod=false) {
  const breaches = await got(prod ? "https://haveibeenpwned.com/api/v2/breaches" : "https://stage.haveibeenpwned.com/api/v2/breaches", {json:true});

  breaches.body
    .filter(b => b.Name.includes("token") || b.Name.includes("hash"))
    .forEach(b => console.log(b.Name));
}

main();
```

groovecoder: Trying to strip the url params off the `pageValue` sent to Google Analytics when the url params include either the user `token` or the user `hash`, which we want to prevent from "leaking" out as much as possible. (This also makes it easier to see the overall aggregate user flow instead of seeing tons of separate pageviews at individualized `/user/verify?token=...` and `/user/unsubscribe?token=...` urls)

I updated to look for `token=` or `hash=` in the url. And I also updated to set both the `location` and `page` properties of ga, and to trim the `Referer` header of calls to GA so they don't leak any of these values either.

groovecoder:
Note: we should check this when it lands on heroku to make sure that trimming the `Referer` header value doesn't strip the pageview locations from the Google Analytics reports. It shouldn't - we're setting and sending the `page` and `location` values in the calls to `ga` ourselves - but just want to make sure.

pdehaan:
I think this looks good. The only other approach I can think of is something like [`URLSearchParams`](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams?token=peter&hash=blerp&utm_content=fancy):

```js
const urlParams = new URLSearchParams(loc.search);
if (urlParams.has("token") || urlParams.has("hash")) {..}
```

PR #366, opened 2018-09-06
Fix reverify error 353

groovecoder:


views/error.hbs
```diff
@@ -0,0 +1,6 @@
+{{!< default }}
+<div class="sub-page-content section-wrapper">
+    <h2>Error</h2>
+    <p> {{ error }} </p>
```
pdehaan: Because I was initially scared and confused, we may want to rename this to `{{ message }}` here, and above in `res.render("error", { error: err.message });`.

I originally thought it was logging an Error object/stack, which is generally a pretty :-1: user experience.

middleware.js
```diff
@@ -9,6 +9,34 @@ function asyncMiddleware (fn) {
...
+  console.log("logErrors");
+  console.error(err.stack);
+  next(err);
+}
+
+
+function clientErrorHandler (err, req, res, next) {
+  console.log("clientErrorHandler");
+  if (req.xhr) {
+    res.status(500).send({ error: err });
```
pdehaan: Should.... Should this be `{ error: err }`, or `{error: err.message }`?
How much stack info are we meaning to potentially disclose?

groovecoder:
Updated.

PR #367, opened 2018-09-06
fix #310: reload breaches to find breach; else error

### LINKED ISSUE #310: Make /hibp/notify resilient to missing breach data
See https://github.com/mozilla/blurts-server/issues/103#issuecomment-417466415 ...

* [x] Request latest breaches.json data when we receive the notification
* [x] If the breach still isn't recognized, respond with a better error message

groovecoder:


groovecoder:
To test this, you can throw an HTTP POST at the `/hibp/notify` endpoint with a breachname that doesn't exist. E.g.,
<details>
<summary>
<pre>
curl -verbose -H "Content-Type: application/json" --data '{"hashPrefix":"323928","breachName":"SvenskaMagick","hashSuffixes":["d826eb487983dce672e87fd5ab306caadb"]}' http://localhost:6060/hibp/notify
</pre>
</summary>
<pre>
*   Trying ::1...
* TCP_NODELAY set
* Connected to localhost (::1) port 6060 (#0)
> POST /hibp/notify HTTP/1.1
> Host: localhost:6060
> User-Agent: curl/7.54.0
> Accept: */*
> Referer: rbose
> Content-Type: application/json
> Content-Length: 106
>
* upload completely sent off: 106 out of 106 bytes
< HTTP/1.1 500 Internal Server Error
< X-DNS-Prefetch-Control: off
< X-Frame-Options: SAMEORIGIN
< Strict-Transport-Security: max-age=15552000; includeSubDomains
< X-Download-Options: noopen
< X-Content-Type-Options: nosniff
< X-XSS-Protection: 1; mode=block
< Content-Security-Policy: base-uri 'none'; default-src 'none'; connect-src 'self' https://code.cdn.mozilla.net/fonts/ https://www.google-analytics.com; font-src 'self' https://code.cdn.mozilla.net/fonts/; frame-ancestors 'none'; img-src 'self' https://www.google-analytics.com; script-src 'self' https://www.google-analytics.com/analytics.js; style-src 'self' https://code.cdn.mozilla.net/fonts/; report-uri /__cspreport__
< X-Content-Security-Policy: base-uri 'none'; default-src 'none'; connect-src 'self' https://code.cdn.mozilla.net/fonts/ https://www.google-analytics.com; font-src 'self' https://code.cdn.mozilla.net/fonts/; frame-ancestors 'none'; img-src 'self' https://www.google-analytics.com; script-src 'self' https://www.google-analytics.com/analytics.js; style-src 'self' https://code.cdn.mozilla.net/fonts/; report-uri /__cspreport__
< X-WebKit-CSP: base-uri 'none'; default-src 'none'; connect-src 'self' https://code.cdn.mozilla.net/fonts/ https://www.google-analytics.com; font-src 'self' https://code.cdn.mozilla.net/fonts/; frame-ancestors 'none'; img-src 'self' https://www.google-analytics.com; script-src 'self' https://www.google-analytics.com/analytics.js; style-src 'self' https://code.cdn.mozilla.net/fonts/; report-uri /__cspreport__
< Referrer-Policy: strict-origin-when-cross-origin
< Content-Type: application/json; charset=utf-8
< Content-Length: 48
< ETag: W/"30-/iYxJ1VmMj+LdMbfQXoycLb7LiY"
< Date: Thu, 06 Sep 2018 21:41:22 GMT
< Connection: keep-alive
<
* Connection #0 to host localhost left intact
{"message":"Unrecognized breach: svenskamagick"}
</pre>
</details>

controllers/hibp.js
```diff
@@ -12,7 +12,16 @@ async function notify (req, res) {
   const subscribers = await DB.getSubscribersByHashes(hashes);
 
   const reqBreachName = req.body.breachName.toLowerCase();
-  const breach = req.app.locals.breaches.find(breach => breach.Name.toLowerCase() === reqBreachName);
+  let breach = req.app.locals.breaches.find(breach => breach.Name.toLowerCase() === reqBreachName);
+  if (!breach) {
+    // If breach isn't found, try to reload breaches from HIBP
+    await HIBP.loadBreachesIntoApp(req.app);
+    breach = req.app.locals.breaches.find(breach => breach.Name.toLowerCase() === reqBreachName);
```
pdehaan: Not sure if we can find a better pattern for this, or hide it in a helper function somewhere...

I think this is now the 3rd or 4th time we're using similar code to filter breaches:

```sh
$ git grep -n locals.breaches.find

controllers/hibp.js:15:  const breach = req.app.locals.breaches.find(breach => breach.Name.toLowerCase() === reqBreachName);
controllers/scan.js:26:    featuredBreach = req.app.locals.breaches.find(breach => breach.Name.toLowerCase() === req.body.featuredBreach.toLowerCase());
```

controllers/hibp.js
```diff
@@ -12,7 +12,16 @@ async function notify (req, res) {
   const subscribers = await DB.getSubscribersByHashes(hashes);
 
   const reqBreachName = req.body.breachName.toLowerCase();
-  const breach = req.app.locals.breaches.find(breach => breach.Name.toLowerCase() === reqBreachName);
+  let breach = req.app.locals.breaches.find(breach => breach.Name.toLowerCase() === reqBreachName);
+  if (!breach) {
+    // If breach isn't found, try to reload breaches from HIBP
+    await HIBP.loadBreachesIntoApp(req.app);
```
pdehaan: Are we throttling this somehow? Like, if I call some API endpoint with a bad breach name and cause it to scan our local breaches and then hit Troy's endpoints, but then do that at scale...

Or did we have some code in place to make sure we aren't hitting HIBP more than once per every 5 minutes?

controllers/hibp.js
```diff
@@ -12,7 +12,16 @@ async function notify (req, res) {
...
+    breach = req.app.locals.breaches.find(breach => breach.Name.toLowerCase() === reqBreachName);
+    if (!breach) {
+      // If breach *still* isn't found, we have a real error
+      throw new Error("Unrecognized breach: " + reqBreachName);
+    }
+  }
```
pdehaan: On L10-12 above we're doing something like:

```js
const hashes = req.body.hashSuffixes.map(suffix => req.body.hashPrefix + suffix);
const subscribers = await DB.getSubscribersByHashes(hashes);
```

Should that move down here after the `throw new Error()`? Otherwise, it seems like we're doing all the heavy lifting of fetching a bunch of subscribers based on hashes for a possibly invalid breach.

middleware.js
```diff
@@ -17,7 +17,7 @@ function logErrors (err, req, res, next) {
 
 
 function clientErrorHandler (err, req, res, next) {
-  if (req.xhr) {
+  if (req.xhr || req.headers["content-type"] === "application/json") {
```
pdehaan: Sorry, I may have missed this earlier... but do we need to check `if (err && req.xhr || req.headers...)` here?
What's the likelihood that `err` is `undefined`, but either `req.xhr` or `req.headers["content-type"] === "application/json"` is truthy, which would cause `err.message` to be accessing an the ".message" property on an undefined object (`err`).

groovecoder: Hmm ... I'm not sure. According to [Express error handling guide](https://expressjs.com/en/guide/error-handling.html) these handlers that are appended after all other middleware and routes will only be reached by thrown `Error`s, so in theory the `err` object will always exist.

groovecoder:
Merging to get this into a 1.9 release.

controllers/hibp.js
```diff
@@ -7,13 +7,21 @@ const HIBP = require("../hibp");
 
 
 async function notify (req, res) {
-  const hashes = req.body.hashSuffixes.map(suffix=>req.body.hashPrefix + suffix);
+  const reqBreachName = req.body.breachName.toLowerCase();
+  let breach = HIBP.getBreachByName(req.app.locals.breaches, reqBreachName);
+  if (!breach) {
+    // If breach isn't found, try to reload breaches from HIBP
+    await HIBP.loadBreachesIntoApp(req.app);
+    breach = HIBP.getBreachByName(req.app.locals.breaches, reqBreachName);
```
pdehaan: Pretty sure we've discussed this before, but...

I'm not exactly sure how this works. AFAICT, logos are added manually, and we were experimenting w/ converting the SVGs to PNGs due to email compatibility issues. So if the breach isn't found in our current dataset and we sync w/ HIBP and then send out the email notifications for the new breach, we're going to be sending out emails w/ broken logo image for the latest breach.

/cc @groovecoder @lesleyjanenorton

PR #375, opened 2018-09-07
fix #360: FXA_ENABLED flag and default to false

### LINKED ISSUE #360: Subscribing with FF accounts the user is redirected directly to the “Confirmed! You’re Subscribed”
[Affected versions]:
- Firefox 62.0
- Firefox 64.0a1

[Affected Platforms]:
- Windows 10 x64
- macOS 10.13
- Ubuntu 16.04 x64

[Steps to reproduce]:
1. Navigate to https://fx-breach-alerts.herokuapp.com/
2. lick on the "Subscribe" button and then on "Subscribe with Firefox Account".
3. Log in with a valid account and observe behaviour. 

[Expected result]:
- The user is asked to verify the subscription as it is in the case of subscribing with an email address. 

[Actual result]:
- The user is redirected directly to the “Confirmed! You’re Subscribed” page and it’s not asked to confirm the account first.

[Regression]:

groovecoder:


pdehaan:
Small nit, but can we add a note to the README.md about this new flag?

Probably somewhere adjacent to: https://github.com/mozilla/blurts-server#firefox-accounts

groovecoder:
Added to README. Will disable on dev too so dev + stage are as close to prod as possible.

Issue #353, opened 2018-09-05
"Error: Token not found" is triggered on website if the user unsubscribes then verifies the subscription again 


CiprianGeorgiu:
[Affected versions]:
- latest Nightly 64.0a1
- Beta 63.0b3
- Release 62.0

[Affected Platforms]:
- Windows 10 x64
- macOS 10.13
- Ubuntu 16.04 x64

[Steps to reproduce]:
1. Launch Firefox.
2. Go to https://fx-breach-alerts.herokuapp.com/
3. Subscribe with a email address.
4. Click on the "Verify Your Subscription" button from confirmation email.
5. Click on the "Unsubscribe" button from the confirmation email.
6. Click again the "Unsubscribe" button from the webpage.
7. Repeat step 4.

[Expected result]:
- "Confirmed! You're Subscribed" page is displayed or a message saying that the user is already subscribed.

[Actual result]:
- The following error is triggered on the page: `Error: Token not found
    at Object.verifyEmailHash (/app/db/DB.js:54:13)`

[Regression]:
- Not a regression.

[Additional info]:
- see screenshot:
![screenshot_4](https://user-images.githubusercontent.com/24842369/45122409-642c2880-b16c-11e8-95ee-25d4161594d3.png)

groovecoder:
I can improve the error message and the display, but we can't verify the user after they've unsubscribed: when they unsubscribe, we remove all their data completely - including their verification token. And we don't want to keep any of their data after they unsubscribe.

CiprianGeorgiu:
Verified fixed on 62.0, across platforms: Windows 10 x64, macOS 10.13 and Ubuntu 16.04 x64.

PR #376, opened 2018-09-07
fix #364: show error when unsubscribe token isn't found

### LINKED ISSUE #364: You can unsubscribe from  Firefox Monitor with the same account more than once
[Affected versions]:
- Firefox 62.0
- Firefox 64.0a1

[Affected Platforms]:
- Windows 10 x64
- macOS 10.13
- Ubuntu 16.04 x64

[Steps to reproduce]:
1. Navigate to https://fx-breach-alerts.herokuapp.com/
2. Click on the Subscribe button.
3. Write a valid email address and confirm the subscription.
4. From the report email "Unsubscribe" and see the process through and close the tab. 
5. Repeat step 4.

[Expected result]:
- An error that the email is already unsubscribed to Firefox Monitor should be displayed. 

[Actual result]:
- The Firefox Monitor homepage is loaded. 

[Regression]:

[Notes]:
- Considering the fact that these emails are already being monitored, there should be a way to notify people that their email is already unsubscribed to this service.

groovecoder:


groovecoder:
Also MOAR TESTS!

controllers/user.js
```diff
@@ -64,7 +64,12 @@ async function verify(req, res) {
 }
 
 
-function getUnsubscribe(req, res) {
+async function getUnsubscribe(req, res) {
+  const subscriber = await DB.getSubscriberByToken(req.query.token);
+  if (!subscriber) {
+    throw new Error("Verification token not found");
```
pdehaan: I mentioned this below too, but almost seems more "in-context" up here...

Should this be an "Invalid token" error, and we should have a separate error/`throw` for the token being "missing"?

Or maybe I'm getting confused with "found" wording. As a user, I supplied a token, but maybe it had a typo or something (missed that last character/digit). But I guess I see the other definition of the token was not found in the database.

groovecoder: In these case I mean "not found *in our database*" ... but we can update the message. 

@mheubush - what do you think the error message should say? See https://github.com/mozilla/blurts-server/issues/364 and https://github.com/mozilla/blurts-server/issues/353 for the user flows that can lead to this error.

groovecoder:
Updated with newer better error message from https://github.com/mozilla/blurts-server/issues/364#issuecomment-420018292

PR #377, opened 2018-09-09
Email updates

lesleyjanenorton:
Email Stuff: 
- Email template overhaul to visually match specs
- Categorizes and presents breaches in 2 lists : Website Breaches & Spam Lists 
- Finishes breachALERT email (to do: some great snazziness for fetching/converting/writing new breach images).

Email breach lists will call for only PNG logos, which I will push up in a separate PR. 

Other stuff: 
- Removes SVG password tip stuff in website partials and replaces with PNGS

PR #381, opened 2018-09-09
Add PNG logos for emails +

lesleyjanenorton:
PNG logos for #377


pdehaan:
Oof, thats a lot of new files.

So is the plan that we're keeping the SVGS+JPG+GIF images and adding fallback PNGs for email compatibility?

Looks like we added one new SVG (presumably a new breach) and modified a few existing PNGs... What did we do in this PR? Conversion and optimization?

I'll have to think about how I can make some script to compare all the logos and make sure the GIFs+JPGs have PNGs or if just the SVGs have the PNG fallback images, and make sure none are missing.

Not sure if we need to add any README.md docs on how we're generating+converting these images for future breaches.

Also, how we're planning on supporting new breaches when we send out emails. Like, what if the image from HIBP is a SVG? Unless we're grabbing new breach logos on the fly and converting them on the server, I imagine the image in the email would just 404 since the image hasn't been added to the site yet and we haven't merged the asset PR into dev->stage->prod.

pdehaan:
If you want to see the side by side logos (in Markdown), you can do something like this:

```js
const fs = require("fs");

const LOGOS = fs.readdirSync(__dirname)
  .filter(file => !file.endsWith(".js"));

console.log("IMAGE | IMAGE.PNG\n------|-------------");
LOGOS
  .filter(file => !file.endsWith(".png"))
  .forEach(file => {
    const pngFile = file.replace(/\.(.*)$/, ".png");

    console.log(`![](https://rawgit.com/mozilla/blurts-server/PNG-Logos/public/img/logos/${file}) | ![](https://rawgit.com/mozilla/blurts-server/PNG-Logos/public/img/logos/${pngFile})`);
    // if (!LOGOS.includes(pngFile)) {
    //   console.log(`Missing ${pngFile}`);
    // }
  });
```

A bunch of the images are difficult to see on GitHub due to the white table backgrounds. Not sure if it'd be easier to create a basic HTML file instead w/ a darker background to see how the PNGs look versus their source SVGs.

groovecoder:
Filed https://github.com/mozilla/blurts-server/issues/387 for following up for new breach logos for notification emails.

PR #420, opened 2018-09-14
throw friendly errors in /user/add

groovecoder:
I ran OWASP ZAP at my localhost and also found another error trace when email was left off.

### DIFF
--- a/controllers/user.js
+++ b/controllers/user.js
@@ -14,7 +14,7 @@ const TIPS = require("../tips");
 async function add(req, res) {
   const email = req.body.email;
 
-  if (!isemail.validate(email)) {
+  if (!email || !isemail.validate(email)) {
     throw new Error("Invalid Email");
   }
   const fxNewsletter = Boolean(req.body.additionalEmails);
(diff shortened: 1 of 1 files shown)


controllers/user.js
```diff
@@ -14,7 +14,7 @@ const TIPS = require("../tips");
 async function add(req, res) {
   const email = req.body.email;
 
-  if (!isemail.validate(email)) {
+  if (!email || !isemail.validate(email)) {
```
pdehaan: Me thinks this works!

```
$ curl -XPOST http://localhost:6060/user/add | grep -A1 "<h2>Error</h2>"

    <h2>Error</h2>
    <p> Invalid Email </p>
```

controllers/user.js
```diff
@@ -37,6 +37,9 @@ async function add(req, res) {
 
 
 async function verify(req, res) {
+  if (!req.query.token) {
+    throw new Error("Verification token is required.");
+  }
```
pdehaan: 👍 

```sh
$ curl http://localhost:6060/user/verify | grep -A1 "<h2>Error</h2>"

    <h2>Error</h2>
    <p> Verification token is required. </p>
```

pdehaan: The error messages are friendlier, but not sure if we want to add UUID validation or anything, or if that is for a future PR...

Like, if I specify a bogus/typoed `?token=3` query param, it seems to do a DB query on the invalid UUID.v4.

```bash
$ curl https://fx-breach-alerts.herokuapp.com/user/verify\?token\=3 | grep -A1 "<h2>Error</h2>"

    <h2>Error</h2>
    <p> This email address is not subscribed to Firefox Monitor. </p>
```

groovecoder:
Merging this as it fixes OWASP scan alerts. Will leave #255 open.

PR #432, opened 2018-09-20
Check hibp scan response 415

groovecoder:
Previously, the `hibp` module code caught and logged errors but didn't throw them up to calling code.

This meant that, for example, a `kAnonReq` scan that encountered an error connecting to HIBP (as we saw when our load test was throttled by Cloudflare) returned an `undefined` response, which was taken by the calling code to mean 0 breach results!

This PR:

* Introduces `HIBP._throttledGot` which adds a configurable throttle on HIBP requests to help respect throttled responses
* Throws HIBP connection errors up to calling code with a friendly message: "Error connecting to HIBP."


groovecoder:
Working on Travis failures ...

hibp.js
```diff
@@ -23,17 +23,37 @@ const HIBP = {
...
 
+  async _throttledGot (url, reqOptions, tryCount = 1) {
+    try {
+      return await got(url, reqOptions);
+    } catch (err) {
+      console.error("got an error: " + err);
+      if (err.statusCode === 429) {
+        console.log("got a 429, tryCount: ", tryCount);
+        if (tryCount >= AppConstants.HIBP_THROTTLE_MAX_TRIES) {
+          throw new Error(err.message);
```
pdehaan: Should we throw a _new_ error, or just rethrow the existing error?

```js
throw err;
```

Not sure if there is anything useful in the original stack or not. 🤷‍♂️

groovecoder: Good idea. Changed to `console.error` the full stack, but throw a more user-friendly "Too many connections to HIBP." error to calling code.

.env-dist
```diff
@@ -44,3 +44,7 @@ HIBP_RELOAD_BREACHES_TIMER=600
 # HIBP API for range search and subscription
 HIBP_KANON_API_ROOT="https://api.haveibeenpwned.com"
 HIBP_KANON_API_TOKEN=""
+# How many seconds to wait before retrying an HIBP request
```
pdehaan: Sorry, small nit/question...

Did you mean "seconds" or "milliseconds"? I believe the code below implies milliseconds, whereas if it's seconds, we're talking about a 33.33 minute backoff.

PR #433, opened 2018-09-20
fix #431: fixes for basket API calls

### LINKED ISSUE #431: Changes to basket integration
I noticed some changes that need to be made for the basket integration to work properly in `basket.js`.

* The parameter basket expects is `optin: "Y"` instead of `validated: true`. See the [basket docs for subscribe](https://basket.readthedocs.io/newsletter_api.html#news-subscribe).
* Basket requires that the content-type of the POST request be `application/x-www-form-urlencoded`. I don't see that being set, but it could be something the request library does.

Looks good apart from those things. Thanks!

groovecoder:


basket.js
```diff
@@ -9,9 +9,11 @@ const Basket = {
 
   async subscribe(email) {
     const params = {
+      form: true,
       email,
       newsletters: AppConstants.BASKET_NEWSLETTER,
       validated: true,
```
pmac: this parameter will be ignored by basket. I've just realized that it's in the docs, but it was removed from the code at some point. I'll update basket's docs, but you can remove it here as well if you want. It won't break anything, to be clear, but it is superfluous at this point.

basket.js
```diff
@@ -9,9 +9,11 @@ const Basket = {
 
   async subscribe(email) {
     const params = {
+      form: true,
```
pmac: I can't tell what this is doing. It appears to be sending `form=true` to basket in the POST, but I don't think that will affect anything.

basket.js
pmac: Oh! this is an option to `got`. I think this should go down in the `reqOptions` object.

basket.js
pmac: Oh! this is an option to got. I think this should go down in the reqOptions object.

basket.js
groovecoder: removed.

basket.js
groovecoder: Moved.

groovecoder:
Merging to spot-check on heroku.

PR #435, opened 2018-09-20
Prefix & Minify statics

groovecoder:
cherry-picked @lesleyjanenorton commits from https://github.com/mozilla/blurts-server/tree/prefix-and-minify to open PR, review, test, and discuss.


package.json
```diff
@@ -28,9 +28,14 @@
     "nodemailer-express-handlebars": "^3.0.0",
     "pg": "^7.4.3",
     "sns-validator": "^0.3.4",
+    "update": "^0.7.4",
```
groovecoder: Is this actually being used?

package.json
```diff
@@ -73,6 +83,17 @@
...
+    "js:concat": "concat -o public/dist/app.min.js public/js/*.js",
+    "css:minify": "cleancss -o public/dist/app.min.css public/dist/app.min.css",
+    "css:autoprefixer": "postcss -u autoprefixer --autoprefixer.browsers '> 5%, ie 11' -r public/dist/app.min.css --no-map",
+    "css:vars": "postcss -u postcss-css-variables -o public/dist/app.min.css public/css/app.css",
+    "build:js": "run-s js:concat js:minify",
+    "build:css": "run-s css:vars css:autoprefixer css:minify",
+    "watch:js": "onchange 'public/js/monitor.js' -- npm run build:js",
+    "watch:css": "onchange 'public/css/app.css' -- npm run build:css",
+    "watch:all": "run-p watch:css watch:js",
+    "build:all": "run-s build:js build:css watch:all",
```
groovecoder: Drop `watch:all` from this - when I ran this I expected the `build:all` command to return.

package.json
```diff
@@ -73,6 +83,17 @@
...
+    "js:minify": "minify public/dist/app.min.js -o public/dist/app.min.js",
+    "js:concat": "concat -o public/dist/app.min.js public/js/*.js",
+    "css:minify": "cleancss -o public/dist/app.min.css public/dist/app.min.css",
+    "css:autoprefixer": "postcss -u autoprefixer --autoprefixer.browsers '> 5%, ie 11' -r public/dist/app.min.css --no-map",
+    "css:vars": "postcss -u postcss-css-variables -o public/dist/app.min.css public/css/app.css",
+    "build:js": "run-s js:concat js:minify",
+    "build:css": "run-s css:vars css:autoprefixer css:minify",
+    "watch:js": "onchange 'public/js/monitor.js' -- npm run build:js",
+    "watch:css": "onchange 'public/css/app.css' -- npm run build:css",
+    "watch:all": "run-p watch:css watch:js",
```
groovecoder: nice proper uses of `run-p` and `run-s`.

package.json
pdehaan: What does **update** do? I can't see any other references to it in this PR.

.eslintignore
```diff
@@ -2,3 +2,4 @@ db/migration
 public/js/vendor
 coverage
 loadtests
+public/dist
```
pdehaan: We may want to add this to .gitignore as well.

public/js/analytics_dnt-helper.js
```diff
@@ -1,3 +1,6 @@
+/*eslint-disable no-var */
```
pdehaan: Why was this file moved/renamed?
It will probably affect the .eslintignore file since we were previously ignoring this file, IIRC.

package.json
lesleyjanenorton: I copied you guys from several lines below.

public/js/analytics_dnt-helper.js
lesleyjanenorton: and here I was wondering why eslint had suddenly taken an interest in this file. It was moved and renamed for the purpose of concatenating the files in the same order they're linked in the website. They're is probably a cooler way to handle this.

package.json
lesleyjanenorton: Removed with no apparent consequences. I actually don't remember this specifically but I expect it's leftover clutter from minification experimentation.

public/js/analytics_dnt-helper.js
pdehaan: > They're is probably a cooler way to handle this.

I imagine we could use **webpack** or something to create a fancy bundle, but this all may be too risky to add for launch.

lesleyjanenorton: and by <i>They're</i> I meant <i>There</i>. How embarrassing.

lesleyjanenorton:
@groovecoder I nixed autoprefixer per our convo on Slack

.gitignore
```diff
@@ -8,3 +8,4 @@ breach_hashsets
 version.json
 loadtests/__pycache__
 loadtests/venv
+public/dist
```
groovecoder: This conflicts with the additions of `public/dist/app.min.[js|css]` below. If we're ever going to check those in, we should keep them in git history.

groovecoder:
If we don't check in the generated `public/dist` directory we still need to:

1. Update `Dockerfile` to include `npm run build:all` before the `COPY . /app` - so the generated `public/dist` files will be included in the Docker containers distributed to stage & prod
2. Update development docs to include a step to `npm run build:dev` in tandem with `npm start` (or else change `start` script to `run-p watch:all server.js`

groovecoder:
https://github.com/mozilla/blurts-server/pull/435/commits/747438c3a2d3195d3efdce741975a6b3a8f9c546 removes the `public/dist/*` resources from the repo, and updates `package.json` and `Dockerfile` to build them for all necessary environments.

The real test for `package.json` is someone else's workstation & heroku; the real test for `Dockerfile` is the stage server.

PR #438, opened 2018-09-21
for #55: convert console to mozlog

groovecoder:


groovecoder:
Could be merged. Though I'm not happy that mozlog squishes error stacktraces to single lines with `\n` characters, even with `fmt: "pretty"`

hibp.js
```diff
@@ -29,14 +31,14 @@ const HIBP = {
...
+      log.error("_throttledGot", {err: err});
       if (err.statusCode === 404) {
         // 404 can mean "no results", return undefined response; sorry calling code
         return response;
       } else if (err.statusCode === 429) {
-        console.log("got a 429, tryCount: ", tryCount);
+        log.info("_throttledGot", {err: "got a 429, tryCount: " + tryCount});
         if (tryCount >= AppConstants.HIBP_THROTTLE_MAX_TRIES) {
-          console.error(err.message);
+          log.error("_throttledGot", {err: err});
```
pdehaan: I'm not 100% familiar w/ mozlog internals, but I'm noticing that sometimes it looks like we're logging an object vs possibly nested error objects (if that matters, and maybe it doesn't):

Above (L39) we do `log.info("_throttledGot", {err: "some string"})`, but here, we're logging an Error object instance, which may be something closer to;

```js
log.error("_throttledGot", {err: {message: "some error message", "stack": "blah"}});
```

So in one branch its `err:string` and in other branches it's `err:object`.
Not saying it's wrong or bad, just invoking "see-something-say-something".

hibp.js
```diff
@@ -80,7 +82,7 @@ const HIBP = {
     } catch (error) {
       throw new Error("Could not load breaches: " + error);
     }
-    console.log("Done loading breaches.");
+    log.info("Done loading breaches");
```
pdehaan: This may be the first instance of a sentence-like first param, where everything else was "loadBreachesIntoApp", "smtpUrl-empty", "removed-subscriber-not-found", "unhandled-eventType", etc.

server.js
```diff
@@ -40,7 +42,7 @@ if (app.get("env") !== "dev") {
   try {
     await HIBP.loadBreachesIntoApp(app);
   } catch (error) {
-    console.error(error);
+    log.error(error);
```
pdehaan: Same as earlier comment... "Above we did something like log.error("generating", {err: err});"

Not sure if we need a "subject" here.

server.js
```diff
@@ -114,8 +116,8 @@ app.use(errorHandler);
 
 EmailUtils.init().then(() => {
   const listener = app.listen(AppConstants.PORT, () => {
-    console.info(`Listening on ${listener.address().port}`);
+    log.info("Listening", { port: listener.address().port });
   });
 }).catch(error => {
-  console.error(error);
+  log.error(error);
```
pdehaan: Ditto... 

> Above we did something like log.error("generating", {err: err});

pdehaan:
Oh, and a quick <kbd>$ git grep "console\."</kbd> shows that we may still have a few lingering `console.*()` statements in 

```sh
$ git grep -n --heading "console\."

public/js/monitor.js
213:  .catch(error => console.error(error));
231:    .catch(error => console.error(error));
255:  .catch(error => console.error(error));

scripts/make-sha1-hashes.js
8:console.log(PROMPT);
13:  console.log(`You entered: [${trimmedString}], sha1 hash of lowercase: ${sha1}`);
14:  console.log(PROMPT);
```

I don't really care about the ones in "scripts/*.js" since those are CLI tools, but the 3x instances in ./public/js/monitor.js may be candidates for mozlog greatness possibly. 🤷‍♂️

groovecoder:
Addressed all comments, except leaving `console` in the client-side `public/` JS code; `mozlog` is intended to help index our logging for server-side logging infrastructure.

PR #442, opened 2018-09-22
Restores Quantum logo in all the places.

### LINKED ISSUE #441: Restore Firefix Quantum logo & icon

lesleyjanenorton:
Fixes #441


pdehaan:
A bit odd that the fav icons aren’t perfectly square, but looking at previous commits it doesn’t look like they ever were.  But this looks great and easily revertable when we’re ready to switch to new icons.
👍

PR #445, opened 2018-09-24
fix #444: check bearer token on /hibp/notify

### LINKED ISSUE #444: Verify HIBP auth token on /hibp/notify
When we lift the restriction on `/hibp/notify` requests after launch, we need to verify that `/hibp/notify` requests are authentically sent from HIBP.

Add code to check `/hibp/notify` requests for `Authorization: Bearer <token>` header with valid token.

groovecoder:


tests/controllers/hibp.test.js
```diff
@@ -9,6 +9,19 @@ const { testBreaches } = require("../test-breaches");
...
+  const testEmail = "victim@spoofattack.com";
+  const testHash = sha1(testEmail);
+  const testPrefix = testHash.slice(0, 6).toUpperCase();
+  const testSuffix = testHash.slice(6).toUpperCase();
+
+  const mockRequest = { body: { hashPrefix: testPrefix, hashSuffixes: [testSuffix], breachName: "SomeSensitiveBreach" } };
+  const mockResponse = { status: jest.fn(), json: jest.fn() };
+
+  await expect(hibp.notify(mockRequest, mockResponse)).rejects.toThrow("HIBP notify endpoint requires valid authorization token.");
+});
```
pdehaan: Do we need another test for an invalid token throwing an error?

groovecoder: Good call. Added.

tests/controllers/hibp.test.js
```diff
@@ -47,7 +73,7 @@ test("notify POST with unknown breach should throw error", async () => {
   const testPrefix = testHash.slice(0, 6).toUpperCase();
   const testSuffix = testHash.slice(6).toUpperCase();
 
-  const mockRequest = { body: { hashPrefix: testPrefix, hashSuffixes: [testSuffix], breachName: "Test" }, app: { locals: { breaches: [] } } };
+  const mockRequest = { token: "unsafe-default-token-for-dev", body: { hashPrefix: testPrefix, hashSuffixes: [testSuffix], breachName: "Test" }, app: { locals: { breaches: [] } } };
```
pdehaan: not sure if these two rascals should be `AppConstants.HIBP_NOTIFY_TOKEN`, but probably doesn't matter since they're hardcoded for dev/testing.

tests/controllers/hibp.test.js
```diff
@@ -9,6 +9,32 @@ const { testBreaches } = require("../test-breaches");
...
+
+  await expect(hibp.notify(mockRequest, mockResponse)).rejects.toThrow("HIBP notify endpoint requires valid authorization token.");
+});
+
+
+test("notify POST with invalid token should throw error", async() => {
+  const testEmail = "victim@spoofattack.com";
+  const testHash = sha1(testEmail);
+  const testPrefix = testHash.slice(0, 6).toUpperCase();
+  const testSuffix = testHash.slice(6).toUpperCase();
```
pdehaan: I keep thinking we should create some wrapper function that takes an email address and returns an object w/ the `testHash`, `testPrefix`, and `testSuffix` since I think we're using it 3-4x in this file. But it's more of a small nit since I think this is the only file that separates out the prefix/suffix.

```js
function doStuff(testEmail) {
  const testHash = sha1(testEmail);
  return {
	testHash,
    testPrefix: testHash.slice(0, 6).toUpperCase(),
    testSuffix: testHash.slice(6).toUpperCase();
  };
}

// ...

const testEmail = "victim@spoofattack.com";
const {testHash, testPrefix, testSuffix} = doStuff(testEmail);
```

controllers/hibp.js
```diff
@@ -13,6 +13,10 @@ const log = mozlog("controllers.hibp");
 
 
 async function notify (req, res) {
+  if (!req.token || req.token !== AppConstants.HIBP_NOTIFY_TOKEN) {
+    const errorMessage = "HIBP notify endpoint requires valid authorization token.";
```
pdehaan: Another small, useless nit... but not sure if we should export this value, since we seem to use it 2-3x below in ./tests/controllers/hibp.test.js

Issue #264, opened 2018-08-20
Serve images (and other statics) straight from nginx; ideally with CDN


groovecoder:
Most likely code.cdn.mozilla.net. We will need to coordinate with Ops (@Micheletto?) to set up the CDN mirroring.

Consider creating a new helper in `hbs-helpers.js` ...

```js
hbs.registerHelper("staticDomain", function() {
  if (AppConstants.NODE_ENV === "production") {
    return AppConstants.CDN_DOMAIN;
  } else {
    return AppConstants.SERVER_URL;
  }
}
```

Then it can be used in the templates:

```html
<script src="{{ staticDomain }}/js/monitor.js"></script>
<link rel="stylesheet" src="{{ staticDomain }}/css/app.css">
<img src="{{ staticDomain }}/img/..." />
```

pdehaan:
And we should also use the CDN for stage verification (so we dont verify the site in Stage but then fails once it goes live to the mass public).

I didn't see any hardcoded image URLs in the CSS file.

pdehaan:
Actually, how does this work?
How would we separate it so that when we do a stage/production deploy, the local assets aren't on the server (so that we aren't deploying a bunch of unused assets)

Not sure if production builds should just not deploy the [/public/ folder](https://github.com/mozilla/blurts-server/tree/master/public), but maybe we don't want to host the favicons and social media icons on a CDN vs locally.

groovecoder:
When clients request assets from code.cdn.mozilla.net, the CDN servers will respond directly if they have the assets in their cache, or will pass the request thru to our origin server if they don't.

So, our stage & production environments should continue to serve the assets as the origin server to the CDN.

groovecoder:
Updated title ... we at least want to serve the static assets directly from nginx without hitting the node.js process. Ideally we will also add CDN in front of nginx, but the nginx part is the most important.

groovecoder:
I think @Micheletto has finished this as of Friday?

groovecoder:
The images are served via nginx now. Moving this to launch week milestone for the CDN part, if we can make it happen this week that will be great.

groovecoder:
Images are now being served properly with `Cache-Control` header that allows `ETag/If-None-Match` negotiation for `304` responses.

Moving out of launch week fix-ups.

EMMLynch:
Closing since we've redesigned the site and functionality since this was created. If you feel that this is still needed, please let me know.

PR #477, opened 2018-09-27
for #264: set Cache-Control on statics

groovecoder:
This should let nginx start serving `304` responses for static assets on stage & prod.

### DIFF
--- a/server.js
+++ b/server.js
@@ -66,7 +66,10 @@ app.use(helmet.contentSecurityPolicy({
   },
 }));
 app.use(helmet.referrerPolicy({ policy: "strict-origin-when-cross-origin" }));
-app.use(express.static("public"));
+app.use(express.static("public", {
+  setHeaders: res => res.set("Cache-Control",
+    "public, maxage=" + 10 * 60 * 1000 + ", s-maxage=" + 24 * 60 * 60 * 1000),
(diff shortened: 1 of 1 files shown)


server.js
```diff
@@ -66,7 +66,10 @@ app.use(helmet.contentSecurityPolicy({
   },
 }));
 app.use(helmet.referrerPolicy({ policy: "strict-origin-when-cross-origin" }));
-app.use(express.static("public"));
+app.use(express.static("public", {
+  setHeaders: res => res.set("Cache-Control",
+    "public, maxage=" + 10 * 60 * 1000 + ", s-maxage=" + 24 * 60 * 60 * 1000),
```
pdehaan: maxage is... 10 minutes and s-maxage is 24 hours?

Also, string above says "s-maxage" whereas comment below says "x-maxage". Not sure which is possible typo.

groovecoder: Updated comment below.

`maxage` > 0 is needed so nginx will actually use `ETag|Last-Modified` and `If-None-Match|If-Modified-Since` values for cache negotiation. But I'm worried that a long client-side cache age could make assets stick, so I'm starting with a lower client-side cache max age. (Pretty much enough to keep the assets cached while a user is performing a few scans and signing up.)

`s-maxage` will make the CDN and/or nginx refresh the statics from upstream at least every 24 hours.

PR #483, opened 2018-09-28
New error pages

lesleyjanenorton:
Adds new static HTML 429 error page and restyles all error pages to match spec.

### DIFF
--- a/views/error-429.html
+++ b/views/error-429.html
@@ -1,5 +1,5 @@
-
 <!DOCTYPE html>
+<!-- Custom static 429 page. See https://bugzilla.mozilla.org/show_bug.cgi?id=1495088 -->
 <html lang="en">
   <head>
     <meta charset="utf-8" />
--- a/views/error-429.html
+++ b/views/error-429.html
@@ -1,0 +1,95 @@
+
+<!DOCTYPE html>
+<html lang="en">
+  <head>
+    <meta charset="utf-8" />
+    <link rel="stylesheet" href="https://monitor.firefox.com/dist/app.min.css">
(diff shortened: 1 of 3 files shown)


views/error-429.html
```diff
@@ -0,0 +1,95 @@
+
+<!DOCTYPE html>
```
groovecoder: Can we add an HTML comment in here to explain why we have this file that is otherwise inaccessible via node? Something like:
```
<!-- Custom static 429 page. See https://bugzilla.mozilla.org/show_bug.cgi?id=1495088 --/>

File: views/error.hbs [hunk 2]
```diff
@@ -1,5 +1,7 @@
 {{!< default }}
 <div id="error-page-content" class="error-page-content section-wrapper">
-    <h2>Error</h2>
-    <p id="error-message"> {{ message }} </p>
+  <div class="half">
+    <h2 class="section-headline">Error</h2>
+    <p id="error-message"> Page not found. </p>
```
groovecoder: This should still be `{{ message }}`.

views/error-429.html
pdehaan: Is it possible to reuse our handlebars templates here and just compile the output to avoid all this duplication and hard coding?
It feels unlikely that we’ll be able to keep this file in sync if it’s hard coded html.

groovecoder: Maybe, but I'm not sure it's worth the effort - especially right now. File a follow-up issue?

pdehaan: Submitted #491 for tracking+discussion.

PR #484, opened 2018-09-28
Drop postcss-css-variables

lesleyjanenorton:
Don't need 'em! Transforming the css variables into static values was causing layout issues (particularly on mobile) and not delivering any performance gain.

### DIFF
--- a/package.json
+++ b/package.json
@@ -48,8 +48,6 @@
     "npm-run-all": "^4.1.3",
     "onchange": "^4.1.0",
     "postcss-cli": "^6.0.0",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -48,8 +48,6 @@
     "npm-run-all": "^4.1.3",
     "onchange": "^4.1.0",
     "postcss-cli": "^6.0.0",
-    "postcss-css-variables": "^0.9.0",
```
groovecoder: Removes 61 total dependencies! Nice!

PR #486, opened 2018-09-29
WIP: start l10n tooling

groovecoder:


public/locales/en/app.ftl
```diff
@@ -0,0 +1,7 @@
+support = Support
+
+feedback = Give Feedback
+
+terms = Terms&nbsp;and&nbsp;Privacy
+
+copyright = Portions of this content are © 1998-2018 by individual mozilla.org contributors. Content available under a Creative Commons license
```
pdehaan: Actual markup below is:

```html
<p data-l10n-id="copyright">
  Portions of this content are &copy; 1998-2018 by individual mozilla.org contributors.
  <br />
  Content available under a <a {{> analytics/link_event eventLabel="Creative Commons"}} href="https://www.mozilla.org/foundation/licensing/website-content/" target="_blank" rel="noopener">Creative Commons license</a>.
</p>
```

Not sure if the .ftl files should use &amp;copy; entity or if we need anything special for link preserving.

public/locales/de/app.ftl
```diff
@@ -0,0 +1,7 @@
+support = Unterstützung
+
+feedback = Feedback geben
+
+terms = Allgemeine&nbsp;Geschäftsbedingungen&nbsp;und&nbsp;Datenschutzerklärung
+
+copyright = Teile dieses Inhalts sind © 1998-2018 von einzelnen mozilla.org-Mitwirkenden. Inhalt verfügbar unter einer Creative Commons-Lizenz
```
pdehaan: /cc @flodolo @stasm

Do either of you have any guidance or suggestions on bundling the .ftl files directly in the source repo, versus doing something like a separate "blurts-server-l10n" repo for translations?

flodolo: Reasons to use a different repository:
* History: Pontoon will write in the repository every 20 minutes, once for each locale (if there are changes). History is going to become messy very quickly.
* Decouplement: You can develop on master and decide when strings are ready for translation.

Cons:
* You need scripts to import strings back.
* You need to expose strings in the l10n repository via PR.

Alternatively, we can also write in a branch instead of master, but then you need to deal with merges back and forth.

Side note: we'll need a dev server updated frequently for l10n testing. For that it might be easier to have the strings in the main repository.



From the point of view of configuration, if you want Pontoon to write here we just need to add its bot.

.env-dist
```diff
@@ -4,6 +4,9 @@ SERVER_URL=http://localhost:6060
 PORT=6060
 COOKIE_SECRET=3895d33b5f9730f5eb2a2067fe0a690e298f55f5e382c032fd3656863412
 
+# available languages for l20n
+AVAILABLE_LANGUAGES="de,en,es,fr,it,ja,pt,ru,zh"
```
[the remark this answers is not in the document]
groovecoder: We should thoroughly test, but the `accept-language` module says that `get` will "Get the most likely language given an Accept-Language string".

middleware.js
```diff
@@ -1,9 +1,21 @@
...
+const AppConstants = require("./app-constants");
 const mozlog = require("./log");
 
 
 const log = mozlog("middleware");
+
+
+function pickLanguage (req, res, next) {
+  acceptLanguage.languages(AppConstants.AVAILABLE_LANGUAGES.split(","));
+  req.app.locals.locale = acceptLanguage.get(req.headers["accept-language"]);
```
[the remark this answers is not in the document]
groovecoder: Code looks like it will default to whatever is the first language tag specified in `acceptLanguage.languages`. So in our case - yes.

public/locales/de/app.ftl
groovecoder: @flodolo can you point us to examples of each kind of setup? I.e., same-repo vs. separate-repo?

stasm:
Thanks for the CC and for kicking this off.

`l20n.js` hasn't been actively developed in more than a year. We shifted focus completely to Fluent. I should really get to deprecating it officially (https://github.com/l20n/spec/issues/17).

Instead, we have developed [`fluent`](https://github.com/projectfluent/fluent.js/tree/master/fluent) (for parsing FTL files and formatting translations) and [`fluent-dom`](https://github.com/projectfluent/fluent.js/tree/master/fluent-dom) (for localizing the DOM), although they might require some additional work to set up. `fluent-dom` doesn't initialize the localization of the document automatically during the load of the page. Some glue code is needed to parse the links to FTL files form `<head>` and start the localization of the document.

There's an experimental packaged called [`fluent-web`](https://github.com/projectfluent/fluent-web) which provides this glue code together with the dependencies. We haven't had a strong driver to work on it recently. @zbraniecki will know the latest status of it.

For language negotiation there's also [`fluent-langneg`](https://github.com/projectfluent/fluent.js/tree/master/fluent-langneg) which is well-tested and used by a number of other Mozilla web projects. It might make sense to use it here, too, although other packages available on `npm` should be OK as well.

public/locales/de/app.ftl
pdehaan: The one separate-repo I can remember off the top of my head is FxA's content server:

https://github.com/mozilla/fxa-content-server
vs
https://github.com/mozilla/fxa-content-server-l10n

I think the trick was putting a package.json in the [*-l10n repo](https://github.com/mozilla/fxa-content-server-l10n/blob/master/package.json) so you can include it as a dependency in your main repo's package.json. But of course, as I went to prove that, I notice that FxA team is using some Bash script[1] to download locales, so we may want to sync up with them to see why they switched to doing it that way (or maybe I'm misremembering how it was done originally).

[1] https://github.com/mozilla/fxa-content-server/blob/master/scripts/download_l10n.sh

groovecoder:
I looked thru send, screenshots, testpilot, and fluent-web and I prefer fluent-web by a large margin. I tried to simply replace `l20n.js` with `fluent-web.js` but it looks like `fluent-web` has some making & building steps that I can't figure out.

How long would it take to get `fluent-web` up to a usable state like `l20n.js`? Can it be done by the end of this week?

If not, I'll start digging into using `fluent-dom` directly.

zbraniecki:
> but it looks like fluent-web has some making & building steps that I can't figure out.

```
hg clone https://github.com/projectfluent/fluent-web
cd fluent-web
npm i
make build
open ./examples/simple.html
```

does that work?

groovecoder:
I get `SyntaxError: import declarations may only appear at top level of a module fluent-web.js:2`

groovecoder:
I'm happy to move this over to a fluent-web issue.

groovecoder:
Still waiting to see if we can use `fluent-web` for the front-end l10n, but I made a significant change to this PR for back-end l10n ...

1. On server start-up:
   * scan `public/locales` directory for available languages
   * instantiate a `FluentBundle` for each language
   * keep both in `app.locals`
   * `app.use(pickLanguage)` middleware
2. During request:
   * `pickLanguage` checks `Accept-Language` and sets appropriate `req.locale` and `req.fluentBundle`
3. In back-end code, e.g., `controllers/home.js`, use `req.fluentBundle.format(req.fluentBundle.getMessage("notFound")` where it used to have the hard-coded English message.
   * (could probably condense this with a `locale.bundleMessage()` helper function

This seems to work fine!

screen shot 2018-10-03 at 4 34 46 pm

And it feels simpler than some of the other approaches I saw in send, testpilot, screenshots, etc.

Does it seem sane?

pdehaan:
One further wrench into the plan is that the presence of a directory/file isn't always an indicator of translation %%.

In other repos (like [**mozilla/send**](https://github.com/mozilla/send/blob/master/scripts/get-prod-locales.js)) we had a script which would only include locales that are 100% translated. I think [**mozilla/fxa-content-server**](https://github.com/mozilla/fxa-content-server/blob/d95118004500ecd522061b4de08ace068e21ae37/grunttasks/l10n-supported-locales.js) had a lower threshold (90%) before it would be considered ready for production.

views/layouts/default.hbs
```diff
@@ -44,9 +52,9 @@
...
-          <li>Support</li>
-          <li><a {{> analytics/link_event eventLabel="Give Feedback"}} href="https://qsurvey.mozilla.com/s3/Firefox-Monitor-Feedback" target="_blank" rel="noopener">Give Feedback</a></li>
-          <li><a {{> analytics/link_event eventLabel="Terms & Privacy"}} href="https://www.mozilla.org/privacy/firefox-monitor/{{> analytics/utm_site_footer }}" target="_blank" rel="noopener">Terms&nbsp;and&nbsp;Privacy</a></li>
+          <li data-l10n-id="support">Support</li>
+          <li><a data-l10n-id="feedback" {{> analytics/link_event eventLabel="Give Feedback"}} href="https://qsurvey.mozilla.com/s3/Firefox-Monitor-Feedback" target="_blank" rel="noopener">Give Feedback</a></li>
+          <li><a data-l10n-id="terms" {{> analytics/link_event eventLabel="Terms & Privacy"}} href="https://www.mozilla.org/privacy/firefox-monitor/{{> analytics/utm_site_footer }}" target="_blank" rel="noopener">Terms&nbsp;and&nbsp;Privacy</a></li>
           <li><a {{> analytics/link_event eventLabel="GitHub"}} href="https://github.com/mozilla/blurts-server/" target="_blank" rel="noopener noreferrer"><img src="/img/github.svg" alt="GitHub Logo" /></a></li>
```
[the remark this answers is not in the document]
groovecoder: They are! Good catch!

groovecoder:
Addressed all comments except the template-level ones. (Still checking if we'll be able to use `fluent-web` for template strings, or if we'll make a handlebars helper for them.)

Consolidated some of the server start-up code into a single `locale.loadLanguagesIntoApp` function.

Made the `pickLanguage` middleware also add a `fluentFormat` helper on the request object, so calling code can be simplified to, e.g., `req.fluentFormat("notFound")`

pdehaan:
> The %-checking heuristic seems neat, but how helpful is it?

I think it's helpful, but you'd probably need to sync up with the FxA team or Send/dcoates and see how well it's working for them after all this time.

> The fxa-content-server check just seems to check that a translated message is non-blank, which ignores up-to-date-ness of the translations; the send check for 100% no errors nor missing translations means (like you say in the PR) that any update to en-US invalidates all translations until they are back to 100%?

IIRC, I think Send stores all it's production locale list in package.json and it's manually managed.

https://github.com/mozilla/send/blob/a2eee15a7de297092575f28da13a0c330e64058e/package.json#L134-L185

The [script](https://github.com/mozilla/send/blob/master/scripts/get-prod-locales.js) just runs a scanner to say which locales are 100% or whatever it does, and you can either manually update package.json if you like the suggestions of adding the new translations, or there may have been some `--write` flag mischief that would auto-write the package.json changes for you.

> Either tooling tactic seems to rely on a significant amount of manual translation support anyway. Does the tooling help notify the human translators of the updates needed?

I don't think the tooling does anything (at least in the case of Send). I _assume_ it's just something that gets run before a deploy to see if there are any new locales that have achieved 90-100% translations and would be good candidates to include in the website.

I imagine we could write some unit test that gets triggered on CI that could do something in a more automated way if you think it'd be useful to be notified regularly of 100% translations. But I imagine that'd just be a bunch of failing PRs due to unrelated strings. Plus, I don't know how actionable updating `app.locals.AVAILABLE_LANGUAGES` is versus just doing it prior to a stage/prod deploy.

locale-utils.js
```diff
@@ -0,0 +1,69 @@
...
+
+    this.fluentID = fluentID;
+  }
+}
+
+
+const LocaleUtils = {
+  init() {
+    const languageDirectories = fs.readdirSync( localesDir ).filter(item => {
+      return (item[0] !== "." && item.length === 2)
```
pdehaan: What does `item.length === 2` do?

It seems like it's checking if the file/folder name is 2 characters long.

middleware.js
```diff
@@ -29,14 +64,18 @@ function clientErrorHandler (err, req, res, next) {
 
 
 function errorHandler (err, req, res, next) {
+  console.log("errorHandler, err.message: ", err.message);
```
pdehaan: `console.log()` or should we use **mozlog** here?

groovecoder: Good catch. We shouldn't actually need this `console.log` since `log.error` in `logErrors` above should take care of the logging.

public/locales/en/app.ftl
```diff
@@ -0,0 +1,35 @@
...
+oauth-confirmed-title = {-product-name} : Subscribed
+
+scan-title = {-product-name} : Scan Results
+
+user-add-invalid-email = Invalid Email
+user-add-email-verify-subject = Verify your subscription to {-product-name}.
+user-add-title = {-product-name} : Confirm Email
+
+user-verify-token-error = Verification token is required.
+user-verify-email-report-subject = Your Firefox Monitor report
```
pdehaan: nit: instead of "Firefox Monitor" we should probably use `{-product-name}` here.

public/locales/en/app.ftl
```diff
@@ -0,0 +1,35 @@
+-product-name = Firefox Monitor
+
+layout-support = Support
+layout-feedback = Give Feedback
+layout-terms = Terms&nbsp;and&nbsp;Privacy
+layout-copyright = Portions of this content are &copy; 1998-2018 by individual mozilla.org contributors. <br />Content available under a <a {{> analytics/link_event eventLabel="Creative Commons"}} href="https://www.mozilla.org/foundation/licensing/website-content/" target="_blank" rel="noopener">Creative Commons license</a>.
```
pdehaan: Not sure how well fluent supports nested HTML, but I'm pretty sure it isn't great from a localizer standpoint since it can get pretty complex (for example, localizers will have to ignore the handlebars partial and that `eventLabel` which shouldn't be localized). I think l20n had ways to avoid this, but i don't know about fluent.

public/locales/en/app.ftl
```diff
@@ -0,0 +1,35 @@
...
+
+error-not-subscribed = This email address is not subscribed to {-product-name}.
+error-hibp-throttled = Too many connections to HIBP.
+error-hibp-connect = Error connecting to HIBP.
```
pdehaan: Not sure how visible these errors are, or if we should use a more verbose "Have I Been Pwned" vs the abbrev.

public/locales/en/app.ftl
zbraniecki: @flodolo ^ ?

public/locales/en/app.ftl
flodolo: In any case, HIBP needs an explanation in a comment.

public/locales/en/app.ftl
```diff
@@ -0,0 +1,35 @@
...
+error-hibp-load-breaches = Could not load breaches.
+
+hibp-notify-email-subject = {-product-name} Alert : Your account was involved in a breach.
```
flodolo: Spaces before a colon don't seem correct in English (also other strings)?

public/locales/en/app.ftl
```diff
@@ -0,0 +1,35 @@
...
+
+user-add-invalid-email = Invalid Email
+user-add-email-verify-subject = Verify your subscription to {-product-name}.
+user-add-title = {-product-name} : Confirm Email
+
+user-verify-token-error = Verification token is required.
+user-verify-email-report-subject = Your Firefox Monitor report
+user-verify-title = {-product-name} : Subscribed
+
+user-unsubscribe-token-error = Unsubscribe requires a token.
```
flodolo: Unsubscribing?

groovecoder:
@stasm @flodolo @pdehaan I'd like to merge & deploy this with English-only at first, just to avoid the bit-rot of a PR that touches almost every file. 😉 Does that seem sane?

flodolo:
> @stasm @flodolo @pdehaan I'd like to merge & deploy this with English-only at first, just to avoid the bit-rot of a PR that touches almost every file. 😉 Does that seem sane?

Works for me. We have space to fix the en-US until we start exposing for localization in Pontoon.

Where is German coming from btw?

groovecoder:
Oh, I was just dumping some of the strings thru Google Translate to make sure the messages were showing up correctly in all the different contexts. (e.g., English in error logs but German on the user-facing error page)

groovecoder:
Also fixed tests!

groovecoder:
@lesleyjanenorton is going to work on the handlebars templates tomorrow. Then we'll remove the temporary `de/app.ftl` file we used for testing, and try to merge & deploy this.

public/locales/en/app.ftl
```diff
@@ -0,0 +1,35 @@
+-product-name = Firefox Monitor
+
+layout-support = Support
+layout-feedback = Give Feedback
+layout-terms = Terms&nbsp;and&nbsp;Privacy
```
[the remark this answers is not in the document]
pdehaan: Nit: we should probably see if we can do this using something like [`white-space`](https://css-tricks.com/almanac/properties/w/whitespace/) in CSS.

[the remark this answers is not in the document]
stasm: Something like the following should do the trick:

```properties
layout-terms = Terms{"\u00A0"}and{"\u00A0"}Privacy
```

hbs-helpers.js
```diff
@@ -16,7 +23,7 @@ function breachDataClasses(dataClasses) {
 
 function prettyDate(date) {
   const jsDate = new Date(date);
-  // TODO: localize this
+  // TODO: l10n
   return jsDate.toLocaleDateString("en-US", {year: "numeric", month: "long", day: "numeric"});
```
stasm: Fluent can format dates, too. You can pass the date object as an argument to `format`. I assume you'd need to make `prettyDate` take `req` as an argument, however. I'm not sure if that's possible given where it's used.

```js
function prettyDate(req, date) {
  return fluentFormat(req, "pretty-date", {date: new Date(date)});
}
```
```properties
# en-US
pretty-date = {DATETIME($date, year: "numeric", month: "long", day: "numeric")}
```
```properties
# Some other language might decide long month names are too long.
pretty-date = {DATETIME($date, year: "numeric", month: "short", day: "numeric")}
```

There's also a not-very-well-documented feature of setting the default formatting options for the arguments.

```js
import {FluentDateTime} from "fluent";

function prettyDate(req, date) {
  return fluentFormat(req, "pretty-date", {
    date: new FluentDateTime(date, {
      year: "numeric",
      month: "long",
      day: "numeric",
    })
  });
}
```
```properties
# en-US
pretty-date = {$date}
```

Other locales can still customize the formatting options, if needed:

```properties
# Some other language might decide long month names are too long.
pretty-date = {DATETIME($date, year: "numeric", month: "short", day: "numeric")}
```

middleware.js
```diff
@@ -1,9 +1,35 @@
...
+
+
+// picks available language by Accept-Language and assigns to request
+function pickLanguage (req, res, next) {
+  acceptLanguage.languages(req.app.locals.AVAILABLE_LANGUAGES);
+  const pickedLanguage = acceptLanguage.get(req.headers["accept-language"]);
+  req.pickedLanguage = pickedLanguage;
+  const fluentBundle = req.app.locals.FLUENT_BUNDLES[pickedLanguage];
+  req.fluentFormat = (message, args = null, errors = null) => {
+    return fluentBundle.format(fluentBundle.getMessage(message, args));
```
stasm: This code will work fine in the optimistic scenario of most localizations being complete :) To better support the more realistic scenario of some localizations missing some translations, I'd like to suggest basing this code on the concept of _languages as a list_.

The idea is to negotiate the _list of languages_ requested by the user against the list of available languages. The output is another list. The default language should always be the last item of the resulting list. You can use the `negotiateLanguages` function from the `fluent-langneg` package to perform this negotation reliably.

Unfortunately, AFAICT, the `accept-language` package doesn't expose a method which returns a list of languages parsed from the header. `fluent-langneg` provides a very simple implementation which should be enough for this use-case.

Here's how I'd use it:

```js
import {negotiateLanguages, acceptedLanguages} from "fluent-langneg";

function pickLanguage (req, res, next) {
  // Parse the header into a list of language tags.
  const requested = acceptedLanguages(req.headers["accept-language"]);
  // Negotiate the list of languages for the current user, in order of their preference.
  const supportedLocales = negotiateLanguages(
    requested,
    req.app.locals.AVAILABLE_LANGUAGES
    {defaultLocale: "en-US"}
  );

  req.fluentFormat = (id, args = null, errors = null) => {
    for (let locale of supportedLocales) {
      let bundle = req.app.locals.FLUENT_BUNDLES[locale];
      if (bundle.hasMessage(id)) {
        const message = bundle.getMessage(id);
        return bundle.format(message, args);
      }
    }
    // If no bundle has this message (even the default),
    // fall back to the identifier.
    return id;
  };

  next();
}
```

stasm: If you don't go this route (although I think you should), note that there's a small bug in the current code. The `args` object should be passed as the second argument to `format` rather than `getMessage`. After all, a single message might be used to produce many different translations in the UI, each for different arguments.

```diff
-return fluentBundle.format(fluentBundle.getMessage(message, args));
+return fluentBundle.format(fluentBundle.getMessage(message), args);
```

locale-utils.js
```diff
@@ -0,0 +1,65 @@
...
+
+class FluentError extends Error {
+  constructor(fluentID = null, ...params) {
+    super(...params);
+
+    if (Error.captureStackTrace) {
+      Error.captureStackTrace(this, FluentError);
+    }
+
+    this.fluentID = fluentID;
```
stasm: 👍 This is a very good pattern, thanks for doing it this way!

pdehaan:
Once we merge this, it'd also be nice to set up a **CODEOWNERS** file and then make @flodolo a required reviewer of any `public/locales/en/*.ftl` string changes (unless he doesn't want to, in which case, don't).

A good reference is how we did it in [**mozilla-services/screenshots** CODEOWNERS](https://github.com/mozilla-services/screenshots/blob/master/CODEOWNERS).

pdehaan:
Re: 
> _"is there tooling around `CODEOWNERS` files? I.e., can we enforce that **flodolo** has to approve PRs that contain changes to `public/locales/en/app.ftl`?"_

https://help.github.com/articles/about-codeowners/ says:

> Code owners are automatically requested for review when someone opens a pull request that modifies code that they own. When someone with admin or owner permissions has enabled required reviews, they also can optionally require approval from a code owner before the author can merge a pull request in the repository.

groovecoder:
@pdehaan brilliant! I love it!

groovecoder:
I brought in https://github.com/mozilla/blurts-server/pull/486/commits/cca431b90b8c21b4a5323c189169279033f3c636 by @lesleyjanenorton which includes wrapping the handlebars/HTML strings with `fluentFormat`. So I think we're at the full i18n level now?

@stasm @flodolo I think @lesleyjanenorton had a question for how to pass args/variables into fluent strings.

hbs-helpers.js
```diff
@@ -1,23 +1,51 @@
...
+function fluentFormat (req, id, args) {
+  return LocaleUtils.fluentFormat(req.supportedLocales, id, args);
+}
+
+function getValue (req, key, number, args) {
+  const fluentID = `${key}${number}`;
+  return LocaleUtils.fluentFormat(req.supportedLocales, fluentID, args);
+}
+
+function breachResponse(req, id, breachTitle, breachCount) {
```
lesleyjanenorton: @stasm @flodolo Trying unsuccessfully to pass breachTitle and breachCount to the app.ftl. Any thoughts on what is going wrong here?

lesleyjanenorton: This is what comes back: screen shot 2018-10-11 at 3 56 43 pm

flodolo:
Note: you might need to add me as a contributor to the repository for CODEOWNERS to work

public/locales/en/app.ftl
```diff
@@ -0,0 +1,35 @@
+-product-name = Firefox Monitor
+
+layout-support = Support
```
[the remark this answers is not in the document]
flodolo: A little more, please? :) Localizers don't see (or read in most cases) code

```
# Support should be treated as a name, used in a link to ???
```

public/locales/en/app.ftl
```diff
@@ -0,0 +1,304 @@
...
+user-unsubscribe-token-email-error = Unsubscribe requires a token and emailHash.
+user-unsubscribe-title = {-product-name} : Unsubscribe
+user-unsubscribe-survey-title = {-product-name} : Unsubscribe Survey
+user-unsubscribed-title {-product-name} : Unsubscribed
+
+## Password Tips
+
+pwt-section-headline = Passwords = Protection
+pwt-section-subhead = Your Private Information Is Only as Safe as Your Passwords
+pwt-section-blurb =  Your passwords protect more than your accounts, they protect every bit of personal information that resides in them. And hackers rely on bad habits, like using the same password everywhere or using common phrases (p@ssw0rd, anyone?), so that if they hack one account, they can hack many. Here are six ways to protect your accounts.
```
flodolo: Extra leading space

public/locales/en/app.ftl
```diff
@@ -0,0 +1,304 @@
...
+
+pwt-headline-1 = Use a Different Password<br />for Every Account
+pwt-headline-2 = Create Strong<br />Passwords
+pwt-headline-3 = Treat Security Questions<br>Like Extra Passwords
```
flodolo: Would be nice to be consistent: `<br>` vs `<br />`

public/locales/en/app.ftl
```diff
@@ -0,0 +1,304 @@
...
+wtd-1 = Change your password on these sites and anywhere else you've used the same password.
+wtd-2 = Make your answers to security questions just as strong as your passwords.
+wtd-3 = Use password managers like 1Password, LastPass, or Dashlane to generate strong passwords, remember them, and fill them into websites so you don't have to.
+wtd-4 = Subscribe to alerts from {-product-name} to learn sooner about your compromised accounts.
+
+
+featured-breach-only = Your email address appears in the <span class="featured-breach-title"> --- </span> breach, but does not appear in any other known breaches.
+
+featured-breach-results = Your email address appears in the { $featuredBreachTitle } breach, { $breachCount ->
+  [1] one other breach.
```
flodolo: Do we really need to spell out numbers up to ten? Do we have data on the average number of breaches?

flodolo:
One more note looking at the size of content from the last time I checked:
* We'll *need* a dev server, updated frequently, for localizers to test their translations.
* We'll need a system to keep partial localizations out of the production server, and only enable them after a certain percentage of completion.

public/locales/en/app.ftl
```diff
@@ -0,0 +1,304 @@
...
+user-unsubscribe-survey-title = {-product-name} : Unsubscribe Survey
+user-unsubscribed-title {-product-name} : Unsubscribed
+
+## Password Tips
+
+pwt-section-headline = Passwords = Protection
+pwt-section-subhead = Your Private Information Is Only as Safe as Your Passwords
+pwt-section-blurb =  Your passwords protect more than your accounts, they protect every bit of personal information that resides in them. And hackers rely on bad habits, like using the same password everywhere or using common phrases (p@ssw0rd, anyone?), so that if they hack one account, they can hack many. Here are six ways to protect your accounts.
+
+pwt-headline-1 = Use a Different Password<br />for Every Account
```
flodolo: This is going to backfire badly as soon as you need to change any of these strings, since you're generating IDs in the code later one. Please let's not do this.

```properties
 pwt-headline-strong-password = Use a Different Password<br />for Every Account 
 pwt-summary-strong-password = You can't prevent a data breach, but you can limit your exposure by always using different passwords for different websites. 
```

Then store the mapping somewhere

```
{
  1: 'strong-password',
  2: 'foo',
}
```

public/locales/en/app.ftl
flodolo: Side note: looks like there's a term for HIBP now, it should be used instead.

public/locales/en/app.ftl
flodolo: It [does](https://pontoon.mozilla.org/en-CA/firefox/browser/browser/preferences/preferences.ftl/?search=%5B1%5D&string=176248). But I'm almost sure the UI would be useful with this many variants, and they would be automatically exposed to all languages, instead of a standard plural form (we should verify it somehow).

public/locales/en/app.ftl
stasm: Even better, use the full identifier (`pwt-headline-strong-password`) for extra-easy `grep` queries :)

public/locales/en/app.ftl
```diff
@@ -0,0 +1,304 @@
...
+  [3] three other breaches.
+  [4] four other breaches.
+  [5] five other breaches.
+  [6] six other breaches.
+  [7] seven other breaches.
+  [8] eight other breaches.
+  [9] nine other breaches.
+  [10] ten other breaches.
+  [other] { $breachCount } other breaches.
+}
```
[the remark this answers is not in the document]
lesleyjanenorton: Success- thank you! Working exactly as expected now after adding the * and indentation.

public/locales/en/app.ftl
lesleyjanenorton: I'm thinking these line breaks need to go since there isn't a way to know exactly where the line breaks need to be placed in each translation.

public/locales/en/app.ftl
```diff
@@ -97,39 +100,35 @@ form-signup-checkbox = Get the latest from {-brand-Mozilla} and {-brand-name}.
...
-found-breaches-plural-blurb = Your email address was compromised in the following breaches.
+found-breaches-plural = Your email address appears in the following { $breachCount } breaches.
+found-breaches-singular = Your email address apppeared in the following breach.
+
+no-breaches-blurb = 
+  Your email address did not appear in our basic scan. 
+  That's good news, but data breaches can happen any time and there is still more you can do. 
+  Subscribe to Firefox Monitor for a full report, alerts when new breaches happen, and tips on protecting your passwords.
+
+featured-breach-results =
```
lesleyjanenorton: @stasm I'm still running into some selector trouble here in "featured-breach-results". As is, this throws a "ctor is not a constructor" error. If I change [one] to [1], the string will return correctly only if the value of $breachCount is exactly 1. I'm totally puzzled as I recreated this over in the Fluent playground where it appears to be working exactly as expected. Have you run into anything like this?

stasm: This might be related to the `Intl.PluralRules` API. It's still a fairly new feature in JavaScript. Looking at the [browser support matrix](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/PluralRules#Browser_compatibility), is there a chance you're testing on Safari? In order to support Safari as well as older browsers, you'll need to polyfill `Intl.PluralRules`. I recommend [`intl-pluralrules`](https://www.npmjs.com/package/intl-pluralrules).

Background: Fluent relies on this API to match variant keys which are not numbers. When you use `[one]` Fluent maps the numerical value of `$breachCount` to a plural category. Categories are [defined in CLDR](http://www.unicode.org/cldr/charts/29/supplemental/language_plural_rules.html). For English, they're `one` and `other`. Other languages have fewer or more.

groovecoder: Aha, node.js is probably the culprit here - it doesn't have support for `Intl.PluralRules` API. I'll see if I can add that to our `package.json` and get this to work.

locale-utils.js
```diff
@@ -0,0 +1,74 @@
+"use strict";
+
+const fs = require("fs");
+const path = require("path");
+
+// node.js needs Intl.PluralRules polyfill
+require("intl-pluralrules");
```
groovecoder: image

groovecoder: its-working-its-working

stasm: Nice! Happy to hear that was it!

groovecoder:
Updated to have `fluentFormat` add messages from all `.ftl` files in the in `public/locales/{lang}/` directories.

public/locales/en/app.ftl
```diff
@@ -0,0 +1,185 @@
...
+featured-breach-results = 
+  Your account appeared in the <span class="bold"> { $featuredBreach } </span> breach, 
+    { $breachCount ->
+        [0] but did not appear in any other known data breaches
+        [1] as well as { $breachCount } other breach
+       *[other] as well as { $breachCount } other breaches
+    }.
+
+scan-results = Your account { $breachCount ->
+    [0] Your email address did not appear in our basic scan.
```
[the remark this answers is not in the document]
lesleyjanenorton: @flodolo thanks for the feedback. Is it possible to translate any variant in Pontoon, or is there a specific issue with the zero case?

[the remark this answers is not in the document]
lesleyjanenorton: screen shot 2018-10-16 at 4 25 07 pm
Was following this example and a similar example of usage here: https://projectfluent.org/play/

public/locales/en/app.ftl
```diff
@@ -0,0 +1,185 @@
...
+-brand-Quantum = Firefox Quantum
+-brand-Mozilla = Mozilla
+-brand-HIBP = HIBP
+-product-name-nowrap = <span class="nowrap">Firefox Monitor</span>
+
+layout-Firefox = {-brand-name}
+# Descriptive headline for a column of links where users can give feedback, or get additional information about, Firefox Monitor.
+layout-support = Support
+# Link that takes the user to a Firefox Monitor survey. 
+give-feedback = <span class="nowrap">Give Feedback</span>
```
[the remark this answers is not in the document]
lesleyjanenorton: Yep, we can move this to template.

public/locales/en/app.ftl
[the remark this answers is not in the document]
Pike: This is perfectly fine Fluent, it's just tricky for localizers to understand what we're asking them to translate.

It becomes easier if we're exposing the full phrase for each variant.

There's things we'll want to do on the l10n tooling side there in the next year, but right now, the extra repetition of the head and tail section would be good.

```properties
single-featured-breach-result = Your account appeared in the <span class="bold"> { $featuredBreach } </span> breach, but did not appear in any other known data breaches.
multiple-featured-breach-results =
  {  $breachCount ->
        [one] Your account appeared in the <span class="bold"> { $featuredBreach } </span> breach, as well as { $breachCount } other breach.
       *[other] Your account appeared in the <span class="bold"> { $featuredBreach } </span> breach, as well as { $breachCount } other breaches.
  }
```

public/locales/en/app.ftl
```diff
@@ -0,0 +1,185 @@
...
+## Password Tips
+
+pwt-section-headline = Passwords = Protection
+pwt-section-subhead = Your Private Information Is Only as Safe as Your Passwords
+pwt-section-blurb =
+  Your passwords protect more than your accounts, they protect every bit of personal information that resides in them. 
+  And hackers rely on bad habits, like using the same password everywhere or using common phrases (p@ssw0rd, anyone?), 
+  so that if they hack one account, they can hack many. Here are six ways to protect your accounts.
+
+pwt-headline-1 = Use a Different Password for Every Account
```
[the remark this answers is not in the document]
lesleyjanenorton: I think in this case the numbers are meaningful because the password tips are specifically presented in the order that they are numbered.  Additionally, the tips are displayed alongside corresponding icons which are also numbered accordingly.  I will happily defer to your judgement here if you think this is going to cause future problems though.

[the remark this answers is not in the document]
lesleyjanenorton: Regarding the colon spacing on L13 (for some reason I can't respond to that comment directly): We just received updated copy for the emails and the extra space has been removed.

public/locales/en/app.ftl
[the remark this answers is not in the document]
lesleyjanenorton: I'll remove it wherever possible. Unfortunately, for cases like L82 and a few others, it will need to remain as is.

public/locales/en/app.ftl
```diff
@@ -0,0 +1,185 @@
+-product-name = Firefox Monitor
+-brand-name = Firefox
+-brand-Quantum = Firefox Quantum
+-brand-Mozilla = Mozilla
+-brand-HIBP = HIBP
+-product-name-nowrap = <span class="nowrap">Firefox Monitor</span>
```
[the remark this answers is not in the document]
lesleyjanenorton: Potentially - we've discussed this in previous pull requests and my understanding is that &nbsp; is deprecated and there are a11y concerns with using it.

public/locales/en/app.ftl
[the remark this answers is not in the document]
lesleyjanenorton: @Pike thanks for the clarification

public/locales/en/app.ftl
```diff
@@ -0,0 +1,185 @@
...
+# Descriptive headline for a column of links where users can give feedback, or get additional information about, Firefox Monitor.
+layout-support = Support
+# Link that takes the user to a Firefox Monitor survey. 
+give-feedback = <span class="nowrap">Give Feedback</span>
+terms-and-privacy = <span class="nowrap">Terms and Privacy</span>
+
+error-not-subscribed = This email address is not subscribed to {-product-name}.
+error-hibp-throttled = Too many connections to {-brand-HIBP}.
+error-hibp-connect = Error connecting to {-brand-HIBP}.
+error-hibp-load-breaches = Could not load breaches.
```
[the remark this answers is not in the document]
lesleyjanenorton: I think L16-L18 are actually a log errors and may not need to be localized at all.

public/locales/en/app.ftl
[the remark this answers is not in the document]
flodolo: What Pike said :)

The UI looks [like this](https://pontoon.mozilla.org/it/firefox/browser/browser/preferences/preferences.ftl/?search=%5B1%5D&string=176248). The fragment before would become another input field, and you would need to drag the input field to see everything.

public/locales/en/app.ftl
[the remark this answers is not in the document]
flodolo: > Additionally, the tips are displayed alongside corresponding icons which are also numbered accordingly.

How is that done? Have you considered other types of numerals?

>  I will happily defer to your judgement here if you think this is going to cause future problems though.

Two likely scenarios:
* You want to change the order. 
* You want to replace or update one of the tips. You'll need to [change the message ID](https://developer.mozilla.org/en-US/docs/Mozilla/Localization/Localization_content_best_practices#Changing_existing_strings) to do that.

In both cases you'll need to add a mapping afterward.

Also, strings are presented sequentially to localizers, so having title and description together help a bit.

public/locales/en/data-classes.ftl
```diff
@@ -0,0 +1,117 @@
...
+device-usage-tracking-data = Device usage tracking data
+drinking-habits = Drinking habits
+drug-habits = Drug habits
+eating-habits = Eating habits
+education-levels = Education levels
+email-addresses = Email addresses
+email-messages = Email messages
+employers = Employers
+ethnicities = Ethnicities
+family-members'-names = Family members' names
```
flodolo: I'm pretty sure the `'` in the ID is a parsing error. The string, on the other hand, should use an apostrophe `’`

public/locales/en/app.ftl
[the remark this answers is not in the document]
flodolo: To be clear, I would do this

```
featured-breach-no-results = 
  Your account appeared in the <span class="bold"> { $featuredBreach } </span> breach, 
  but did not appear in any other known data breaches.

featured-breach-results =
    { $breachCount ->
        [one] Your account appeared in the <span class="bold"> { $featuredBreach } </span> breach, as well as { $breachCount } other breach.
      *[other] Your account appeared in the <span class="bold"> { $featuredBreach } </span> breach, as well as { $breachCount } other breaches.
    }

scan-no-results = 
    Your email address did not appear in our basic scan.
    That's good news, but data breaches can happen any time and there is still more you can do. 
    Subscribe to Firefox Monitor for a full report, alerts when new breaches happen, and tips on protecting your passwords.
    
scan-results = { $breachCount ->
    [one] Your account appeared in the following breach.
  *[other] Your account appeared in the following { $breachCount } breaches.
}
```

Note also that your current string would result in `Your account Your email address did not appear in our basic scan. etc`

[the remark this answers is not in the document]
stasm: > I really think we should split out the 0 case. If nothing else for practical reasons: it would be impossible to translate this in Pontoon, where there no text area for variants.

From the Fluent perspective, I think using `[0]` like this is fine and it adds flavor to the translation. The [Good Practices document](https://github.com/projectfluent/fluent/wiki/Good-Practices-for-Developers#prefer-separate-messages-over-variants-for-ui-logic) has more discussion about this.

However, if Pontoon doesn't support extra variants right now, we should avoid it (for now). Sorry, @lesleyjanenorton! We'll get there eventually :)

groovecoder:
After @lesleyjanenorton's updates, I added commits to remove the test `de/app.ftl` file, to move `public/locales` to `locales` at the base of the repo, and to check for `NODE_ENV === "heroku"` to enable pontoon JS for in-page localization.

I've also pushed this branch directly to https://fx-breach-alerts.herokuapp.com/ so we can see this code running live before our upcoming meeting.

(Note: @lesleyjanenorton I'm getting `error {"stack":"TypeError: dataClasses.split is not a function
    at Object.breachDataClasses (/Users/lcrouch/code/mozilla/blurts-server/hbs-helpers.js:28:71)` on http://localhost:6060/user/verify?token=87dea900-428d-4b71-aad9-16b9ceb7ed92. Trying to track it down now.)

flodolo:
>  to check for `NODE_ENV === "heroku"` to enable pontoon JS for in-page localization.

I'm not exactly sure if that works reliably, or for Fluent projects. @mathjazz ?

mathjazz:
> I'm not exactly sure if that works reliably, or for Fluent projects. @mathjazz ?

In theory it works, but requires additional testing. AFAICT, it hasn't been used in combination with a Fluent.

I suggest we don't block enabling localization on it and enable in-place localization as a follow-up.

groovecoder:
In our case, I've specifically set `NODE_ENV=heroku` using `heroku:config` so it should work for us. Not sure what the broader `NODE_ENV` convention is.

I've also fixed the error on `/user/verify` so the text actually shows, though the page styles are messed up.

screen shot 2018-10-17 at 11 17 44 am

locales/en/app.ftl
```diff
@@ -0,0 +1,193 @@
+# String IDs beginning with "-product" and "-brand" should remain in English.
+# They should not be:
+# - Declined to adapt to grammatical case.
+# - Transliterated.
+# - Translated.
+-product-name = Firefox Monitor
```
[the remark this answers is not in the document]
groovecoder: Filed https://github.com/mozilla/blurts-server/issues/535 for a follow-up on this.

groovecoder:
I'm merging this now, as it touches so many files and is preventing us from doing much else until it's merged.

Filed https://github.com/mozilla/blurts-server/issues/533 for email localization, https://github.com/mozilla/blurts-server/issues/534 for the heroku/in-page localization follow-up, https://github.com/mozilla/blurts-server/issues/535 for `&nbsp;` discussion & decision, and  https://github.com/mozilla/blurts-server/issues/524 to verify that @mozilla-pontoon can/will create the launch `.ftl` files in the repository for launch.

They are all in [the `l10n` label](https://github.com/mozilla/blurts-server/issues?q=is%3Aissue+is%3Aopen+label%3Al10n) now.

PR #487, opened 2018-09-30
start saving signup language for future emails

groovecoder:
So we can send the right translated emails to subscribers, we need to start saving their `Accept-Language` value when they sign up.

We should merge and deploy this ASAP - the longer we run without this, the more users are going to have to set their language later. 😢


groovecoder:
Added `signup_language` expectation to existing test.

PR #509, opened 2018-10-06
Replace missing breach images -- temporary image bandaids.

lesleyjanenorton:
Bandaids for missing breach images while we sort out the final process.


hbs-helpers.js
```diff
@@ -88,6 +91,17 @@ function breachMath(lValue, operator = null, rValue = null) {
   return returnValue;
 }
 
+function checkImageSrc(imageName) {
```
groovecoder: nit: function name can be something like `logoImgSrc`

views/email/email_partials/breach_alert.hbs
```diff
@@ -1,6 +1,6 @@
     <tr>
       <td valign="top" style="padding-top: 3px; padding-bottom: 10px; width: 70px;">
-        <img src="{{ SERVER_URL }}/img/email_images/missing-logo-icon.png" width="50" style="width: 50px;" />
+        <img src="{{checkImageSrc this.Name}}" alt="" width="50" style="width: 50px"; />
```
groovecoder: Is `alt` left intentionally blank? Or should it be something like `alt="{{ this.Name }} logo icon"`?

hbs-helpers.js
```diff
@@ -88,6 +91,17 @@ function breachMath(lValue, operator = null, rValue = null) {
...
+  try {
+    fs.statSync(`public/img/logos/${imageName}.png`);
```
pdehaan: Not all logos are PNG, I think only the occasional SVGs that you manually convert to PNG (for email reasons) are PNG. Not sure if it matters or not, but if we're going to hardcode PNG in places, we may want to consider just dumping the JPG/SVG logos and just using PNG everywhere.

Actually, this may be moot since I think this helper may only be used in the email templates... 🤷‍♂️

pdehaan: Also, doing a filesystem statSync is probably pretty heavy if we're calling it for each logo in each email. Not sure if it makes sense to somehow just grab all *.png logos on the filesystem once during server setup and then just check our local object to see if the file exists. Like, if we stored all the logos in a `Map()`, I think we could just use ```EmailLogos.has(`${imageName}.png`)``` or something.

public/js/monitor.js
```diff
@@ -458,3 +463,9 @@ if (document.querySelectorAll("button")) {
     button.addEventListener("click", (e) => doButtonRouting(e));
   });
 }
+
+if (document.querySelector(".breach-image")) {
```
pdehaan: Not sure what this `if` is doing. IIUC, the `document.querySelectorAll()` would return an empty array if the selector wasn't found, so the `.forEach()` would be a noop. Not sure we need the extra check here.

hibp.js
```diff
@@ -64,16 +62,35 @@ const HIBP = {
     return await this._throttledGot(url, reqOptions);
   },
 
+  // make list of all PNG logos
+  makeLogoListForEmails() {
+    const emailLogos = [];
+    const images = fs.readdirSync("./public/img/logos/");
+    images.forEach(file => {
+      if (file.endsWith("png")) {
```
pdehaan: Can't imagine many cases where it'd cause an issue, or if you want to add slightly more footgun protection by using `.endsWith(".png")`.

hibp.js
```diff
@@ -117,7 +134,7 @@ const HIBP = {
     // ]
     for (const breachedAccount of response.body) {
       if (sha1.toUpperCase() === sha1Prefix + breachedAccount.hashSuffix) {
-        foundBreaches = allBreaches.filter(breach => breachedAccount.websites.includes(breach.Name));
+        foundBreaches = allBreaches.filter(breach => breachedAccount.websites.includes(breach.Title));
```
pdehaan: This feels like a curious change... Why is foundBreaches using `.Title`s now instead of `.Name`?

views/email/email_partials/email_single_breach.hbs
```diff
@@ -1,6 +1,6 @@
     <tr>
       <td valign="top" style="padding-top: 13px; padding-bottom: 10px; width: 70px;">
-        <img src="{{ SERVER_URL }}/img/logos/{{ this.Name }}.png" width="40" style="width: 50px" />
+        <img src="{{ SERVER_URL }}/img/logos/{{ this.ImgSrc }}.png" alt="" width="50" style="width: 50px"; />
```
pdehaan: The `;` is now in the wrong place in the inline `style=""`

hibp.js
```diff
@@ -64,16 +62,35 @@ const HIBP = {
...
         // const breach = breachesResponse.body[breachIndex];
-        // purify the description
-        breach.Description = DOMPurify.sanitize(breach.Description, {ALLOWED_TAGS: []});
+
+        // check if we have png logo & set breach.ImgSrc to "missing-logo-icon" if not.
+        if (!pngLogoList.includes(`${breach.Name}.png`)) {
+          log.info(`missing image ${breach.Name}`);
+          breach.ImgSrc = "missing-logo-icon";
+        } else {
+          breach.ImgSrc = `${breach.Name}`;
```
pdehaan: nit: just use `breach.ImgSrc = breach.Name;`. I don't think it needs interpolation.

lesleyjanenorton:
Closing this in favor of an actual automated process for getting images.

PR #522, opened 2018-10-08
script to calculate quickest breach alert

groovecoder:
A quick utility script I wrote to calculate the fastest response time for HIBP breach alerts.


groovecoder:
Just merging this so it's not only on my computer anymore.

PR #525, opened 2018-10-09
Verification email updates

lesleyjanenorton:
- Moves 'this is an automated email' copy down into footer so that it appears on each page.
- Updates copy and button copy in the verification email.


views/email/layouts/default_email.hbs
```diff
@@ -14,7 +14,7 @@
         color: #333333;
         background: #ededf0;
       }
-
+
```
groovecoder: ⚡️ this blank space.

PR #537, opened 2018-10-18
Add CODEOWNERS and l10n.toml

flodolo:
CODEOWNERS will ping me for review whenever FTL files in /en are changed

l10n.toml can be used for automation, i.e. install `compare-locales` 

```
pip install compare-locales
```

Then run from the root

```
compare-locales l10n.toml . it
```

You can also get JSON and use it for automation (level of completion, errors).

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -1,0 +1,2 @@
+# flod as main contact for string changes
+locales/en/*.ftl @flodolo
--- a/l10n.toml
+++ b/l10n.toml
@@ -1,0 +1,14 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+basepath = "."
+
+locales = []
+
+[env]
+    l = "{l10n_base}/locales/{locale}/"
+
+[[paths]]
+    reference = "locales/en/**"
+    l10n = "{l}**"


groovecoder:
Cool, thanks.

pdehaan:
@flodolo I'm getting the following errors. Not sure if I don't have it configured correctly, or there are a couple errors/warnings we need to address:

```sh
$ compare-locales l10n.toml . `ls locales`

locales/en
  app.ftl
      ERROR: Unparsed content "user-unsubscribed-title {-product-name} : Unsubscribed" from line 48 column 1 to line 48 column 55
      WARNING: Parser error in en-US
      ERROR: Unparsed content "signup-with-fxa - Sign Up with {-brand-name} Account" from line 107 column 1 to line 107 column 53
      WARNING: Parser error in en-US
  data-classes.ftl
      ERROR: Unparsed content "school-grades-(class-levels) = School grades (class levels)" from line 96 column 1 to line 96 column 60
      WARNING: Parser error in en-US
en:
errors            3
unchanged       226
unchanged_w    1175
0% of entries changed
```

pdehaan:
I think in the case of `school-grades-(class-levels)`, the problem is coming from:
https://github.com/mozilla/blurts-server/blob/2cfcffd4b54a7e2b1b34ae9921c1f590b160b93d/hbs-helpers.js#L21-L32

We probably need a better algorithm/function that replaces any non a-Z0-9 with a hyphen or something. 

It looks like we're currently only doing a `'` (single quote) to (empty string) conversion as well as converting a space to a hyphen. Oddly looks like we're only calling `.toLowerCase()` in one of the branches, which may be suspicious too.

Maybe something like https://github.com/blakeembrey/param-case logic.

PR #538, opened 2018-10-18
I20n updates

lesleyjanenorton:
- Updates templates and helpers to send/accept req.supportedLocales instead of req.
- Adds additional logic to scan.hbs for scans coming from featured breach pages.
- Adds strings to ftl file.

### DIFF
--- a/locales/en/app.ftl
+++ b/locales/en/app.ftl
@@ -114,18 +114,28 @@ form-signup-error = Must be a valid email
 found-breaches-headline = This could be a problem
 no-breaches-headline = So far, so good
 
+-no-breaches = 
+        Your email address did not appear in our basic scan.
+        That’s good news, but data breaches can happen any time and there is still more you can do. 
+        Subscribe to Firefox Monitor for a full report, alerts when new breaches happen, and tips on protecting your passwords.
+
 featured-breach-results =  
   { $breachCount ->
-      [0] Your account appeared in the <span class="bold"> { $featuredBreach } </span> breach, but did not appear in any other known data breaches.
+      [0] Your account appears in the <span class="bold">{ $featuredBreach }</span> breach, but does not appear in any other known data breaches.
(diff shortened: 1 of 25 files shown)


locales/en/app.ftl
```diff
@@ -114,18 +114,28 @@ form-signup-error = Must be a valid email
...
   { $breachCount ->
-      [0] Your account appeared in the <span class="bold"> { $featuredBreach } </span> breach, but did not appear in any other known data breaches.
+      [0] Your account appears in the <span class="bold">{ $featuredBreach }</span> breach, but does not appear in any other known data breaches.
       [1] Your account appeared in the <span class="bold"> { $featuredBreach } </span> breach, as well as one other breach.
       *[other] Your account appeared in the <span class="bold"> { $featuredBreach } </span> breach, as well as { $breachCount } other breaches.
   }
 
+featured-breach-not-compromised = 
+  { $breachCount ->
+    [0] {-no-breaches}
```
groovecoder: nit: the indentation of these variations are even with the `$` character while the variations inother items are indented 2 spaces more then the `$` character.

locales/en/app.ftl
```diff
@@ -114,18 +114,28 @@ form-signup-error = Must be a valid email
...
+    [1] Your account did not appear in the <span class="bold">{ $featuredBreach }</span> breach, but did appear in one other breach.
```
flodolo: For English the practical result is the same, but I'd prefer this to be `[one]`.

locales/en/app.ftl
```diff
@@ -114,18 +114,28 @@ form-signup-error = Must be a valid email
 found-breaches-headline = This could be a problem
 no-breaches-headline = So far, so good
 
+-no-breaches =
```
flodolo: This shouldn't be a term. You can [reference standard messages](https://projectfluent.org/fluent/guide/references.html) in other messages.

PR #539, opened 2018-10-18
Updates website strings

lesleyjanenorton:
Updates website strings with latest copy and restructures What To Do section to accommodate larger amounts of copy.


pdehaan:
Since we're editing these files, I found a few other l10n typos/bugs in https://github.com/mozilla/blurts-server/pull/537#issuecomment-431165930

Sorry, I should have submitted my comments in a new ticket instead of piggybacking off a merged PR, so I'll migrate those comments to a new issue shortly.

hbs-helpers.js
```diff
@@ -22,8 +22,8 @@ function localizedBreachDataClasses(supportedLocales, dataClasses, args) {
   const localizedDataClasses = [];
   if (dataClasses.constructor === Array) {
     dataClasses.forEach(dataClass => {
-      dataClass = dataClass.replace("'", "");
-      localizedDataClasses.push(LocaleUtils.fluentFormat(supportedLocales, dataClass.split(" ").join("-").toLowerCase(), args));
+      dataClass = dataClass.replace(/[^-a-z0-9]/ig,"-").toLowerCase();
```
pdehaan: I feel like this will be needed on L30 below as well, and maybe a candidate for a new function maybe. Not sure if we care about how pretty the string is with multiple concurrent hyphens or trailing hyphens. Maybe not worth extra fanciness if it’s only an l10n key.

locales/en/data-classes.ftl
```diff
@@ -93,7 +93,7 @@ relationship-statuses = Relationship statuses
 religions = Religions
 reward-program-balances = Reward program balances
 salutations = Salutations
-school-grades-(class-levels) = School grades (class levels)
+school-grades--class-levels- = School grades (class levels)
```
pdehaan: This (and L41 above) are what I meant about double hyphens and trailing hyphens.

Also curious what would happen if HIBP adds a new data class. How would we detect that and add it to our l10n chain?

hbs-helpers.js
pdehaan: Here's a pretty crude `slugify()`, if it helps:

```js
function slugify(str) {
  return str.toLowerCase()
    .replace(/[^-a-z0-9]/ig, "-") // replace non a-z and 0-9 w/ hyphens. 
    .replace(/-{2,}/g, "-") // collapse multiple hyphens into single hyphen.
    .replace(/^\-?(.+?)\-?$/g, "$1"); // trim any leading or trailing hyphens.
}

console.log(slugify("Peter is lame")); // peter-is-lame
console.log(slugify("Peter is as lame (as a duck)")); // peter-is-as-lame-as-a-duck
console.log(slugify("Peter's lameness intensifies!")); // peter-s-lameness-intensifies
console.log(slugify("'(fancy) pork', they shouted!")); // fancy-pork-they-shouted
console.log(slugify("----peter-IS-LAAAMMMMEEEE---- and that's a guarantee!!!")); // peter-is-laaammmmeeee-and-that-s-a-guarantee
```

pdehaan: Actually, here's the diff from the current parser [1] versus the "new" `slugify` parser [2]. But basically this was just an excuse for me to test out **fluent-syntax** for parsing .ftl files.

```js
const {readFileSync} = require("fs");
const {parse} = require("fluent-syntax");

const dataClasses = readFileSync("./locales/en/data-classes.ftl", "utf-8");

parse(dataClasses).body.forEach(item => {
  if (item.type === "GroupComment") return;
  const value = item.value.elements.map(({value}) => value).join(" ");
  const slug1 = item.id.name;
  const slug2 = slugify(value);
  if (slug1 !== slug2) {
    console.log(`${value}\n\t[1]: ${slug1}\n\t[2]: ${slug2}`);
  }
});

function slugify(str) {
  return str.toLowerCase()
    .replace(/[^-a-z0-9]/g, "-")
    .replace(/-{2,}/g, "-")
    .replace(/(^-|-$)/g, "");
}
```

```
Family members’ names
        [1]: family-members--names
        [2]: family-members-names

School grades (class levels)
        [1]: school-grades--class-levels-
        [2]: school-grades-class-levels
```

views/scan.hbs
```diff
@@ -16,7 +16,7 @@
           <p>{{{fluentFormat req.supportedLocales "featured-breach-not-compromised" featuredBreach=featuredBreach.Title breachCount=foundBreaches.length}}}</p>
         {{/if}}
       {{else}}
-        <p>{{fluentFormat req.supportedLocales "scan-results" breachCount=foundBreaches.length}}</p>
+        <p>{{{fluentFormat req.supportedLocales "scan-results" breachCount=foundBreaches.length}}}</p>
```
groovecoder: This made me file https://github.com/mozilla/blurts-server/issues/542 for follow-up, but doesn't need to block.

hbs-helpers.js
lesleyjanenorton: Thanks @pdehaan ! Got mired in email stuff yesterday and missed these until much later. They're just l10n keys, but I think it's worth getting rid of trailing / multiple consecutive hyphens anyway. Just pushed up something to format the l10n keys when we load breaches, but it doesn't address what happens when new data classes are added.

hbs-helpers.js
```diff
@@ -19,16 +19,11 @@ function getStringID (supportedLocales, key, number) {
 
 
 function localizedBreachDataClasses(supportedLocales, dataClasses, args) {
-  const localizedDataClasses = [];
-  if (dataClasses.constructor === Array) {
+    const localizedDataClasses = [];
     dataClasses.forEach(dataClass => {
```
pdehaan: I’m confused here. Before we were checking if dataClasses way an array and then doing X. Now we’re just always assuming it’s an Array and calling `.forEach()`. Was there a case/reason where it was either an array vs a plain string?

hibp.js
```diff
@@ -65,6 +63,25 @@ const HIBP = {
...
+
+  formatDataClassesArray(dataCategories) {
+    const formattedArray = [];
+    if (dataCategories.constructor === Array) {
+      dataCategories.forEach(category => {
+        formattedArray.push(this.matchFluentID(category));
+      });
+      return formattedArray;
+    } else {
+      return formattedArray.push(this.matchFluentID(dataCategories));
```
pdehaan: This looks suspicious. 
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/push

Looks like this would return the length of the array (always 1 in this case), whereas the `if` above is returning an actual array.

🎤 👇

locales/en/app.ftl
```diff
@@ -200,4 +238,6 @@ copyright-info =
   Portions of this content are &#x24B8; 1998-2018 by individual mozilla.org contributors. <br />
   Content available under a  <a href="https://www.mozilla.org/foundation/licensing/website-content/" target="_blank" rel="noopener">Creative Commons license</a>.
 
+# This string is followed by a link to Have I Been Pwned, the code for which is created elsewhere.
+# The final string reads: Breach data provided by Have I Been Pwned.
```
pdehaan: Ref: 
https://github.com/mozilla/blurts-server/pull/543#discussion_r226927638

I think @flodolo was saying in the other PR that the concat isn’t great for l10n and we should do a placeholder link so localizers can rearrange where the link best fits on a per-language basis.

hbs-helpers.js
groovecoder: the mSpy breach has a single data class, which blew up the old `localizedBreachDataClasses` logic before we checked that the argument value was an array.

But now this branch adds `formatDataClassesArray` which transforms the `breach.DataClasses` value from HIBP into an array that we can localize, so now it's always an array.

locales/en/app.ftl
groovecoder: Let's do the fix in https://github.com/mozilla/blurts-server/pull/543#discussion_r227007616 so we can merge this PR.

Issue #541, opened 2018-10-19
Run compare-locales in travis-ci


Pike:
@pdehaan mentioned this in a couple PRs, it'd be great to run compare-locales in automation.

I think it makes sense to just run that in general, one for the validation mode, and one for the l10n check.

groovecoder:
Thanks @Pike. I submitted https://github.com/mozilla/blurts-server/pull/544 which runs the validation in Travis to help make sure we don't merge any invalid `.ftl` changes.

What do you mean by "the l10n check"? I could run:
```
compare-locales l10n.toml . `ls locales`
```
And just make sure it returns with `0`?

Pike:
`compare-locales --validate l10n.toml .`

reports errors in en-US as actual errors, while they're just warnings if you run them across locales.

en-US validation and l10n validation cannot be run in one go right now, so you want both, eventually.

groovecoder:
Ah okay so I can just add 2 calls to `compare-locales`? One with `--validate` and one without?

pdehaan:
@groovecoder Can we close this now that #544 was merged, or is there further tasks to do?

groovecoder:
Yup. I had it open to remind me we should do something like https://github.com/mozilla/blurts-server/issues/550. But now we have https://github.com/mozilla/blurts-server/issues/550 open for that.

PR #544, opened 2018-10-22
for #541: run compare-locales validate on Travis

groovecoder:


Pike:
I see that you're already doing the `--validate` that I suggested in #541.

Now we just need gh to tell travis to actually do something ;-)

groovecoder:
Still just waiting on GitHub ... ["We are working through the backlogs of webhook deliveries"](https://status.github.com/messages)

Pike:
OK, this doesn't work in the node build.

We'll need a matrix of python builds for compare-locales and node builds for the rest.

https://stackoverflow.com/questions/27644586/how-to-set-up-travis-ci-with-multiple-languages is what I've found on the tubes.

pdehaan:
I _think_ the failed build is due to some old Python 2.7 version on Travis CI.

Error seems to be related to https://urllib3.readthedocs.io/en/latest/advanced-usage.html#ssl-warnings

> ### SSL Warnings
> ...
> **SNIMissingWarning**
> This happens on Python 2 versions older than 2.7.9. These older versions lack SNI support. This can cause servers to present a certificate that the client thinks is invalid. Follow the pyOpenSSL guide to resolve this warning.

I can't see anywhere in the output where it says what exact version of Python we're using, apart from vague references to "2.7":

> $ pip install compare-locales
Collecting compare-locales
>
> /usr/local/lib/python2.7/dist-packages/pip/_vendor/requests/packages/urllib3/util/ssl_.py:318: SNIMissingWarning: An HTTPS request has been made, but the SNI (Subject Name Indication) extension to TLS is not available on this platform. This may cause the server to present an incorrect TLS certificate, which can cause validation failures. You can upgrade to a newer version of Python to solve this. For more information, see https://urllib3.readthedocs.io/en/latest/security.html#snimissingwarning.
  SNIMissingWarning
> 
> /usr/local/lib/python2.7/dist-packages/pip/_vendor/requests/packages/urllib3/util/ssl_.py:122: InsecurePlatformWarning: A true SSLContext object is not available. This prevents urllib3 from configuring SSL appropriately and may cause certain SSL connections to fail. You can upgrade to a newer version of Python to solve this. For more information, see https://urllib3.readthedocs.io/en/latest/security.html#insecureplatformwarning.
  InsecurePlatformWarning

Although https://docs.travis-ci.com/user/languages/python/#travis-ci-uses-isolated-virtualenvs seems to say that for "Trusty", this means Python **2.7.6** and **3.4.3**.

I haven't found a good answer on how to force Circle/Ubuntu to use the Python3 executable by default yet, apart from switching to a matrix of containers.
I'll also read a bit more on it and see if there is a way we can just ignore these TLS errors, because whatever.

pdehaan:
Actually, the error message was this, and I think the other messages are warnings:

> OSError: [Errno 13] Permission denied: '/usr/local/lib/python2.7/dist-packages/pytoml'

.travis.yml
```diff
@@ -6,7 +6,11 @@ services:
   - postgresql
 env:
   - NODE_ENV=tests
+install:
```
pdehaan: Interesting, now we're getting the following error:

> sh: 1: npm-run-all: not found

I wonder if this is causing <kbd>$ npm install</kbd> (or <kbd>$ npm ci</kbd>) to not get called, and maybe we should rename this to "before_install:", similar to https://docs.travis-ci.com/user/languages/javascript-with-nodejs/#using-a-specific-npm-version

groovecoder: Yeah, just updated `install` to do both `pip` and `npm`

groovecoder:
Yay! now with both `pip install` and `npm install` during `install`

groovecoder:
Cool, I added a commit that introduced a parse error in the `locales/en/app.ftl` file and Travis failed it.

So I think this is good to go!

pdehaan:
> I think we should change the l10n check to not do `ls` but instead add the list of locales to `l10n.ini`.
> 
> That's going to make it easier to switch things on, and infer l10n status from the configuration files.
> 
> And yes, in [bug 1501220](https://bugzilla.mozilla.org/show_bug.cgi?id=1501220) I'm tracking to add validation to that data based on local data.

Filed new bug ___

PR #548, opened 2018-10-22
Restores lost featured breach colons

lesleyjanenorton:
Restores colons to breaches on /?breach= pages that were lost sometime during i-EIGHTEEN-n.


pdehaan:
🤷‍♂️ /summon @flodolo to answer whether colons should be inside or outside of the .ftl files.

flodolo:
Ugh :-\

It should definitely be in the FTL. That's a question mark for Greek, French needs a non-breaking space before it, etc.

If you can merge this quickly, I can try to let you update the strings without a new ID and manually fixing existing localization afterwards.

groovecoder:
Merged.

PR #549, opened 2018-10-22
fix #547: fix locale args to hbs helpers

### LINKED ISSUE #547: Change calls to `toLocaleString` and `toLocaleDateString` to pass a single language
I kinda mis-filed this but I was able to work out what we really need to do.  Submitting PR.

groovecoder:


package.json
```diff
@@ -20,6 +20,7 @@
     "express-handlebars": "^3.0.0",
     "fluent": "^0.8.1",
     "fluent-langneg": "^0.1.0",
+    "full-icu": "^1.2.1",
```
groovecoder: This installs all the icu data that `Intl` needs to format dates & numbers with long-form months, etc.

groovecoder: See https://nodejs.org/api/intl.html#intl_embed_the_entire_icu_full_icu

groovecoder:
Easiest way to spot-check is to go thru the visit + scan + signup + verify + breach flow while your browser has "Italian" as its preferred language.

Dates & Numbers should look like this:

screen shot 2018-10-22 at 3 38 47 pm

PR #551, opened 2018-10-23
Add strings for l10n

lesleyjanenorton:
Adds site description, and "Sent!" (in modal, after resend) to ftl strings.

### DIFF
--- a/locales/en/app.ftl
+++ b/locales/en/app.ftl
@@ -120,6 +120,9 @@ signup-modal-verify-blurb = We sent a verification link to <span id="submitted-e
 signup-modal-verify-expiration = This link expires in 24 hours.
 signup-modal-verify-resend = Not in inbox or spam folder? Resend.
 
+# Appears after Firefox Monitor has sent a verification email to a new user. 
+signup-modal-sent = Sent!
+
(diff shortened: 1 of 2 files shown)


locales/en/app.ftl
```diff
@@ -120,6 +120,9 @@ signup-modal-verify-blurb = We sent a verification link to <span id="submitted-e
 signup-modal-verify-expiration = This link expires in 24 hours.
 signup-modal-verify-resend = Not in inbox or spam folder? Resend.
 
+# Appears after Firefox Monitor has sent a verification email to a new user. 
+signup-modal-sent = Sent!
```
pdehaan: Where is this being used?
I did <kbd>$ git grep "signup-modal-sent"</kbd> but didn't get any results.

lesleyjanenorton: So as of this PR the "Sent!" is still being generated by the CSS but I wanted to get the additional strings in asap and at the same time. I'll push something up today that implements the localized string.

pdehaan: Ah, perfect!

```sh
$ git grep -n "Sent\!"

public/css/app.css:1115:  content: "Sent!";
```

PR #559, opened 2018-10-24
Disable Unicode isolation

lesleyjanenorton:
This disables Unicode isolation and gets rid of "PDI" and "FSI" boxes all over the place in Edge. What are the potential negative ramifications of setting { useIsolating: false } for all browsers?

Our current situation in Edge:
before

### DIFF
--- a/locale-utils.js
+++ b/locale-utils.js
@@ -40,7 +40,7 @@ const LocaleUtils = {
     });
     for (const lang of languageDirectories) {
       try {
-        const langBundle = new FluentBundle(lang);
+        const langBundle = new FluentBundle(lang, {useIsolating: false});
         const ftlFiles = fs.readdirSync(path.join(localesDir, lang)).filter(item => {
           return (item.endsWith(".ftl"));
         });
--- a/public/css/app.css
+++ b/public/css/app.css
@@ -645,6 +645,8 @@ h5,
 
 .bold {
   font-weight: 600;
+  direction: ltr;
+  unicode-bidi: embed;
 }
 
 /* internal link styling  */


pdehaan:
🤷‍♂️ 

Summoning @stasm and @punamdahiya since I noticed a similar change in [**mozilla-services/screenshots**](https://github.com/mozilla-services/screenshots/pull/5025), which was reverted recently, so they may have some keen insight.

I think both TestPilot and Firefox Send are still setting this to `false`:

| ISSUE | SOURCE |
|:------|:-------|
| https://github.com/mozilla/testpilot/issues/1828 | [/frontend/src/app/containers/App/index.js:223](https://github.com/mozilla/testpilot/blob/3e96a91464d13d4e4534fb94f0f918d83ed74a8f/frontend/src/app/containers/App/index.js#L223)
https://github.com/mozilla/send/issues/476 | [/build/fluent_loader.js:46](https://github.com/mozilla/send/blob/62d507120c7d925372df90f2b7a4194db3e9ab1c/build/fluent_loader.js#L46)

punamdahiya:
@pdehaan useisolating:false is the [recommended workaround ](https://github.com/mozilla-services/screenshots/issues/4863#issuecomment-422896098) to get rid of the unicode characters in windows until https://github.com/projectfluent/fluent.js/issues/173 is fixed
However the PR caused regression and broke localization for screenshots and had to revert it sooner https://github.com/mozilla-services/screenshots/pull/5025

@stasm can help explain better why the change regressed localization. Thanks

pdehaan:
Interesting, per @zbraniecki's comment in https://github.com/projectfluent/fluent.js/issues/173#issuecomment-375386568:

> _For now, using `useIsolating=false` will "fix" it (but regress BiDi)_

stasm:
I'd like to provide a bit more detail about this behavior and the reasons for turning it on by default.

### Why is isolation needed?

Most of the time, the [Unicode BiDi Algorithm](http://unicode.org/reports/tr9/) handles bi-directional text very well. In some rare cases, however, it needs a little bit of help. This happens sometimes when a run of RTL text is embedded into a larger run of LTR text, or the other way round. It's due to the fact that some characters accord their directionality with the characters around them.

> Each character has an implicit bidirectional type. The bidirectional types left-to-right and right-to-left are called strong types, and characters of those types are called strong directional characters. The bidirectional types associated with numbers are called weak types, and characters of those types are called weak directional characters. With the exception of the directional formatting characters, the remaining bidirectional types and characters are called neutral. The algorithm uses the implicit bidirectional types of the characters in a text to arrive at a reasonable display ordering for text.

If the embedded text starts or ends with a weak character (such as `/`, `:` or `!`), the algorithm needs help to determine if the character should inherit the directionality of the embedded text, or the text around the embedding.


### What does it look like?

Assume the following Fluent message which receives a URL as an argument from the app:

```properties
privacy-more = Visit {$PRIVACY_URL} to learn more.
```

Let the URL end with a forward slash, like so:

```json
{"PRIVACY_URL": "https://www.mozilla.org/privacy/"}
```

Without isolation, the rendered result shows the trailing slash 

> <p dir="rtl">تفضل بزيارة https://www.mozilla.org/privacy/ لمعرفة المزيد. ❌</p>

With isolation, the forward slash is displayed correctly:

> <p dir="rtl">تفضل بزيارة ⁨https://www.mozilla.org/privacy/⁩ لمعرفة المزيد. ✅</p>


### Why is it on by default?

`useIsolating` is set to `true` by default for two reasons:

1. The Unicode Consortium recommends that isolation be used as the default for all future inline bidirectional text embeddings. From [UAX #9](http://unicode.org/reports/tr9/#Directional_Formatting_Characters):

    >  [T]he use of the directional isolates instead of embeddings is encouraged in new documents – once target platforms are known to support them. 

    The W3C [recommends the same practice](https://www.w3.org/International/articles/inline-bidi-markup/#nomarkup) in situations when markup cannot be used.

2. For HTML, where markup is available, both the Unicode Consortium and the W3C recommend using the `<bdi>` element or the `dir="auto"` attribute. However, `fluent.js` can't always rely on these to be available. Translations formatted by Fluent might be displayed to the user in other environments than HTML, even if they're still using JS. Think: node.js, `alert()` or React, which would require the use of `dangerouslySetInnerHTML` to allow arbitrary HTML markup in translations.

### Can I turn it off?

If `useIsolation` is set to `false`, there's a risk that some translations won't look correctly in bidi scenarios. For RTL languages (Arabic, Hebrew and a few more), this happens when interpolated variables are LTR and start or end with weak characters (URLs, book or website titles, citations, Wi-Fi network names, addon names, etc.). For LTR languages, this happens when the interpolated variables are RTL and have leading or trailing weak characters (names and title of all sorts are the most common use-case here).

In the particular case of Firefox Monitor, since all translations are formatted on the server side and inserted into Handlebars templates, you could also use what W3C recommends: turn `useIsolating` off and use markup instead.

```properties
privacy-more = Visit <bdi>{$PRIVACY_URL}</bdi> to learn more.
```

This will have the same effect but it requires a bit more work. @flodolo What do you think about this approach? Ideally, every translation using interpolations would be reviewed to decide if it needs `<bdi>` or not in `en-US`. You could also use a group comment within the Fluent file to tell localizers that they can:

1. Use `<bdi>` around interpolations, or
2. Add `dir="auto"` to elements already found in the translations, e.g. `<span class="bold">{$featuredBreach}</span>`.

Group comments start with two hashes (`##`) and Pontoon shows them next to every message in  the group, i.e. until the next group comment.

groovecoder:
@zbraniecki @stasm so what's your recommendation here? Should we disable the unicode isolation to fix Edge, or should we wait for Edge to release their own fix?

Because we initialize our `FluentBundle`s on server start-up, I don't think we could detect Edge from each request's `User-Agent` and disable unicode isolation only for Edge users.

stasm:
Realistically, this only affects `<span class="bold"> { $featuredBreach } </span>` in `app.ftl`. What kind of content is possible as values on `$featuresBreach`?

1. If it's just titles of websites, like _Amazon_, _Facebook_, or _Bitly_, then I would recommend disabling `useIsolating` for all users and nothing more.

2. If it's URLs or if many titles have punctuation at the end (like _Yahoo!_), then additionally, I would reach out to our RTL localizers (via the dev-l10n mailing list) and let them know that they might want to use `<bdi>` or `<span dir="auto">` in their translations.

3. If some titles of the breaches are written in RTL scripts, we need _all_ LTR languages to use isolation, too. I would recommend adding `dir="auto"` to all `<span class="bold"> { $featuredBreach } </span>` in `locales/en/app.ftl` in that case. @flodolo, can we do this without having to change message ids?

stasm:
Oh, if the titles of breaches are guaranteed to be LTR, you could also disable `useIsolating` and use CSS to solve the _Yahoo!_ case:

```css
span.bold {
    direction: ltr;
    unicode-bidi: embed;
}
```

PR #561, opened 2018-10-24
Add period to unsubscribe-link string and create new string ID.

### LINKED ISSUE #555: period for unsubscribe-link should be localizable inside the string
https://github.com/mozilla/blurts-server/blob/d77784a816581865a3e38248097890b99f004895/views/email/email_partials/email_footer.hbs#L12,

The period should be moved into the string, as not every languages use period as a full stop.

lesleyjanenorton:
Fixes #555

### DIFF
--- a/locales/en/email-strings.ftl
+++ b/locales/en/email-strings.ftl
@@ -13,8 +13,8 @@ automated-message =
 # Without HTML markup reads: We sent this message to [user’s email address] because the email address opted into alerts from Firefox Monitor.
 we-sent-this-alert = We sent this message to { $userEmail } because the email address opted into alerts from {-product-name}.
 
-unsubscribe-link  = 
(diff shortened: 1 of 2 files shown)


groovecoder:
@flodolo if we want to launch Nov 12/13, what date would you suggest to freeze the strings for l10n?

flodolo:
I currently put a deadline on November 7. We should freeze a few days before that.

Alternatively, we could freeze content on that date, but I would still try to add/change strings only for emergencies, because it's frustrating to complete a project and find it incomplete a couple of days after.

groovecoder:
Dang, the code here looked right, but when I merged I now see this in the emails (in both zh-CN and en). 😢 

screen shot 2018-10-25 at 9 58 32 am

lesleyjanenorton:
@groovecoder  I just checked en and it's working for me. Maybe try restarting your server?

PR #566, opened 2018-10-30
bug #1490830: add sentry and capture errors

groovecoder:


middleware.js
```diff
@@ -53,6 +54,7 @@ function asyncMiddleware (fn) {
 
 function logErrors (err, req, res, next) {
   log.error("error", {stack: err.stack});
+  Sentry.captureException(err);
```
pdehaan: Do we want to capture stacks on dev/stage/prod? or just stage/prod? prod only?
Or are we just overriding the `SENTRY_DSN` env var on production so it routes to a different bucket?

groovecoder: I'd like to, but I can't tell from [the docs](https://docs.sentry.io/learn/capturing/?platform=node) how we're supposed to pass the stacktraces into `captureException` 😢

groovecoder:
Updated with calls I verified worked on https://sentry.prod.mozaws.net/operations/firefox-monitor-prod/ and including an environment property so we can distinguish dev, heroku, stage, prod.

groovecoder:
Merging to trigger deploy to Heroku which will show if the `environment` property is working as expected.

PR #568, opened 2018-11-02
Add email mockups for easier l10n testing. 

lesleyjanenorton:
- Consolidates email templates into web templates folder
- Adds new route and sets up new layout that calls in existing email partials so that translators can see and spot-check localized emails.


controllers/email-l10n.js
```diff
@@ -0,0 +1,45 @@
+"use strict";
+
+const HBSHelpers = require("../hbs-helpers");
+
+
+function testEmails(req, res) {
+  const unsafeBreachesForEmail = [];
+  const email = "example@email.com";
```
pdehaan: Is this the "From:" email?
Either way, feels a bit suspiciously "debuggy".

controllers/email-l10n.js
```diff
@@ -0,0 +1,45 @@
...
+
+  if (req.query.type) {
+    buttonValue = req.fluentFormat("report-scan-another-email");
+    if(req.query.type === "singleBreach" || req.query.type === "breachAlert") {
+      unsafeBreachesForEmail.push(req.app.locals.breaches.filter(breach => breach.Name === "Experian")[0]);
+      if(req.query.type === "breachAlert") {
+        breachAlert = unsafeBreachesForEmail[0];
+      }
+    } else if (req.query.type === "multipleBreaches") {
+      const breachArray = ["Experian", "Dropbox", "Apollo"];
```
pdehaan: What is this hardcoded list of breaches for?
I see we're looping over it below, but curious why we need to hardcoded 3 breach names here.

controllers/email-l10n.js
```diff
@@ -0,0 +1,45 @@
...
+        unsafeBreachesForEmail.push(req.app.locals.breaches.filter(breach => breach.Name === name)[0]);
+      });
+    }
+  }
+
+  res.render("test_emails", {
+    layout: "email_l10n_testing.hbs",
+    unsafeBreachesForEmail: unsafeBreachesForEmail,
+    supportedLocales: req.supportedLocales,
+    whichView: `email_partials/${req.query.view}`,
```
pdehaan: Does this need some validation?
I see we're guarding against a null/missing `req.query.view` on L12-14 above, but not sure what would happen if the query `?view=` is somehow malformed or something. Although looking at this again, I see its a `testEmail()` method and "test_emails" view, so maybe this is just debugging code.

email-utils.js
```diff
@@ -15,12 +15,12 @@ const log = mozlog("email-utils");
...
     extname: ".hbs",
-    layoutsDir: __dirname + "/views/email/layouts",
+    layoutsDir: __dirname + "/views/layouts",
     defaultLayout: "default_email",
-    partialsDir: __dirname + "/views/email/email_partials/",
+    partialsDir: __dirname + "/views/partials",
     helpers: HBSHelpers,
   },
-  viewPath: __dirname + "/views/email/",
+  viewPath: __dirname + "/views/layouts",
```
pdehaan: views and layouts are now in the same directory? (comparing L23 and L18 above)

public/css/email_l10n_testing.css
```diff
@@ -0,0 +1,176 @@
...
+.email-logo,
+.email-body,
+.email-footer {
+  padding: 30px;
+}
+
+.email-content {
+  overflow: hidden;
+  border-radius: 3px;
+  box-shadow: 2px 2px 5px #7788995c;
```
pdehaan: Not sure on your feelings on #RRGGBBAA versus `rgba(rr, gg, bb, aa)` above on L5. Not sure if one has better cross-browser support, or something.

views/partials/email_partials/email_single_breach.hbs
```diff
@@ -0,0 +1,13 @@
+    <tr class="email-breach-listing">
+      <td class="email-breach-image" valign="top" style="padding-top: 13px; padding-bottom: 10px; width: 55px;">
+        <img src="{{ SERVER_URL }}/img/logos/{{ this.Name }}.png" width="40" style="width: 40px" />
```
pdehaan: I can't remember from the other PRs, but does this take into account a new breach named "Foo" but we don't have any logos for it yet? Or will this send out a link to "Foo.png" which will 404 in the email?

controllers/email-l10n.js
lesleyjanenorton: We really only actually need two but it's there so that translators can see an example of a report with multiple breaches as well as a singular breach since the strings are different.

public/css/email_l10n_testing.css
lesleyjanenorton: Oops- RGBA is the clear winner according to canIuse. Thanks!

controllers/email-l10n.js
lesleyjanenorton: No debugging here,  just totally crummy nomenclature on my part! Changing this to something less completely misleading and will add something to catch query malformations.

views/partials/email_partials/email_single_breach.hbs
lesleyjanenorton: In a report scenario, it will 404 all over the place. The breach alerts are hard-coded to show the missing-image icon for now.

controllers/email-l10n.js
```diff
@@ -0,0 +1,59 @@
...
+    supportedLocales: req.supportedLocales,
+    whichPartial: `email_partials/${req.query.partial}`,
+    date: HBSHelpers.prettyDate(req.supportedLocales, new Date()),
+    buttonValue,
+    breachAlert,
+    email,
+  });
+}
+
+function notFound(req, res) {
```
groovecoder: I don't think we need this extra `notFound` controller & route. I'd like to merge this to deploy for l10n QA ASAP, so I filed https://github.com/mozilla/blurts-server/issues/578 as a follow-up.

PR #579, opened 2018-11-07
UI Nits 

lesleyjanenorton:


lesleyjanenorton:
Addresses ui bugs noticed while testing locales and does some code cleaning.

PR #603, opened 2018-11-15
Updates sensitive featured breach pages

lesleyjanenorton:
- Replaces scan form on sensitive featured breach pages with a "Get Your Report" button. Currently if a user lands on a sensitive featured breach page (/?breach=AshleyMadison, for example) and scans their email address they will be told that their email address was not compromised in that breach regardless of whether or not it actually was compromised since we are not showing sensitive breach results to the public via the website.

- Removes the "Get Your Report" button in the sign up banner on sensitive featured breach pages and expands the blurb about Firefox Monitor reports.


flodolo:
Noting here since it's somehow related to the topic

```
signup-modal-verify-headline = Verify Your subscription
```

The Dutch localized noticed the weird capitalization on this one. Is this expected? In case, you can fix it directly, no need for a new message ID.

lesleyjanenorton:
@flodolo not expected- thank you for the catch! I will update.

controllers/home.js
```diff
@@ -5,15 +5,21 @@
 function home(req, res) {
   let featuredBreach = null;
   let scanFeaturedBreach = false;
+  let breachIsSensitive = false;
```
groovecoder: Could be a ternary assignment expression in the `if (req.query.breach)` block:
```
const breachIsSensitive = featuredBreach.IsSensitive ? true : false;
```

groovecoder: Nvm, looks like we need to default the assignment to `false` first.

PR #617, opened 2018-11-20
Adds new strings for sharing by email and social media.

lesleyjanenorton:
Adds copy for new site-description, new email message, and social media share posts/tweets. 

@flodolo some of the previously added strings will become obsolete once we implement the strings in this PR. Should the obsolete strings remain in the FTL file or is it safe to delete them?

### DIFF
--- a/locales/en/email-strings.ftl
+++ b/locales/en/email-strings.ftl
@@ -82,7 +82,7 @@ legal = Legal
 share-by-email-subject = See if you’ve been part of a data breach.
 
 # Share Firefox Monitor by email message. Followed by steps about how Firefox Monitor works. 
-share-by-email-message = Hi, { $markup }{-brand-name} has a free service where you can check to see if you've been part of a data breach. Here’s how it works:
+share-by-email-message = Hi, { $markup }{-brand-name} has a free service where you can check to see if you’ve been part of a data breach. Here’s how it works:
 
 share-by-email-step-1 = 1. Go to { $link } and search your email.
 share-by-email-step-2 = 2. See if your online accounts have been exposed in a data breach.
--- a/locales/en/email-strings.ftl
+++ b/locales/en/email-strings.ftl
@@ -1,5 +1,7 @@
 # Firefox Monitor is a product name and should not be translated.
 -product-name = Firefox Monitor
+# Firefox is a brand name and should not be translated.
+-brand-name = Firefox
 
 click-to-verify = 
   Select the Verify My Email button within 24 hours to confirm your Firefox Monitor account. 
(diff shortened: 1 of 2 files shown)


locales/en/email-strings.ftl
```diff
@@ -75,3 +77,13 @@ report-pwt-summary-4 =
 
 # A link to legal information about mozilla products.
 legal = Legal
+
+# Share Firefox Monitor by email subject line
+share-by-email-subject = See if you’ve been part of a data breach.
+
+# Share Firefox Monitor by email message. Followed by steps about how Firefox Monitor works. 
+share-by-email-message = Hi, { $markup }{-brand-name} has a free service where you can check to see if you've been part of a data breach. Here's how it works:
```
flodolo: What's `$markup` here?

This string is using a straight single quote instead of an apostrophe, you should try to use the latter consistently.

lesleyjanenorton: Thanks for catching the quote, I will update to an apostrophe. 
`{ $markup  }` will be `<br>`.

lesleyjanenorton:
@flodolo Regarding timelines, I believe the plan is to deploy the features that will use these new strings at the end of next week (if possible), or at the beginning of the week following All Hands.

locales/en/email-strings.ftl
flodolo: And we can't have the actual ´<br>`?

lesleyjanenorton: @flodolo these strings will be used together in mailto link email where the line break just looks like "\n" before it's encoded. I'm worried the "\n" would confuse.

pdehaan: What about something like adding a `-newline` variable to the top of the .ftl files:
```ini
-newline = "\n"
```

And then changing this string to:
```ini
share-by-email-message = Hi,{-newline}{-brand-name} has a free service...
```

It may remove some of the ambiguity around what we could possibly be injecting in `{ $markup }` (which I was assuming was some name/email address the first time I saw it).

flodolo: I think the `-newline` would be even more confusing. Have we discarded a multiline string? Would it work in your code? Not sure if @stasm has other ideas for this use case.

```
 share-by-email-message = 
 	Hi, 
	{-brand-name} has a free service where you can check to see if you've been part of a data breach. Here’s how it works:
```

lesleyjanenorton: @flodolo as far as I know, line-breaks like this in the ftl won't carry over to the template.

stasm: Fluent preserves newlines in text. If there's a line break after `Hi,`, the `href` attribute of the link will end up being written over multiple lines in the HTML source. As @lesleyjanenorton says, the line break will then not carry over to the email body.

I tested this quickly and it looks like the only way for the line break to carry over is when it's written as `%0A` or `%0D%0A`. That's `LF` and `CRLF`, respectively, encoded using hexadecimal escape sequences (same as the result of `escape()` or `encodeURI()`). Interestingly, [RFC 2368](https://tools.ietf.org/html/rfc2368#page-3), which specifies the `mailto:` URL scheme, mentions `%0D%0A` exclusively.

A term (`-newline`) or a variable (`$newline`) could work here, but I agree with @flodolo: this is not what we usually use them for and it might end up being confusing. Some tools also require the translation to reference all terms and variables which are present in the original message. So using a term might have the unintended effect of forcing all translations to use it, even if they don't need it. (In the future, Fluent might offer a way to tell which terms are required and which ones are optional, which would help in cases like this one.)

I would suggest the following:

```properties
# {"%0D%0A"} is a line break encoded for the mailto: scheme.
share-by-email-message = 
    Hi,{"%0D%0A"}{-brand-name} has a free service where…
```

As a matter of fact, you could also just write `Hi,%0D%0A{-brand-name}` and it would work, too. I find the string literal, `{"%0D%0A"}`, more explicit and easier to spot, however. It highlights that there's something special about this part of the translation.

PR #618, opened 2018-11-21
Adds new breach image (HTHStudios.png)

lesleyjanenorton:
Image for newly released 'High Tail Hall' breach (HTHStudios.com).


pdehaan:
Ah, I totally missed this breach because I didn't see it included in the [RSS feed](https://feeds.feedburner.com/HaveIBeenPwnedLatestBreaches) that I've been following. But I guess maybe the RSS feed doesn't include `IsSensitive` breaches.

Do we get notified of new breaches so we know to add logos, etc? Or curious if we need something that will scrape the [/api/v2/breaches API](https://haveibeenpwned.com/api/v2/breaches) on a cron job and submit GitHub issues or something.

lesleyjanenorton:
@pdehaan No prior notice, just happened to catch an article (https://gizmodo.com/hackers-hit-adult-furry-website-exposing-hundreds-of-t-1830589098) in the breach-alerts Slack channel earlier today. I think we occasionally get a heads up if an about-to-be-released breach has affected a particularly large number of people, but we don't find out the name of the company/organization until everyone else does. Regarding future logos: I heard today that breach images going forward will be PNGs. The major grief I have encountered whilst endeavoring to set up an automated system for breach logos has revolved around SVGs so I think that - if they really are going to be PNGs - there shouldn't be any trouble getting an automated image-getting script up and running. 🎉

pdehaan:
Fair enough... I managed to hack https://github.com/pdehaan/blurts-breaches-diff which will display a possibly useless JSON diff of a locally cached "./breaches.json" with a version that it scrapes from the https://haveibeenpwned.com/api/v2/breaches endpoint.

So in theory it should work for telling us if a new breach was added (assuming we run this regularly, or something) or if something mysteriously gets retired. But it may also tell us if any of the existing data changes (although in my local testing w/ hacking the breaches.json, the output is pretty much useless if only a single DataClass or other key is modified since the diff doesn't give you enough context in that case.

Or maybe a local "breaches.json" cache file is silly and it would be better to try getting the full breach data from the Blurts-Server directly so we can compare our version versus the latest HIBP data.

PR #625, opened 2018-11-26
Serve email-l10n routes on env=dev only. 

### LINKED ISSUE #612: Only serve /email-l10n routes on Heroku?
Not sure if these routes should exist on stage/prod (and only live on Heroku):

https://monitor.firefox.com/email-l10n/?partial=report&type=singleBreach

And maybe this should be wrapped in a `if (app.get("env") === "dev") {...}` check:

https://github.com/mozilla/blurts-server/blob/6ce8007c1571d174978cc609055ffaa4d06d29ac/server.js#L158

lesleyjanenorton:
I can't think of any reason these really need to be on stage/production.
Fixes #612

### DIFF
--- a/server.js
+++ b/server.js
@@ -155,7 +155,7 @@ if (AppConstants.FXA_ENABLED) {
 app.use("/scan", ScanRoutes);
 app.use("/ses", SesRoutes);
 app.use("/user", UserRoutes);
-if (app.get("env") === "dev" || AppConstants.NODE_ENV === "heroku") {
+if (["dev", "heroku"].includes(AppConstants.NODE_ENV)) {
   app.use("/email-l10n", EmailL10nRoutes);
 }
 app.use("/", HomeRoutes);
--- a/server.js
+++ b/server.js
@@ -155,7 +155,7 @@ if (AppConstants.FXA_ENABLED) {
 app.use("/scan", ScanRoutes);
 app.use("/ses", SesRoutes);
 app.use("/user", UserRoutes);
-if (app.get("env") === "dev") {
+if (app.get("env") === "dev" || AppConstants.NODE_ENV === "heroku") {
   app.use("/email-l10n", EmailL10nRoutes);
 }
 app.use("/", HomeRoutes);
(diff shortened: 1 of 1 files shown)


server.js
```diff
@@ -155,7 +155,9 @@ if (AppConstants.FXA_ENABLED) {
 app.use("/scan", ScanRoutes);
 app.use("/ses", SesRoutes);
 app.use("/user", UserRoutes);
-app.use("/email-l10n", EmailL10nRoutes);
+if (app.get("env") === "dev") {
```
groovecoder: On heroku we have:
```
NODE_ENV:                            heroku
```
So, this should be `if (app.get("env") === "heroku") {`

In fact, we could move this up into the existing `heroku` check at https://github.com/mozilla/blurts-server/blob/master/server.js#L73-L82

lesleyjanenorton: @groovecoder updated. I wasn't able to view locally with just ==="heroku" but maybe I am goofing something?

pdehaan: Would you need to run `$ NODE_ENV=heroku npm start`?

server.js
```diff
@@ -155,7 +155,9 @@ if (AppConstants.FXA_ENABLED) {
 app.use("/scan", ScanRoutes);
 app.use("/ses", SesRoutes);
 app.use("/user", UserRoutes);
-app.use("/email-l10n", EmailL10nRoutes);
+if (app.get("env") === "dev" || AppConstants.NODE_ENV === "heroku") {
```
groovecoder: Hmm ... `app.get("env")` _should_ be the same as `AppConstants.NODE_ENV`, but can we keep this consistent? So:
```
if (["dev", "heroku"].includes(AppConstants.NODE_ENV)) {
```

server.js
lesleyjanenorton: Works! But I think anyone trying to view the email-l10n routes locally will have to have a self-signed cert for their localhost or disable the https redirect.

server.js
lesleyjanenorton: Works! I think it will mean that anyone poking around on email-l10n routes locally will either need a self-signed cert for their localhost or will just need to disable the https redirect.

lesleyjanenorton:
@groovecoder Were there additional changes needed here or are we good to merge?

PR #629, opened 2018-11-28
Update confirmation page & condense subpages.

### LINKED ISSUE #628: Update confirmation page with new share by email and social media sharing options.
Our existing confirmation page needs to be updated to use latest copy and match this new spec (https://e-pang.github.io/firefoxmonitor/#artboard14). 

-Sharing option buttons should send unique GA pings on click.
-Email client links should open in either gmail.com yahoo.com, outlook.com, or the user's default mail client if "Other" is clicked. 
-The email should be prepopulated with translated strings.

lesleyjanenorton:
The confirmation, unsubscribe, unsubscribe survey, and error pages now all use the same layout (since the confirmation page redesign). This creates a single view for all of these pages, uses partials for page specific elements, and updates the confirmation page to the new spec. Routes updated where necessary.

New to the confirmation page:
- Option added for the user to share a link to Firefox Monitor by email, with options for opening the prepopulated email in the user's default mail client or in Gmail, Yahoo, or Outlook email web apps.
- GA pings sent for each of these on click

Fixes #628


controllers/user.js
```diff
@@ -54,28 +82,27 @@ async function verify(req, res) {
...
   await EmailUtils.sendEmail(
     verifiedEmailHash.email,
     req.fluentFormat("user-verify-email-report-subject"),
     "default_email",
     {
-      supportedLocales,
+      supportedLocales: EmailUtils.getSupportedLocales(req),
       email: verifiedEmailHash.email,
-      date: HBSHelpers.prettyDate(supportedLocales, new Date()),
+      date: HBSHelpers.prettyDate(req.supportedLocales, new Date()),
```
pdehaan: Is `EmailUtils.getSupportedLocales(req)` (L90) different than the `req.supportedLocales` (L92)?

It looks like `req.supportedLocales` gets set in the `pickLanguage` middleware via

https://github.com/mozilla/blurts-server/blob/61d2a44d7ab55abd3325092276b8681143e71061/middleware.js#L27-L49

groovecoder: I think there was a subtle difference, but I've forgotten what it was. It may just be dupe logic that can be removed.

controllers/home.js
```diff
@@ -26,7 +26,9 @@ function home(req, res) {
 
 function notFound(req, res) {
   res.status(404);
-  res.render("error", { message: req.fluentFormat("home-not-found") });
+  res.render("subpage", {
+    headline: "Error",
```
groovecoder: I ses this was a bug in the old `error.hbs` template too, but this needs to be wrapped in `req.fluentFormat("error")` ?

controllers/user.js
```diff
@@ -39,6 +39,34 @@ async function add(req, res) {
...
 }
 
+function getShareByEmail(req) {
+
+  const subject = encodeURIComponent(req.fluentFormat("share-by-email-subject"));
+  const body = encodeURIComponent(req.fluentFormat("share-by-email-message"));
+
+  return {
+    "gmail" : {
+      client: "Gmail",
```
groovecoder: Are we sure these email client names don't need translation?

controllers/user.js
```diff
@@ -120,8 +150,11 @@ function getUnsubSurvey(req, res) {
   if(!req.session.unsub) {
     throw new FluentError("error-not-subscribed");
   }
-  res.render("unsubscribe_survey", {
+  res.render("subpage", {
   title: req.fluentFormat("user-unsubscribe-survey-title"),
+  headline: req.fluentFormat("unsub-survey-headline"),
+  subhead: req.fluentFormat("unsub-survey-blurb"),
+  whichPartial: "subpages/unsubscribe_survey",
```
groovecoder: nit (it was broken in the previous code too): indent these properties.

middleware.js
```diff
@@ -84,7 +84,10 @@ function clientErrorHandler (err, req, res, next) {
 
 function errorHandler (err, req, res, next) {
   res.status(500);
-  res.render("error", { message: err.message });
+  res.render("subpage", {
+    headline: "Error",
```
groovecoder: See https://github.com/mozilla/blurts-server/pull/629/files#r237335082

controllers/user.js
```diff
@@ -54,28 +82,27 @@ async function verify(req, res) {
...
-      buttonValue,
-      whichView,
+      buttonValue: req.fluentFormat("report-scan-another-email"),
+      whichView: "email_partials/report",
     }
   );
 
-  res.render("confirm", {
+  res.render("subpage", {
+    headline: req.fluentFormat("confirmation-headline"),
```
groovecoder: Looks like this should be `confirmed` not `confirmation-headline` ?

controllers/user.js
```diff
@@ -54,28 +82,27 @@ async function verify(req, res) {
...
+    subhead: req.fluentFormat("confirmation-blurb"),
```
groovecoder: `confirmed-blurb` not `confirmation-blurb` ?

controllers/user.js
groovecoder: Nvm. Just needed to rebase on master.

controllers/user.js
groovecoder: Rebase fixed this.

controllers/user.js
```diff
@@ -39,6 +39,34 @@ async function add(req, res) {
...
+    },
+    "yahoo" : {
+      client: "Yahoo",
+      class: "yahoo",
+      href: `https://compose.mail.yahoo.com/?subject=${subject}&body=${body}`,
+    },
+    "outlook" : {
+      client: "Outlook",
+      class: "outlook",
+      href: `https://outlook.live.com/owa/?path=/mail/action/compose&to=service%40domain.com&subject=${subject}&body=${body}`,
```
groovecoder: Should drop the `to=service%40domain.com` url param so the compose message isn't pre-filled with this in the `To:` field.

views/partials/subpages/unsubscribe_survey.hbs
```diff
@@ -0,0 +1,20 @@
+<section id="unsubscribe-survey" class="unsubscribe-survey section-wrapper drop-shadow">
+  <div class="whole">
+    <form action="" method="post" id="unsubscribe-survey-form" class="form-group">
+      <span class="secondary-title" tabindex="1">{{{fluentFormat req.supportedLocales "unsub-survey-form-label"}}}</span>
+      {{#each UNSUB_REASONS}}
+        <div class="radio-button-group" tabindex="1">
+          <input id="reason{{ id_number }}" type="radio" name="unsubscribe-reason" data-analytics-label="{{ value }}" value="{{getStringID ../req.supportedLocales "unsub-reason-" id_number}}" />
+          <span class="radio-dot"></span>
+          <label for="reason{{ id_number }}">{{{getStringID ../req.supportedLocales "unsub-reason-" id_number}}}</label>
```
groovecoder: Regression: this re-introduced the bug where the user has to click on the `<label>` element - i.e., clicking on the `<span class="radio-dot">` element doesn't work anymore? Firefox Nightly 65.0a1 (2018-11-24)

controllers/user.js
lesleyjanenorton: Dupe. Except for `res.vary`. Could the lack of `res.vary` in `EmailUtils.getSupportedLocales` have caused #630 ? Will get rid of it.

controllers/user.js
lesleyjanenorton: Pretty sure, but to be totally sure...
@flodolo should Gmail, Yahoo, and Outlook be translated?

flodolo: I'd say no, but can you give me the context in which they are used?

lesleyjanenorton: @flodolo They appear in a list of email clients a user can select to open an email in. Currently, only "Email" and "Other" get translated.
screen shot 2018-11-29 at 9 16 24 am

controllers/user.js
groovecoder: Ooo, that's a good insight and idea. It's worth investigating for sure.

controllers/user.js
flodolo: OK, then I would say to keep them not translatable (I was afraid they would appear in the middle of a sentence), and see if anyone complains about it.

lesleyjanenorton:
Anything else that needs work here or can we get this merged in so SoftVision can do their thing?

locales/en/app.ftl
```diff
@@ -38,6 +38,7 @@ user-add-invalid-email = Invalid Email
 user-add-email-verify-subject = Verify your subscription to {-product-name}.
 user-add-title = {-product-name} : Confirm Email
 
+error = Error
```
flodolo: Is this a title? Maybe it would be nice to have a different ID, e.g. error-title, to make the context clearer and avoid the temptation to reuse it in a different context

groovecoder:
To close the loop on the integrity of this ... @lesleyjanenorton and I discussed this PR with @jgmize on the way to, and at, all-hands, and @jgmize offered to merge it while I was unavailable so that SoftVision could start the QA cycle on heroku.

PR #646, opened 2018-11-29
Load test breaches endpoint

groovecoder:
Bad branch name, but this updates our load tests to more closely match real site traffic that we see post-Launch.


groovecoder:
Closing and deleting branch as we're serving breach data from Remote Settings now - not from Monitor.

PR #647, opened 2018-11-29
migration to add index to email column

groovecoder:


groovecoder:
Approving, merging, and deploying myself so we can fix GCP.

PR #663, opened 2018-12-11
Additional breach filtering / Hide non-website breaches everywhere. 

lesleyjanenorton:
There was a lot of confusion about the "You've Been Scraped" breach alert we sent out on 12/6. Earlier today a decision was made to hide non-website breaches until new content has been created to explain them.
This hides breaches in website scan results, email reports, and prevents the sending of new breach alerts when  `breach.IsFabricated`, `breach.IsSpamList`, and when `breach.Domain === ""`.

### DIFF
--- a/controllers/hibp.js
+++ b/controllers/hibp.js
@@ -36,6 +36,13 @@ async function notify (req, res) {
       throw new Error("Unrecognized breach: " + reqBreachName);
     }
   }
+  
+  if (breachAlert.IsSpamList || breachAlert.Domain === "" || breachAlert.IsFabricated) {
+    log.info(`${breachAlert.Name} is fabricated, a spam list, or not associated with a website. \n Breach Alert not sent.`);
+    return res.json(
+      {info: "Breach loaded into database."}
+    );
+  }
 
   const hashes = req.body.hashSuffixes.map(suffix=>reqHashPrefix + suffix.toLowerCase());
   const subscribers = await DB.getSubscribersByHashes(hashes);
@@ -49,7 +56,6 @@ async function notify (req, res) {
     log.info("notify", {subscriber});
 
     const email = subscriber.email;
-    // need to create unsubscribe button for these
     const requestedLanguage = subscriber.signup_language ? acceptedLanguages(subscriber.signup_language) : "";
     const supportedLocales = negotiateLanguages(
       requestedLanguage,
--- a/controllers/hibp.js
+++ b/controllers/hibp.js
@@ -37,9 +37,9 @@ async function notify (req, res) {
     }
   }
 
-  if (breachAlert.IsSpamList || breachAlert.Domain === "" || breachAlert.IsFabricated) {
+  if (breachAlert.IsSpamList || breachAlert.Domain === "" || breachAlert.IsFabricated || !breachAlert.IsVerified) {
     log.info(`${breachAlert.Name} is fabricated, a spam list, or not associated with a website. \n Breach Alert not sent.`);
-    return res.json(
+    return res.status(200).json(
       {info: "Breach loaded into database."}
     );
   }
--- a/controllers/hibp.js
+++ b/controllers/hibp.js
@@ -36,7 +36,7 @@ async function notify (req, res) {
       throw new Error("Unrecognized breach: " + reqBreachName);
     }
   }
-  
+
   if (breachAlert.IsSpamList || breachAlert.Domain === "" || breachAlert.IsFabricated) {
     log.info(`${breachAlert.Name} is fabricated, a spam list, or not associated with a website. \n Breach Alert not sent.`);
     return res.json(
--- a/controllers/hibp.js
+++ b/controllers/hibp.js
@@ -40,7 +40,7 @@ async function notify (req, res) {
   if (breachAlert.IsSpamList || breachAlert.Domain === "" || breachAlert.IsFabricated || !breachAlert.IsVerified) {
     log.info(`${breachAlert.Name} is fabricated, a spam list, not associated with a website, or unverified. \n Breach Alert not sent.`);
     return res.status(200).json(
-      {info: "Breach loaded into database."}
+      {info: "Breach loaded into database. Subscribers not notified."}
     );
   }
 
--- a/controllers/hibp.js
+++ b/controllers/hibp.js
@@ -38,7 +38,7 @@ async function notify (req, res) {
   }
 
   if (breachAlert.IsSpamList || breachAlert.Domain === "" || breachAlert.IsFabricated || !breachAlert.IsVerified) {
-    log.info(`${breachAlert.Name} is fabricated, a spam list, or not associated with a website. \n Breach Alert not sent.`);
+    log.info(`${breachAlert.Name} is fabricated, a spam list, not associated with a website, or unverified. \n Breach Alert not sent.`);
     return res.status(200).json(
       {info: "Breach loaded into database."}
     );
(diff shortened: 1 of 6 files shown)


controllers/hibp.js
```diff
@@ -37,6 +37,13 @@ async function notify (req, res) {
     }
   }
 
+  if (breachAlert.IsSpamList || breachAlert.Domain === "" || breachAlert.IsFabricated) {
```
pdehaan: Q: Should we move this logic into some helper?
It looks like we're using it here, and a similar (but inverted) version below in /hibp.js:L104-107.

hibp.js
```diff
@@ -101,7 +101,9 @@ const HIBP = {
   async getUnsafeBreachesForEmail(sha1, allBreaches) {
     const allFoundBreaches = await this.getBreachesForEmail(sha1, allBreaches, true);
     return allFoundBreaches.filter(
-      breach => !breach.IsSpamList
+      breach => !breach.IsSpamList &&
+                !breach.IsFabricated &&
```
pdehaan: The logic here changed to now exclude any `.IsFabricated` breaches, is that expected?

hibp.js
```diff
@@ -139,10 +141,11 @@ const HIBP = {
 
   filterOutUnsafeBreaches(breaches) {
     return breaches.filter(
-      breach => breach.IsVerified &&
-                !breach.IsRetired &&
+      breach => !breach.IsRetired &&
                 !breach.IsSensitive &&
-                !breach.IsSpamList
+                !breach.IsSpamList &&
+                !breach.IsFabricated &&
```
pdehaan: Similarly, this logic changed to add the `!breach.IsFabricated` check, but removed the `.IsVerified` check.

hibp.js
lesleyjanenorton: Expected, but potentially unnecessary. I think there is only one "breach.IsFabricated" and it gets filtered out by `!breach.IsVerified` (added back in [24450c4](https://github.com/mozilla/blurts-server/pull/663/commits/24450c4)).

controllers/hibp.js
```diff
@@ -37,9 +37,9 @@ async function notify (req, res) {
     }
   }
 
-  if (breachAlert.IsSpamList || breachAlert.Domain === "" || breachAlert.IsFabricated) {
+  if (breachAlert.IsSpamList || breachAlert.Domain === "" || breachAlert.IsFabricated || !breachAlert.IsVerified) {
     log.info(`${breachAlert.Name} is fabricated, a spam list, or not associated with a website. \n Breach Alert not sent.`);
-    return res.json(
+    return res.status(200).json(
```
lesleyjanenorton: @groovecoder or @nhnt11 Sorry, I know you guys are on PTO but do we need to give Troy a heads up that he might be seeing a new response from us?

hibp.js
pdehaan: I think I lack a lot of context/knowledge to be the best reviewer... Perhaps we can summon @groovecoder to give a set of 👀 as well.

I think my confusion is around how we're loading and filtering breaches and why the `.IsSpamList` and `.IsFabricated`, and other flags logic lives in multiple places.

Can/should we do that filtering somewhere when we cron-import the breaches list in `loadBreachesIntoApp()`? If we know we'll never show spam list results, it may make sense to just filter them at import before shimming them into `app.locals.breaches` array. Then the only checks we may need to do are checking whether the breach has `.IsSensitive` set or not (depending on the context of web results vs emails). It may simplify the logic somewhat.
But like I say, maybe I'm missing something obvious and we're already doing that somewhere and I should simmer down.

https://github.com/mozilla/blurts-server/blob/e3cecb9a31065909ab20b995f0647bd4fd3395db/hibp.js#L81-L99

hibp.js
lesleyjanenorton: Added .IsVerified check back in.

controllers/hibp.js
lesleyjanenorton: Yes, that would certainly be cleaner. Did a bit of DRYing up in HIBP in 24450c4 with this in mind but it doesn't help us out any here.

lesleyjanenorton:
@pdehaan Made some additional updates. If you could give them another look-through, that would be wonderful! Thank you!

controllers/hibp.js
```diff
@@ -37,6 +37,13 @@ async function notify (req, res) {
     }
   }
 
+  if (breachAlert.IsSpamList || breachAlert.Domain === "" || breachAlert.IsFabricated || !breachAlert.IsVerified) {
+    log.info(`${breachAlert.Name} is fabricated, a spam list, or not associated with a website. \n Breach Alert not sent.`);
```
pdehaan: "or is not verified" (but this is just a `log.info()` that goes into the ether and not anything user facing...

hibp.js
pdehaan: Cool, I think this makes sense. We lost the `!breach.IsSensitive` check here, but I see it doing the filter above on L127 (if we aren't checking `includeSensitive` branch in `getBreachesForEmail()`).

hibp.js
```diff
@@ -126,23 +120,27 @@ const HIBP = {
       }
     }
 
-    if (includeUnsafe) {
-      return foundBreaches;
+    if (includeSensitive) {
+      return this.filterBreaches(foundBreaches);
     }
-    return this.filterOutUnsafeBreaches(foundBreaches);
+    return this.filterBreaches(foundBreaches).filter(
```
pdehaan: Both of these branches call `this.filterBreaches()`. Not sure if it makes sense to move this up around L118 (where we populate `foundBreaches` initially), but putting it all in a single command is probably a bit too aggressive:

```js
foundBreaches = this.filterBreaches(allBreaches.filter(breach => breachedAccount.websites.includes(breach.Name)));

// vs

foundBreaches = allBreaches.filter(breach => breachedAccount.websites.includes(breach.Name));
foundBreaches = this.filterBreaches(foundBreaches);
```

Which would reduce this code to something like:

```js
if (includeSensitive) {
  return foundBreaches;
}
return foundBreaches.filter(breach => !breach.IsSensitive);
```

I guess the alternative would be to pass the `includeSensitive` argument to the `this.filterBreaches()` method and do the additional check there, which could reduce this `if..else` code to the following:

```js
return this.filterBreaches(foundBreaches, includeSensitive);
```

But all these are pretty nano-nits, and pointless code juggling. Feel free to merge whenever.

hibp.js
lesleyjanenorton: Hey, sorry for the lack of context here. So just for background, last week we sent out an alert for the "You've Been Scraped" breach. Since "You've Been Scraped" isn't associated with a website, it caused a lot of confusion and resulted in a number of users replying to the alert wondering wth.

While content is off creating new copy to address the non-website breaches (spam lists included) of the future, we want to temporarily disable alerts for them and hide them everywhere except the "?breach=" pages. Once we have new content, these breach categories will be reintroduced but no word yet as to whether we'll be sending alerts for them, disclosing them in website scan results, or restricting them to email reports.

We shouldn't have been sending alerts for unverified or fabricated breaches to begin with, and they could be filtered out in loadBreachesIntoApp() except for the fact that we <i>are</i> still showing them on the "/?breach=" pages. I'm going to find out if this is still desired behavior, given that the doorhanger doesn't appear on sites with unverified or fabricated breaches. If it's not, we'll get rid of them and any other unshowables in loadBreachesIntoApp(), delete their logos, and forget all about them.

controllers/hibp.js
nhnt11: We probably should. Also, if the API calls for sending back the number of notified subscribers, we should send a zero value to indicate we didn't notify anyone, no?

lesleyjanenorton: Great point, will update.

PR #669, opened 2018-12-20
start FxA fixes for new templates

groovecoder:
We still need to ...

* [x] Change our OAuth client to request `profile` scope and not just `profile:email` (needs FxA team to approve)
  * [x] Store the full `profile` JSON in our DB
* [x] Include `access_type=offline` in our query params to get a long-life refresh token
  * [x] Store the refresh token in our DB

### DIFF
--- a/db/DB.js
+++ b/db/DB.js
@@ -62,6 +62,7 @@ const DB = {
     return verifiedSubscriber[0];
   },
 
+  // TODO: refactor into an upsert? https://jaketrent.com/post/upsert-knexjs/
   // Used internally, ideally should not be called by consumers.
   async _getSha1EntryAndDo(sha1, aFoundCallback, aNotFoundCallback) {
     const existingEntries = await knex("subscribers")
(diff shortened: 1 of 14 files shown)


groovecoder:
Set `FXA_ENABLED=1` and this should fix the errors when being returned back to Monitor from FxA.

But, I don't see a report email in my Inbox, so there may still be a bug to work out.

groovecoder:
Example of full FxA profile JSON I was able to request and store:
```
{"email":"lcrouch@mozilla.com","locale":"en-US,en;q=0.5","amrValues":["pwd","email"],"twoFactorAuthentication":false,"uid":"<redacted>","avatar":"https://stable.dev.lcip.org/profile/a/4672b0a97c98ed5728a8b442b5b08c5b","avatarDefault":false,"displayName":"Luke"}
```
To display the profile picture, we will probably have to add `accounts.firefox.com` to our CSP.

db/migrations/20181227100332_add_fxa_columns.js
```diff
@@ -0,0 +1,15 @@
+"use strict";
+
+exports.up = function(knex, Promise) {
+  return knex.schema.table("subscribers", table => {
+    table.string("fxa_refresh_token");
+    table.json("fxa_profile_json");
```
jgmize: You should be able to use `table.jsonb` here, as https://github.com/tgriesser/knex/issues/991 was fixed in https://github.com/tgriesser/knex/pull/1044

groovecoder: Updated to `jsonb`

jgmize: While this should work in 9.4, https://github.com/crowdAI/crowdai/issues/68#issuecomment-285289202 needed to switch to Postgres 9.6 (and ubuntu trusty) to get jsonb working on travisci.

groovecoder:
Got Travis CI fixed. @pdehaan - did you want to give this a once-over? I'm feeling pretty good about it now.

db/DB.js
```diff
@@ -101,18 +101,40 @@ const DB = {
...
+   * @param {string} fxaRefreshToken from Firefox Account Oauth
+   * @param {string} fxaProfileData from Firefox Account
+   * @returns {object} subscriber knex object added to DB
+   */
+  async addSubscriber(email, fxaRefreshToken=null, fxaProfileData=null) {
     const emailHash = await this._addEmailHash(getSha1(email), email, true);
-    const verifiedSubscriber = await this._verifySubscriber(emailHash);
-    return verifiedSubscriber[0];
+    const verified = await this._verifySubscriber(emailHash);
+    const verifiedSubscriber = verified[0];
```
pdehaan: I realize this is the same as the old code behavior, but should we check that `verified[0]` (or `verifiedSubscriber`) is non-null before using it?

groovecoder: Good point. I'll add a check, and I'll actually add a low-level `try-catch` block around the subscription to basket since we may get errors beyond our control there, and we don't need/want to show those low-level errors to the user.

locales/en/app.ftl
```diff
@@ -19,6 +19,7 @@ about-firefox-alerts = About Firefox Alerts
 give-feedback = Give Feedback
 terms-and-privacy = Terms and Privacy
 
+error-could-not-add-email = Could not add email to database.
```
flodolo: email -> email address?

groovecoder: Updated.

PR #673, opened 2018-12-31
Download logos from HIBP.

lesleyjanenorton:
- Checks for corresponding logo when a new breach is loaded in.
- Downloads logo from HIBP if necessary.
- Uses imagemin & imagemin-pngquant to compress images if necessary.
- Sets the `newBreach.LogoPath` to `missing-logo-icon.png` if something goes wrong during the download.
- Updates breach alert template to use LogoPath instead of `missing-logo-icon.png`.
- Adds `missing-logo-icon.png` to img/logos folder.


breach-logos.js
```diff
@@ -0,0 +1,86 @@
...
+    stream.on("response", res => {
+
+        // Check response and error out on certain conditions.
+        if (res.headers["content-type"] !== "image/png") { /* eslint-disable quotes */
+          throw new Error(`response.headers["content-type"] !== "image/png"`); /* eslint-enable quotes */
+        }
+        if (res.statusCode !== 200) {
+          throw new Error(`Status: ${res.statusCode}`);
+        }
+        if (parseInt(res.headers["content-length"]) > 1000000) {
```
lesleyjanenorton: This number was a bit arbitrary. Not sure what we think a reasonable cut off figure is.

breach-logos.js
```diff
@@ -0,0 +1,86 @@
...
+    const localSrc = `${this.localDir}${fileName}`;
+    let fileIsBig = false;
+    let logoData = [];
+
+    const stream = await got.stream(remoteSrc, {
+      retries: 3,
+      timeout: {
+        connect: 1000,
+        socket: 1000,
+        request: 5000,
```
groovecoder: Are these suggested values somewhere? 1s connect timeout seems kinda low.

breach-logos.js
groovecoder: Is this 1MB? Do we know the average or median HIBP image size?

lesleyjanenorton: Anecdotally, they're less than 10kb. There's really no reason we should expect to see an image this big come floating across but just need to decide where to draw the line in the event of an unexpected-huge-file-occurence.

lesleyjanenorton: Sounds good to me!

controllers/hibp.js
```diff
@@ -36,6 +37,17 @@ async function notify (req, res) {
...
   }
+  // If we don't have a logo of the breach, get the logo.
+  if (!fs.existsSync(`./public/img/logos/${breachAlert.LogoPath}`)) {
+    try {
+      await BreachLogos.downloadLogo(breachAlert.LogoPath);
+    } catch(err) {
+      // Set LogoPath to missing-logo-icon.png if download fails
+      // LogoPath will reset when the server restarts
+      breachAlert.LogoPath = "missing-logo-icon.png";
+      log.error(`Unable to download logo for ${breachAlert.LogoPath}.`);
```
pdehaan: Does someone read/monitor these logs so we know there was a logo fetch failure?

pdehaan: Oh, also, I think your `log.error()` won't be useful, because if I'm reading it correctly, it will _always_ say "Unable to download logo for missing-logo-icon.png" (vs using the `breachAlert.Name` or something more descriptive).

breach-logos.js
```diff
@@ -0,0 +1,86 @@
...
+        if (res.headers["content-type"] !== "image/png") { /* eslint-disable quotes */
+          throw new Error(`response.headers["content-type"] !== "image/png"`); /* eslint-enable quotes */
+        }
+        if (res.statusCode !== 200) {
+          throw new Error(`Status: ${res.statusCode}`);
+        }
+        if (parseInt(res.headers["content-length"]) > 1000000) {
+          throw new Error(`${fileName} is too large. Downloaded Aborted.`);
+        }
+        if (parseInt(res.headers["content-length"]) > 10000) {
```
pdehaan: Not sure if we want constants for the magic numbers here (and L53 above). I'm not sold either way, but this is basically just checking if the file size is around 9.75 KB, and if so, we should optimize the image.
Not sure if we should just always optimize the image, regardless of file size, or if that is too aggressive.

breach-logos.js
pdehaan: I actually looked into this once briefly.

So far our worst offending image I've found [from HIBP] was Parapa.png, at 231 KB:

```
464 -rw-r--r--    1 pdehaan  staff   231K 31 Dec 23:46 Parapa.png
```

But, if I view the image in the repo, it's a lovely optimized 38 KB;
```
 80 -rw-r--r--    1 pdehaan  staff    38K 31 Dec 23:48 Parapa.png
```

So I'm not sure what the best approach is... If the image is considered "large" at 500 KB, but it's almost arbitrary and we would really only care about the file size of the _optimized_ logo, and maybe error out if the minified file size is over 100 KB.

lesleyjanenorton: I distinctly remembered the panic level exclamations associated with the soul-crushing Parapa logo and used it to test this set up! Pulling the existing logo from HIBP and pushing it through png-quant resulted locally in an a 37kb end-product.

controllers/hibp.js
lesleyjanenorton: You're totally right. I'm also now wondering if any retries will hit this catch then look for "missing-logo-icon.png" from HIBP on subsequent requests?

breach-logos.js
lesleyjanenorton: I'll try this again with imagemin-pngquant and see how it goes with optimizing all images by default no matter their size. I encountered (possibly with imagemin-pngquant, possibly with one of the other set-ups I tried) two strange things when optimizing images that were already quite small: in some cases they turned into black squares, and in other cases they actually increased in file size.

controllers/hibp.js
groovecoder: These errors are thrown up to Sentry, which sends alerts. Do either of you get them? I can add you to the Sentry group if you like.

Also still need to update these `mozlog` calls to have the 2 expected arguments.

controllers/hibp.js
```diff
@@ -36,6 +37,17 @@ async function notify (req, res) {
       throw new Error("Unrecognized breach: " + reqBreachName);
     }
   }
+  // If we don't have a logo of the breach, get the logo.
+  if (!fs.existsSync(`./public/img/logos/${breachAlert.LogoPath}`)) {
```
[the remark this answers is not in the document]
groovecoder: After white-boarding, we're asking Ops to provision us some https://cloud.google.com/storage/ space for these images, and we'll use the GCP node module to write to it.

It may still be helpful to merge this branch though - the emails will start to include the logos.

controllers/hibp.js
pdehaan: @groovecoder I don't think I'm getting Sentry emails, but would like to be added if it means I'll see errors.

PR #676, opened 2019-01-03
 fix #675: store accept-language as signup_language during fxa oauth 

### LINKED ISSUE #675: When a user signs up with Firefox Account, their signup_language is blank
STR:

1. Set `FXA_ENABLED=1`
2. Sign up with Firefox Account
3. `psql blurts`
4. `SELECT signup_language FROM subscribers WHERE fxa_profile_json is not null`

Expected results:

The `Accept-Language` value from the browser - e.g., `en-US,en;q=0.5`

Actual results:

blank value 😢

groovecoder:


db/DB.js
```diff
@@ -78,7 +78,7 @@ const DB = {
   },
 
   // Used internally.
-  async _addEmailHash(sha1, email, verified = false) {
+  async _addEmailHash(sha1, email, signup_language, verified = false) {
```
pdehaan: Nit: not sure if this should be `signupLanguage` to be consistent w/ `async addSubscriber()` implementation below on L120.

groovecoder: This also made me look back at some kind of node ORM library. @pdehaan - do you know if any other Mozilla node projects are using https://bookshelfjs.org/ ?

PR #686, opened 2019-01-08
Catch and replace missing breach logos in templates.

lesleyjanenorton:
Replaces broken breach logos with `missing-logo-icon.png` so this doesn't happen in Safari.
screen shot 2019-01-08 at 10 00 35 am


public/js/monitor.js
```diff
@@ -427,7 +427,6 @@ function restoreInputs() {
   });
 }
 
-//prevents footer from covering stuff up
```
groovecoder: this line in paritcular.

PR #687, opened 2019-01-08
Throw error/404 on unknown breaches.

### LINKED ISSUE #649: Change breach landing page to show "Not Found"/404 for unkown breaches
https://monitor.firefox.com/?breach=List

Shows "Cannot read property 'IsSensitive' of undefined" but it should just show a 404.

lesleyjanenorton:
Fixes #649

### DIFF
--- a/controllers/home.js
+++ b/controllers/home.js
@@ -1,6 +1,6 @@
 "use strict";
 
-
+const { FluentError } = require("../locale-utils");
 
 function home(req, res) {
   let featuredBreach = null;
@@ -10,6 +10,9 @@ function home(req, res) {
   if (req.query.breach) {
     const reqBreachName = req.query.breach.toLowerCase();
     featuredBreach = req.app.locals.breaches.filter(breach => breach.Name.toLowerCase() === reqBreachName)[0];
+    if (!featuredBreach) {
+      throw new FluentError("home-not-found");
+    }
     scanFeaturedBreach = true;
     if (featuredBreach.IsSensitive) {
       breachIsSensitive = true;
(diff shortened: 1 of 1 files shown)


controllers/home.js
```diff
@@ -10,6 +10,9 @@ function home(req, res) {
   if (req.query.breach) {
     const reqBreachName = req.query.breach.toLowerCase();
     featuredBreach = req.app.locals.breaches.filter(breach => breach.Name.toLowerCase() === reqBreachName)[0];
```
pdehaan: Technically outside the scope of this PR, but looking at this code again, I'm curious if using the `.find()` API here may be slightly more appropriate than using `.filter()[0]`.

Where:

> The `find()` method returns the **value** of the **first element** in the array that satisfies the provided testing function. Otherwise `undefined` is returned.

vs

> The `filter()` method creates a new array with all elements that pass the test implemented by the provided function.

pdehaan:
And here's what it looks like locally for me for a non-existent `?breach=adobeX` breach:

mozilla_firefox

controllers/home.js
lesleyjanenorton: oh nice! This is way better. Pretty sure there are a few other places where this would also make more sense.

controllers/home.js
```diff
@@ -10,6 +10,9 @@ function home(req, res) {
...
+    if (!featuredBreach) {
+      throw new FluentError("home-not-found");
```
groovecoder: `throw` makes this return with `500` status code.

We should try to change this to do like the `notFound` route handler does: set `res.status(404)` and render the error subpage with the right message.

But I'm not 100% sure if that's easy, and this would be better than what we have now. So if we can't figure it out quickly, I'm okay to merge this and leave the `404` status code as a follow-up.

groovecoder: This isn't as simple as it should be, but I'm going to go ahead and merge so we can land the half-fix.

PR #692, opened 2019-01-10
Cherry picked automated images

groovecoder:


controllers/hibp.js
```diff
@@ -36,6 +36,17 @@ async function notify (req, res) {
...
     }
   }
+  // If we don't have a logo of the breach, get the logo.
+  if (!fs.existsSync(`./public/img/logos/${breachAlert.LogoPath}`)) {
+    try {
+      await HIBP.downloadLogo(breachAlert.LogoPath);
+    } catch(err) {
+      // Set LogoPath to missing-logo-icon.png if download fails
+      // LogoPath will reset when the server restarts
+      log.error(`Unable to download logo for ${breachAlert.LogoPath}.`);
```
pdehaan: I think this needs a string `op` as the first param and an object for the second param:

Something like:
```js
log.error("notify", {message: `Unable to download logo for ${breachAlert.LogoPath}`});
```

controllers/hibp.js
```diff
@@ -36,6 +36,17 @@ async function notify (req, res) {
...
+      breachAlert.LogoPath = "missing-logo-icon.png";
+    }
+  }
 
   if (breachAlert.IsSpamList || breachAlert.Domain === "" || breachAlert.IsFabricated || !breachAlert.IsVerified) {
     log.info(`${breachAlert.Name} is fabricated, a spam list, not associated with a website, or unverified. \n Breach Alert not sent.`);
```
pdehaan: Unrelated, but I filed #694 for this curious `log.info()`...

hibp.js
```diff
@@ -168,6 +176,69 @@ const HIBP = {
...
+      if (parseInt(res.headers["content-length"]) > 500000) {
+        throw new Error(`${fileName} is too large. Downloaded Aborted.`);
+      }
+      if (parseInt(res.headers["content-length"]) > 10000) {
+        fileIsBig = true;
+      } else {
+        // Conditions met, open up writeStream
+        const writeStream = fs.createWriteStream(localSrc);
+        res.pipe(writeStream);
+        writeStream.on("error", err => log.error(err));
```
pdehaan: I think this `log.error()` may need to be something like:

```js
log.error("download-logo.error", {err});
```

hibp.js
```diff
@@ -168,6 +176,69 @@ const HIBP = {
...
+        writeStream.on("finish", () => {
+          log.info(`${fileName} logo successfully downloaded.`);
```
pdehaan: Similarly, maybe:

```js
log.info("download-logo.finish", {message: `${fileName} logo successfully downloaded.`});
```

hibp.js
```diff
@@ -168,6 +176,69 @@ const HIBP = {
...
+    });
+    stream.on("end", async () => {
+      if(fileIsBig) {
+        logoData = Buffer.concat(logoData);
+        try {
+          logoData = await this.minifyLogoData(logoData);
+          await writeFile(localSrc, logoData, "binary");
+          log.info(`${fileName} logo successfully minified and downloaded.`);
+        } catch (err) {
+          log.error(`Error compressing and saving ${fileName}: ${err}`);
```
pdehaan:
```js
log.info("download-logo.end", {message: `${fileName} logo successfully minified and downloaded`});
...
log.error("download-logo.end-error", {message: `Error compressing and saving ${fileName}`, err});
```

hibp.js
```diff
@@ -168,6 +176,69 @@ const HIBP = {
...
+        }
+      }
+    });
+    stream.on("error", err => {
+      log.error(`Error downloading ${fileName} image: ${err.message}`);
```
pdehaan:
```js
log.error("download-logo.end-error", {message: `Error downloading ${fileName} image: ${err.message}`);
```

pdehaan:
I think I have a basic question about how this works, after looking at the PR again.

Troy discloses a new breach for example.com. Only our production site gets notified and the latest breach data gets loaded. Does this mean only the production site will then fetch the remote logos?

Hypothetically we now have 3 breaches that have been auto-loaded into monitor.firefox.com and we loaded+minified+stored the logos on the production server. But now what happens when we redeploy? Since those logos were never committed to source control, wouldn't they be lost with a new production deploy, unless somebody still commits them to GitHub? Which, if any of that guessing was correct, means that on production, the logos all work perfectly, until we do a new production deploy, at which point the `` handler kicks in, and shows the missing-logo.png style fallback logo for those 3 breaches that were auto-loaded between production deploys.

lesleyjanenorton:
@pdehaan I think the idea is to move all existing images to a Cloud Storage bucket (https://bugzilla.mozilla.org/show_bug.cgi?id=1518173), write future logos to the bucket, and then be done with having to commit new images to GitHub.

groovecoder:
Closing for now, as we're moving this work to our backlog until we have GCP-based storage all set up. Will keep the branch alive for future reference.

PR #696, opened 2019-01-10
Send analytics ping on HIBP clicks.

lesleyjanenorton:
Sends a ping anytime a user clicks on the HIBP attribution link. I'm sure we were collecting these clicks at one point but I think the analytics attributes were mistakenly left out during localization.


groovecoder:
Can confirm. 
screen shot 2019-01-14 at 11 10 16 am

PR #699, opened 2019-01-11
Add utm parameters to CTA buttons and unsubscribe links in emails.

### LINKED ISSUE #439: Google Analytics: /user/verify and /user/unsubscribe pageviews are not recorded
I suspect it's the code trying to remove the `token` and `hash` values here:
https://github.com/mozilla/blurts-server/blob/master/public/js/analytics_tracking_protection.js#L64-L68

### LINKED ISSUE #636: Add UTM tags to "Scan Another Email in Firefox Monitor" email CTA
We should add specific UTM parameters to the breach alert email about scanning another email as that is a clear signal of a user getting value and they want more value.

It would be interesting to find out how many loops the average user goes through over time where a loop would be defined as signing up for monitor, getting a breach alert email, and then setting up another email. I would imagine after you set up one email, get value that you likely probably set up other emails (if they have them) to scan.

### LINKED ISSUE #650: Add utm_content url param to unsubscribe links in emails
We include an unsubscribe link in the verification email. But, if a user doesn't verify their subscription within 24 hours, we delete their verification token, so that unsubscribe link sitting in their email inbox will send them to a "This email address is not subscribed to Firefox Monitor." error.

lesleyjanenorton:
- Adds utm parameters to the CTA buttons and unsubscribe links in emails.
- Moves and updates the function responsible for clipping out utm parameters from the url to prevent`hash=` and `token=` in unsubscribe routes from also getting clipped.

Fixes #636, fixes #650, closes #439

### DIFF
--- a/controllers/hibp.js
+++ b/controllers/hibp.js
@@ -47,6 +47,9 @@ async function notify (req, res) {
   const hashes = req.body.hashSuffixes.map(suffix=>reqHashPrefix + suffix.toLowerCase());
   const subscribers = await DB.getSubscribersByHashes(hashes);
 
+  const utmID = "breach-alert";
+  const buttonHref = EmailUtils.getScanAnotherEmailUrl(utmID);
+
 
   log.info("notification", { length: subscribers.length, breachAlertName: breachAlert.Name });
 
@@ -62,7 +65,6 @@ async function notify (req, res) {
       req.app.locals.AVAILABLE_LANGUAGES,
       {defaultLocale: "en"}
     );
-    const unsubscribeUrl = EmailUtils.unsubscribeUrl(subscriber);
 
     if (!notifiedSubscribers.includes(email)) {
       await EmailUtils.sendEmail(
@@ -73,10 +75,11 @@ async function notify (req, res) {
           email,
           supportedLocales,
           date: HBSHelpers.prettyDate(supportedLocales, new Date()),
-          unsubscribeUrl,
           breachAlert,
           SERVER_URL: req.app.locals.SERVER_URL,
           buttonValue: LocaleUtils.fluentFormat(supportedLocales, "report-scan-another-email"),
+          buttonHref: buttonHref,
+          unsubscribeUrl: EmailUtils.getUnsubscribeUrl(subscriber, utmID),
           whichView: "email_partials/report",
         },
       );
--- a/controllers/hibp.js
+++ b/controllers/hibp.js
@@ -48,7 +48,7 @@ async function notify (req, res) {
   const subscribers = await DB.getSubscribersByHashes(hashes);
 
   const utmID = "breach-alert";
-  const buttonHref = EmailUtils.getScanAnotherEmailUrl(utmID);
+  const scanAnotherEmailHref = EmailUtils.getScanAnotherEmailUrl(utmID);
 
 
   log.info("notification", { length: subscribers.length, breachAlertName: breachAlert.Name });
@@ -77,8 +77,7 @@ async function notify (req, res) {
           date: HBSHelpers.prettyDate(supportedLocales, new Date()),
           breachAlert,
           SERVER_URL: req.app.locals.SERVER_URL,
-          buttonValue: LocaleUtils.fluentFormat(supportedLocales, "report-scan-another-email"),
-          buttonHref: buttonHref,
+          scanAnotherEmailHref: scanAnotherEmailHref,
           unsubscribeUrl: EmailUtils.getUnsubscribeUrl(subscriber, utmID),
           whichView: "email_partials/report",
         },
--- a/email-utils.js
+++ b/email-utils.js
@@ -78,15 +78,26 @@ const EmailUtils = {
     });
   },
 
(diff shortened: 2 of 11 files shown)


email-utils.js
```diff
@@ -78,15 +78,26 @@ const EmailUtils = {
     });
   },
 
-  verifyUrl (subscriber) {
-    return `${AppConstants.SERVER_URL}/user/verify?token=${encodeURIComponent(subscriber.verification_token)}`;
+  utmParams(campaign, content) {
+    return `?utm_source=fx-monitor-emails&utm_medium=email&utm_campaign=${campaign}&utm_content=${content}`;
```
groovecoder: Because `utmParams` returns `?utm...`, it makes `getVerificationUrl` return http://localhost:6060/user/unsubscribe?token=null&hash=12eeb998fe0c7600b3a3aacf518aa760b1dbd775?utm_source=fx-monitor-emails&utm_medium=email&utm_campaign=unsubscribe&utm_content=report

Note that `?utm_source` is appended after `&hash=...` (and that `token=null`?)

The `utmParams` function should combine the utm params with any existing url params. In [the new FxA oauth `init` code](https://github.com/mozilla/blurts-server/blob/5932394e2ba15439a6f41dba01a1c4cf7a9b51f5/controllers/oauth.js#L38-L50), [we use `URLSearchParams` and `append` to do this](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams). Could do the same thing here.

controllers/hibp.js
```diff
@@ -73,10 +75,11 @@ async function notify (req, res) {
           email,
           supportedLocales,
           date: HBSHelpers.prettyDate(supportedLocales, new Date()),
-          unsubscribeUrl,
           breachAlert,
           SERVER_URL: req.app.locals.SERVER_URL,
           buttonValue: LocaleUtils.fluentFormat(supportedLocales, "report-scan-another-email"),
```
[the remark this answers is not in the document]
lesleyjanenorton: We can definitely move this into `email_button.hbs` and toggle the values based on which kind of email is being sent for now. The verification email will presumably be gone post-FxA integration anyway leaving "Scan another email.." as the single call to action button value.

controllers/hibp.js
```diff
@@ -73,10 +75,10 @@ async function notify (req, res) {
           email,
           supportedLocales,
           date: HBSHelpers.prettyDate(supportedLocales, new Date()),
-          unsubscribeUrl,
           breachAlert,
           SERVER_URL: req.app.locals.SERVER_URL,
-          buttonValue: LocaleUtils.fluentFormat(supportedLocales, "report-scan-another-email"),
+          scanAnotherEmailHref: scanAnotherEmailHref,
+          unsubscribeUrl: EmailUtils.getUnsubscribeUrl(subscriber, utmID),
```
groovecoder: There's some other kind of future refactor in here ... something where we put all URL-generators together and make them available with a handlebars helper function or something. But doesn't need to happen now.

tests/controllers/user.test.js
```diff
@@ -102,7 +102,7 @@ test("user unsubscribe GET request with valid token returns error", async () =>
   const validToken = "0e2cb147-2041-4e5b-8ca9-494e773b2cf0";
 
   // Set up mocks
-  const req = { fluentFormat: jest.fn(), query: { token: validToken } };
+  const req = { fluentFormat: jest.fn(), query: { token: validToken, hash: "testHash" } };
```
groovecoder: Maybe pedantic, but should we make the hash `ad9c69bcc69b3399775d2ddbe9b0b229369fca42` (i.e., sha1 of "testHash") so it's slightly more clear to someone reading the test that the hash value is a sha1 hash?

views/partials/email_partials/email_button.hbs
```diff
@@ -5,8 +5,12 @@
         <table>
           <tr>
             <td class="email-button-wrapper" align="center" bgcolor="#0060df" style="width: 100%; max-width: 530px; border-radius: 2px;">
-              <a class="email-button" href="{{ href }}" rel="noopener" style="text-align: center; text-decoration: none; color: #ffffff; padding-top: 12px; padding-right: 30px; padding-left: 30px; padding-bottom: 12px; border: 1px solid #0060df; display: inline-block; font-size: 13px;">
-               {{ buttonValue }}
+              <a class="email-button" href="{{ scanAnotherEmailHref }}{{ verificationHref }}" rel="noopener" style="text-align: center; text-decoration: none; color: #ffffff; padding-top: 12px; padding-right: 30px; padding-left: 30px; padding-bottom: 12px; border: 1px solid #0060df; display: inline-block; font-size: 13px;">
```
groovecoder: Whoa, this is a crazy trick that works for now, but it "smells" like it may not be future-proof. I.e., if we add an email that contains both `verificationHref` and `scanAnotherEmailHref`, they will conflict in here.

lesleyjanenorton: Very good point. Did some moving around to fix this.

PR #702, opened 2019-01-14
FxA Updates

lesleyjanenorton:
More FxA integration work with template updates hidden behind the FXA_ENABLED pref.


views/partials/fxa_enabled/fxa_navigation.hbs
```diff
@@ -0,0 +1,21 @@
...
+      <div class="fxa-menu">
+        <div class="fxa-links">
+          <a href="/">See Full Report</a>
+          <a href="https://accounts.firefox.com/">Manage Firefox Accounts</a>
+          <a href="/user/logout">Sign out</a>
+        </div>
+      </div>
+    </div> 
+  {{/if}}
+  {{#unless req.session.user }}
```
groovecoder: Just curious: What's the advantage of doing an `{{#unless req.session.user }}` block instead of using `{{else}}` block in the `{{#if req.session.user }}`

PR #707, opened 2019-01-17
bug 1517343 - add index to verified column

groovecoder:


groovecoder:
Merged my own PR like a tyrant so we can cut a release tag to deploy to stage & prod.

Issue #512, opened 2018-10-08
The "Firefox Monitor report" is sent as many times as the "Confirmed! You're Subscribed"  page is refreshed


obotisan:
[Affected versions]:
- Firefox 62.0
- Firefox 64.0a1

[Affected Platforms]:
- Windows 10 x64
- macOS 10.13
- Ubuntu 16.04 x64

[Steps to reproduce]:
1. Navigate to https://fx-breach-alerts.herokuapp.com/
2. Do the necessary steps in order for the "Confirmed! You're Subscribed" page to be displayed.
3. Refresh the page a few times and check the your email.

[Expected result]:
- The "Firefox Monitor report" is sent only once.

[Actual result]:
- The "Firefox Monitor report" is sent as many times as the "Confirmed! You're Subscribed"  page is refreshed.

[Regression]:
- I don't think this is a regression. I can reproduce this issue using Nightly from 2018-08-28. 
- I can reproduce the issue using https://blurts-server.stage.mozaws.net/, too. It reproduces after you close the "Confirm your account" pop-up.

groovecoder:
This is being fixed via #712

PR #716, opened 2019-01-24
fix #712: check if token is already verified before sending email

### LINKED ISSUE #712: User Says They're Getting Multiple Breach Reports Emails
multiplebreachreports

groovecoder:


groovecoder:
This also fixes https://github.com/mozilla/blurts-server/issues/512

PR #742, opened 2019-02-05
fix #738: FXA.revokeOAuthToken

### LINKED ISSUE #738: When a user unsubscribes from Monitor, revoke their OAuth token
When a user unsubscribes from Monitor, send an OAuth token revocation request to FxA.

https://github.com/mozilla/fxa-auth-server/blob/master/fxa-oauth-server/docs/api.md#post-v1destroy

groovecoder:


controllers/oauth.js
```diff
@@ -98,7 +97,7 @@ async function confirmed(req, res, next, client = FxAOAuthClient) {
         date: HBSHelpers.prettyDate(req.supportedLocales, new Date()),
         unsafeBreachesForEmail: unsafeBreachesForEmail,
         scanAnotherEmailHref: EmailUtils.getScanAnotherEmailUrl(utmID),
-        unsubscribeUrl: unsubscribeUrl,
+        unsubscribeUrl: EmailUtils.getUnsubscribeUrl(verifiedSubscriber, utmID),
```
groovecoder: Note: this addresses https://github.com/mozilla/blurts-server/issues/763 but doesn't add any content to the unsubscribe page to tell the user that unsubscribing from Monitor alerts will leave their Firefox Account intact.

pdehaan: > doesn't add any content to the unsubscribe page to tell the user that unsubscribing from Monitor alerts will leave their Firefox Account intact.

Do we need a new bug for that?

db/seeds/test_subscribers.js
```diff
@@ -12,8 +12,9 @@ exports.seed = function(knex) {
         {
           sha1: getSha1("firefoxaccount@test.com"),
           email: "firefoxaccount@test.com",
-          verification_token: "",
+          verification_token: "0e2cb147-2041-4e5b-8ca9-494e773b2cf1",
           verified: true,
+          fxa_refresh_token: "4a4792b89434153f1a6262fbd6a4510c00834ff842585fc4f4d972da158f0fc1",
```
pdehaan: I assume these are just random, dummy values and I shouldn't be scared?

lib/fxa.js
```diff
@@ -0,0 +1,35 @@
...
+      method: "POST",
+      headers: {"Authorization": `Bearer ${AppConstants.OAUTH_CLIENT_SECRET}`},
+      json: true,
+      body: tokenDestroyParams,
+    };
+
+    try {
+      return await got(tokenDestroyUrl, tokenDestroyOptions);
+    } catch (e) {
+      log.error("fxa", {stack: e.stack});
```
pdehaan: Not sure if this matters, but should it be `log.error("fxa", ...)`, or `log.error("revokeOAuthToken", ...)`?

lib/fxa.js
```diff
@@ -0,0 +1,35 @@
+"use strict";
+
+const got = require("got");
+const { URL } = require("url");
+
+const AppConstants = require("../app-constants");
+const mozlog = require("../log");
+
+
+const log = mozlog("fxa");
```
pdehaan: I [searched for current "mozlog" usage](https://github.com/mozilla/blurts-server/search?q=mozlog&type=Code), but this may be the first instance in lib/*. But for controllers/ses.js, we used `const log = mozlog("controllers.ses");` (but possibly only to differentiate from `const log = mozlog("routes.ses")`).

Not sure if we want to use `mozlog("lib.fxa")` for consistency, or if that is the tiniest of papercuts.

tests/controllers/user.test.js
```diff
@@ -118,11 +119,13 @@ test("user verify request with invalid token returns error", async () => {
 });
 
 
-test("user unsubscribe GET request with valid token returns error", async () => {
-  const validToken = "0e2cb147-2041-4e5b-8ca9-494e773b2cf0";
+test("user unsubscribe GET request with valid token and hash returns 200 without error", async () => {
+  // from db/seeds/test_subscribers.js
+  const subscriberToken = "0e2cb147-2041-4e5b-8ca9-494e773b2cf1";
+  const subscriberHash = getSha1("firefoxaccount@test.com");
```
pdehaan: I had to do a bit of scrolling to verify these looked the same.
Not sure if it's worth the hassle to possibly export the dummy user from db/seeds/test_subscribers.js to avoid having to duplicate the UUIDs in 2+ places.

tests/fxa.test.js
```diff
@@ -0,0 +1,23 @@
...
+
+const FXA = require("../lib/fxa");
+
+
+jest.mock("got");
+
+
+test("revokeOAuthToken calls oauth destroy with fxa_refresh_token", async () => {
+  // from db/seeds/test_subscribers.js
+  const token = "4a4792b89434153f1a6262fbd6a4510c00834ff842585fc4f4d972da158f0fc1";
```
pdehaan: Same comment as above. Should we duplicate these UUIDs vs trying to export them from db/seeds/test_subscribers.js (or somewhere else shared). Or possibly not worth it. 🤷‍♀️

controllers/oauth.js
groovecoder: Filed https://github.com/mozilla/blurts-server/issues/773

db/seeds/test_subscribers.js
groovecoder: Correct. They are old refresh tokens on dummy accounts on the FxA dev server that have already been deleted.

lib/fxa.js
groovecoder: Good catch. It should be `revokeOAuthToken` because `mozlog("fxa")` already includes the `fxa` string in the log messages.

lib/fxa.js
groovecoder: Good catch, but I'll fix it in a future PR.

I'm about to move a bunch of modules into `lib/` and I think the module names will all be unique enough to know they are `lib` modules. But if it looks like we'll have module names that are common across `controllers/` and `lib/` I'll update the `mozlog` calls.

tests/controllers/user.test.js
groovecoder: Good idea. I made `test_subscribers.js` export `TEST_DATA` for use in tests.

PR #747, opened 2019-02-06
fix #688: add csrf & encrypted page tokens for scans

### LINKED ISSUE #688: Add bot mitigation to scans
To prevent bots from submitting scan requests without valid browser sessions, add bot mitigations to the /scan endpoint.

groovecoder:
Note: This breaks the "Scan another email" form at the bottom of the `/scan` results. Not sure yet how to un-break it, but I need some other eyes on this ASAP.

### DIFF
--- a/controllers/home.js
+++ b/controllers/home.js
@@ -34,7 +34,7 @@ async function home(req, res) {
   let authenticatedUser = false;
 
   // for #688: use a page token to check for bot scans
-  const encryptedPageToken = AppConstants.PAGE_TOKEN_TIMER > 0 ? _generatePageToken(req) : "";
+  const pageToken = AppConstants.PAGE_TOKEN_TIMER > 0 ? _generatePageToken(req) : "";
 
   if (req.query.breach) {
     const reqBreachName = req.query.breach.toLowerCase();
@@ -83,7 +83,7 @@ async function home(req, res) {
   res.render("monitor", {
     title: req.fluentFormat("home-title"),
     csrfToken: req.csrfToken(),
-    encryptedPageToken: encryptedPageToken,
+    pageToken: pageToken,
     featuredBreach: featuredBreach,
     scanFeaturedBreach,
     foundBreaches,
--- a/controllers/home.js
+++ b/controllers/home.js
@@ -59,6 +59,7 @@ async function home(req, res) {
 
   res.render("monitor", {
     title: req.fluentFormat("home-title"),
+    csrfToken: req.csrfToken(),
     featuredBreach: featuredBreach,
     scanFeaturedBreach,
     foundBreaches,
--- a/controllers/home.js
+++ b/controllers/home.js
@@ -12,7 +12,7 @@ const sha1 = require("../sha1-utils");
 function _generatePageToken(req) {
   const pageToken = {ip: req.ip, date: new Date(), nonce: uuidv4()};
   const cipher = crypto.createCipher("aes-256-cbc", AppConstants.COOKIE_SECRET);
-  const encryptedPageToken = [cipher.update(JSON.stringify(pageToken), "utf8", "hex"), cipher.final("hex")].join("");
+  const encryptedPageToken = [cipher.update(JSON.stringify(pageToken), "utf8", "base64"), cipher.final("base64")].join("");
   return encryptedPageToken;
 
   /* TODO: block on scans-per-ip instead of scans-per-timespan
--- a/controllers/home.js
+++ b/controllers/home.js
@@ -1,9 +1,29 @@
 "use strict";
 
+const crypto = require("crypto");
 const { URL } = require("url");
+const uuidv4 = require("uuid/v4");
 
+const AppConstants = require("../app-constants");
 const HIBP = require("../hibp");
 const sha1 = require("../sha1-utils");
+
+
+function _generatePageToken(req) {
+  const pageToken = {ip: req.ip, date: new Date(), nonce: uuidv4()};
+  const cipher = crypto.createCipher("aes-256-cbc", AppConstants.COOKIE_SECRET);
+  const encryptedPageToken = [cipher.update(JSON.stringify(pageToken), "utf8", "hex"), cipher.final("hex")].join("");
+  return encryptedPageToken;
+
+  /* TODO: block on scans-per-ip instead of scans-per-timespan
+  if (req.session.scans === undefined){
+    console.log("session scans undefined");
+    req.session.scans = [];
+  }
+  req.session.numScans = req.session.scans.length;
+  */
+}
+
 
 async function home(req, res) {
 
@@ -12,6 +32,9 @@ async function home(req, res) {
   let foundBreaches = [];
   let userAccountCompromised = false;
   let authenticatedUser = false;
+
+  // for #688: use a page token to check for bot scans
+  const encryptedPageToken = AppConstants.PAGE_TOKEN_TIMER > 0 ? _generatePageToken(req) : "";
 
   if (req.query.breach) {
     const reqBreachName = req.query.breach.toLowerCase();
@@ -60,6 +83,7 @@ async function home(req, res) {
   res.render("monitor", {
     title: req.fluentFormat("home-title"),
     csrfToken: req.csrfToken(),
+    encryptedPageToken: encryptedPageToken,
     featuredBreach: featuredBreach,
     scanFeaturedBreach,
     foundBreaches,
--- a/controllers/scan.js
+++ b/controllers/scan.js
@@ -20,7 +20,7 @@ function _decryptPageToken(encryptedPageToken) {
 
 
 function _validatePageToken(pageToken, req) {
-  const requestIP = req.ip;
+  const requestIP = req.headers["x-real-ip"] || req.ip;
   const pageTokenIP = pageToken.ip;
   if (pageToken.ip !== requestIP) {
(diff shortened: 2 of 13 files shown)


views/partials/scan_form.hbs
```diff
@@ -1,4 +1,6 @@
 <form action="/scan" class="email-scan form-group" method="post" data-no-csrf novalidate>
```
pdehaan: Personal curiosity, but now that we're passing hidden `_csrf` fields w `{{csrfToken}}`s, do we still need/want `data-no-csrf` form attributes?

controllers/scan.js
```diff
@@ -1,15 +1,68 @@
...
   let foundBreaches = [];
 
+  // for #688: use a page token to check for bot scans
+  if (AppConstants.PAGE_TOKEN_TIMER > 0) {
+    if (!encryptedPageToken) {
+      throw new FluentError("error-no-page-token");
+    }
+    const decryptedPageToken = _decryptPageToken(encryptedPageToken);
+    if (!_validatePageToken(decryptedPageToken, req)) {
+      throw new FluentError("error-invalid-page-token");
```
pdehaan: I didn't find `"error-no-page-token"` (L57) or `"error-invalid-page-token"` in any of the .ftl files. Not sure if we need to add the strings still since they appear to be user/bot facing.

views/partials/scan_form.hbs
psiinon: Yeah, if you're supplying a CSRF token then you should definitely remove the data-no-csrf attribute :)

groovecoder:
Now that it's the next day, I think the encrypted page token effectively provides CSRF protection. @psiinon - what do you think?

psiinon:
Quite possibly - how is the encrypted page token created, stored and used?

controllers/home.js
```diff
@@ -1,10 +1,30 @@
...
 const { URL } = require("url");
+const uuidv4 = require("uuid/v4");
 
+const AppConstants = require("../app-constants");
 const HIBP = require("../hibp");
 const sha1 = require("../sha1-utils");
 
+
+function _generatePageToken(req) {
+  const pageToken = {ip: req.ip, date: new Date(), nonce: uuidv4()};
```
jvehent: just to sanity check: `req.ip` contains the client ip from the x-forwarded-for, not the load balancer?

controllers/home.js
```diff
@@ -1,10 +1,30 @@
...
+  const cipher = crypto.createCipher("aes-256-cbc", AppConstants.COOKIE_SECRET);
```
jvehent: can we use `aes-256-gcm` instead?

controllers/home.js
```diff
@@ -1,10 +1,30 @@
...
+  const encryptedPageToken = [cipher.update(JSON.stringify(pageToken), "utf8", "hex"), cipher.final("hex")].join("");
```
jvehent: nit: base64 may be shorter than hex

controllers/home.js
groovecoder: Correct, because we've got [`app.enable("trust proxy");`](https://expressjs.com/en/4x/api.html#trust.proxy.options.table)

groovecoder:
tl;dr the code 

1. On a GET to `/` I create a token of the client IP, timestamp, and a nonce. I encrypt the token using our `COOKIE_SECRET` env var value.
2. The scan form on `/` includes the encrypted token in a `<input type="hidden">` field.
3. POST to `/scan` receives the token, decrypts it, checks that the request IP address matches the IP address in the token, and that the timestamp in the token is less than 60 seconds ago. (60s value configurable by `PAGE_TOKEN_TIME` env var so we can adjust if too many humans are getting errors.)

The protections this adds:

1. Every `POST /scan` needs a valid unique token that it receives from the form in `/` (effective CSRF protection, right?)
2. Each IP has a unique token, so a token can't be re-used across IP addresses (i.e., each client has to `GET /` to get its own token)
3. The token expires in 60s, so each client has to `GET /` at least once every 60s

controllers/home.js
jvehent: From that doc: `If true, the client’s IP address is understood as the left-most entry in the X-Forwarded-* header.`

Which is a big fail, because we know bots inject x-forwarded-for headers to pass random ips as the leftmost entry. @Micheletto can confirm that our glb doesn't sanitize those, which is sad, but effectively means this control won't work.

jvehent: Ok, that library supports `Trusting the nth hop from the front-facing proxy server as the client.` Is that how we're currently setup?

views/partials/scan_form.hbs
groovecoder: Ah, I think I remember now ... this attribute is a signal to ZAP? Or another of our scanners?

controllers/home.js
g-k: In that case we need to remember to update https://github.com/mozilla/blurts-server/pull/747/files#diff-3e18cca2cad1fa80c47052b459cd06a0R16 too

package.json
```diff
@@ -14,6 +14,7 @@
     "client-oauth2": "^4.2.1",
     "client-sessions": "^0.8.0",
     "cpr": "^3.0.1",
+    "csurf": "^1.9.0",
```
g-k: NB: This is kind of unmaintained (last published 3 years ago; older open PRs). https://npms.io/search?q=csurf

I think it's solid though and don't know of any alternatives of the top of my head.

controllers/home.js
g-k: Doesn't look like it? I think this should be a number instead of `true` now: https://github.com/mozilla/blurts-server/blob/1e2bc90b3e20a0a59438c0f6b9bda9241b831ed8/server.js#L136-L140

Sanitizing XFF seems like pretty basic functionality, disappointing that GLB doesn't do that.

Micheletto: > @Micheletto can confirm that our glb doesn't sanitize those, which is sad, but effectively means this control won't work.

That is correct.  We could set a header that is passed to the app with the real-ip though.

controllers/home.js
g-k: Also, is the confidentiality of the token data important? Could we sign the data and submit nonce, date, and sig instead? I guess that might be larger.

controllers/home.js
groovecoder: Nope, but I can set it to whatever N we want. Do we know it will always be the *n*th hop from the front-facing proxy server?

controllers/scan.js
groovecoder: Yes, will need to add some strings for this. But I don't know if it needs @betsymi to write them, or if I can just write them myself. @betsymi - do you want to write error messages here?

controllers/home.js
groovecoder: When I read [how to use GCM](https://lollyrock.com/posts/nodejs-encryption/), it looks like we would need a per-encryption IV, and we don't currently have a back-end data store for holding those IVs and associating them with each session ID. Is `-cbc` okay "for now" ?

jvehent: The IV and AAD should just be stored in cleartext alongside the encrypted data, but yes, cbc is fine for now.

controllers/home.js
groovecoder: I'm going to drop the `trust proxy` setting here, and we're going to have nginx explicitly set a custom header `X-Real-IP` that I will read from.

controllers/scan.js
betsymi: Yes, happy to. I'm having a hard time following what the errors are themselves though. @groovecoder could you help me out by drafting what these are supposed to communicate, and I can work on making them more user-friendly?

groovecoder:
I added the same token to the "Scan another email" form to fix that form for humans that use the form before the page token expires. Also added an error message when the token is invalid, so @flodolo will have to review before merge.

flodolo:
> I added the same token to the "Scan another email" form to fix that form for humans that use the form before the page token expires. Also added an error message when the token is invalid, so @flodolo will have to review before merge.

Do we expect more strings, or that's just this one?

controllers/scan.js
groovecoder: @betsymi @flodolo right now I just have a single error string: "Your network may be performing too many scans. Please try again later."

Is that good?

flodolo: I wonder if that's a bit confusing for normal users? (i.e. what's "my network"?) Not sure what's the target and frequency of this message showing up though.

Maybe something like: "We detected too many scans originating from your address and the request has been blocked. Please try again later."

betsymi: What makes this error appear? Is it that a bot is trying to scan too many email addresses in quick succession?

groovecoder: So far yes - this is aimed at bots. The normal users who are likely to encounter this would be people who are using "Scan another email" box repeatedly.

I agree "network" may not be the right term. But "address" might feel weird too - will people think we're somehow looking at their physical address? Should we be afraid of using the term "IP Address"? Google and many others use it all the time when they don't want me using their services from Tor.

(But as a scraper-writing colleague of mine pointed out, we shouldn't be anti-bot 😜 ... i.e., we don't really want humans to be scanning  dozens of email addresses either.)

betsymi: You tried to search too many emails in a short time period. For security reasons, this IP address is temporarily blocked from new searches. You can try again later.

groovecoder:
Updated with new error message. @flodolo - how does this one look?

PR #749, opened 2019-02-07
String fixes. 

### LINKED ISSUE #725: Update copyright year in footer?
Randomly spotted on the site, not sure if it needs updating to **2019** or not...

> Portions of this content are Ⓒ **1998-2018** by individual mozilla.org contributors. 
Content available under a Creative Commons license.

https://github.com/mozilla/blurts-server/blob/ecc85281efd054343d7f9dc9e862805a36d367b1/locales/en/app.ftl#L243-L246

### LINKED ISSUE #741: Missing word in one of the new FxA strings
It looks like a word is missing in this string `fxa-what-to-do-blurb-2`

`Hackers may try to that same password and your email to get into other accounts. ` should probably be `Hackers may try to *reuse* that same password and your email to get into other accounts.`

lesleyjanenorton:
- Punctuation updates requested by content. 
- Fixes #741 
- Fixes #725

### DIFF
--- a/locales/en/app.ftl
+++ b/locales/en/app.ftl
@@ -241,12 +241,6 @@ features = Features
 # beta-nightly-developer-edition refers to additional versions of Firefox Browser
 beta-nightly-developer-edition = Beta, Nightly, Developer Edition
 
-# The following string contains HTML markup which should not be translated. 
-# Without HTML markup: copyright-info = Portions of this content are 1998-2018 by individual mozilla.org contributors. Content available under a Creative Commons license.
-copyright-info = 
-  Portions of this content are &#x24B8; 1998-2018 by individual mozilla.org contributors. <br />
-  Content available under a  <a href="https://www.mozilla.org/foundation/licensing/website-content/" target="_blank" rel="noopener">Creative Commons license</a>.
-
 # Breach data provided by Have I Been Pwned.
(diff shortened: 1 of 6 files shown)


views/partials/footer.hbs
```diff
@@ -33,7 +33,10 @@
             </ul>
           </div>
           <div class="row legal">
-            <p class="half copyright-info">{{{fluentFormat req.supportedLocales "copyright-info"}}}</p>
```
pdehaan: Do we need to delete the "copyright-info" string from the .ftl file?

flodolo: Yes, we should

locales/en/app.ftl
```diff
@@ -446,3 +446,8 @@ fxa-what-to-do-blurb-4 =
   on your phone and computer so you don’t have to remember them all.
 
 fb-landing-headline = Was your info exposed in the { $breachName } data breach?
+
+copyright = 
+  Portions of this content are { $years } by individual mozilla.org contributors.
```
flodolo: Can we have only the final year exposed as a variable? Right now the code is making a lot of assumptions on spacing, separators, etc.

```
copyright = 
  Portions of this content are © 1998-{ $year } by individual mozilla.org contributors.
```

lesleyjanenorton: Yep, will do!

PR #750, opened 2019-02-07
HIBP attribution updates.

lesleyjanenorton:
- Adds attribution to breach detail pages (isn't in spec, but should be there).
- Updates HIBP styling/positioning to match spec.


views/partials/hibp_attribution.hbs
```diff
@@ -1,3 +1,3 @@
-<span class="source-info">
+<span class="source-info {{ class }}">
```
groovecoder: Will this be confusing later that `source-info` is a hard-coded class for this attribute, and then `class` argument only adds to that - i.e., it doesn't replace the entire `class` property value?

Would it be clearer to do something like  `{{> hibp_attribution additionalClasses="align-left"`}}` and `<span class="source-info {{ additionalClasses }}">`

lesleyjanenorton: Makes sense! Updated in the name of walking The Path of Most Future Sanity.

PR #754, opened 2019-02-08
https://nvd.nist.gov/vuln/detail/CVE-2018-16487

groovecoder:


package.json
```diff
@@ -30,6 +30,7 @@
     "isemail": "^3.1.3",
     "jsdom": "^11.11.0",
     "knex": "^0.14.6",
+    "lodash": "4.17.11",
```
pdehaan: LGTM. Only observation was the other dependencies use the `^` semver range operator, whereas the new **lodash** dependency is pinned to "4.17.11".

PR #760, opened 2019-02-11
Use #loop to loop through unsub_reasons and cleanup.

lesleyjanenorton:
-Deletes unused tips.js
-Deletes unsubscribe_reasons.js
-Uses {{#loop}} instead of {{#each}} to iterate through unsubscribe reasons


groovecoder:
I couldn't get to the unsubscribe page because of https://github.com/mozilla/blurts-server/issues/763, but this code looks good to go.

PR #765, opened 2019-02-12
Consolidate scan result logic for easier current/future templating. 

lesleyjanenorton:
- Moves scan logic that was living in various/multiple controllers into standalone `scan-results.js` that returns one big object for sending to the templates.
- Updates templates with new/renamed vars.
   - `latestBreaches` is now `userDash`
   - `authenticatedUser` is now `selfScan`
   - `featuredBreach` is now `specificBreach`  (in `scan-results.js` and various result partials)
- Fixes some logic in the templates unearthed while working on this. Specifically:
   - Single breach results not showing up on `fullReport` and `userDash` pages.
   - Some incorrect response headlines/subheads when signed in users scan emails other than their own from featured breach pages.
   - Fixes rder and appearance of the sign up banner, the download banner, and the what to do tips on result pages.
   - See Full Report incorrectly showing up when a signed in user scans an email that is not their own from a featured breach page.
- Sets us up to move all of the iffing/elsing/unlessing/ifComparing out of the handlebars templates and into something a bit easier to wrangle.

### DIFF
--- a/routes/scan.js
+++ b/routes/scan.js
@@ -2,6 +2,7 @@
 
 const express = require("express");
 const bodyParser = require("body-parser");
+const csrf = require("csurf");
 
 const {asyncMiddleware} = require("../middleware");
 const {post, get, getFullReport, getUserDashboard} = require("../controllers/scan");
@@ -9,10 +10,12 @@ const {post, get, getFullReport, getUserDashboard} = require("../controllers/sca
 
 const router = express.Router();
 const urlEncodedParser = bodyParser.urlencoded({ extended: false });
+const csrfProtection = csrf();
 
-router.post("/", urlEncodedParser, asyncMiddleware(post));
+ 
+router.post("/", urlEncodedParser, csrfProtection, asyncMiddleware(post));
 router.get("/", get);
 router.get("/full_report", getFullReport);
-router.get("/user_dashboard", getUserDashboard);
+router.get("/user_dashboard", urlEncodedParser, csrfProtection, asyncMiddleware(getUserDashboard));
 
 module.exports = router;
--- a/routes/scan.js
+++ b/routes/scan.js
@@ -4,7 +4,7 @@ const express = require("express");
 const bodyParser = require("body-parser");
 
 const {asyncMiddleware} = require("../middleware");
-const {post, get, getFullReport, getLatestBreaches} = require("../controllers/scan");
+const {post, get, getFullReport, getUserDashboard} = require("../controllers/scan");
 
 
 const router = express.Router();
@@ -13,6 +13,6 @@ const urlEncodedParser = bodyParser.urlencoded({ extended: false });
 router.post("/", urlEncodedParser, asyncMiddleware(post));
 router.get("/", get);
 router.get("/full_report", getFullReport);
-router.get("/latest_breaches", getLatestBreaches);
+router.get("/user_dashboard", getUserDashboard);
 
 module.exports = router;
--- a/scan-results.js
+++ b/scan-results.js
@@ -5,11 +5,13 @@ const { URL } = require("url");
 const HIBP = require("./hibp");
 const sha1 = require("./sha1-utils");
 
+
 const scanResult = async(req, selfScan=false) => {
 
   const allBreaches = req.app.locals.breaches;
   let scannedEmail = null;
 
+  const title = req.fluentFormat("scan-title");
   let foundBreaches = [];
   let specificBreach = null;
   let doorhangerScan = false;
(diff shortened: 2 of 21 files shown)


routes/scan.js
```diff
@@ -13,6 +13,6 @@ const urlEncodedParser = bodyParser.urlencoded({ extended: false });
 router.post("/", urlEncodedParser, asyncMiddleware(post));
 router.get("/", get);
 router.get("/full_report", getFullReport);
-router.get("/latest_breaches", getLatestBreaches);
+router.get("/user_dashboard", getUserDashboard);
```
groovecoder: Let's consider if we can just show the user dashboard at `/` url instead of redirecting signed-in users to `/user_dashboard`. @sandysage - should we file and issue to consider this? Or just discuss and decide in a meeting or something?

scan-results.js
```diff
@@ -0,0 +1,95 @@
+"use strict";
+
+const { URL } = require("url");
+
+const HIBP = require("./hibp");
+const sha1 = require("./sha1-utils");
+
```
groovecoder: nit: I always put 2 blank lines between the `require`s and the module code.

scan-results.js
```diff
@@ -0,0 +1,95 @@
...
+  if (req.session.user) {
+    signedInUser = req.session.user;
+  }
+
+  // Checks if the user scanning their own verified email.
+  if (req.body && req.body.emailHash) {
+    scannedEmail = req.body.emailHash;
+    if (!selfScan && signedInUser) {
+      if (sha1(signedInUser.email).toUpperCase() === req.body.emailHash) {
+        selfScan = true;
```
groovecoder: 3 nested levels of `if` blocks may mean we should consider if we can simplify the logic here, or at least move it to its own helper function with a descriptive name?

scan-results.js
```diff
@@ -0,0 +1,95 @@
...
+      }
+    }
+  }
+
+  const url = new URL(req.url, req.app.locals.SERVER_URL);
+
```
groovecoder: Could we simplify the below if blocks into here:
```js
const fullReport = url.pathname === "/full_report";
const userDash = url.pathname === "/user_dashboard";
```

scan-results.js
```diff
@@ -0,0 +1,95 @@
...
+  }
+
+  if (doorhangerScan || specificBreach) {
+    const specificBreachIndex = foundBreaches.findIndex(breach => breach.Name === specificBreach.Name);
+
+    // Checks foundBreaches for specificBreach and if found,
+    // brings specificBreach to front of foundBreaches list.
+    if (specificBreachIndex !== -1) {
+      userCompromised = true;
+      if (foundBreaches.length > 1) {
```
groovecoder: 3 nested `if` blocks again.

scan-results.js
```diff
@@ -0,0 +1,95 @@
...
+    foundBreaches: foundBreaches,
+    specificBreach: specificBreach,
+    doorhangerScan: doorhangerScan,
+    userCompromised: userCompromised,
+    signedInUser: signedInUser,
+    selfScan: selfScan,
+    fullReport: fullReport,
+    userDash: userDash,
+    title: req.fluentFormat("scan-title"),
+  };
```
groovecoder: Since all the property value variables are the same as the property names, this could be condensed to:
```js
const title = req.fluentFormat("scan-title");
return { 
  foundBreaches, specificBreach, doorhangerScan, userCompromised,
  signedInUser, selfScan, fullReport, userDash, title
};
```

tests/controllers/scan.test.js
```diff
@@ -39,6 +39,8 @@ test("scan POST with hash should render scan with foundBreaches", async () => {
   mockRequest.body = { emailHash: sha1(testEmail) };
   mockRequest.app = { locals: { breaches: testBreaches } };
   mockRequest.session = { user: null };
+  mockRequest.url = { url: "https://www.mozilla.com" };
+  mockRequest.app.locals.SERVER_URL = AppConstants.SERVER_URL;
```
groovecoder: Not clear what this url param is testing?

tests/controllers/home.test.js
```diff
@@ -28,17 +29,21 @@ test("home GET without breach renders monitor without breach", () => {
...
-  mockRequest.url = "https://www.mozilla.com";
   mockRequest.session = { user: null };
-  const mockResponse = { render: jest.fn() };
+  mockRequest.url = { url: "https://www.mozilla.com" };
+  mockRequest.app.locals.SERVER_URL = AppConstants.SERVER_URL;
 
 
+  const mockResponse = { render: jest.fn(), redirect: jest.fn() };
   home.home(mockRequest, mockResponse);
+  const scanRes = await scanResult(mockRequest);
```
groovecoder: We could move this test code into a new test module that only tests the `scanResult` function in isolation from the rest of this test logic.

scan-results.js
```diff
@@ -0,0 +1,95 @@
...
+
+  if (req.session.user) {
+    signedInUser = req.session.user;
+  }
+
+  // Checks if the user scanning their own verified email.
+  if (req.body && req.body.emailHash) {
+    scannedEmail = req.body.emailHash;
+    if (!selfScan && signedInUser) {
+      if (sha1(signedInUser.email).toUpperCase() === req.body.emailHash) {
```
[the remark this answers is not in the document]
lesleyjanenorton: Nope! After some console logging and poking around, it appears that the sha1 function in monitor.js returns an uppercase string which then gets `.toUpperCased()` again in `HIBP.getBreachesForEmail()` anyway.  Regarding your other sha1() comment below - I think we can absolutely have `sha1()` (in sha1-utils.js) return an uppercase string and then be done with all of the `sha1(...).toUpperCase()` all over the place.  🎉🎉🎉

pdehaan:
Looks like Travis is failing due to https://www.npmjs.com/advisories/755

groovecoder:
I would *NOT* try to re-base this PR on latest commit with the `hbs` fix for `npm audit`, or you'll have to re-do all the conflict resolutions.

Instead, we could merge the other PR(s) first, and then cherry pick just the `hbs` removal into this PR. Or we can just merge this thru the fails, since we know it'll be fixed in master.

groovecoder:
`push` job is failing, but `pr` job that has the `hbs` fix is now passing. merging.

PR #768, opened 2019-02-13
dev and heroku can trust proxy

groovecoder:


server.js
```diff
@@ -130,6 +136,7 @@ app.set("view engine", "hbs");
 
 const cookie = {httpOnly: true, sameSite: "lax"};
 
+(devOrHeroku ? app.set("trust proxy", true) : "");
```
lesleyjanenorton: Very snazzy

pdehaan: What does this do? Isn't it just

```js
if (devOrHeroku) {
  app.set("trust proxy", true);
}
```

with an odd `""` at the end?

groovecoder: Yeah it's a bit odd. I turned it into a single-liner to line up more closely with the surrounding lines. I guess it could just be `if (devOrHeroku) app.set("trust proxy", true);` ?

PR #776, opened 2019-02-15
fix #775: dev, heroku, and stage don't need to force https

### LINKED ISSUE #775: Bypass https redirect for Dockerflow routes on stage
When the Dockerflow service + agent runs, it hits the stage server over http. We need to make sure we can serve the Dockerflow routes over http on stage.

groovecoder:


groovecoder:
This will fix an issue on the stage server where we need to be able to serve the `/__lbheartbeat__` endpoint over http.

server.js
```diff
@@ -39,7 +39,8 @@ function devOrHeroku() {
 }
 
 
-if (app.get("env") !== "dev") {
+// dev, heroku, and stage environments don't need to force https for all urls
+if (!["dev", "heroku", "stage"].includes(app.get("env"))) {
```
pdehaan: Why stage? Shouldn't stage and prod be virtually identical?

groovecoder: Good question, and we may hit a similar issue in prod. Hmm ... I may have to dig into why removing `trust proxy` seemed to break the Dockerflow heartbeat routes.

groovecoder:
Okay, I restored the `trust proxy` setting for all non-dev environments, since the `pageToken` code will use the `X-Real-IP` header. This should fix stage issue where Dockerflow routes were not serving correctly to the Dockerflow agent that was requesting them.

groovecoder:
Heads-up @GeorgiuCiprian, @obotisan, @sorinaflorean, @sv-sdeiac ...

This code is now live on the stage server, along with a new "page token" feature that is meant to mitigate bot scans.

So, you might see this error message if you're scanning too fast: "You tried to scan too many email addresses in a short time period. For security reasons, we’ve temporarily blocked you from new searches. You’ll be able to try again later."

You should be able to scan again if you go back to the home page and wait 90 seconds. Please let us know if you think you're hitting the error too often and we can increase the timeout.

GeorgiuCiprian:
@groovecoder Thanks for the heads-up!

We just tested this on stage and it looks that we're getting the error message from the first account scanned. Returning to the homepage and waiting for 90 secs doesn't seem to make any difference either; we're still seeing the error: "You tried to scan too many email addresses in a short time period. For security reasons, we’ve temporarily blocked you from new searches. You’ll be able to try again later."

Any thoughts about this, Luke?

pdehaan:
+1. Can confirm I was able to scan multiple accounts as quick as possible on Heroku this AM with no issues. But scanning on [stage](https://stage.firefoxmonitor.nonprod.cloudops.mozgcp.net/scan) threw up an error on the first scan:

> **Error**
> You tried to scan too many email addresses in a short time period. For security reasons, we’ve temporarily blocked you from new searches. You’ll be able to try again later.

groovecoder:
Hmm ... can you try https://blurts-server-review.herokuapp.com/ too? I just pushed this up to that server too.

I have a feeling it's something to do with `X-Real-IP` on stage (which will also need to be on prod).

groovecoder:
@Micheletto - can you check that stage environment is sending the `x-real-ip` header as expected by the page token code?

PR #793, opened 2019-02-19
fix #710 fix #721: delete subscriber on fxa delete

### LINKED ISSUE #710: When a user deletes their FxA, delete their Monitor records + data
FxA currently broadcasts account deletions to an SQS queue. We will need to either:

* Connect directly to the SQS queue from Monitor code with credentials (like we do with SES)
* See if FxA is using SNS to publish to SQS; if so we can ask to get into their SNS flow
* Ask FxA to create an https callback feature for account deletions

### LINKED ISSUE #721: Store fxa_uid on sign-up
To prepare for FxA users with multiple email addresses subscribed, store their fxa_uid when they sign up, so we can insert it with their future subscribers records.

groovecoder:
Store fxa uid when adding subscribers; handle SNS callback messages from
FXA delete events, and delete the corresponding Monitor subscriber.

### DIFF
--- a/db/DB.js
+++ b/db/DB.js
@@ -167,9 +167,11 @@ const DB = {
    * @returns {object} updated subscriber knex object in DB
    */
   async _updateFxAData(subscriber, fxaRefreshToken, fxaProfileData) {
+    const fxaUID = JSON.parse(fxaProfileData).uid;
     const updated = await knex("subscribers")
     .where("id", "=", subscriber.id)
     .update({
+      fxa_uid: fxaUID,
       fxa_refresh_token: fxaRefreshToken,
       fxa_profile_json: fxaProfileData,
     })
@@ -217,6 +219,10 @@ const DB = {
       .where("verified", false)
       .andWhere("created_at", "<", expiredTimeStamp)
       .del();
+  },
+
+  async deleteSubscriberByFxAUID(fxaUID) {
+    await knex("subscribers").where("fxa_uid", fxaUID).del();
   },
 
   async createConnection() {
--- a/controllers/ses.js
+++ b/controllers/ses.js
@@ -34,8 +34,34 @@ async function notification(req, res) {
 
 
 async function handleNotification(notification) {
-  log.info("received-SES", { id: notification.MessageId });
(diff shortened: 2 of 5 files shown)


groovecoder:
To spot-check, you'll need to run db migrations to add the new `fxa_uid` column.

`knex --knexfile db/knexfile.js migrate:latest`

groovecoder:
FWIW I just spot-checked this functionality on https://blurts-server-review.herokuapp.com/ and it works end-to-end! I deleted my Firefox Account on https://stable.dev.lcip.org/, and it triggered the SNS-http callback which deleted the `subscribers` record from the DB.

db/DB.js
```diff
@@ -219,6 +221,10 @@ const DB = {
       .del();
   },
 
+  async deleteSubscriberByFxAUID(fxaUID) {
+    await knex("subscribers").where("fxa_uid", fxaUID).del();
```
pdehaan: I think this is OK. I was a bit confused since the `where` clause up on L172 above has `.where("id", "=", subscriber.id)`, but this doesn't have an "=" operator. But it looks like the docs imply that it is optional: https://knexjs.org/#Builder-wheres

The only other interesting thing from the docs was that an error may get thrown if `fxaUID` is `undefined` above:

> **Important:** Supplying knex with an `undefined` value to any of the where functions will cause knex to throw an error during sql compilation. This is both for yours and our sake. Knex cannot know what to do with `undefined` values in a where clause, and generally it would be a programmatic error to supply one to begin with. The error will throw a message containing the type of query and the compiled query-string.

Not sure if we want to add a bit more footgun protections by wrapping in a `try..catch` block, or if bubbling errors are caught higher up the flow and I'm just not seeing everything.

groovecoder: Yeah, good idea. I'll add that tomorrow before merge.

groovecoder: After looking at the flow, I think it's okay to let the error throw from here so that it can record into Sentry with as much of the stack as possible. These requests are never sent by humans, so we don't need to return a human-friendly error message here.

PR #802, opened 2019-02-21
Add new strings and update partials.

### LINKED ISSUE #773: Unsubscribe page should tell the user that their Firefox Account will NOT be deleted
After #742 lands, Firefox Account users will be able to unsubscribe from Firefox Monitor.

When a user clicks their unsubscribe link, we should tell them that unsubscribing *DOES NOT* delete their Firefox Account. (We may even want to provide a link to their Firefox Account settings page if they want to delete that account too?)

lesleyjanenorton:
- Adds legally sanctioned, latest copy.
- Updates partials to use new strings.
- Changes sign up button text when foundBreaches "===" 0.
- Changes sign up banner text and removes sign up button from the sign up banner on sensitive breach detail pages.

-Fixes #773

### DIFF
--- a/controllers/user.js
+++ b/controllers/user.js
@@ -110,7 +110,7 @@ async function getUnsubscribe(req, res) {
 
   if (AppConstants.FXA_ENABLED) {
     const fxaUrl = new URL(AppConstants.OAUTH_PROFILE_URI).origin;
-    const manageFxaLink = `<a class="manage-fxa" href="${fxaUrl}" target="_blank" rel="noopener"><span>${req.fluentFormat("manage-fxa")}</span>`;
+    const manageFxaLink = `<a class="manage-fxa" href="${fxaUrl}" target="_blank" rel="noopener">${req.fluentFormat("manage-fxa")}</a>`;
     headline = req.fluentFormat("fxa-unsub-headline");
     subhead = `${req.fluentFormat("fxa-unsub-blurb")}${manageFxaLink}`;
   }
--- a/controllers/user.js
+++ b/controllers/user.js
@@ -2,6 +2,10 @@
 
 const crypto = require("crypto");
 const isemail = require("isemail");
+const { URL } = require("url");
+
+
+const AppConstants = require("../app-constants");
 const DB = require("../db/DB");
 const EmailUtils = require("../email-utils");
 const { FluentError } = require("../locale-utils");
@@ -101,10 +105,20 @@ async function getUnsubscribe(req, res) {
     throw new FluentError("error-not-subscribed");
   }
 
+  let headline = req.fluentFormat("unsub-headline");
+  let subhead = req.fluentFormat("unsub-blurb");
+
+  if (AppConstants.FXA_ENABLED) {
+    const fxaUrl = new URL(AppConstants.OAUTH_PROFILE_URI).origin;
+    const manageFxaLink = `<a class="manage-fxa" href="${fxaUrl}" target="_blank" rel="noopener"><span>${req.fluentFormat("manage-fxa")}</span>`;
+    headline = req.fluentFormat("fxa-unsub-headline");
+    subhead = `${req.fluentFormat("fxa-unsub-blurb")}${manageFxaLink}`;
+  }
+
   res.render("subpage", {
     title: req.fluentFormat("user-unsubscribe-title"),
-    headline: req.fluentFormat("unsub-headline"),
-    subhead: req.fluentFormat("unsub-blurb"),
+    headline: headline,
+    subhead: subhead,
     whichPartial: "subpages/unsubscribe",
     token: req.query.token,
     hash: req.query.hash,
@@ -138,8 +152,8 @@ function getUnsubSurvey(req, res) {
   }
   res.render("subpage", {
     title: req.fluentFormat("user-unsubscribe-survey-title"),
-    headline: req.fluentFormat("unsub-survey-headline"),
-    subhead: req.fluentFormat("unsub-survey-blurb"),
+    headline: req.fluentFormat("unsub-survey-headline-v2"),
+    subhead: req.fluentFormat("unsub-survey-blurb-v2"),
     whichPartial: "subpages/unsubscribe_survey",
   });
 }
--- a/controllers/user.js
+++ b/controllers/user.js
@@ -2,10 +2,8 @@
 
 const crypto = require("crypto");
 const isemail = require("isemail");
-const { URL } = require("url");
 
 
-const AppConstants = require("../app-constants");
 const DB = require("../db/DB");
 const EmailUtils = require("../email-utils");
 const { FluentError } = require("../locale-utils");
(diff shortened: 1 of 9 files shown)


controllers/user.js
```diff
@@ -101,10 +105,20 @@ async function getUnsubscribe(req, res) {
     throw new FluentError("error-not-subscribed");
   }
 
+  let headline = req.fluentFormat("unsub-headline");
+  let subhead = req.fluentFormat("unsub-blurb");
+
+  if (AppConstants.FXA_ENABLED) {
+    const fxaUrl = new URL(AppConstants.OAUTH_PROFILE_URI).origin;
+    const manageFxaLink = `<a class="manage-fxa" href="${fxaUrl}" target="_blank" rel="noopener"><span>${req.fluentFormat("manage-fxa")}</span>`;
```
pdehaan: Does this need a `</a>`?

Also, what's the purpose of the `<span>` wrapper? Are we using that in CSS or something?

locales/en/app.ftl
```diff
@@ -210,10 +210,11 @@ unsub-headline = Unsubscribe from {-product-name-nowrap}
...
 unsub-button = Unsubscribe
 
-unsub-survey-headline = You are no longer subscribed.
-unsub-survey-blurb = 
-  Your email is unsubscribed from {-product-name-nowrap}. Thank you for using this service. 
-  Will you take a moment to answer one question about your experience?
+fxa-unsub-headline = Unsubscribe from {-product-name} alerts.
+fxa-unsub-blurb = You’ll no longer receive {-product-name} alerts. 
+  Your {-brand-fxa} will remain active, and you may receive other 
+  account-related communications.
```
pdehaan: Nit: should we include a link to manage their FxA settings if they want to nuke that too?

controllers/user.js
lesleyjanenorton: Definitely doesn't need the span, definitely does need the `</a>`.

locales/en/app.ftl
lesleyjanenorton: Yep, the link from the above comment follows this string.

controllers/user.js
groovecoder: I think we should move this `FXA_ENABLED` check into the view and use the `getFxaUrl` helper for the `href`.

groovecoder: I.e., move all this logic into the `.hbs`

locales/en/app.ftl
```diff
@@ -334,16 +340,19 @@ user-fb-compromised-single =
...
+    Create a free {-brand-fxa} for your full report of past breaches, new breach alerts, 
+    and info about other {-brand-Mozilla} services.
+
+guest-fb-compromised-blurb-v2 =
   { $breachCount ->
-      [one] This email also appeared in { $breachCount } other breach. Sign up for a 
-            {-brand-fxa} to get your full report and alerts about new breaches.
-     *[other] This email also appeared in { $breachCount } other breaches. Sign up for a 
-              {-brand-fxa} to get your full report and alerts about new breaches.
+      [one] This email also appeared in { $breachCount } other breach. Create a
```
pdehaan: I don't know that we have a rule or not, but I did notice in other strings we hardcode "1" or sometimes "one". Not sure it makes a difference to anything in life, but see-something-say-something:

1. [`featured-breach-results[one]`](https://github.com/mozilla/blurts-server/blob/d1e1fcdb9098a89d09bbdbd7f52e4dca201c6fb4/locales/en/app.ftl#L149-L154) = 

    > ```html
    > [one] Your account appeared in the <span class="bold"> { $featuredBreach } </span> breach, as well as one other breach.
    > ```

1. [`featured-breach-not-compromised[one]`](https://github.com/mozilla/blurts-server/blob/d1e1fcdb9098a89d09bbdbd7f52e4dca201c6fb4/locales/en/app.ftl#L156-L161) =

    > ```html
    > [one] Your account did not appear in the <span class="bold">{ $featuredBreach }</span> breach, but did appear in one other breach.
    > ```

1. [`user-scan-results-headline[one]`](https://github.com/mozilla/blurts-server/blob/d1e1fcdb9098a89d09bbdbd7f52e4dca201c6fb4/locales/en/app.ftl#L307-L311) =

    > ```html
    > [one] { $userEmail } appeared in 1 data breach.
    > ```

1. [`guest-scan-results-headline[one]`](https://github.com/mozilla/blurts-server/blob/d1e1fcdb9098a89d09bbdbd7f52e4dca201c6fb4/locales/en/app.ftl#L313-L317) = 

    > ```html
    > [one] This email appeared in 1 data breach.
    > ```

But not always... here's a couple places where we _didn't_ hardcode a number/string for the [one] case:

1. https://github.com/mozilla/blurts-server/blob/d1e1fcdb9098a89d09bbdbd7f52e4dca201c6fb4/locales/en/app.ftl#L163-L168

1. https://github.com/mozilla/blurts-server/blob/d1e1fcdb9098a89d09bbdbd7f52e4dca201c6fb4/locales/en/app.ftl#L327-L331

flodolo: The practical result for en-US is the same, other locales will have to use the variables. Probably good to use the variable at that point.

lesleyjanenorton: When surfacing the number of breaches a user has, the rule going forward is to use the numeral - unless that numeral is zero. Zero should still get spelled out because it looks too much like an uppercase letter "O". 

I don't think it affects anything but I'll use { $breachCount } where possible instead of hardcoding the 1s in the future.

lesleyjanenorton:
@flodolo These are intended for the FxA launch, if possible.

Regarding your other comment, which variable do you mean?

flodolo:
> @flodolo These are intended for the FxA launch, if possible.

Again, this is a lot of string changes, and late. When we were handed off strings, I wasn't really expecting this to happen.

We should check how things are looking in terms of localization completion in a week or so.

> Regarding your other comment, which variable do you mean?

The number of breaches, instead of hard-coding 1 (I thought I replied under another series of comments).

locales/en/app.ftl
```diff
@@ -334,16 +340,19 @@ user-fb-compromised-single =
...
   change your passwords.
 user-generic-fb-compromised-single = This breach exposed the following personal info.
-guest-fb-compromised-single = 
-  This breach exposed the following personal info. Sign up for a {-brand-fxa} 
-  to get your full report and alerts about new breaches.
 
-guest-fb-compromised-blurb =
+guest-fb-compromised-single-v2 = This breach exposed the following personal info. 
+    Create a free {-brand-fxa} for your full report of past breaches, new breach alerts, 
+    and info about other {-brand-Mozilla} services.
```
flodolo: Note to self: I've overlooked this when we created terms, we should be consistent and use kebab-case for message IDs (in the future).

pdehaan: > and use kebab-case for message IDs (in the future).

It looks like only a handful of mixed-Case terms, if it's something you think we should change:

```
Term => brand-Quantum
Term => brand-Mozilla
Term => brand-HIBP
Term => brand-Chrome
Message => layout-Firefox
```

<details>
<summary>Source</summary>

```js
const fs = require("fs");
const { parse } = require("fluent-syntax");

const argv = process.argv.slice(2);

checkLocale(...argv);

function checkLocale(ftlFile = "./locales/en/app.ftl") {
  const ftl = fs.readFileSync(ftlFile, "utf-8");
  const { body } = parse(ftl, { withSpans: false });
  const checkCase = str => str === str.toLowerCase();

  body.forEach(resource => {
    switch (resource.type) {
      case "Term":
      case "Message":
        if (!checkCase(resource.id.name)) {
          console.log(`${resource.type} => ${resource.id.name}`);
        }
        break;
    }
  });
}
```
</details>

flodolo: As said, it's something to keep in mind for the future. Changing any of these would invalidate all existing strings.

PR #814, opened 2019-02-26
FxA Top of Funnel pings

### LINKED ISSUE #711: FxA "self-hosted email form" metrics tracking
FxA would like to measure the "top of funnel" metrics for FxA signups. For "self-hosted email forms" like Monitor has, we need to send [a `fetch` request to their `metrics-flow` url](https://github.com/mozilla/application-services/blob/master/docs/accounts/metrics.md#self-hosted-email-forms-and-metrics-tracking-aka-the-fxa-email-first-flow).

But first, we need to ask FxA if our email scan form constitutes a "self-hosted email form" as part of their top-of-funnel. @sandysage can you handle coordinate that with FxA team?

### LINKED ISSUE #807: Record the number of breaches a scan returns
Via conversation with @changecourse, this data could be used to show the value of monitor to potential users (e.g. "Users of monitor were found to have had on average XX breaches"), and might inform other design issues.

Of course, if we were to consider collecting such measures we would want to be very careful and get review from policy. This could be ripe for an application of [differential privacy](https://privacytools.seas.harvard.edu/differential-privacy)

### LINKED ISSUE #819: Record number of breaches from a signed in 'full report'
In addition to #807, we want to know roughly how many breaches are our signed in users see, and how different is that count from the 'free scan'. This will inform how we communicate breach results to our users in visual displays of the signed in state and how we promote the 'free scan'

lesleyjanenorton:
- Sends "top of funnel" pings to FxA `/metrics-flow` on page load for each FxA entrypoint (Sign In button in header, Sign Up button, Sign in Link, etc). 
- Stores `flowId` and `flowBeginTime` for each entrypoint element in that element's dataset.
- When a user clicks on a given entrypoint, the `entrypoint`, `flowId`, and `flowBeginTime` values are picked out and bundled up with `utm_source` and `utm_campaign` before being sent off to the /oauth flow.

Fixes #711 
Fixes #819 
Fixes #807

### DIFF
--- a/controllers/oauth.js
+++ b/controllers/oauth.js
@@ -35,16 +35,20 @@ const FxAOAuthClient = new ClientOAuth2({
   scopes: ["profile"],
 });
 
+
 function init(req, res, next, client = FxAOAuthClient) {
   // Set a random state string in a cookie so that we can verify
   // the user when they're redirected back to us after auth.
   const state = crypto.randomBytes(40).toString("hex");
   req.session.state = state;
   const url = new URL(client.code.getUri({state}));
+  const fxaParams = new URL(req.url, AppConstants.SERVER_URL);
+
   url.searchParams.append("access_type", "offline");
   url.searchParams.append("action", "email");
-  if (req.query.scanned) {
-    url.searchParams.append("email", req.query.scanned);
+
+  for (const param of fxaParams.searchParams.keys()) {
+    url.searchParams.append(param, fxaParams.searchParams.get(param));
   }
   res.redirect(url);
 }
--- a/package-lock.json
+++ b/package-lock.json
@@ -309,6 +309,405 @@
         }
       }
     },
+    "@cnakazawa/watch": {
+      "version": "1.0.3",
+      "resolved": "https://registry.npmjs.org/@cnakazawa/watch/-/watch-1.0.3.tgz",
+      "integrity": "sha512-r5160ogAvGyHsal38Kux7YYtodEKOj89RGb28ht1jh3SJb08VwRwAKKJL0bGb04Zd/3r9FL3BFIc3bBidYffCA==",
+      "dev": true,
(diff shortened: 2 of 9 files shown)


lesleyjanenorton:
For testing locally: Change `PORT` from `6060` to `8000` and manually set `redirectUri` (L34 
 of `oauth.js`) to `http://localhost:6060`.

irrationalagent:
I just pulled down your last commit, its looking good. 

One question, what would we expect `utm_source` to be set to in production? right now I'm seeing 
`utm_source=localhost%253A8000`. Is this being pulled in from the original URL?

lesleyjanenorton:
@irrationalagent it's currently the `SERVER_URL` environment variable with the `https://`clipped out. On prod this works out to `monitor.firefox.com`. I can manually set `utm_source` to be `monitor.firefox.com` for each environment if that would be better?

controllers/oauth.js
```diff
@@ -35,16 +35,20 @@ const FxAOAuthClient = new ClientOAuth2({
...
   // the user when they're redirected back to us after auth.
   const state = crypto.randomBytes(40).toString("hex");
   req.session.state = state;
   const url = new URL(client.code.getUri({state}));
+  const fxaParams = new URL(req.url, AppConstants.SERVER_URL);
+
   url.searchParams.append("access_type", "offline");
   url.searchParams.append("action", "email");
-  if (req.query.scanned) {
-    url.searchParams.append("email", req.query.scanned);
```
groovecoder: Hmm ... I thought I saw a bug where this makes the url parameter in the redirect to `stable.dev.lcip.org/oauth/signin?`...`scanned=<email>` ... which breaks the pre-population logic on the FxA page:

image

But I can't reproduce it. So I won't block this but let's keep it in mind in case we see this bug happen during today's deploy.

PR #817, opened 2019-02-27
Implement a script to add breaches missing from Remote Settings.

nhnt11:
This is my script to diff the breaches in the fxmonitor-breaches Remote Settings collection against HIBP's breach list, and add the missing ones.

I modified my script to use `got` which is already a dependency in this repo.

The post request to actually add records is untested at the moment after converting from `request-promise` to `got`.

This needs to be updated to use credentials from the environment instead of a pasted bearer token.

### DIFF
--- a/scripts/updatekinto.js
+++ b/scripts/updatekinto.js
@@ -1,10 +1,21 @@
 "use strict";
 
+const DEBUG = true;
+
 const got = require("got");
 
 const SERVER = "https://settings-writer.prod.mozaws.net/v1";
 const CID = "fxmonitor-breaches";
-const DEBUG = true;
+/* eslint-disable-next-line no-process-env */
+const USERNAME = process.env.KINTO_USERNAME;
+/* eslint-disable-next-line no-process-env */
+const PASSWORD = process.env.KINTO_PASSWORD;
+if (!USERNAME || !PASSWORD) {
+  console.error("Please set credentials in the environment.");
+  return;
+}
+
+const AUTH = Buffer.from(`${USERNAME}:${PASSWORD}`).toString("base64");
 
 async function run() {
   const RemoteSettingsBreachesSet = new Set((await got(
@@ -48,7 +59,7 @@ async function run() {
       await got.post(`${SERVER}/buckets/main-workspace/collections/${CID}/records`, {
         headers: {
           "Content-Type": "application/json",
-          "authorization": "Bearer <insert token here>",
+          "authorization": `Basic ${AUTH}`,
         },
         body: JSON.stringify({data: data}),
       });
--- a/scripts/updatekinto.js
+++ b/scripts/updatekinto.js
@@ -1,0 +1,62 @@
+"use strict";
+
+const got = require("got");
+
+const SERVER = "https://settings-writer.prod.mozaws.net/v1";
+const CID = "fxmonitor-breaches";
+const DEBUG = true;
+
+async function run() {
+  const RemoteSettingsBreachesSet = new Set((await got(
+    "https://firefox.settings.services.mozilla.com/v1/buckets/main/collections/fxmonitor-breaches/records",
+    { json: true }
+  )).body.data.map(b => b.Name));
+
+  const hibp_breaches = (await got(
+    "https://haveibeenpwned.com/api/v2/breaches",
+    { json: true }
+  )).body;
+
+  const new_breaches = [];
+
+  for (const breach of hibp_breaches) {
+    if (breach.IsSpamList || breach.IsRetired || !breach.IsVerified || !breach.Domain) {
+      continue;
+    }
+
+    if (RemoteSettingsBreachesSet.has(breach.Name)) {
+      continue;
+    }
+
+    new_breaches.push(breach);
+  }
+  for (const breach of new_breaches) {
+    const data = {
+      Name: breach.Name,
+      Domain: breach.Domain,
+      BreachDate: breach.BreachDate,
+      PwnCount: breach.PwnCount,
+      AddedDate: breach.AddedDate,
+    };
+
+    if (DEBUG) {
+      console.log(data);
+      continue;
+    }
+
+    try {
+      await got.post(`${SERVER}/buckets/main-workspace/collections/${CID}/records`, {
+        headers: {
+          "Content-Type": "application/json",
+          "authorization": "Bearer <insert token here>",
+        },
+        body: JSON.stringify({data: data}),
+      });
+    } catch (e) {
+      console.log(e);
+      return;
+    }
+  }
+}
+
+run();
--- a/scripts/updatekinto.js
+++ b/scripts/updatekinto.js
@@ -1,17 +1,21 @@
 "use strict";
 
-const DEBUG = true;
+
+/* eslint-disable no-process-env */
+
+const DEBUG = !process.env.PUSH_TO_KINTO;
 
 const got = require("got");
 
-const SERVER = "https://settings-writer.prod.mozaws.net/v1";
-const CID = "fxmonitor-breaches";
-/* eslint-disable-next-line no-process-env */
+const KINTO_UPDATE_ENDPOINT = "https://settings-writer.prod.mozaws.net/v1/buckets/main-workspace/collections/fxmonitor-breaches/records";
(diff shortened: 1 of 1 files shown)


scripts/updatekinto.js
```diff
@@ -0,0 +1,62 @@
...
+    if (DEBUG) {
+      console.log(data);
+      continue;
+    }
+
+    try {
+      await got.post(`${SERVER}/buckets/main-workspace/collections/${CID}/records`, {
+        headers: {
+          "Content-Type": "application/json",
+          "authorization": "Bearer <insert token here>",
```
pdehaan: I saw your commit message,

> _"This needs to be updated to use credentials from the environment instead of a pasted bearer token."_

But not sure if we're planning on doing that in this PR or a future PR and we should slap a `// TODO` comment on this. Although I'm not sure if we're really tracking TODO comments in this repo or if we should remove that ESLint rule, since it may just be adding noise to the CI logs.

scripts/updatekinto.js
```diff
@@ -0,0 +1,62 @@
...
+        },
+        body: JSON.stringify({data: data}),
+      });
+    } catch (e) {
+      console.log(e);
+      return;
```
pdehaan: Should we do a `process.exit(1)` or `process.exitCode = 1` or something here so we exit with a non-zero status?

scripts/updatekinto.js
```diff
@@ -0,0 +1,73 @@
...
+if (!USERNAME || !PASSWORD) {
+  console.error("Please set credentials in the environment.");
+  return;
+}
+
+const AUTH = Buffer.from(`${USERNAME}:${PASSWORD}`).toString("base64");
+
+async function run() {
+  const RemoteSettingsBreachesSet = new Set((await got(
+    "https://firefox.settings.services.mozilla.com/v1/buckets/main/collections/fxmonitor-breaches/records",
```
pdehaan: Wasn't sure if this should be added as a `const` above since it looks semi-similar to the `SERVER` const.

Or if we should disambiguate with a "READ_SERVER" and "WRITE_SERVER", and this may turn into:

```js
const READ_SERVER = "https://firefox.settings.services.mozilla.com/v1";
const WRITE_SERVER = "https://settings-writer.prod.mozaws.net/v1";
const CID = "fxmonitor-breaches";

//...

`${READ_SERVER}/buckets/main/collections/${CID}/records`,

//...

await got.post(`${WRITE_SERVER}/buckets/main-workspace/collections/${CID}/records`, {
```

scripts/updatekinto.js
```diff
@@ -0,0 +1,73 @@
...
+
+    new_breaches.push(breach);
+  }
+  for (const breach of new_breaches) {
+    const data = {
+      Name: breach.Name,
+      Domain: breach.Domain,
+      BreachDate: breach.BreachDate,
+      PwnCount: breach.PwnCount,
+      AddedDate: breach.AddedDate,
```
pdehaan: Do we need others like `.LogoPath` and `.DataClasses` and `.Title`?

https://github.com/mozilla/blurts-server/blob/a7a5730bf0bd83d6183c185f21fde70e98b99415/hibp.js#L88-L89

scripts/updatekinto.js
```diff
@@ -0,0 +1,73 @@
...
+
+const SERVER = "https://settings-writer.prod.mozaws.net/v1";
+const CID = "fxmonitor-breaches";
+/* eslint-disable-next-line no-process-env */
+const USERNAME = process.env.KINTO_USERNAME;
+/* eslint-disable-next-line no-process-env */
+const PASSWORD = process.env.KINTO_PASSWORD;
+if (!USERNAME || !PASSWORD) {
+  console.error("Please set credentials in the environment.");
+  return;
```
pdehaan: Possibly add a non-zero exit code here? `process.exitCode = 1;`. Not that I imagine we'll run this script in CI or anything, but if it gets run on some sort of cron or OPs script, it may be nice knowing it hasn't been silently failing.

scripts/updatekinto.js
pdehaan: I guess since we're updating Kinto in a loop here, it would make more sense to use `process.exitCode` otherwise `.exit(1)` will abort after the first error.

Not sure if there is a way to just push all new records up to Kinto in a single request, or if it's worth the effort.

scripts/updatekinto.js
```diff
@@ -0,0 +1,73 @@
...
+  for (const breach of hibp_breaches) {
+    if (breach.IsSpamList || breach.IsRetired || !breach.IsVerified || !breach.Domain) {
+      continue;
+    }
+
+    if (RemoteSettingsBreachesSet.has(breach.Name)) {
+      continue;
+    }
+
+    new_breaches.push(breach);
```
pdehaan: We're pushing the original HIBP breach object into the `new_breaches` array here, and then seemingly using a subset of the fields below on L46-50. Not sure if we should just store the 5-8 keys we actually care about here:

```js
new_breaches.push({
  Name: breach.Name,
  Domain: breach.Domain,
  BreachDate: breach.BreachDate,
  PwnCount: breach.PwnCount,
  AddedDate: breach.AddedDate,
  // See note below
  LogoPath: breach.LogoPath,
  DataClasses: breach.DataClasses,
  Title: breach.Title
});
```

scripts/updatekinto.js
nhnt11: There is a batch mode, but there's a limit on how many records you can upload per batch if I remember correctly - I tested that like a month or two ago.

I don't think it's worth the effort, since this will likely run something like every day and I'd expect either 0 or 1 new breach added per run MOST of the time.

nhnt11: > Should we do a `process.exit(1)` or `process.exitCode = 1` or something here so we exit with a non-zero status?

Yes indeed, proper exit codes is on my to-do list for this WIP. Thanks!

scripts/updatekinto.js
nhnt11: Nope, our Kinto collection only maintains these 5 properties. We can add new properties if/when the need arises - the data is sync'd with Firefox clients and I'd like to avoid having users download unnecessary data.

scripts/updatekinto.js
nhnt11: Yup, this would work. I'd like to keep the explicit fields in the post request though, which will prevent us from accidentally sending extra data in the future if we make changes. (just as a human-error safety net)

scripts/updatekinto.js
pdehaan: > I don't think it's worth the effort, since this will likely run something like every day and I'd expect either 0 or 1 new breach added per run MOST of the time.

Yeah, it definitely looks like having 2+ breaches per day is rare.
With a daily high of 8 on Nov 7 2016!

```
total breaches: 345

48 days had 2+ breaches added
12 days had 3+ breaches added
 4 days had 4+ breaches added
 2 days had 5+ breaches added

12/27/2016 ==> 4
11/7/2016 ==> 8
 9/5/2016 ==> 4
12/3/2013 ==> 5
```

<details>
<summary>source</summary>

```js
const { getBreaches } = require("monitor-breaches-axios");

async function main(minPerDay = 2) {
  const { breaches } = await getBreaches();
  const sortedBreaches = breaches.sort(
    (breachA, breachB) => breachB.AddedDate - breachA.AddedDate
  );
  const breachDays = {};
  sortedBreaches.forEach(breach => {
    const day = breach.AddedDate.toLocaleDateString();
    const res = breachDays[day] || [];
    breachDays[day] = [...res, breach.Title];
  });

  let count = 0;
  Object.entries(breachDays).forEach(([key, value]) => {
    if (value.length >= minPerDay) {
      console.log(key, "==>", value.length);
      count++;
    }
  });
  console.log(`\n${count} days had ${minPerDay}+ breaches added`);
  console.log(`total breaches: ${breaches.length}`);
}

main(2);
```
</details>

nhnt11: Nice, thanks for verifying my claim!

nhnt11:
I've created a new repo for this script to live in: https://github.com/mozilla/fxmonitor-remotesettings-updater

Please see my comment here for my reasoning: https://github.com/mozilla/fxmonitor-remotesettings-updater/pull/1#issue-257469657

PR #829, opened 2019-03-12
use FXA_SETTINGS_URL for manage fxa link

groovecoder:


groovecoder:
Merging with tests failing on the new env variable until I can fix that later.

Issue #465, opened 2018-09-26
Publish HIBP breach data to Remote Settings service


groovecoder:
See https://mana.mozilla.org/wiki/pages/viewpage.action?pageId=66655528 for getting access to publish to kinto-settings/Remote Settings.

See https://kinto.readthedocs.io/en/latest/tutorials/first-steps.html for docs on publishing.

pdehaan:
From Slack chat today:

@pdehaan says:
> @groovecoder is the plan to have the server use the same Kinto/RemoteSettings endpoint as the add-on? or is that a V2/future thing?

@groovecoder:
> Interesting … do you know if Remote Settings supports web(site) clients?

@pdehaan: 
> Not sure. I was reading https://remote-settings.readthedocs.io/ and thought RemoteSettings was just a wrapper around Kinto. But maybe it isnt that simple.

> I just figured if it was possible, it may be easiest to all pull the breach data from the same bucket so things are always in sync, and the add-on doesn’t know about a breach before the server.

@nhnt11:
> great idea I think. To be fair, the way it's set up now, the server is the one that has chances of knowing about breaches before the addon. We don't have automation in place to update the kinto collection yet. But I think it would be a good long term setup to make a bot that syncs kinto, and the addon as well as the server can rely on kinto as the source of truth. We don't need to change much, I think. The kinto endpoint serves the data as JSON very similar to HIBP. One change we'll need to make is the actual data we're storing on kinto- currently it omits stuff like data classes that the website depends on.

---

I know [**mozilla/notes**](https://github.com/mozilla/notes) uses [**kinto-http**](https://github.com/Kinto/kinto-http.js/blob/master/README.md), but not sure about if RemoteSettings and kinto-http are interchangeable.

groovecoder:
We could do this in [hibp.js](https://github.com/mozilla/blurts-server/blob/master/controllers/hibp.js) when HIBP sends us breach load callbacks.

groovecoder:
Based on today's meeting, we will add this script to our repo and have Ops set up a scheduled job to run it.

sandysage:
@nhnt11 now that the FxA integration has landed, can we pick this back up? I _believe_ we have some working code that would address this. Is there already a branch or pull request?

groovecoder:
This is done early last month.

PR #855, opened 2019-04-04
for #465: start updatebreaches.js script

groovecoder:


scripts/updatebreaches.js
```diff
@@ -0,0 +1,97 @@
...
+
+const got = require("got");
+
+const AppConstants = require("../app-constants");
+const HIBP = require("../hibp");
+
+
+const FX_RS_RECORDS = `${AppConstants.FX_REMOTE_SETTINGS_WRITER_SERVER}/buckets/main-workspace/collections/fxmonitor-breaches/records`;
+const FX_RS_COLLECTION = `${AppConstants.FX_REMOTE_SETTINGS_WRITER_SERVER}/buckets/main-workspace/collections/fxmonitor-breaches`;
+const FX_RS_PREVIEW_COLLECTION = `${AppConstants.FX_REMOTE_SETTINGS_WRITER_SERVER}/buckets/main-preview/collections/fxmonitor-breaches`;
```
nhnt11: Might want to put our collection name in a const as well and re-use it everywhere.

scripts/updatebreaches.js
```diff
@@ -0,0 +1,97 @@
...
+
+
+if (!FX_RS_USERNAME || !FX_RS_PASSWORD) {
+  console.error("updatebreaches requires FX_REMOTE_SETTINGS_USERNAME and FX_REMOTE_SETTINGS_PASSWORD.");
+  process.exitCode = 1;
+  return;
+}
+
+
+function whichBreachesAreNotInRemoteSettingsYet(remoteSettingsBreachesSet, hibpBreaches) {
```
nhnt11: emotional nit: Overall this function feels a bit strange because the two lists it's diff'ing aren't expected to be of the same type, and it doesn't handle the sourcing of the data so technically it can be used to compare any two lists of breaches. I'd expect either for this function to source the data instead of accepting arguments, or for this code to be inline'd. Up to you!

groovecoder: Ah that's a good point ... and I requested a similar refactor on another project to make a function always *return* the same data *type*, so I totally agree. Hmm ...

My main motivation here was to break things up into smaller functions with fewer logic branches (`if`, `for` statements, etc.). To get the same effect, I'll move the hibp breaches call above the call to this function, and only pass in the hibp breaches. So, this function will handle sourcing the data from Remote Settings *and* returning only the HIBP breaches that aren't already there.

Good catch!

scripts/updatebreaches.js
```diff
@@ -0,0 +1,97 @@
...
+  const newBreaches = [];
+  for (const breach of hibpBreaches) {
+    if (breach.IsSpamList || breach.IsRetired || !breach.IsVerified || !breach.Domain) {
+      continue;
+    }
+    if (remoteSettingsBreachesSet.has(breach.Name)) {
+      continue;
+    }
+    newBreaches.push(breach);
+  }
```
pdehaan: Do we need to `return newBreaches;` somewhere?
I see we have a `const newBreaches = whichBreachesAreNotInRemoteSettingsYet(remoteSettingsBreachesSet, hibpBreaches);` on L45 below, but not sure how that wouldn't always be `undefined`, unless I'm missing a trick somewhere.

scripts/updatebreaches.js
```diff
@@ -0,0 +1,97 @@
...
+  console.error("updatebreaches requires FX_REMOTE_SETTINGS_USERNAME and FX_REMOTE_SETTINGS_PASSWORD.");
+  process.exitCode = 1;
+  return;
+}
+
+
+function whichBreachesAreNotInRemoteSettingsYet(remoteSettingsBreachesSet, hibpBreaches) {
+  const newBreaches = [];
+  for (const breach of hibpBreaches) {
+    if (breach.IsSpamList || breach.IsRetired || !breach.IsVerified || !breach.Domain) {
```
pdehaan: Do we need to check `breach.IsFabricated`, per:

https://github.com/mozilla/blurts-server/blob/39e81fecfd01ff8b57aa2b3005b00b2a9aea7272/hibp.js#L143-L151

I'd say bonus points if we could reuse that filter function instead of duplicating the logic in 2+ places, but I agree w/ @nhnt11 that this may be better suited for it's own repo so that you don't have to clone+install all the server dependencies just to run this one script.

scripts/updatebreaches.js
groovecoder: Good catch. I caught that while addressing @nhnt11's comments too.

scripts/updatebreaches.js
groovecoder: Good call!

groovecoder:
Updated for comments. Still need to set up my own test & debugging runs.

scripts/updatebreaches.js
```diff
@@ -0,0 +1,97 @@
...
+
+  if (!newBreaches) {
+    console.log("No new breaches found.");
+    return;
+  }
+
+  console.log(`${newBreaches.length} new breach(es) found.`);
+
+  const FX_RS_AUTH = Buffer.from(`${FX_RS_USERNAME}:${FX_RS_PASSWORD}`).toString("base64");
+  const remoteSettingsPreviewSet = new Set((await got(
```
groovecoder: Not to self: This should also move into `whichBreachesAreNotInRemoteSettingsYet`

scripts/updatebreaches.js
leplatrem: Couldn't all this written like:
```js
const newBreaches = hibpBreaches.filter(({ Name }) => !remoteSettingsBreachesSet.has(Name))
```
?

scripts/updatebreaches.js
```diff
@@ -0,0 +1,97 @@
...
+      PwnCount: breach.PwnCount,
+      AddedDate: breach.AddedDate,
+    };
+
+    try {
+      // Create the record
+      await got.post(FX_RS_RECORDS, {
+        headers: {
+          "Content-Type": "application/json",
+          "authorization": `Basic ${FX_RS_AUTH}`,
```
leplatrem: For `Basic` authentication to work, you'll need to ask for the creation of a Remote Settings account (eg. https://bugzilla.mozilla.org/show_bug.cgi?id=1512466).
If you plan on using your personal credentials to run this script, you'll have to use Bearer tokens (see [FAQ](https://remote-settings.readthedocs.io/en/latest/frequently-asked-questions.html#how-do-i-automate-the-publication-of-records))

lib/remote-settings.js
```diff
@@ -0,0 +1,51 @@
...
+    return breaches.filter( ({Name}) => !remoteSettingsBreachesSet.has(Name) );
+  },
+
+  async postNewBreachToBreachesCollection(data) {
+    // Create the record
+    await got.post(FX_RS_RECORDS, {
+      headers: {
+        "Content-Type": "application/json",
+        "authorization": `Bearer ${FX_RS_BEARER_TOKEN}`,
+      },
```
pdehaan: Not sure if it helps here or not, but it looks like we can create a **got** client by using [`got.extend()`](https://github.com/sindresorhus/got#gotextendoptions) and possibly get away w/ declaring the `Content-Type`, `Authorization` headers and `json: true` in a single place [maybe].

```js
const client = got.extend({
  json: true,
  headers: {
    "Content-Type": "application/json", // possibly unneeded if we define `json: true` above.
    "Authorization": `Bearer ${FX_RS_BEARER_TOKEN}`
  }
});

// 1
const fxRSRecords = await client(FX_RS_RECORDS);
const remoteSettingsBreachesSet = new Set(fxRSRecords.body.data.map(b => b.Name));

// 2
await client.post(FS_RS_RECORDS, {
  body: JSON.stringify({data})
});

// 3
await client.patch(FS_RS_COLLECTION, {
  body: JSON.stringify({data: {status: "to-review"}})
});
```

---

This worked locally, but I had to do some mild tweaking since I didn't have a bearer token or know the exact `FX_REMOTE_SETTINGS_WRITER_SERVER` URL. Not sure if we should add that to the .env-dist file or somethin'.

<details>
<summary>Expand</summary>

```js
const got = require("got");

const AppConstants = {
  FX_REMOTE_SETTINGS_WRITER_SERVER: "https://firefox.settings.services.mozilla.com/v1"
};
const BREACHES_COLLECTION = "fxmonitor-breaches";

const FX_RS_COLLECTION = `${AppConstants.FX_REMOTE_SETTINGS_WRITER_SERVER}/buckets/main/collections/${BREACHES_COLLECTION}`;
const FX_RS_RECORDS = `${FX_RS_COLLECTION}/records`;

const client = got.extend({
  json: true,
  headers: {
    "Content-Type": "application/json"
  }
});

main();

async function main() {
  const fxRSRecords = await client(FX_RS_RECORDS);
  const remoteSettingsBreachesSet = new Set(fxRSRecords.body.data.map(b => b.Name).sort());
  console.log(remoteSettingsBreachesSet);
}
```
</details>

groovecoder: I filed https://github.com/mozilla/blurts-server/issues/859 for this. I'd like to keep the script as-is for merging, because I just ran it and it added the new Bukalapak breach perfectly.

Issue #863, opened 2019-04-22
Add "breaches_last_shown" column to database


groovecoder:
When the user visits the home page, we want to show them new breaches.

To do this, we need to store a new `subscribers.breaches_last_shown` timestamp value in the DB.

PR #866, opened 2019-04-23
wip for #863: add breaches_last_shown to DB

groovecoder:
Still to do:

* [x] Update `subscribers.breaches_last_shown` when a signed-in user visits the home page which shows their "new" breaches.

### DIFF
--- a/controllers/user.js
+++ b/controllers/user.js
@@ -111,14 +111,19 @@ async function getAllEmailsAndBreaches(user, allBreaches) {
 
 
 async function getDashboard(req, res) {
+  _requireSessionUser(req);
   const allBreaches = req.app.locals.breaches;
   const user = req.session.user;
   const { verifiedEmails, unverifiedEmails } = await getAllEmailsAndBreaches(user, allBreaches);
+
+  const newBreachesFound = allBreaches.filter(breach => breach.AddedDate >= user.breaches_last_shown);
+  await DB.setBreachesLastShownNow(user);
 
   res.render("dashboards", {
     title: req.fluentFormat("user-dash"),
     verifiedEmails,
     unverifiedEmails,
+    newBreachesFound,
     whichPartial: "dashboards/breaches-dash",
   });
 }
@@ -219,11 +224,11 @@ async function postUnsubscribe(req, res) {
 
 
 async function getPreferences(req, res) {
+  _requireSessionUser(req);
   const allBreaches = req.app.locals.breaches;
   const user = req.session.user;
   const { verifiedEmails, unverifiedEmails } = await getAllEmailsAndBreaches(user, allBreaches);
 
-  _requireSessionUser(req);
   res.render("dashboards", {
     title: req.fluentFormat("email-add-title"),
     whichPartial: "dashboards/preferences",
--- a/controllers/user.js
+++ b/controllers/user.js
@@ -110,14 +110,27 @@ async function getAllEmailsAndBreaches(user, allBreaches) {
 }
 
 
+function getNewBreachesForEmailEntriesSinceDate(emailEntries, date) {
+  const breaches = new Set();
+  for (const emailEntry of emailEntries) {
+    const newBreachesForEmail = emailEntry.breaches.filter(breach => breach.AddedDate >= date);
+    for (const newBreachForEmail of newBreachesForEmail) {
+      breaches.add(newBreachForEmail);
+    }
+  }
+  return breaches;
+}
+
+
 async function getDashboard(req, res) {
   _requireSessionUser(req);
   const allBreaches = req.app.locals.breaches;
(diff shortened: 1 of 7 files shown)


db/DB.js
```diff
@@ -212,6 +212,19 @@ const DB = {
     return updatedSubscriber;
   },
 
+  async setBreachesLastShownNow(subscriber) {
+    const nowDateTime = Date.now();
+    const nowTimeStamp = nowDateTime.toISOString();
```
pdehaan: Curious... I don't think this is a thing.
[`Date.now()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/now) returns an integer. And it looks like [`.toISOString()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toISOString) is a method on the Date object.

This was giving me the following error when trying to run in a devtools console or in Node directly:

```
// const nowTimeStamp = nowDateTime.toISOString();
//                                  ^

// TypeError: nowDateTime.toISOString is not a function
//     at Object.<anonymous> (/Users/pdehaan/dev/github/mozilla/fingerprinting-cryptomining-retention-study/tmp.js:2:34)
//     at Module._compile (internal/modules/cjs/loader.js:701:30)
//     at Object.Module._extensions..js (internal/modules/cjs/loader.js:712:10)
//     at Module.load (internal/modules/cjs/loader.js:600:32)
//     at tryModuleLoad (internal/modules/cjs/loader.js:539:12)
//     at Function.Module._load (internal/modules/cjs/loader.js:531:3)
//     at Function.Module.runMain (internal/modules/cjs/loader.js:754:12)
//     at startup (internal/bootstrap/node.js:283:19)
//     at bootstrapNodeJSCore (internal/bootstrap/node.js:622:3)
```

Workarounds would be:

1. Wrap the `nowDateTime` in a `new Date()` constructor:

    ```js
    const nowDateTime = Date.now();
    // const nowTimeStamp = nowDateTime.toISOString();
    const nowTimeStamp = new Date(nowDateTime).toISOString();
    ```

1. Use the Date constructor and one line it:

    ```js
    const nowTimeStamp = new Date().toISOString();
    console.log(nowTimeStamp);
    ```

groovecoder: Ah yes, meant to restore `new Date()` in there somewhere.

pdehaan:
Submitted #867 for the CI dependency failures.
I noticed we also have a failing test in the CI runs:

### error: column "email" of relation "subscribers" does not exist

---

> blurts-server@0.0.1 test:db:seed /home/travis/build/mozilla/blurts-server
> NODE_ENV=tests knex seed:run --knexfile db/knexfile.js --env tests
>
> Working directory changed to ~/build/mozilla/blurts-server/db
> Using environment: tests

```
error: column "email" of relation "subscribers" does not exist
    at Connection.parseE (/home/travis/build/mozilla/blurts-server/node_modules/pg/lib/connection.js:553:11)
    at Connection.parseMessage (/home/travis/build/mozilla/blurts-server/node_modules/pg/lib/connection.js:378:19)
    at Socket.<anonymous> (/home/travis/build/mozilla/blurts-server/node_modules/pg/lib/connection.js:119:22)
    at Socket.emit (events.js:193:13)
    at addChunk (_stream_readable.js:295:12)
    at readableAddChunk (_stream_readable.js:276:11)
    at Socket.Readable.push (_stream_readable.js:231:10)
    at TCP.onStreamRead (internal/stream_base_commons.js:150:17)
```

db/DB.js
```diff
@@ -212,6 +212,19 @@ const DB = {
     return updatedSubscriber;
   },
 
+  async setBreachesLastShownNow(subscriber) {
+    const nowDateTime = new Date(Date.now());
```
pdehaan: I believe `new Date()` returns the current date, so you shouldn't need the `Date.now()`.

controllers/user.js
```diff
@@ -111,14 +111,19 @@ async function getAllEmailsAndBreaches(user, allBreaches) {
 
 
 async function getDashboard(req, res) {
+  _requireSessionUser(req);
   const allBreaches = req.app.locals.breaches;
   const user = req.session.user;
   const { verifiedEmails, unverifiedEmails } = await getAllEmailsAndBreaches(user, allBreaches);
 
+  const newBreachesFound = allBreaches.filter(breach => breach.AddedDate >= user.breaches_last_shown);
```
pdehaan: Just so I'm clear, is the intent here to display all new HIBP breaches since the user's last visit, and not just new breaches that the user was involved in?

lesleyjanenorton: Should be only new breaches that the user was involved in so we'll just need to loop through their `foundBreaches` instead of `allBreaches`.

PR #882, opened 2019-05-06
Adds new strings for V2 website.

lesleyjanenorton:
After talking to @flodolo, the plan is to add the new strings to the existing .ftl file and clean out the old strings after the V2 launch. This will likely be less problematic than attempting to merge separate .ftl files later.

### DIFF
--- a/locales/en/app.ftl
+++ b/locales/en/app.ftl
@@ -19,7 +19,7 @@ layout-support = Support
 about-firefox-alerts = About Firefox Alerts
 # Link that takes the user to a Firefox Monitor survey. 
 give-feedback = Give Feedback
-terms-and-privacy = Terms and Privacy
+terms-and-privacy = Terms & Privacy
 
 error-scan-page-token = You tried to scan too many email addresses in a short time period. For security reasons, we’ve temporarily blocked you from new searches. You’ll be able to try again later.
 error-could-not-add-email = Could not add email address to database.
@@ -464,7 +464,6 @@ frequently-asked-questions=Frequently Asked Questions
 
 about-firefox-monitor = About {-product-name}
 mozilla-dot-org = Mozilla.org
-terms-and-privacy-v2 = Terms & Privacy
 
 preferences = Preferences
 
@@ -601,7 +600,8 @@ feat-enroll-multiple = Enroll multiple emails in breach monitoring
 
 sign-up-for-fxa = Sign up for a {-brand-fxa}
 
-
+# This string is shown beneath each of the user’s email addresses to indicate
+# how many known breaches that email address was found in. 
 appears-in-x-breaches =
   { $breachCount ->
         [one] Appears in { $breachCount } known breach.
@@ -762,3 +762,18 @@ passwords-exposed =
 
 # Button
 see-additional-breaches = See Additional Breaches
+
+# A button on the All Breaches page that restores all of the breaches
+# back to the page if the user has filtered some of them out.
+see-all-breaches = See All Breaches
+
+scan-results-known-breaches =
+  { $breachCount ->
+      [one] This email appeared in 1 known data breach.
+     *[other] This email appeared in { $breachCount } known data breaches.
+  }
+
+
+# This string is shown at the top of the scan results page and indicates
+# the email that was searched.
+results-for-x = Results for { $userEmail }
--- a/locales/en/app.ftl
+++ b/locales/en/app.ftl
@@ -464,7 +464,7 @@ frequently-asked-questions=Frequently Asked Questions
 
 about-firefox-monitor = About {-product-name}
 mozilla-dot-org = Mozilla.org
-terms-and-privacy = Terms & Privacy
+terms-and-privacy-v2 = Terms & Privacy
 
 preferences = Preferences
 
@@ -484,7 +484,7 @@ fxa-account = {-brand-fxa}
 open-fxa-menu = Open {-brand-fxa} navigation
 
 # Appears above a snippet about the breach most recently reported to Firefox Monitor.
-latest-breach= LATEST BREACH ADDED
+latest-breach = LATEST BREACH ADDED
 
 breach-added = Breach reported:
 breach-discovered = Breach discovered:
@@ -536,8 +536,8 @@ delayed-reporting-copy = It can sometimes take months or years for credentials e
 about-fxm-headline = About {-product-name}
 about-fxm-blurb = {-product-name} warns if your online accounts were involved in a 
   data breach. Find out if you’ve been in a data breach, get alerts about new breaches, 
-  and take steps to protect your online accounts. {-product-name} is a service 
-  is provided by {-brand-Mozilla}.
+  and take steps to protect your online accounts. {-product-name} is provided 
+  by {-brand-Mozilla}.
 
 fxm-warns-you = {-product-name} warns you if your email address has been exposed 
   in an online data breach. See if your information has been exposed, learn how 
@@ -545,7 +545,7 @@ fxm-warns-you = {-product-name} warns you if your email address has been exposed
   appears in a new breach.
 
 
-## How FXM Works
+# How Firefox Monitor works
 how-fxm-works = How {-product-name} works
 
 how-fxm-1-headline = Conduct a basic search
@@ -593,8 +593,6 @@ stop-reusing-pw = Stop reusing the same passwords
 create-unique-pw = Create unique passwords and save them somewhere safe, like a password manager.
 five-myths = 5 myths about password managers
 
-show-all-breaches = Show All Breaches
-
 create-a-fxa = Create a {-brand-fxa} for your full report of breaches and to get alerts.
 
 feat-security-tips = Security tips to protect your accounts
@@ -607,18 +605,18 @@ sign-up-for-fxa = Sign up for a {-brand-fxa}
 appears-in-x-breaches =
   { $breachCount ->
         [one] Appears in { $breachCount } known breach.
-       *[other] appears in { $breachCount } known breaches.
+       *[other] Appears in { $breachCount } known breaches.
   }
 
 
 see-if-breached = See if you’ve been in an online data breach.
-check-for-breaches = Check for breaches.
+check-for-breaches = Check for breaches
 find-out-what-hackers-know = Find out what hackers already know about you. Learn how to stay a step ahead of them.
 search-for-your-email = Search for your email address in public data breaches back to 2007.
 
 back-to-top = Back to Top
 
-comm-opt-0 = Email me if one of my email addreseses below appears in a data breach.
+comm-opt-0 = Email me if one of my email addresses below appears in a data breach.
 comm-opt-1 = Send all breach alerts to { $primaryEmail }.
 
 stop-monitoring-this = Stop monitoring this email.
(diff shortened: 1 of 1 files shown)


locales/en/app.ftl
```diff
@@ -459,3 +459,305 @@ get-your-report-and-sign-up= Get your report and sign up for alerts.
...
+security-tips = Security Tips
+
+fxa-account = {-brand-fxa}
+
+
+# Aria button message to open menu. "Open Firefox Account Navigation"
+open-fxa-menu = Open {-brand-fxa} navigation
+
+# Appears above a snippet about the breach most recently reported to Firefox Monitor.
+latest-breach= LATEST BREACH ADDED
```
pdehaan: nit: space before `=`

locales/en/app.ftl
```diff
@@ -459,3 +459,305 @@ get-your-report-and-sign-up= Get your report and sign up for alerts.
...
+get-alerted = Get alerted about new breaches.
+
+was-your-info-exposed = Was your info exposed in the { $breachName } data breach?
+find-out-if = Find out if your data was exposed in this breach.
+
+fb-not-comp = This email did not appear in the { $breachName } breach.
+fb-comp = This email appeared in { $breachCount } known data breaches, including { $breachName }. 
+fb-comp-only = This email appeared in the { $breachCount } breach.
+
+other-breaches-found = However, it did appear in
```
flodolo: Let's do this. Also, *two* is not a CLDR plural category.

```
other-breaches-found = 
  { $breachCount ->
    [one] However, it did appear in { $breachCount } other breach.
   *[other] However, it did appear in { $breachCount } other breaches. 
  }
```

locales/en/app.ftl
```diff
@@ -459,3 +459,305 @@ get-your-report-and-sign-up= Get your report and sign up for alerts.
...
+fb-comp = This email appeared in { $breachCount } known data breaches, including { $breachName }. 
+fb-comp-only = This email appeared in the { $breachCount } breach.
+
+other-breaches-found = However, it did appear in 
+  { $breachCount ->
+    [one] { $breachCount } other breach.
+   *[two] { $breachCount } other breaches. 
+  }
+
+fb-comp-and-others = This email appeared in { $breachCount } known data breaches, including { $breachName }.
```
flodolo: This needs a plural.

locales/en/app.ftl
```diff
@@ -459,3 +459,305 @@ get-your-report-and-sign-up= Get your report and sign up for alerts.
...
+no-other-breaches-found = No other breaches found from a basic search.
+
+no-results-blurb = Sorry, that breach is not in our database.
+all-breaches-headline = All breaches in {-product-name}
+search-breaches = Search Breaches
+
+# Tells the user which breach types are currently being displayed. In page, it would look like "Showing: All Breaches"
```
flodolo: What are the possible alternative values besides "All Breaches"?

locales/en/app.ftl
```diff
@@ -459,3 +459,305 @@ get-your-report-and-sign-up= Get your report and sign up for alerts.
...
+  emails {-product-name} sent you. Check your inbox for messages from 
+  {-brand-team-email}. Select the unsubscribe link at the bottom of the email.
+
+
+login-link-pre = Have an account?
+login-link = Log in
+
+
+# This string follows a large numeral that indicates the total number of email address a user has signed up for monitoring.
+# For instance, "7 Email addresses being monitored"
```
flodolo:
```
# This string is displayed under a large numeral that indicates the total number
# of email address a user has signed up for monitoring. Don't add $emails to
# your localization, because it would result in the number showing twice.
```

flodolo:
A few errors showing up:

```
    WARNING: sign-up-for-alerts occurs 2 times
    WARNING: terms-and-privacy occurs 2 times
    ERROR: sign-up-for-alerts occurs 2 times
    ERROR: terms-and-privacy occurs 2 times
    WARNING: Plural categories missing: other at line 703, column 6 for other-breaches-found
```

locales/en/app.ftl
```diff
@@ -459,3 +459,305 @@ get-your-report-and-sign-up= Get your report and sign up for alerts.
...
+
+appears-in-x-breaches =
+  { $breachCount ->
+        [one] Appears in { $breachCount } known breach.
+       *[other] appears in { $breachCount } known breaches.
+  }
+
+
+see-if-breached = See if you’ve been in an online data breach.
+check-for-breaches = Check for breaches.
```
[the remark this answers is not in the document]
lesleyjanenorton: Fixed, thanks!

locales/en/app.ftl
```diff
@@ -459,3 +459,305 @@ get-your-report-and-sign-up= Get your report and sign up for alerts.
...
+remove-fxm = Remove {-product-name}
+remove-fxm-blurb = Turn off {-product-name} alerts. Your {-brand-fxa} will remain active, and you may receive 
+  other account-related communications.
+
+
+manage-email-addresses = Manage Email Addresses
+
+latest-breach-link = See if you were in this breach
+
+welcome-back = Welcome back, { $userName }!
```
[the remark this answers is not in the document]
lesleyjanenorton: We don't. Is there a gender neutral greeting that would be appropriate for all genders?

locales/en/app.ftl
lesleyjanenorton: None, it disappears when the breaches have been filtered. It's broken up like this so that we add markup to the "Showing:" string. I'll update the comment.

locales/en/app.ftl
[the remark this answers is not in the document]
flodolo: No, weird forms like "Benvenuto/a". Anyhow, that's a problem in any other localization system, it was just wishful thinking.

locales/en/app.ftl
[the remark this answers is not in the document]
lesleyjanenorton: dupe!

locales/en/app.ftl
```diff
@@ -459,3 +459,305 @@ get-your-report-and-sign-up= Get your report and sign up for alerts.
...
+feat-sensitive = Advanced search in sensitive breaches
+feat-enroll-multiple = Enroll multiple emails in breach monitoring
+
+sign-up-for-fxa = Sign up for a {-brand-fxa}
+
+
+appears-in-x-breaches =
+  { $breachCount ->
+        [one] Appears in { $breachCount } known breach.
+       *[other] appears in { $breachCount } known breaches.
```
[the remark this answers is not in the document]
pdehaan: Not sure if this should be uppercase or lowercase. This was the only context I could find, and it looks like the string is prefixed by the user's email address. Should the email address be in this ftl string in case it should appear in a different order in different locales?

https://github.com/mozilla/blurts-server/blob/856d840348caad99c36e0f5abe260d53dca66e7c/views/partials/email-card.hbs#L3-L12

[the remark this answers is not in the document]
flodolo: Yes, the email should be part of the string. My point was that the two forms should be consistent.

locales/en/app.ftl
```diff
@@ -459,3 +459,305 @@ get-your-report-and-sign-up= Get your report and sign up for alerts.
...
+
+comm-opt-0 = Email me if one of my email addreseses below appears in a data breach.
+comm-opt-1 = Send all breach alerts to { $primaryEmail }.
+
+stop-monitoring-this = Stop monitoring this email.
+resend-verification = Resend verification email
+
+add-new-email = Add a new email address
+send-verification = Send Verification Link
+global-communication = Global communication
```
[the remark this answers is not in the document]
pdehaan: Seems to be a header on the https://monitor-v2.herokuapp.com/user/preferences page:

email-add-title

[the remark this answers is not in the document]
flodolo: OK, this helps. Can we add a comment? :)

locales/en/app.ftl
```diff
@@ -464,7 +464,7 @@ frequently-asked-questions=Frequently Asked Questions
 
 about-firefox-monitor = About {-product-name}
 mozilla-dot-org = Mozilla.org
-terms-and-privacy = Terms & Privacy
+terms-and-privacy-v2 = Terms & Privacy
```
pdehaan: Interesting... What's the difference between this and the previous string, or, why can't we use the same string in both places?

```sh
$ cat ./locales/en/app.ftl | grep "terms-and-privacy"

terms-and-privacy = Terms and Privacy
terms-and-privacy-v2 = Terms & Privacy
```

lesleyjanenorton: The ampersand in `terms-and-privacy-v2` is the only difference. After V2 is released, `terms-and-privacy` (no ampersand) will be retired. I think it makes more sense to just use the `terms-and-privacy` string we already have, but I need to double-check that content/ux does not have any strong feelings about this. Or ,@flodolo can we just update the existing `terms-and-privacy` string to use the ampersand for english only?

flodolo: > Or ,@flodolo can we just update the existing `terms-and-privacy` string to use the ampersand for english only?

Yes, that would be absolutely OK. In the end, it's just a styling thing for en-US

flodolo: Side note: you can move the existing string here, if it helps.

locales/en/app.ftl
```diff
@@ -459,3 +459,306 @@ get-your-report-and-sign-up= Get your report and sign up for alerts.
...
+create-a-fxa = Create a {-brand-fxa} for your full report of breaches and to get alerts.
+
+feat-security-tips = Security tips to protect your accounts
+feat-sensitive = Advanced search in sensitive breaches
+feat-enroll-multiple = Enroll multiple emails in breach monitoring
+
+sign-up-for-fxa = Sign up for a {-brand-fxa}
+
+
+appears-in-x-breaches =
```
flodolo: Can you add a comment to explain what's the subject, and how this string is displayed?

locales/en/app.ftl
flodolo: This still need to be address (we need a proper plural form, with ONE and OTHER for English).

lesleyjanenorton:
@flodolo I found a few previously undiscovered strings in Betsy's doc and had to modify another. Please let me know if there are any additional changes. Thank you!

locales/en/app.ftl
```diff
@@ -773,7 +773,20 @@ scan-results-known-breaches =
      *[other] This email appeared in { $breachCount } known data breaches.
   }
 
+# This string is shown at the top of the scan results page and is followed
+# by the email address that the user searched.
+# In page, it reads "Results for: searchedEmail@monitor.com"
+results-for = Results for
```
pdehaan: Should this have a placeholder email variable in case certain locales need to control order?
/cc @flodolo

locales/en/app.ftl
```diff
@@ -11,6 +11,7 @@
 -brand-HIBP = Have I Been Pwned
 -brand-fxa = Firefox Account
 -brand-Chrome = Chrome
+-brand-team-email = breach-alerts@mozilla.com
```
pdehaan: Where is this used?

flodolo: More important. There's no reason to make this localizable at all

locales/en/app.ftl
flodolo: The example in the comment is different: why there's a colon there, and not in the string?

But yes, a placeholder is safer.

locales/en/app.ftl
lesleyjanenorton: It's used in `error-invalid-unsub-blurb`.

PR #892, opened 2019-05-09
Fix moar v2 tests

groovecoder:


groovecoder:
Let's fix the others in additional PRs. Inching towards a green button for when we do `monitor_v2` -> `master` !!

PR #895, opened 2019-05-09
Update l10n linter to moz-l10n-lint

### LINKED ISSUE #888: Update l10n linter to `moz-l10n-lint`
I've updated `compare-locales --validate l10n.toml .` to `moz-l10n-lint l10n.toml`.

That's going to grow more features soon, and has better output.

I suggest to migrate to that.

Nit that might be nice to fix along, I made `l10n.toml` work w/out a `locales` field, and that has a different meaning now than `locales = []`. So I suggest to either remove the entry, or make the locale list right.

pdehaan:
Fixes #888 
/cc @Pike

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -31,7 +31,7 @@ matrix:
           install:
               - sudo pip install compare-locales
           script:
-              - compare-locales --validate l10n.toml .
+              - moz-l10n-lint l10n.toml
               - compare-locales l10n.toml . `ls locales`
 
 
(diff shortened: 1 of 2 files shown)


Pike:
Yeah, both problems reported by the linter are real.

LGTM, r+.

pdehaan:
Thanks @Pike 
I think both issues have been fixed in the [open] https://github.com/mozilla/blurts-server/pull/882 PR (**add-v2-strings** branch).

Issue #915, opened 2019-05-14
“Email addresses” string reused in different contexts


TheoChevalier:
“Email addresses” (adresses électroniques in the screenshots) is coming from data-classes.ftl and it doesn’t work because it’s two different contexts:

In data-classes: strings are concatenated as a coma separated list, and thus are not capitalized in French.
In preferences pane: used as a title, so should be capitalized.

Please add a new “Email addresses” string for the two titles below. Here the same string can be shared, because it’s the exact same context

![Screenshot_2019-05-14 Vérifiez si vous êtes concerné⋅e par une fuite de données](https://user-images.githubusercontent.com/1294206/57693370-f57b6500-7640-11e9-9a7f-d9dd281ce5e2.png)

![Screenshot_2019-05-14 Vérifiez si vous êtes concerné⋅e par une fuite de données(1)](https://user-images.githubusercontent.com/1294206/57693384-f9a78280-7640-11e9-9271-f034552ac495.png)

Issue #921, opened 2019-05-14
Missing l10n string for "additional-information-including"


pdehaan:
https://fx-breach-alerts.herokuapp.com/breach-details/VerificationsIO

![https___fx-breach-alerts_herokuapp_com_breach-details_VerificationsIO_and_Slack_-_Mozilla](https://user-images.githubusercontent.com/557895/57735126-7003af00-7658-11e9-9875-7e9ec92e5685.png)

Issue #919, opened 2019-05-14
Need text for when a breach list search has no results


pdehaan:
## Steps to reproduce

1. Go to https://fx-breach-alerts.herokuapp.com/breaches
2. Search for a random string, like "qwerty".

### Actual results
"empty" results page

### Expected results
A friendly error message saying something like "No matches found, please try again".
Uncertain if we should pre-populate the search form with the current search query.

lesleyjanenorton:
It's in there, just needs to be implemented. String ID is`no-results-blurb`. This is on my list for tomorrow, along with the other filtering quirks.

groovecoder:
Thanks for filing all these @pdehaan. I'm going thru and adding the "bug" label and Project to the ones we need to make sure to fix for launch.

pdehaan:
> I'm going thru and adding the "bug" label and Project to the ones we need to make sure to fix for launch.

Awesome, thanks! It's super helpful if you can flag all the issues that we want to fix for the v2 launch in a milestone or project column, or even just a single label so QA can get a sense of how many issues are left before launch to see if we're 98% of the way there, or 42%. 🍻

lesleyjanenorton:
Fixed in #956

Issue #875, opened 2019-05-03
A few missing data classes in locales/en/data-classes.ftl


pdehaan:
Spotted via https://monitor-v2.herokuapp.com/?breach=Appartoo which lists "employment-statuses"

```sh
$ npx pdehaan/blurts-server-data-classes

Missing "employment-statuses" data class from Appartoo breach
Missing "bios" data class from EyeEm breach
Missing "social-security-numbers" data class from GoldSilver breach
```

https://monitor.firefox.com/?breach=appartoo
https://monitor.firefox.com/?breach=eyeem
https://monitor.firefox.com/?breach=goldsilver

pdehaan:
Not sure if we should be verifying data classes whenever we import new breaches (and automatically adding them to the en/data-class.ftl file?) or if we should have some other front-end error handling when an  data class is encountered and then send an error to Sentry or something so we can add it manually.

PR #923, opened 2019-05-14
Update v2 strings

lesleyjanenorton:
@flodolo We have a number of string edits and a few new strings. We unanimously agree that we owe you champagne at Whistler.

Question:
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp; This string: 
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;`see-if-youve-been-part = See if you’ve been part of an online data breach.`

&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;Is very close to these strings:
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;`see-if-breached = See if you’ve been in an online data breach.`
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;`scan-label = See if you’ve been involved in a data breach.`

Could we update one of the old strings and avoid creating a new string ID for this?


pdehaan:
Not sure if we want to address any of these l10n issues at the same time:

- #875 &mdash; A few missing data classes in locales/en/data-classes.ftl
- #915 &mdash; “Email addresses” string reused in different contexts
- #919 &mdash; Need text for when a breach list search has no results
- #921 &mdash; Missing l10n string for "additional-information-including"

flodolo:
> Could we update one of the old strings and avoid creating a new string ID for this?

Either we use the existing string, or we create a new ID if you want the updated text. Your choice.

We should also fix the other l10n issues that Peter pointed out, and we should do it as soon as possible (like, yesterday). In particular the one pointed out by Théo, given French is a target locale.

Any update on pushing back the launch date one week? We really need that extra time.

lesleyjanenorton:
@flodolo It's attributable to a variety factors: 1.) Me missing things 2.) Discrepancies between the visual spec and the copy deck and 3.) The content strategist feeling that there were tweaks needed post "final copy" handover. I'll fix #915 and #875 and get those pushed up.

locales/en/data-classes.ftl
```diff
@@ -12,6 +12,7 @@ avatars = Avatars
 bank-account-numbers = Bank account numbers
 beauty-ratings = Beauty ratings
 biometric-data = Biometric data
+bios = Bios
```
flodolo: Is this "bios" and in "biographies", or something else? Can we get a comment?

locales/en/data-classes.ftl
```diff
@@ -101,6 +103,7 @@ smoking-habits = Smoking habits
 sms-messages = SMS messages
 social-connections = Social connections
 social-media-profiles = Social media profiles
+social-security-numbers = Social security numbers
```
pdehaan: SSNs sounds scary. Should we add it to the priority data classes list? https://github.com/mozilla/blurts-server/blob/d99aa8133628202a15ce2471869bf5e3714ed04f/template-helpers/breach-detail.js#L44

lesleyjanenorton: Definitely yes.

pdehaan: We can also do it in a separate pr, I just wanted to call it out while I saw it.
But if this PR is ready and has @flodolo’s blessing, I’d rather land it quickly and not block on the priority data class stuff.

flodolo: I'd really like this to land soon. But I also need a comment on the *bios* thing

pdehaan: @flodolo already done, see https://github.com/mozilla/blurts-server/pull/923/files#diff-bd84432b6e22fdc22c58a93217eeefa9

flodolo: Then yes, it should land ASAP

PR #935, opened 2019-05-15
Add breach-overview-string.

lesleyjanenorton:


locales/en/app.ftl
```diff
@@ -828,4 +828,7 @@ email-addresses-title = Email Addresses
 # This is a standardized breach overview blurb that appears on all breach detail pages.
 # $breachTitle is the name of the breached company or website.
 # $breachDate and $addedDate are calendar dates.
-On { $breachDate }, { $breachTitle } suffered a breach. Once the breach was discovered and verified it was added to our database on { $addedDate }.
+breach-overview = On { $breachDate }, { $breachTitle } suffered a breach. Once the breach was discovered and verified it was added to our database on { $addedDate }.
+
+# Title appearing on the Preferences dashboard. 
+monitor-preferences = {-brand-monitor} Preferences
```
flodolo: We have `-product-name`, which is "Firefox Monitor". Can we use that?

flodolo: Also, is it OK to give up on the bold for Preferences?

lesleyjanenorton: Yes, it would have needed to go either way.

flodolo: To be clear: either we add another term for *Monitor* (and it would be `-product-short-name`), or we update the string. `-brand-monitor` doesn't exist.

locales/en/app.ftl
```diff
@@ -823,3 +823,12 @@ additional-information-including = Additional information, including:
 
 # Title
 email-addresses-title = Email Addresses
+
+
+# This is a standardized breach overview blurb that appears on all breach detail pages.
+# $breachTitle is the name of the breached company or website.
+# $breachDate and $addedDate are calendar dates.
+breach-overview = On { $breachDate }, { $breachTitle } suffered a breach. Once the breach was discovered and verified it was added to our database on { $addedDate }.
```
pdehaan: This loses the `<b>` tags around the title and dates, but I know adding HTML markup in the strings isn't always the best. Unless you're planning on injecting HTML markup around the title and localized dates, etc.

lesleyjanenorton: It needed to go anyway.

locales/en/app.ftl
lesleyjanenorton: The headline is "Monitor Preferences", so I think it will require a new string for { -brand-Monitor }

locales/en/app.ftl
```diff
@@ -11,6 +11,7 @@
 -brand-HIBP = Have I Been Pwned
 -brand-fxa = Firefox Account
 -brand-Chrome = Chrome
+-product-short-name = Monitor
```
flodolo: nit: can you move it close to the other -product-name terms?

locales/en/app.ftl
```diff
@@ -823,3 +824,23 @@ additional-information-including = Additional information, including:
...
+# Title appearing on the Preferences dashboard. 
+monitor-preferences = {-product-short-name} Preferences
+
+# When a user is signed in, this appears in the drop down menu 
+# and is followed by the user's primary Firefox Account email. 
+signed-in-as = Signed in as: { $userEmail }
+
+# Appears on the All Breaches page and is followed by a list of filter options
+# that a user can filter the visible breaches by.
+filter-by = Filter By:
```
flodolo: What about the CATEGORY part? Are you dropping it?

flodolo:
@lesleyjanenorton 
This requires changes on the code side. Do you have a timeline for that?

PR #944, opened 2019-05-15
fix #927: disallow duplicate email addresses

### LINKED ISSUE #927: Dashboard allows me to set primary and secondary emails to same address
## Steps to reproduce
1. Sign in via FxA.
2. Go to https://fx-breach-alerts.herokuapp.com/user/preferences
3. Enter a secondary email thats the same as your primary account.

### Actual results:
You can.

Firefox_Monitor

But then it gets weird, because on https://fx-breach-alerts.herokuapp.com/user/dashboard all your breach summary stats are doubled because it counts the same email address twice as two separate things, and displays the results twice (with the same info).

Firefox_Monitor

groovecoder:


controllers/user.js
```diff
@@ -69,15 +69,29 @@ async function updateCommunicationOptions(req, res) {
   return res.json("Comm options updated");
 }
 
-async function add(req, res) {
-  // need to decide how to handle resends
 
-  _requireSessionUser(req);
+function _checkForDuplicateEmail(sessionUser, email) {
+  if (email === sessionUser.primary_email) {
+    throw new FluentError("user-add-duplicate-email");
```
pdehaan: Can you confirm this ID? I did a quick <kbd>git grep "user-add-duplicate-email"</kbd> and didn't find any results in **master** branch in any of the .ftl files.

pdehaan: I also ran <kbd>git grep "new FluentError" | sort | uniq</kbd> and didn't find any .ftl results for the "must-be-signed-in" key:

```sh
git grep "must-be-signed-in" locales/en/
```

Unrelated, but I also spotted this on L92 below, which is inconsistent with other `FluentError()` usages:

```js
throw new FluentError("Error message for this verification email timed out or something went wrong.");
```

groovecoder:
I was hoping to get a list of IDs for strings uncovered during development, and then add them all at once. Think we should do them ad-hoc? @betsymi @flodolo - which way do you prefer?

A single follow-up PR with all missing messages together, or add them as we're able to?

groovecoder:
Channeling @tcinotto ...

Per the timeline in [our ongoing meeting notes](https://docs.google.com/document/d/1gYRx-jQI8of-4cyPUZcItSK_FCYHjD87arAG6dT5QG4/edit), it looks like our final "dev freeze" (aka string stop) will be May 24, to be "Release Ready" on May 28. (And I think we're planning to launch right on June 4?)

tcinotto:
Hi Luke,

Dev freeze: so all blocking bug fixes should be complete by May 24.
We are targeting to release firefox monitor v2 on production on May 28th at
10am PT-12pm PT
Then June 4th would just be marketing launch

PR #971, opened 2019-05-22
Title case sign in button for en locales.

### LINKED ISSUE #970: Title case Sign In button in nav
Capitalize the In in Sign In. Here it is in the design comp:

image

lesleyjanenorton:
Fixes #970

### DIFF
--- a/public/css/app.css
+++ b/public/css/app.css
@@ -294,6 +294,10 @@ a.btn-light:active {
   align-items: center;
 }
 
+.capitalize {
+  text-transform: capitalize;
+}
+
 ul {
   padding: 0;
   display: flex;


public/css/app.css
```diff
@@ -294,6 +294,10 @@ a.btn-light:active {
   align-items: center;
 }
 
+.capitalize {
+  text-transform: capitalize;
+}
```
pdehaan: Since it looks like this may only impact 1-3 locales ("en", and possibly "en-CA"/"en-GB"), would it just be easier to manually edit the .ftl files?

https://github.com/mozilla/blurts-server/blob/c25f2ab1b3f286e1fdd10abfef84610cf5cdc1af/public/js/monitor.js#L198-L201

lesleyjanenorton: That make a lot more sense. @flodolo can I manually update this string in en-US, en-CA, and en-GB?

flodolo: Yes, that would OK in this case.

Issue #961, opened 2019-05-20
Audit new preferences operations for CSRF protection


groovecoder:
There are new create/update/delete operations on the /user/preferences page. We need to make sure we have implemented our CSRF protections on each of them.

PR #982, opened 2019-05-22
fix #941: implement remove button

### LINKED ISSUE #941: Clicking "Remove Firefox Monitor" doesn't do anything
In the console I see

```
NotFoundError: Node was not found context-buttons.js:60
```

groovecoder:
Okay, this is quite a bit of change and I'd like to get some more eyes on it.

@lesleyjanenorton - can you spot check on this branch a bunch? Sign up/in, add more email addresses, click their unsubscribe links, remove them from preferences, do the full "Remove Firefox Monitor" flow? I tried to do it a bunch but need some more eyes on it.

@pdehaan - also, can you give me a code sanity check here? I hate adding DB-touching code and would like some extra eyes if I'm making mistakes.


controllers/user.js
```diff
@@ -226,45 +225,73 @@ async function verify(req, res) {
...
+    title: req.fluentFormat("remove-fxm"),
+    subscriber: sessionUser,
+    whichPartial: "subpages/remove_fxm",
+  });
+}
+
+
+async function postRemoveFxm(req, res) {
+  const sessionUser = _requireSessionUser(req);
+  await DB.removeSubscriber(sessionUser);
```
groovecoder: This definitely needs #961 before launch.

controllers/user.js
```diff
@@ -226,45 +225,73 @@ async function verify(req, res) {
...
+  // Token is for a primary email address,
+  // redirect to preferences to remove Firefox Monitor
+  if (subscriber) {
+    return res.redirect("/user/preferences");
+  }
 
-  //throws error if user backs into and refreshes unsubscribe page
-  if (!subscriber) {
+  const emailAddress = await DB.getEmailByToken(req.query.token);
+  if (!subscriber && !emailAddress) {
```
pdehaan: Not sure about this logic.
If `subscriber` is truthy on L237-239 above, we `return` and exit this function, which means that the `!subscriber` check is probably always true here. Maybe kinda?

public/js/dashboard.js
```diff
@@ -51,3 +52,11 @@ if (document.querySelector(".email-card")) {
     option.addEventListener("click", sendCommunicationOption);
   });
 }
+
+if (document.querySelector(".remove-fxm")) {
+  const removeMonitorButton = document.querySelector(".remove-fxm");
```
pdehaan: minor nit, feel free to ignore... but if we move the `const` above the `if`, we can just do:

```js
if (removeMonitorButton) {...}
```

and avoid a second DOM traversal.

tests/controllers/user.test.js
```diff
@@ -258,10 +258,16 @@ test("user unsubscribe GET request with valid token and hash returns 200 without
...
 
-  expect(resp.statusCode).toEqual(200);
+  expect(resp.statusCode).toEqual(302);
+  expect(resp._getRedirectUrl()).toEqual("/user/preferences");
 });
 
 
+// TODO: test("remove-email POST");
+// TODO: test("remove-fxm get");
+// TODO: test("remove-fxm POST");
```
pdehaan: Are you doing these tests in _this_ PR, or is this work for Future Luke(tm)?

groovecoder: Planning to add in _this_ PR.

groovecoder: Now with tests.

controllers/user.js
groovecoder: Good catch.

public/js/dashboard.js
groovecoder: also nice.

lesleyjanenorton:
@groovecoder this looks great! Just went through the various sign up/sign in/delete-this/add-that/unsubscribe-this-secondary-email-from-an-email/etc flows and they all appear to be flowing as expected!

groovecoder:
@pdehaan and/or @lesleyjanenorton - want to double-check my test logic? Make sure I'm testing the right actions with the right DB/fixture users and the right expectations.

Issue #951, opened 2019-05-17
[Intermittent] An error page is displayed when you log out


obotisan:
**Affected versions:**
- Firefox 67.0

**Affected Platforms:**
- Windows 10 x64
- macOS 10.13
- Ubuntu 18.04 x64

**Steps to reproduce:**
1. Navigate to https://fx-breach-alerts.herokuapp.com/
2. Sign in with a valid account. 
3. Wait 10 minutes and log out.

**Expected result:**
- The user is still signed in. 

**Actual result:**
- An error is displayed on the page. 

**Additional Notes:**
- Sometimes you don't have to manually log out, the error is displayed either way. 
- This is the error page:
![pic5](https://user-images.githubusercontent.com/42799381/57932588-eb1ccd80-78c3-11e9-8fcb-836f80a7cc49.png)

PR #989, opened 2019-05-23
fix pre-FxA unsubscribe links

groovecoder:
I couldn't catch this until https://github.com/mozilla/blurts-server/pull/982 landed on https://fx-breach-alerts.herokuapp.com/ where we have pre-FxA subscribers.

I dug up one of those old pre-FxA subscriber emails and clicked its "unsubscribe" link. It incorrectly tried to redirect to `/users/preferences`.

So I updated a localhost user to match that old pre-FxA record and updated the `getUnsubscribe` code to fix that.

### DIFF
--- a/controllers/user.js
+++ b/controllers/user.js
@@ -13,6 +13,7 @@ const sha1 = require("../sha1-utils");
 
 function _requireSessionUser(req,res) {
   if (!req.session || !req.session.user) {
+    // TODO: can we do a nice redirect to sign in instead of an error?
     throw new FluentError("must-be-signed-in");
   }
   return req.session.user;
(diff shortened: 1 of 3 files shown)


controllers/user.js
```diff
@@ -13,6 +13,7 @@ const sha1 = require("../sha1-utils");
 
 function _requireSessionUser(req,res) {
   if (!req.session || !req.session.user) {
+    // TODO: can we do a nice redirect to sign in instead of an error?
```
groovecoder: I think this is also https://github.com/mozilla/blurts-server/issues/951

groovecoder:
Merging to get latest onto fx-breach-alerts. Wouldn't mind a post-merge code review though.

PR #999, opened 2019-05-23
fix #992: destroy access and refresh tokens on Remove

### LINKED ISSUE #992: Remove Monitor does not seem to remove the service from the Account client list
I signed into monitor (https://fx-breach-alerts.herokuapp.com/) and confirmed the "blurts-heroku" is added to the list: https://stable.dev.lcip.org/settings/clients

I then opted to "remove Monitor". When I return to this client list, I still see "blurts-heroku" in the list even after refreshing several times
Screenshot_2019-05-23 Firefox Accounts Devices apps(1)

groovecoder:
This could be a fast-follow after release if we think it's too risky to land before. This code:

1. Adds a db column to hold the FxA access token in addition to the refresh token
2. Updates the OAuth code to store the access token it receives from FxA
3. Updates the OAuth code to update existing subscribers' FxA data every time they sign in (to get existing subscribers' access tokens, and this should help update their avatar images too.)
4. During `removeFxm` calls FxA `/destroy` endpoint for both the access token and the refresh token

### DIFF
--- a/controllers/oauth.js
+++ b/controllers/oauth.js
@@ -80,7 +80,7 @@ async function confirmed(req, res, next, client = FxAOAuthClient) {
     // req.session.newUser determines whether or not we show "fxa_new_user_bar" in template
     req.session.newUser = true;
     const signupLanguage = req.headers["accept-language"];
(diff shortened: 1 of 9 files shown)


groovecoder:
Oh crap those tests are going to be so broken

groovecoder:
Tests fixed!

PR #1009, opened 2019-05-24
fix #985: update SES callback code for multiple emails

### LINKED ISSUE #985: Check that SES bounce callbacks correctly remove email addresses

groovecoder:


db/DB.js
```diff
@@ -264,18 +279,35 @@ const DB = {
...
+    const subscriber = await this.getSubscriberByEmail(email);
+    if (!subscriber) {
+      const emailAddress = await this.getEmailAddressByEmail(email);
+      if (!emailAddress) {
+        log.warn("removed-subscriber-not-found");
+        return;
+      }
+      await knex("email_addresses")
+        .where({
+          "email": email,
```
pdehaan: I think I'm confused here.
Above on L286 we're trying to find a `subscriber` by `email`. But here we're in a `if (!subscriber) {...}` block, so I'm assuming we didn't find a match, but here on L295, we're trying to find a match `WHERE email = email`.
Also curious is that we're trying to do a `emailAddress = this.getEmailAddressByEmail(email)`, but then not really "using" the `emailAddress` after that. Not sure if this line should be `.where({"email": emailAddress, ...)`.

db/DB.js
```diff
@@ -63,6 +63,21 @@ const DB = {
...
+  async getEmailAddressByEmail(email) {
+    const emailAddresses = await knex("email_addresses").where({
+      "email": email, verified: true,
+    });
+    if (!emailAddresses) {
+      return null;
+    }
+    if (emailAddresses.length > 1) {
+      // TODO: handle multiple emails in separate(?) subscriber accounts?
+      log.warn("getEmailAddressByEmail found the same email multiple times");
```
pdehaan: I'm a bit uncertain about the syntax here.
Below (L290), we do `log.warn("removed-subscriber-not-found")`, but here we're using a sentence.
I was trying to figure out the [mozlog](https://github.com/mozilla/mozlog) syntax, which seems to imply that the first param should be an `op`, a la:

```js
log.info(op, { some: details });
```

Not sure if this should be something like:

```js
log.warn("multiple-emails-found", {msg: "getEmailAddressByEmail found the same email multiple times"});
```

... which seems to be consistent with most of our current usages [except [controllers/hibp.js:51](https://github.com/mozilla/blurts-server/blob/8651f80e461007f661d36e1b3fc8d884eefa2813/controllers/hibp.js#L51)]:

```sh
$ git grep -n "\slog\."

basket.js:32:      log.error("subscribe", {stack: e.stack});
controllers/dockerflow.js:18:  log.info("generating");
controllers/dockerflow.js:23:    log.error("generating", {err: err});
controllers/hibp.js:51:    log.info(`${breachAlert.Name} is fabricated, a spam list, not associated with a website, or unverified. \n Breach Alert not sent.`);
controllers/hibp.js:61:  log.info("notification", { length: recipients.length, breachAlertName: breachAlert.Name });
controllers/hibp.js:68:    log.info("notify", {recipient});
controllers/hibp.js:96:  log.info("notified", { length: notifiedRecipients.length });
controllers/oauth.js:64:  log.debug("fxa-confirmed-fxaUser", fxaUser);
controllers/oauth.js:71:  log.debug("fxa-confirmed-profile-data", data.body);
controllers/scan.js:25:    log.error("_validatePageToken", {msg: "IP mis-match", pageTokenIP, requestIP});
controllers/scan.js:29:    log.error("_validatePageToken", {msg: "expired pageToken"});
controllers/ses.js:19:        log.error("notification", { err: err });
controllers/ses.js:37:  log.info("received-SNS", { id: notification.MessageId });
controllers/ses.js:54:      log.info("unhandled-event", { event: message.event });
controllers/ses.js:73:      log.info("unhandled-eventType", { type: message.eventType });
db/DB.js:273:      log.info("removed-subscriber", { id: aEntry.id });
db/DB.js:276:      log.warn("removed-subscriber-not-found");
email-utils.js:36:      log.info("smtpUrl-empty", { message: "EmailUtils will log a JSON response instead of sending emails." });
email-utils.js:75:          log.info("JSONTransport", { message: info.message.toString() });
hibp.js:33:      log.error("_throttledGot", {err: err});
hibp.js:38:        log.info("_throttledGot", {err: "got a 429, tryCount: " + tryCount});
hibp.js:40:          log.error("_throttledGot", {err: err});
hibp.js:82:    log.info("loadBreachesIntoApp");
hibp.js:99:    log.info("done-loading-breaches");
lib/fxa.js:28:      log.error("revokeOAuthToken", {stack: e.stack});
locale-utils.js:59:        log.error("loadFluentBundle", {stack: e.stack});
locale-utils.js:62:    log.info("LocaleUtils.init", {availableLanguages});
locale-utils.js:63:    log.info("LocaleUtils.init", {fluentBundles});
middleware.js:62:  log.error("error", {stack: err.stack});
routes/ses.js:17:    log.info("ses-notification-body", { body: req.body });
server.js:59:  log.error("try-load-languages-error", { error: error });
server.js:91:    log.error("try-load-breaches-error", { error: error });
server.js:213:    log.info("Listening", { port: listener.address().port });
server.js:216:  log.error("try-initialize-email-error", { error: error });
template-helpers/hbs-helpers.js:143:  log.error("ifCompare", {message: `${operator} not found`});
```

db/DB.js
groovecoder: The first check looks in the `subscribers` table. If we can't find the email address in there, the next check looks in the `email_addresses` table.

I'm using the `if (!emailAddress)` just to check if it wasn't found and log the warning in that case.

db/DB.js
groovecoder: Updated the `log.warn` syntax. Good catch.

groovecoder:
Okay, this is ready for a last r? and merge.

PR #1021, opened 2019-05-28
Update breach logos.

lesleyjanenorton:
This replaces some of our breach logos that are vertically oriented, white, or otherwise difficult to read.


pdehaan:
Actually, 2 minor things of interest was I think 8 images we modified from `100644 → 100755` permissions.

Oh, and I think it was the two R2 breaches had different logos (one was 100x100 and the other was 200x200).

groovecoder:
Merge-spree to get these small(-ish) bug-fixes in for post-launch release.

pdehaan:
Actually, just noticed that some of our "special case" logos (ie: List.png) are transparent+white. I'll file a new issue.

https://fx-breach-alerts.herokuapp.com/breach-details/DataAndLeads

pdehaan:
Filed a followup logo issue in #1037

PR #1028, opened 2019-05-28
Update looping in sign up banner.

lesleyjanenorton:


template-helpers/sign-up-banners.js
```diff
@@ -1,35 +1,28 @@
...
+    {
+      "title": "Advanced search in sensitive breaches",
+      "stringId": "feat-sensitive",
+    },
+    {
+      "string": "Security tips to protect your accounts",
+      "stringId": "feat-security-tips",
+    },
+  ];
+  bulletPoints.forEach(bulletPoint => {
```
groovecoder: This `forEach` "smells" like a potential place where we might be losing the assignment data?

template-helpers/sign-up-banners.js
```diff
@@ -1,35 +1,28 @@
...
+    bulletPoint["stringId"] = LocaleUtils.fluentFormat(locales, bulletPoint["stringId"]);
```
groovecoder: Or this assignment of a new value into the same item of the array?

template-helpers/sign-up-banners.js
```diff
@@ -1,35 +1,28 @@
...
+  });
```
groovecoder: Maybe we should store the translated bullet points into a whole new array ...

```
const translatedBulletPoints = ...
```

PR #1042, opened 2019-05-29
Highlight "Home" navigation when signed in.

### LINKED ISSUE #1027: "Home" navigation isn't highlighted when you are signed in
Firefox_Monitor
**Figure 1:** https://monitor.firefox.com/user/dashboard (signed in)

---

Firefox_Monitor
**Figure 2:** https://monitor.firefox.com/ (signed out)

---

https://github.com/mozilla/blurts-server/blob/8ce597e75b0d9fc96ff8279d3c2816fd52c7a61b/public/js/monitor.js#L167-L178

The logged in URL is https://monitor.firefox.com/user/dashboard, but the top nav link is:

```html
<a class="nav-link" ... href="/">Home</a>
```

Which means the `if` statement evaluates to something like this, which is why the ".active-link" class never gets appended to the class list:

```js
     if ("/" === "/user/dashboard") { 
       link.classList.add("active-link"); 
     } 
```

lesleyjanenorton:
- Adds highlighted "home" navigation for signed in users on the "/user/dashboard" page.
- Fixes bug where the "Security Tips" link does not appear highlighted if the user navigates to that page via one of the article links on another page.

Fixes #1027

### DIFF
--- a/public/js/monitor.js
+++ b/public/js/monitor.js
@@ -160,6 +160,22 @@ function toggleHeaderStates(header, win) {
   }
 }
 
+function styleActiveLink(locationHref) {
+  let queryString = `.nav-link[href='${locationHref}']`;
+  const activeLink = document.querySelector(queryString);
+  if (activeLink) {
+    return activeLink.classList.add("active-link");
+  }
+
+  if (~locationHref.indexOf("/dashboard")) {
+    queryString = queryString.replace("user/dashboard", "");
+    return document.querySelector(queryString).classList.add("active-link");
+  }
+  if (~locationHref.indexOf("/security-tips")) {
+    return document.querySelector(".nav-link[href*='/security-tips']").classList.add("active-link");
+  }
+}
+
 ( async() => {
   document.addEventListener("touchstart", function(){}, true);
   const win = window;
@@ -169,14 +185,7 @@ function toggleHeaderStates(header, win) {
     if (previousActiveLink) {
       previousActiveLink.classList.remove("active-link");
     }
-    const navLinks = document.querySelectorAll(".nav-link");
-
-    navLinks.forEach(link => {
-      if (link.href === win.location.href) {
-        link.classList.add("active-link");
-      }
-    });
-
+    styleActiveLink(win.location.href);
     if (win.location.search.includes("utm_") && win.history.replaceState) {
       win.history.replaceState({}, "", win.location.toString().replace(/[?&]utm_.*/g, ""));
     }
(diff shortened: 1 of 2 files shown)


public/js/monitor.js
```diff
@@ -160,6 +160,22 @@ function toggleHeaderStates(header, win) {
...
 }
 
+function styleActiveLink(locationHref) {
+  let queryString = `.nav-link[href='${locationHref}']`;
+  const activeLink = document.querySelector(queryString);
+  if (activeLink) {
+    return activeLink.classList.add("active-link");
+  }
+
+  if (~locationHref.indexOf("/dashboard")) {
```
groovecoder: Don't need to use bitwise OR (`~`) here. Let's just do `locationHref.indexOf("/dashboard") > -1` or `locationHref.indexOf("/dashboard") !== -1`

public/js/monitor.js
```diff
@@ -160,6 +160,22 @@ function toggleHeaderStates(header, win) {
...
+    queryString = queryString.replace("user/dashboard", "");
+    return document.querySelector(queryString).classList.add("active-link");
+  }
+  if (~locationHref.indexOf("/security-tips")) {
```
groovecoder: Same here.

public/js/monitor.js
pdehaan: I ain't gonna lie. I had to Bing exactly what the bitwise `~` operator was.
+1 to using `!locationHref.includes("/dashboard")` or what @groovecoder suggests above.
I also had to refresh my memory on the 2 logged in routes, which were:
- /user/preferences
- /user/dashboard

So my guess is that when the user is on /user/preferences, nothing will be highlighted, but that seems reasonable.

I can't really think of any other better solutions, unless it makes sense to check for the `/user/dashboard` route first, and overwrite it with the `/` route:

```js
function setActiveLink(locationHref) {
  if (locationHref === "/user/dashboard") {
    locationHref = "/";
  }
  const queryString = `.nav-link[href='${locationHref}']`;
  const activeLink = document.querySelector(queryString);
  if (activeLink) {
    return activeLink.classList.add("active-link");
  }
  ...
}
```

pdehaan: There is also an ESLint [`no-bitwise`](https://eslint.org/docs/rules/no-bitwise) rule which disallows bitwise operators:

> **disallow bitwise operators** (`no-bitwise`)
> The use of bitwise operators in JavaScript is very rare and often `&` or `|` is simply a mistyped `&&` or `||`, which will lead to unexpected behavior.

Doesn't throw any existing errors in our code, not sure if we want to bother adding it to our ESLint config.

PR #1047, opened 2019-05-29
Update dashboard and preferences pages.

lesleyjanenorton:
- Adds form for adding additional emails to the dashboard page and shows an "Email Sent!" message.
- Adds new string to replace the "Global Communication" string.


locales/en/app.ftl
```diff
@@ -851,3 +863,26 @@ join-firefox = There is a way to protect your privacy. Join {-brand-name}.
...
+
+
+email-sent = Email Sent!
+
+# Form title
+want-to-add = Want to add another email?
+
+# This is part of a confirmation message that appears after a user has submitted
+# the form to add an additional email to Firefox Monitor.
+verify-the-link = Verify the link sent to { $userEmail } to add it to Firefox Monitor.
```
flodolo: Firefox Monitor -> `{ -product-name }`

locales/en/app.ftl
```diff
@@ -851,3 +863,26 @@ join-firefox = There is a way to protect your privacy. Join {-brand-name}.
...
+# to the Preferences page. The code and text for the link is generated elsewhere
+# using the {-preferences} string.
+manage-all-emails = Manage all email addresses in { $preferencesLink }.
+
+# This string is a header on the user preferences page and
+# appears above a check-box list of user options which allow
+# the user to choose whether or not they want to receive breach
+# alerts for all of their monitored email addresses to a single 
+# email address.
+breach-alert-notifications = Breach Alert Notifications
```
flodolo: Is this replacing "Global communications"? If so, why isn't the old string removed?

lesleyjanenorton: It does replace "Global Communications" but I was under the impression we should not remove strings at this point.

lesleyjanenorton:
@flodolo I think (@betsymi , correct me if I'm wrong), the timing is June 4th. The thinking is that the form to add another email is too buried and possibly difficult to find for brand new users.

locales/en/app.ftl
flodolo: When are we going to remove the strings from the previous version, given we launched v2? That would be a good start ;-)

flodolo:
> the timing is June 4th. The thinking is that the form to add another email is too buried and possibly difficult to find for brand new users.

That seems like a really bad idea to me. We're going to have all the focus from marketing, and websites partially translated, even in focus locales like German and French.

This seems like a fix that can be deployed after the initial launch (that doesn't mean that we can't expose strings earlier).

On a side note: it's public holiday here today, and I was hoping to have a long weekend away from the computer.

locales/en/app.ftl
flodolo: The old string is unused, the moment you push this code out, we're either going to see English or the new translation. The only reason to keep it is that we only expose the strings, but not the code, so we can still push updates to production.

template-helpers/dashboard.js
```diff
@@ -94,9 +94,25 @@ function getUserPreferences(args) {
   return args.fn(user);
 }
 
+function getLastAddedEmailStrings(args) {
+  const locales = args.data.root.req.supportedLocales;
+  let lastAddedEmail = args.data.root.lastAddedEmail;
+  lastAddedEmail = `<span class="bold">${lastAddedEmail}</span>`;
```
groovecoder:
```suggestion
  const lastAddedEmailSpan = `<span class="bold">${lastAddedEmail}</span>`;
```

template-helpers/dashboard.js
```diff
@@ -94,9 +94,25 @@ function getUserPreferences(args) {
...
+
+  let preferencesLinkString = LocaleUtils.fluentFormat(locales, "preferences");
+  preferencesLinkString = `<a class="demi text-link" href="/user/preferences">${preferencesLinkString}</a>`;
+
+  const lastAddedEmailStrings = [
+    LocaleUtils.fluentFormat(locales, "verify-the-link", { userEmail: lastAddedEmail }),
```
groovecoder:
```suggestion
    LocaleUtils.fluentFormat(locales, "verify-the-link", { userEmail: lastAddedEmailSpan }),
```

template-helpers/dashboard.js
```diff
@@ -94,9 +94,25 @@ function getUserPreferences(args) {
   return args.fn(user);
 }
 
+function getLastAddedEmailStrings(args) {
+  const locales = args.data.root.req.supportedLocales;
+  let lastAddedEmail = args.data.root.lastAddedEmail;
```
groovecoder:
```suggestion
  const lastAddedEmail = args.data.root.lastAddedEmail;
```

template-helpers/dashboard.js
```diff
@@ -94,9 +94,25 @@ function getUserPreferences(args) {
...
+  lastAddedEmail = `<span class="bold">${lastAddedEmail}</span>`;
+
+  let preferencesLinkString = LocaleUtils.fluentFormat(locales, "preferences");
```
groovecoder:
```suggestion
  const preferencesLinkString = LocaleUtils.fluentFormat(locales, "preferences");
```

template-helpers/dashboard.js
```diff
@@ -94,9 +94,25 @@ function getUserPreferences(args) {
...
+  preferencesLinkString = `<a class="demi text-link" href="/user/preferences">${preferencesLinkString}</a>`;
```
groovecoder:
```suggestion
  const preferencesLink = `<a class="demi text-link" href="/user/preferences">${preferencesLinkString}</a>`;
```

template-helpers/dashboard.js
```diff
@@ -94,9 +94,25 @@ function getUserPreferences(args) {
...
+
+  const lastAddedEmailStrings = [
+    LocaleUtils.fluentFormat(locales, "verify-the-link", { userEmail: lastAddedEmail }),
+    LocaleUtils.fluentFormat(locales, "manage-all-emails", { preferencesLink: preferencesLinkString }),
```
groovecoder:
```suggestion
    LocaleUtils.fluentFormat(locales, "manage-all-emails", { preferencesLink }),
```

controllers/user.js
```diff
@@ -167,11 +174,18 @@ async function getDashboard(req, res) {
   const user = _requireSessionUser(req, res);
   const allBreaches = req.app.locals.breaches;
   const { verifiedEmails, unverifiedEmails } = await getAllEmailsAndBreaches(user, allBreaches);
+  let lastAddedEmail = null;
```
groovecoder: To avoid `let`, change to:

```js
const lastAddedEmail = req.session.lastAddedEmail;
if (lastAddedEmail) {
  delete req.session.lastAddedEmail; // or req.session["lastAddedEmail"] = undefined;
}
```

and you can drop the whole other `if (req.session.lastAddedEmail)` block.

lesleyjanenorton: For reasons most mystérieux, `delete req.session.lastAddedEmail` does not work for me locally.

groovecoder:
Once the conflicts are resolved we can merge this, tag, and tee up a production deploy.

flodolo:
> Once the conflicts are resolved we can merge this, tag, and tee up a production deploy.

I disagree, we should leave a buffer for localization to avoid having English content for all locales.

groovecoder:
We've merged the strings from this branch into `master`. We will check the translation status tomorrow and/or Monday for merging & deploying this change.

groovecoder:
Ping @lesleyjanenorton - are you able to resolve conflicts for this for a merge & deploy today or tomorrow?

PR #1050, opened 2019-05-30
Add fixed "Join Firefox" branding strip to top of all pages.

lesleyjanenorton:
- Adds `branding-strip.hbs`.
- Makes adjustments to page content spacing to account for the added height of the top bar.
- Adds script to toggle the header in and out of view on mobile (< 600px) devices.
- Removes conflicting back to top button from `/breaches`.


public/js/monitor.js
```diff
@@ -136,20 +136,48 @@ function toggleArticles() {
   });
 }
 
-function toggleMobileFeatures() {
-  const windowWidth = window.innerWidth;
+function hideShowNavBars(win, navBar) {
+  win.onscroll = function(e) {
```
groovecoder:
```suggestion
  win.addEventListener('scroll', (e) => {
```

groovecoder: Though I'm not sure if this will work with `this` usage below?

public/js/monitor.js
```diff
@@ -136,20 +136,48 @@ function toggleArticles() {
...
       return;
     }
 
   const closeActiveEmailCards = document.querySelectorAll(".col-9.email-card.active");
     closeActiveEmailCards.forEach(card => {
       card.classList.remove("active");
     });
+
+    if (windowWidth < 600) {
+      win.addEventListener("scroll", hideShowNavBars(win, topNavBar));
```
groovecoder: This adds `hideShowNavBars` as an event listener on window scroll, but `hideShowNavBars` also uses `onscroll` to add an event listener to window scrolling?

views/partials/branding-strip.hbs
```diff
@@ -0,0 +1,8 @@
+<div class="top-bar flx jst-cntr row-full-width">
+  <div class="join-fx-wrap flx">
+    <span class="join-fx-img"></span>
+    <div class="join-fx-copy-wrap">
+      <span class="join-fx-copy">{{ getString "join-firefox" }}<a class="learn-more" taget="_blank" rel="noopener" href="https://www.mozilla.org/firefox/accounts/?utm_source=monitor.firefox.com&utm_medium=banner&utm_campaign=trailhead&utm_content=protect-your-privacy">{{ getString "learn-more-link" }}</a></span>
```
groovecoder:
```suggestion
      <span class="join-fx-copy">{{ getString "join-firefox" }}<a class="learn-more" target="_blank" rel="noopener" href="https://www.mozilla.org/firefox/accounts/?utm_source=monitor.firefox.com&utm_medium=banner&utm_campaign=trailhead&utm_content=protect-your-privacy">{{ getString "learn-more-link" }}</a></span>
```

views/partials/header/header.hbs
```diff
@@ -1,23 +1,26 @@
...
-        </nav>
-      </section>
-      <!--mobile navigation-->
-      <section class="mobile-nav show-mobile">
-        <span class="nav-link drop-down-menu">{{ getString "menu" }} {{> svg/toggle-down }}</span>
-        <nav class="mobile-menu flx-cntr row-full-width">
-          {{> header/nav-links }}
-        </nav>
-      </section>
+      {{> branding-strip }}
```
groovecoder:
```suggestion
      {{#unless req.session.user}}
        {{> branding-strip }}
      {{/unless}}
```

lesleyjanenorton: So I asked about this specifically and we are supposed to show this to users who are signed in.

PR #1070, opened 2019-06-05
Improve `/breaches` loading.

lesleyjanenorton:
#### Broad strokes:
- Loads `all-breaches.min.js` separately (and without the `defer` attribute).
- Reduces the size of the breach array sent to the template by removing more unused `breachCard` properties and sending the localized `breachCard` strings once as`LocalizedBreachCardStrings`.
- Builds only the first 15 breaches on page load with a "Show All" button to see the rest.

#### `breachCard` updates:
- Renames the artist formerly known as `breachCard.String` to `breachCard.LocalizedBreachCardStrings` to match the `/breaches` breach cards. 
- Stops assigning `breachCard.Category` since we aren't doing anything with the category in the cards now anyway.


public/js/all-breaches/all-breaches.js
```diff
@@ -85,112 +86,113 @@ function makeBreaches(breachArray, breachCardWrapper) {
...
 
-    makeBreaches(breachArray, breachCardWrapper);
-    const showAllBreaches = document.getElementById("show-all-breaches");
     const noResultsBlurb = document.getElementById("no-results-blurb");
 
     const fuzzyFindInput = document.getElementById("fuzzy-find-input");
     const fuzzyFinder = document.getElementById("fuzzy-form");
 
-    showAllBreaches.addEventListener("click", (e) => {
+    const [fuzzyShowAll, showHiddenBreaches] = document.querySelectorAll(".show-all-breaches");
```
groovecoder: Are we guaranteed that these will always come back in this order? Or should we do it more like this?

```js
for (const showAllBreachesElement of document.querySelectorAll(".show-all-breaches")) {
  if (showAllBreachesElement.classList.contains("fuzzy-find-input")) {
    // do the code for fuzzyShowAll and return
  }
  // do the code for showHiddenBreaches (and return)
}
```

template-helpers/breaches.js
```diff
@@ -32,70 +30,28 @@ function dataClassesforCards(breach, locales) {
...
+  return breaches;
+}
 
-    breachCard.AddedDate = prettyDate(breach.AddedDate, locales);
-    breachCard.PwnCount = localeString(breach.PwnCount,locales);
-    breachCard.DataClasses = dataClassesforCards(breach, locales);
+function makeBreachCards(breaches, locales) {
+  const formattedBreaches = [];
+  const breachCardStrings = getLocalizedBreachCardStrings(locales);
+  breaches = JSON.parse(JSON.stringify(breaches));
```
groovecoder: Do you have to `JSON.parse(JSON.stringify())`? Can you just use `breaches` without `stringify` and `parse`?

template-helpers/breaches.js
```diff
@@ -106,21 +62,28 @@ function lastAddedBreach(options) {
...
+  breach.AddedDate = prettyDate(breach.AddedDate, locales);
+  breach.PwnCount = localeString(breach.PwnCount,locales);
+  breach.DataClasses = dataClassesforCards(breach, locales);
+  return breach;
+}
+
+function getBreachArray(breaches, args) {
   const locales = args.data.root.req.supportedLocales;
-  let allBreaches = args.data.root.breaches;
+  breaches = JSON.parse(JSON.stringify(breaches));
```
groovecoder: Again do we need `JSON.parse(JSON.stringify())` together?

template-helpers/breaches.js
lesleyjanenorton: There may be a way, but after lots of trying different approaches and stack overflowing I wasn't able to find it.

template-helpers/breaches.js
lesleyjanenorton: Same as previous comment.

public/js/all-breaches/all-breaches.js
```diff
@@ -85,112 +86,113 @@ function makeBreaches(breachArray, breachCardWrapper) {
     makeBreachInfoSpans("breach-title", breach.Title, wrapper);
 
     // added date
-    makeBreachInfoSpans("breach-key", breach.String.AddedDate, wrapper);
+    makeBreachInfoSpans("breach-key", LocalizedBreachCardStrings.BreachReported, wrapper);
```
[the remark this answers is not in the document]
lesleyjanenorton: Thank you for calling this out. Our many specs attach a multitude of labels to the `breach.AddedDate` in breach cards - ("Added:"... "Reported on:"... "Breach discovered:"...), but after speaking with Betsy I can confirm we are going with "Breach added: `breach.AddedDate` " - which unfortunately necessitates creating a new string

public/js/all-breaches/all-breaches.js
lesleyjanenorton: We could do this but then we'd still need to assign them to variables because they are also used down the line.

pdehaan:
A couple interesting notes from doing some quick scanning of the _current_ https://fx-breach-alerts.herokuapp.com/breaches page:

- All the contents from the `div#breach-array-json` element is **214K** worth of data.
- All the rendered content of the final HTML page is **469K**.

public/js/all-breaches/all-breaches.js
groovecoder: Hmm .... maybe a helper function then? But not sure how much of the vars we'd have to pass in to make it work? Might not be worth it.

```js
 function addShowAllListeners() {
   const showAllElements = {};
   for (const showAllBreachesElement of document.querySelectorAll(".show-allbreaches")) {
     if (showAllBreachesElement.classList.contains("fuzzy-find-input")) {
       showAllBreachesElement.addEventListener("click", (e) => {
         e.preventDefault();
      fuzzyFindInput.value = "";
      doBreaches(breaches);
      noResultsBlurb.classList = [""];
      fuzzyShowAll.classList = ["fuzzy-find-show-breaches"];
     });
     showAllElements.fuzzyShowAll = showAllBreachesElement;
   } else {
     showAllBreachesElement.addEventListener("click", (e) => {
    doBreaches(breaches);
    showHiddenBreaches.classList.add("hide");
     });
     showAllElements.showHiddenBreaches = showAllBreachesElement;
   }
 }
 return showAllElements;
 }

const { fuzzyShowAll, showHiddenBreaches } = addShowAllListeners();
```

PR #1074, opened 2019-06-05
Show $userEmail in footer.

### LINKED ISSUE #1068: $userEmail is not expanded in email report footer
The footer of a report email I received has:

> We sent this message to $userEmail because the email address opted into alerts from Firefox Monitor.

lesleyjanenorton:
Fixes #1068 (despite the branch name claiming otherwise).

### DIFF
--- a/controllers/user.js
+++ b/controllers/user.js
@@ -54,7 +54,7 @@ async function resendEmail(req, res) {
     email,
     req.fluentFormat("user-add-email-verify-subject"),
     "default_email",
-    { email,
+    { recipientEmail: email,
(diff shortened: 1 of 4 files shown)


controllers/user.js
```diff
@@ -110,13 +109,13 @@ async function add(req, res) {
     req.fluentFormat("user-add-email-verify-subject"),
     "default_email",
     { breachedEmail: email,
+      recipientEmail: email,
```
groovecoder: This also needs to go into the `confirmed` handler in `controllers/oauth.js`

PR #1078, opened 2019-06-06
Add new dashboard and preferences strings.

lesleyjanenorton:


locales/en/app.ftl
```diff
@@ -848,6 +848,28 @@ join-firefox = There is a way to protect your privacy. Join {-brand-name}.
...
+want-to-add = Want to add another email?
+
+# This is part of a confirmation message that appears after a user has submitted
+# the form to add an additional email to Firefox Monitor.
+verify-the-link = Verify the link sent to { $userEmail } to add it to {-product-name}.
+
+# This is part of a confirmation message that appears after a user has submited the
+# form to add an additional email to Firefox Monitor. { $preferencesLink } is a link
+# to the Preferences page. The code and text for the link is generated elsewhere
+# using the {-preferences} string.
```
flodolo: Is this the `monitor-preferences` string?

lesleyjanenorton: It's the `preferences` string that we use as a link title in the menu that drops down off the user avatar.

flodolo: My point is that there's not a `-preferences` string

PR #1081, opened 2019-06-07
Adjust space above "latestBreach" card.

lesleyjanenorton:
Adds more space between the `latestBreach` card and the content above it on smaller screens.

### DIFF
--- a/public/css/monitor.css
+++ b/public/css/monitor.css
@@ -128,12 +128,6 @@ h3.sb-callout {
   max-width: 350px;
 }
 
-@media screen and (max-width: 1100px) {
-  /* .landing-bottom-bar {
-    padding-top: 4vh;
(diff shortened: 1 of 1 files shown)


public/css/monitor.css
```diff
@@ -180,7 +173,7 @@ h3.sb-callout {
     min-height: 96vh;
   }
 
-  /* .landing-bottom-bar {
-    padding-top: 6vh;
-  } */
+  .landing-content .form-desc {
+    padding-bottom: 3rem;
```
groovecoder: Just a note: at some point we should go thru and change our properties to [logical properties](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Logical_Properties/Basic_concepts). But that's a bigger project.

Issue #1008, opened 2019-05-24
"fn is not a function" error on dev /hibp/breaches endpoint


pdehaan:
https://fx-breach-alerts.herokuapp.com/hibp/breaches

I use our endpoint to scrape the breach list, but I guess I've always used the monitor.firefox.com prod link, and maybe nothing internally uses the /hibp/breaches endpoint which is why we havent noticed this throwing an error.

![https___fx-breach-alerts_herokuapp_com_hibp_breaches](https://user-images.githubusercontent.com/557895/58345605-9b7f5a00-7e0d-11e9-8c3e-db5f4e65aa03.png)

groovecoder:
Oh, I removed this endpoint, as we decided not to use it for the Firefox door-hanger UI.

pdehaan:
So, is there any way to get a list of all the breaches, or should I just point all my scripts at HIBP's endpoint now?

lesleyjanenorton:
Is there any harm in us bringing this back? Seems like it might be useful.

PR #1096, opened 2019-06-26
Add D3Scene and SocialEngineered logos.

lesleyjanenorton:


pdehaan:
| Name | Logo |
|:-----|:-----|
| D3Scene | Firefox_Monitor |
| SocialEngineered | Firefox_Monitor |

Interestingly, the Logo is there:
https://fx-breach-alerts.herokuapp.com/img/logos/D3Scene.png

But it's not rendering like I expected. Looking at the source on the dev server, it looks like it may be a D3scene vs D3Scene issue, and the server OS is case sensitive:

```html
<img alt="" class="breach-logo lazy-loaded" data-src="/img/logos/D3scene.png" src="/img/logos/missing-logo-icon.png">
```

https://haveibeenpwned.com/api/v2/breach/D3scene

But sadly we removed our [/hibp/breaches](https://fx-breach-alerts.herokuapp.com/hibp/breaches) endpoint, so I can't easily inspect what our modified HIBP data structure is.

lesleyjanenorton:
Thank you for catching this @pdehaan . Should be fixed in #1098. Was the removal of that endpoint intentional? The error message I get when I land there looks like a handlebars helper error.

pdehaan:
> Was the removal of that endpoint intentional? The error message I get when I land there looks like a handlebars helper error.

Intentional per https://github.com/mozilla/blurts-server/commit/a24f3d2205a0ae909d0b63d5b9acfe2e00294dd0 commit and https://github.com/mozilla/blurts-server/issues/1008#issuecomment-495758202.

But unfortunately that also broke virtually every script/tool I've written for blurts-server, since it always expected a JSON blob of breaches at that URL. Although I managed to refactor most of my scripts to use the upstream HIBP API v2 endpoints, and then redid all the logic around our breach logo naming or DataClasses modifying, etc.

pdehaan:
Confirmed.

Firefox_Monitor

PR #1098, opened 2019-06-27
Fix casing on D3scene image.

lesleyjanenorton:


pdehaan:
LGTM, but I'd delete the old file so we don't have **D3Scene.png** _and_ **D3scene.png** and never remember which is the live one.

lesleyjanenorton:
@pdehaan good call but I think I may have to do that in a follow up PR? I renamed the image and so, in this branch anyway, there is no `D3Scene.png` to delete.

lesleyjanenorton:
closing in favor of #1099

PR #1103, opened 2019-06-28
Reinstate /hibp/breaches endpoint.

### LINKED ISSUE #1008: "fn is not a function" error on dev /hibp/breaches endpoint

lesleyjanenorton:
Fixes #1008

### DIFF
--- a/controllers/hibp.js
+++ b/controllers/hibp.js
@@ -116,7 +116,15 @@ async function notify (req, res) {
   );
 }
 
+async function breaches (req, res, next) {
+  res.append("Last-Modified", req.app.locals.mostRecentBreachDateTime);
+  res.json(req.app.locals.breaches);
+}
+
+
+
 
 module.exports = {
   notify,
+  breaches,
 };


groovecoder:
Interesting. This is/was just a proxy for https://haveibeenpwned.com/api/v2/breaches right? Does your script want to make sure *our* environment has all the breaches? Or can your script use the HIBP endpoint directly?

groovecoder:
Even if we decide to remove our own endpoint, we should removing [the corresponding code from `routes/hibp.js`](https://github.com/mozilla/blurts-server/blob/master/routes/hibp.js#L9) too.

pdehaan:
> This is/was just a proxy for [haveibeenpwned.com/api/v2/breaches](https://haveibeenpwned.com/api/v2/breaches) right?

IIRC, it returned a modified version of the upstream HIBP breach data (where we tweak the `.DataClasses[]` and `.LogoPath`).

groovecoder:
Oh yeah, we do some customization to the data. Sounds good to restore.

Issue #125, opened 2018-05-30
Rename repo'n'stuff to firefox-monitor-server?


pdehaan:
Possibly mildly disruptive, but may be nice to do before this goes out as a study or project.

pdehaan:
We'd need to make sure we sync up w/ @Micheletto before renaming anything, as he'll need to make tweaks in #135; "Complete Dockerflow setup for stage & prod deployment pipelines".

From bobm:
> It has some minimal impact to the Dockerflow puppet I've got started, but not much that a find and replace can't fix.  The most time consuming changes would be changing the Docker hub and CircleCI configuration.  If the dev team can handle the CircleCI stuff, I can worry about the hub configuration.

pdehaan:
We decided (offline) that this can/should wait until post shield study to be less disruptive as we try and wrap everything up in time for the Shield Study.

EMMLynch:
Closing since we've redesigned the site and functionality since this was created. If you feel that this is still needed, please let me know.

PR #1106, opened 2019-06-29
Footer updates: Add a link to the repo.

### LINKED ISSUE #924: Add GitHub icon in footer?
Currently, I don't think there is an easy way [in the v2 design] to find the GitHub repo if you want to contribute or submit a bug.

### V1 footer
Firefox_Monitor

### V2 footer
Firefox_Monitor

lesleyjanenorton:
Closes #924 

Hopefully this will make it easier for people to find the repo. 
Maybe we also want to consider #125?

### DIFF
--- a/locales/en/app.ftl
+++ b/locales/en/app.ftl
@@ -12,7 +12,6 @@
 -brand-HIBP = Have I Been Pwned
 -brand-fxa = Firefox Account
 -brand-Chrome = Chrome
--github = Github
 
 
 layout-Firefox = {-brand-name}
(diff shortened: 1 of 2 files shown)


locales/en/app.ftl
```diff
@@ -12,6 +12,7 @@
 -brand-HIBP = Have I Been Pwned
 -brand-fxa = Firefox Account
 -brand-Chrome = Chrome
+-github = Github
```
flodolo: That's not how you write GitHub ;-)

locales/en/app.ftl
```diff
@@ -23,6 +24,8 @@ about-firefox-alerts = About Firefox Alerts
 give-feedback = Give Feedback
 terms-and-privacy = Terms & Privacy
 
+github-link-title = {-github}
```
flodolo: If you're only using it once, you shouldn't create a term for it, just 


```
github-link-title = GitHub
```

locales/en/app.ftl
lesleyjanenorton: thank you!

groovecoder:
We shouldn't need to tag it with utm params ourselves. [Our GitHub repo traffic page](https://github.com/mozilla/blurts-server/graphs/traffic) already shows referring sites.

Or do you mean for us to see how many visitors actually click this?

PR #1111, opened 2019-07-10
 fix #1110: add /user/breach-stats endpoint 

### LINKED ISSUE #1110: Add /user/getBreachStats endpoint
It accepts an OAuth access token in the `Authorization: Bearer` header. (The access token should have a `monitor` scope.)

If the access token is for a user who has an active Monitor subscription, return a JSON summary of their breach statistics:
* number of Email addresses being monitored
* number of data breaches that have exposed their data
* how many times their password(s) have been exposed

See [invision spec](https://mozilla.invisionapp.com/d/main#/console/17296602/360923276/preview).

groovecoder:
For #1110, I want to get this up for review ASAP.

Still TODO:
* [x] Change endpoint to accept OAuth token in `Authorization` HTTP header instead of url parameter

### DIFF
--- a/controllers/user.js
+++ b/controllers/user.js
@@ -8,6 +8,7 @@ const EmailUtils = require("../email-utils");
 const { FluentError } = require("../locale-utils");
 const FXA = require("../lib/fxa");
 const HIBP = require("../hibp");
+const { resultsSummary } = require("../scan-results");
 const sha1 = require("../sha1-utils");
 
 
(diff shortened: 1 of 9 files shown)


controllers/user.js
```diff
@@ -337,6 +338,20 @@ async function getPreferences(req, res) {
 }
 
 
+async function getBreachStats(req, res) {
+  const fxaResponse = await FXA.verifyOAuthToken(req.query.fxa_oauth_token);
```
pdehaan: I'm guessing if an `fxa_auth_token` wasn't passed via the querystring, that'd just be `null` or `undefined`. Not sure if we'd need to do a check for a valid-looking oauth token before proceeding or if that is just overkill.

controllers/user.js
```diff
@@ -337,6 +338,20 @@ async function getPreferences(req, res) {
...
+  const user = await DB.getSubscriberByFxaUid(fxaResponse.body.user);
```
pdehaan: Should we do any checks here to verify that `user` found a match?

scan-results.js
```diff
@@ -95,4 +95,52 @@ const scanResult = async(req, selfScan=false) => {
...
+    },
+    passwords: {
+      subhead: "",
+      affectedAccounts: {},
+      count: 0,
+    },
+  };
+  let foundBreaches = [];
+
+  breachStats.monitoredEmails.count = verifiedEmails.length;
```
pdehaan: Why not set this on L102 above?

scan-results.js
```diff
@@ -95,4 +95,52 @@ const scanResult = async(req, selfScan=false) => {
...
+  breachStats.monitoredEmails["allVerifiedEmails"] = verifiedEmails;
```
pdehaan: Same here, why not set this in the initial object above?

controllers/user.js
groovecoder: Added checks and appropriate responses.

groovecoder:
Addressed all comments and updated the code to get the token from `Authorization:` header instead of URL parameter.

groovecoder:
Updated to fix tests too. @lesleyjanenorton r?

controllers/user.js
```diff
@@ -337,6 +338,35 @@ async function getPreferences(req, res) {
...
+    });
+  }
+  const fxaResponse = await FXA.verifyOAuthToken(req.token);
+  if (!fxaResponse) {
+    return res.status(404).json({
+      errorMessage: "Cannot find FXA for that OAuth token.",
+    });
+  }
+  const user = await DB.getSubscriberByFxaUid(fxaResponse.body.user);
+  if (!fxaResponse) {
```
pdehaan: I think this is unreachable sine we check-n-bail above on L348.
Should this be `if (!user)`?

groovecoder: Oh, yes. Duh! Good catch.

Issue #1064, opened 2019-06-04
Travis CI failures: Failed: "Access denied. Certificate could not be retrieved"


groovecoder:
We receive intermittent Travis CI failures during the "ses notification with Permanent bounce unsubscribes recipient" test:

```
FAIL tests/controllers/ses.test.js
  ● ses notification with Permanent bounce unsubscribes recipient
    Failed: "Access denied. Certificate could not be retrieved"
      24 | 
      25 | 
    > 26 | test("ses notification with Permanent bounce unsubscribes recipient", async () => {
         | ^
      27 |   // TODO: restore tests for ["General", "NoEmail", "Suppressed"] sub types
      28 |   const testEmail = "bounce@simulator.amazonses.com";
      29 |   const testHashes = [getSha1(testEmail)];
      at Env.it (node_modules/jest-jasmine2/build/jasmineAsyncInstall.js:91:24)
      at Object.<anonymous> (tests/controllers/ses.test.js:26:1)
```

Issue #1034, opened 2019-05-28
Copy refinements for V2 website


betsymi:
These copy changes didn't make it in yet, so documenting them here. 

Location | New Copy | Old Copy | Done
-- | -- | -- | --
[Home](https://monitor.firefox.com/) | See if you’ve been part of an online data breach. | See if you’ve been in an online data breach. | ✅
[Home](https://monitor.firefox.com/) | Get ongoing breach monitoring for multiple email addresses. | Search all your email addresses for breaches and get alerts about new threats. | ✅
[Breach Detail: Website](https://monitor.firefox.com/breach-details/MyFitnessPal) | On January 1, 2018, MyFitnessPal was breached. Once the breach was discovered and verified, it was added to our database on February 2, 2019. | On January 1, 2018, MyFitnessPal suffered a breach. Once the breach was discovered and verified it was added to our database on February 2, 2019.
[Breach Detail: Website](https://monitor.firefox.com/breach-details/MyFitnessPal) | What to do after a website breach: | What to do after a data breach:
[Breach Detail: Data Aggregator](https://monitor.firefox.com/breach-details/exactis) | On June 1, 2018, Exactis was breached. Once the breach was discovered and verified, it was added to our database on July 25, 2018. | On June 1, 2018, Exactis suffered a breach. Once the breach was discovered and verified it was added to our database on July 25, 2018.
[Breach Detail: Data Aggregator](https://monitor.firefox.com/breach-details/exactis) | What to do after a data aggregator breach: | What to do after a data aggregator breach
[Breach Alert Promo ](https://monitor.firefox.com/?breach=DataCamp)| Firefox Monitor warns you if your online accounts were involved in a data breach. Find out if you’ve been part of a data breach, get alerts about new breaches, and take steps to protect your online accounts. Firefox Monitor is a service provided by Mozilla. | Firefox Monitor warns if your online accounts were involved in a data breach. Find out if you’ve been in a data breach, get alerts about new breaches, and take steps to protect your online accounts. Firefox Monitor is a service provided by Mozilla.
[Dashboard](https://monitor.firefox.com/user/dashboard) | Known data breaches have exposed your info | Data breaches have exposed your information
[Preferences](https://monitor.firefox.com/user/preferences) | Update link on Change Primary Email Address button | This should link to: https://support.mozilla.org/en-US/kb/change-primary-email-address-firefox-accounts
[...]

Issue #1024, opened 2019-05-28
"What do do after a data breach" has a trailing colon


pdehaan:
via jdavidson in Slack:

> https://monitor.firefox.com/breach-details/Houzz — see the blue link under the Change your password tip that says “What to do after a data breach:” — it looks like that’s an error?

![Slack_-_Mozilla](https://user-images.githubusercontent.com/557895/58499970-114d3380-8136-11e9-9a14-b810b31d0bfb.png)

Ryan Gaddis also pointed to this design spec:

![Slack_-_Mozilla](https://user-images.githubusercontent.com/557895/58500020-33df4c80-8136-11e9-9517-563314cda2e2.png)

---

via _Originally posted by @pdehaan in https://github.com/mozilla/blurts-server/issues/1011#issuecomment-496014499_

The label for the "Change your password" section seems wrong (only one with a semicolon).

![](https://user-images.githubusercontent.com/557895/58384720-eaf29100-7f99-11e9-9e27-06e98db68d03.png)

---

**UPDATE 1:** Rogue trailing semicolon, it's also visible on the homepage:

![Firefox_Monitor](https://user-images.githubusercontent.com/557895/58494781-bd891d00-812a-11e9-87e6-5e0726bf89f4.png)

---

**UPDATE 2:** Also on the security tips page heading:

![Firefox_Monitor](https://user-images.githubusercontent.com/557895/58504985-b967fa00-8140-11e9-997b-2e5a3a78b05f.png)

pdehaan:
The semi-interesting thing is that "What do do after a data breach:" is used as both a heading and link text. So I _think_ we need two l10n strings, one with a trailing colon and one without.

betsymi:
Addressed this in #1034, but putting it here too for visibility since two things need to change:

+ Remove colon from "What to do after a data breach" string entirely so it no longer appears here or on Security Tips content block that surfaces elsewhere
+ Adjust headline to **What to do after a website breach:** with a colon. On the handful of data aggregators, this should read **What to do after a data aggregator breach:** with a colon. 

![Colon Confusion](https://user-images.githubusercontent.com/43555492/58568789-79654d80-81fa-11e9-8ec1-28e1e3df2946.png)

PR #1113, opened 2019-07-11
Remove semi-colon from list item heading

arku:
This is a super minor PR removing the semi-colon from "What to do after a data breach:". The motivation is to keep the list item headings consistent.

**Before**

image

**After**

image

### DIFF
--- a/locales/en-CA/app.ftl
+++ b/locales/en-CA/app.ftl
@@ -239,7 +239,7 @@ your-info-was-discovered-headline = Your information was discovered in a new dat
 your-info-was-discovered-blurb =
     You’re signed up to receive { -product-name } alerts 
     when your email appears in a data breach. Here’s what we know about this breach.
-what-to-do-after-breach = What to do after a data breach:
+what-to-do-after-breach = What to do after a data breach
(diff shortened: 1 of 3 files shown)


arku:
@flodolo The tests fail with `Failed: "Access denied. Certificate could not be retrieved"` error which is unrelated to my code changes.

Related to https://github.com/mozilla/blurts-server/issues/1064

flodolo:
I'm flagged automatically for review, but this is not really on me to review. That's a question for English copywriters.

arku:
> I'm flagged automatically for review, but this is not really on me to review. That's a question for English copywriters.

@flodolo Ah, no worries!

pdehaan:
Ref: #1024 

I think the problem was that we’re using that string in two different places/contexts, and we may need two different versions of the string (one with semicolon and one without).

groovecoder:
@betsymi have we decided what to do here? Are we going to remove or change this text/blurb in any of the upcoming copy changes?

betsymi:
This was addressed in #1034. I do not know if those copy changes were put through. See also #1024

groovecoder:
Okay, then I'm going to call this fixed and close it.

PR #1124, opened 2019-07-14
Copy refinements.

lesleyjanenorton:
Addressing more of the issues in #1034

### DIFF
--- a/locales/en/app.ftl
+++ b/locales/en/app.ftl
@@ -23,6 +23,7 @@ error-hibp-throttled = Too many connections to {-brand-HIBP}.
 error-hibp-connect = Error connecting to {-brand-HIBP}.
 error-hibp-load-breaches = Could not load breaches.
 error-must-be-signed-in = You must be signed in to your {-brand-fxa}.
+error-to-finish-verifying = To finish verifying this email for {-product-name}, you must be signed in under your primary account email.
 
 home-title = {-product-name}
 home-not-found = Page not found.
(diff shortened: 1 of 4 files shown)


locales/en/app.ftl
```diff
@@ -225,8 +228,8 @@ how-fxm-3-blurb = If you use {-brand-name}, you’ll receive a notification if y
   site that’s been breached. Find out right away if you were part of that breach 
   and what you can do about it.
 
-wtd-after-website = What to do after a website breach
-wtd-after-data-agg = What to do after a data aggregator breach
+wtd-after-website = What to do after a website breach:
```
flodolo: Do we care about other localizations picking up this change?

locales/en/app.ftl
```diff
@@ -33,6 +34,8 @@ scan-title = {-product-name} : Scan Results
 
 user-add-invalid-email = Invalid Email
 user-add-email-verify-subject = Verify your subscription to {-product-name}.
+user-add-duplicate-email = This email has already been added to {-product-name}.
+user-add-duplicate-email-part-2 = Visit your { $preferencesLink } to check the status of { $userEmail }.
```
flodolo: What is replacing `$preferencesLink`? Please add a comment

locales/en/app.ftl
lesleyjanenorton: What are your thoughts on this? If it doesn't change much in other locales, then I suppose not?

flodolo: Let's change it without new IDs. After all, they are supposed to look at the website and decide if it works or not for their language.

groovecoder:
The Travis failures are un-related to these changes (and are fixed in latest master). Merging.

groovecoder:
Oh f. I overlooked @flodolo's comments in my hurry to merge this before going out on PTO.

flodolo:
At this point it's only missing a comment. The sooner we fix it, the better.

PR #1138, opened 2019-07-19
fix #1129: echo errors from FXA to client

### LINKED ISSUE #1129: Return `401 Unauthorized` when receiving an invalid FxA OAuth token
The `/breach-stats` endpoint added in #1111 accepts an FxA OAuth bearer token for authorization, and returns a `404 Not Found` error response if the token is invalid. Please consider returning a `401 Unauthorized` in this case, if only for consistency with other FxA-OAuth-powered services.

This will also make it easier for client code to distinguish between "the oauth token was invalid, so we can't tell you anything about the user" and "the oauth token was fine, and we don't know anything about the user".

More broadly, I suggest echoing any non-200 status from the FxA `/verify` endpoint back to the client, e.g. if you get a `503 Service Unavailable` from FxA then also report a `503 Service Unavailable` back to the client. This generally tends to be helpful for debugging any failures that might occur.

groovecoder:


controllers/user.js
```diff
@@ -347,9 +347,9 @@ async function getBreachStats(req, res) {
     });
   }
   const fxaResponse = await FXA.verifyOAuthToken(req.token);
-  if (!fxaResponse) {
-    return res.status(404).json({
-      errorMessage: "Cannot find FXA for that OAuth token.",
+  if (fxaResponse.name === "HTTPError") {
```
pdehaan: Are we sure `fxaResponse` will be non-null here?
I _assume_ so, looking at the `catch..return` below, but just checking.

groovecoder: Yes, there will always be some kind of `fxaResponse` value returned.

PR #1139, opened 2019-07-19
DNM: fix #1128: check fxa token for Monitor scope

### LINKED ISSUE #1128: Check for an appropriate scope on FxA access tokens
The `/breach-stats` endpoint added in https://github.com/mozilla/blurts-server/pull/1111 accepts an FxA OAuth bearer token for authorization, but it does not check anything about the token beyond "is it valid". This means that anyone who can get any OAuth token to access anything on the user's account (e.g. other FxA reliers like AMO or Pocket, and potentially third-party RPs in future) could use their token to access this endpoint and read the user's breach-protection status.

Instead, we should check that the received token is actually intended for accessing this service.

The way we do that in FxA currently is via scopes. We should allocate a unique scope value for this service, e.g. "https://identity.mozilla.com/apps/monitor", and then:

* Have the client code generate a token with this scope when accessing monitor
* Have the server code explicitly check that this value is in the list of scopes returned when verifying the OAuth token

/cc @shane-tomlinson; does that line up with your expectations for how this should work?

groovecoder:
Don't merge this unless/until [the client-side code](https://phabricator.services.mozilla.com/D38228) has landed in central.

### DIFF
--- a/controllers/user.js
+++ b/controllers/user.js
@@ -12,6 +12,9 @@ const { resultsSummary } = require("../scan-results");
 const sha1 = require("../sha1-utils");
 
 
+const FXA_MONITOR_SCOPE = "https://identity.mozilla.com/apps/monitor";
+
+
 async function _requireSessionUser(req,res) {
   if (!req.session || !req.session.user) {
     // TODO: can we do a nice redirect to sign in instead of an error?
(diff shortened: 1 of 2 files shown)


lissyx:
@rfk @groovecoder It looks like https://phabricator.services.mozilla.com/D38228 landed, is it possible this is why we are hitting https://github.com/mozilla/blurts-server/issues/1146 ?

rfk:
That desktop patch *should* work with both the existing monitor code and the new version from this PR, but it's possible something is not going quite right. I'll see what I can find out...

ashishthukral:
@rfk i had reported #1146 . i tried resetting the password. then clearing out existing logged in sessions from firefox account for monitor.firefox.com. But both steps did not help

groovecoder:
We're fixing #1146 in https://github.com/mozilla/blurts-server/pull/1149.

Merging this so to make it into this week's release for testing with production Firefox (Nightly) clients.

PR #1150, opened 2019-07-29
fix #1109: inlude BreachDataClasses in remote settings record

### LINKED ISSUE #1109: Add DataClasses to fxmonitor-breaches collection records in Remote Settings
For https://bugzilla.mozilla.org/show_bug.cgi?id=1563802, we **DON'T** want to show breach indicators on logins for sites where the breach does **NOT** include passwords.

To do so, we need to include some kind of `containedPasswords` field in the `fxmonitor-breaches` records in Remote Settings. We will likely want the ability to implement other password manager UI features based on the data classes involved in breaches. So, we may as well add all of the `DataClasses` field to the records while we are implementing this.

groovecoder:
Now that https://github.com/mozilla-services/remote-settings-permissions/pull/62 is merged and deployed, this works on both stage & prod Remote Settings servers.

### DIFF
--- a/scripts/updatebreaches.js
+++ b/scripts/updatebreaches.js
@@ -35,6 +35,7 @@ if (
       BreachDate: breach.BreachDate,
       PwnCount: breach.PwnCount,
       AddedDate: breach.AddedDate,
+      DataClasses: breach.DataClasses,
     };
 
     console.log("New breach detected: \n", data);


groovecoder:
Note: we may need to empty the collections in order to force the script to re-upload all of the breaches with the data classes included.

scripts/updatebreaches.js
```diff
@@ -35,6 +35,7 @@ if (
       BreachDate: breach.BreachDate,
       PwnCount: breach.PwnCount,
       AddedDate: breach.AddedDate,
+      BreachDataClasses: breach.DataClasses,
```
nhnt11: Maybe consider calling this simply "DataClasses"? That would keep the parity since everything else follows the HIBP names.

groovecoder: Dang, you're right ... but I already updated the Remote Settings schema with this new field name. I imagine it's harder to update a field than to add a new one? I'll check it out tomorrow though.

groovecoder: Updated name to just `DataClasses` and filed https://github.com/mozilla-services/remote-settings-permissions/pull/68 for the corresponding change to the RS schema.

groovecoder:
https://github.com/mozilla-services/remote-settings-permissions/pull/68 is merged, deployed on stage & prod, and I spot-checked on stage. So I'm merging this for the prod job to start adding this data.

I have a review request on https://settings-writer.prod.mozaws.net/v1/admin/#/buckets/main-workspace/collections/fxmonitor-breaches/records to remove a single record to test that tonight's `updatebreaches.js` script restores the record including is `DataClasses` before I remove all the records (tomorrow).

Issue #1148, opened 2019-07-29
Switch to server-side session storage


groovecoder:
https://github.com/mozilla/blurts-server/issues/1146 and https://github.com/mozilla/blurts-server/issues/1084 show that we can't rely on client-sessions forever. As we put more data into user session objects, we need to move it to server-side session storage.

@mozilla/svcops - what are the suggested storage backends for (node) sessions?

Micheletto:
> @mozilla/svcops - what are the suggested storage backends for (node) sessions?

We should use Google's [MemoryStore](https://cloud.google.com/memorystore/) (which is Redis) for this type of data.

Issue #1146, opened 2019-07-23
Cant sign in to monitor.firefox.com. Error - "You must be signed in to your Firefox Account."


ashishthukral:
On trying to sign in to https://monitor.firefox.com/user/dashboard, it simply keeps giving error "You must be signed in to your Firefox Account."

been facing this since 07/22, even though i was able to add multiple emails to monitoring service initially. but been unable to login to the monitor service on chrome or firefox.

lissyx:
Looks like you are not alone: https://bugzilla.mozilla.org/show_bug.cgi?id=1559718

I'm also hit by this.

ddurst:
(Carrying over NI from bugzilla for @groovecoder and @lesleyjanenorton just for completeness.)

ashishthukral:
linked PR - https://github.com/mozilla/blurts-server/pull/1139

rfk:
As far as I can tell, the change in #1139 (and corresponding desktop patch in https://phabricator.services.mozilla.com/D38228) don't seem related to this issue; they are for access monitor data within Firefox UI as part of the protection report, but this issue seems to be something to do with the state of the monitor website.

Just to confirm: @lissyx and @ashishthukral, this happens to you every time you try to sign in with your Firefox Account? You click on "sign in", go through the Firefox Accounts OAuth flow successfully, it lands you on https://monitor.firefox.com/user/dashboard, and you get an error message displayed on that page?

I *think* this means that something on the monitor side is failing to set `session.user`, or perhaps setting `session.user` to a falsey value, at the end of the FxA OAuth flow. But I can't see any obvious code paths that would do that (without throwing some other unrelated error first).

rfk:
One thing that would be valuable to check is, after you do the OAuth flow and get this signin error, open the dev console, go to the "storage" tab, and see what cookies you have for monitor.firefox.com. There should be a cookie there named "session" and it would be useful to know how big the value is that it contains (the actual value is encrypted).

ashishthukral:
> As far as I can tell, the change in #1139 (and corresponding desktop patch in https://phabricator.services.mozilla.com/D38228) don't seem related to this issue; they are for access monitor data within Firefox UI as part of the protection report, but this issue seems to be something to do with the state of the monitor website.
> 
[...]

PR #1152, opened 2019-07-31
for #1146: limit to 15 email addresses until #1148 is done

groovecoder:


lesleyjanenorton:
Do we want to add anything to this message to indicate that the maximum number is 15 and that the number of maximum emails will increase soon?

locales/en/app.ftl
```diff
@@ -33,6 +33,7 @@ oauth-invalid-session = Invalid session
 scan-title = {-product-name} : Scan Results
 
 user-add-invalid-email = Invalid Email
+user-add-too-many-emails = You are monitoring the maximum number of email addresses.
```
flodolo: I assume you'd get this after trying to add another email, so maybe "you are already…"?

lesleyjanenorton: +1

groovecoder: The message could potentially show up in 2 places.

1. It always replaces the "add another email form" when the user already has the maximum number
2. If the user somehow hacks together a form request to add another email address when they're at the maximum, they will get an error page displaying this same message.

So, since the 2nd case should never really happen, I dropped the "already" term.

views/partials/dashboards/preferences.hbs
```diff
@@ -32,8 +32,14 @@
         {{/each}}
       </div>
       <div class="pref">
-        <h4 class="email-pref add-new">{{ getString "add-new-email" }}</h4>
-        {{> forms/add-another-email-form csrfToken=../csrfToken}}
+        {{!-- TODO: remove this when https://github.com/mozilla/blurts-server/issues/1148 is fixed --}}
+        {{#ifCompare emails.total ">=" 15}}
+          <h4 class="email-pef">{{ getString "user-add-too-many-emails" }}</h4>
```
pdehaan: is this a typo? Should it be `class="email-pref"`?

groovecoder: Whoops, yeah.

pdehaan: Fixed via #1153 PR

PR #1157, opened 2019-08-01
Forward utms to fxa

### LINKED ISSUE #844: Forward UTM parameters to FxA login/registration from referrals to the landing page
SUMO is cross-promoting Monitor, and proper attribution (`utm_source`, `utm_medium`, `utm_campaign`, `utm_content`, and `utm_term`) needs to get forwarded to the FxA signup form on the landing page. This would support cross-promotion from other referral sources in the future as well.

lesleyjanenorton:
This saves the UTM parameters of referring sites and forwards them to FxA when a user clicks one of the FxA entrypoints (sign in / log in links and buttons) from the landing page.

Closes #844

### DIFF
--- a/public/js/fxa-analytics.js
+++ b/public/js/fxa-analytics.js
@@ -49,15 +49,38 @@ async function sendPing(el, eventAction, eventLabel = null) {
   }
 }
 
+
 function getFxaUtms(url) {
-  const utmSource = encodeURIComponent(document.body.dataset.serverUrl.replace(/(^\w+:|^)\/\//g, ""));
-  url.searchParams.append("utm_source", utmSource);
-  url.searchParams.append("utm_campaign", document.body.dataset.utmCampaign);
+  const bodyDataset = document.body.dataset;
+  getUTMNames().forEach(param => {
+    if (bodyDataset[param]) {
+      url.searchParams.append(param, encodeURIComponent(bodyDataset[param]));
+    }
+  });
   url.searchParams.append("form_type", "email");
   return url;
 }
 
+function saveReferringPageData(utmParams) {
+  const bodyDataset = document.body.dataset;
+  getUTMNames().forEach(param => {
+    if (utmParams.has(param)) {
+      const cleanParam = utmParams.get(param);
+      bodyDataset[param] = cleanParam.replace(/[&<>"',.`=\/]/g, "");
+    }
+  });
+}
+
+function getUTMNames() {
+  return ["utm_source", "utm_medium", "utm_campaign", "utm_term", "utm_content"];
+}
+
 (() => {
+  const win = window;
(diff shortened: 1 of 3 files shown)


public/js/fxa-analytics.js
```diff
@@ -49,15 +49,38 @@ async function sendPing(el, eventAction, eventLabel = null) {
...
+  getUTMNames().forEach(param => {
+    if (utmParams.has(param)) {
+      const cleanParam = utmParams.get(param);
+      bodyDataset[param] = cleanParam.replace(/[&<>"',.`=:/]/g, "");
+    }
+  });
+}
+
+function getUTMNames() {
+  return ["utm_source", "utm_medium", "utm_campaign", "utm_term", "utm_content"];
```
pdehaan: Is this everything? Or should we do some generic parser that looks for any query param that starts with `utm_*`?
Sorry, I'm not a GA expert and don't know how flexible these query params are.

public/js/fxa-analytics.js
```diff
@@ -49,15 +49,38 @@ async function sendPing(el, eventAction, eventLabel = null) {
...
+}
+
 (() => {
+  const win = window;
+  if (win.location.search.includes("utm_") && win.history.replaceState) {
+    saveReferringPageData(new URL(win.location).searchParams);
+    win.history.replaceState({}, "", win.location.toString().replace(/[?&]utm_.*/g, ""));
```
pdehaan: This may also be a bit greedy.

I tried in regexr.com with this[1] made up URL, and it takes everything after `utm_` to the end of the line, even the &age=999&foo=bar. Not sure if we ever expect anything else on the querystring apart from UTMs, but just throwing that out there.

```
[1] https://monitor.firefox.com/scanner/?name=peter&utm_source=ok&age=999&foo=bar
```

public/js/fxa-analytics.js
lesleyjanenorton: This is a good point. These were the parameters we were specifically asked to forward in #844 and my thinking was that limiting the parameters to just these would prevent us from potentially keeping and forwarding a bunch of unnecessary junk.

public/js/fxa-analytics.js
lesleyjanenorton: So I think the UTMs are technically supposed to be at the end of the URL, in which case clipping everything after the `?utm_` is what we want and I don't think we're currently wanting any values that could feasibly come after the UTMs to stick around in the url bar after the UTMs have been clipped out in any case.

groovecoder:
Code and spot-check looks good. This didn't send any of the `utm_*` params to the `metrics-flow` endpoint, but #844 was only about landing page analytics for GA.

So, I want to merge this to fix #844 and we can fix anything else in follow-ups.

PR #1159, opened 2019-08-04
Updated email-strings.ftl

### LINKED ISSUE #567: Small grammar issue (log into > log in to) in email-strings.ftl
A line says "They can use that password to log into your other accounts."

The "log into" part should be changed to "log in to". Also see [here](https://github.com/mozilla/blurts-server/commit/301341f43e6671f076b9444d0e846ee2cc2f31a5#diff-967a8adff6e98f3795337ef56f485989R62).

Patil2099:
Fixes #567
Small grammar issue (log into > log in to) in email-strings.ftl

### DIFF
--- a/locales/en-CA/email-strings.ftl
+++ b/locales/en-CA/email-strings.ftl
@@ -34,7 +34,7 @@ report-pwt-blurb =
 report-pwt-headline-1 = Use a different password for every account
 report-pwt-summary-1 =
     Reusing the same password everywhere opens the door for hackers. 
-    They can use that password to log into your other accounts.
+    They can use that password to log in to your other accounts.
 report-pwt-headline-2 = Create strong, unique passwords
 report-pwt-summary-2 =
     Hackers use lists of common passwords to try to guess yours. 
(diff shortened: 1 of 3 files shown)


flodolo:
Change looks good to me, but in this specific case it might make sense to also fix en-CA and en-GB directly in this pull request.

pdehaan:
Semi-related, and I'm not great with words, but can somebody do a quick sanity check of the other "into" usages in the EN strings?

```sh
# Check for "into" usages, but ignore any commented lines:
$ git grep " into " locales/en | grep -v ":#"

1. locales/en/app.ftl:  passwords securely and fill them into websites for you. Use a password manager

2. locales/en/email-strings.ftl:we-sent-this-alert = We sent this message to { $userEmail } because the email address opted into alerts from {-product-name}.

3. locales/en/email-strings.ftl:  They can use that password to log into your other accounts.

4. locales/en/email-strings.ftl:  and fill them into websites so you don’t have to remember every single one.
```

We don't have to address them in this PR, just curious if they're obviously incorrect and I should open a new issue.

PR #1160, opened 2019-08-05
Restore "see-all-breaches" string.

lesleyjanenorton:


flodolo:
Is this only used in emails? Should it live in emails.ftl?

lesleyjanenorton:
@flodolo This string is only being used in emails and would be better housed in `emails.ftl` but I noticed that it is still available in `locales/cak/app.ftl`. If it can live in `emails.ftl` though I will move it.

flodolo:
It's still in cak because they're behind, and Pontoon never had a chance to clean up the file. Ignore cak ;-)

groovecoder:
code & spot-check look good.

PR #1163, opened 2019-08-05
bug 1569750: return entire FxA response to client

groovecoder:


groovecoder:
image

groovecoder:
Oh crap, I meant to put the scope-check up in the controller. Will restore ASAP.

rfk:
(IMHO, FxA should make this easier for you by accepting a `scope` parameter as part of the verification request, but it doesn't behave that way right now)

pdehaan:
> but it doesn't behave that way right now

Dearest @rfk,
Do you have an upstream bug or something we can follow for this?

rfk:
>  Do you have an upstream bug or something we can follow for this?

This was proposed a while ago in https://github.com/mozilla/fxa-oauth-server/issues/381, but it's not currently being tracked or worked on. It's probably more expedient to do the check locally here in monitor for now.

PR #1164, opened 2019-08-05
restore check for body.scope.includes monitor

groovecoder:


pdehaan:
Ref: #1163

controllers/user.js
```diff
@@ -354,7 +354,7 @@ async function getBreachStats(req, res) {
     });
   }
   const fxaResponse = await FXA.verifyOAuthToken(req.token, FXA_MONITOR_SCOPE);
-  if (fxaResponse.name === "HTTPError") {
+  if (fxaResponse.name === "HTTPError" || !fxaResponse.body.scope.includes(FXA_MONITOR_SCOPE)) {
```
pdehaan: nit: Semi-related, but it doesn't look like `FXA.verifyOAuthToken()` uses the `scope` argument anymore.

https://github.com/mozilla/blurts-server/blob/c9dc4941cddd2d97db6a1eea9317536a27bb5617/lib/fxa.js#L33-L40

Not sure if it's any cleaner to remove that argument to reduce visual clutter (and unnecessary arg passing).

rfk: I recommend removing it to reduce possible future confusion, esp. since there's no test that would prevent a future refactor from assuming that `verifyOAuthToken` checks the passed-in scope.

controllers/user.js
```diff
@@ -354,7 +354,7 @@ async function getBreachStats(req, res) {
...
     return res.status(fxaResponse.statusCode).json({
```
rfk: In the `FXA_MONITOR_SCOPE` case, this is going to return a `200` response with an error message; you'll need to special-case it to return a `401` or the client will get confused.

groovecoder:
Finally got back to update this. But I'm having trouble spot-checking this on my local (using https://stable.dev.lcip.org/). Does stable give the `https://identity.mozilla.com/apps/monitor` scope to tokens for the `blurts-server (local)` app? (client ID `edd29a80019d61a1`)? Or do I need to make a new OAuth app for it to give the monitor scope?

rfk:
>  Does stable give the https://identity.mozilla.com/apps/monitor scope to tokens for the
> blurts-server (local) app? (client ID edd29a80019d61a1)?

From slack convo, it sounds like @vladikoff flipped the necessary config bits to allow this.

groovecoder:
I was able to spot-check:

1. Nightly with a profile using oauth-stable for FxA
2. Sign into localhost Monitor on this branch
3. Send the oauth-stable Nightly OAuth token thru localhost /user/breach-stats and receive the expected `{"monitoredEmails":1,"numBreaches":1,"passwords":0}` output.

I've got a task to go thru this code-base to audit test coverage. I'll cover as many of these cases as I can ... but yeah - sometimes setting up all the mocks is just as brittle as spot-checking the real service. 😐 I'll do my best to cover it though.

@rfk or @vladikoff - can we make sure this scope is being given to client tokens on all environments?

PR #1188, opened 2019-09-04
Fix #1182

### LINKED ISSUE #1182: Zero results after signing in with FxA and email case doesn't match
So, apparently I appear in 13 breaches if I don't sign in, but if I sign in, my same email appears in no breaches at all. Not a very good system at monitoring if the same exact email signed in isn't being detected properly. And yes, I literally mean the same exact. I logged out and checked the same email copy and pasted and it shown up in 13. So signing in shows zero breaches while signing out shows all.

image

lesleyjanenorton:
- Lower cases new emails before adding to DB.
-  Updates `bundleVerifiedEmails` to convert existing user emails to lowercase letters before making a new sha1 hash (of the newly lowercased email) to use for scanning breaches. 

Screen Shot 2019-09-04 at 5 58 25 PM
Screen Shot 2019-09-04 at 5 58 31 PM

### DIFF
--- a/controllers/user.js
+++ b/controllers/user.js
@@ -96,7 +96,7 @@ function _checkForDuplicateEmail(sessionUser, email) {
 
 async function add(req, res) {
   const sessionUser = await _requireSessionUser(req);
-  const email = req.body.email;
+  const email = req.body.email.toLowerCase();
   if (!email || !isemail.validate(email)) {
     throw new FluentError("user-add-invalid-email");
   }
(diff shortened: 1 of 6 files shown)


controllers/user.js
```diff
@@ -136,7 +136,8 @@ async function add(req, res) {
 }
 
 async function bundleVerifiedEmails(email, emailSha1, ifPrimary, id, verificationStatus, allBreaches) {
-  const foundBreaches = await HIBP.getBreachesForEmail(emailSha1, allBreaches, true);
+  const lowerCaseEmailSha = sha1(email.toLowerCase());
```
pdehaan: Is `emailSha1` now unused?

pdehaan:
> * Lower cases new emails before adding to DB.

Do we need to update the existing records in the DB and lowercase some emails/SHAs?

controllers/user.js
lesleyjanenorton: indeed it is!

lesleyjanenorton:
@pdehaan re: your last comment, I would think that updating existing records would be ideal right? @groovecoder what would that look like/is it feasible? 

Also curious: I'm no longer able to add variations of `lesleyjanenorton@gmail.com` without triggering the duplicate email error.

Issue #903, opened 2019-05-12
[v2] Breach cards stretch to fill space


pdehaan:
# Steps to reproduce

1. Go to https://monitor-v2.herokuapp.com/breaches
2. Select "Sensitive Breaches" and "Data Aggregator Breaches".
3. Scroll to bottom.

## Actual results
Last card is 100% width instead of 33% width.

![All_the_breaches](https://user-images.githubusercontent.com/557895/57578428-22f9c000-7441-11e9-827d-4c7361b6a830.png)

If you select all breaches, the last 2 cards are 50% each:

![All_the_breaches](https://user-images.githubusercontent.com/557895/57578431-43297f00-7441-11e9-827a-e7fc5d9e21bb.png)

EMMLynch:
Closing since we've redesigned the site and functionality since this was created. If you feel that this is still needed, please let me know.

PR #1189, opened 2019-09-05
Fix shrinking breach cards in dashboard.

lesleyjanenorton:
Noticed this while working on #1188. If you have an email with five breaches, the fifth breach card will appear much smaller than the other four. Related: #903 (which I think is now resolved?). 
Screen Shot 2019-09-05 at 1 01 38 PM

### DIFF
--- a/public/css/breach-cards.css
+++ b/public/css/breach-cards.css
@@ -138,10 +138,6 @@ a.latest-breach-link {
   max-width: 47%;
 }
 
-.email-card .two-up {
-  max-width: 100%;
-}
-
 .three-up {
   flex: 1 1 30%;
   min-width: 260px;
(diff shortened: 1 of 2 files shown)


groovecoder:
I couldn't reproduce the original bug on either localhost or on https://fx-breach-alerts.herokuapp.com/. I used test4@mailinator.com which has 5 known breaches.

pdehaan:
I can reproduce w/ the logged in state, but not the logged out state.

### Logged in

Firefox_Monitor

---

### Logged out

Firefox_Monitor___Scan_Results

groovecoder:
Reproduced (with test4@mailinator.com) and saw the fix.

PR #1195, opened 2019-09-23
New breach logos

lesleyjanenorton:
Adds logos for **KiwiFarms**, **Minehut**, and **LuminPDF**...still rooting around for a suitable "void.to" logo.


pdehaan:
> still rooting around for a suitable "void.to" logo.

Submitted as #1199

Issue #816, opened 2019-02-26
Refresh FxA profile data on page-loads


sandysage:
Based on Vijay's findings in the [Monitor and FxA Integration](https://docs.google.com/document/d/12Fah9i1tYZLiPwUIUpjslfY1dpKZIyI7b6N4Uv0GzQw/edit#heading=h.lxh3sysaocd7), after someone deletes their account, the profile information is still available on Monitor for 15 minutes - When a user deletes their Firefox Account, their Monitor login session remains valid until it expires in 15 minutes.

We want to get this addressed so that the FxA profile information is updated more frequently - on page load. With a timeout of 500ms or 1s.

sandysage:
Let's also use this ticket to update both the avatar image AND the email address.

groovecoder:
Fix for this started in https://github.com/mozilla/blurts-server/pull/1352 but I'm not 100% sure I'm happy with the approach yet.

PR #1197, opened 2019-09-24
fix #1131: destroy old refresh tokens

### LINKED ISSUE #1131: Delete any previous `refresh_token` when processing a new user signin
As described in https://github.com/mozilla/fxa/issues/1802, each time a user signs in to the Monitor website we end up generating a new FxA `refresh_token` and storing it in the Monitor backend DB, overwriting any existing `refresh_token` that was generated by a previous login.  Over time, this results in an accumulation of orphaned `refresh_token`s that are held by the FxA server, but have been deleted from Monitor.

Please consider destroying any existing `refresh_token` when writing a new one to the db.

groovecoder:


db/DB.js
```diff
@@ -277,6 +278,9 @@ const DB = {
     })
     .returning("*");
     const updatedSubscriber = Array.isArray(updated) ? updated[0] : null;
+    if (updatedSubscriber) {
+      FXA.destroyOAuthToken({refresh_token: subscriber.fxa_refresh_token});
```
pdehaan: Should this be `subscriber.` or `updatedSubscriber.`?

And how is `subscriber.fxa_refresh_token` different from `fxaRefreshToken`? Or is the `subscriber.fxa_refresh_token` just the _old_ token that you want to destroy, I guess...

groovecoder: It should be `subscriber` - the *old* subscriber data with the `fxa_refresh_token` that we want to destroy.

pdehaan: Does this mean you couldn't be logged into two browsers/sessions at once?
I'm also trying to think of any cases where we'd have orphaned session tokens, or if the logic for "delete all of this subscriber's FxA OAuth tokens _except_ for this latest one, makes any sense.

But yeah, R+ (unless @lesleyjanenorton has any thoughts or objections). Once this lands on dev we can try running a few more scenarios.

groovecoder:
Multiple sessions should be okay, as we currently only use the refresh token during sign in/up. If/when we get to https://github.com/mozilla/blurts-server/issues/816 we might break multiple sessions at once unless we find a way to only destroy tokens that are *not* associated with an active session.

PR #1200, opened 2019-09-26
Add bento

lesleyjanenorton:
Screen Shot 2019-09-26 at 12 23 12 PM

### Relevant Context & Vocabulary
**Bento** (pictured above) - Hereinafter "bento" means a button that you can press to reveal (in our case) a menu of Firefox apps and services. You may also have heard this referred to as "Dandelion" because it promotes cross-pollination.

The idea is that Monitor with implement The Bento by Skyline and that eventually The Bento might move to its own repository (and maybe server) where other Firefox product pages (tbd) can easily adopt it.

**This PR...** adds the necessary Bento files and rearranges existing header elements to accommodate The Bento and match the spec.

`fx-bento.js` looks for the `<firefox-apps>` custom element then builds The Bento items from there using localized strings fetched from `/getBentoStrings` endpoint and sprites from `/img/fx-bento-sprites.png`. The l10n fetch may be overkill if other product pages want to do their own thing with localization instead of fetching localized strings. 

The `._currentSite` is determined by a data attribute -`document.body.dataset.bentoAppId`- which stores the site's Bento ID (will require future organization between Bento maintainers and any consuming website). The current site's link should appear at the top of the link list inside The Bento.

On small screens The Bento menu should span the full height and width of the screen and a round purple button with an "X" in it should appear in the upper right hand corner. This button should close The Bento menu when clicked. There is also an invisible pseudo element  - `.fx-bento-mobile-close::after`-  which should span the upper-most portion of the mobile Bento menu and also close The Bento menu when clicked.

#### To be thought about now and figured out later:
We may need to figure out a unified metrics system (not Amplitude, too expensive) for measuring basic operational Bento data across all the websites on which The Bento maintains a residence. The other option is that each Bento-consuming website can implement their own analytics (Send and FxA don't use GA) but that means more work for all involved and more time spent hunting down data).


locales/en/bento.ftl
```diff
@@ -0,0 +1,10 @@
+bento-button-title = Firefox apps and services
+fx-makes-tech = Firefox is tech that fights for your online privacy.
+made-by-Mozilla = Made by Mozilla
+fx-desktop = Firefox Browser for Desktop
+fx-mobile = Firefox Browser for Mobile
```
lesleyjanenorton: @flodolo  Do you know if there is a consensus on how these should be localized? I'm assuming the "Browser for Desktop" and "Browser for Mobile" bits should be translated.

locales/en/bento.ftl
```diff
@@ -0,0 +1,10 @@
+bento-button-title = Firefox apps and services
```
flodolo: Can we use the brand names from app.ftl here? e.g. `{ -brand-name } apps and services`.

locales/en/bento.ftl
flodolo: I kind of remember seeing a conversation about considering the whole "Firefox Browser" as a brand, but I'm not sure who can confirm if there are guidelines. I wouldn't be super happy if that's the case, to be honest.

locales/en/bento.ftl
```diff
@@ -0,0 +1,10 @@
...
+fx-makes-tech = Firefox is tech that fights for your online privacy.
+made-by-Mozilla = Made by Mozilla
+fx-desktop = Firefox Browser for Desktop
+fx-mobile = Firefox Browser for Mobile
+fx-lockwise = Firefox Lockwise
```
flodolo: I think we should create terms for all these brands in app.ftl (Lockwise, Pocket, Monitor, Send), and here reference those terms.

views/partials/imports.hbs
```diff
@@ -20,10 +23,10 @@
   <link rel="stylesheet" href="/css/sign-up-banner.css">
   <link rel="stylesheet" href="/css/subpage.css">
 
-
   <script type="text/javascript" src="/js/all-breaches/all-breaches.js"></script>
   <script type="text/javascript" src="/js/analytics_dnt-helper.js" defer></script>
   <script type="text/javascript" src="/js/analytics_tracking_protection.js" defer></script>
+  <script type="text/javascript" src="/js/fx-bento.js"></script>
```
pdehaan: Does this need a `defer`?

views/partials/imports.hbs
```diff
@@ -1,3 +1,5 @@
+
+
```
pdehaan: nit: remove these 2 blank lines

pdehaan:
CSS lint seems to be failing on CI with the following:

```
$ npm run lint:css
> blurts-server@0.0.1 lint:css /home/travis/build/mozilla/blurts-server
> stylelint --fix 'public/css/**/*.css'
public/css/fx-bento.css
   1:1  ✖  Unexpected unknown type selector "firefox-apps"   selector-type-no-unknown
 217:3  ✖  Unexpected unknown type selector "firefox-apps"   selector-type-no-unknown
 227:3  ✖  Unexpected unknown type selector "firefox-apps"   selector-type-no-unknown
```

Which seems to be choking on [`<firefox-apps></firefox-apps>`](https://github.com/mozilla/blurts-server/pull/1200/files?short_path=cae1e5d#diff-9cd89c0c23b215f2cdc8fe8c02b9d75bR13). Not sure what kind of browser support custom elements has. Possibly high support, if it is https://caniuse.com/#feat=custom-elementsv1.

Probably safe to disable that rule maybe?

locales/en/app.ftl
```diff
@@ -10,6 +10,10 @@
 -brand-Mozilla = Mozilla
 -brand-HIBP = Have I Been Pwned
 -brand-fxa = Firefox Account
+-brand-pocket = Pocket
+-brand-lockwise = Firefox Lockwise
+-brand-firefox-browser = Firefox Browser
```
flodolo: Wait, aren't we allow languages to localize browser? If so, this shouldn't be a brand, and strings should use `{-brand-name} Browser`

flodolo: One more note: we should change the comment above into a section comment (with two hashes, and an empty line after), and reset the section comment at the end of the brands block with just `##`.

locales/en/app.ftl
```diff
@@ -1,4 +1,5 @@
-# String IDs beginning with "-product" and "-brand" should remain in English.
+## String IDs beginning with "-product" and "-brand" should remain in English.
```
flodolo: Better comment:

```
## The following messages are brand and should be kept entirely in English
## unless otherwise indicated.
```

locales/en/app.ftl
flodolo: That's unfortunate. At this point, we should expose the entire "brand", otherwise it's a mess to localize piece by piece.

```
# "Firefox Browser" needs to remain in English, "for Desktop" can be localized
 -brand-firefox-browser-desktop = Firefox Browser for Desktop

# "Firefox Browser" needs to remain in English, "for Mobile" can be localized
 -brand-firefox-browser-mobile = Firefox Browser for Mobile
```

To clarify one point: the reason to have terms assigned to brands, is that it makes a lot easier to check what happens across languages.
- A message missing a reference brand will raise a warning in compare-locales
- You only need to look at that string, not all strings using the brand

flodolo:
Schermata 2019-09-28 alle 07 13 32

This is unlikely, but I think you want to have a max width for the menu, and have elements wrapping?

locales/en/app.ftl
flodolo: Let's hold on a bit more, the ongoing conversations seem to force us to use the current approach instead (keep "Firefox Browser" as a brand).

locales/en/bento.ftl
flodolo: Given the conversation that happened in Slack, I think we need to go back to your previous solution, with `Firefox Browser` defined as a term (and a clear comment that it needs to stays in English), and only that referenced in these strings.

It's unfortunate, but that's the only way to enforce that the term stays untranslated :-\

locales/en/app.ftl
```diff
@@ -1,16 +1,18 @@
...
-# - Transliterated.
-# - Translated.
+## The following messages are brand and should be kept entirely in English
+## unless otherwise indicated.
+
 -product-name = Firefox Monitor
 -product-name-nowrap = <span class="nowrap">{-product-name}</span>
 -product-short-name = Monitor
 -brand-name = Firefox
+-brand-firefox-browser = Firefox Browser
```
flodolo: This has been a rollercoaster. We're back to "Browser" being localizable, so we can remove this term.

locales/en/bento.ftl
```diff
@@ -0,0 +1,19 @@
+fx-monitor = {-product-name}
+pocket = {-brand-pocket}
+fx-send = {-brand-send}
+fx-lockwise = {-brand-lockwise}
+
+fx-desktop = {-brand-firefox-browser} for Desktop
+fx-mobile = {-brand-firefox-browser} for Mobile
```
flodolo: These two go back to need `{-brand-name}`, i.e.

```
fx-desktop = {-brand-name} Browser for Desktop
fx-mobile = {-brand-name} Browser for Mobile
```

groovecoder:
There's a conflict in `locales/en/bento.ftl` now?

locales/en/bento.ftl
```diff
@@ -0,0 +1,24 @@
+fx-monitor = {-product-name}
+pocket = {-brand-pocket}
+fx-send = {-brand-send}
+fx-lockwise = {-brand-lockwise}
+
+# If possible, keep "Firefox Browser" in English, but feel free 
+# to translate browser if that doesn't work for your language.
```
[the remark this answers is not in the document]
flodolo: Looks like this one was addressed on master but not in this branch

PR #1205, opened 2019-09-29
Add pre fxa message to breach alert emails.

lesleyjanenorton:
Screen Shot 2019-09-29 at 4 24 40 PM

Adds this message to breach alert emails sent to subscribers who signed up for Monitor before a Firefox Account was required.
Updates `/email-l10n` route for breach alerts accordingly.


controllers/hibp.js
```diff
@@ -24,18 +24,21 @@ function getAddressesAndLanguageForEmail(recipient) {
         recipientEmail: recipient.primary_email,
         breachedEmail: recipient.email,
         signupLanguage,
+        preFxaSubscriber: false,
```
lesleyjanenorton: @groovecoder Is there a better way/place to check this? I'm not totally familiar with all of this snazziness yet and am kind of winging it here.

locales/en/email-strings.ftl
```diff
@@ -160,3 +160,19 @@ faq-v2-3 = I just found out I’m in a data breach. What do I do next?
 
 # Link Title
 faq-v2-4 = How does {-product-name} treat sensitive sites?
+
+
+# This is the title of a link that is nested inside { pre-fxa-subhead }.
```
flodolo: This approach is painful but usable when it comes to a short link (like sign-in), patching together entire sentences is less than ideal, and should be avoided if possible.

How much code could we have in a string? Could we use a span, for example, and then replace it with a link? That would make things less confusing.

lesleyjanenorton: @flodolo We could, but I think that would introduce its own set of issues and edgecases. @betsymi Are you ok with only reducing the clickable/linked portion of this string to "Firefox Account" ?

betsymi: I'm OK with that, but account should be lowercased. We are moving away from capital A Account.

lesleyjanenorton: @betsymi thanks! In that case, do we think "account" should now be localized? And should this change be reflected everywhere we see "Firefox Account"?

flodolo: > @flodolo We could, but I think that would introduce its own set of issues and edgecases. @betsymi Are you ok with only reducing the clickable/linked portion of this string to "Firefox Account" ?

Can you clarify? What's I'm asking is to have something like

```
create-a-free-fxa = <a>Create a free {-brand-fxa}</a>, and you can add up to 15 email addresses.
```

That shouldn't affect how much of the sentence you linkify.

controllers/hibp.js
groovecoder: In general, if `recipient.fxa_uid` is null/empty, then they are a pre-FxA record. So, we could move this check up outside of these `if` conditionals ...

```
function getAddressesAndLanguageForEmail(recipient) {
  const signupLanguage = recipient.signup_language;
  const preFxaSubscriber = !recipient.fxa_uid;

File: controllers/hibp.js [hunk 3]
```diff
@@ -15,7 +15,9 @@ const log = mozlog("controllers.hibp");
 
 // Get addresses and language from either subscribers or
 // email_addresses fields
+
 function getAddressesAndLanguageForEmail(recipient) {
+  const preFxaSubscriber = recipient.fxa_uid ? true : false;
```
groovecoder: I added a commit to fix this logic - i.e., an `fxa_uid` means the user is *NOT* a pre-FXA subscriber.

groovecoder:
Note for future reference, to check a real breach alert email to a non-subscriber:
```
curl -d "{\"breachName\": \"Adobe\", \"hashPrefix\":\"f7d068\", \"hashSuffixes\":[\"69d0454f75028548bc2a51138d06f91610\"]}" -H "Content-Type: application/json" -H "Authorization: Bearer unsafe-default-token-for-dev" http://localhost:6060/hibp/notify
```
This will go to [nonfxasubscriber@mailinator.com](https://www.mailinator.com/v3/index.jsp?zone=public&query=nonfxasubscriber#/#inboxpane)

PR #1245, opened 2019-10-03
fix #609: add supported languages to __version__

### LINKED ISSUE #609: Expose supported locales in /__version__ endpoint?
Currently https://monitor.firefox.com/__version__ shows:

```js
{
  commit: "61d2a44d7ab55abd3325092276b8681143e71061",
  version: "v3.1.1",
  source: "https://github.com/mozilla/blurts-server",
  build: "https://circleci.com/gh/mozilla/blurts-server/2077"
}
```

But I have no idea which locales are supported on that build (or which locales were 80%+ translated at that exact moment when the build was built/deployed), since the logic is controlled by a `process.env.SUPPORTED_LOCALES` var.

Not sure if it'd be possible/helpful if we exposed the `SUPPORTED_LOCALES` env variable in the `/__version__` endpoint, or in some `<meta/>` tag or something.

https://github.com/mozilla/blurts-server/blob/f57f1974b4c54d94b19709980d23955d7c5da3b4/.env-dist#L58-L59

I vaguely recall l20n having a concept of `defaultLanguage` and `availableLanguages` which were exposed via `<meta/>` tags:

```html
<meta name="defaultLanguage" content="en-US">
<meta name="availableLanguages" content="de, en-US, fr, pl">
```

But at a minimum, adding it to `__version__` means I could hypothetically do one GET request to get the JSON payload, extract the available locales, then use that to scrape the site in the various supported locales (or whatever other silly stuff we want to do).

groovecoder:


pdehaan:
Verified in https://fx-breach-alerts.herokuapp.com/__version__

```json
{
  "commit": "c336e12eb0e2aaf1f4b57b875391de729394d516",
  "version": "c336e12eb0e2aaf1f4b57b875391de729394d516",
  "source": "https://github.com/mozilla/blurts-server",
  "languages": "cs,cy,da,de,en,en-CA,es-AR,es-CL,es-ES,es-MX,fr,fy-NL,hu,ia,id,it,ja,kab,ms,nl,pt-BR,pt-PT,ro,ru,sk,sl,sq,sv-SE,tr,uk,vi,zh-CN,zh-TW"
}
```

groovecoder:
I'm afraid it may only work on heroku and not stage nor prod. :( Because it looks at the env var, which is empty when we build the docker image that's deployed to stage + prod.

pdehaan:
Ah yes... I believe @Micheletto pulls the supported languages from `firefoxMonitorSupportedLocales` config in a cloudops-infra repo. Not sure if there would be a way to inject the currently supported languages when/if OPs is generating a static version.json file (or how they're detecting+injecting the Git commit SHA).

I also notice that we don't seem to have "kab" as a supported locale in the cloudops-infra repo, which means it probably isn't enabled on production yet.

PR #1267, opened 2019-10-10
Accepts emails on internationalized domain names

ywg-jean:
Since the service sends an email to verify the email address and the form already has input type `email`, using an out of date regex copy pasted from a stackoverflow answer which doesn't account for RFC822 and RFC2822 to verify email is not necessary.

### DIFF
--- a/public/js/monitor.js
+++ b/public/js/monitor.js
@@ -25,14 +25,6 @@ function toggleEl(e) {
   });
 }
 
-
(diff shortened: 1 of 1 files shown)


groovecoder:
Since the function is just returning `true` now, it's not necessary. Just remove it and its calling code completely.

ywg-jean:
@groovecoder I have amended my commit as per your recommandations.

PR #1270, opened 2019-10-11
Revert "Accepts emails on internationalized domain names"

groovecoder:
This reverts commit c74a67b53fc8ca74ab354591aab3613f305693ad.


groovecoder:
The email address is not just sent for scanning - it is also saved into session storage and sent to FxA for email first flow. So this change needs more thought before we can make it.

PR #1308, opened 2019-10-23
Update knex

groovecoder:
Supersedes https://github.com/mozilla/blurts-server/pull/1285 with a fix for using `Promise` in knex migrations after [the `Promise` argument was removed from knex migrations in 0.18.4](https://github.com/knex/knex/issues/3375)


pdehaan:
:+1: on **dompurify** and **jsdom** removal.
The only reference I could find via the mediocre GitHub search is https://github.com/mozilla/blurts-server/commit/c50e4b11f78ab796e5ddcc4cc6aef67a29e84bc8 where they were added to the repo.

As close as I can tell, we only used **dompurify** and **jsdom** for making breach `.Descriptions` site-safe, but it looks like we're currently deleting the descriptions (and other unused keys) from the breach object anyways, per:

https://github.com/mozilla/blurts-server/blob/cd5dae074cf97670222990c226bee207995430f9/template-helpers/breaches.js#L85-L96

groovecoder:
That migration is the only one that *uses* the `Promise` object. In the new knex version, `Promise` is not passed as an argument to migrations. But `Promise` is still available in the global context. So, removing the `Promise` parameter from the `exports.up|down` functions means that JS uses the global context `Promise` instead.

The other migrations don't *need* the parameter removed to work, so I didn't remove it to limit the code churn among migrations that really shouldn't ever be edited after-the-fact anyway. I only changed this one because Travis CI and other new environments with knex 0.19.5 fail when trying to run it. 😐

All new migrations in the repo will be generated by knex 0.19.5, so they will no longer have the `Promise` parameter.

PR #1310, opened 2019-10-23
revert to doc.pathname and doc.search for GA

groovecoder:
https://github.com/mozilla/blurts-server/pull/1275/files was an over-fix. Didn't need to use `document.location.href` for the GA `location` and `page` values. Only needed to move the `ga("set" ...)` calls out of the `loc.search` check.


groovecoder:
The expected audit failures.

Issue #1265, opened 2019-10-09
Monitor not sending utm parameters to GA


hoosteeno:
When you visit monitor.firefox.com with a set of arbitrary UTM parameters, those parameters are not sent to GA servers for the monitor property. This impacts our ability to measure impact of e.g. in-product or other upstream links. 

STR:

* Visit https://monitor.firefox.com/?utm_source=test&utm_campaign=test&utm_content=test&utm_medium=referral
a) Look for realtime traffic from that source -- it won't appear: https://analytics.google.com/analytics/web/#/realtime/rt-traffic/a77033033w177898028p180612539/filter.list=14==REFERRAL;7~~=test;&metric.type=0/
b) Use [Google's Analytics Debugger](https://chrome.google.com/webstore/detail/google-analytics-debugger/jnkmfdileelhofjcijamephohjechhna) to inspect the contents of the GA payload sent (see devtools console). It includes the host name, but not the parameters:

```
Running analytics_debug.js. This script is intended for testing and debugging only.
log @ analytics_debug.js:16
analytics_debug.js:16 Initializing Google Analytics.
analytics_debug.js:16 Running command: ga("create", "UA-77033033-16")
analytics_debug.js:16 Creating new tracker: t0
analytics_debug.js:16 Running command: ga("set", "anonymizeIp", true)
analytics_debug.js:16 Running command: ga("set", "transport", "beacon")
analytics_debug.js:16 Running command: ga("send", "pageview")
analytics_debug.js:16 Sending hit with transport method "beacon"
analytics_debug.js:16 
Sent beacon:
v=1&_v=j79d&aip=1&a=100790930&t=pageview&_s=1&dl=https%3A%2F%2Fmonitor.firefox.com%2F&ul=en-us&de=UTF-8&dt=Firefox%20Monitor&sd=24-bit&sr=1680x1050&vp=1506x851&je=0&_u=QACAAAAB~&jid=&gjid=&cid=138610038.1568737118&tid=UA-77033033-16&_gid=607632993.1570650957&z=1767132027


analytics_debug.js:16 _j1              (&jid)  
analytics_debug.js:16 _j2              (&gjid) 
analytics_debug.js:16 adSenseId        (&a)    100790930
analytics_debug.js:16 anonymizeIp      (&aip)  1
analytics_debug.js:16 apiVersion       (&v)    1
analytics_debug.js:16 clientId         (&cid)  138610038.1568737118
[...]

PR #1322, opened 2019-10-30
Analytics Fix

lesleyjanenorton:
- Send `document.location.href` (minus hash/token values) to GA instead of `document.location.pathname + document.location.search` (which sometimes evaluates to "undefined").

- Update regex to also capture "-" + any additional numerals or letters that follow. This will fix the strange "verify?-039802" URLs we are seeing in the GA dash.

### DIFF
--- a/public/js/analytics_tracking_protection.js
+++ b/public/js/analytics_tracking_protection.js
@@ -50,15 +50,15 @@ if(typeof(ga) !== "undefined") {
   ga("set", "anonymizeIp", true);
   ga("set", "transport", "beacon");
 
-
-  // Strip token and hash values from pings sent to GA
   const loc = document.location;
-  let pageValue = loc.pathname + loc.search;
+  let pageValue = loc.href;
+
+    // Replace token and hash values before sending URL to GA
   if (loc.search && (loc.search.includes("token=") || loc.search.includes("hash="))) {
-    pageValue = loc.href.replace(/token=\w+&?/, "").replace(/hash=\w+&?/, "");
+    pageValue = loc.href.replace(/\?token=[A-Za-z0-9_-]+/, "").replace(/&hash=[A-Za-z0-9_-]+/, "");
   }
+
   ga("set", "location", pageValue);
   ga("set", "page", pageValue);
-
 	ga("send", "pageview");
 }


public/js/analytics_tracking_protection.js
```diff
@@ -50,15 +50,15 @@ if(typeof(ga) !== "undefined") {
...
-
-  // Strip token and hash values from pings sent to GA
   const loc = document.location;
-  let pageValue = loc.pathname + loc.search;
+  let pageValue = loc.href;
+
+    // Replace token and hash values before sending URL to GA
   if (loc.search && (loc.search.includes("token=") || loc.search.includes("hash="))) {
-    pageValue = loc.href.replace(/token=\w+&?/, "").replace(/hash=\w+&?/, "");
+    pageValue = loc.href.replace(/\?token=[A-Za-z0-9_-]+/, "").replace(/&hash=[A-Za-z0-9_-]+/, "");
```
zerokarmaleft: If `href` is `undefined`, will this throw?

public/js/analytics_tracking_protection.js
```diff
@@ -50,15 +50,15 @@ if(typeof(ga) !== "undefined") {
...
   }
+
   ga("set", "location", pageValue);
```
lesleyjanenorton: @groovecoder not sure if this matters/relates to the current situation in the dashboard but it looks like before #1275 we were only explicitly setting the "page" and "location" values if `loc.search.includes("token=") || loc.search.includes("hash=")`. I'm wondering if we should move this back to inside the `if` block?

groovecoder: I was afraid that only setting the `page` and `location` values in the `if` block was causing #1265. 😢 Or at least, moving them outside the `if` block seemed to fix the bug described in those STR.

But if we can perform the same STR on this branch and verify we're not regressing the fix for #1265, then yeah - I'm happy to move them back how they were.

groovecoder: I performed a spot-check of the #1265 STR on this branch and I still see the `utm_*` params in the `location` and `page` values in the GA debugger. So this change does NOT regress #1265 . 👍

groovecoder:
I'd like to merge, tag, and deploy this today so we can start checking the live GA reports & dashboards ASAP.

PR #1323, opened 2019-10-30
No rated R breaches on the landing page / latest breach card.

### LINKED ISSUE #1321: Omit sensitive breaches from the "Latest Breach Added" section on the homepage
Using the boolean .IsSensitive flag via HIBP to prevent sensitive breach cards from appearing on our home page, even if a sensitive breach is technically the latest breach in the db.

lesleyjanenorton:
Fixes #1321

### DIFF
--- a/hibp.js
+++ b/hibp.js
@@ -154,6 +154,9 @@ const HIBP = {
     let latestBreach = {};
     let latestBreachDateTime = new Date(0);
     for (const breach of breaches) {
+      if (breach.IsSensitive) {
+        continue;
+      }
       const breachAddedDate = new Date(breach.AddedDate);
       if (breachAddedDate > latestBreachDateTime) {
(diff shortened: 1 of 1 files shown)


groovecoder:
Merged to make sure it's included in the next deploy. Can clean up logic or syntax in a follow-up if @pdehaan wants.

pdehaan:
Yeah, the logic is good, and it looks like we only refer to the method in 2 places (one in app, one in tests).

For future verifications, here the current Heroku homepage:

Firefox_Monitor

And I'll update this bit here after this deploy rides the commit train to party town and the Heroku instance is born anew.

I don't think we list recent breaches anywhere on the site (the /breaches page is random A.F.) but according to haveibeenpwned.com, the latest non-IsSensitive breach should be (drumroll) "StreetEasy accounts".

PR #1326, opened 2019-10-31
Analytics fixing 2.0

lesleyjanenorton:
- Moves ga script to `fxa-analytics.js` and removes `analytics_tracking_protection.js` and the browser detection code that is no longer in use.

- Waits for `hitCallback` before removing utm parameters so that these parameters can be sent to GA without having to set "page" and "location" explicitly, which I think was causing other issues.

From the Google Analytics Debugger in Chrome:
Screen Shot 2019-10-31 at 12 34 04 PM


groovecoder:
Dang; conflicts because I merged the previous branch first.

groovecoder:
Looks like a good approach. On this branch, there's not more "document" field sent?

image

@hoosteeno, @zerokarmaleft - does this look right to you?

Issue #1300, opened 2019-10-22
Global Communication Preferences: UI/UX Improvements


changecourse:
Options for global communication should be styled as radio buttons, not checkboxes. Additionally, there should be a visual affordance for a successful change between the two options, so let's fade in the same type of green status indicator as we use for adding a new email, when someone changes this. (Fade in, hold for 5 seconds, fade out) 

Zeplin: https://zpl.io/aMWEgKd

EMMLynch:
Closing since we've redesigned the site and functionality since this was created. If you feel that this is still needed, please let me know.

PR #1336, opened 2019-11-04
Restyle radio buttons on preferences dashboard

### LINKED ISSUE #1051: Checkboxes behave like radio buttons in Preferences
The preferences page has the following section:

image

Based on the visual design I expected these settings to behave like checkboxes. So I can have both or none checked, which would also make sense. However they appear to actually be radio buttons.

lesleyjanenorton:
- Adds a new "Changes saved!" string which will will be added to the template once localized to notify a user that they've successfully updated their communication settings.
- Restyles the radio buttons to look like radio buttons instead of checkboxes.
- Removes now unusued `checkbox-check.svg`.

Fixes #1051 and gets the ball rolling on #1300


public/css/forms.css
```diff
@@ -181,33 +181,48 @@ input[type="radio"] {
 
 .checkmark {
   background-color: rgba(255, 255, 255, 1);
-  height: 32px;
-  width: 32px;
-  border-radius: 4px;
-  display: block;
+  height: 24px;
+  width: 24px;
+  border-radius: 100%;
```
maxxcrawford: This may result in less ideal rounded corners on longer text strings. Would it possible to use more specific (in `px`) rounded corner radius?

Issue #1167, opened 2019-08-07
Website Styling Updates


changecourse:
Matching site to latest design spec: updated typography, type scales, padding, and slight tweaks/optimization of link styles and button styles throughout the site. 

## Acceptance criteria
- [x] Form field updates (email address & search) | 320px for all email address fields, placeholder text color (match to breaches search) | https://zpl.io/2yqgE7o
- [ ]  Typography Overhaul
   - consistent type scales (including letter and line spacing)
   - https://app.zeplin.io/project/5c93a4e0eb472f24e353125a/dashboard?seid=5d8b878865a7831666001e92
- [x] Button consistency across website
   - Large (check for breaches)
   - Medium (FxA callouts)
   - Small (Preferences)
- [x] Hover/interaction states for buttons
   - need to regroup and determine
- [x] Tweaks to breach cards
   - minor adjustments to align to spec (icon positioning, and type tweaks for margins and padding) 
   - https://zpl.io/aNmNnOD
- [x] Latest Breach Added' Card Styling
   - revise to match spec in Zeplin (sizing and padding)
   - https://zpl.io/2yqgE7o
- [x] Sign up for breach monitoring with a Firefox Account, on home page
   - matching to spec, button style, resizing/balancing tout images
   - https://zpl.io/2yqgE7o
- [x] Footer updates
   - change from light to dark, and update link styling
   - https://zpl.io/2yqgE7o
- [ ] Authenticated Dashboard - Updates
   - align to spec
   - https://zpl.io/bA7gJJn
- [ ] Monitor Preferences - Updates
   - align to spec
   - https://zpl.io/aMWEgKd
- [x] Breach Detail Updates
   - one column for compromised data, adjustments to padding between sections and subsections, adjustments to type scales of content
   - https://zpl.io/a7WP5m8
- [x] Breaches - Updates
   - decide on standard card size vs. truncation of compromised data, update title placement and padding to spec
   - https://zpl.io/bLdvnj7
- [x] Breaches - No Breaches Found
   - instead of tooltip copy, create placeholder text at base of field for 0 breaches found
   - https://zpl.io/V4oNN4X
- [x] Align breach promo landing pages to spec
   - https://zpl.io/aX3jO1g

## Visual design
InVision: https://mozilla.invisionapp.com/share/RGRIETZ7NAJ
Zeplin: https://zpl.io/VDBgQ7w

lesleyjanenorton:
- Typography overhaul is in progress
[...]

PR #1338, opened 2019-11-04
Update /breaches page.

lesleyjanenorton:
Restyles and moves the "Sorry, that breach is not in our database" message into the central part of the page.

Fixes part of #1167


pdehaan:
## Before

Firefox_Monitor

---

## After

Firefox_Monitor

PR #1339, opened 2019-11-04
 fix #1329: redirect user to FxA sign-in instead of error 

### LINKED ISSUE #1329: Evolve monitor.firefox.com/user/dashboard to be more user friendly
When logged into the Monitor site, the direct link to /dashboard looks great, but for those logged out, the first thing a user sees is "Error" (with login copy below - see screenshot)

Can we evolve the page to make it clearer to users that they just simply need to login so that any promotional communications that speak directly about the dashboard or targeted to known Monitor users (like via email) can be linked to this URL?

Screen Shot 2019-10-31 at 1 46 54 PM

groovecoder:


middleware.js
```diff
@@ -92,6 +93,25 @@ function errorHandler (err, req, res, next) {
...
+    // make sure the user object has all subscribers and email_addresses properties
+    return DB.getSubscriberById(req.session.user.id);
+  }
+  return null;
+}
+
+async function requireSessionUser(req, res, next) {
+  const user = await _getRequestSessionUser(req);
+  if (!user) {
+    return res.redirect("/oauth/init");
```
lesleyjanenorton: Just ran this locally and I was redirected, but when I added UTM parameters, they weren't passed along to FxA. I think we just need to check for any search params and append them to the /ouath/init route.

groovecoder: Updated to preserve utm params.

PR #1352, opened 2019-11-12
fix #816: refresh fxa profile data on page loads

### LINKED ISSUE #816: Refresh FxA profile data on page-loads

groovecoder:


controllers/oauth.js
```diff
@@ -66,14 +46,9 @@ async function confirmed(req, res, next, client = FxAOAuthClient) {
...
   log.debug("fxa-confirmed-fxaUser", fxaUser);
-  const data = await got(FxAOAuthUtils.profileUri,
-    {
-    headers: {
-      Authorization: `Bearer ${fxaUser.accessToken}`,
-    },
-  });
-  log.debug("fxa-confirmed-profile-data", data.body);
-  const email = JSON.parse(data.body).email;
+  const fxaProfileData = FXA.getProfileData(fxaUser.accessToken);
```
pdehaan: Should this have an `await`?
```js
async getProfileData(accessToken) {
```

lib/fxa.js
```diff
@@ -53,7 +72,20 @@ const FXA = {
...
   },
 
+  async getProfileData(accessToken) {
+    const data = await got(FxAOAuthUtils.profileUri,
+      {
+      headers: {
+        Authorization: `Bearer ${accessToken}`,
+      },
+    });
+    return data.body;
```
pdehaan: Should this return a string, or should we be returning a JSON blob?

lib/fxa.js
```diff
@@ -53,7 +72,20 @@ const FXA = {
     await this.destroyOAuthToken({ refresh_token: subscriber.fxa_refresh_token });
   },
 
+  async getProfileData(accessToken) {
+    const data = await got(FxAOAuthUtils.profileUri,
+      {
+      headers: {
```
pdehaan: Indenting here feels a bit "off".

db/DB.js
```diff
@@ -296,6 +296,15 @@ const DB = {
     return updatedSubscriber;
   },
 
+  async updateFxAProfileData(subscriber, fxaProfileData) {
+    await knex("subscribers").where("id", subscriber.id)
+    .update({
+      fxa_profile_json: fxaProfileData,
+    })
+    .returning("*");
+    return this.getSubscriberById(subscriber.id);
```
pdehaan: Q: On the previous line we use `.returning("*")`, but then ignore the [presumably] returned row/record. Now here we're doing _another_ query to get a subscriber by `subscriber.id`. Is that needed, or can we just return the result of `await knex().update()`?

http://knexjs.org/#Builder-returning

controllers/oauth.js
```diff
@@ -66,14 +46,9 @@ async function confirmed(req, res, next, client = FxAOAuthClient) {
...
+  log.debug("fxa-confirmed-profile-data", fxaProfileData);
+  const email = JSON.parse(fxaProfileData).email;
```
pdehaan: Looks like this line might be breaking the tests/CI:

```
confirmed request checks session cookie, calls FXA for token and email, adds new subscriber with signup language, and redirects
    SyntaxError: Unexpected token o in JSON at position 1
        at JSON.parse (<anonymous>)
      49 |   const fxaProfileData = FXA.getProfileData(fxaUser.accessToken);
      50 |   log.debug("fxa-confirmed-profile-data", fxaProfileData);
    > 51 |   const email = JSON.parse(fxaProfileData).email;
         |                      ^
      52 | 
      53 |   const existingUser = await DB.getSubscriberByEmail(email);
      54 |   req.session.user = existingUser;
      at parse (controllers/oauth.js:51:22)
      at Object.<anonymous> (tests/controllers/oauth.test.js:57:3)
```

lib/fxa.js
groovecoder: I'm returning the json just like the `confirmed` controller code did before. Trying to minimize the code changes in this PR, and also because the code puts this data in the `fxa_profile_json` column in the DB, which is `jsonb` type.

db/DB.js
groovecoder: We can probably drop the `returning("*")` call. But I'm calling `this.getSubscriberById` because it calls `joinEmailAddressesToSubscriber` which is required for the `req.user` and `req.session.user` objects.

groovecoder:
@pdehaan how does this look now?

lib/fxa.js
pdehaan: Hhmm, then I'm a bit confused by 

https://github.com/mozilla/blurts-server/pull/1352/files#diff-075ff2d4bc68f5a42b68daf27075ab5bR49-R51

```js
const fxaProfileData = await FXA.getProfileData(fxaUser.accessToken);
log.debug("fxa-confirmed-profile-data", fxaProfileData);
const email = JSON.parse(fxaProfileData).email;
```

If we're returning a JSON object, I thought that we wouldn't need the extra `JSON.parse()` bit there. But let me try pulling the PR locally and add some console loggin' and see why I'm confused. I thought if you needed to specify [`{json: true}`](https://www.npmjs.com/package/got#json) for **`got.get()`** if you wanted Got to auto `JSON.parse()` the response.

PR #1356, opened 2019-11-13
Add GA event to "Change your password" pathways on Breach Details page

### LINKED ISSUE #1301: Measure success of 'change password'
As we look to introduce this new 'call to action', we'll want to include event tracking.

We believe Monitor users want to be able to take action upon learning about a breach. We will know this to be true when 5% or more site visitors click to change their password.

Roughly 3-4% of all site visitors navigate to /security-tips, which is where we send any intent to take a 'next step' ("What to do after a website breach"): https://analytics.google.com/analytics/web/#/report/content-overview/a77033033w177898028p180612539/_u.date00=20190501&_u.date01=20191021 

We expect a more actionable 'change your password' option that is more visible on the breach detail to convert a higher number of visitors to take an action. But also we know the 'gold standard' for conversion on changing a password is ~26% (per [Google UR in this space](https://www.usenix.org/conference/usenixsecurity19/presentation/thomas), which is based on in-context breach notifications.

@irrationalagent I'd like your thoughts on how we should measure intent here as well. Let me know if there is a better way we might evaluate the success of this initial feature.

child of #1225

maxxcrawford:
Fixes #1301

This PR added a function that can add dynamic GA events on elements that have a class of `send-ga-ping`. Once clicked, it looks for three specific data-attributes on that element to fire a GA event.

### DIFF
--- a/public/js/fxa-analytics.js
+++ b/public/js/fxa-analytics.js
@@ -124,6 +124,16 @@ function getUTMNames() {
         removeUtmsFromUrl();
       },
     });
+
+    document.querySelectorAll("send-ga-ping").forEach((el) => {
+      el.addEventListener("click", async(e) => {
+        const eventCategory = e.target.dataset.eventCategory;
+        const eventAction = e.target.dataset.eventAction;
+        const eventLabel = e.target.dataset.eventLabel;
+        ga("send", "event", eventCategory, eventAction, eventLabel);
+      });
+    });
+
   } else {
     removeUtmsFromUrl();
   }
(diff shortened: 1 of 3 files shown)


public/js/fxa-analytics.js
```diff
@@ -124,6 +124,16 @@ function getUTMNames() {
         removeUtmsFromUrl();
       },
     });
+
+    document.querySelectorAll("send-ga-ping").forEach((el) => {
```
lesleyjanenorton: needs an ".". My fault.

PR #1359, opened 2019-11-14
fix #1191: ping fxa on email visits

### LINKED ISSUE #1191: Ping FxA metrics for visits to the Monitor website from a breach alert email
Monitor’s primary value to users is sending emails, and they may get all of the value they want out of their Monitor/FxA with that - may never come back to sign into the site at all.

We're not currently recording email notifications for FxA accounts. We should start recording this.

groovecoder:
Still need to add the actual ping the FXA, but here's the skeleton of the functionality.

@lesleyjanenorton - can you help me consider how we can get the `subscriberId` into ALL of the email links?

@pdehaan - is this crazy? 😄

### DIFF
--- a/controllers/hibp.js
+++ b/controllers/hibp.js
@@ -85,6 +85,9 @@ async function notify (req, res) {
 
   for (const recipient of recipients) {
     log.info("notify", {recipient});
+    // Get subscriber ID from "subscriber_id" property (if email_addresses record)
+    // or from "id" property (if subscribers record)
+    const subscriberId = recipient.subscriber_id || recipient.id;
     const { recipientEmail, breachedEmail, signupLanguage, preFxaSubscriber } = getAddressesAndLanguageForEmail(recipient);
     const campaignId = (preFxaSubscriber) ? "preFXA-see-all-breaches" : "see-all-breaches";
     const ctaHref = EmailUtils.getEmailCtaHref(utmID, campaignId);
@@ -104,9 +107,10 @@ async function notify (req, res) {
         {
           breachedEmail,
           recipientEmail,
+          subscriberId,
           supportedLocales,
           breachAlert,
-          SERVER_URL: req.app.locals.SERVER_URL,
+          SERVER_URL: AppConstants.SERVER_URL,
           unsubscribeUrl: EmailUtils.getUnsubscribeUrl(recipient, utmID),
           ctaHref: ctaHref,
           whichPartial: "email_partials/report",
--- a/controllers/hibp.js
+++ b/controllers/hibp.js
@@ -90,7 +90,7 @@ async function notify (req, res) {
     const subscriberId = recipient.subscriber_id || recipient.id;
     const { recipientEmail, breachedEmail, signupLanguage, preFxaSubscriber } = getAddressesAndLanguageForEmail(recipient);
     const campaignId = (preFxaSubscriber) ? "preFXA-see-all-breaches" : "see-all-breaches";
-    const ctaHref = EmailUtils.getEmailCtaHref(utmID, campaignId);
+    const ctaHref = EmailUtils.getEmailCtaHref(utmID, campaignId, subscriberId);
 
     const requestedLanguage = signupLanguage ? acceptedLanguages(signupLanguage) : "";
     const supportedLocales = negotiateLanguages(
--- a/email-utils.js
+++ b/email-utils.js
@@ -81,7 +81,7 @@ const EmailUtils = {
 
   appendUtmParams(url, campaign, content) {
     const utmParameters = {
-      utm_source: "fx-monitor-email",
+      utm_source: "fx-monitor",
       utm_medium: "email",
       utm_campaign: campaign,
       utm_content: content,
--- a/email-utils.js
+++ b/email-utils.js
@@ -99,8 +99,9 @@ const EmailUtils = {
     return req.fluentFormat("email-subject-found-breaches");
   },
 
-  getEmailCtaHref(emailType, campaign) {
-    const url = new URL(AppConstants.SERVER_URL);
+  getEmailCtaHref(emailType, campaign, subscriberId=null) {
+    const subscriberParamPath = (subscriberId) ? `/?subscriber_id=${subscriberId}` : "/";
+    const url = new URL(subscriberParamPath, AppConstants.SERVER_URL);
     return this.appendUtmParams(url, campaign, emailType);
   },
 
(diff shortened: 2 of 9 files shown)


controllers/hibp.js
```diff
@@ -104,9 +107,10 @@ async function notify (req, res) {
         {
           breachedEmail,
           recipientEmail,
+          subscriberId,
           supportedLocales,
           breachAlert,
-          SERVER_URL: req.app.locals.SERVER_URL,
+          SERVER_URL: AppConstants.SERVER_URL,
```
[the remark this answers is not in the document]
groovecoder: Can you file a follow-up issue to clean up our usage of `req.app.locals` to replace with `AppConstants` wherever we can?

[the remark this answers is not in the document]
pdehaan: Done. #1390

middleware.js
```diff
@@ -52,6 +52,20 @@ function pickLanguage (req, res, next) {
...
 
+async function recordVisitFromEmail (req, res, next) {
+  if (req.query.utm_source && req.query.utm_source !== "fx-monitor-email") {
+    next();
+  }
+  if (!req.query.subscriber_id || !Number.isInteger(req.query.subscriber_id)) {
+    next();
+  }
+  const subscriber = await DB.getSubscriberById(req.query.subscriber_id);
+  const fxaMetricsFlowPath = `metrics-flow?event_type=engage&uid=${subscriber.fxa_uid}&service=${AppConstants.OAUTH_CLIENT_ID}`;
```
pdehaan: What are the chances there that `DB.getSubscriberById()` returns nothing, `subscriber` is null/undefined, and then `${subscriber.fxa_uid}` throws up?

template-helpers/emails.js
```diff
@@ -35,8 +35,8 @@ function emailBreachStats(args) {
 function getPreFxaUtmParams(serverUrl, content, userEmail) {
   const url = new URL(`${serverUrl}/oauth/init`);
   const utmParams = {
-    utm_source: "fx-monitor",
-    utm_medium: "fx-monitor-email",
+    utm_source: "fx-monitor-email",
+    utm_medium: "email",
```
pdehaan: These seem like important changes, will this break any backwards compat with older things being logged?

email-utils.js
```diff
@@ -99,8 +99,9 @@ const EmailUtils = {
     return req.fluentFormat("email-subject-found-breaches");
   },
 
-  getEmailCtaHref(emailType, campaign) {
-    const url = new URL(AppConstants.SERVER_URL);
+  getEmailCtaHref(emailType, campaign, subscriberId=null) {
+    const subscriberParamPath = (subscriberId) ? `/?subscriber_id=${subscriberId}` : "/";
```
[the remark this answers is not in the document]
groovecoder: It's the integer `id` column from either the `subscribers` or `email_addresses` table.

middleware.js
groovecoder: High enough to check for the property first!

template-helpers/emails.js
groovecoder: I'll check with @lesleyjanenorton about it. AFAIK no-one but ourselves is looking into our own email traffic sources.

middleware.js
groovecoder: Updated.

lesleyjanenorton: Does this need an `entrypoint` value?

template-helpers/emails.js
lesleyjanenorton: These won't get logged in our analytics since they're appended to the `/oauth/init` route which redirects directly to to FxA.  I think `utm_source` should remain `fx-monitor` to be consistent with other FxA pings and then perhaps we add `entrypoint: "pre-fxa-email"`.

middleware.js
```diff
@@ -52,6 +52,20 @@ function pickLanguage (req, res, next) {
 }
 
 
+async function recordVisitFromEmail (req, res, next) {
+  if (req.query.utm_source && req.query.utm_source !== "fx-monitor-email") {
```
lesleyjanenorton: Depending on how we want FxA to categorize these events, we may want to just explicitly set `utm_source` here to `fx-monitor`.

template-helpers/emails.js
groovecoder: Oh dang, this is much ado about nothing now since we're already running the pre-FXA email blast.

So, I'm going to leave this as-is/was in `master`, so later when we look at GA reports and see the traffic from `utm_medium` of `fx-monitor-email` we'll see this is where it came from.

middleware.js
groovecoder: I'll update to be `utm_source=fx-monitor` but that will affect all our other uses of `appendUtmParams`. So we're basically converting all our email traffic from:
* `utm_source=fx-monitor-email&utm_medium=email`
to: 
* `utm_source=fx-monitor&utm_medium=email`

Which I think makes sense. We just may want to make a note in our GA reports about this change-over.

groovecoder:
Updated to consolidate and clean up `utm_`* params to what I think are reasonable and consistent values.

views/partials/email_partials/email_single_breach.hbs
```diff
@@ -17,7 +17,7 @@
...
           <span class="text-medium" style="font-weight:600; font-size: 15px; font-family: sans-serif; margin: 0px; color: #20123a;">{{ DataClasses }}</span>
         {{/if}}
-        <p class="more-info-link-wrap" style="padding-top: 20px; margin: 0;"><a class="more-info-link" target="_blank" rel="noopener noreferrer" style="color: #0060df; font-family: sans-serif; font-weight: 500; text-decoration: none;" href="https://monitor.firefox.com/breach-details/{{ Name }}?utm_source=fx-monitor-email&utm_medium=email&utm_campaign=more-about-this-breach-link">{{ LocalizedBreachCardStrings.MoreInfoLink }}</a></p>
+        <p class="more-info-link-wrap" style="padding-top: 20px; margin: 0;"><a class="more-info-link" target="_blank" rel="noopener noreferrer" style="color: #0060df; font-family: sans-serif; font-weight: 500; text-decoration: none;" href="{{ SERVER_URL }}/breach-details/{{ Name }}?utm_source=fx-monitor-email&utm_medium=email&utm_campaign=more-about-this-breach-link&subscriber_id={{ subscriberId }}">{{ LocalizedBreachCardStrings.MoreInfoLink }}</a></p>
```
lesleyjanenorton: This `fx-monitor-email` in particular 🌪

Issue #1351, opened 2019-11-11
Optimize /scan Results for Marketing Campaign


changecourse:
Optimizing the page for account conversion, which means de-prioritizing the current primary callout for secondary email scan, and instead promoting FxA for future breach alerts. 

Design: 
![Scan Results, Multiple Found - Desktop](https://user-images.githubusercontent.com/6164801/69271365-efd37e80-0b91-11ea-9afd-11e9b8a5d739.png)

SVG Illustration (159x154px):
[firefox-monitor_stay-alert.svg.zip](https://github.com/mozilla/blurts-server/files/3871066/firefox-monitor_stay-alert.svg.zip)

Zeplin: https://zpl.io/aNmNnOD

cherrypsf:
I've heard you can do mini-A/B split tests on buttons by sprinkling some javascript. Engineering would have to validate this rumor of mine, but if we did that here, we would be set up to try split testing CTAs in the future.

groovecoder:
Note: Google Optimize requires `unsafe-inline` in our CSP, so we can't use that one. 😢 Will keep looking into implementation options for the A/B test.

groovecoder:
I think we can do the A/B split like so:

* Create a new [custom dimension in our GA view](https://analytics.google.com/analytics/web/#/a77033033w177898028p180612539/admin/custom-dimensions/)
* In `scan` controller, `const newScanPage = (Math.random() >= 0.5) ? true : false`
* In `scan` template, `{% if newScanPage %}`:
  * Display the new scan page
  * `ga('set', 'cd5', 'newScanPage');`

That should be enough for us to show [the signup button conversion report](https://analytics.google.com/analytics/web/#/report/conversions-goals-overview/a77033033w177898028p180612539/overview_goaloption_ALL-graphOptions.primaryConcept=analytics.goalConversionRate&_.goalOption=4&overview_goaloption_4-graphOptions.primaryConcept=analytics.goalConversionRate4) with both segments side-by-side.

lesleyjanenorton:
> * In `scan` controller, `const newScanPage = (Math.random() >= 0.5) ? true : false`
> * In `scan` template, `{% if newScanPage %}`:
>   * Display the new scan page

-- Done in [this branch](https://github.com/mozilla/blurts-server/tree/new-marketing-scan-page)

<br />

>   * `ga('set', 'dimension5', 'newScanPage');`

-- In Progress

lesleyjanenorton:
[...]

PR #1360, opened 2019-11-14
Fix / add / update ga pings

lesleyjanenorton:


public/js/all-breaches/all-breaches.js
```diff
@@ -132,6 +136,7 @@ function initBreaches() {
     const [fuzzyShowAll, showHiddenBreaches] = document.querySelectorAll(".show-all-breaches");
 
     showHiddenBreaches.addEventListener("click", (e) => {
+      sendPing(e.target, "Click", "All Breaches Page");
```
groovecoder: Could we achieve the same effect (i.e., to send a GA ping) by adding `send-ga-ping` class and data attributes?

groovecoder:
We might do this in a follow-up when we optimize the scan results page (#1351 ), but there's no ping when someone clicks "Alert me about new breaches" button on scan results?

image

groovecoder:
Oops, that was because I had configured my firefox to disable the sendBeacon API. Looks good!

PR #1376, opened 2019-11-18
Add new scan page. 

### LINKED ISSUE #1351: Optimize /scan Results for Marketing Campaign

lesleyjanenorton:
- adds a new version of the scan results page.
- sets a `newScanPage` custom dimension to all Google Analytics events occurring on the new `/scan` so that we can compare them to the GA events occurring on the existing `/scan`.\

Fixes #1351


scan-results.js
```diff
@@ -86,6 +86,8 @@ const scanResult = async(req, selfScan=false) => {
     }
   }
 
+  const newScanPage = ((Math.random() * (10 - 1) + 1) >= 5);
```
groovecoder: Do we need the `-1` and `+1`? Why not just `Math.random() * 10 >= 5`

pdehaan: Does this mean I could get a different UI on each page reload?

lesleyjanenorton: @pdehaan Yes - you could very feasibly get a different UI on reload or subsequent scans.

pdehaan: I don't know if it matters or not (it's above my pay grade).
I do recall that the FxA team had built a little A/B testing thing in the past, but not sure if it's still used or if they moved onto something else: https://github.com/dannycoates/able

It might depend on how many A/B tests we plan on doing in the future and if keeping people "in" the experiment matters or not.

groovecoder:
`push` fail is `npm audit` fail which is already fixed in `master`

PR #1381, opened 2019-11-19
drop email_addresses subquery

groovecoder:


db/DB.js
```diff
@@ -80,13 +80,8 @@ const DB = {
...
   async getPreFxaSubscribers() {
-    // We don't want to get subscribers records where the email address has
-    // since been added "under" an FxA subscription
-    const allEmailAddressesSubquery = knex("email_addresses").select("email");
     const preFxaSubscribers = await knex("subscribers").whereRaw(
       "(fxa_uid = '') IS NOT FALSE"
-    ).andWhere(
-      "primary_email", "not in", allEmailAddressesSubquery
     );
     return preFxaSubscribers;
```
pdehaan: This feels weird.
Now that we're not using a subquery, should this whole function just be something like this &mdash; instead of putting the result in a temporary variable:

```js
async getPreFxaSubscribers() {
  return knex("subscribers").whereRaw("(fxa_uid = '') IS NOT FALSE");
}

groovecoder: Updated.

groovecoder:
Note: Travis will fail on the `audit` task. I have a separate fix incoming for that.

PR #1383, opened 2019-11-20
use knex-paginate to paginate pre-FXA query

groovecoder:
Run like so:

```
PAGE_SIZE=5 node scripts/send-email-to-pre-fxa-subscribers.js
```

### DIFF
--- a/scripts/send-email-to-pre-fxa-subscribers.js
+++ b/scripts/send-email-to-pre-fxa-subscribers.js
@@ -1,5 +1,7 @@
 "use strict";
 
+/* eslint-disable no-process-env */
+
 const { negotiateLanguages, acceptedLanguages } = require("fluent-langneg");
 
 const AppConstants = require("../app-constants");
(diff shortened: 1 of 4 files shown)


scripts/send-email-to-pre-fxa-subscribers.js
```diff
@@ -8,42 +10,52 @@ const EmailHelpers = require("../template-helpers/emails.js");
...
+
+  const subscribersResult = await DB.getPreFxaSubscribersPage({ perPage: PAGE_SIZE, currentPage: 1 });
+  console.log(`Found ${subscribersResult.pagination.total} subscriber records with empty fxa_uid.`);
+  console.log(`Will process ${subscribersResult.pagination.lastPage} pages of size ${PAGE_SIZE}.`);
+  const lastPage = subscribersResult.pagination.lastPage;
+
+
+  for (let currentPage = 1; currentPage <= lastPage; currentPage++) {
+    console.log(`Processing page ${currentPage} of ${lastPage}.`);
+    const subscribersPageResult = await DB.getPreFxaSubscribersPage({ perPage: PAGE_SIZE, currentPage });
```
pdehaan: This line looks the same as the one above on L22 (both fetching `currentPage=1`). But the first usage seems to just get stats, so I don't think we're processing anything twice or whatever. I think.

scripts/send-email-to-pre-fxa-subscribers.js
```diff
@@ -8,42 +10,52 @@ const EmailHelpers = require("../template-helpers/emails.js");
 const EmailUtils = require("../email-utils");
 const { LocaleUtils } = require ("../locale-utils");
 
+const PAGE_SIZE = process.env.PAGE_SIZE;
```
pdehaan: Should this have a default (ie: `process.env.PAGE_SIZE || 100` -- or whatever @Micheletto thinks is a sane default)?
I can't imagine OPs would try running this without that env var set, but I can think this might fail spectacularly below if it tries to use an `undefined` `perPage` page size.

groovecoder: Oh dang, good point. Running without a `PAGE_SIZE` will try to run without any pagination. So it'll just run/fail the same as it does now. But I'll make the point on the bug.

scripts/send-email-to-pre-fxa-subscribers.js
groovecoder: Yeah first usage is to get the `lastPage` number to put into the `for` loop. But trying to write the `for` loop to accommodate the first-page-is-already-queried case made the logic more complicated.

PR #1392, opened 2019-11-22
Update Bento Lockwise link.

lesleyjanenorton:


groovecoder:
Confirmed:

* [x] Clicking on desktop Mac goes to https://www.mozilla.org/en-US/firefox/lockwise/
* [x] Clicking on iOS without app installed goes to app store
* [x] Clicking on Android with app installed opens app
* [x] Clicking on Android without app goes to app store

PR #1398, opened 2019-11-23
add utm_ params to fxa_rp - engage ping

groovecoder:


middleware.js
```diff
@@ -70,7 +70,10 @@ async function recordVisitFromEmail (req, res, next) {
     next();
     return;
   }
-  const fxaMetricsFlowPath = `metrics-flow?entrypoint=breach-alert-email&event_type=engage&uid=${subscriber.fxa_uid}&service=${AppConstants.OAUTH_CLIENT_ID}`;
+  const breachDetailsRE = /breach-details\/(\w*)$/;
+  const capturedMatch = req.path.match(breachDetailsRE);
+  const utmContent = (capturedMatch) ? `&utm_content=${capturedMatch[1]}` : "";
+  const fxaMetricsFlowPath = `metrics-flow?utm_source=${req.query.utm_source}&utm_medium=${req.query.utm_medium}${utmContent}&event_type=engage&uid=${subscriber.fxa_uid}&service=${AppConstants.OAUTH_CLIENT_ID}`;
```
lesleyjanenorton: probably doesn't make a huge diff either way but do we want to hardcode `utm_source` and `utm_medium` here since they should always be `fx-monitor` and `email`?

groovecoder: I usually try to reduce hard-coding. We might add other re-engagements in the future, like a push notification or something.

groovecoder:
Didn't want to merge this while we were hustling to land the feature, but this is good-to-merge now.

PR #1399, opened 2019-11-23
Mention that passwords are being cracked automatically

devurandom:
"figure out" sounds as if someone sits down and thinks about your password, which might seem less dangerous. Mention that it is an automated process.

References: #1400


groovecoder:
Flagging this for review from @betsymi for content strategy. We also need to consider localization anytime we update strings. So, it's unlikely this PR will merge as-is. But we'll take this under advisement.

groovecoder:
Closing this in favor of https://github.com/mozilla/blurts-server/issues/1407 which will include more comprehensive content updates and changes.

PR #1424, opened 2019-12-02
fix #887: un-pin node image to pick up bug-fixes

### LINKED ISSUE #887: Update our docker alpine
See https://bugzilla.mozilla.org/show_bug.cgi?id=1550155

groovecoder:


Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM node:10.16.3-alpine
+FROM node:10-alpine
```
pdehaan: LGTM, per @jbuck's Slack comment:

> **Jon Buckley 11:47 AM**
> @groovecoder I’d recommend switching to `node:10-alpine` so you’re constantly picking up bug fixes (unless you’re really good about following the patch releases)

groovecoder:
Updated to include `chown` during the `COPY` command, as @jbuck suggested.

PR #1425, opened 2019-12-02
fix #1411: add breaches_resolved JSON to subscribers table

### LINKED ISSUE #1411: Add a breaches_resolved JSON column to subscribers table
Add a `breaches_resolved` JSON column to `subscribers` table to store data like:
```
[
  {"email_address": <primary_email_address>, "recency_indices":[0,2]},
  {"email_address": <another_email_address>, "recency_indices":[1]},
]
```

groovecoder:


groovecoder:
I didn't add an index here, because it's a `jsonb` column, and I don't see any requirements to query by values in the JSON. So, we'll include the full JSON value with each query and do any querying/logic at the code level.

@Micheletto the data structure in the column will be like so ...
```
[
  {"email_address": <primary_email_address>, "recency_indices":[0,2]},
  {"email_address": <another_email_address>, "recency_indices":[1]},
]
```

Do you see any problem with adding this much storage (in the DB) or this much bandwidth (from the DB to web containers)? Should we query for some metrics?

* Average number of email addresses per user
* Average number of breaches per email address (a recent data test showed 75% of email addresses in HIBP have between 2 and 3 breaches)

Multiply those together and by the JSON blob size to get the expected record column size?

groovecoder:
Merging so I can deploy and run this migration on:

https://fx-breach-alerts.herokuapp.com/
https://blurts-server-review.herokuapp.com/
https://monitor-v2.herokuapp.com/

PR #1429, opened 2019-12-06
Add /breach-details recommendations

lesleyjanenorton:
This updates `/breach-details` pages and replaces the current "What to do" tips with recommendations specific to the breach's exposed data classes.

**Recommendation Rules & Regulations:**
- A minimum of four recommendations should be shown for each breach.
- If passwords were exposed, show the first three password recommendations followed by any other specific recommendations, and push the fourth password recommendation to the end of the list.
- If fewer than four recommendations are available, the list should be back-filled with generic recommendations until there are four. 
- If there are more than four recommendations, the first four should be visible on page load with the remaining number becoming visible on "See Additional Recommendations" click.
- Some recommendations should show up differently depending on the user's browser and locale.

**New Analytics Events:**
- A ping is sent each time the "See Additional Recommendations" button is clicked.
```
Ping Structure 
  {
    eventCategory: See Additional Recommendations,
    eventAction: Click,
    eventLabel: Name of Breach,
  }
```
- A ping is sent each time one of the recommendation CTAs is clicked.
```
Ping Structure
  {
    eventCategory: Breach Detail: Recommendation CTA Click,
    eventAction: Click,
    eventLabel: Name of Breach,
  }
```

-Adds recommendations.ftl
-Adds template-helpers/recommendations.js
-Adds recommendations-icons/*.svg

TO DO: Get Monitor whitelisted to open `about:logins` so that we can do something like what happens when you click "Open in Firefox" here https://www.mozilla.org/firefox/lockwise/.


locales/en/recommendations.ftl
```diff
@@ -0,0 +1,140 @@
...
+  Look for accounts, loans, or credit cards you don’t recognize.
+
+# Recommendation subhead
+rec-pw-primary-subhead = Change your password
+# Link title
+rec-pw-primary-cta = Change password for this site
+rec-pw-primary = 
+  Make this password unique and different from any others you use. 
+  A good strategy to follow is to combine two or more unrelated 
+  words to create an entire pass phrase.
```
flodolo: Isn't *passphrase* a single word normally?

locales/en/recommendations.ftl
```diff
@@ -0,0 +1,140 @@
...
+
+# Recommendation subhead
+rec-pw-secondary-subhead = Update other logins using the same password
+# Link title
+rec-pw-secondary-cta-fx = View logins in {-brand-name}
```
flodolo: nit: can we keep empty spaces in placeables, i.e. `{ -brand-name }`?

locales/en/recommendations.ftl
```diff
@@ -0,0 +1,140 @@
...
+rec-gen-2 =
+  Put your login details in a secure place only you can access, such as a 
+  password manager. This also makes it easy to keep track of all your different passwords.
+
+# Recommendation subhead
+rec-gen-3-subhead = Be cautious about giving out personal information
+# Link title
+rec-gen-3-cta = Read more security tips
+rec-gen-3 =
+  Don’t hand out personal data if you don't have to. If you are asked to
```
flodolo: Wrong apostrophe in the second *don't*

template-helpers/recommendations.js
```diff
@@ -0,0 +1,266 @@
...
+            recommendationCopy: {
+              subhead: "rec-ssn-cta",
+              cta: "rec-ssn-cta",
+              body: "rec-ssn",
+            },
+            ctaHref: "https://www.annualcreditreport.com/index.action",
+            ctaShouldOpenNewTab: true,
+            recIconClassName: "rec-ssn",
+          },
+        ] : "",
```
pdehaan: This feels a bit funny. This will either return an array of object, or an empty string?

views/partials/breach-detail-content-group.hbs
```diff
@@ -1,14 +1,29 @@
-              <h2 class="section-headline txt-purple7">{{ this.headline }}</h2>
-
+              
+            <div class="mw-700 breach-detail {{ sectionId }}">
+              {{#if smallerSectionHeadline}}
+                <h3 class="small-section-headline txt-purple7 bold">{{ this.headline }}</h3>
+              {{else}}
+                <h2 class="section-headline txt-purple7">{{ this.headline }}</h2>
+              {{/if}}
```
pdehaan: This feels wrong. I don't think you're supposed to jump headings for a11y reasons.
Not sure if we should always just use an `<h2>` here and just change the classes to make the UI different, vs toggling between an H2 or H3 (unless we know they'll always be in a logical H1=>H2=>H3 order, in which case feel free to ignore me).

views/partials/svg/glyphs/ip-addresses.hbs
```diff
@@ -1,4 +1,4 @@
...
...7989726,7.5323923 22.1669056,7.7724646 21.5161693,7.74991171 L17.7193745,13.7611735 C18.1968363,14.6503574 18.0558172,15.7408496 17.3674748,16.4823927 C16.6791324,17.2239357 15.5930899,17.4553424 14.6573594,17.059846 C13.7216289,16.6643495 13.1396075,15.727918 13.2049807,14.7230704 L9.46458587,12.2553892 C9.10841452,12.4306037 8.71369143,12.5150954 8.31618746,12.5012073 L4.51939269,18.510094 C5.02926571,19.4592775 4.83134628,20.6290243 4.03684211,21.3620726 C3.24233794,22.0951209 2.04935806,22.2086857 1.12821853,21.6389567 C0.207079006,21.0692278 -0.212534717,19.9582666 0.105151518,18.9303042 C0.422837752,17.9023419 1.39860912,17.2137 2.4841955,17.2513154 L6.28099026,11.2400536 C5.8035285,10.3508697 5.9445476,9.26037748 6.63288997,8.51883443 C7.32123233,7.77729138 8.40727482,7.5458847 9.34300535,7.94138114 C10.2787359,8.33687758 10.8607572,9.27330914 10.795384,10.2781567 L14.5357789,12.7458379 C14.8920573,12.5710335 15.2867751,12.4869512 15.6841773,12.5012073 L19.4809721,6.48994554 C19.0121593,5.61716504 19.1385753,4.54828748 19.7983749,3.8062277 C20.4581745,3.06416792 21.5134111,2.80407098 22.4478433,3.15318117 C23.3822755,3.50229135 24.0005382,4.38761896 24.0001658,5.3760452 Z"/>
```
pdehaan: Hard to see from diffs. I blindly (and foolishly) assume the only thing that we changed here was the `fill=""` color (here and all the other glyphs/*.hbs files).
I wonder if it's worth seeing if we can modify the SVG fill colors via CSS to avoid big diffs in the future (if we ever need to change the fill colors), or if that gets infinitely tricky because these are inlined SVGs via HBS.

locales/en/recommendations.ftl
```diff
@@ -0,0 +1,140 @@
...
+rec-ip-subhead = Use a service that masks your IP address
+# Link title
+rec-ip-us-cta = Try {-brand-fpn}
+rec-ip-us =
+  Your Internet Protocol address (IP address) pinpoints your location 
+  and internet service provider. A service like {-brand-fpn} 
+  masks your IP address to hide your location.
+rec-ip-non-us = 
+  Your Internet Protocol address (IP address) pinpoints your location 
+  and internet service provider. With a VPN, you can hide your location
```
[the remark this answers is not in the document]
lesleyjanenorton: I think it would be helpful to define it here, even if it's defined elsewhere. @betsymi What are your thoughts here?

locales/en/recommendations.ftl
```diff
@@ -0,0 +1,140 @@
...
+  Use {-brand-lockwise} to securely access the passwords you’ve saved in 
+  {-brand-name} from anywhere — even outside of the browser.
+rec-pw-tertiary-non-fx = 
+  Use {-brand-lockwise} to keep track of all your different 
+  passwords and access them securely from your phone or tablet.
+
+# Recommendation subhead
+rec-pw-quaternary-subhead = Set up two-factor authentication
+# Link title
+rec-pw-quaternary-cta = See sites that offer 2FA
```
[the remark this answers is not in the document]
lesleyjanenorton: @betsymi

public/css/breach-detail.css
```diff
@@ -104,20 +98,135 @@ main.breach-detail {
...
+
+.rec-pw-secondary {
+  background-image: url("/img/recommendation-icons/change-password.svg");
+}
+
+.rec-pw-tertiary {
+  background-image: url("/img/recommendation-icons/store-safe-place.svg");
+}
+
+.rec-pw-quaternary {
```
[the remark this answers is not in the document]
lesleyjanenorton: "Quaternary" was a new vocab word for me as well. No reason other than I was just matching the naming that was set forth in the recommendation copy deck. I'll update to numbers!

template-helpers/breach-detail.js
```diff
@@ -1,110 +1,64 @@
...
+  return { locales, breach, changePWLink, isUserBrowserFirefox };
 }
 
-// REMAINING QUESTIONS:
-// How to actually categorize these?
-// How to dynamically detect Data Aggregator breaches?
 
 function getBreachCategory(breach) {
-  if (["Exactis", "Apollo", "YouveBeenScraped", "ElasticsearchSalesLeads", "Estonia", "MasterDeeds", "PDL"].includes(breach.Name)) {
+  if (["Exactis", "Apollo", "YouveBeenScraped", "ElasticsearchSalesLeads", "Estonia", "MasterDeeds"].includes(breach.Name)) {
```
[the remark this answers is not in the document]
lesleyjanenorton: I goofed. It should still be there.

public/css/breach-detail.css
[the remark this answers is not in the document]
flodolo: Welcome to Latin 101 :P (it's easy up to 4, from 5 it becomes a lot less logical)

lesleyjanenorton:
> What would be the timeline to start showing this? That's a lot of content to translate.

@flodolo As far as I know there isn't a specific release date for these yet. What do you think is a realistic amount of time for these to be localized?

flodolo:
> As far as I know there isn't a specific release date for these yet. What do you think is a realistic amount of time for these to be localized?

2-3 weeks sounds safe, although we're getting into the Christmas vacation period. If copy is set, it could also be a good idea to preland the strings to gain time (if needed).

public/css/breach-detail.css
```diff
@@ -104,20 +98,135 @@ main.breach-detail {
...
+  min-width: 52px;
+  height: 48px;
+  margin: 0 auto;
+  background-position: top center;
+  background-size: contain;
+  background-repeat: no-repeat;
+}
+
+.rec-pw-1 {
+  background-image: url("/img/recommendation-icons/change-password.svg");
```
pdehaan: **TL;DR:** All the image paths look 👍. 

I was compelled to write some crude linter to verify all these CSS file paths.

```js
"use strict";

const exec = require("child_process").exec;
const fs = require("fs").promises;
const path = require("path");
const { promisify } = require("util");

const execp = promisify(exec);
const cmd = `git grep -Eno '/img/recommendation-icons/.*?\\.svg'`;

main(cmd);

async function main(cmd) {
  const { stdout, stderr } = await execp(cmd);
  if (stderr) {
    console.error(stderr);
    process.exit(1);
  }

  const css = stdout
    .toString()
    .split("\n")
    .filter(Boolean)
    .map(line => line.trim());

  for (const row of css) {
    const regexp = /^(?<fileinfo>.*):(?<src>.*?)$/;
    const { fileinfo, src } = regexp.exec(row).groups;
    const svgPath = path.join("./public", src);
    try {
      await fs.stat(svgPath);
    } catch (err) {
      console.error(`[${fileinfo}] Cannot find '${svgPath}'`);
      process.exitCode = 2;
    }
  }
}
```

locales/en/recommendations.ftl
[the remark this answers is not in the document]
betsymi: Sure. I'm good with **Set up two-factor authentication (2FA)**

locales/en/recommendations.ftl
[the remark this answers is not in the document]
betsymi: Your Internet Protocol address (IP address) pinpoints your location and internet service provider. With a Virtual Private Network (VPN), you can hide your location and mask your IP address.

The glossary is a good idea, but not currently on the table.

public/css/breach-detail.css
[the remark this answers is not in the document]
betsymi: Numbers are so much better. Thank you. I don't have a good reason why we started using this jargon. Good thing we are not surfacing it to users!

template-helpers/recommendations.js
```diff
@@ -0,0 +1,266 @@
...
+        weight: 101,
+        pathToGlyph: "svg/glyphs/social-security-numbers",
+        recommendations: isUserLocaleEnUs ? [
+          {
+            recommendationCopy: {
+              subhead: "rec-ssn-cta",
+              cta: "rec-ssn-cta",
+              body: "rec-ssn",
+            },
+            ctaHref: "https://www.annualcreditreport.com/index.action",
```
groovecoder: I'm very unimpressed with this site's security score. 😢 Can we remove the CTA/link from this one? (@betsymi )

betsymi: Sure, we can remove the CTA.

groovecoder: Although it's probably better this than to send users on a Google search for credit reports. So, let's keep this as-is.

groovecoder:
Can we add GA pings for views of the recommendations? That way we can measure the conversion rate of people clicking/seeing recommendations. In this PR or in a follow-up.

public/js/fxa-analytics.js
```diff
@@ -124,6 +124,15 @@ function getUTMNames() {
       },
     });
 
+    // Temporarily send "View" and "Engage" pings for recommendation CTAs.
+    document.querySelectorAll(".recommendation-cta").forEach(cta => {
+      const eventLabel = cta.dataset.eventLabel;
+      ga("send", "event", "Breach Detail: Recommendation CTA", "View", eventLabel);
```
groovecoder: Non-blocking fix-up: This fired even though the element wasn't showing yet. I.e., it was behind "See additional recommendations"

Issue #1287, opened 2019-10-21
Hover on breach card unstable


sandysage:
![monitor-card-zooming](https://user-images.githubusercontent.com/850418/67232205-1caf3d00-f3fe-11e9-9a72-abd74d089ced.gif)

sandysage:
Closing as a duplicate of #1256

changecourse:
This is more pronounced in a disorienting way on Chrome vs. Firefox. 

I think a proposed change to this behavior would be to not zoom the text in the card, but rather just the card itself, keep the cursor change, and keep the darker shadow, on hover.

groovecoder:
Hmm ... I can't reproduce this?

lesleyjanenorton:
Closed in #1430

PR #1430, opened 2019-12-06
Fix breach card wiggling on hover.

### LINKED ISSUE #1419: On focus - animation is janky
[Affected versions]:
 - Firefox 71

[Affected Platforms]:
- Windows 10 x64
- MacOS 10.14
- Ubuntu 18.04 x64

[Steps to reproduce]:
1. Navigate to: https://monitor-v2.herokuapp.com/
2. In the upper side of the page, click on "Breaches"
3. Move your mouse over the existing breaches 

[Expected result]:
The existing breaches are properly displayed on mouse hover.

[Actual result]:
On mouse hover, the animation is janky. 

IssueBreaches

lesleyjanenorton:
Use `box-shadow` to simulate a slight scaling effect instead of `transform: scale()`, which was causing a noticeable wiggle in some browsers.

Fixes #1419 , #1287

### DIFF
--- a/public/css/app.css
+++ b/public/css/app.css
@@ -529,15 +529,13 @@ ul li {
 }
 
 .drop-shadow {
-  transform: scale(1);
   box-shadow: 0 0 1px 1px #607d8b08, 1px 1px 10px #4341571a;
-  transition: box-shadow 0.15s ease-in-out, transform 0.15s ease-in-out;
+  transition: all 0.15s ease-in-out;
 }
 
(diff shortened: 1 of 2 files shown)


lesleyjanenorton:
Thanks! UX has given us grace to remove the scaling. Regarding wiggle fixes: If you can stack-overflow it, believe me I tried it. Hacks were partially, but not completely effective and unfortunately `will-change` alone, or in concert with hacks, `would-not-change` the problem either.

PR #1433, opened 2019-12-06
Add linking data to English homepage

### LINKED ISSUE #1432: Add json-ld data to improve search results
Google supports the [JSON Linking Data](https://json-ld.org/) format to create custom search results. 

The Monitor one currently looks like this:
Screen Shot 2019-12-06 at 14 42 52

@rraue created a batch of JSON for MEAO to implement on mozilla.org (that's where I'm from 👋 ) but the Monitor data should really go on the homepage of the monitor site.

```
{
    "@context": "https://schema.org/",
    "@type": "Product",
    "@id": "https://www.mozilla.org/#firefoxmonitor",
    "url": "https://monitor.firefox.com",
    "logo": "{{ SERVER_URL }}/img/firefox-monitor-logo.png",
    "image": [
        "{{ SERVER_URL }}/img/linking_data/monitor.png",
        "{{ SERVER_URL }}/img/firefox-monitor.png",
        "{{ SERVER_URL }}/img/linking_data/alert.png"
    ],
    "name": "Firefox Monitor",
    "brand": {
        "@type": "Brand",
        "@id": "https://www.mozilla.org/#firefoxbrand",
        "name": "Firefox"
        },
    "sameAs": [
        "https://www.wikidata.org/wiki/Q64828890",
        "https://monitor.firefox.com",
        "https://en.wikipedia.org/wiki/Firefox_Monitor"
        ],
    "offers": {
        "@type": "Offer",
        "@id": "https://www.mozilla.org/#firefoxmonitor",
        "priceCurrency": "USD",
        "price": "0",
        "availability": "https://schema.org/InStock"
        }
}
```

Note that the logo and images must be fully qualified links to JPG or PNG files.

Ideally it is only on the English homepage.

stephaniehobson:
Fix #1432

This json-ld data will enhance the search results listing for Firefox Monitor.

- add json-ld
    - properly formatted json
    - fully qualified urls
- add supporting product images
    - there's nothing special about these, swap them out if there are better ones
    - keep in mind they must be png or jpg
- should be on EN homepage only


stephaniehobson:
I could not get the site running locally because I couldn't create a local database. So this PR is 💯  untested.

It's either a useful starting point for #1432 or a complete mess 😜

One thing it definitely isn't doing is constraining including the JSON to only the English homepage (though I think I got it to be the homepage only).

I'm happy to work to improve it or to have someone take it over if that means it'll be done faster.

groovecoder:
Thanks @stephaniehobson ! Is there a way we can preview the custom search results rendering of this JSON-LD ?

views/partials/linking-data/firefox-monitor-product.hbs
```diff
@@ -0,0 +1,34 @@
+<script type="application/ld+json">
```
groovecoder: I think this would require `unsafe-inline` in our CSP which we can't do. Is there an example of `application/ld+json` scripts hosted at a `src=` url?

pdehaan: https://support.google.com/webmasters/thread/9589175?hl=en
https://webmasters.stackexchange.com/questions/104024/linking-external-json-ld-schema-from-schema-org

Seems to imply that it isn’t possible/supported.
First link is a couple months old. Second link is a couple years. Not sure if we could push a branch to Heroku and test, or do a merge/revert if it doesn’t work.

stephaniehobson:
@groovecoder There's no way to know for sure what Google will do with the info. There's a validation tool here: https://search.google.com/structured-data/testing-tool but it doesn't render the enhanced search results. One hopes it would remove at least remove the outdated browser logo.

stephaniehobson:
Since the client doesn't have to do anything with the script for this to work, as long as it doesn't cause CSP errors it could be fine to leave it and see if Google consumes it anyway.

It's also possible to communicate the same info using [Microdata or RDFa](https://developers.google.com/search/docs/guides/intro-structured-data#structured-data-format) but that is more work since it needs to be part of the HTML structure and a lot of that HTML is rendered client side right now but Google needs it server side.

stephaniehobson:
I've found a few references saying a script element with a non executable type value doesn't violate CSP but I'm having trouble tracking down a reference that isn't just some coder in stackoverflow comments.

[Here's Mathias Bynens saying it,](https://mathiasbynens.be/notes/json-dom-csp#script) and he's a good authority. But again, no link to a standard.

groovecoder:
Oh, good point re: `type="application/ld+json"` being non-executable. I'll go with @pdehaan suggestion and push this up to a heroku review server to test. Thanks!

groovecoder:
Ran this locally and saw no CSP violation. Pushed to blurts-server-review.herokuapp.com/ and ran it thru the Google validator and it looks good. Thanks @stephaniehobson !!

image

Issue #1404, opened 2019-11-25
Modify and expand recommendation logic `/breach-details` pages.


lesleyjanenorton:
Expand/modify existing logic on `/breach-details` pages to show recommendations based on a breach's exposed data classes instead of the breach "type". Update `/breach-details` UI to reflect changes.

Recommendations: https://docs.google.com/spreadsheets/d/1MlRHSs5PVEO0baDduWLTBJWWSQuwOCVZewhPuDN3ZKE/edit#gid=0

New / revised copy: https://docs.google.com/document/d/1DHQytRPHF8xPa-VE_PrCqJTIDIGwQ6TiMRui09lgdNs/edit#heading=h.yt4u26dg3ru9

pdehaan:
I love this idea. Do we have any rules yet, or a list of data classes we're targeting?
According to <kbd>npx pdehaan/blurts-data-classes-stats</kbd>, we currently have 120 unique data classes &mdash; so this might be a bit tricky to verify that the UI looks good for each distinct breach (since data classes are so random).

lesleyjanenorton:
I'm fishing around the proposed rules and working copy and will share the with you when I find them but IIRC, the proposed rules are along these lines:

- Provide 4 recommendations for each breach, regardless of total number of exposed data classes.
- Provide specific recommendations for exposed "priority" data classes. Current list is [here](https://github.com/mozilla/blurts-server/blob/d3f8b89e70a3ce5e9ee5cbabd22587937345fee6/template-helpers/breach-detail.js#L45)
- If there are greater than 4 priority data classes exposed, pick the four with the highest weight.
- If there are fewer than 4 priority data classes exposed, back-fill with more generic recommendations.
- When the exposed data classes are less actionable, and zero priority data classes have been exposed, show 4 general digital hygiene recommendations.

lesleyjanenorton:
@pdehaan new recommendations here: https://docs.google.com/spreadsheets/d/1MlRHSs5PVEO0baDduWLTBJWWSQuwOCVZewhPuDN3ZKE/edit#gid=0

pdehaan:
Future notes for Future @pdehaan:
Here's a list of the priority data classes and the top breached site with said data class:

```js
// npm i pdehaan/hibp-chain-api -S
const HIBP = require("hibp-chain-api");

const priorityDataClasses = {
  "government-issued-ids" : {
    weight: 101,
    pathToGlyph: "svg/glyphs/social-security-numbers",
  },
[...]

PR #1434, opened 2019-12-09
Filter out breaches where breach.Domain === ""

lesleyjanenorton:
This restores the filtering out of breaches with no domain.

### DIFF
--- a/controllers/hibp.js
+++ b/controllers/hibp.js
@@ -67,7 +67,7 @@ async function notify (req, res) {
     }
   }
 
-  if (breachAlert.IsSpamList || breachAlert.IsFabricated || !breachAlert.IsVerified) {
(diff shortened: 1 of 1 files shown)


controllers/hibp.js
```diff
@@ -67,7 +67,7 @@ async function notify (req, res) {
     }
   }
 
-  if (breachAlert.IsSpamList || breachAlert.IsFabricated || !breachAlert.IsVerified) {
+  if (breachAlert.IsSpamList || breachAlert.IsFabricated || !breachAlert.IsVerified || breachAlert.Domain === "") {
```
pdehaan: Looks like this was originally changed in #1397.
Did we only change it short term for the popular PDL breach?

lesleyjanenorton: Yep - this was just a short term change for PDL.

groovecoder: Started https://github.com/mozilla/blurts-server/issues/1404#issuecomment-564018724 to consider making this change permanent as we improve recommendations.

PR #1435, opened 2019-12-09
Add recommendation strings

lesleyjanenorton:


pdehaan:
I just did a quick glance through, but everything looked the same as in #1429 :+1:

lesleyjanenorton:
@pdehaan thanks!

locales/en/recommendations.ftl
```diff
@@ -0,0 +1,140 @@
...
+rec-pw-1-cta = Change password for this site
+rec-pw-1 = 
+  Make this password unique and different from any others you use. 
+  A good strategy to follow is to combine two or more unrelated 
+  words to create an entire passphrase.
+
+# Recommendation subhead
+rec-pw-2-subhead = Update other logins using the same password
+# Link title
+rec-pw-2-cta-fx = View logins in { -brand-name }
```
groovecoder: We can use a [`WebChannel`](https://developer.mozilla.org/en-US/docs/Mozilla/JavaScript_code_modules/WebChannel.jsm) to have this open `about:logins` though it needs a corresponding patch in Firefox.

PR #1442, opened 2019-12-13
Add strings for breach resolution feature.

lesleyjanenorton:
This adds the new strings needed for the upcoming breach resolution feature.

### DIFF
--- a/locales/en/app.ftl
+++ b/locales/en/app.ftl
@@ -563,9 +563,16 @@ rec-section-subhead-no-pw = Though passwords weren’t exposed in this breach, t
 # Button
 see-additional-recs = See Additional Recommendations
 
-# This string contains nested markup that becomes a link later in the code.
-# Please do not modify or remove "<a>" and "</a>".
+## This string contains nested markup that becomes a link later in the code.
+## Please do not modify or remove "<a>" and "</a>".
+
 resolve-top-notification = { $affectedEmail } appeared in this breach. <a>What to do next</a>
+
+resolve-top-notification-plural =
+  { $numAffectedEmails ->
+    *[other] { $numAffectedEmails } of your email addresses appeared in this breach. <a>What to do next</a>
+  }
+##
 
 marking-this-subhead = Marking this breach as resolved
 
@@ -640,10 +647,6 @@ progress-complete-message =
   update other logins with stronger passwords.
 ##
 
-# A status indicator that appears in the top right corner of breach cards
-# once they have been resolved.
-resolved = Resolved
-
 resolve-this-breach-link = Resolve this breach
 
 # This string appears in resolved breach cards and is followed by 
@@ -652,3 +655,18 @@ marked-resolved = Marked resolved:
 
 hide-resolved-button = Hide Resolved
 show-resolved-button = Show Resolved
+
+unresolved-passwords-exposed =
+  { $passwords ->
+    [one] Password exposed in unresolved breaches
+   *[other] Passwords exposed in unresolved breaches
+  }
+
+known-data-breaches-resolved = 
+  { $resolvedBreaches ->
+    [one] Known data breach marked as resolved
+   *[other] Known data breaches marked as resolved
+  }
+
+# A status indicator that appears in the top right corner of new breach cards
+new-breach = New
--- a/locales/en/app.ftl
+++ b/locales/en/app.ftl
@@ -657,13 +657,13 @@ hide-resolved-button = Hide Resolved
 show-resolved-button = Show Resolved
 
 unresolved-passwords-exposed =
-  { $passwords ->
+  { $numPasswords ->
     [one] Password exposed in unresolved breaches
    *[other] Passwords exposed in unresolved breaches
   }
 
 known-data-breaches-resolved = 
-  { $resolvedBreaches ->
+  { $numResolvedBreaches ->
     [one] Known data breach marked as resolved
    *[other] Known data breaches marked as resolved
   }
--- a/locales/en/app.ftl
+++ b/locales/en/app.ftl
@@ -576,14 +576,12 @@ marking-this-body =
   you can mark it as resolved. You can still access details about the breach 
   from your dashboard at any time.
 
-# Button
-mark-as-resolve = Mark as Resolved
 
-# Button
-marked-as-resolved = Marked as Resolved
+mark-as-resolve-button = Mark as Resolved
+marked-as-resolved-label = Marked as Resolved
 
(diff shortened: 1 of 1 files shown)


locales/en/app.ftl
```diff
@@ -562,3 +562,97 @@ rec-section-subhead-no-pw = Though passwords weren’t exposed in this breach, t
...
+## These strings contain nested markup that is later used to style the text inside of it.
+## Please do not modify or remove "<span>" and "</span>".
+progress-message-1 = 
+  <span>You’re off to a great start!</span> Check out the remaining breaches to learn 
+  what steps to take.  
+progress-message-2 = 
+  <span>Keep it up!</span> Small changes like updating passwords have a big impact on 
+  keeping your personal info safe.
+progress-message-3 = <span>Nice work resolving those breaches!</span> Keep it up. You’ve got a few more to go.
+progress-message-4 = <span>Nice work resolving those breaches!<span> Keep it up. You’ve got a few more to go.
```
pdehaan: message 3 and 4 are the same, except this line has an improperly closed `</span>` tag.

locales/en/app.ftl
```diff
@@ -562,3 +562,97 @@ rec-section-subhead-no-pw = Though passwords weren’t exposed in this breach, t
...
+marking-this-body = 
+  <span>Once you’ve taken the steps you can to address this breach</span>,
+  you can mark it as resolved. You can still access details about the breach 
+  from your dashboard at any time.
+
+# Button
+mark-as-resolve = Mark as Resolved
+
+# Button
+marked-as-resolved = Marked as Resolved
```
flodolo: Is this really a button?

locales/en/app.ftl
```diff
@@ -562,3 +562,97 @@ rec-section-subhead-no-pw = Though passwords weren’t exposed in this breach, t
...
+
+# Link
+return-to-breach-details = Return to breach details
+
+# Link
+go-to-dashboard = Go to Dashboard
+
+# This string appears above a breach resolution progress bar and indicates
+# the percentage of breaches a user has resolved. For instance, "27% complete".
+progress-percent-complete = { $percentComplete } complete
```
flodolo: We should pass the number here, not the formatted value (what if a language wants a space, e.g. `27 %`?)

locales/en/app.ftl
```diff
@@ -562,3 +562,97 @@ rec-section-subhead-no-pw = Though passwords weren’t exposed in this breach, t
...
+
+# This string appears in the purple callouts at the top of the user dashboard and shows
+# the total number of breaches a user has resolved. For instance, "5 Resolved".
+num-resolved = { $numResolved } Resolved
```
flodolo: I know that for English it's the same, but this should be a plural string to trigger the plural UI in Pontoon (adjectives are declined for plurals in most languages).

locales/en/app.ftl
```diff
@@ -562,3 +562,97 @@ rec-section-subhead-no-pw = Though passwords weren’t exposed in this breach, t
...
+
+progress-intro-subhead = New in { -product-name }: Mark breaches as resolved
+progress-intro-message =
+  After reviewing the details about a breach and taking steps to protect 
+  your personal info, you can mark breaches as resolved.
+progress-status = { $numResolvedBreaches} out of { $numTotalBreaches } breaches marked as resolved
```
flodolo: Same here. It doesn't affect English, but this should be a plural string based on `$numTotalBreaches`

locales/en/app.ftl
```diff
@@ -562,3 +562,97 @@ rec-section-subhead-no-pw = Though passwords weren’t exposed in this breach, t
...
+progress-complete = All known breaches have been marked as resolved
+
+## These strings contain nested markup that is later used to style the text inside of it.
+## Please do not modify or remove "<span>" and "</span>".
```
flodolo: Nit: empty line after a section comment

locales/en/app.ftl
```diff
@@ -562,3 +562,97 @@ rec-section-subhead-no-pw = Though passwords weren’t exposed in this breach, t
...
+# A status indicator that appears in the top right corner of breach cards
+# once they have been resolved.
+resolved = Resolved
+
+# Link
+resolve-this-breach = Resolve this breach
+
+# This string appears in resolved breach cards and is followed by 
+# the date the user marked the breach as resolved.
+marked-resolved = Marked resolved:
```
flodolo: How is the date going to be displayed? Can it be exposed as part of the string?

locales/en/app.ftl
```diff
@@ -562,3 +562,97 @@ rec-section-subhead-no-pw = Though passwords weren’t exposed in this breach, t
...
+marking-this-subhead = Marking this breach as resolved
+
+# This string contains nested markup that is later used to style the text inside of it.
+# Please do not modify or remove "<span>" and "</span>".
+marking-this-body = 
+  <span>Once you’ve taken the steps you can to address this breach</span>,
+  you can mark it as resolved. You can still access details about the breach 
+  from your dashboard at any time.
+
+# Button
```
flodolo: Instead of adding comments, you could put `-button` in the string ID.

locales/en/app.ftl
lesleyjanenorton: Here is how it will show up in resolved breach cards:
Screen Shot 2019-12-18 at 4 54 16 PM
We could expose the date in the string but we'd probably have to add markup around the date and then do some additional juggling in the template.

locales/en/app.ftl
```diff
@@ -562,3 +562,111 @@ rec-section-subhead-no-pw = Though passwords weren’t exposed in this breach, t
...
+
+# This string appears in resolved breach cards and is followed by 
+# the date the user marked the breach as resolved.
+marked-resolved = Marked resolved:
+
+hide-resolved-button = Hide Resolved
+show-resolved-button = Show Resolved
+
+unresolved-passwords-exposed =
+  { $passwords ->
```
flodolo: Can we use `$numPasswords` for consistency with other strings?

locales/en/app.ftl
```diff
@@ -562,3 +562,111 @@ rec-section-subhead-no-pw = Though passwords weren’t exposed in this breach, t
...
+    [one] Password exposed in unresolved breaches
+   *[other] Passwords exposed in unresolved breaches
+  }
+
+known-data-breaches-resolved = 
+  { $resolvedBreaches ->
```
flodolo: Same here, `$numResolvedBreaches`

PR #1443, opened 2019-12-13
Remove old scan-res bits and pieces.

lesleyjanenorton:
This clips out partials and code related to the previous version of the scan results page.


views/partials/sign-up-banners/breach-detail-sign-up-banner.hbs
```diff
@@ -1,7 +1,7 @@
 
       <div class="sign-up-wrapper breach-detail-sign-up drop-shadow fxa-signup-gradient">
         <div class="mw-8 sign-up-banner">
-          <div class="sign-up-headline-list-wrapper flx flx-col">
+          <div class="flx flx-col">
```
maxxcrawford: Should the `sign-up-headline-list-wrapper` class be removed from CSS to, or is it still in use on the Breach Details signup?

PR #1452, opened 2019-12-18
fix #1426: handle HTTPError during getProfileData

### LINKED ISSUE #1426: Error page after revoking Monitor from FXA
See https://bugzilla.mozilla.org/show_bug.cgi?id=1598314.

groovecoder:
Steps to Test:

1. Sign in http://localhost:6060/ and go to http://localhost:6060/user/dashboard
2. In another tab, go to https://stable.dev.lcip.org/settings/clients
3. Click "Revoke" next to the most recent "blurts-server (local)"
4. On the first tab, refresh http://localhost:6060/user/dashboard

Expected results:
Should be signed out and redirected to home page

(Old result: the page would hang.)

### DIFF
--- a/middleware.js
+++ b/middleware.js
@@ -139,6 +139,10 @@ async function requireSessionUser(req, res, next) {
     return res.redirect(`/oauth/init?${queryParams}`);
   }
   const fxaProfileData = await FXA.getProfileData(user.fxa_access_token);
+  if (fxaProfileData.hasOwnProperty("name") && fxaProfileData.name === "HTTPError") {
+    delete req.session.user;
+    return res.redirect("/");
+  }
   await DB.updateFxAProfileData(user, fxaProfileData);
   req.session.user = user;
   req.user = user;
--- a/lib/fxa.js
+++ b/lib/fxa.js
@@ -73,10 +73,15 @@ const FXA = {
   },
 
   async getProfileData(accessToken) {
-    const data = await got(FxAOAuthUtils.profileUri,
-      { headers: { Authorization: `Bearer ${accessToken}` } }
-    );
-    return data.body;
+    try {
+      const data = await got(FxAOAuthUtils.profileUri,
+        { headers: { Authorization: `Bearer ${accessToken}` } }
+      );
+      return data.body;
+    } catch (e) {
+      log.warn("getProfileData", {stack: e.stack});
+      return e;
+    }
   },
 
   async sendMetricsFlowPing(path) {
--- a/tests/middleware.test.js
+++ b/tests/middleware.test.js
@@ -17,6 +17,7 @@ test("requireSessionUser calls getProfileData, updateFxAProfileData, and sets re
   jest.mock("../db/DB");
   jest.mock("../lib/fxa");
   FXA.getProfileData = jest.fn();
+  FXA.getProfileData.mockReturnValueOnce({});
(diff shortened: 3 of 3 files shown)


lesleyjanenorton:
After revocation should I still be able to log in to Monitor with my prior credentials and still see all of my non-primary emails?

middleware.js
```diff
@@ -139,6 +139,10 @@ async function requireSessionUser(req, res, next) {
     return res.redirect(`/oauth/init?${queryParams}`);
   }
   const fxaProfileData = await FXA.getProfileData(user.fxa_access_token);
+  if (fxaProfileData.hasOwnProperty("name") && fxaProfileData.name === "HTTPError") {
+    delete req.session.user;
```
lesleyjanenorton: Should we remove their info from our db here or should hitting revoke just end the session?

lesleyjanenorton: Should we also delete their info here or does "Revoke" just end the session?

groovecoder: Hmm ... have to think about that one a bit ... I don't think we can/should delete DB data here, because the HTTP request for profile data might fail for some other reason.

 @mozilla/fxa-devs - even if we check for a `401` status code in the response (to the request for profile data), are there other reasons we would get that status code back from FXA?

Also, it's easier to mis-click on the "Revoke" button in the FXA UI than it is to go back thru Monitor and re-add up to 15 email addresses.

So, it seems the canonical UX for removing Monitor data should be the Monitor-side `user/preferences` page. What do you think @changecourse ?

changecourse: We would likely need to specify what exactly we'd want to do in both the description text of the "Remove Firefox Monitor" section within Preferences, as well as re-emphasize it in the confirmation modal... as for which approach, let's discuss in today's meeting.

rfk: > even if we check for a 401 status code in the response (to the request for profile data),
> are there other reasons we would get that status code back from FXA?

If you get a `401` here, that means your *access token* has expired or been revoked or similar. Dealing with expired access tokens is a normal and expected part of OAuth and doesn't necessarily mean that you've lost access to the user's account.

In the face of a `401` here, I suggest trying to [use your refresh token to obtain a fresh access token](https://oauth.net/2/grant-types/refresh-token/), and then trying to request the profile data again. If you get a `400` error with errno `108` when trying to use your refresh token, that's a very strong indicator that the user has explicitly disconnected Monitor from their account.

groovecoder: Cool, thanks for the info!

groovecoder: @changecourse after some discussion in https://github.com/mozilla/fxa/issues/3744, here's what I'm thinking ...

* If we get a 401 response to the request with our access token, we'll try to use our refresh token to get a fresh access token and use that to get the profile data
* If we get a 400 response to the request with our access token, with errno: 108, inform the user that Monitor has been revoked from their Firefox Account, and prompt them to either:
   * Reconnect Monitor to their FxA (Trigger the oauth flow again)
   * Delete Firefox Monitor from their account

changecourse: sounds good to me @groovecoder, thanks for the update. Please let me know if you need anything from me, design wise.

groovecoder:
I filed https://github.com/mozilla/blurts-server/issues/1464 for prompting the user about deleting data when they revoke Monitor from their FXA. So, this is good for new r? @lesleyjanenorton

PR #1457, opened 2019-12-23
Fix the flickering active-link underline

### LINKED ISSUE #1415: Header options flickers when changed
[Affected versions]:
 - Firefox 70.0.1

[Affected Platforms]:
 - Windows 10 x64
 - MacOS 10.14
 - Ubuntu 18.04 x64

[Steps to reproduce]:
 1. Navigate to: https://monitor-v2.herokuapp.com/
 2. Navigate through the header options but wait for the page to be completed loaded before moving the mouse.

[Expected result]:
 - Navigating through the header option is successfully done.

[Actual result]:
 - A flicker is noticed each time you navigate through the header options. Please see the screen-cast for more details.

FlickerIssue

lesleyjanenorton:
Fixes #1415

- Sets the `.active-link` class from the template instead of `monitor.js`.
- Adds mouseenter/mouseleave event listeners to the other header links to handle showing and hiding the gradient underline.

### DIFF
--- a/public/css/header.css
+++ b/public/css/header.css
@@ -95,7 +95,6 @@ nav {
   font-weight: 700;
 }
 
-.active-link span::after,
 .nav-link span::after {
   height: 2px;
   background: var(--monitorGradient);
--- a/public/css/header.css
+++ b/public/css/header.css
@@ -86,17 +86,17 @@ nav {
   font-size: 16px;
 }
 
+.nav-link span {
+  position: relative;
+  pointer-events: none; /* Prevents conflicts with GA events that should fire off the wrapping <a> tag. */
+}
+
 .active-link {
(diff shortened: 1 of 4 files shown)


public/css/header.css
```diff
@@ -86,17 +86,17 @@ nav {
...
 
-.active-link-underline {
-  position: relative;
-  pointer-events: none; /* Make sure span attributes aren't used for GA ping */
+.active-link {
+  font-weight: 700;
 }
 
-.active-link-underline::after {
+.active-link span::after,
```
maxxcrawford: `.active-link span::after` has classes both here and on L114. Should that be the case?

lesleyjanenorton: Fixed - thanks!

PR #1467, opened 2020-01-09
 fix #1412: annotate breaches with IsResolved and recencyIndex 

### LINKED ISSUE #1412: Augment `bundleVerifiedEmails` function with logic to add properties to the `emailEntry` objects
Augment `bundleVerifiedEmails` function with logic to add properties to the `emailEntry` objects:
  - `"resolved": true|false`
  - `"recency_index": Number`

groovecoder:


hibp.js
```diff
@@ -118,9 +118,14 @@ const HIBP = {
       if (sha1.toUpperCase() === sha1Prefix + breachedAccount.hashSuffix) {
         foundBreaches = allBreaches.filter(breach => breachedAccount.websites.includes(breach.Name));
         foundBreaches = this.filterBreaches(foundBreaches);
+
+        // NOTE: DO NOT CHANGE THIS SORT LOGIC
+        // We store breach resolutions by recency indices,
+        // so that our DB does not contain any part of any user's list of accounts
```
groovecoder: self-note: we should move this sorting function into a utility function and use it everywhere when we mark breaches as resolved or when we show breaches as resolved/not.

controllers/user.js
```diff
@@ -125,16 +125,36 @@ async function add(req, res) {
   res.redirect("/user/preferences");
 }
 
-async function bundleVerifiedEmails(email, ifPrimary, id, verificationStatus, allBreaches) {
+function getResolvedBreachesForEmail(user, email) {
+  if (typeof user.resolved_breaches !== "object") {
```
lesleyjanenorton: ...and TIL you don't need parentheses!

hibp.js
lesleyjanenorton: One thing we should probably consider is what happens if we add back in other breach types (IsSpamList, etc)?

groovecoder: Updated to annotate the breaches with `recencyIndex` BEFORE `filter`ing the breaches.

groovecoder: Great catch, BTW!

groovecoder:
I'm going to say unlikely for now, which is a sure-fire way to make it happen ASAP. I'd like to skip that flexibility for now just to keep things as simple as possible. If we add another source we can add the complexity for resolutions then.

tests/controllers/user.test.js
```diff
@@ -30,6 +29,71 @@ function expectResponseRenderedSubpagePartial(resp, partial) {
...
+  expect(renderCallContext.verifiedEmails.length).toEqual(2);
+  expect(renderCallContext.unverifiedEmails.length).toEqual(1);
+  expect(renderCallContext.verifiedEmails[0].breaches.length).toEqual(mockReturnedBreaches.length);
+});
+
+
+test("user getDashboard with valid user renders verified emails and resolved breaches", async () => {
+  const testSubscriberEmail = TEST_SUBSCRIBERS.firefox_account.primary_email;
+  const testSubscriber = await DB.getSubscriberByEmail(testSubscriberEmail);
+  // TODO: figure out how the F to put resolved_breaches in the test db seed?
```
groovecoder: When I try to add a `resolved_breaches: {"firefoxaccount@test.com": [0]},` line in the `firefox_account` object of `TEST_SUBSCRIBERS` in the `db/seeds/test_subscribers.js` file, knex can't seem to insert into the `subscribers` table. 😢

groovecoder:
Added a couple tests for `getDashboard` to at least verify the right number of breaches and are included and marked as resolved.

groovecoder:
Closing in favor of https://github.com/mozilla/blurts-server/pull/1481.

PR #1468, opened 2020-01-09
POST breach resolution endpoint 1413

groovecoder:
Depends on https://github.com/mozilla/blurts-server/pull/1467 so we should merge that first. But I want a sanity check on this approach sooner than later ... @lesleyjanenorton @pdehaan ?

This should accept either of the following HTTP POST requests:
```
POST /user/breach-resolution

emailAddressId=123&recencyIndex=0
```

Will mark the logged-in subscriber's email_addresses.id's 0th breach as resolved.

```
POST /user/breach-resolution

emailAddressId=&recencyIndex=0
```

Will mark the logged-in subscriber's primary_email's 0th breach as resolved.

Could easily change this to accept a JSON body instead of form-encoded body.

### DIFF
--- a/controllers/user.js
+++ b/controllers/user.js
@@ -125,16 +125,36 @@ async function add(req, res) {
   res.redirect("/user/preferences");
 }
 
-async function bundleVerifiedEmails(email, ifPrimary, id, verificationStatus, allBreaches) {
+function getResolvedBreachesForEmail(user, email) {
+  if (typeof user.resolved_breaches !== "object") {
+    return [];
+  }
+  return user.resolved_breaches.hasOwnProperty(email) ? user.resolved_breaches[email] : [];
+}
+
+function annotateFoundBreachesAsResolvedOrNot(foundBreaches, resolvedBreaches) {
+  const annotatedFoundBreaches = [];
+  foundBreaches.forEach( (breach, index) => {
+    const IsResolved = resolvedBreaches.includes(index) ? true : false;
+    const annotatedBreach = Object.assign({IsResolved, recencyIndex: index}, breach);
+    annotatedFoundBreaches.push(annotatedBreach);
+  });
+  return annotatedFoundBreaches;
+}
+
+async function bundleVerifiedEmails(options) {
+  const { user, email, recordId, recordVerified, allBreaches} = options;
   const lowerCaseEmailSha = sha1(email.toLowerCase());
   const foundBreaches = await HIBP.getBreachesForEmail(lowerCaseEmailSha, allBreaches, true);
+  const resolvedBreaches = getResolvedBreachesForEmail(user, email);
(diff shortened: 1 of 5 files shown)


controllers/user.js
```diff
@@ -72,6 +72,19 @@ async function updateCommunicationOptions(req, res) {
...
+async function resolveBreach(req, res) {
+  const sessionUser = req.user;
+  // TODO: verify that req.body.emailAddressId belongs to sessionUser
+  const updatedSubscriber = await DB.setResolvedBreach({
+    subscriber: sessionUser,
+    emailAddresses: req.body.emailAddressId,
+    recencyIndex: req.body.recencyIndex,
+  });
+  req.session.user = updatedSubscriber;
+  return res.json("Breach marked as resolved.");
```
pdehaan: Q: Is this string user facing at all?

controllers/user.js
```diff
@@ -72,6 +72,19 @@ async function updateCommunicationOptions(req, res) {
 }
 
 
+async function resolveBreach(req, res) {
+  const sessionUser = req.user;
+  // TODO: verify that req.body.emailAddressId belongs to sessionUser
+  const updatedSubscriber = await DB.setResolvedBreach({
+    subscriber: sessionUser,
+    emailAddresses: req.body.emailAddressId,
+    recencyIndex: req.body.recencyIndex,
```
pdehaan: Q: Do we need to verify that the `req.body` object includes the `emailAddressId` and `recencyIndex`?

controllers/user.js
```diff
@@ -125,16 +138,36 @@ async function add(req, res) {
...
+  if (typeof user.resolved_breaches !== "object") {
+    return [];
+  }
+  return user.resolved_breaches.hasOwnProperty(email) ? user.resolved_breaches[email] : [];
+}
+
+function annotateFoundBreachesAsResolvedOrNot(foundBreaches, resolvedBreaches) {
+  const annotatedFoundBreaches = [];
+  foundBreaches.forEach( (breach, index) => {
+    const IsResolved = resolvedBreaches.includes(index) ? true : false;
```
pdehaan: Using an `index` feels odd, since I'm not 100% sure how reliable those are? Couldn't they change if new breaches are added/removed/resorted? Wasn't sure if we could use something like the `breach.Name` or something more reliable.

Also, I believe `.includes()` returns a Boolean, so the `? true : false` probably isn't needed.

pdehaan: Well, I guess this is answered below in the excellent comment:

>         // NOTE: DO NOT CHANGE THIS SORT LOGIC
>         // We store breach resolutions by recency indices,
>         // so that our DB does not contain any part of any user's list of accounts

groovecoder:
Closing in favor of https://github.com/mozilla/blurts-server/pull/1481 which is rebased on front-end changes too.

Issue #1471, opened 2020-01-11
strengthen-pin-security.svg is 404ing


pdehaan:
https://monitor.firefox.com/breach-details/vtech

Failed to load resource: the server responded with a status of 404 (strengthen-pin-security.svg:1)

![Firefox_Monitor](https://user-images.githubusercontent.com/557895/72199077-85012f80-33eb-11ea-81cd-799473079be9.png)

Shivansh2407:
I would like to work on this issue.

Shivansh2407:
.rec-dob must need to be added to the css styling . I am submitting a PR for the same .Please check and let me know.

pdehaan:
@Shivansh2407 Thanks for the investigation. I have another theory about possibly just incorrect case in the icon name in https://github.com/mozilla/blurts-server/pull/1472#issuecomment-573362248.

❤️

lesleyjanenorton:
Fixed in #1472

PR #1472, opened 2020-01-11
Update breach-detail.css

Shivansh2407:
Issue:
strengthen-pin-security.svg is 404ing #1471

Solution :
I have seen that .rec-dob item was not added along with other items in height styling.That might have created the issue. Please review the PR and let me know soon


pdehaan:
@Shivansh2407 Interesting find! Thanks for the PR.

I'm not 100% familiar with the CSS, but from the looks of the DOM and CSS, it almost looks like the icons use both the `.rec-img` and `.rec-dob` classes, so I'm assuming we only overwrote the heights for a few select icons.
/cc @lesleyjanenorton @maxxcrawford who might have more insight.

But I wouldn't expect a missing/incorrect SVG/icon height to cause the hard 404 response code.

```html
<div class="rec-img rec-dob"></div>
```

https://github.com/mozilla/blurts-server/blob/5095da29cf8bb8a178d2642aef5dc690611498a4/public/css/breach-detail.css#L216-L231

---

But what is most interesting, is that I managed to fix this via DevTools locally by checking the filesystem and modifying the filename to match what we have in GitHub.

## Original:

https://github.com/mozilla/blurts-server/blob/5095da29cf8bb8a178d2642aef5dc690611498a4/public/css/breach-detail.css#L160-L162

## Modified:
```css
.rec-dob {
    background-image: url(/img/recommendation-icons/strengthen-PIN-security.svg);
}
```

Where all the icons are located in https://github.com/mozilla/blurts-server/tree/master/public/img/recommendation-icons

A quick scan shows 3 files that are mixed case, and possibly worth more investigation:

1. [/public/img/recommendation-icons/set-2FA.svg](https://github.com/mozilla/blurts-server/blob/master/public/img/recommendation-icons/set-2FA.svg)
2. [/public/img/recommendation-icons/strengthen-PIN-security.svg](https://github.com/mozilla/blurts-server/blob/master/public/img/recommendation-icons/strengthen-PIN-security.svg)
3. [/public/img/recommendation-icons/use-mask-IP-service.svg](https://github.com/mozilla/blurts-server/blob/master/public/img/recommendation-icons/use-mask-IP-service.svg)

Although, I believe "set-2FA.svg" and "use-mask-IP-service.svg" are correct, per:

https://github.com/mozilla/blurts-server/blob/5095da29cf8bb8a178d2642aef5dc690611498a4/public/css/breach-detail.css#L124-L126

https://github.com/mozilla/blurts-server/blob/5095da29cf8bb8a178d2642aef5dc690611498a4/public/css/breach-detail.css#L144-L146

Looks like we're possibly just bitten by the old "case sensitive file system" trick again (which might explain why it potentially worked locally, but failed once it was deployed to a more strict OS/stack). 🤷‍♂

pdehaan:
Re: file sensitive filenames, I think this might be the only incorrect file name in the CSS that uses `url()`. 

But warning, my code is sketchy at best. At _BEST_.

```js
// git grep -Po "url\((.*?)\)" public/**/*.css | grep -v ".woff" > svg-files.txt

const cp = require("child_process");

const fs = require("fs").promises;
const path = require("path");

// Look for the `url()` pattern in all the public/**/*.css files...
const CMD = `git grep -Po "url\\((.*?)\\)" public/**/*.css`;
const cssFiles = cp
  .execSync(CMD)
  .toString()
  .trim()
  .split("\n");

const fileMap = new Map();

main(cssFiles);

async function main(matches = []) {
  let files = matches.map(line => {
    return line
      .split(":", 2)
      .pop()
      .match(/url\(["']?(.*?)["']?\)/)[1];
  });
  // Dedupe the files...
  files = new Set(files);
  for (const file of [...files]) {
    const dir = path.join("./public", path.dirname(file));
    if (!fileMap.has(dir)) {
      // Cache the directory files...
      fileMap.set(dir, await fs.readdir(dir));
    }
    const dirfiles = fileMap.get(dir);
    if (!dirfiles.includes(path.basename(file))) {
      console.log(`Unable to find file, ${file}`);
    }
  }
}
```

Current output is only this one file in question:

```sh
Unable to find file, /img/recommendation-icons/strengthen-pin-security.svg
```

Shivansh2407:
What I do next in order to fix this?

pdehaan:
@Shivansh2407 Lets see if fixing the filename case fixes the issue. My macOS isn't case sensitive, but you can see if changing the filename or the CSS so they match fixes the issue locally for you. I'm guessing we might not be able to fully verify until it lands on Heroku dev server or the staging server which seems to be case sensitive.

Thanks!

lesleyjanenorton:
@Shivansh2407 I agree with @pdehaan- changing the filename for the SVG in question from `strengthen-PIN-security.svg` to `strengthen-pin-security.svg` should fix this.

Shivansh2407:
Will Do the changes and let you know @lesleyjanenorton

Shivansh2407:
Done . Please Check and let me know

Shivansh2407:
@lesleyjanenorton Mam the changes are done to the CSS file. Now it is done .

public/css/breach-detail.css
```diff
@@ -227,8 +227,7 @@ button.fade-out {
   .rec-ip-non-us,
   .rec-ip-us,
   .rec-sec-qa,
-  .rec-dob,
-  .rec-pins {
+  .rec-dob {
```
lesleyjanenorton: Almost there! Looks like we just need to remove the `.rec-dob` selector here and then we'll be all set.

Shivansh2407:
Removed .rec-dob selector.

Shivansh2407:
Thank You @lesleyjanenorton and @pdehaan .

pdehaan:
FWIW, here's a better version of my script above that fetches the minified CSS from the production server and compares it to the local filesystem:

```js
const fs = require("fs").promises;
const path = require("path");

const axios = require("axios");

main("../../mozilla/blurts-server/public");

async function main(
  publicDir = "./public",
  cssUrl = "https://monitor.firefox.com/dist/app.min.css"
) {
  const { data: css } = await axios.get(cssUrl);
  let images = css
    .match(/url\((.*?)\)/g)
    .map(url => url.replace(/^url\((.*?)\)$/, "$1"))
    .sort();
  // Dedupe images.
  images = [...new Set(images)];

  for (const image of images) {
    const fullpath = path.join(publicDir, image);
    try {
      await fs.stat(fullpath);
    } catch (err) {
      console.error(`Unable to find "${image}"`);
      process.exitCode = 1;
    }
  }
}
```

This approach also comes with it's own caveats though (where undeployed changes between Git and production might cause some false results), but it might be worth the risk.

Currently it shows zero errors, but I can reproduce the previous error if I rename a file back to a case insensitive version.
The other interesting (and probably very relevant) tip is that I repartitioned my macOS HD and moved all my GitHub-related code into a _case-sensitive_ drive (which means all my apps can still be on the case-insensitive bit to avoid future headaches). But I'm hoping that being on a case sensitive drive might help me catch errors in the future (even if this particular bug was specific to certain breaches, and hidden behind a toggle button).

PR #1488, opened 2020-01-17
Fix false negative breach report emails

### LINKED ISSUE #1461: False negatives: Firefox Monitor found no known breaches (e-mail) when adding an address for which there are breaches
### Steps

1. https://monitor.firefox.com/user/dashboard
2. add (another) e-mail address

### Expected

3. an e-mail that tells the true number of breaches for the address

### Actual result

3. an e-mail that gives a false sense of security

#### E-mail view of things

image

> **Firefox Monitor found no known breaches**

> … appeared in 0 known  data breaches
> 
> Firefox Monitor warns you about data breaches involving your personal info. So far, no breaches were found. We’ll send you an alert if your email address appears in a new breach. …

### A subsequent dashboard view of the two addresses that are reportedly breach-free

image

lesleyjanenorton:
Fixes #1461

### DIFF
--- a/controllers/user.js
+++ b/controllers/user.js
@@ -198,7 +198,7 @@ async function _verify(req) {
   const verifiedEmailHash = await DB.verifyEmailHash(req.query.token);
   let unsafeBreachesForEmail = [];
   unsafeBreachesForEmail = await HIBP.getBreachesForEmail(
-    sha1(verifiedEmailHash.email),
+    sha1(verifiedEmailHash.email.toLowerCase()),
     req.app.locals.breaches,
     true,
   );


controllers/user.js
```diff
@@ -198,7 +198,7 @@ async function _verify(req) {
   const verifiedEmailHash = await DB.verifyEmailHash(req.query.token);
   let unsafeBreachesForEmail = [];
   unsafeBreachesForEmail = await HIBP.getBreachesForEmail(
-    sha1(verifiedEmailHash.email),
+    sha1(verifiedEmailHash.email.toLowerCase()),
```
pdehaan: Q: Should we lowercase the email _here_, or within the `getSha1()` method?
Curious if there are other places that we might be not lowercasing that is causing potential edge cases.

lesleyjanenorton: Yeah it seems like that would make a lot more sense to just do it within `getSha1()` but I think I remember this being discussed before and there being some reason (blanking on specifics) why we didn't take that approach. I filed #1489 in the interest of getting this totally sorted out.

lesleyjanenorton: although... we DO lowercase the email address from within sha1() when we hash emails from the client....   https://github.com/mozilla/blurts-server/blob/a09591d8190404e6191edc067e2426fd95e726d0/public/js/scan-email.js#L8

groovecoder: Since this fixes the email issue, let's merge it as-is. I'll add #1489 to the top of our backlog/todos as a defect since it has bitten us at least twice now.

lesleyjanenorton: Ok I will push the rest of my updates up in a new PR but I'm pretty sure we need to 1.) run `lowercase-all-records.js` again on production to correct any hashes created and stored for email addresses with the uppercase letters. and 2.) update our sha1 function and all of the places we're calling it.

PR #1494, opened 2020-01-21
Audit usage of sha1, .toLowerCase() and .toUpperCase() - WIP

lesleyjanenorton:
This attempts to bring come consistency to the way we hash emails for scanning and storing. It's WIP but I'm pushing it up early so that we can start discussing.

A few findings/confirmations from the ongoing sha1 deep dive:
#### In order to return correct scan results from HIBP:
- An email should be lowercased before hashing
- The hash itself should be numbers and uppercase letters only

#### The confusing gotcha:
- Hashes are stored in our database in _lowercase_ letters
- When HIBP callbacks hit our `/notify` endpoint, we first have to lowercase everything in order to match the hash coming from HIBP to what is stored in our DB.

Maybe we can consider converting our existing hashes to uppercase letters to match HIBP (this, I believe, is the path of greatest sanity). I think w may need to rerun `lowercase-all-records.js` on production because we may have stored hashes for emails that were not first converted to lowercase letters (which caused the false negative emails, did not affect dashboard since those breach results are retrieved differently, and I'm concerned could result in us not correctly sending breach alerts to some affected individuals).

### DIFF
--- a/db/DB.js
+++ b/db/DB.js
@@ -23,6 +23,14 @@ const log = mozlog("DB");
 
 
 const DB = {
+
+  // Hashes are stored in the db using lowercase letters
+  getLowerCaseHash(email) {
+    email = getSha1(email);
+    return email.toLowerCase();
+  },
+
+
   async getSubscriberByToken(token) {
     const res = await knex("subscribers")
       .where("primary_verification_token", "=", token);
@@ -117,7 +125,7 @@ const DB = {
     const res = await knex("email_addresses").insert({
       subscriber_id: user.id,
       email: email,
-      sha1: getSha1(email),
+      sha1: this.getLowerCaseHash(email),
       verification_token: uuidv4(),
       verified: false,
     }).returning("*");
@@ -160,6 +168,7 @@ const DB = {
     }
   },
 
+
   // Used internally.
   async _addEmailHash(sha1, email, signup_language, verified = false) {
     try {
@@ -169,7 +178,7 @@ const DB = {
           const res = await knex("subscribers")
             .update({
               primary_email: email,
-              primary_sha1: getSha1(email.toLowerCase()),
+              primary_sha1: this.getLowerCaseHash(email),
               primary_verified: verified,
               updated_at: knex.fn.now(),
             })
@@ -183,7 +192,7 @@ const DB = {
         // Always add a verification_token value
         const verification_token = uuidv4();
         const res = await knex("subscribers")
-          .insert({ primary_sha1: getSha1(email.toLowerCase()), primary_email: email, signup_language, primary_verification_token: verification_token, primary_verified: verified })
+          .insert({ primary_sha1: this.getLowerCaseHash(email), primary_email: email, signup_language, primary_verification_token: verification_token, primary_verified: verified })
           .returning("*");
         return res[0];
       });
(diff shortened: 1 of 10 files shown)


scripts/lowercase-all-records.js
```diff
@@ -23,10 +22,10 @@ async function subscribeLowercaseHashToHIBP(emailAddress) {
   const subsWithUpperCount = subRecordsWithUpperChars.length;
   console.log(`found ${subsWithUpperCount} subscribers records with primary_email != lower(primary_email). fixing ...`);
   for (const subRecord of subRecordsWithUpperChars) {
-    const lowerCasedSha1 = await subscribeLowercaseHashToHIBP(subRecord.primary_email);
+    const emailHash = await subscribeLowercaseHashToHIBP(subRecord.primary_email);
     await knex("subscribers")
       .update({
-        primary_sha1: lowerCasedSha1,
+        primary_sha1: emailHash.toLowerCase(),
```
lesleyjanenorton: Because we didn't explicitly lowercase hashes before storing them in our db when we last ran `lowercase-all-records.js`, does this potentially mean we may not be able to correctly match them since we lowercase all hashes when they come in from HIBP?

scripts/lowercase-all-records.js
```diff
@@ -10,10 +10,9 @@ const getSha1 = require("../sha1-utils");
 
 
 async function subscribeLowercaseHashToHIBP(emailAddress) {
-  const lowerCasedEmail = emailAddress.toLowerCase();
-  const lowerCasedSha1 = getSha1(lowerCasedEmail);
-  await HIBP.subscribeHash(lowerCasedSha1);
-  return lowerCasedSha1;
+  const emailHash = getSha1(emailAddress);
+  await HIBP.subscribeHash(emailHash);
+  return emailHash;
```
pdehaan: This bit feels interesting here. If I'm following the diffs correctly, previously this would have been a lowercased SHA, but now we're explicitly calling `.toUpperCase()` on the `return` statement of `getSha1Hash()` below. Not sure if there will be any fallout from changing the case here, but I suspect not. Especially since it looks like we call `.toLowerCase()` on the string again below on L28 and L42 when inserting the hashes into the DB.
Juggling the string cases feels a bit fragile. Not sure if it's easier to just always return a lowercased hash and then force the string to uppercase in the `HIBP.subscribeHash()` method, if "Hashes must be uppercased to return correct results from HIBP".

db/DB.js
```diff
@@ -23,6 +23,14 @@ const log = mozlog("DB");
 
 
 const DB = {
+
+  // Hashes are stored in the db using lowercase letters
+  getLowerCaseHash(email) {
+    email = getSha1(email);
+    return email.toLowerCase();
```
pdehaan: Same comment as below. It feels odd that `getSha1()` returns the uppercased version of the hash (see snippet below), and then we seem to be calling `.toLowerCase()` again in most places and flipping the hash strings.

```js
  // Hashes must be uppercased to return correct results from HIBP
  return sha1EmailHash.toUpperCase();
```

lesleyjanenorton: Completely agree. I'm closing this in favor of #1498 - which is (I think) a saner approach.

lesleyjanenorton:
Closing in favor of #1498 , will push the clean-up updates to `scan-results.js`.

PR #1498, opened 2020-01-22
Audit Sha1() usage

lesleyjanenorton:
Replaces our single `sha1()` function with `getSha1ForHIBP()` and `getLowerCasedSha1()` which should return hashes correctly formatted for the purpose. This will hopefully help us cut down on the .toUpperCase()/.toLowerCase()/.toBasketCase() shenanigans. 

`getSha1ForHIBP(emailAddress)` 
- converts emailAddress to lowercase letters before hashing
- converts the hash to uppercase letters.
- should be used anytime we're creating a hash for scanning HIBP

`getLowerCaseSha1(emailAddress)` 
- converts emailAddress to lowercase letters before hashing
- leaves the hash in lowercase letters to be stored in our DB.
- should be used anytime a hash is being created for storing in our DB, or for finding something in our DB. 

TODO:
It would be great to add some related tests and a readme documenting the Laws of Monitor Casing.

### DIFF
--- a/scripts/subscribe-lowercase-hashes.js
+++ b/scripts/subscribe-lowercase-hashes.js
@@ -11,8 +11,7 @@ const getSha1 = require("../sha1-utils");
 
 
 async function subscribeLowercaseHashToHIBP(emailAddress) {
-  const lowerCasedEmail = emailAddress.toLowerCase();
-  const lowerCasedSha1 = getSha1(lowerCasedEmail);
+  const lowerCasedSha1 = getSha1.getSha1ForHIBP(emailAddress);
   await HIBP.subscribeHash(lowerCasedSha1);
   return lowerCasedSha1;
 }
--- a/sha1-utils.js
+++ b/sha1-utils.js
@@ -2,8 +2,23 @@
 
 const crypto = require("crypto");
 
-function getSha1(email) {
-  return crypto.createHash("sha1").update(email).digest("hex");
+// Email addresses must be lowercased before hashing
+// Hashes sent to HIBP should be uppercased
+function getSha1ForHIBP (email) {
+  const formattedEmail = email.toLowerCase();
+  const emailHash = crypto.createHash("sha1").update(formattedEmail).digest("hex");
+  return emailHash.toUpperCase();
 }
 
-module.exports = getSha1;
+// Email addresses must be lowercased before hashing
+// Hashes that are stored in our DB should be lowercased
+// The "crypto" hash library returns hashes with lowercase letters by default
+function getLowerCaseSha1(email) {
+  const formattedEmail = email.toLowerCase();
+  return crypto.createHash("sha1").update(formattedEmail).digest("hex");
+}
+
+module.exports = {
+  getSha1ForHIBP,
+  getLowerCaseSha1,
+};
--- a/controllers/oauth.js
+++ b/controllers/oauth.js
@@ -65,7 +65,7 @@ async function confirmed(req, res, next, client = FxAOAuthClient) {
     let unsafeBreachesForEmail = [];
 
     unsafeBreachesForEmail = await HIBP.getBreachesForEmail(
-      sha1(email),
+      sha1.getSha1ForHIBP(email),
       req.app.locals.breaches,
       true,
     );
--- a/controllers/scan.js
+++ b/controllers/scan.js
@@ -61,7 +61,7 @@ async function post (req, res) {
     }
   }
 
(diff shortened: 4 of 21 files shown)


sha1-utils.js
```diff
@@ -2,8 +2,23 @@
 
 const crypto = require("crypto");
 
-function getSha1(email) {
-  return crypto.createHash("sha1").update(email).digest("hex");
+// Email addresses must be lowercased before hashing
+// Hashes sent to HIBP should be uppercased
+function getSha1ForHIBP (email) {
+  const formattedEmail = email.toLowerCase();
+  const emailHash = crypto.createHash("sha1").update(formattedEmail).digest("hex");
```
pdehaan: Silly Q: Thoughts on replacing some of this w/ `return getLowerCaseSha1(email).toUpperCase()`? and keeping it somewhat DRYer?

scripts/subscribe-lowercase-hashes.js
```diff
@@ -11,8 +11,7 @@ const getSha1 = require("../sha1-utils");
 
 
 async function subscribeLowercaseHashToHIBP(emailAddress) {
-  const lowerCasedEmail = emailAddress.toLowerCase();
-  const lowerCasedSha1 = getSha1(lowerCasedEmail);
+  const lowerCasedSha1 = getSha1.getSha1ForHIBP(emailAddress);
```
pdehaan: Q: the variable name is `lowerCasedSha1`, but I believe that `getSha1ForHIBP()` returns an uppercased string.

pdehaan:
Another interesting fact, it doesn't [currently] look like any files use both the `.getSha1ForHIBP()` _and_ `.getLowerCaseSha1()` APIs.

We could potentially do something like:

```js
const sha1 = require("../sha1-utils").getSha1ForHIBP;
```

Which just pulls the single method we need and aliases it to `sha1()`.

**PROS:** The PR diff would be minimal.
**CONS:** Makes it a bit less obvious what the code is doing I suppose, and would make it a bit more awkward if in some places we did eventually need to use both the `getSha1ForHIBP()` and `getLowerCaseSha1()` methods in the same file.

I'm just thinking out loud here, don't feel like this is a good, or even mediocre solution.

@groovecoder Have any other ideas or suggestions here?

groovecoder:
This is a great and thorough audit! I had a bunch of line-by-line comments, but I think a summary here is better ...

@pdehaan - I wouldn't want to use `const sha1 = require("../sha1-utils").[getSha1ForHIBP|getLowerCaseSha1]` - it seems like another new point where confusion or accidents could creep into the code.

This PR changes all of the _calling code_ to use explicit `getLowerCasedSha1()` and `getSha1ForHIBP()` functions, but there's probably a more robust way to enforce the 2 major rules we have:

* For HIBP, upper-case the hash before sending API calls
* For database, lower-case the hash before writing or reading

It's probably better to also enforce the proper casing of hashes at the HIBP & DB touch-points.

E.g., the `hibp` module has just 2 functions that actually make use of the hash: `getBreachesForEmail` and `subscribeHash`. If we enforce `toUpperCase` in those functions, we can be sure the hash will be upper-cased before it goes to HIBP, even if we forget to call `getSha1ForHIBP` in the calling code first. (If we send an upper-case hash, there's no harm in calling `toUpperCase` on it again.)

Similarly, the `DB` module has 9 places that actually reference the `subscribers.primary_sha1` or the `email_addresses.sha1` fields. If we enforce `toLowerCase` in those functions, we can be sure the hash will be lower-cased before it goes into the DB, even if we forget to call `getLowerCasedSha1` in the calling code first. (If we send a lower-case hash, there's no harm in calling `toLowerCase` on it again.)

lesleyjanenorton:
Would it be too much of a nightmare to uppercase the hashes in our db and then simply add `.toUpperCase()` in our existing sha1 function?

scripts/subscribe-lowercase-hashes.js
lesleyjanenorton: It's totally confusing. I think the `lowerCased` bit refers to the letters of the email address being lowercased before being encrypted.

groovecoder:
It wouldn't be technically hard. I think it would simply be:

```sql
UPDATE subscribers set primary_sha1 = upper(primary_sha1);
UPDATE email_addresses set sha1 = upper(sha1);
```

But it's a bit risky to try to update all the data in the entire DB. 😐

groovecoder:
Not planning to merge this.

PR #1499, opened 2020-01-22
fix #1453: stop adding PDL to Remote Settings

### LINKED ISSUE #1453: Update load script to skip breaches without domains
The breach script keeps trying to load the PDL breach into Remote Settings. Need to update it to skip breaches without domains.

groovecoder:


hibp.js
```diff
@@ -145,7 +145,7 @@ const HIBP = {
                 !breach.IsSpamList &&
                 !breach.IsFabricated &&
                 breach.IsVerified &&
-                !(breach.Domain === "" && breach.Name !== "PDL")
+                !(breach.Domain === "")
```
pdehaan: nit: Should this just be `breach.Domain !== ""` instead of the extra step of `===` and then negating it with `!`?

Issue #1514, opened 2020-01-27
Add missing Tout breach and DailyObjects logos


pdehaan:
- https://monitor.firefox.com/breach-details/Tout
- https://monitor.firefox.com/breach-details/DailyObjects

```sh
npx pdehaan/blurts-missing-logos -p

[
  {
    "logo": "https://monitor.firefox.com/img/logos/DailyObjects.png",
    "details": "https://monitor.firefox.com/breach-details/DailyObjects",
    "status": 404
  },
  {
    "logo": "https://monitor.firefox.com/img/logos/Tout.png",
    "details": "https://monitor.firefox.com/breach-details/Tout",
    "status": 404
  }
]
```

```sh
npx pdehaan/blurts-kinto

Missing breaches
[
  {
    "Name": "DailyObjects",
    "Title": "DailyObjects",
    "Domain": "dailyobjects.com",
    "BreachDate": "2018-01-01",
    "AddedDate": "2020-01-28T10:50:00Z",
    "ModifiedDate": "2020-01-28T10:50:00Z",
    "PwnCount": 464260,
    "Description": "In approximately January 2018, a collection of more than 464k customer records from the Indian online retailer <a href=\"https://www.dailyobjects.com/\" target=\"_blank\" rel=\"noopener\">DailyObjects</a> were leaked online. The data included names, physical and email addresses, phone numbers and &quot;pincodes&quot; stored in plain text. After multiple attempts to contact them, DailyObjects responded and received a copy of the data for verification, however failed to respond to multiple contact attempts following that.",
    "LogoPath": "https://haveibeenpwned.com/Content/Images/PwnedLogos/DailyObjects.png",
    "DataClasses": [
      "Email addresses",
      "Names",
      "Passwords",
      "Phone numbers",
      "Physical addresses"
    ],
    "IsVerified": true,
    "IsFabricated": false,
    "IsSensitive": false,
    "IsRetired": false,
    "IsSpamList": false
  },
  {
    "Name": "Tout",
    "Title": "Tout",
    "Domain": "tout.com",
    "BreachDate": "2014-09-11",
    "AddedDate": "2020-01-25T06:12:28Z",
    "ModifiedDate": "2020-01-25T06:12:28Z",
    "PwnCount": 652683,
[...]

PR #1515, opened 2020-01-27
Add job-applications data class

pdehaan:
Ref: https://github.com/mozilla/blurts-server/issues/1514#issuecomment-578874307

See https://monitor.firefox.com/breach-details/EuropaJobs

### DIFF
--- a/locales/en/data-classes.ftl
+++ b/locales/en/data-classes.ftl
@@ -63,6 +63,7 @@ imsi-numbers = IMSI numbers
 income-levels = Income levels
 instant-messenger-identities = Instant messenger identities
 ip-addresses = IP addresses
+job-applications = Job applications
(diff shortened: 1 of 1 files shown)


locales/en/data-classes.ftl
```diff
@@ -63,6 +63,7 @@ imsi-numbers = IMSI numbers
 income-levels = Income levels
 instant-messenger-identities = Instant messenger identities
 ip-addresses = IP addresses
+job-applications: Job applications
```
flodolo: This needs an equal sign ;-)

pdehaan: Bah, sorry. I've been playing w/ YAML all morning and am asleep at the wheel.
Never trust a QA to do a Dev's work. Fixed.

Issue #275, opened 2018-08-22
Refactor AppConstants to more gracefully handle missing env vars


groovecoder:
```throw new Error(`Required environment variable was not set: ${v}`);``` makes spinning up a new environment very flaky. 😢 We may want to relax it a bit.

pdehaan:
+1 to switching to something like [**convict**](https://www.npmjs.com/package/convict) to handle env vars and configs.

EMMLynch:
Closing since we've redesigned the site and functionality since this was created. If you feel that this is still needed, please let me know.

PR #1538, opened 2020-02-05
for #1469: add LOGOS_ORIGIN to serve breach logos

### LINKED ISSUE #1469: Automate acquisition and optimization of potential logo asset as much as possible
This issue is the pre-requisite for https://bugzilla.mozilla.org/show_bug.cgi?id=1608543. As much as we can automate getting the logo asset (or what might be the logo asset) from a breach, we should -- to minimize turnaround time.

groovecoder:


app-constants.js
```diff
@@ -40,7 +41,6 @@ const kEnvironmentVariables = [
   "DATABASE_URL",
   "PAGE_TOKEN_TIMER",
   "SENTRY_DSN",
-  "SERVER_URL",
```
groovecoder: this was a dupe of line 10.

views/partials/breach-groups.hbs
```diff
@@ -1,5 +1,5 @@
         {{#each firstFourBreaches }}
-          {{> breach-card cardType=./../cardType}}
+          {{> breach-card cardType=./../cardType LOGOS_ORIGIN=./../LOGOS_ORIGIN}}
```
groovecoder: This doesn't seem to be the right path to the parent context with `LOGOS_ORIGIN` in it. :/ The dashboard breach groups still render without the logos origin value.

views/partials/breach-groups.hbs
```diff
@@ -9,7 +9,7 @@
           <!-- foundBreaches 4-foundBreaches.length -->
           <div id="additional-breaches" class="show-additional-breaches hide flx">
             {{#each remainingBreaches }}
-              {{> breach-card cardType=./../cardType}}
+              {{> breach-card cardType=./../cardType LOGOS_ORIGIN=./../LOGOS_ORIGIN}}
```
groovecoder: This doesn't seem to be the right path to the parent context with `LOGOS_ORIGIN` in it. :/ The dashboard breach groups still render without the logos origin value.

views/partials/breach-groups.hbs
lesleyjanenorton: Might be easier to do this in the `makeBreachCards` helper.  Otherwise, we'll have to send the value like this everywhere there are breach cards or breach images (latest breach, scan results, breach details, dashboard, emails) which could be painful.

views/breach-detail.hbs
```diff
@@ -3,7 +3,7 @@
   <main id="breach-detail" class="breach-detail clear-header bg-white flx jst-cntr" data-page-label="Breach Detail">
     <div class="mw-700 detail-section flx-cntr flx flx-col">
       <div class="breach-detail-logo-wrapper">
-        <img class="breach-detail-logo breach-logo" alt="{{ breach.Name }} logo" src="/img/logos/{{ breach.LogoPath }}" />
+        <img class="breach-detail-logo breach-logo" alt="{{ breach.Name }} logo" src="{{ ../LOGOS_ORIGIN }}/img/logos/{{ breach.LogoPath }}" />
```
pdehaan: Unrelated to life, but should this be `{{ breach.Name }}` or `{{ breach.Title }}`?

views/partials/breach-groups.hbs
lesleyjanenorton: Might be easier to do this in the `makeBreachCards` helper.

views/partials/email_partials/email_single_breach.hbs
```diff
@@ -3,9 +3,9 @@
       <td class="email-breach-image" valign="top" style="padding-bottom: 10px; width: 40px;">
         <img class="breach-logo" 
           {{#if breachAlert }}
-            src="{{ SERVER_URL }}/img/logos/missing-logo-icon.png"
+            src="{{ LOGOS_ORIGIN }}/img/logos/missing-logo-icon.png"
```
lesleyjanenorton: would we still be using the `missing-logo-icon` here?

views/partials/breach-groups.hbs
pdehaan: I'm not an .hbs expert (but I should really be better at it by this point), but is there a way to just specify globals in Handlebars instead of trying to pass relative pointers to variables?

app-constants.js
pdehaan: Did a quick search+filter and looks like this was the only duped key, good catch!

pdehaan:
OK, I _think_ I understand the context here, but just wanted to confirm the solution vs my simple theories.
Is the plan that localhost/dev/stage would all use their respective WWW server to serve images from, but the production server would redirect to some CDN? Or would stage+production share the CDN and localhost/Heroku would serve from localhost? Or are all logos being served via a singluar CDN?

views/partials/breach-groups.hbs
groovecoder: I looked for a way to access "globals" in the partials, but couldn't find one. But it sounds like @lesleyjanenorton 's idea to move this logic into the helper function will effectively avoid the relative pointers too.

views/partials/email_partials/email_single_breach.hbs
groovecoder: If it's in the `/logos/` directory alongside the other breach logos, yup.

groovecoder:
The setup would be:

| app host (i.e., `SERVER_URL`) | breach logo host (i.e., `LOGOS_ORIGIN` |
|---|-----|
| `http://localhost:6060` | `http://localhost:6060` |
| `https://fx-breach-alerts.herokuapp.com` | `https://fx-breach-alerts.herokuapp.com` |
| `https://blurts-server-review.herokuapp.com` | `https://blurts-server-review.herokuapp.com` |
| `https://monitor-v2.herokuapp.com` | `https://monitor-v2.herokuapp.com` |
| `https://stage.firefoxmonitor.nonprod.cloudops.mozgcp.net/` | `https://monitor.cdn.mozilla.net/` |
| `https://monitor.firefox.com/` | `https://monitor.cdn.mozilla.net/` |

groovecoder:
Simplified per @lesleyjanenorton's suggestion to add a new property - `LogoUrl` - to the breach objects.

views/partials/email_partials/email_single_breach.hbs
```diff
@@ -3,9 +3,9 @@
       <td class="email-breach-image" valign="top" style="padding-bottom: 10px; width: 40px;">
         <img class="breach-logo" 
           {{#if breachAlert }}
-            src="{{ SERVER_URL }}/img/logos/missing-logo-icon.png"
+            src="{{ LogoUrl }}"
           {{else}}
-            src="{{ SERVER_URL }}/img/logos/{{ LogoPath }}"
+            src="{{ LogoUrl }}"
```
pdehaan: Looks like the `{{#if}}` and `{{else}}` blocks are the same here. Not sure if we can remove the check, or if we need different logic in one of the paths.

views/partials/breach-groups.hbs
pdehaan: FWIW, it looks like we might be able to use the `helpers` to inject some globals: https://github.com/pdehaan/express-hbs-helpers-test
Not sure if its a good or bad idea, or if it works outside of my silly little sandbox test case, but 🤷‍♂.

<kbd>/giphy "the more you know"</kbd>

views/partials/email_partials/email_single_breach.hbs
lesleyjanenorton: Well right but I thought we were going back to using the correct image path here then adding a configuration to serve `missing-logo-icon.png` only if the real logo isn't available so that `missing-logo-icon.png` doesn't appear even after we've uploaded the logo.

views/partials/email_partials/email_single_breach.hbs
lesleyjanenorton: What he said

views/partials/email_partials/email_single_breach.hbs
lesleyjanenorton: ...just noticed the yellow "Outdated" flag above this. I'll be on my way now!

.env-dist
```diff
@@ -2,6 +2,7 @@
 NODE_ENV=dev
 SERVER_URL=http://localhost:6060
 PORT=6060
+LOGOS_ORIGIN=http://localhost:6060
```
pdehaan: I guess this is only a sample .env file, but what if the default value was just an empty string?
In theory that should change all these fully qualified URLs `${AppConstants.LOGOS_ORIGIN}/img/logos/${breachCard.LogoPath}` into just `/img/logos/${breachCard.LogoPath}` which seems like the behavior we'd want for localhost/Heroku, and only the stage/production configs would need to specify an ENV var for the CDN.

Although I suspect our envs currently blow up anyways if we don't add them to each Heroku environment anyways, so maybe its moot.

groovecoder: Yeah there's a big issue (https://github.com/mozilla/blurts-server/issues/275) to refactor AppConfig to more gracefully handle missing env vars. But I'll change this to be an empty value - it works.

groovecoder:
I set my `LOGOS_ORIGIN=https://monitor.cdn.mozilla.net` and saw we needed to add the origin to our CSP. Got that working, and included the `LOGOS_ORIGIN` value in the site-wide template context so it's included correctly in the `<body data-logos-origin>` attribute.

PR #1548, opened 2020-02-12
Update breach alert and report emails -  #1395

### LINKED ISSUE #1395: Update Breach Alert Email to include "resolve this breach"
Pending our new epic for resolving breaches, we need to update the breach alert email to appropriately route folks to the appropriate page (breach detail instead of dashboard) in an authenticated state. 

Design and approach TBD

Tied to #1224

lesleyjanenorton:
Fix #1395

### DIFF
--- a/controllers/hibp.js
+++ b/controllers/hibp.js
@@ -89,7 +89,7 @@ async function notify (req, res) {
     // or from "id" property (if subscribers record)
     const subscriberId = recipient.subscriber_id || recipient.id;
     const { recipientEmail, breachedEmail, signupLanguage, preFxaSubscriber } = getAddressesAndLanguageForEmail(recipient);
-    const campaignId = (preFxaSubscriber) ? "preFXA-see-all-breaches" : "see-all-breaches";
+    const campaignId = "go-to-dashboard-link";
     const ctaHref = EmailUtils.getEmailCtaHref(utmID, campaignId, subscriberId);
 
     const requestedLanguage = signupLanguage ? acceptedLanguages(signupLanguage) : "";
--- a/middleware.js
+++ b/middleware.js
@@ -61,21 +61,37 @@ async function recordVisitFromEmail (req, res, next) {
     next();
     return;
   }
-  if (!req.query.subscriber_id || !Number.isInteger(Number(req.query.subscriber_id))) {
+
+  // Send engagement ping to FXA if req.query contains a valid subscriber ID
+  if (req.query.subscriber_id && Number.isInteger(Number(req.query.subscriber_id))) {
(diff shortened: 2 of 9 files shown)


middleware.js
```diff
@@ -61,21 +61,37 @@ async function recordVisitFromEmail (req, res, next) {
...
+    const fxaMetricsFlowPath = `metrics-flow?utm_source=${req.query.utm_source}&utm_medium=${req.query.utm_medium}${utmContent}&event_type=engage&uid=${subscriber.fxa_uid}&service=${AppConstants.OAUTH_CLIENT_ID}`;
+    const fxaResult = await FXA.sendMetricsFlowPing(fxaMetricsFlowPath);
+    log.info(`fxaResult: ${fxaResult}`);
   }
-  const subscriber = await DB.getSubscriberById(req.query.subscriber_id);
-  if (!subscriber.fxa_uid || subscriber.fxa_uid === "") {
+
+  // Redirect users who have clicked "Resolve this breach" from an email
+  // and aren't signed in to the /oauth flow.
+  if (req.query.utm_campaign && req.query.utm_campaign === "resolve-this-breach-link") {
```
groovecoder: This is true even if the user is signed-in. So, when I was signed in and clicked "Resolve this breach" in my email, I was still redirected thru FxA sign-in. Needs a check with `_getRequestSessionUser` above here.

```js
if (__getRequestSessionUser(req) !== null) {
  next();
  return;
}
```

lesleyjanenorton: Tried this and kept wheeling out trying to get to "/" when not signed in so I tried something slightly different. You now should be able to, when signed in, click "Resolve this breach" or "Go to dashboard" and not get rerouted to the `/oauth` flow.

controllers/hibp.js
```diff
@@ -89,7 +89,7 @@ async function notify (req, res) {
     // or from "id" property (if subscribers record)
     const subscriberId = recipient.subscriber_id || recipient.id;
     const { recipientEmail, breachedEmail, signupLanguage, preFxaSubscriber } = getAddressesAndLanguageForEmail(recipient);
-    const campaignId = (preFxaSubscriber) ? "preFXA-see-all-breaches" : "see-all-breaches";
+    const campaignId = "go-to-dashboard-link";
```
groovecoder: Can we change the utm params so that all of these are in the same campaign, but with different content?

E.g., on the "Go to dashboard" link ...

`utm_campaign=breach-resolution&utm_source=fx-monitor&utm_medium=email&utm_content=go-to-dashboard`

And then on the "Resolve this breach" link ...

`utm_campaign=breach-resolution&utm_source=fx-monitor&utm_medium=email&utm_content=resolve-this-breach`

PR #1551, opened 2020-02-14
CircleCI config to sync logos to CDN

oremj:


.circleci/scripts/sync-static
```diff
@@ -0,0 +1,19 @@
...
+set -o errexit -o nounset -o pipefail
+
+cd "$(dirname $0)/../.."
+
+export AWS_SECRET_ACCESS_KEY
+export AWS_ACCESS_KEY_ID
+
+aws s3 sync \
+    --acl "public-read" \
+    --cache-control "max-age=3600" \
```
groovecoder: Is this a standard `max-age` value across our services? We could probably increase this quite a bit? Or does CloudFront set its own?

oremj: We can set it to whatever you'd like.

Issue #1556, opened 2020-02-18
Automated test suite


sandysage:
Looking to Benjamin to complete the description here :)

jrbenny35:
Going to need to update the baseline images for visual regression with #1501, #1502, #1504, #1505 before April 7th, or Fenix release date.

jrbenny35:
Closed in #1557

Issue #1539, opened 2020-02-06
Add browser tests


lesleyjanenorton:
We will be adding to this list but for now we want to add browser tests to verify that:
- [ ] Plaintext emails are never passed to server
- [ ] UTM parameters are stored in FxA entrypoint elements
- [ ] UTM parameters are accurately passed to FxA on entrypoint click

EMMLynch:
Closing since we've redesigned the site and functionality since this was created. If you feel that this is still needed, please let me know.

PR #1557, opened 2020-02-18
Initial commit adding integration tests.

jrbenny35:
Works on #1539 #1556 

Adds a total of 6 tests. 3 are visual regression tests that use a base image screenshot, The others are using selenium to validate the page loads the correct data

1. ```should load correct number of breaches from an email input```: Checks the number of breaches shown matches the number.
2. ```should allow secondary email to be added```: Flow that logs in and adds an email from the user dashboard.
3. ```should load the latest breach card```: Checks breach card loads on home page


package.json
```diff
@@ -116,6 +131,9 @@
     "test:db:migrate": "NODE_ENV=tests knex migrate:latest --knexfile db/knexfile.js --env tests",
     "test:tests": "NODE_ENV=tests HIBP_THROTTLE_DELAY=1000 HIBP_THROTTLE_MAX_TRIES=3 jest --runInBand --coverage tests/",
     "test:coveralls": "cat ./coverage/lcov.info | coveralls",
+    "test:integration": "./node_modules/.bin/wdio tests/integration/wdio.conf.js",
```
pdehaan: nit: Typically, npm scripts will add `./node_modules/.bin/` to the current `$PATH`, so you might be able to just call `"wdio tests/integration/wdio.conf.js"` without the prefixes (here and on L135+136 below)

tests/integration/pages/desktop/dashboard.page.js
```diff
@@ -0,0 +1,22 @@
+/* global $ */
```
pdehaan: Was this just to appease the ESLint gods, or was this needed for something else?

tests/integration/pages/desktop/home.page.js
```diff
@@ -0,0 +1,31 @@
...
+  waitForPageToLoad() {
+    this.monitorLogo.waitForExist(5000);
+    return this;
+  }
+
+  get monitorLogo() { return $(".fx-monitor-logotype"); }
+  get breachCard() { return new BreachCard(); }
+  get breachEmailAddress() { return $("#scan-email"); }
+  get checkBreachesButton() {
+    $(".input-group-button > input:nth-child(1)").click();
```
pdehaan: n-th child!
Not sure if we want to put a more convenient class/handle on that submit button to make this any easier on you.
Or not sure if we can abuse the system with something like:

```js
$(".input-group-button > input[type='submit']).click(); // if this even works
```

I'd be reluctant to bend it too far and try reading the button's label, since that will change with different locales (ie: `input[value='Check for Breaches']` feels a bit risky risky).

tests/integration/pages/desktop/scanResults.page.js
```diff
@@ -0,0 +1,19 @@
...
+
+class ScanResultsPage {
+
+  waitForPageToLoad() {
+    $(".scan-results").waitForExist(5000);
+    return this;
+  }
+
+  get numberOfBreaches() {
+      return $(".headline > span:nth-child(1)").getText();
```
pdehaan: :+1:

tests/integration/tests/test-create-baseline.js
```diff
@@ -0,0 +1,17 @@
...
+describe("Firefox Monitor homepage", function() {
+    this.retries(2);
+
+    beforeEach(function() {
+      browser.url("http://localhost:6060/security-tips");
+    });
+
+    it("should look like normal", function() {
+      browser.saveFullPageScreen("SecurityTips_Page", { /* some options*/ });
+      expect(browser.checkFullPageScreen("SecurityTips_Page", {})).to.equal(0);
```
pdehaan: Should this be `.to.be.within(0, 5)` (instead of `.to.equal(0)`)?

tests/integration/tests/test-home-page.js
```diff
@@ -0,0 +1,80 @@
...
+    homePage.waitForPageToLoad();
+    navBar.signIn.click();
+    // FxA login
+    // Check if fxa email is needed, if error, just continue
+    try {
+      $(".email").waitForExist(5000);
+      $(".email").click();
+      $(".email").setValue(global.primaryEmail);
+      $("#submit-btn").click();
+    } catch (error) {
```
pdehaan: Curious, why would this error?

tests/integration/regions/navbar.region.js
```diff
@@ -0,0 +1,11 @@
+/* global $ */
+
+"use strict";
+
+class NavBar {
+	/* Represents the navbar */
+
+	get signIn() { return $("#sign-in-btn"); }
```
pdehaan: Just noticed this file uses tabs. Not sure if ESLint was catching that, or if we maybe need to tweak the ESLint config (which is well outside the scope of this PR, but I'm just thinking out loud).

tests/integration/wdio.conf.js
```diff
@@ -0,0 +1,273 @@
...
+     * @param {Array.<Object>} capabilities list of capabilities details
+     * @param {Array.<String>} specs List of spec file paths that are to be run
+     */
+    before: function (capabilities, specs) {
+      const chai = require("chai");
+
+      global.expect = chai.expect;
+      chai.Should();
+
+      global.baseUrl = "http://localhost:6060";
```
pdehaan: Hhmm. this is defined here, and on L95 above:

```js
baseUrl: "http://localhost:6060",
```

Not sure if they are the same thing, or can be combined or extracted into a single variable that is shared between them.

tests/integration/tests/test-breaches-page.js
```diff
@@ -0,0 +1,15 @@
+/* global browser */
+
+"use strict";
+
+describe("Firefox Monitor Breaches Page", function() {
+  this.retries(2);
+
+  beforeEach(function() {
+    browser.url("http://localhost:6060/breaches");
```
pdehaan: Since we already define the `baseUrl` in the configs, do we need the fully qualified URL here? Or can this just be a vague "/breaches", which infers the domain from the config?

tests/integration/tests/test-create-baseline.js
pdehaan: Oh, or should this use `0` since we're trying to verify the screenshot that we just took on L14 above.?

tests/integration/tests/test-home-page.js
```diff
@@ -0,0 +1,80 @@
...
+
+"use strict";
+
+const assert = require("assert");
+const HomePage = require("../pages/desktop/home.page");
+const NavBar = require("../regions/navbar.region");
+const UserDashboardPage = require("../pages/desktop/dashboard.page");
+
+describe("Firefox Monitor homepage", function() {
+  //this.retries(2);
```
pdehaan: Should this be uncommented (like in L7 of previous file)?

scripts/wait-for-it.sh
```diff
@@ -0,0 +1,182 @@
+#!/usr/bin/env bash
```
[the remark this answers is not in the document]
pdehaan: OK, after marinating on this overnight and giving it a few Bing searches, what about something like this: https://www.npmjs.com/package/wait-on

> wait-on is a cross-platform command line utility which will wait for files, ports, sockets, and http(s) resources to become available (or not available using reverse mode). Functionality is also available via a Node.js API. Cross-platform - runs everywhere Node.js runs (linux, unix, mac OS X, windows)

I've never used it in a project yet, but 🤷‍♂.

Our current implementation looks something like this:

```yaml
command: ["./scripts/wait-for-it.sh", "postgres:5432", "--", "npm", "run", "db:migrate"]
```

Not sure if that would be similar to:
```yaml
command: ["wait-on", "tcp:5432", "&&", "npm", "run", "db:migrate"]
```

Or if we just add another npm script in package.json which does "ci:db:migrate" so the command looks less scary.

tests/integration/pages/desktop/dashboard.page.js
jrbenny35: For eslint. The ```$``` is needed for wdio though

tests/integration/tests/test-create-baseline.js
jrbenny35: Ah this test shouldn't be added to the repo. I'll remove it

tests/integration/tests/test-home-page.js
jrbenny35: Sometimes the email doesn't show as monitor passes the email you used for checking breaches into the fxa sign in

tests/integration/regions/navbar.region.js
jrbenny35: Wrong size tabs?

tests/integration/wdio.conf.js
jrbenny35: Not the same, I'll remove the wrong one.

tests/integration/tests/test-breaches-page.js
jrbenny35: Well the baseurl doesn't do much. I could just do ```browser.url(global.baseUrl)```. There's not really any standard on doing baseurl with wdio.

.circleci/config.yml
```diff
@@ -49,6 +49,23 @@ jobs:
...
+            - run:
+                name: Update npm to latest
+                command: docker run blurts-server npm install npm@latest -g
+            - run:
+                name: Build image and run Intergration Test
+                command: |
+                  cp .env-dist .env
+                  npm install
+                  npm install --only=dev
+                  docker-compose -f tests/interation/docker-compose.yml up --build -d
```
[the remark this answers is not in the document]
jrbenny35: yep :sweat_smile:

scripts/wait-for-it.sh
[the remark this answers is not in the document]
jrbenny35: Not sure what would be best here. I am open to suggestions.

tests/integration/tests/test-home-page.js
```diff
@@ -0,0 +1,80 @@
...
+      $(".email").setValue(global.primaryEmail);
+      $("#submit-btn").click();
+    } catch (error) {
+      // eslint-disable-next-line no-console
+      console.log(error);
+    }
+    // FxA password login
+    $("#password").waitForExist(5000);
+    $("#password").click();
+    $("#password").setValue(global.mailinatorPassword);
```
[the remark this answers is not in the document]
jrbenny35: Good point!

tests/integration/wdio.conf.js
```diff
@@ -0,0 +1,273 @@
...
+      global.primaryEmail = "test@mailinator.com";
+      global.secondaryEmail = "test" + Math.random() + "@mailinator.com";
+      global.mailinatorPassword = process.env.MAILINATOR_PASSWORD || "a_secure_password";
+      browser.setWindowSize(1920, 1080);
```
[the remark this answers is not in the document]
jrbenny35: For mobile I'll have to set a screensize as well. We could maybe have another ```*.conf.js``` for mobile that has that set differently.

tests/integration/pages/desktop/home.page.js
jrbenny35: Classes on everything would make it easier but I know that is boring so..this works, and if it gets updated it's an easy change

tests/integration/regions/navbar.region.js
jrbenny35: Wrong size indeed. Darn Visual Studio

tests/integration/tests/test-breaches-page.js
jrbenny35: Well I guess I should read. Seems they have baseUrl support but maybe it didn't work before. I was probably doing it wrong..

jrbenny35:
> I'm not sure the Breaches_Page baselines will really work in the long run. It seems like any recent breach will cause the items to reorder and we'll need to regen the baselines after each freshly added breach. Possibly multiple times after we upload a new breach logo to the CDN.
> 
> Even the Home_Page baseline has a recent breach card. Not sure how viable it would be to hide that DOM element before taking/comparing images, or if that would defeat the purpose and reduce confidence (less brittle test that doesn't test a core thing). man_shrugging

I wonder how exact it is. I think it is worth having but if it comparing words, then yeah, that will be a problem. That is why I wanted to use a range. It calculates a percentage based on how "off" the page is.

pdehaan:
> It calculates a percentage based on how "off" the page is.

AAAHHHH... OK, my misunderstanding. The tool we used at a different company measured by _pixels_ so we could only have 0-5 pixels different from a baseline, not 0-5%. I say we run with it for a bit and see how it works in the long run.

jrbenny35:
> > It calculates a percentage based on how "off" the page is.
> 
> AAAHHHH... OK, my misunderstanding. The tool we used at a different company measured by _pixels_ so we could only have 0-5 pixels different from a baseline, not 0-5%. I say we run with it for a bit and see how it works in the long run.

I agree. If the words make up for 10% or less, I think we are still good (as we discussed on slack).

tests/integration/tests/test-breaches-page.js
pdehaan: I am not a Selenium/wdio expert, but my understanding was that if you passsed a relative URL, it'd use the `baseUrl` from the config. So basically, could we use `browser.url("/breaches")` and have it infer the `baseUrl` from the config. But really, I should be pulling the PR locally and testing that theory.

tests/integration/pages/desktop/dashboard.page.js
pdehaan: Ah, OK. You can also define [per-glob] globals in the ESLint config file (.eslintrc.js in root of project) instead of needing to define in each template.

For example, in the .eslintrc.js `overrides:` section, you can tweak the tests/**/*.js glob (around L32) and inject your `$`, `$$`, and `browser` globals there &mdash; if that is easier.

```diff
    {
      files: [
        'tests/**/*.js',
      ],
      env: {
        jest: true,
      },
+     globals: {
+       "$": "readonly",
+       "$$": "readonly",
+       "browser": "readonly",
+     },
    },
```

tests/integration/wdio.conf.js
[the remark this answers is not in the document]
pdehaan: @groovecoder did you have a specific device/viewport in mind?
Somewhere between an Alcatel Go Flip 2 and a Samsung Galaxy S20 Ultra 5G Turbo: Hyper Fighting edition?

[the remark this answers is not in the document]
jrbenny35: I usually use an iPhone 'Plus' model

tests/integration/tests/test-home-page.js
jrbenny35: I am thinking retries aren't so bad, as there are multiple things that could go wrong. What do you think @pdehaan

pdehaan: :+1: on retries, and 2 is a good number.

tests/integration/wdio.conf.js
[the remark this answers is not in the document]
pdehaan: Looks like the latest iPhone [11 Pro Max], has the following viewport:

```js
    'viewport': {
      'width': 414,
      'height': 896,
      'deviceScaleFactor': 3,
      'isMobile': true
    }
```

&mdash; per the [**playwright** deviceDescriptors file](https://github.com/microsoft/playwright/blob/33824aa1d748a82f11f53b027018b5facf1d8cd3/src/deviceDescriptors.ts#L441-L450).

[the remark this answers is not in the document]
pdehaan: > For mobile I'll have to set a screensize as well. We could maybe have another `*.conf.js` for mobile that has that set differently.

I think all that would need to change is the `browser.setWindowSize(w, h)` dimensions, correct?
Wondering if we could convert that into ENV vars (maybe with a desktop fallback). That way we wouldn't need any new config files and we could possibly hack something like this:

```json
"test:integration:mobile": "BROWSER_WIDTH=414 BROWSER_HEIGHT=896 wdio tests/integration/wdio.conf.js",
```

Then, above we'd just need to change to this psesudo-code mess:
```js
// NOTE: Env vars come in as strings, iirc, so we cast them to numbers here.
browser.setWindowSize(
    Number(process.env.BROWSER_WIDTH) || 1920,
    Number(process.env.BROWSER_HEIGHT) || 1080
);
```

tests/integration/pages/desktop/dashboard.page.js
jrbenny35: Well that's pretty neat!

tests/integration/wdio.conf.js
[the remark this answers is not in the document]
jrbenny35: This is good thinking. @groovecoder thoughts on this?

scripts/wait-for-it.sh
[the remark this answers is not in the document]
pdehaan: @jrbenny35 and I were noodling on this a bit, and we might be able to remove the +x flag on the .sh file and just invoke this script using <kbd>bash ./wait-for-it.sh</kbd>. Testing locally, that seems to still work in a simple test case even though the file is non-executable.

pdehaan:
From @jrbenny35 in a Slack channel:

> there is a new breach added to monitor, the visual tests failed with 6.55% diff

jrbenny35:
> From @jrbenny35 in a Slack channel:
> 
> > there is a new breach added to monitor, the visual tests failed with 6.55% diff

The homepage failed with a 5.8%ish diff and the "Breaches" page failed with 6.55%. If this keeps gets worse with more breaches, we will have to revist this and maybe create a better baseline.

pdehaan:
> If this keeps gets worse with more breaches, we will have to revisit this and maybe create a better baseline.

Yeah, I can imagine the /breaches/ endpoint will be a hot mess(tm). Currently the breaches are sorted by the breach date (which isn't a displayed field, making the ordering appear completely random). So if a breach is new added to the site, it could be displayed first (invalidating the order of all 15 items), or could appear 8th in the list (invalidating only 7-8 of the 15 displayed items), or could be displayed as the 200th most recent breach, which is then hidden behind the "Show All" button, and may not even appear in the image diff (depending on whether you baseline before or after hitting the "Show All" button).

So the fact the "Slickwraps" breach is currently the lastest breach date and date added (putting it in the first slot of the /breaches/ page), seems to be our worst case scenario -- image-diff-wise speaking.

tests/integration/tests/test-home-page.js
```diff
@@ -0,0 +1,76 @@
...
+
+  beforeEach(function() {
+    browser.url("/");
+  });
+
+  it("should load the latest breach card", function() {
+    const homePage = HomePage;
+
+    homePage.waitForPageToLoad();
+    assert(
```
pdehaan: @jrbenny35 This seems to be the only instance of `assert()` I can find. Should this use chai's `expect()`?

jrbenny35: Good find!

jrbenny35:
If you want to try these locally you will need to follow these instructions:
1. Copy ```.env-dist``` to ```.env```.
2. Set the following env variables: ```FXA_ENABLED=1, HIBP_KANON_API_TOKEN=token123, HIBP_API_TOKEN=token123, MONITOR_FXA_PASSWORD=password```. Replace the values in "token123" and "password" with the correct ones.
3. In the same shell instance, start the server and build the docker image ```docker-compose -f tests/integration/docker-compose.yml up --build -d```.
4. Run the tests: ```npm run test:integration```.

tests/integration/tests/test-create-baseline.js
```diff
@@ -0,0 +1,27 @@
...
+*/
+
+"use strict";
+
+
+describe("Firefox Monitor homepage", function() {
+    this.retries(2);
+
+    beforeEach(function() {
+      browser.url("http://localhost:6060/");
```
pdehaan: Should this be localhost:6060 or just `/` (since `baseUrl` is `baseUrl: "http://localhost:6060",`)?

tests/integration/wdio.conf.js
```diff
@@ -0,0 +1,271 @@
...
+    //
+    // If you only want to run your tests until a specific amount of tests have failed use
+    // bail (default is 0 - don't bail, run all tests).
+    bail: 0,
+    //
+    // Set a base URL in order to shorten url command calls. If your `url` parameter starts
+    // with `/`, the base url gets prepended, not including the path portion of your baseUrl.
+    // If your `url` parameter starts without a scheme or `/` (like `some/path`), the base url
+    // gets prepended directly.
+    baseUrl: "http://localhost:6060",
```
pdehaan: Interesting... Currently in our .env-dist, we have an env var for the current `SERVER_URL`:

```ini
```

Wonder if we could do `baseUrl: process.env.SERVER_URL,` and if that means tests could run against stage/production, or if that is just silly and I'm mixing different ideas/goals.

tests/integration/tests/test-create-baseline.js
jrbenny35: Should just be ```/```

tests/integration/wdio.conf.js
jrbenny35: Well I am treating the server as a black box. it _should_ work against stage or prod

tests/integration/README.md
```diff
@@ -0,0 +1,22 @@
...
+- [geckodriver]
+- [npm]
+- [nodejs]
+
+## Firefox Setup
+
+Make sure Firefox is in your system path if you want to watch the tests run inside your browser.
+
+1. Follow setup instructions [here](https://github.com/mozilla/blurts-server#install).
+2. Set appropriate ENV variables, HIBP_API_TOKEN, HIBP_KANON_API_TOKEN, MALINATOR_PASSWORD
```
pdehaan: typo: should this be `MAILINATOR_PASSWORD` (missing first "I")?

tests/integration/README.md
```diff
@@ -0,0 +1,22 @@
...
+2. Build and start the containers: ```docker-compose -f tests/integration/docker-compose.yml up --build -d```.
+3. Run the tests: ```npm run test:integration```.
+   They can also be run headlessly: ```npm run test:integration-headless```, and inside docker: ```npm run test:integration-docker```.
+
+[geckodriver]: https://github.com/mozilla/geckodriver/releases
```
pdehaan: Ah, interesting! looks like geckodriver is on npm too: https://www.npmjs.com/package/geckodriver (by our very own vladikoff!)

tests/integration/wdio.docker.js
```diff
@@ -23,7 +23,11 @@ exports.config = Object.assign({}, common.config, {
       startDelay: 2000,
     },
     options: {
-      e: ["MOZ_HEADLESS=1", "MALINATOR_PASSWORD=${MAILINATOR_PASSWORD}"],
+      e: ["MOZ_HEADLESS=1",
+          "MALINATOR_PASSWORD=${MAILINATOR_PASSWORD}",
```
pdehaan: same typo here? "MALINATOR" missing first "I"

tests/integration/README.md
jrbenny35: Should I link that instead?

pdehaan: No, sorry, that was just me thinking out loud and bookmarking it for future reference. It'd only make sense to switch if we added it as a project dependency in package.json and got rid of the explicit dependency of having it preinstalled. But I'd have to audit the npm install to see how big that bundle is. In fact, I can do that right now.

Looks like running <kbd>npm i geckodriver -D</kbd> in a fresh directory adds about 9.7 MB of node_modules goodness, including the following files in _./node_modules/geckodriver/_:

```
-rwxr-xr-x   5.0M 10 Oct 15:53 geckodriver
-rw-r--r--   1.9M  5 Mar 10:26 geckodriver.tar.gz
```

But let's just leave this as-is for now and can refactor in the future after this lands.

jrbenny35: Good info to have though!

jrbenny35:
@groovecoder @lesleyjanenorton would love to land this this week

tests/integration/docker-compose.yml
```diff
@@ -0,0 +1,32 @@
...
+  postgres:
+    image: postgres
+    restart: always
+    environment:
+      - POSTGRES_USER=postgres
+      - POSTGRES_DB=blurts
+    expose:
+      - "5432"
+  migration:
+    image: blurts-server_server:latest
```
lesleyjanenorton: Just following up on our Slack thread... Should we change this line to `integration_server:latest` ?

tests/integration/docker-compose.yml
```diff
@@ -0,0 +1,32 @@
+version: "3.7"
+services:
+  server:
+    build:
+      context: ../..
+      dockerfile: tests/integration/Dockerfile.integration-test.yml
+    ports:
+      - "6060:6060"
```
lesleyjanenorton: Can we make this use the SERVER_URL environment var?

jrbenny35:
@groovecoder @lesleyjanenorton any updates?

pdehaan:
@groovecoder @lesleyjanenorton Do either of you have any bandwidth to assist with this? We've rebased and are still seeing the test failure on Travis-CI (https://travis-ci.org/github/mozilla/blurts-server/jobs/666489293?utm_medium=notification&utm_source=github_status) but I'm still unclear if it's related to our PR since it doesn't look like we've touched any of the failing code. Although @jrbenny35 rebased and it looks like we should be in sync w/ latest changes to **master** branch.

I'm at a bit of a loss.

.eslintrc.js
```diff
@@ -35,7 +35,12 @@ module.exports = {
       ],
       env: {
         jest: true,
-      }
+      },
+      globals: {
+        "$": "readonly",
+        "$$": "readonly",
+        "browser": "readonly",
```
groovecoder, line 42: Can we mark these as `globals` only in the `tests/integration/` directory?

.env-dist
```diff
@@ -41,19 +41,20 @@ BASKET_NEWSLETTER=mozilla-and-you
 # leave FXA_ENABLED empty to disable FXA
 FXA_ENABLED=
 FXA_SETTINGS_URL="https://stable.dev.lcip.org/settings"
-OAUTH_CLIENT_ID=cb1bef9d06bb9bc9
-OAUTH_CLIENT_SECRET=f5fb99de6e0af18ab17e013ac1d439903179a97a1c510fc10bc3bd50bbce089b
+OAUTH_CLIENT_ID=edd29a80019d61a1
+OAUTH_CLIENT_SECRET=a80feaad77c847275d39ac989ee12a873ef6b54cbc184128f86f2afecdf003b5
```
pdehaan, line 45: I assume these are OK to change? (I know it's just a dist file, but wanted to confirm these aren't critical creds that need to be regenerated now).

.env-dist
```diff
@@ -41,19 +41,20 @@ BASKET_NEWSLETTER=mozilla-and-you
...
 # HIBP API for breach data
-HIBP_API_ROOT="https://stage.haveibeenpwned.com/api/v2/"
 # How many seconds to wait before refreshing upstream breach data from HIBP
 HIBP_RELOAD_BREACHES_TIMER=600
 # HIBP API for range search and subscription
 HIBP_KANON_API_ROOT="https://api.haveibeenpwned.com"
-HIBP_KANON_API_TOKEN=""
+HIBP_KANON_API_TOKEN=
+HIBP_API_ROOT="https://haveibeenpwned.com/api/v2"
+HIBP_API_TOKEN=
```
pdehaan, line 57: Are these OK to be "naked", or should they be empty strings, a la `=""`?

tests/integration/docker-compose.yml
```diff
@@ -16,18 +16,40 @@ services:
       - LOGOS_ORIGIN=null
       - BREACH_RESOLUTION_ENABLED=false
```
pdehaan, line 17: Curious, should this be "true"? I think breach resolutions are enabled in production. Curious what effect this would have, apart from the logged in admin being slightly different.

tests/integration/docker-compose.yml
```diff
@@ -16,18 +16,40 @@ services:
...
       - MONITOR_FXA_PASSWORD
+      - DATABASE_URL=postgres://postgres@postgres:5432/blurts
```
pdehaan, line 19: I was curious if you could recycle the environment's `DATABASE_URL` variable, but i think this is subtlety different than the .env-dist file above, which has:

```ini
DATABASE_URL="postgres://postgres@localhost:5432/blurts"
```
(an `@localhost:5432` instead of `@postgres:5432`).

tests/integration/pages/desktop/scanResults.page.js
```diff
@@ -6,7 +6,7 @@
 class ScanResultsPage {
 
   waitForPageToLoad() {
-    $(".scan-results").waitForExist(5000);
+    $(".scan-results").waitForExist(100000);
```
pdehaan, line 9: This almost feels like a big leap. I'll assume ms here, since the previous value of "5000" ms is 5 seconds, which meaning this timeout is now 100000ms or 100 seconds or 1.6 minutes, although I concede the screenshots are pretty big, so maybe it takes a bit for the UI to fully bake.

tests/integration/wdio.docker.js
```diff
@@ -24,7 +24,7 @@ exports.config = Object.assign({}, common.config, {
     },
     options: {
       e: ["MOZ_HEADLESS=1",
-          "MAILINATOR_PASSWORD=${MAILINATOR_PASSWORD}",
+          "MAILINATOR_PASSWORD=${MONITOR_FXA_PASSWORD}",
```
pdehaan, line 27: And I _think_ this is OK here? Seems almost like unrelated passwords (mailinator disposable passwords vs a potentially sensitive MONITOR_FXA_PASSWORD)? But considering the following two API keys are probably much more secret, I'm guessing there isn't a way for anybody to ssh or pull a docker image and get those secrets and be naughty(tm).

pdehaan:
Awesome work @jrbenny35, as always!
Re-reviewed the latest changes and left a few random thoughts and comments.
@groovecoder Can you give this one last 👀 before we merge this into the big code repo in the sky?

.env-dist
groovecoder: Yup, these are the values I have on my local file and it's working.

tests/integration/docker-compose.yml
groovecoder: The breach detail pages will also be different for logged-in users.

tests/integration/pages/desktop/scanResults.page.js
groovecoder: I was frequently getting errors with the 5s limit. :/

tests/integration/wdio.docker.js
groovecoder: Yeah, this password is for a test account on https://stable.dev.lcip.org/. So even if someone got this, I think the worse they could do is to DoS that account which would break our tests. :/

.eslintrc.js
pdehaan: @jrbenny35 Let me know if you need a hand with this. I think it should just be copying this one override config and creating a new one for `files: ['tests/integration/**/*.js']` with our fancy globals.

tests/integration/pages/desktop/scanResults.page.js
jrbenny35: I did this for testing...thanks for catching it.

tests/integration/docker-compose.yml
jrbenny35: Yes different because docker-compose. Using ```@postgres``` points to the postgres container

.env-dist
jrbenny35: Yes, according to, https://www.npmjs.com/package/dotenv

.eslintrc.js
jrbenny35: Does it have to be named specially?

tests/integration/wdio.docker.js
jrbenny35: I highly doubt anyone would crack that password.

.eslintrc.js
jrbenny35: @pdehaan ^

pdehaan: Sorry, I'm not so good with words. I think the answer here is to remove these new globals from the "tests/**/*.js" glob here, and create a new entry with:

```js
    {
      files: [
        "tests/integration/**/*.js",
      ],
      globals: {
        "$": "readonly",
        "$$": "readonly",
        "browser": "readonly",
      },
    },
```

PR #1574, opened 2020-03-02
Hide dashboard relay callout from users who have already signed up

lesleyjanenorton:


template-helpers/dashboard.js
```diff
@@ -2,6 +2,19 @@
 
 const { LocaleUtils } = require("./../locale-utils");
 const { makeBreachCards } = require("./breaches");
+const { hasUserSignedUpForRelay } = require("./../controllers/utils");
+
+function showRelayCallout(args) {
+  const enLocaleIsSupported = args.data.root.req.supportedLocales.includes("en");
```
groovecoder: `supportedLocales` always includes `en` because we specify it as our `defaultLocale`. This check needs to specifically check the `args.data.root.req.headers["accept-language"]` instead.

```suggestion
  const enLocaleIsSupported = args.data.root.req.headers["accept-language"].includes("en");
```

PR #1585, opened 2020-03-13
fix #1569: add a SKIP_LIST to updatebreaches.js

### LINKED ISSUE #1569: Turn off desktop browser notification for Factual

groovecoder:


groovecoder:
Good consideration, though in this case the point of skipping the breach is so that we don't show the pop-up on the domain. So, it's fine to filter out multiple breaches of the same domain.

PR #1602, opened 2020-03-19
Opt-in/out FxA Experiment

maxxcrawford:
## To-do: 

- [x] Add UTM params to FxA Button Click, making sure they are routed/passed through on OAuth return
- [x] Setup GA Goal/Conversion
- [x] Update PR with testing steps
- [x] Pass CI Checks
- [x] Add tests (where necessary) 
- [x] ~Confirm .FTL usage, if this experiment is only targeting EN users.~ Move .FTL strings to hardcode

## Testing Steps

To directly access a specific experiment branch, use the following query string to your testing URL: 
- Control/vA: `?experimentBranch=va`
- Opt-in/vB: `?experimentBranch=vb`
- Opt-out/vC: `?experimentBranch=vc`

To tell which branch you're in WITHOUT the query param, inspect the `body` tag data attribute: `data-experiment="{branch}"`.

### Expected results:

- Control/vA: 
image

- Opt-in/vB:
image

- Opt-out/vC: 
image


maxxcrawford:
@flodolo This experiment is only targeting EN users, but I set the string in the EN fluent file. Because of this scope, should I do something different? (Add text directly to handlebars template, specific "Don't translate" notation in Fluent file, etc.)

flodolo:
If they're not meant to be translated, please hardcode them in the template.

public/js/monitor.js
```diff
@@ -369,4 +369,24 @@ function addBentoObserver(){
...
+
+  if (document.body.dataset.experiment) {
+    const submitBtn = document.querySelector("#scan-user-email input[type='submit']");
+    const createFxaCheckbox = document.getElementById("createFxaCheckbox");
+
+    submitBtn.addEventListener("click", (e)=> {
+      if (createFxaCheckbox.checked) {
+        e.preventDefault();
+        if (typeof(ga) !== "undefined") {
+          // Send "Click" ping for #see-additional-recs click
```
lesleyjanenorton: Why are we commenting this part out?

maxxcrawford: WIP Comment. Will be fixed during first to-do item above.

controllers/home.js
```diff
@@ -13,6 +13,15 @@ async function home(req, res) {
     csrfToken: req.csrfToken(),
   };
 
+  const coinFlipNumber = Math.random() * 100;
+  const experimentBranch = getExperimentBranch(req, coinFlipNumber);
+
+  const isUserInExperiment = isExperiment(experimentBranch);
+  let experimentBranchB = false;
```
lesleyjanenorton: I think this could also just be `const experimentBranchB = (experimentBranch === "vb" && isUserInExperiment)`

controllers/home.js
```diff
@@ -13,6 +13,15 @@ async function home(req, res) {
     csrfToken: req.csrfToken(),
   };
 
+  const coinFlipNumber = Math.random() * 100;
+  const experimentBranch = getExperimentBranch(req, coinFlipNumber);
+
+  const isUserInExperiment = isExperiment(experimentBranch);
```
lesleyjanenorton: Since we're only calling `isExperiment` once, I think this could be 
`const isUserInExperiment = (experimentBranch === "vb" || experimentBranch === "vc")`

public/js/monitor.js
```diff
@@ -36,17 +36,90 @@ function isValidEmail(val) {
...
+  // If so, we set it similiar to what would happen on form submit.
+  //
+  // Options was added to limit how the watched email input field is injected
+  // in this function. It only moves it if  options.emailWatch is set to TRUE;
+
   let url = new URL("/oauth/init", document.body.dataset.serverUrl);
   url = getFxaUtms(url);
-  ["flowId", "flowBeginTime", "entrypoint"].forEach(key => {
+
+  ["flowId", "flowBeginTime", "entrypoint", "form_type"].forEach(key => {
```
lesleyjanenorton: is `form_type` always `email` ? If so, let's stick this in the oauth controller.

lesleyjanenorton: Caveat: unless `form_type` needs to also be sent in `View` pings.

lesleyjanenorton:
Not sure what's up yet but going to https://monitor-v2.herokuapp.com with no specified query params seems to default to the `vc` branch each time.

views/partials/forms/scan-email-form.hbs
```diff
@@ -15,7 +24,7 @@
     {{#if featuredBreach}}
       <input type="submit" value="{{ getString "find-out" }}"/>
     {{else}}
-      <input type="submit" value="{{ getString "check-for-breaches" }}"/>
+      <input {{#if isUserInExperiment}} {{> analytics/fxa id="fx-monitor-check-for-breaches-blue-btn" }} {{/if}} type="submit" value="{{ getString "check-for-breaches" }}"/>
```
lesleyjanenorton: It looks like the FxA `metrics-flow` values are not being added to this element, so we need to just add something like `class="add-metrics-flow-values"` after the `{{#if isUserInExperiment}}` and then add that class to the selector on L156 of `fxa-analytics.js` so that it looks  something like this:

```
document.querySelectorAll(".open-oauth, .add-metrics-flow-values").forEach( async(el) => {
```

lesleyjanenorton: It looks like the FxA `metrics-flow` values are not being added to this element, so we need to just add something like `class="add-metrics-flow-values"` after the `{{#if isUserInExperiment}}` and then add that class to the selector on L156 of `fxa-analytics.js` so that it looks  something like this:

```
document.querySelectorAll(".open-oauth, .add-metrics-flow-values").forEach( async(el) => {
```

controllers/oauth.js
```diff
@@ -23,12 +28,20 @@ function init(req, res, next, client = FxAOAuthClient) {
   const url = new URL(client.code.getUri({state}));
   const fxaParams = new URL(req.url, AppConstants.SERVER_URL);
 
+  req.session.utmContents = {};
+
+  req.session.returnTo = "/scans";
```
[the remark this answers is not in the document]
groovecoder: Wait, it looks like we're not ever using this `returnTo` value?

controllers/oauth.js
```diff
@@ -53,6 +66,16 @@ async function confirmed(req, res, next, client = FxAOAuthClient) {
   const existingUser = await DB.getSubscriberByEmail(email);
   req.session.user = existingUser;
 
+  const returnURL = new URL("/user/dashboard", AppConstants.SERVER_URL);
+
+
```
groovecoder:
```suggestion
```

controllers/oauth.js
```diff
@@ -53,6 +66,16 @@ async function confirmed(req, res, next, client = FxAOAuthClient) {
...
+  getUTMNames().forEach(param => {
+    if (req.session.utmContents[param]) {
+      returnURL.searchParams.append(param, req.session.utmContents[param]);
+      // returnURL.searchParams.append(param, encodeURIComponent(req.session.utmContents[param]));
```
groovecoder:
```suggestion
```

controllers/oauth.js
```diff
@@ -67,7 +90,7 @@ async function confirmed(req, res, next, client = FxAOAuthClient) {
     unsafeBreachesForEmail = await HIBP.getBreachesForEmail(
       sha1(email),
       req.app.locals.breaches,
-      true,
```
groovecoder: this comma in particular.

controllers/user.js
```diff
@@ -552,7 +552,16 @@ async function getBreachStats(req, res) {
...
+    // Persist experimentBranch across session reset
+    let experimentBranch = req.session.experimentBranch;
+    req.session.reset();
+    req.session.experimentBranch = experimentBranch;
+  } else {
+    req.session.reset();
+  }
+
+
   res.redirect("/");
```
groovecoder: Hmm ... I was going to say get rid of this `else` block, but I know express is a pain about `return` statements when/before using `res.*` methods.

controllers/user.js
```diff
@@ -552,7 +552,16 @@ async function getBreachStats(req, res) {
 
 
 function logout(req, res) {
-  req.session.reset();
+  if (req.session.experimentBranch) {
+    // Persist experimentBranch across session reset
+    let experimentBranch = req.session.experimentBranch;
```
groovecoder:
```suggestion
    const experimentBranch = req.session.experimentBranch;
```

views/partials/forms/scan-email-form.hbs
maxxcrawford: Should this only fire when the user is going to open a FxA OAuth flow? (Example: User unchecks box)

views/layouts/default.hbs
```diff
@@ -21,7 +21,7 @@
     <link rel="icon" href="/img/favicons/favicon-128.png" sizes="128x128" />
     <link rel="icon" href="/img/favicons/favicon-256.png" sizes="256x256" />
   </head>
-  <body {{> analytics/default_dataset }} data-bento-app-id="fx-monitor">
+  <body {{> analytics/default_dataset }} {{#if experimentBranch }} {{> analytics/experiment }} data-utm_term="{{experimentBranch}}"  data-experiment="{{experimentBranch}}" form_type="email" data-utm_content="{{#if experimentBranchB }}opt_out{{else}}opt_in{{/if}}" {{/if}} data-bento-app-id="fx-monitor">
```
groovecoder: Any way to move all this if logic "up" into the `analytics/experiment` partial?

views/partials/forms/scan-email-form.hbs
lesleyjanenorton: Yes - if the user un-checks the box, then you probably do not want to send an "Engage" ping to FxA once that form is submitted.

lesleyjanenorton: From the requirements doc, it looks like we need to add these params as well. 
```
entrypoint_experiment = growthuserflow1
entrypoint_variation = (va, vb, vc)

File: controllers/home.js (hunk 2 above)
Comment (maxxcrawford) [reply controllers/home.js]: Revised!

File: controllers/home.js (hunk 3 above)
Comment (maxxcrawford) [reply controllers/home.js]: Revised!

File: public/js/monitor.js (hunk 4 above)
Comment (maxxcrawford) [reply public/js/monitor.js]: Revised!

File: views/partials/forms/scan-email-form.hbs (hunk 5 above)
Comment (maxxcrawford) [reply views/partials/forms/scan-email-form.hbs]: Added those attributes!

File: controllers/home.js [hunk 13]
```diff
@@ -50,9 +61,50 @@ async function home(req, res) {
...
+
+function getExperimentBranch(req, sorterNum) {
+
+  if (!req.headers["accept-language"].includes("en") ){
+    return false;
+  }
+
+  // If URL param has experimentBranch entry, use that branch;
+  if (req.query.experimentBranch) {
+    if (req.query.experimentBranch !== "va" && req.query.experimentBranch !== "vb" && req.query.experimentBranch !== "vc" ) {
```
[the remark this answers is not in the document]
maxxcrawford: Revised!

controllers/home.js
```diff
@@ -50,9 +61,50 @@ async function home(req, res) {
...
+      return false;
+    } else {
+      req.session.experimentBranch = req.query.experimentBranch;
+      return req.query.experimentBranch;
+    }
```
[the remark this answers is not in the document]
maxxcrawford: Revised!

controllers/oauth.js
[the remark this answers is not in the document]
maxxcrawford: Good catch. Removing!

controllers/oauth.js
```diff
@@ -12,9 +12,14 @@ const HIBP = require("../hibp");
 const mozlog = require("../log");
 const sha1 = require("../sha1-utils");
 
-
 const log = mozlog("controllers.oauth");
 
+const utmArray = ["utm_source", "utm_medium", "utm_campaign", "utm_term", "utm_content"];
+
+function getUTMNames() {
+  return ["utm_source", "utm_medium", "utm_campaign", "utm_term", "utm_content"];
```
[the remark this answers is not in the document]
maxxcrawford: Revised.

controllers/oauth.js
maxxcrawford: Whitespace removed.

controllers/oauth.js
maxxcrawford: Line removed.

controllers/oauth.js
maxxcrawford: Was matching the linter. 🙈

controllers/user.js
maxxcrawford: Yup. Leaving as is. ☹️

controllers/user.js
maxxcrawford: Revised!

views/layouts/default.hbs
maxxcrawford: Moved!

public/js/monitor.js
```diff
@@ -36,19 +37,100 @@ function isValidEmail(val) {
...
+  // there is an email address entered into #scan-user-email form.
+  // If so, we set it similiar to what would happen on form submit.
+  //
+  // Options was added to limit how the watched email input field is injected
+  // in this function. It only moves it if  options.emailWatch is set to TRUE;
+
   let url = new URL("/oauth/init", document.body.dataset.serverUrl);
   url = getFxaUtms(url);
+
   ["flowId", "flowBeginTime", "entrypoint"].forEach(key => {
```
lesleyjanenorton: We'll want to add `entrypoint_experiment` and `entrypoint_variation` to this, otherwise those values won't make it to the OAuth flow. Since not every entrypoint element will have `entrypoint_experiment` and `entrypoint_variation` attributes we'll need to check before appending to the url though.

public/js/monitor.js
```diff
@@ -36,19 +37,100 @@ function isValidEmail(val) {
...
+      }
+    }
+    window.location.assign(url);
+    return;
+  }
+
+  // Growth Experiment: This logic is complex to handle the different scenarios of users logging into FxA.
+  let email = false;
+
+  let growthTaggedURL = new URL("/oauth/init", document.body.dataset.serverUrl);
```
lesleyjanenorton: I think we can just continue to use `url` since this is the same as `url`

public/js/monitor.js
```diff
@@ -36,19 +37,100 @@ function isValidEmail(val) {
...
+  growthTaggedURL = appendFxaParams(growthTaggedURL, document.body.dataset);
+
+  ["flowId", "flowBeginTime", "entrypoint", "form_type", "entrypoint_experiment", "entrypoint_variation"].forEach(key => {
```
lesleyjanenorton: And then just append "entrypoing_experiment" and "entrypoint_variation" here?

public/js/monitor.js
maxxcrawford: Revised!

public/js/monitor.js
maxxcrawford: Great catch. Done.

public/js/monitor.js
maxxcrawford: I added one qualifier – making sure there is a value to inject before placing the tag.

public/js/monitor.js
```diff
@@ -36,17 +36,96 @@ function isValidEmail(val) {
...
+      url.searchParams.append(key, encodeURIComponent(el.dataset[key]));
+    }
   });
+
+  if (!sessionStorage) {
+    window.location.assign(url);
+    return;
+  }
+
+  const scannedEmailId = document.querySelector("#scan-user-email input[name=scannedEmailId]");
```
lesleyjanenorton, line 61: This will be `null` if an FxA entrypoint on other pages where there isn't a scan form is clicked.

server.js
```diff
@@ -103,6 +104,7 @@ const connectSrc = [
   "https://code.cdn.mozilla.net/fonts/",
   "https://www.google-analytics.com",
   "https://accounts.firefox.com",
+  "https://accounts.stage.mozaws.net/metrics-flow",
```
groovecoder, line 107: Hmm ... makes me a little nervous. But we can remove this when the experiment is done, yes?

PR #1614, opened 2020-03-24
bug 1616983: delete-user script for when we need it

groovecoder:


scripts/delete-user.js
```diff
@@ -0,0 +1,26 @@
...
+
+(async () => {
+  rl.question("What FXA primary email address? ", first_answer => {
+    rl.question("Please re-type the email address to confirm. ", async (second_answer) => {
+      if (first_answer !== second_answer) {
+        console.error("Email addresses do not match.");
+        process.exit();
+      }
+      const subscriber = await DB.getSubscriberByEmail(second_answer);
+      await DB.deleteEmailAddressesByUid(subscriber.id);
```
pdehaan: Do we need to check if the subscriber was found and `subscriber` isnt null/undefined before trying to yank it's `id` and `fxa_uid` properties? Maybe not since this is just a CLI tool and probably doesnt hurt anything if we get a runtime error of accessing props on undefined objects.

scripts/delete-user.js
```diff
@@ -0,0 +1,26 @@
...
+  input: process.stdin,
+  output: process.stdout,
+});
+
+(async () => {
+  rl.question("What FXA primary email address? ", first_answer => {
+    rl.question("Please re-type the email address to confirm. ", async (second_answer) => {
+      if (first_answer !== second_answer) {
+        console.error("Email addresses do not match.");
+        process.exit();
```
pdehaan: Exit with an error code here?

groovecoder:
Updated.

groovecoder:
As far as Monitor is concerned, we'll wait for FXA to verify requests to delete accounts.

PR #1639, opened 2020-04-13
Fix breaches page

lesleyjanenorton:


groovecoder:
LGTM.

pdehaan:
But why did this break now?
I tried doing a "blame" in GitHub for those lines and it looks like the code hand't changed in 10 months.
Curious what else changed recently that would have potentially caused this.

jrbenny35:
> But why did this break now?
> I tried doing a "blame" in GitHub for those lines and it looks like the code hand't changed in 10 months.
> Curious what else changed recently that would have potentially caused this.

Looks like it started failing with an NPM update: https://circleci.com/gh/mozilla/blurts-server/10038, PR https://github.com/mozilla/blurts-server/pull/1637

First thoughts are that the failures could've been unrelated, but not sure.

groovecoder:
Merging to un-block stage (and then prod) deployment.

maxxcrawford:
Jumping into this belated – PR #1637 was my doing. Are there any takeaways for next time?

groovecoder:
I think we need to take the CI failures more critically, even in the case of `npm audit` fixes. I know in the past my thinking has been that `npm audit` is a just-merge-it kind of update. But we should be more diligent about CI failures even on seemingly small changes.

PR #1649, opened 2020-04-20
Update integration test baseline images

maxxcrawford:
Created new baseline images for both headless and non-headless test variations.


groovecoder:
I think this has been superseded now. Closing.

PR #1667, opened 2020-04-27
Update Integration Test - OS Differences 

maxxcrawford:
Increased tolerance for Security Tips page to account for font renderng issues

### DIFF
--- a/tests/integration/tests/test-securityTips-page.js
+++ b/tests/integration/tests/test-securityTips-page.js
@@ -8,6 +8,6 @@ describe("Firefox Monitor Security Tips Page", function() {
   });
 
   it("should look like normal", function() {
-    expect(browser.checkFullPageScreen("SecurityTips_Page", {})).to.be.within(0, 5);
+    expect(browser.checkFullPageScreen("SecurityTips_Page", {})).to.be.within(0, 15);
   });
 });
--- a/tests/integration/tests/test-securityTips-page.js
+++ b/tests/integration/tests/test-securityTips-page.js
@@ -8,6 +8,6 @@ describe("Firefox Monitor Security Tips Page", function() {
   });
 
   it("should look like normal", function() {
-    expect(browser.checkFullPageScreen("SecurityTips_Page", {})).to.be.within(0, 15);
+    expect(browser.checkFullPageScreen("SecurityTips_Page", {ignoreAntialiasing: true})).to.be.within(0, 15);
   });
 });


jrbenny35:
I think this is fine, seeing as the page is pretty long. The idea behind the visual checks is just to make sure the page is working correctly.

tests/integration/tests/test-securityTips-page.js
```diff
@@ -8,6 +8,6 @@ describe("Firefox Monitor Security Tips Page", function() {
   });
 
   it("should look like normal", function() {
-    expect(browser.checkFullPageScreen("SecurityTips_Page", {})).to.be.within(0, 5);
+    expect(browser.checkFullPageScreen("SecurityTips_Page", {})).to.be.within(0, 15);
```
jrbenny35, line 11: Please add ```ignoreAntialiasing: true``` to the array

pdehaan:
Ha, the one where @pdehaan suggests injecting/overwriting some [redacted](https://changelog.com/posts/redacted-open-source-font) font to try and fix cross-OS issues.

jrbenny35:
@pdehaan if you create a baseline with that font I can try on my system as well

tests/integration/tests/test-securityTips-page.js
pdehaan: @jrbenny35 And just to remind myself, the `.within(0, 15)` means that 0-15% of the pixels in the screenshot can be wrong (versus the baseline image)?

jrbenny35: Yes. 15% of the overall page can be wrong. Since this is a big page, its a big percentage of that page.

jrbenny35:
@lesleyjanenorton @pdehaan would it make more sense to have 2 different sets of images based on the OS of the user running the tests?

pdehaan:
> would it make more sense to have 2 different sets of images based on the OS of the user running the tests?

I think that sounds reasonable (assuming it also means we can avoid the 15% pixel difference).
I was trying to calculate how wrong the security tips page could be before it ever failed, and it seems like a massive 1920px-by-18136px page with 15% tolerance can be off by a factor of 5.2m pixels before our tests would fail. Or, roughly the first 1920x2700 of the page could be completely missing and the test would still pass, assuming the rest of the page was 99.x% pixel perfect.

**PROS:**
- Fewer cross OS problems/failures (we hope).

**CONS:**
- Difficult for a single person to regen baselines for both OSes unless they have both OSes.
- Unclear if it means we can push a failing PR up to GitHub, let CI fail, then grab the failure artifacts as baselines for an specific OS so you don't need to figure out how to set up a Windows VM just to generate Windows baselines.
- Might need to see if we can get CircleCI to run on multiple OSes (Linux+Windows)?

jrbenny35:
> I was trying to calculate how wrong the security tips page could be before it ever failed, and it seems like a massive 1920px-by-18136px page with 15% tolerance can be off by a factor of 5.2m pixels before our tests would fail. Or, roughly the first 1920x2700 of the page could be completely missing and the test would still pass, assuming the rest of the page was 99.x% pixel perfect.

Damn, that's too much imo.

jrbenny35:
Maybe we just limit the testing to docker. The docker has a vnc server built in so you can still watch the tests if needed. And we could disable the visual tests if you are not within the docker container. Thoughts @pdehaan ?

groovecoder:
I'd prefer to make the security tips page test more resilient, rather than make it more specific and add more maintenance. Image-based tests should catch if an entire page is goes blank for some reason, but I can't think of anything that would make the first viewport blank without destroying the rest of the page too.

Maintaining even more image baselines for unclear test resilience seems like the opposite direction we want to go. So let's just bump the tolerance up to 15%.

jrbenny35:
> I'd prefer to make the security tips page test more resilient, rather than make it more specific and add more maintenance. Image-based tests should catch if an entire page is goes blank for some reason, but I can't think of anything that would make the first viewport blank without destroying the rest of the page too.
> 
> Maintaining even more image baselines for unclear test resilience seems like the opposite direction we want to go. So let's just bump the tolerance up to 15%.

I do think this goes with the idea of the visual tests. It's not trying to match exact words, but making sure the page is there and working

groovecoder:
That's why I like the 15% tolerance change on this particular page, which we know is substantially larger than all other pages, and particularly susceptible to minor visual differences cascading into larger visual effects on the overall page without breaking the page.

So, I've approved this change as submitted. Is there something else we wanted to also change before merging?

maxxcrawford:
@groovecoder @jrbenny35 Updated per Ben's request and caught this back up with `master`. LMK if there's more needed.

groovecoder:
Cool, thanks all.

PR #1670, opened 2020-04-28
Fixed #1669 - Set default dataset to use SERVER_URL.hostname

### LINKED ISSUE #1669: Set utm_source to use URL hostname rather than full server URL
Specifically, while on an experiment, it sets it correctly: 
https://github.com/mozilla/blurts-server/pull/1641/files#diff-e0a21ca0a3feea047cb7b35cfbdac200

maxxcrawford:
## To verify: 

Start the OAUth process and examine the URL. Check the `utm_source` param. It should be the `.hostname` of whatever dev site URL you're testing on. 

Example: On dev master: `utm_source=fx-breach-alerts.herokuapp.com`

### DIFF
--- a/public/js/fxa-analytics.js
+++ b/public/js/fxa-analytics.js
@@ -46,7 +46,7 @@ function appendFxaParams(url, storageObject) {
 
 function getFxaUtms(url) {
   if (sessionStorage) {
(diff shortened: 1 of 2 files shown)


groovecoder:
integration test failure is known.

Issue #1643, opened 2020-04-14
Set nonInteraction: true on "View" pings in GA


groovecoder:
We should set all of our GA "View" pings to include `{nonInteraction: true}`, since those events are non-interactive. This will fix our bounce rate metrics in GA.

PR #1672, opened 2020-04-29
Exp 2 Analytics Fix

maxxcrawford:


controllers/utils.js
```diff
@@ -2,6 +2,8 @@
 
 const crypto = require("crypto");
 const uuidv4 = require("uuid/v4");
+const mozlog = require("../log");
+const log = mozlog("controllers.utils");
```
maxxcrawford, line 6: Added debug logging. Information should show up in terminal when running locally.

controllers/utils.js
```diff
@@ -33,39 +35,57 @@ function hasUserSignedUpForRelay(user) {
 
 function getExperimentBranch(req, sorterNum) {
 
+  if (req.session.excludeFromExperiment && !req.query.experimentBranch) {
```
maxxcrawford, line 38: This is the primary fix for #1671. In summary, any time a session would return `false` for `getExperimentBranch()`, it would be forgotten when the function was called again (on another page load, refresh, etc). This applies specifically to a use who would be eligible to be in the experiment. They would have the experiment sorter run until it landed them in one of the experiment branches. 

One note about this if statement is that it includes an override for a query parameter. This is for testing purposes. If you add `?experimentBranch=vb`, it will place you in that cohort, regardless of getting discounted previously. One caveat is that if your language settings rule you out, this query request is denied (Again – for testing purposes).

controllers/utils.js
```diff
@@ -33,39 +35,57 @@ function hasUserSignedUpForRelay(user) {
...
   }
 
   // If URL param has experimentBranch entry, use that branch;
   if (req.query.experimentBranch) {
     if (!["va", "vb"].includes(req.query.experimentBranch)) {
+      log.debug("The requested branch is unknown: ", req.query.experimentBranch);
       return false;
     }
+    log.debug("This session has been set to the requested branch: ", req.query.experimentBranch);
+    req.session.excludeFromExperiment = false;
```
maxxcrawford, line 65: Because this session variable is only set during `return false` occasions, this reset here was necessary.

public/js/fxa-analytics.js
```diff
@@ -126,7 +126,6 @@ function sendRecommendationPings(ctaSelector) {
       hitCallback: function() {
         removeUtmsFromUrl();
       },
-      nonInteraction: true,
```
maxxcrawford, line 129: Because the Growth team was having issues tracking the experiment in GA, we want to minimize changes at this time. I propose we re-open #1643 to track that issue in the future.

public/js/monitor.js
```diff
@@ -62,6 +62,16 @@ function doOauth(el) {
         url.searchParams.append(key, document.body.dataset[key]);
       }
     });
+
+    if (typeof(ga) !== "undefined") {
```
maxxcrawford, line 66: Added back this event to mirror what was fired in the first experiment.

views/partials/analytics/experiment.hbs
```diff
@@ -1,3 +1,3 @@
 data-utm_term={{experimentBranch}}
 data-experiment={{experimentBranch}}
-data-utm_campaign="growthuserflow2"
+data-utm_campaign="growthuserflow3"
```
maxxcrawford, line 3: Revised this ID for a clean break in the data

tests/controllers/home.test.js
```diff
@@ -62,27 +71,51 @@ test("notFound set status 404 and renders 404", () => {
...
+  mockRequest.session = {
+    excludeFromExperiment: false,
+  };
+
+  // The session is assigned to the control group when the coin flip is 0;
   let experimentNumber = 0;
   let experimentBranch = getExperimentBranch(mockRequest, experimentNumber);
   expect(experimentBranch).toBe("va");
 
+  mockRequestSessionReset(mockRequest);
```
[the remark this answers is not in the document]
maxxcrawford: Copy that. I did a session earlier this week with @gregtatum and we talked about resetting the request/env between each test. This function is a compromise – the extra step (that I did not opt for) is to break this test out at each `expect` to ensure it's a clean start every time.

controllers/user.js
```diff
@@ -554,9 +554,12 @@ async function getBreachStats(req, res) {
 function logout(req, res) {
   if (AppConstants.EXPERIMENT_ACTIVE) {
     // Persist experimentBranch across session reset
+    const excludeFromExperiment = req.session.excludeFromExperiment;
     const experimentBranch = req.session.experimentBranch;
     req.session.reset();
+    // Reset session vars after sign out event
     req.session.experimentBranch = experimentBranch;
+    req.session.excludeFromExperiment = excludeFromExperiment;
```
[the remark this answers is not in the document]
maxxcrawford: My in-code comment is a little murky. We want to make sure that any user assigned to a specific experiment (including being excluded) is remembered past the session reset. 

The `req.session.reset` removes both of these variables.

public/js/fxa-analytics.js
```diff
@@ -120,13 +120,15 @@ function sendRecommendationPings(ctaSelector) {
       // If an experiment is active, set the "Growth Experiment Version"
       // Custom Dimension to whichever branch is active.
       ga("set", "dimension7", `${document.body.dataset.experiment}`);
+      ga("set", "dimension8", `${document.body.dataset.experiment}`);
+      ga("set", "campaignName", `${document.body.dataset.utm_campaign}`);
+      ga("set", "campaignTerm", `${document.body.dataset.utm_term}`);
```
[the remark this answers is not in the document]
maxxcrawford: Great catch. Fixing this  – that property doesn't exist. I was following naming conventions (and that property is the one that contains the experiment branch name). Looks like I need to set it to [campaignKeyword](https://developers.google.com/analytics/devguides/collection/analyticsjs/field-reference#campaignKeyword). 

/cc @zerokarmaleft

views/partials/analytics/experiment.hbs
zerokarmaleft: Can we keep and append to `growthuserflow2`, e.g. `growthuserflow2a` so we can keep them sequentially in sync with our project docs?

public/js/fxa-analytics.js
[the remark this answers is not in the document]
zerokarmaleft: Yea, `campaignKeyword` is the right choice for `utm_term`

PR #1674, opened 2020-04-30
GA Bounce Rate Tracking Test

maxxcrawford:


public/js/fxa-analytics.js
```diff
@@ -25,13 +25,17 @@ function getLocation() {
 }
 
 
-async function sendPing(el, eventAction, eventLabel = null) {
+async function sendPing(el, eventAction, eventLabel = null, options = null) {
```
maxxcrawford, line 28: Added options object to include `nonInteraction: true` on some pings.

public/js/resolve-breaches.js
```diff
@@ -32,7 +32,12 @@
 
   function sendBreachDetailAnalyticsPing(eventCategory, eventAction, eventLabel) {
     if (typeof(ga) !== "undefined") {
-      ga("send", "event", eventCategory, eventAction, eventLabel);
+      // Set view pings as nonInteraction:true to get accurate bounce rate
+      let options = {};
+      if (eventAction !== "Engage"){
+        options = {nonInteraction: true};
```
maxxcrawford, line 38: Same as on `fx-analytics.js` – We are wanting to make sure only user-interacted events get marked as interactive in GA.

public/js/resolve-breaches.js
```diff
@@ -32,7 +32,12 @@
 
   function sendBreachDetailAnalyticsPing(eventCategory, eventAction, eventLabel) {
     if (typeof(ga) !== "undefined") {
-      ga("send", "event", eventCategory, eventAction, eventLabel);
+      // Set view pings as nonInteraction:true to get accurate bounce rate
+      let options = {};
+      if (eventAction !== "Engage"){
```
groovecoder, line 37: @lesleyjanenorton - do we have any "Click" event actions?

maxxcrawford: Both of these events would get marked as interactive:
https://github.com/mozilla/blurts-server/blob/master/public/js/resolve-breaches.js#L104
https://github.com/mozilla/blurts-server/blob/master/public/js/resolve-breaches.js#L92

lesleyjanenorton: Does this mean that other events like "Submit" will be marked as non-interactive?

lesleyjanenorton: What about changing this conditional to look specifically for eventAction(s) that should be marked as `nonInteraction: true` instead of adding `nonInteraction: true` to any eventAction that isn't `"Engage"`?

Issue #1665, opened 2020-04-24
Add check-box for Firefox Account on the Monitor homepage


winstonbowden:
Graduate Monitor Experiment 1 - Issue #1587 

![Screen Shot 2020-05-04 at 9 55 08 AM](https://user-images.githubusercontent.com/6164801/80986191-76e97b80-8ded-11ea-8164-a28a49538ac8.png)

maxxcrawford:
- Copy will need to be localized

winstonbowden:
EN only for now.

changecourse:
@betsymi would you be willing to do a drive-by review of the string associated with the checkbox? This string ran for our experiment 1 already, and thus we didn't do a formal review, but since the data came back positively, we're wanting to do our due diligence now.

betsymi:
> @betsymi would you be willing to do a drive-by review of the string associated with the checkbox? This string ran for our experiment 1 already, and thus we didn't do a formal review, but since the data came back positively, we're wanting to do our due diligence now.

Sure, thanks for asking. Has legal had eyes on it as well? Are you asking for me to comment my drive-by string, right here in this thread? Or do I need to put it somewhere else?

changecourse:
Legal has approved it (correct @winstonbowden?) 

Right here on this thread is perfect, and I can work with the team from there.

winstonbowden:
Thanks for asking about legal. They did have a request, which is call it out as a "known" breach.  Something like this:
Stay Safe: get automated alerts when we know about an account breach.

changecourse:
Updated string from @betsymi 

>Get email alerts when your info appears in a known breach

@winstonbowden let's discuss whether or not we need to re-run vs. moving forward with this. IMHO, I don't think we're at risk to limit impact with this string vs. our original experiment string, with regards to impact/growth.

winstonbowden:
@changecourse thanks! I'm good with the proposed string as long as we keep "Stay safe" in front of it.

changecourse:
"Final" string: 
Stay safe: get email alerts when your info appears in a known breach

![Screen Shot 2020-05-04 at 9 13 37 AM](https://user-images.githubusercontent.com/6164801/80981740-9bdaf000-8de7-11ea-8104-0475c80c64d6.png)

Ready for graduation, @winstonbowden @maxxcrawford

maxxcrawford:
The string is now in `master` and will get localized.

Issue #1587, opened 2020-03-16
Opt-in/Opt-out Experiment


maxxcrawford:
### Summary: 

The Growth team will conduct an experiment on Monitor, centered around when to present users with the FxA signup/signin. This experiment will help us learn more about the FxA funnel as it relates to Monitor. 

#### Notes: 

- **What question does this experiment answer?** If we ask a user to sign up for future email breach alerts immediately after they submit their email address on Firefox Monitor, we will increase the volume of Accounts being generated from Firefox Monitor.
- **Hypothesis:** How can we adapt the Firefox Monitor user flow to improve the conversion rate from a breach result to a Firefox account?
- **Scope:** This experiment focuses specifically on the homepage-entry point breach check. There are other ways in the site (ex: breah detail) to check your email address for known breaches. 

### Experiment Branches: 

- Control
  - The current flow – users enter email address, are shown known associated breaches to that email address and provided with button to get future alerts (FxA) 
- Treatment A: Opt-in
  - User is presented with a check box (unchecked) to sign up for alerts for future breaches. If checked, user is routed through FxA account create/sign-in flow and then redirected back to breach results. 
- Treatment B: Opt-out
  - Works exactly as Treatment A, except the check box is already checked.

### UX: 

[InVision Mockups](https://mozilla.invisionapp.com/share/AHWCN84JPN9#/screens/408543238_Growth_Experiment_Landing)

### Implementation: 

We will use a similar method to how the New Scan page experiment was conducted last November. https://github.com/mozilla/blurts-server/pull/1376 

### Caveats

- #1579 - Add modal for when a user has an FxA account associated with the email they entered. 
- Data will be a blend of both Google Analytics Custom Dimensions and Amplitude
FxA currently has a single redirect URI – we will need to add a feature to set this dynamically, based on point of entry
- Focused only on EN-US (No localization)

PR #1681, opened 2020-05-05
Add new text string for homepage checkbox

maxxcrawford:
- This PR is going in before the rest of the dev work for #1665 to give the l10n team time to get the string localized. 

This feature is a graduation from the first growth team experiment. (#1587)

### DIFF
--- a/locales/en/app.ftl
+++ b/locales/en/app.ftl
@@ -280,6 +280,7 @@ appears-in-x-breaches =
 
 check-for-breaches = Check for Breaches
 find-out-what-hackers-know = Find out what hackers already know about you. Learn how to stay a step ahead of them.
+get-email-alerts = Stay safe: Get email alerts when your info appears in a known breach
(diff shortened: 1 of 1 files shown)


flodolo:
Note: I don't mind trimming the whitespaces, but it should be done in a separate PR, as a stand-alone change.

maxxcrawford:
@flodolo Sorry about that. Whitespace removed!

locales/en/app.ftl
```diff
@@ -280,6 +280,7 @@ appears-in-x-breaches =
 
 check-for-breaches = Check for Breaches
 find-out-what-hackers-know = Find out what hackers already know about you. Learn how to stay a step ahead of them.
+get-email-alerts = Stay safe: Get email alerts when your info appears in a known breach
```
maxxcrawford, line 283: Note – I confirmed with @changecourse and @winstonbowden  that there is no punctuation on this string.

PR #1689, opened 2020-05-14
Add survey signup to dashboard

lesleyjanenorton:


public/js/monitor.js
```diff
@@ -410,6 +438,34 @@ function addBentoObserver(){
...
+    }
+  };
+  const promoHasAlreadyBeenDismissed = (researchPromo && sessionStorage.getItem("fxMonitorResearchPromo") === "dismissed");
+  const sessionStorageDoesNotExist = (researchPromo && !sessionStorage);
+  const acceptedLanguages = navigator.languages;
+  const acceptedLocalesIncludesEnglish =
+    (acceptedLanguages.includes("en-US") || acceptedLanguages.includes("en"));
+
+  if (promoHasAlreadyBeenDismissed || sessionStorageDoesNotExist || !acceptedLocalesIncludesEnglish) {
+    researchPromo.style.display = "none";
```
groovecoder, line 454: Could we save a lot of `researchPromoIsVisible` logic if we `return;` here?

public/js/monitor.js
```diff
@@ -410,6 +438,34 @@ function addBentoObserver(){
       });
     }
   }
+  const researchPromo = document.querySelector(".research-recruitment-wrapper");
+  const sendRecruitmentPing = (action, interactionStatus) => {
+    if (gaAvailable) {
+      ga("send", "event", "Research Recruitment Promo", action, "Recruitment Promo", {nonInteraction: interactionStatus});
```
maxxcrawford, line 444: You'll need to send additional options on the click/Engage event (but not the "view" event) 
`{transport: "beacon"}`

Similar to this: https://github.com/mozilla/blurts-server/blob/master/public/js/fxa-analytics.js#L75

public/js/monitor.js
lesleyjanenorton: We could! Thank you!

groovecoder:
r+

maxxcrawford:
r+

PR #1691, opened 2020-05-17
Fixed #1665 - Added FxA checkbox to the homepage 

### LINKED ISSUE #1665: Add check-box for Firefox Account on the Monitor homepage

maxxcrawford:
- Graduating Growth Experiment 1 (#1587) 
- Review site: https://monitor-v2.herokuapp.com/ 

## To verify: 

- Scan for breaches from the homepage with the checkbox checked
- FxA OAuth flow show start


views/partials/forms/scan-email-form.hbs
```diff
@@ -15,7 +24,7 @@
     {{#if featuredBreach}}
       <input type="submit" value="{{ getString "find-out" }}"/>
     {{else}}
-      <input type="submit" value="{{ getString "check-for-breaches" }}"/>
+      <input class="add-metrics-flow-values" data-form_type="other" {{> analytics/fxa id="fx-monitor-check-for-breaches-blue-btn" }} type="submit" value="{{ getString "check-for-breaches" }}"/>
```
maxxcrawford, line 27: I'm intentionally not using `.open-oauth`class as it needs to be conditional based if the user opts-in. `'.open-oauth` has other events attached to it that would fire regardless.

public/js/monitor.js
```diff
@@ -36,7 +36,7 @@ function isValidEmail(val) {
 }
 
 
-function doOauth(el) {
+function doOauth(el, {emailWatch = false} = {}) {
```
groovecoder, line 39: whoa.

public/js/monitor.js
```diff
@@ -51,15 +51,68 @@ function doOauth(el) {
     return;
   }
 
-  // Preserve entire control function
+  if (!emailWatch) {
```
groovecoder, line 54: Do we need type-safe checking here? In case someone later sends in `{emailWatch: 0}` ?

groovecoder, line 54: Do we need type-safe checking here? In case someone later sends in `{emailWatch: 0}` ?

public/js/monitor.js
```diff
@@ -51,15 +51,68 @@ function doOauth(el) {
...
+      const lastScannedEmail = sessionStorage.getItem("lastScannedEmail");
+      if (lastScannedEmail) {
+        url.searchParams.append("email", lastScannedEmail);
+      }
+    }
+    window.location.assign(url);
+    return;
+  }
+
+  let email = false;
```
groovecoder, line 66: `let` sniff. any way to make this a `const`?

public/js/monitor.js
```diff
@@ -51,15 +51,68 @@ function doOauth(el) {
...
+          break;
+        case !lastScannedEmail:
+          // The last saved email address and the current entry DIFFER, so create
+          // a new entry, launch a new FxA login session with new email prefilled.
+          sessionStorage.removeItem("lastScannedEmail");
+          sessionStorage.setItem("lastScannedEmail", email);
+          scannedEmailId.value = sessionStorage.length;
+          break;
+      }
+    } else {
```
groovecoder, line 97: `else` sniff. any way to make this a helper function with an early `return` statement?

public/js/monitor.js
```diff
@@ -51,15 +51,68 @@ function doOauth(el) {
...
+      if (lastScannedEmail) {
+        // Control method. User set this by checking breach results
+        email = lastScannedEmail;
+      }
     }
+  } else if (email && sessionStorage) {
```
groovecoder, line 103: `else if` sniff. helper function with early `return` statement?

public/js/monitor.js
```diff
@@ -436,6 +489,50 @@ function resizeDashboardMargin() {
   }
 
   const dropDownMenu = document.querySelector(".mobile-nav.show-mobile");
+  const submitBtn = document.querySelector("#scan-user-email input[type='submit']");
+
   dropDownMenu.addEventListener("click", () => toggleDropDownMenu(dropDownMenu));
 
+  const acceptedFirstLocalesIncludesEnglish =
+  (acceptedLanguages[0].includes("en-US") || acceptedLanguages[0].includes("en"));
```
groovecoder, line 497: Is `acceptedLanguages[0]` a string? If so, `.includes("en")` will return `true` for any `en` language, so, the `en-US`-specific check isn't needed?

public/js/monitor.js
```diff
@@ -436,6 +489,50 @@ function resizeDashboardMargin() {
   }
 
   const dropDownMenu = document.querySelector(".mobile-nav.show-mobile");
+  const submitBtn = document.querySelector("#scan-user-email input[type='submit']");
+
   dropDownMenu.addEventListener("click", () => toggleDropDownMenu(dropDownMenu));
 
+  const acceptedFirstLocalesIncludesEnglish =
```
groovecoder, line 496: nit: variable name `firstAcceptedLanguageIsEnglish` might be more descriptive.

public/js/monitor.js
maxxcrawford: Would it be easier to just make it the following? 
```
doOauth(el, emailWatch = fase) {…
```

public/js/monitor.js
```diff
@@ -51,68 +67,34 @@ function doOauth(el, {emailWatch = false} = {}) {
     return;
   }
 
+  const lastScannedEmail = sessionStorage.getItem("lastScannedEmail");
+
+  if (typeof emailWatch !== "boolean") {
+    throw new Error("invalid argument option in doOauth");
```
groovecoder, line 73: And someday we all use typescript just like Greg always said we should.

public/js/monitor.js
maxxcrawford: Added. 

https://github.com/mozilla/blurts-server/blob/1665-add-homepage-fxa-checkbox/public/js/monitor.js#L72-L74

public/js/monitor.js
maxxcrawford: 🎯  

cc/ @gregtatum

public/js/monitor.js
maxxcrawford: acceptedLanguages[0] is a string. Revised!

public/js/monitor.js
maxxcrawford: Revised name to be `acceptedFirstLanguageIsEnglish`

PR #1695, opened 2020-05-19
Relaunch: Homepage FxA Experiment (Growth Exp 2)

maxxcrawford:
### Treatment

**Note: When testing locally, add the variable `EXPERIMENT_ACTIVE=1` to your `.ENV` file.**

1. Go to https://monitor-v2.herokuapp.com/?experimentBranch=vb
1. Click "Get Started" Button
1. Go through FxA Flow 
1. Once completed, user should be on the `/user/dashboard/` page

## Expected Results

**Control:** 
image

**Treatment:** 
_Note - The text is still being worked on._

image

### DIFF
--- a/controllers/home.js
+++ b/controllers/home.js
@@ -3,7 +3,9 @@
 const AppConstants = require("../app-constants");
 const DB = require("../db/DB");
 const { scanResult } = require("../scan-results");
-const { generatePageToken } = require("./utils");
+const { generatePageToken, getExperimentBranch } = require("./utils");
+
+const EXPERIMENTS_ENABLED = (AppConstants.EXPERIMENT_ACTIVE === "1");
 
 async function home(req, res) {
 
@@ -15,8 +17,22 @@ async function home(req, res) {
   let featuredBreach = null;
   let scanFeaturedBreach = false;
 
+  // Growth Experiment
+  let experimentBranch = null;
+  let isUserInExperiment = null;
+  let experimentBranchB = null;
+
   if (req.session.user && !req.query.breach) {
     return res.redirect("/user/dashboard");
+  }
+
+  // Growth Experiment
+  if (EXPERIMENTS_ENABLED) {
+    experimentBranch = getExperimentBranch(req, "en");
+    if (!experimentBranch) { req.session.excludeFromExperiment = true; }
+    req.session.experimentBranch = experimentBranch;
+    isUserInExperiment = (experimentBranch === "vb");
+    experimentBranchB = (experimentBranch === "vb" && isUserInExperiment);
   }
 
   if (req.query.breach) {
@@ -40,6 +56,8 @@ async function home(req, res) {
       scanFeaturedBreach,
       pageToken: formTokens.pageToken,
       csrfToken: formTokens.csrfToken,
+      experimentBranch,
+      experimentBranchB,
     });
   }
 
@@ -49,6 +67,9 @@ async function home(req, res) {
     scanFeaturedBreach,
     pageToken: formTokens.pageToken,
     csrfToken: formTokens.csrfToken,
+    experimentBranch,
+    isUserInExperiment,
+    experimentBranchB,
   });
 }
 
--- a/controllers/home.js
+++ b/controllers/home.js
@@ -28,7 +28,7 @@ async function home(req, res) {
 
   // Growth Experiment
   if (EXPERIMENTS_ENABLED) {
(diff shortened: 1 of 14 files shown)


template-helpers/homepage.js
```diff
@@ -28,6 +32,11 @@ function makeLanding(args) {
     });
   }
 
+  // Growth Experiment
+  if (featuredBreach && experimentBranchB) {
+    landingCopy.subhead = "Your free Firefox account alerts you if you’ve been involved in a known data breach.";
```
lesleyjanenorton, line 37: Just verifying - do we want this to be English only?

controllers/user.js
```diff
@@ -230,6 +229,18 @@ async function getDashboard(req, res) {
   const allBreaches = req.app.locals.breaches;
   const { verifiedEmails, unverifiedEmails } = await getAllEmailsAndBreaches(user, allBreaches);
 
+  let experimentBranch = null;
+  let isUserInExperiment = null;
+  let experimentBranchB = null;
```
groovecoder, line 234: `let` = code sniff. Can we wrap these up in a single object and do something more like:
```js
const experimentFlags = _getExperimentFlags(req);
...
res.render("dashboards", {
...
  experimentFlags,
});
```

controllers/home.js
```diff
@@ -15,10 +17,24 @@ async function home(req, res) {
   let featuredBreach = null;
   let scanFeaturedBreach = false;
 
+  // Growth Experiment
+  let experimentBranch = null;
+  let isUserInExperiment = null;
+  let experimentBranchB = null;
```
groovecoder, line 23: `let` = code sniff. Can we wrap these up in a single object and do something more like:
```js
const experimentFlags = _getExperimentFlags(req);
...
res.render("dashboards", {
...
  experimentFlags,
});
```

scan-results.js
```diff
@@ -6,28 +6,28 @@ const HIBP = require("./hibp");
...
   const allBreaches = req.app.locals.breaches;
   let scannedEmail = null;
 
   // Growth Experiment
-  // let experimentBranch = null;
-  // let isUserInExperiment = null;
-  // let experimentBranchB = null;
+  let experimentBranch = null;
+  let isUserInExperiment = null;
+  let experimentBranchB = null;
```
groovecoder, line 20: `let` = code sniff. Can we wrap these up in a single object and do something more like:
```js
const experimentFlags = _getExperimentFlags(req);
...
res.render("dashboards", {
...
  experimentFlags,
});
```

template-helpers/homepage.js
```diff
@@ -5,6 +5,7 @@ const { LocaleUtils } = require("./../locale-utils");
 function makeLanding(args) {
   const locales = args.data.root.req.supportedLocales;
   const featuredBreach = args.data.root.featuredBreach;
+  const experimentBranchB = args.data.root.experimentBranchB;
```
groovecoder, line 8: If we change the controllers to add an `experimentFlags` object to the template context, we'll need to update these references to `args.data.root.experimentFlags.experimentBranchB`, etc.

views/monitor.hbs
```diff
@@ -1,5 +1,5 @@
 {{!< default }}
-<main class="container monitor-homepage clear-header {{#if featuredBreach}} featured-breach {{/if}} {{#if featuredBreach.IsSensitive}} sensitive-landing {{/if}} data-page-label={{> analytics/landing_page_label }}>
+<main class="container monitor-homepage clear-header {{#if featuredBreach}} featured-breach {{/if}} {{#if featuredBreach.IsSensitive}} sensitive-landing {{/if}} {{#if experimentBranchB}} experiment-subhead {{/if}}" data-page-label={{> analytics/landing_page_label }}>
```
groovecoder, line 2: Might need to be `experimentFlags.experimentBranchB`

template-helpers/homepage.js
maxxcrawford: @lesleyjanenorton @groovecoder Correct – it's being filtered in this argument: 
https://github.com/mozilla/blurts-server/blob/1638-homepage-fxa-experiment/controllers/home.js#L31

We're using a direct string like this in order to not deploy any l10n resources.

maxxcrawford:
> Can we try to remove the `let` assignments and introduce a `getExperimentFlags` function instead?
> 
> If that becomes too much work, we can merge without the refactor.

Migrated the fix into another issue (#1698) and PR (https://github.com/mozilla/blurts-server/pull/1699)

PR #1700, opened 2020-05-22
Updated ESlint rules to use Protocol standards

maxxcrawford:
Added ESlint rules from https://github.com/mozilla/eslint-config-protocol project to standardize JavaScript layout. 

This PR makes that change and then fixes all errors, adding excepts where test-breaking level changes occurred.


public/js/monitor.js
```diff
@@ -7,529 +7,524 @@
...
+    const scripts = document.getElementsByTagName("script")[0];
+    cryptoScript.src = "/dist/edge.min.js";
+    scripts.parentNode.insertBefore(cryptoScript, scripts);
 }
 
-
 function findAncestor(el, cls) {
-  while ((el = el.parentElement) && !el.classList.contains(cls));
-  return el;
+    // FIXME: no op
```
maxxcrawford, line 17: This function is empty.

public/js/monitor.js
```diff
@@ -7,529 +7,524 @@
...
-  win.addEventListener("resize", () => {
-    const newWindowWidth = win.outerWidth;
-      if (newWindowWidth !== windowWidth) {
-      toggleMobileFeatures(topNavigation);
-      toggleArticles();
-      windowWidth = newWindowWidth;
-      resizeDashboardMargin();
-    }
-  });
+    // FIXME: no op
```
maxxcrawford, line 363: This function is empty.

__mocks__/sns-validator.js
```diff
@@ -1,13 +1,15 @@
 "use strict";
 
 // See https://jestjs.io/docs/en/manual-mocks
+// FIXME: no op
```
maxxcrawford, line 4: This function is empty.

scripts/send-breach-notification.js
```diff
@@ -9,44 +9,46 @@ const { LocaleUtils } = require("../locale-utils");
...
-      console.log(JSON.stringify(resp));
-
-      rl.close();
-
-      process.exit();
+    await EmailUtils.init();
+    let emailAddress,
+        breachName;
+    const resp = {
+        // FIXME: no op
```
maxxcrawford, line 25: This function is empty.

public/js/monitor.js
pdehaan: Technically, I think the loop is reassigning the value of `el` on each iteration until it finds a match, so I don't think it's a noop.

basket.js
```diff
@@ -5,35 +5,33 @@ const got = require("got");
...
-      log.error("subscribe", {stack: e.stack});
-    }
-  },
+    async subscribe(email) {
+        const params = {
+            email,
+            newsletters: AppConstants.BASKET_NEWSLETTER,
+            optin: "Y",
+            "api-key": AppConstants.BASKET_API_KEY,
+            // eslint-disable-next-line camelcase
```
pdehaan, line 18: Rather than shotgun this throughout the codebase, should we just blanket disable the rule in .eslint config?

db/.eslintrc
```diff
@@ -0,0 +1,5 @@
+{
+  "rules": {
+    "camelcase": ["error", {properties: "never"}]
+  }
+}
```
pdehaan, line 5: My guts tell me that we should add this as an override in the base .eslint config file (similar to other rules specific to folders).

email-utils.js
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable camelcase */
```
pdehaan, line 1: Same here. Looks like we have 3 standard ways now of disabling the `camelcase` rule for a line, file, folder. I vote we just disable the rule globally and remove these.

public/js/dashboard.js
```diff
@@ -1,96 +1,91 @@
...
+    return await response.json();
 }
 
 async function sendCommunicationOption(e) {
-  const { formAction, commOption, csrfToken } = e.target.dataset;
-  sendForm(formAction, { communicationOption: commOption, _csrf: csrfToken })
-    .then(data => {}) /*decide what to do with data */
-    .catch(e => {})/* decide how to handle errors */;
+    const { formAction, commOption, csrfToken } = e.target.dataset;
+    sendForm(formAction, { communicationOption: commOption, _csrf: csrfToken });
```
pdehaan, line 23: This behavior is slightly different, in that now we don't silently ignore any errors via the `.catch()` stuff. 🤷

public/js/dashboard.js
```diff
@@ -1,96 +1,91 @@
...
-    }, 1000);
-  }) /*decide what to do with data */
-  .catch(e => {})/* decide how to handle errors */;
+    await sendForm(formAction, { _csrf: csrfToken, emailId })
+        .then(data => {
+            setTimeout( ()=> {
+                const span = resendEmailBtn.nextElementSibling;
+                span.classList.remove("hide");
+            }, 1000);
+        });
```
pdehaan, line 37: Same as above. Now errors aren't silently ignored.

public/js/fx-bento.js
```diff
@@ -1,303 +1,302 @@
 "use strict";
-
+/*eslint radix: ["error", "as-needed"]*/
```
pdehaan, line 2: This feels odd. Why only in this file specifically?

public/js/fx-bento.js
```diff
@@ -1,303 +1,302 @@
...
-    // close Bento on mouse clicks outside the Bento menu
-    if (hasParent(event.target, "FIREFOX-APPS") === null) {
-      this._active = !this._active;
-      return closeBento();
-    }
+        const hasParent = (el, selector) => {
+            while (el.parentElement) {
+                el = el.parentElement;
+                if (el.tagName === selector)
+                {return el;}
```
pdehaan, line 149: nit: I had no idea why this line needed to be wrapped in braces, but it seems to be for the `if` block.

routes/breach-details.js
```diff
@@ -4,10 +4,9 @@ const express = require("express");
 const { getBreachDetail } = require("../controllers/breach-details");
 const { asyncMiddleware } = require("../middleware");
 
-
+// eslint-disable-next-line new-cap
```
pdehaan, line 7: This seems peppered throughout as well. Not sure if we disable the rule globally (or for the ./routes/ folder only) or see if there are ignore pattern options maybe.

scripts/delete-user.js
```diff
@@ -1,32 +1,33 @@
+/* eslint-disable camelcase */
```
pdehaan, line 1: 🐫

template-helpers/.eslintrc
```diff
@@ -0,0 +1,5 @@
+{
+  "rules": {
+    "camelcase": ["error", {properties: "never"}]
+  }
+}
```
pdehaan, line 5: Same as comment way above, should this be in the global .eslint config as an override (or disabled globally since it seems to be a very common pattern to ignore this anyways).

pdehaan:
PROTIP for future travellers. Disabling the whitespace changes will hide all the indenting diffs which makes this much easier (although it will still show all the empty line deletions, which is unfortunate).

Updated_ESlint_rules_to_use_Protocol_standards_by_maxxcrawford_·_Pull_Request__1700_·_mozilla_blurts-server

groovecoder:
I suggest we close this? Lots of conflicts now. :/

maxxcrawford:
@groovecoder Yeah. This was more of a learning exercise.

PR #1707, opened 2020-05-26
remove old, un-used loadtests

groovecoder:
https://github.com/mozilla/blurts-server/pull/1703 reminded me we still have these old load tests in the repo. We don't use them anymore - should just remove them.


pdehaan:
Fail CI? That's unpossible!

PR #1726, opened 2020-06-04
Growth Experiment #4 - i10n Homepage FxA Experiment

maxxcrawford:
### Steps to test: 
- Control: https://monitor-v2.herokuapp.com/?experimentBranch=va 
- Treatment: https://monitor-v2.herokuapp.com/?experimentBranch=vb

Verify you are enrolled in the experiment by looking at the `data-experiment` attribute on `<body>` tag. 

**Treatment:**
- Go to website (with FR or DE set as the language)
- Click the checkbox
- It should open FxA flow.
- When you land on the /user-dashboard after completing the flow, it should still have the same meta data as mentioned above. 

#### Notes: 
- The user MUST have either German or French variant as the primary/top language. Any other deviation will be excluded from the experiment. If the first language is NOT English, then they should not get the checkbox. If English (any variant), the default state is now to HAVE a checkbox. 

### Treatment Screenshots

**French Homepage**
image

**French Homepage (Featured Breach)**
image

**German Homepage**
image

**German Homepage (Featured Breach)**
image

### DIFF
--- a/controllers/home.js
+++ b/controllers/home.js
@@ -31,11 +31,13 @@ async function home(req, res) {
   }
 
   const experimentFlags = getExperimentFlags(req, EXPERIMENTS_ENABLED);
-  req.session.experimentFlags = experimentFlags;
 
   // Growth Experiment
   if (EXPERIMENTS_ENABLED) {
-    getExperimentBranch(req);
+    getExperimentBranch(req, false, ["de", "fr"], {
+      "va": 25,
+      "vb": 25,
+    });
   }
 
   if (req.query.breach) {
--- a/controllers/home.js
+++ b/controllers/home.js
@@ -4,7 +4,7 @@ const AppConstants = require("../app-constants");
 const DB = require("../db/DB");
 const HIBP = require("../hibp");
 const { scanResult } = require("../scan-results");
-const { generatePageToken, getExperimentBranch } = require("./utils");
+const { generatePageToken, getExperimentBranch, getExperimentFlags } = require("./utils");
 
 const EXPERIMENTS_ENABLED = (AppConstants.EXPERIMENT_ACTIVE === "1");
 
@@ -15,7 +15,6 @@ function _getFeaturedBreach(allBreaches, breachQueryValue) {
   const lowercaseBreachValue = breachQueryValue.toLowerCase();
   return HIBP.getBreachByName(allBreaches, lowercaseBreachValue);
 }
-
 
 async function home(req, res) {
 
@@ -31,11 +30,13 @@ async function home(req, res) {
     return res.redirect("/user/dashboard");
   }
 
-  if (EXPERIMENTS_ENABLED) {
-    // Growth Experiment
-  }
+  const experimentFlags = getExperimentFlags(req, EXPERIMENTS_ENABLED);
+  req.session.experimentFlags = experimentFlags;
 
-  // TODO: Set session experiment variables
+  // Growth Experiment
+  if (EXPERIMENTS_ENABLED) {
+    getExperimentBranch(req);
+  }
 
   if (req.query.breach) {
     featuredBreach = _getFeaturedBreach(req.app.locals.breaches, req.query.breach);
@@ -57,7 +58,7 @@ async function home(req, res) {
       scanFeaturedBreach,
       pageToken: formTokens.pageToken,
       csrfToken: formTokens.csrfToken,
-      // TODO: Expose session experiment variables to template
+      experimentFlags,
     });
   }
 
@@ -67,7 +68,7 @@ async function home(req, res) {
     scanFeaturedBreach,
     pageToken: formTokens.pageToken,
     csrfToken: formTokens.csrfToken,
-    // TODO: Expose session experiment variables to template
+    experimentFlags,
   });
 }
 
--- a/controllers/home.js
+++ b/controllers/home.js
@@ -4,7 +4,9 @@ const AppConstants = require("../app-constants");
 const DB = require("../db/DB");
 const HIBP = require("../hibp");
 const { scanResult } = require("../scan-results");
-const { generatePageToken } = require("./utils");
+const { generatePageToken, getExperimentBranch } = require("./utils");
+
+const EXPERIMENTS_ENABLED = (AppConstants.EXPERIMENT_ACTIVE === "1");
 
 function _getFeaturedBreach(allBreaches, breachQueryValue) {
   if (!breachQueryValue) {
@@ -29,6 +31,12 @@ async function home(req, res) {
     return res.redirect("/user/dashboard");
   }
 
+  if (EXPERIMENTS_ENABLED) {
+    // Growth Experiment
+  }
+
+  // TODO: Set session experiment variables
+
   if (req.query.breach) {
     featuredBreach = _getFeaturedBreach(req.app.locals.breaches, req.query.breach);
 
@@ -49,6 +57,7 @@ async function home(req, res) {
       scanFeaturedBreach,
       pageToken: formTokens.pageToken,
       csrfToken: formTokens.csrfToken,
+      // TODO: Expose session experiment variables to template
     });
   }
 
@@ -58,6 +67,7 @@ async function home(req, res) {
     scanFeaturedBreach,
     pageToken: formTokens.pageToken,
     csrfToken: formTokens.csrfToken,
+    // TODO: Expose session experiment variables to template
   });
 }
 
--- a/controllers/user.js
+++ b/controllers/user.js
@@ -11,6 +11,8 @@ const HIBP = require("../hibp");
 const { resultsSummary } = require("../scan-results");
 const sha1 = require("../sha1-utils");
 
+const EXPERIMENTS_ENABLED = (AppConstants.EXPERIMENT_ACTIVE === "1");
+
 const FXA_MONITOR_SCOPE = "https://identity.mozilla.com/apps/monitor";
 
 async function removeEmail(req, res) {
@@ -227,6 +229,11 @@ async function getDashboard(req, res) {
   const allBreaches = req.app.locals.breaches;
   const { verifiedEmails, unverifiedEmails } = await getAllEmailsAndBreaches(user, allBreaches);
 
+  if (EXPERIMENTS_ENABLED && req.session.experimentBranch) {
+    // Growth Experiment
+    // TODO: Set session experiment variables
+  }
+
   let lastAddedEmail = null;
 
   req.session.user = await DB.setBreachesLastShownNow(user);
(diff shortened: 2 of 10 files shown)


controllers/home.js
```diff
@@ -29,6 +30,16 @@ async function home(req, res) {
...
   }
 
+  const experimentFlags = getExperimentFlags(req, EXPERIMENTS_ENABLED);
+
+  // Growth Experiment
+  if (EXPERIMENTS_ENABLED) {
+    getExperimentBranch(req, false, ["de", "fr"], {
+      "va": 25,
+      "vb": 25,
+    });
```
maxxcrawford, line 40: This uses the same syntax as [Traffic Cop](https://github.com/mozilla/trafficcop/blob/master/documentation.md#type-a-callback)

controllers/utils.js
```diff
@@ -33,71 +33,161 @@ function hasUserSignedUpForRelay(user) {
...
 
   // If URL param has experimentBranch entry, use that branch;
   if (req.query.experimentBranch) {
-    if (!["va", "vb"].includes(req.query.experimentBranch)) {
+    if (!Object.keys(variations).includes(req.query.experimentBranch)) {
       log.debug("The requested branch is unknown: ", req.query.experimentBranch);
+      session.excludeFromExperiment = true;
+      session.experimentBranch = false;
+      session.treatmentBranch = false;
       return false;
```
maxxcrawford, line 130: DRY Check: Anywhere it returns false, we set these three vars. May make sense to put this as a separate function?

controllers/utils.js
```diff
@@ -33,71 +33,161 @@ function hasUserSignedUpForRelay(user) {
...
-    log.debug("This session has been randomly assigned to the treatment group. (vb)");
-    req.session.experimentBranch = "vb";
-    return "vb";
+  if (sorterNum === false) {
+    sorterNum = Math.floor(Math.random() * 10000) + 1;
+    sorterNum = sorterNum/100;
+    // sorterNum = Math.floor(Math.random() * 100);
+    log.debug("No coinflip number provided. Coinflip number is ", sorterNum);
+  } else {
+    log.debug("Coinflip number provided. Coinflip number is ", sorterNum);
```
maxxcrawford, line 150: This else was added to add the coin flip result to the debug console.

public/js/monitor.js
```diff
@@ -434,14 +483,15 @@ function resizeDashboardMargin() {
...
   const acceptedFirstLanguageIsEnglish = acceptedLanguages[0].includes("en");
 
   if (!acceptedFirstLanguageIsEnglish && document.getElementById("fxaCheckbox")) {
-    document.getElementById("fxaCheckbox").style.display = "none";
-    return;
+    // Growth Experiment -This overrides the language filter if they've been sorted into the treatment cohort.
+    if (document.body.dataset.experiment !== "vb" || !document.body.dataset.experiment) {
+      document.getElementById("fxaCheckbox").style.display = "none";
+      return;
+    }
```
maxxcrawford, line 494: This is the crux of the experiment – if the user is on branch VB, show the checkbox.

controllers/home.js
```diff
@@ -4,7 +4,9 @@ const AppConstants = require("../app-constants");
 const DB = require("../db/DB");
 const HIBP = require("../hibp");
 const { scanResult } = require("../scan-results");
-const { generatePageToken } = require("./utils");
+const { generatePageToken, getExperimentBranch, getExperimentFlags } = require("./utils");
+
+const EXPERIMENTS_ENABLED = (AppConstants.EXPERIMENT_ACTIVE === "1");
```
groovecoder, line 9: Just to cut down on code churn from future experiments ... we could push a lot of this logic up into some kind of `ExperimentUtils.init()` module + call. That way the linter will like it but we won't have to keep adding and removing lines like this.

groovecoder: And then `ExperimentUtils.getExperimentBranch()`, `ExperimentUtils.getExperimentFlags()`, etc.

controllers/user.js
```diff
@@ -252,7 +259,7 @@ async function _verify(req) {
   unsafeBreachesForEmail = await HIBP.getBreachesForEmail(
     sha1(verifiedEmailHash.email.toLowerCase()),
     req.app.locals.breaches,
-    true,
```
groovecoder, line 255: this comma in paritcular

controllers/user.js
```diff
@@ -555,6 +562,23 @@ async function getBreachStats(req, res) {
...
+  if (EXPERIMENTS_ENABLED && req.session.experimentFlags) {
+    // Persist experimentBranch across session reset
+    const excludeFromExperiment = req.session.experimentFlags.excludeFromExperiment;
+    const experimentBranch = req.session.experimentFlags.experimentBranch;
+    req.session.reset();
+    // Reset session vars after sign out event
+    req.session.experimentFlags = {
+      experimentBranch,
+      excludeFromExperiment,
+    };
```
groovecoder, line 575: There's some clever de-structuring and re-structuring way to do this with less code, but I'm not clever.

lesleyjanenorton: Are there other experiment variables in `req.session.experimentFlags` that you want to reset or could this maybe just be something like...
```
if (EXPERIMENTS_ENABLED && req.session.experimentFlags) {
   const sessionExperimentFlags = req.session.experimentFlags
   req.session.reset();
   req.session.experimentFlags = sessionExperimentFlags
};

File: public/js/monitor.js [hunk 8]
```diff
@@ -62,6 +62,35 @@ function doOauth(el, {emailWatch = false} = {}) {
     }
   });
 
+  // Growth Experiment: OAuth Entry Point IDs are unique to the experiment.
+  const oAuthEntryPointIds = [
+    "fx-monitor-check-for-breaches-blue-btn",
```
lesleyjanenorton, line 67: maybe we should add "exp" to these so that it's easier later to sort these out.

controllers/utils.js
```diff
@@ -33,71 +33,161 @@ function hasUserSignedUpForRelay(user) {
...
+  // calculate and store total percentage of variations
+  for (const v in variations) {
+    if (variations.hasOwnProperty(v) && typeof variations[v] === "number") {
+      // multiply by 100 to allow for percentages to the hundredth
+      // (and avoid floating point math errors)
+      totalPercentage += (variations[v] * 100);
+    }
+  }
+
+  totalPercentage = totalPercentage / 100;
```
groovecoder, line 49: Looks like you could refactor this logic into a `const totalPercentage = getTotalPercentage(variations)` helper function?

controllers/user.js
maxxcrawford: @lesleyjanenorton Ohh. That's exactly it. There aren't any. We can/should just save the entire object. 🤝

controllers/user.js
maxxcrawford: 🌟🧹 Linter Magic

public/js/monitor.js
maxxcrawford: Marking this as an issue for next time. We're using these IDs in amplitude already, but we can make this change in future experiments.

maxxcrawford: Filed https://github.com/mozilla/blurts-server/issues/1737

controllers/utils.js
groovecoder: Yeah, could be this easy?
```js
return unEnrollSession(session);
...
function unEnrollSession(session) {
    session.excludeFromExperiment = true;
    session.experimentBranch = false;
    session.treatmentBranch = false;
    return false;
}
```

controllers/utils.js
```diff
@@ -33,71 +33,161 @@ function hasUserSignedUpForRelay(user) {
...
+      }
+    }
+
+    return choice;
+
+}
+
+function getExperimentBranch(req, sorterNum = false, language = false, variations) {
+
+  const session = req.session.experimentFlags;
```
groovecoder, line 83: Should be more explicit here? `const sessionFlags = req.session.experimentFlags` ?

public/js/monitor.js
```diff
@@ -62,6 +62,35 @@ function doOauth(el, {emailWatch = false} = {}) {
...
+    "fx-monitor-find-out-blue-btn",
+    "fx-monitor-alert-me-blue-btn-top",
+    "fx-monitor-alert-me-blue-btn-bottom",
+  ];
+
+  if (oAuthEntryPointIds.includes(el.dataset.entrypoint)) {
+    // Growth Experiment: Reset UTMs from in-line body tag data elements.
+    ["utm_source", "utm_medium", "utm_campaign", "utm_term", "utm_content" ].forEach(key => {
```
groovecoder, line 75: DRY: put this list of values into a constant and use it here and in `handleFormSubmits`

public/js/monitor.js
```diff
@@ -141,6 +170,26 @@ function handleFormSubmits(formEvent) {
 
   const formClassList = thisForm.classList;
 
+  // Growth Experiment
+  if (formClassList.contains("skip")) {
```
groovecoder, line 174: Should this class be more descriptive? `skipExperiment`? `experimentSkip` ?

public/js/monitor.js
```diff
@@ -141,6 +170,26 @@ function handleFormSubmits(formEvent) {
...
+  if (document.body.dataset.experiment) {
+    const scanFormActionURL = new URL(thisForm.action);
+
+    ["utm_source", "utm_medium", "utm_campaign", "utm_term", "utm_content" ].forEach(key => {
+      if (document.body.dataset[key]) {
+        scanFormActionURL.searchParams.append(key, document.body.dataset[key]);
+      }
+    });
+
+    const revisedActionURL = scanFormActionURL.pathname + scanFormActionURL.search;
```
groovecoder, line 188: Yay! `const` variables!

controllers/utils.js
```diff
@@ -33,71 +33,161 @@ function hasUserSignedUpForRelay(user) {
...
+          // tally variation percentages for the next loop iteration
+          runningTotal += variations[v];
+      }
+    }
+
+    return choice;
+
+}
+
+function getExperimentBranch(req, sorterNum = false, language = false, variations) {
```
groovecoder, line 81: Since this gets the experiment branch AND sets it into the session, can we re-name this to `setExperimentBranch`?

controllers/utils.js
lesleyjanenorton: Wondering if we really need the `excludeFromExperiment` session var or if maybe it would just be enough to check for `request.session.experimentBranch` and assume that anything without `req.session.experimentBranch` is excluded from the experiment. Or set `req.session.experiment` equal to `null`.

public/js/monitor.js
lesleyjanenorton: sounds good but if they're already being used it may just cause more grief to switch them up now in any case.

controllers/home.js
maxxcrawford: Filed! https://github.com/mozilla/blurts-server/issues/1738

controllers/utils.js
maxxcrawford: Filed! https://github.com/mozilla/blurts-server/issues/1739

PR #1749, opened 2020-06-15
Add supportedLocales to package.json and version endpoint

### LINKED ISSUE #1333: Add supported locales to stage/production /__version__ endpoint
Ref #609 and #1245 PR and https://github.com/mozilla/blurts-server/pull/1245#issuecomment-538105086

Currently the production [`/__version__`](https://monitor.firefox.com/__version__) endpoint has the following:

```js
{
  "commit": "6280822d4db00ed58a3d149e5abd51a3df443c83", 
  "version": "v9.0.2",
  "source": "https://github.com/mozilla/blurts-server",
  "build": "https://circleci.com/gh/mozilla/blurts-server/7318"
}
```

But the Heroku [`/__version__`](https://fx-breach-alerts.herokuapp.com/__version__) endpoint also has a jazzy `languages` string of supported locales:

```js
{
  "commit": "46fc20a687a39fa4588ec1934f80791cb9ed8222",
  "version": "46fc20a687a39fa4588ec1934f80791cb9ed8222",
  "source": "https://github.com/mozilla/blurts-server",
  "languages": "cs,cy,da,de,en,en-CA,es-AR,es-CL,es-ES,es-MX,fr,fy-NL,hu,ia,id,it,ja,kab,ms,nl,pt-BR,pt-PT,ro,ru,sk,sl,sq,sv-SE,tr,uk,vi,zh-CN,zh-TW"
}
```

It'd be great if we can find a way to add the "languages" string/array to the stage/production endpoints as well so QA/l10n can know exactly which languages are supported. I know we have a Top Secret OPs repo (URL redacted) with configs and the `firefoxMonitorSupportedLocales` array, but not everybody has security clearance access to that repo, and somehow shimming the `firefoxMonitorSupportedLocales` config into the `/__version__` endpoint would solve that issue and give l10n people a definitive way of seeing what locale support is currently in production, and know if the locale list is stale or not.

/cc @flodolo @Micheletto

pdehaan:
Fixes #1333


controllers/dockerflow.js
```diff
@@ -41,7 +40,7 @@ function vers (req, res) {
       commit: process.env.HEROKU_SLUG_COMMIT,
       version: process.env.HEROKU_SLUG_COMMIT,
       source: homepage,
-      languages: AppConstants.SUPPORTED_LOCALES,
+      languages: "*",
```
pdehaan, line 43: Hardcoding Heroku supported locales here.

locale-utils.js
```diff
@@ -36,7 +37,6 @@ class FluentError extends Error {
 
 const LocaleUtils = {
   init() {
-    const supportedLocales = AppConstants.SUPPORTED_LOCALES;
```
pdehaan, line 39: Cutting this since we defined it above on L13 (from ./package.json value).

package.json
```diff
@@ -133,5 +133,6 @@
     "test:integration-headless-ci": "MOZ_HEADLESS=1 ERROR_SHOTS=1 wdio tests/integration/wdio.conf.js",
     "test:integration-docker": "MOZ_HEADLESS=1 wdio tests/integration/wdio.docker.js",
     "test": "run-s test:db:migrate test:tests test:coveralls"
-  }
+  },
+  "supportedLocales": "cak,cs,cy,da,de,el,en,en-CA,en-GB,es-AR,es-CL,es-ES,es-MX,fi,fr,fy-NL,gn,hu,kab,ia,id,it,ja,nb-NO,nl,nn-NO,pt-BR,pt-PT,ro,ru,sk,sl,sq,sv-SE,tr,uk,vi,zh-CN,zh-TW"
```
pdehaan, line 137: Stole this from the current k8s config.

pdehaan:
> all emails, including breach alert - still come thru in German too.

We'd need to consult with the great oracle, @Micheletto. I'm guessing everything still works because the k8s config is still handling all the lifting. I assume there is still an OPs portion to this that would include removing the locales from the k8s config and using the ones defined in the version.json file. Otherwise we'd have two very vaguely similar concepts floating around and quickly out of sync.

Micheletto:
> We'd need to consult with the great oracle, @Micheletto. I'm guessing everything still works because the k8s config is still handling all the lifting. I assume there is still an OPs portion to this that would include removing the locales from the k8s config and using the ones defined in the version.json file. Otherwise we'd have two very vaguely similar concepts floating around and quickly out of sync.

The magic happens in the [configmap.yaml](https://github.com/mozilla-services/cloudops-infra/blob/master/projects/firefoxmonitor/k8s/charts/firefoxmonitor/templates/configmap.yaml#L46) chart, and [values.yaml ](https://github.com/mozilla-services/cloudops-infra/blob/master/projects/firefoxmonitor/k8s/charts/firefoxmonitor/values.yaml#L83) file.  We could remove that relatively easily, and you supply it in the code.  That's fine with me.

groovecoder:
> We'd need to consult with the great oracle, @Micheletto. I'm guessing everything still works because the k8s config is still handling all the lifting. I assume there is still an OPs portion to this that would include removing the locales from the k8s config and using the ones defined in the version.json file. Otherwise we'd have two very vaguely similar concepts floating around and quickly out of sync.

This was on my local `runserver` so no `k8s` involved. As far as the app is concerned, all the app code is moved off the environment variable and into the `package.json` file.

But yeah - to deploy this we will want to remove this from the k8s config too, just so we don't have un-used config floating around.

Micheletto:
> But yeah - to deploy this we will want to remove this from the k8s config too, just so we don't have un-used config floating around.

r+ then.  @groovecoder would you file a bug for that please?

groovecoder:
I'll include it in the deploy bug instructions.

Issue #1766, opened 2020-06-29
Update VPN and Proxy content


groovecoder:
VPN is becoming "Mozilla VPN", so let's update the FPN blurb copy to:

Your Internet Protocol address (IP address) can reveal your location and internet service provider. A service like Firefox Private Network hides your IP address and location for Firefox and a service like Mozilla VPN hides your IP address and location for your entire device

(Approved already by Legal and Content)

maxxcrawford:
Looks like this will require localization: 

https://github.com/mozilla/blurts-server/blob/master/locales/en/app.ftl#L14

```
-brand-fpn = Firefox Private Network
```

https://github.com/mozilla/blurts-server/blob/master/locales/en/recommendations.ftl#L60-L67
```
# Recommendation subhead
rec-ip-subhead = Use a service that masks your IP address
# Link title
rec-ip-us-cta = Try { -brand-fpn }
rec-ip-us =
  Your Internet Protocol address (IP address) pinpoints your location 
  and internet service provider. A service like { -brand-fpn } 
  masks your IP address to hide your location.
```

groovecoder:
I tried to start this, but it may need some additional template code & logic to enable 2 CTA's in a promo. @lesleyjanenorton - how hard is that do you think?

lesleyjanenorton:
Not hard but I just need more context: Why the two CTAs in a single promo?

groovecoder:
I believe this is done.

PR #1767, opened 2020-06-29
Added string needed for change on #1766. 

maxxcrawford:
This does NOT solve #1766. Once localized broadly, we can update the referenced string.

### DIFF
--- a/locales/en/app.ftl
+++ b/locales/en/app.ftl
@@ -12,6 +12,7 @@
 -brand-lockwise = Firefox Lockwise
 -brand-send = Firefox Send
 -brand-fpn = Firefox Private Network
+-brand-fpn-2 = Mozilla VPN
 ##
 
 # “account” can be localized, “Firefox” must be treated as a brand,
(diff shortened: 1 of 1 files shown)


locales/en/app.ftl
```diff
@@ -12,6 +12,7 @@
 -brand-lockwise = Firefox Lockwise
 -brand-send = Firefox Send
 -brand-fpn = Firefox Private Network
+-brand-fpn-2 = Mozilla VPN
```
flodolo, line 15: Why not `-brand-mozilla-vpn` at this point?

maxxcrawford: Good call. That'll fix any confusions later too. Updating!

groovecoder:
@flodolo - are you saying we should include the updates to the promo CTA where this new brand name will be used?

flodolo:
> @flodolo - are you saying we should include the updates to the promo CTA where this new brand name will be used?

I think so, there's not much value in exposing the new term alone, when it's not used anywhere.

lesleyjanenorton:
@maxxcrawford do you want to update `-brand-fpn-2` to `-brand-mozilla-vpn` or should we close this in favor of #1771 ?

lesleyjanenorton:
Closing in favor of #1771

PR #1769, opened 2020-06-30
Update the URL for the VPN

rlr:
The VPN moved to vpn.mozilla.org. We miss the traffic.


groovecoder:
I think the real fix is to fix #1766 with #1767. I tagged @lesleyjanenorton to see how much work it might be to update the promo template logic to include 2 CTAs.

lesleyjanenorton:
@groovecoder  I can get this done today but just need to know 1.) What are the two CTAs 2.) What are the circumstances that determine when we want to show one CTA instead of the other?

rlr:
I agree. This was meant just to be a stop gap for the real fix. We can close this.

lesleyjanenorton:
Just opened #1771 which adds the copy for the new Mozilla VPN / FPN recommendation. We can update the templates to use the new string and two CTAs once they have been localized.

PR #1780, opened 2020-07-11
fix #1148: change sessions to redis

### LINKED ISSUE #1148: Switch to server-side session storage

groovecoder:
To test this locally, will need to install and run a `redis` server. On Mac the easiest way to do this was:

* `brew install redis`
* `redis-server`

When the redis server is running, `npm start` should work fine. Then everything should work just like before.


groovecoder:
I provisioned a `REDIS_URL` on https://blurts-server-review.herokuapp.com/ and this branch works.

.env-dist
```diff
@@ -76,3 +76,6 @@ PRODUCT_PROMOS_ENABLED=1
 
 # Experiment Flag
 EXPERIMENT_ACTIVE=0
+
+REDIS_URL=redis://127.0.0.1:6379
+MAX_NUM_ADDRESSES=5
```
pdehaan, line 81: This was hard coded to 15 below. Will it still be 15 in prod, or we dropping the limit?

groovecoder: Dropping the limit to 5. 99.9% of our subscribers are below 5 addresses already.

groovecoder: Sorry, [99.7%](https://bugzilla.mozilla.org/show_bug.cgi?id=1648582#c7).

pdehaan: Interesting. What will be the impact to the 0.3% of the users who are **over** the new 5 address limit? Will they be able to keep their additional addresses, or will it randomly delete addresses so they fit within the new 5 address limit?

groovecoder: They'll be able to keep all their existing addresses, but won't be able to add any new ones.

lesleyjanenorton:
Probably user-error but I wheeling out when trying to log in locally.

groovecoder:
Worked with @lesleyjanenorton on this. Her error was un-related to the branch. Moving forward with merge + deploy of this branch.

.env-dist
Flux3PO: Well I'm one of the 0.3% and wonder why you upgrade the system to an "unlimited" system when lowering the value to 5.
I'm really dissatisfied :(

PR #1810, opened 2020-07-27
Invert sentence for not using short password

hamdouni:


template-helpers/articles.js
```diff
@@ -184,7 +184,7 @@ const articleCopy = {
             },
             {
               do: "Do make your passwords at least 8 characters long. Aim for 12-15 characters.",
-              dont: "Use short, one-word passwords, like sunshine, monkey, or football.",
+              dont: "Don't use short, one-word passwords, like sunshine, monkey, or football.",
```
groovecoder, line 187: Good catch. This is under the "Don't" column on the page, but the rest of the `dont` tips also start with "Don't" verbiage.

PR #1864, opened 2020-08-03
update nodemon for dot-prop high priority vuln

groovecoder:


package.json
```diff
@@ -60,7 +60,7 @@
     "jest": "24.9.0",
     "jest-tap-reporter": "1.9.0",
     "node-mocks-http": "1.8.1",
-    "nodemon": "1.19.4",
+    "nodemon": "^2.0.4",
```
pdehaan, line 63: This has a range operator, but the others are pinned to a specific version. But it’s green and passes the CI tests, so I dare not tempt fate by tweaking anything at this point (and I assume dependabot or whomever will pin on a future version PR).

groovecoder:
🎉

PR #1869, opened 2020-08-04
Gradaute FxA Homepage Checkbox to German/French Languages 

### LINKED ISSUE #1842: Gradaute FxA Homepage Checkbox to German/French Languages
Related: 
- #1638 
- #1722

maxxcrawford:
Fixed #1842 - Revised show/hide logic for FxA Homepage checkbox to show in English, German or French.

### DIFF
--- a/public/js/monitor.js
+++ b/public/js/monitor.js
@@ -312,6 +312,19 @@ function resizeDashboardMargin() {
   }
 }
 
+function checkIfTier1(preferredLanguage) {
+  const tier1Languages = ["de", "en", "fr"];
+  let isTier1Language = false;
+
+  tier1Languages.forEach( lang => {
+    if (preferredLanguage.includes(lang) ) {
+      isTier1Language = true;
+    }
+  });
+
+  return isTier1Language;
+}
+
 ( async() => {
   document.addEventListener("touchstart", function(){}, true);
   const win = window;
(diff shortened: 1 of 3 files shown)


public/js/monitor.js
```diff
@@ -312,6 +312,19 @@ function resizeDashboardMargin() {
...
+  const tier1Languages = ["de", "en", "fr"];
+  let isTier1Language = false;
+
+  tier1Languages.forEach( lang => {
+    if (preferredLanguage.includes(lang) ) {
+      isTier1Language = true;
+    }
+  });
+
+  return isTier1Language;
```
groovecoder, line 325: This seems to work too, right?
```suggestion
  const tier1Languages = ["de", "en", "fr"];

  tier1Languages.forEach( lang => {
    if (preferredLanguage.includes(lang) ) {
      return true;
    }
  });

  return false;
```

maxxcrawford: Sure does. Accepted!

pdehaan: I... I dont think that's how it works. I think your nested `return true` doesnt return the _outer_ function it returns the nested forEach arrow function, so you'll probably always `return false` here.

```js
console.log("ru:", test("ru"));
console.log("de:", test("de"));
console.log("en:", test("en"));
console.log("fr:", test("fr"));

function test(preferredLanguage="") {
  const tier1Languages = ["de", "en", "fr"];
  tier1Languages.forEach( lang => {
    if (preferredLanguage.includes(lang) ) {
      return true;
    }
  });
  return false;
}

/*
ru: false
de: false
en: false
fr: false
*/
```

PR #1887, opened 2020-08-14
Growth Experiment #6 - Homepage FxA Checkbox (Worldwide) 

maxxcrawford:
### Steps to test: 
- Control: https://monitor-v2.herokuapp.com/?experimentBranch=va 
- Treatment: https://monitor-v2.herokuapp.com/?experimentBranch=vb

Verify you are enrolled in the experiment by looking at the `data-experiment` attribute on `<body>` tag. 

**Treatment:**
- Go to website (with any of the listed languages set as the language)
- Click the checkbox
- It should open FxA flow.
- When you land on the /user-dashboard after completing the flow, it should still have the same meta data as mentioned above. 

#### Notes: 
- The user MUST have either German or French variant as the primary/top language. Any other deviation will be excluded from the experiment. If the first language is NOT English, then they should not get the checkbox. If English (any variant), 
-  These languages were selected from [Pontoon[(https://pontoon.mozilla.org/projects/firefox-monitor-website). Each selected language has the string ["get-email-alerts" from app.ftl](https://github.com/mozilla/blurts-server/blob/master/locales/en/app.ftl#L284) translated and has at least 75% of all the strings translated. _(As of Aug 14, 2020)_
- This test applies to every where BUT Tier 1 languages (EN, DE, and FR). If in one of those locales, the default state is now to HAVE a checkbox. (Once #1869 is merged) 
- Each language entry includes variations of that language, so all Spanish versions (Spain, Mexico, etc) are included by including only "ES"
- From the top 20 languages, this test covers all of them except for Polish (14) and Japanese (15)

### DIFF
--- a/controllers/home.js
+++ b/controllers/home.js
@@ -6,6 +6,7 @@ const HIBP = require("../hibp");
 const { scanResult } = require("../scan-results");
 const {
   generatePageToken,
+  getExperimentBranch,
   getExperimentFlags,
   getUTMContents,
 } = require("./utils");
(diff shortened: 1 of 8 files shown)


groovecoder:
Some conflicts here need to be reconciled.

groovecoder:
Code LGTM and spot-check works. Branch needs conflict resolutions.

maxxcrawford:
@groovecoder We need to hold on this until the active experiment comes down (#1897)  and then rebase on it.

PR #1957, opened 2020-10-06
RECRUITMENT_BANNER_* for a site-wide recruitment banner

groovecoder:
Doesn't look great, but it could help us use Monitor for some quick research links. (Surveys, panel recruitment, etc.)

image


views/partials/header/header.hbs
```diff
@@ -1,4 +1,7 @@
     <header id="header" class="{{ addUpgradeClasses }}">
+      {{#if RECRUITMENT_BANNER_LINK }}
+      <div class="recruitment-banner"><a href="{{ RECRUITMENT_BANNER_LINK }}" target="_blank" rel="noopener noreferrer">{{ RECRUITMENT_BANNER_TEXT }}</a></div>
```
pdehaan, line 3: No localization?

groovecoder: Hmm ... if we have/had a generic recruitment text we could localize that. @tcinotto - should we have a default recruitment text string? Or should each recruitment link run have custom text?

pdehaan: Either that, or maybe have custom strings and get them pre-translated in pontoon before publishing a survey. The `RECRUITMENT_BANNER_TEXT` could be used as a pontoon id.
Not sure if that'd be enough to get us localized banners, and fallback to en-US if a translation is missing.

I guess neither solution would work if we wanted to exclude some banners from specific locales (if we were promoting the VPN or something).

tcinotto:
@groovecoder Can we make this banner show only on en locales? I would imagine our survey and research would only be in English only. I think it's probably better to have the string configurable, just incase there are any needs from the research team to be more specific.

groovecoder:
Added code to restrict the recruitment banner to browser that have English in their `Accept-Language`

views/partials/header/header.hbs
groovecoder: Going with english-only recruitment right now.

pdehaan:
Verified locally and things looked OK at various viewport resolutions in Firefox+Chrome.

I did notice that the email field on the homepage overlaps the headers, but it doesn't seem related to this PR (since I can reproduce it on production, but only when I use the Private Relay add-on):
Firefox_Monitor

My only other minor concern would be if the text looks clickable enough. It isn't underlined or very button looking, so not sure if users will think its an informational statement or a clickable element.
But I'm fine with merging the new functionality now and tweaking the UI in the future.

Another thing I wasn't able to adequately test was the behavior on a non-EN browser. I installed FR and DE locales in my Firefox Beta, but it seemingly refuses to let me delete my en-CA locale, so it always showed the header, regardless of what locale the site was using.

Einstellungen

Firefox_Monitor

template-helpers/hbs-helpers.js
```diff
@@ -21,6 +21,25 @@ function getSupportedLocales(args) {
 }
 
 
+function englishInAcceptLanguages(args) {
+  const acceptedLanguages = args.data.root.req.acceptsLanguages();
+  return acceptedLanguages.some(locale => locale.startsWith("en"));
+}
```
pdehaan, line 27: Confirmed by hacking my `intl.accept_languages` in about:config

Erweiterte_Einstellungen

template-helpers/hbs-helpers.js
```diff
@@ -21,6 +21,25 @@ function getSupportedLocales(args) {
...
+function recruitmentBanner(args) {
+  if (!AppConstants.RECRUITMENT_BANNER_LINK || !AppConstants.RECRUITMENT_BANNER_TEXT) {
+    return;
+  }
+
+  if (!englishInAcceptLanguages(args)) {
+    return;
+  }
+
+  return `<div class="recruitment-banner"><a id="recruitment-banner" href="${AppConstants.RECRUITMENT_BANNER_LINK}"  target="_blank" rel="noopener noreferrer" data-ga-link="" data-event-category="Recruitment" data-event-label="${AppConstants.RECRUITMENT_BANNER_TEXT}">${AppConstants.RECRUITMENT_BANNER_TEXT}</a></div>`;
```
pdehaan, line 39: Is it OK that `data-event-label` will have spaces and other stuff?

Also, are we expecting that `RECRUITMENT_BANNER_TEXT` is always just a simple string and will never include markup (like a `<br>` or `<i>` or `<strong>`), or any double-quotes (which would probably just throw up right around this attribute -- since we aren't doing any escaping).

groovecoder: I'll have it escape `"` and `'` chars here and we'll have to remember this about the text.

groovecoder:
@pdehaan @maxxcrawford @lesleyjanenorton this look good enough to merge now?

pdehaan:
Sorry, silly Q, but is this PR essentially the same as the https://github.com/mozilla/fx-private-relay/pull/707 PR?

Q: Do the banners need to look the same on each site? Should they?

groovecoder:
Different banners on the sites seem fine. The sites don't even link to each other yet.

PR #1979, opened 2020-11-06
fix #1868: use initialization vectors for ciphers

### LINKED ISSUE #1868: 'crypto.createCipher' was deprecated since v10.0.0. Use 'crypto.createCipheriv()' instead
via #1853 logs

```sh
/home/travis/build/mozilla/blurts-server/controllers/scan.js
  15:20  error    'crypto.createDecipher' was deprecated since v10.0.0. Use 'crypto.createDecipheriv()' instead  node/no-deprecated-api

/home/travis/build/mozilla/blurts-server/controllers/utils.js
  13:18  error    'crypto.createCipher' was deprecated since v10.0.0. Use 'crypto.createCipheriv()' instead  node/no-deprecated-api
```

stanleykerr:
# Summary
This fixes the usage of the deprecated Crypto.createCipher/Crypto.createDecipher, which throws a warning as shown in #1868.

# Explanation
I implemented the usage of the Crypto.createCipheriv/Crypto.createDecipheriv functions that use non-static initialization vectors (IV), as well as using the Crypto implementation for pbkdf2<sup>[[1]](#notes)</sup> to generate the key based on the password. The password used is AppConstants.COOKIE_SECRET, which should be at least 64 bits per [PBKDF2's specification sheet](https://tools.ietf.org/html/rfc8018#section-4), but 128 bits is [recommended by the NIST](https://nvlpubs.nist.gov/nistpubs/Legacy/SP/nistspecialpublication800-132.pdf).

The page token is now generated as "`base64(<IV>).base64(<Encrypted Page Token>)`"

This allows the decryption function to check for the separator in the page token string, where it will use the IV to decrypt the page token. If it isn't found, it will use the Crypto.createDecipher for users that have the tokens generated by the old algorithm, although it may be unnecessary to keep this later on, as new page tokens won't be generated in that format.

###### <b name="notes">[1]</b> Note: Another KDF algorithm such as scrypt can also be used over pbkdf2, as it's more secure, although it could be more costly to compute, and might not offer many benefits based on the pertaining data.


controllers/scan.js
```diff
@@ -12,9 +12,25 @@ const sha1 = require("../sha1-utils");
 const log = mozlog("controllers.scan");
 
 function _decryptPageToken(encryptedPageToken) {
-  const decipher = crypto.createDecipher("aes-256-cbc", AppConstants.COOKIE_SECRET);
-  const decryptedPageToken = JSON.parse([decipher.update(encryptedPageToken, "base64", "utf8"), decipher.final("utf8")].join(""));
-  return decryptedPageToken;
+  let decipher;
```
groovecoder, line 15: `let` always makes me nervous. Can we refactor into a more pure function? Maybe:

```js
function _getDecipherAndToken(encryptedPageToken) {
  if (encryptedPageToken.slice(24, 25) === ".") {
    // iv becomes 24 characters long when represented as base64: ceil(16 / 3) * 4 = 24
    const iv = Buffer.from(encryptedPageToken.slice(0, 24), "base64");
    const key = crypto.pbkdf2Sync(AppConstants.COOKIE_SECRET, iv.toString(), 10000, 32, "sha512");
    return {
      decipher: crypto.createDecipheriv("aes-256-cbc", key, iv),
      encryptedToken: encryptedPageToken.slice(25),
    };
  }
  return {
    decipher: crypto.createDecipher("aes-256-cbc", AppConstants.COOKIE_SECRET),
    encryptedToken: encryptedToken,
  }
}

function _decryptPageToken(encryptedPageToken) {
  const {decipher, encryptedToken} = _getDecipherAndToken(encryptedPageToken);

  const decryptedPageToken = Buffer.concat([
    decipher.update(Buffer.from(encryptedToken, "base64")),
    decipher.final(),
  ]).toString("utf8");

  return JSON.parse(decryptedPageToken);
}
```

controllers/utils.js
```diff
@@ -9,10 +9,18 @@ const AppConstants = require("../app-constants");
 
 
 function generatePageToken(req) {
-  const pageToken = {ip: req.ip, date: new Date(), nonce: uuidv4()};
-  const cipher = crypto.createCipher("aes-256-cbc", AppConstants.COOKIE_SECRET);
-  const encryptedPageToken = [cipher.update(JSON.stringify(pageToken), "utf8", "base64"), cipher.final("base64")].join("");
-  return encryptedPageToken;
+  const pageToken = { ip: req.ip, date: new Date(), nonce: uuidv4() };
```
groovecoder, line 12: Not sure we need the `nonce` anymore - if I remember correctly, I was using it as make-shift 128-bit "salt" for each token. But a 16-byte IV should have the same effect?

controllers/scan.js
```diff
@@ -12,9 +12,25 @@ const sha1 = require("../sha1-utils");
...
+
+  if (encryptedPageToken.slice(24, 25) === ".") {
+    // iv becomes 24 characters long when represented as base64: ceil(16 / 3) * 4 = 24
+    const iv = Buffer.from(encryptedPageToken.slice(0, 24), "base64");
+    const key = crypto.pbkdf2Sync(AppConstants.COOKIE_SECRET, iv.toString(), 10000, 32, "sha512");
+
+    decipher = crypto.createDecipheriv("aes-256-cbc", key, iv);
+    encryptedPageToken = encryptedPageToken.slice(25);
+  } else {
+    decipher = crypto.createDecipher("aes-256-cbc", AppConstants.COOKIE_SECRET);
```
groovecoder, line 25: I assume this could be removed later? E.g., after the server has processed all outstanding requests that are sending with an "old" token?

groovecoder (approved):
Code LGTM and works fine. A few comments that are non-blocking.

controllers/scan.js
stanleykerr: Yeah, there will be no need for the else statement after all "old" tokens are gone. The if could still be used in order to ensure the token is the right format. We may also want to wrap the deciphering in a try/catch just in case input other than properly formatted/encrypted data is sent in a request.

Issue #1581, opened 2020-03-12
Update "0 Breaches Resolved" Placeholder Image


changecourse:
Per a conversation with @betsymi... "When I logged into Monitor and first saw breach resolution in context of my dashboard, I found it a bit jarring. I'm concerned that it looks like a warning and could be giving users the wrong signal — that this is something bad, not good. Because we're all about breaches, we should avoid anything that might raise alarm."  

Old placeholder: 
![image(2)](https://user-images.githubusercontent.com/6164801/76544164-356fcc00-644d-11ea-91e9-4b7c245a9842.png)

Updated placeholder image:
<img width="741" alt="Screen Shot 2020-03-12 at 10 37 19 AM" src="https://user-images.githubusercontent.com/6164801/76544344-7bc52b00-644d-11ea-97f1-ee4aedcf2cbb.png">

Assets:
[assets.zip](https://github.com/mozilla/blurts-server/files/4325400/assets.zip)

trevorb1637:
@changecourse I would like to work on this issue as my first contribution

Vinnl:
This image is no longer present on the current website, so I'm closing this issue.

PR #1980, opened 2020-11-06
#1581 issue update

trevorb1637:
In reference to issue #1581, I have updated the svg image with the new breach resolution intro icon


lesleyjanenorton (approved):
Thank you @trevorb1637 !

PR #2018, opened 2021-01-21
Fix #2011 - Add logic to not overwrite already set URL params

### LINKED ISSUE #2011: Incorrectly structured query to FxA /metrics endpoint
👋 Hello!

Was just poking through our sentry errors for FxA and noticed a validation error that may be related to some code here...

`child "utm_source" fails because ["utm_source" with value "about-protections,about-protections" fails to match the required pattern: /^[\w\/.%-]+$/], `

``` json
{
  entrypoint: fx-monitor-check-for-breaches-blue-btn,
  entrypoint_experiment: none,
  entrypoint_variation: none, 
  utm_campaign: none,
  utm_content: none,
  utm_medium: none,
  utm_source: about-protections,about-protections,
  utm_term: none
}
```
The `entrypoint` made it through, which points us to:
https://github.com/mozilla/blurts-server/blob/0f59faa7142de7d3c11f93cb31ecf6df5dae97b4/views/partials/forms/scan-email-form.hbs#L26

I'm not sure if the bug actually lies here, but somewhere between `about:protections` and FxA `/metrics` endpoint something is getting messed up.

maxxcrawford:
## Testing Steps

- Start the URL with an entry point and utm_source alreay set: 
  - Go to `localhost:6060/?entrypoint=protection_report_monitor&utm_source=about-protections`
- Complete the form on the homepage with the checkbox checked (to login/create a new FxA account) 
- On the Accounts page, check the URL. It should have only one `utm_source` entry. (From this example, it should be set to `about-protections`)

### DIFF
--- a/public/js/fxa-analytics.js
+++ b/public/js/fxa-analytics.js
@@ -37,7 +37,10 @@ async function sendPing(el, eventAction, eventLabel = null, options = null) {
 
 function appendFxaParams(url, storageObject) {
   getUTMNames().forEach(param => {
-    if (storageObject[param]) {
+    if (storageObject[param] && !url.searchParams.get(param)) {
+      // Bug #2011 - This logic only allows params to be set/passed
+      // on to FxA isn't that param isn't already set.
+      // (Example: Overwriting a utm_source)
       url.searchParams.append(param, encodeURIComponent(storageObject[param]));
     }
   });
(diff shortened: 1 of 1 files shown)


public/js/fxa-analytics.js
```diff
@@ -37,7 +37,10 @@ async function sendPing(el, eventAction, eventLabel = null, options = null) {
 
 function appendFxaParams(url, storageObject) {
   getUTMNames().forEach(param => {
-    if (storageObject[param]) {
+    if (storageObject[param] && !url.searchParams.get(param)) {
+      // Bug #2011 - This logic only allows params to be set/passed
+      // on to FxA isn't that param isn't already set.
```
groovecoder, line 42:
```suggestion
      // on to FxA if that param isn't already set.
```

groovecoder (changes requested):

groovecoder (approved):

PR #2033, opened 2021-02-02
add relay as a recommendation

groovecoder:


flodolo (commented):
Can you help me understand why two sets of identical messages? How does the context change, i.e. are they used in different places?

groovecoder:
> Can you help me understand why two sets of identical messages? How does the context change, i.e. are they used in different places?

Sorry, I thought I had to use 2 different messages to show this in 2 separate places.

I consolidated it down to 1 message.

groovecoder:
The difference in context is only on the back-end. This message is shown as a recommendation on a breach detail page:

Some breaches included "priority" data categories, so we have specific recommendations:

image

For breaches that have no priority data classes, we show "general" recommendations:

image

So the contextual difference to the end user is very small. It's mostly a technical detail of which recommendations we show on which breaches.

flodolo (approved):

maxxcrawford:
Added icon for the new recommendation: 
image

PR #2043, opened 2021-02-22
start code for supr-q and pmf microsurveys

groovecoder:


maxxcrawford:
Added commit with additional styles, including hiding the survey on screens smaller than `700px`. 

Screenshots: 
image

image

template-helpers/hbs-helpers.js
```diff
@@ -44,6 +44,71 @@ function recruitmentBanner(args) {
...
+
+  // don't show micro survey if user is not signed in
+  if (!args.data.root.req.session.user){
+    return;
+  }
+
+  const bannerOpeningDiv = "<div id=\"micro-survey-banner\" class=\"micro-survey-banner hidden\">";
+  const nowSecond = new Date().getTime() / 1000;
+  let surveyElements;
+  switch (nowSecond) {
```
pdehaan, line 67: How does `nowSecond` work? I tried the snippet above in DevTools and `new Date().getTime() / 1000` returned `1614107470.262`, which would fail every `switch..case` below (except `default`). Did we only want to look at the last digit of the time?

maxxcrawford: Can we wrap it in a `Math.floor()` or something?

groovecoder:
@pdehaan @maxxcrawford I think this is ready and good to go. Agree?

maxxcrawford (approved):

PR #2076, opened 2021-03-24
Add Astoria breach logo

maxxcrawford:
Preview: 
Astoria


maxxcrawford:
Note – The artifacting around this logo made it noisy/too large to clip. Keeping the background on this one.

pdehaan:
For anybody else curious, looks like the Astoria breach appears on the homepage as the latest breach, but not on the /breaches/ page; presumably because the breach is flagged as `IsVerified=false` in https://haveibeenpwned.com/api/v3/breach/Astoria

And we filter out non-verified breaches on L165 below
https://github.com/mozilla/blurts-server/blob/0f59faa7142de7d3c11f93cb31ecf6df5dae97b4/hibp.js#L160-L168

PR #2077, opened 2021-03-26
remove data waitlist page experiment

groovecoder:
This re-uses a bunch of old Relay waitlist page experiment code in pretty dumb ways, so we probably want to re-name at least a bunch of elements and styles before we would merge and launch this.

### DIFF
--- a/public/css/remove-data.css
+++ b/public/css/remove-data.css
@@ -351,7 +351,7 @@
   }
 
   .relay-toggle-button:hover {
-    background-color: rgba(255, 255, 255, 0.1);
+    background-color: rgba(0, 0, 0, 0.1);
     transition: all 0.2s ease;
   }
 
--- a/public/css/remove-data.css
+++ b/public/css/remove-data.css
@@ -1,0 +1,421 @@
+
+.txt-light-grey {
+  color: rgb(237, 237, 240);
+}
+
+.txt-white {
+  color: rgba(255, 255, 255, 1);
+}
+
+.relay-hero {
+  margin: auto auto 60px auto;
+  max-width: 700px;
+}
+
+.relay-headline {
+  font-size: 44px;
+  line-height: 46px;
+  margin: 0 auto 12px auto;
+}
+
+.relay-headline::before {
+  content: "";
+  display: block;
+  height: 150px;
+  width: 100%;
+  background-image: url("/img/svg/data-removal-large.svg");
+  background-repeat: no-repeat;
+  background-position: center center;
+  background-size: contain;
+  margin: 60px auto 40px auto;
+}
+
+.relay-subhead {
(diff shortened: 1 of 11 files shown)


public/js/monitor.js
```diff
@@ -336,6 +336,80 @@ function recruitmentLogic() {
...
+          body: JSON.stringify({"emailToAdd": "add-user-email"}),
+        });
+        if (response && response.status === 200) {
+          setTimeout(()=> {
+            signUpCallout.classList.add("email-sent");
+            signUpCallout.classList.remove("sending-email");
+          }, 500);
+        }
+      } catch(e) {
+        // we need error messaging
```
maxxcrawford, line 363: Yes, we do.

public/js/monitor.js
```diff
@@ -336,6 +336,80 @@ function recruitmentLogic() {
...
+}
+
+function addWaitlistObservers() {
+
+  const privateRelayCtas = document.querySelectorAll(".private-relay-cta");
+
+  if (privateRelayCtas.length === 0) {
+    return;
+  }
+  const availableIntersectionObserver = ("IntersectionObserver" in window);
```
maxxcrawford, line 376: I had a question about adding IntersectionObserver from a [browser compat](https://github.com/mozilla/blurts-server/blame/f80d96d09b4d6dc2bdb997c9af1612a84a37f5e1/public/js/resolve-breaches.js#L51) stand point, but looks like we have had it [on the site for a while](https://github.com/mozilla/blurts-server/blame/f80d96d09b4d6dc2bdb997c9af1612a84a37f5e1/public/js/resolve-breaches.js#L51) without issue? 

Can you confirm that Monitor does not support browsers that don't support this? (Namely, IE11)

public/css/remove-data.css
```diff
@@ -0,0 +1,421 @@
...
+    margin: 0;
+  }
+
+  .relay-toggle-button {
+    padding-left: 16px;
+    pointer-events: all;
+  }
+
+  .relay-toggle-button:hover {
+    background-color: rgba(255, 255, 255, 0.1);
```
maxxcrawford, line 354: This background appears brighter on mobile. I would recommend going darker instead of lighter. 

```suggestion
    background-color: rgba(0, 0, 0, 0.1);
```

image

maxxcrawford (approved):
Tested on Firefox stable, Chrome stable and iOS and Android (Firefox Mobile) 

I only ran into one issue: When opening up one of the FAQs on mobile, the highlight appears stronger than on desktop. See comment for more info. 

Gooooood to go!

public/js/monitor.js
groovecoder: Confirm.

PR #2188, opened 2021-06-21
fix #2123: don't redirect /oauth/confirmed

### LINKED ISSUE #2123: FxA UTM Params Login Issue
Found an odd bug related to a change in FxA. FxA now passes through UTM parameters during login flow, Monitor apparently was not ready for this, and will redirect on oauth confirmation back to FxA when UTM parameters are intact. When I manually removed the UTM params from the URL and finished FxA signin, then monitor was ok with it.

**Impact: This means clicking the Resolve links in Monitor emails will never drop you on the Monitor page if you need to sign-in, as there'll be a redirect loop.**

groovecoder:


maxxcrawford (approved):
Works!

PR #2197, opened 2021-06-24
Fix #2193 - Add new pricing copy to the VPN promo banner

### LINKED ISSUE #2193: Replace Firefox Monitor: Get Mozilla VPN banner with 4.99 pricing special banner
We are running a new campaign for Mozilla VPN, before we end our introductory pricing and need to replace the banner on Firefox Monitor.

Link to asset: https://drive.google.com/drive/folders/16ghFYfkLYPGm60ZqSacXS-eMCU5ItAXs?usp=sharing

This banner should only go live in US, UK, CA, Singapore, New Zealand, and Malaysia in English only. 

The banner should be dismissable.

UTM parameter: https://www.mozilla.org/products/vpn/?utm_source=firefox-monitor&utm_medium=firefox-monitor&utm_campaign=intro-pricing 

Asset should stay live until July 13th then be removed. I'll file another issue for changing the asset back to the existing asset.

maxxcrawford:
Note – These strings are intentionally not being translated. It is only for the next few weeks and will revert back to the previous text.

### DIFF
--- a/public/css/vpn-promo-banner.css
+++ b/public/css/vpn-promo-banner.css
@@ -31,8 +31,12 @@ body {
   max-height: 0;
 }
 
+.vpn-promo-banner + .recruitment-banner {
+  display: none;
+}
+
 .vpn-promo-banner-wrapper {
-  max-width: 900px;
+  max-width: 1000px;
   margin: 0 auto;
   padding: 1rem 2rem;
   text-align: center;
@@ -50,6 +54,11 @@ body {
   max-width: 320px;
   margin: 0 auto 1rem;
   line-height: 1.2rem;
+}
+
+.vpn-promo-copy strong {
+  display: block;
+  margin: 0 0 2px;
 }
 
 .vpn-promo-banner a {
@@ -138,7 +147,8 @@ body {
 
 @media screen and (min-width: 800px) {
   .vpn-promo-copy {
-    max-width: 400px;
+    max-width: none;
+    padding: 0 1rem;
   }
   .vpn-banner-visible .clear-header {
     padding-top: 16rem;
@@ -149,5 +159,8 @@ body {
   .vpn-promo-banner button {
     top: 0.25rem;
     right: 0.25rem;
+  }
+  .vpn-promo-banner a {
+    flex-shrink: 0;
   }
 }
--- a/public/css/vpn-promo-banner.css
+++ b/public/css/vpn-promo-banner.css
@@ -56,9 +56,18 @@ body {
   line-height: 1.2rem;
 }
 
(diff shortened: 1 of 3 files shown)


public/css/vpn-promo-banner.css
```diff
@@ -31,8 +31,12 @@ body {
   max-height: 0;
 }
 
+.vpn-promo-banner + .recruitment-banner {
```
maxxcrawford, line 34: This statement only allows for one "banner" to be visible at a time.

public/css/vpn-promo-banner.css
```diff
@@ -150,4 +160,7 @@ body {
     top: 0.25rem;
     right: 0.25rem;
   }
+  .vpn-promo-banner a {
+    flex-shrink: 0;
+  }
```
maxxcrawford, line 165: Added to adjust the design better for non-English languages. It allows the button copy to stay on one-line when on larger screens.

public/js/monitor.js
```diff
@@ -528,10 +542,14 @@ function vpnBannerLogic() {
...
   // Only show banner if users first language is some English-locale variant
-  if (["en", "de", "fr"].some(lang=>preferredLanguages[0].includes(lang))) {
+  if (["de", "fr"].some(lang=>preferredLanguages[0].includes(lang))) {
     vpnBannerLogic();
   }
 
+  if (["en"].some(lang=>preferredLanguages[0].includes(lang))) {
+    vpnBannerLogic({isEnglish: true});
+  }
+
```
maxxcrawford, line 552: This logic is just reused from above. If there's a better way to do this without the `.some`, let me know.

views/partials/vpn-promo.hbs
```diff
@@ -5,7 +5,7 @@
 			<strong>{{ getString "vpn-promo-headline" }}</strong>
 			<span>{{ getString "vpn-promo-copy" }}</span>
 		</div>
-		<a id="" href="https://www.mozilla.org/products/vpn/?utm_source=monitor&utm_medium=monitor&utm_campaign=mozillavpn-monitor-top-banner">{{ getString "vpn-promo-cta" }}</a>
+		<a class="vpn-promo-cta" href="https://www.mozilla.org/products/vpn/?utm_source=monitor&utm_medium=monitor&utm_campaign=mozillavpn-monitor-top-banner">{{ getString "vpn-promo-cta" }}</a>
```
maxxcrawford, line 8: Added this class for CSS targeting (and removed the empty `id`).

groovecoder (approved):
Code looks good. Did not spot check though.

Issue #2210, opened 2021-07-13
UI Tooling Updates


maxxcrawford:
- [x] Add/Copy [Protocol Library](https://www.npmjs.com/package/@mozilla-protocol/core) via NPM
- [x] Migrate all CSS to SCSS, using gulp to compile
- [x] Update `gitignore` to ignore compiled CSS
- [x] Update docs
- [x] Update stylelint for SCSS

PR #2215, opened 2021-07-15
idea: only add gulp & scss to intermediate build image

groovecoder:
Just an idea ... can we use the same technique we use in Relay, where we:

1. Move all the gulp & SCSS dependencies to `devDependencies`
2. Add an intermediate docker `builder` image just for building the statics
3. Copy the build statics into the final app image
4. Only install the production-necessary node modules in the final app image

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,3 +1,11 @@
+FROM node:10-alpine AS builder
+WORKDIR /app
+COPY package*.json ./
+COPY gulpfile.js ./
+COPY public ./public/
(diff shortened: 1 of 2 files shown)


maxxcrawford (commented):
Quick note – I'm commandeering this PR. Caught it back up with `main` since #2210 merged. Testing on v2 now.

toufali:
Closing this as our build-phase now uses ESBuild and without Sass.  Cool idea tho!

PR #2216, opened 2021-07-19
Added Vastaamo breach logo (ignore)

codemist:
Added the Vastaamo breach logo.

**Preview:** 
Vastaamo

cc: @pdehaan @maxxcrawford


maxxcrawford (changes requested):
This image is 250x250. Can you revise it to be 150x150?

PR #2221, opened 2021-07-21
Adding licence-plates from recent breach

pdehaan:
See https://monitor.firefox.com/breach-details/ParkMobile

### DIFF
--- a/locales/en/data-classes.ftl
+++ b/locales/en/data-classes.ftl
@@ -66,7 +66,8 @@ instant-messenger-identities = Instant messenger identities
 ip-addresses = IP addresses
 job-applications = Job applications
 job-titles = Job titles
-licence-plates = Licence Plates
+# This string refers to vehicle license plates. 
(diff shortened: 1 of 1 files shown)


locales/en/data-classes.ftl
```diff
@@ -66,6 +66,7 @@ instant-messenger-identities = Instant messenger identities
 ip-addresses = IP addresses
 job-applications = Job applications
 job-titles = Job titles
+licence-plates = Licence Plates
```
flodolo, line 69: I know it's silly, but can we add a note that these reference vehicles? Also, what's up with the British spelling?

pdehaan: Added a note. Ha! I just copy-pasted the strong from the JSON file. I think Troy is Australian, so he probably uses the British version (same as Canada). Good catch!

flodolo (approved):

PR #2229, opened 2021-07-22
fix #2228 - dev server restart loop during build

### LINKED ISSUE #2228: Fix dev server restart during build-phase
Currently, Nodemon is set up to watch javascript files and restart the dev server on-change.  This triggers ~20 plus server restarts during the Gulp build-phase, as all the files are generated into `public/`.

We should throttle or restrict Nodemon's watch to avoid unnecessary server restarts.  The server restarts could cause issues, but in the least clutter the build-phase terminal output which is necessary for debugging.

toufali:
Nodemon should not care about files in `public/`.  By ignoring all the files built in that folder, we avoid the many server restarts triggered by Nodemon's watch.

### DIFF
--- a/package.json
+++ b/package.json
@@ -132,7 +132,7 @@
     "lint:css": "stylelint --fix public/scss/ public/css/app.css",
     "lint:audit": "npm-audit-ci-wrapper -t 'high'",
     "get-hashsets": "node scripts/get-hashsets",
-    "server": "NODE_ICU_DATA=./node_modules/full-icu nodemon server.js",
+    "server": "NODE_ICU_DATA=./node_modules/full-icu nodemon server.js --ignore public/",
     "start": "run-p build:all watch:all server",
     "test:db:migrate": "NODE_ENV=tests knex migrate:latest --knexfile db/knexfile.js --env tests",
     "test:tests": "NODE_ENV=tests HIBP_THROTTLE_DELAY=1000 HIBP_THROTTLE_MAX_TRIES=3 jest --runInBand --coverage tests/",


groovecoder (commented):
Looks good for the `npm run server` command. But I'm not sure that was the script causing re-compiles before? `npm start` still causes re-compiles on this branch. I think we need to get nodemon AND gulp to ignore the public dir?

maxxcrawford:
Gulp is only watching the SCSS folder for changes, so Gulp shouldn't be trigger any extra actions – however, when you run `gulp default`, it runs a build step and then the watch step, which…triggers one extra build. So, the two SCSS builds are expected from gulp.

package.json
```diff
@@ -132,7 +132,7 @@
     "lint:css": "stylelint --fix public/scss/ public/css/app.css",
     "lint:audit": "npm-audit-ci-wrapper -t 'high'",
     "get-hashsets": "node scripts/get-hashsets",
-    "server": "NODE_ICU_DATA=./node_modules/full-icu nodemon server.js",
+    "server": "NODE_ICU_DATA=./node_modules/full-icu nodemon server.js --ignore public/",
     "start": "run-p build:all watch:all server",
```
maxxcrawford, line 136: Calling this issue out while checking – running `build:all` and `watch:all` in parallel may be another issue (at least in watching the stream that follows `npm start`)

I was playing with it locally, and by dropping `watch:all` from the `start` function, and instead putting into the `build:dev` function, it makes the output ot the `build`/`watch` function more logical.

maxxcrawford (approved):
@toufali Thanks for jumping on this! I had one more optimization but it can be punted to a later time. This is a great, thoughtful improvement!

toufali:
> Looks good for the `npm run server` command. But I'm not sure that was the script causing re-compiles before? `npm start` still causes re-compiles on this branch. I think we need to get nodemon AND gulp to ignore the public dir?

@groovecoder, I might be missing something, but `npm start` calls `npm run server`.  Without the fix above I get many nodemon-specific restarts in terminal:
Screen Shot 2021-07-23 at 9 08 41 AM

(The above shows only some of the restarts -- there are many more throughout the logs)

With the fix in place running `npm start` my terminal shows all that goes away – I see almost the entire log on 1 screen with no nodemon-specific restarts.

Can you let me know if you see something different?

package.json
toufali: Agreed @maxxcrawford , this was meant to be a "touch as little as possible" fix as I get up to speed.  As a larger exercise, we could look into refactoring the entire build process with more modern practices – a tall order!  This fix is low-hanging fruit.  🍒

maxxcrawford: Agreed. Let's start with this PR!

PR #2234, opened 2021-07-23
CSS visual fixes

maxxcrawford:
I ran an audit from the compiled CSS from [v13.4.8](https://github.com/mozilla/blurts-server/releases/tag/v13.4.8) against the compiled CSS. I made a few adjustments and now when comparing the dev site version prod, it's much much closer. 

- Reordered when the main SCSS file (previously app.css) gets compiled. (Previously, it was the second file to be concat'd, so I moved it back to that position) 

Notes:
- The new CSS includes additional font weights. (Ex: The main navigation had a font-weight of `600`. Previously, bc of which font weights were available, it defaulted to `700`. Now that `600` is present, the weight changed. 

STR: 
- Open both https://monitor.firefox.com/ and https://monitor-v2.herokuapp.com/
- **Expected:** Both sites should look visually the same

### DIFF
--- a/public/scss/app.scss
+++ b/public/scss/app.scss
@@ -14,6 +14,7 @@
 
 // Previous Styles
 @import "partials/all-breaches.scss";
+@import "partials/main.scss";
 @import "partials/articles.scss";
 @import "partials/breach-cards.scss";
 @import "partials/breach-detail.scss";
@@ -27,7 +28,6 @@
 @import "partials/fx-bento.scss";
 @import "partials/header.scss";
 @import "partials/latest-breach.scss";
-@import "partials/main.scss";
 @import "partials/monitor.scss";
 @import "partials/product-promos.scss";
 @import "partials/remove-data.scss";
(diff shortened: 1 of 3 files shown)


public/scss/app.scss
```diff
@@ -14,6 +14,7 @@
 
 // Previous Styles
 @import "partials/all-breaches.scss";
+@import "partials/main.scss";
```
pdehaan, line 17: Random Q: But why this shuffle?

pdehaan (approved):
Compared the two sites and they indeed looked similar.
Tried comparing to https://fx-breach-alerts.herokuapp.com/, but that currently looks bricked w/ an application error.

public/scss/app.scss
pdehaan: Nevermind, I just saw your note in the commit msg:
> Reordered when the main SCSS file (previously app.css) gets compiled. (Previously, it was the second file to be concat'd, so I moved it back to that position)

PR #2240, opened 2021-08-02
upgrade to node 14

toufali:
Opening this for testing since I'm not as familiar with the inner-workings of the app.  On the surface, all seems to be working as expected.


toufali:
Right off the bat I see integration test fail -- I'll take a look at the docker setup.

toufali:
Upgraded node-postgres to resolve Knex error: https://github.com/knex/knex/issues/3836

toufali:
Latest progress: CircleCI integration test is failing.  The server log keeps suggesting incorrect PORT var or that it cannot find .env.  I tried a couple solutions to no avail including:
- adding PORT to CircleCI .env
- fixing the syntax of the command (supposed to be in quotes, according to docs)
- explicitly specifying `--env-file`

Any thoughts/ideas welcome!

toufali:
The `PORT` var error was misleading the integration test was failing because NVM was not correctly using v14.  Corrected this by explicitly installing 14 and making it default, instead of just `use` command.

toufali:
The major pain point for this PR was an outdated image for https://hub.docker.com/r/b4handjr/selenium-firefox .  After working with that image's code-owner, I was able to resolve integration test fails.  

It was cleaner to start over, so I'm closing this in favor of https://github.com/mozilla/blurts-server/pull/2244

PR #2244, opened 2021-08-06
upgrade to Node 14

toufali:
This PR upgrades Node from version 10 to 14 (current LTS).  This should enable package-updates to resolve vulnerabilities, and at the same time allow the latest ES features.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -71,7 +71,7 @@ jobs:
                 command: |
                     export NVM_DIR="/opt/circleci/.nvm"
                     [ -s "$NVM_DIR/nvm.sh" ] && \. "$NVM_DIR/nvm.sh"
-                    nvm install v14.17.5 --default
+                    nvm install 14 --default
                     node -v
                     cp .env-dist .env
                     npm install
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -72,7 +72,6 @@ jobs:
                     export NVM_DIR="/opt/circleci/.nvm"
                     [ -s "$NVM_DIR/nvm.sh" ] && \. "$NVM_DIR/nvm.sh"
                     nvm install v14 --default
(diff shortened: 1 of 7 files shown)


toufali:
The issue that complicated/blocked this upgrade was an older image incompatible with Node 14.  With help from that image's code owner, @jrbenny35 , we were able to update `b4handjr/selenium-firefox:python3-latest` for Node 14 and pass the integration test.

maxxcrawford (approved):
This works as expected for me! I was able to do the following locally: 

- Check for breaches
- Log in
- Resolve a breach
- Add secondary email (and verify via email) 

@groovecoder When you are checking this, could you trigger a breach email (or maybe the onboarding email flow)? You may also need to check the `about:protections` page too.  

@moz-astults Can you also review this from a prod-setup standpoint? 

@pdehaan Is there any way to simulate HIBP pushing new breach data to make sure everything works as expected?

groovecoder (approved):
Code LGTM. I spot-checked some DB-changing things too, as I saw this updates the `pg` dependency too. Adding and removing email addresses worked, and sending a breach alert worked too. Ship it!

.circleci/config.yml
```diff
@@ -71,7 +71,7 @@ jobs:
                 command: |
                     export NVM_DIR="/opt/circleci/.nvm"
                     [ -s "$NVM_DIR/nvm.sh" ] && \. "$NVM_DIR/nvm.sh"
-                    nvm use 10
+                    nvm install v14.17.5 --default
```
pdehaan, line 74: Q: Should this be pinned to a specific version? Or wouldn't pinning generically to v14 use the latest (security) release in that channel? I guess my concern is that we'd have to constantly bump this if new versions of v14 are added for future security releases, etc.

toufali: I think you're right – updated.

PR #2249, opened 2021-08-08
Update data-classes.ftl

pdehaan:
Add "Appointments" from recent [MMG Fusion](https://monitor.firefox.com/breach-details/MMGFusion) breach.
https://haveibeenpwned.com/api/v3/breach/MMGFusion

> In December 2020, the dental practice management service MMG Fusion was the victim of a data breach which exposed 2.6M unique email addresses. The data also included patient appointments, names, phone numbers, dates of birth, genders and physical addresses. A small number of records also included passwords stored as bcrypt hashes.

### DIFF
--- a/locales/en/data-classes.ftl
+++ b/locales/en/data-classes.ftl
@@ -4,6 +4,7 @@ account-balances = Account balances
 address-book-contacts = Address book contacts
 age-groups = Age groups
 ages = Ages
+# Refers to medical appointments.
 appointments = Appointments
 apps-installed-on-devices = Apps installed on devices
 astrological-signs = Astrological signs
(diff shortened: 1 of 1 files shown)


locales/en/data-classes.ftl
```diff
@@ -4,6 +4,7 @@ account-balances = Account balances
 address-book-contacts = Address book contacts
 age-groups = Age groups
 ages = Ages
+appointments = Appointments
```
flodolo, line 7: Please add a quick note that this refers to doctor appointments. I have a feeling a few language might confuse this with something else.

flodolo (approved):

PR #2255, opened 2021-08-18
fix incorrect font reference to "Inter"

toufali:
Correctly link body font family to "Inter".  

I'm not up on original designs, but reviewers should please keep an eye out for layout shifts or strange text blocks as a result of this update.


maxxcrawford (approved):

maxxcrawford:
Great catch! I wasn't seeing major design differences (but I also had Inter on my machine).

PR #2260, opened 2021-08-23
check .nav-user-email exists before referencing

groovecoder:
When the user is signed-out, the new code in `dashboard.js` throws this error:
```
Uncaught TypeError: can't access property "textContent", document.querySelector(...) is null
```
NBD in local & heroku, because it's the last code in `dashboard.js`, so the error doesn't cause any other effects. But in stage and prod, `dashboard.js` script is bundled into the single `app.min.js` file and the error stops all the following JS from executing - yikes! Breaks a lot of functionality, including the "Sign in" button.

This changes the code so it checks the element exists before trying to access its `textContent` property.

### DIFF
--- a/public/js/dashboard.js
+++ b/public/js/dashboard.js
@@ -96,11 +96,14 @@ if (removeMonitorButton) {
 }
 
 const relayLink = document.querySelector("[data-event-label='Try Firefox Relay']");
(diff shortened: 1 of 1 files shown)


maxxcrawford (approved):
🎯

PR #2267, opened 2021-08-24
allow self-signed certs on dev and heroku

groovecoder:


heaversm:
I think you just need to add `pg-connection-string` as a dependency to the `package.json`

db/knexfile.js
```diff
@@ -1,19 +1,26 @@
 "use strict";
 
+
+const { parse } = require("pg-connection-string");
+
 const AppConstants = require("../app-constants");
+const connectionObj = parse(AppConstants.DATABASE_URL);
+if (["dev", "heroku"].includes(AppConstants.NODE_ENV)) {
```
heaversm, line 8: I think it should just be `if (["heroku"]` - when I keep dev in there, I get an error locally that says SSL is not supported.

Issue #2235, opened 2021-07-23
Add a banner to Firefox Monitor that educates users about exposed IP addresses


tcinotto:
PRD: https://docs.google.com/document/d/1uuOPVRLABRvaG7ofxFTDaRT0BwUFNhVQK4bVFOpXPy8/edit?usp=sharing

tcinotto:
@lesleyjanenorton just wanted to flag you so you're aware of this idea! Feel free to add into the PRD

toufali:
PR #2284 opened. Still outstanding:
- Guardian integration for VPN Protected status
  - test user IP against protected servers
  - fallback when server array is blank or Guardian connection failed
- K8 integration for MaxMind location database
  - City/State/Country review needed for locales outside of US
- "Protected" version of banner
- Analytics: return Monitor page name variable
- Localization: support locales that use VPN

EMMLynch:
Closing since we've redesigned the site and functionality since this was created. If you feel that this is still needed, please let me know.

PR #2284, opened 2021-09-13
IP location banner for VPN education/promotion

toufali:
Adds banner to Firefox Monitor that displays the user's geographic location derived from their publicly accessible IP address.  The banner also informs users whether their IP address is being masked by the Mozilla VPN. The banner expands downward to provide contextual information about why users might need the Mozilla VPN for extra protection.

Ticket #2235


locales/en/app.ftl
```diff
@@ -798,3 +798,21 @@ vpn-promo-cta = Get { -brand-mozilla-vpn }
 
 vpn-promo-headline-new = Save 50% with a full year subscription
 vpn-promo-copy-new = Protect your online data—and choose a VPN subscription plan that works for you.
+
+# VPN educational/promotional banner
+vpn-banner-location = Your location:
+vpn-banner-protect-yourself = Protect yourself
+vpn-banner-with-moz-vpn = with { -brand-mozilla-vpn }.
```
flodolo, line 805: Please don't split sentences, this should be a single string.

See also [this doc](https://mozilla-l10n.github.io/documentation/localization/dev_best_practices.html#make-strings-properly-localizable) (this is basically concatenation).

views/partials/vpn-banner.hbs
```diff
@@ -0,0 +1,42 @@
+<aside class="vpn-banner">
+	{{#with (vpnBannerData)}}
+	<section class="vpn-banner-top">
+		<label>{{ getString "vpn-banner-location" }} <output>{{ shortLocation }}</output>&nbsp; • </label>
```
flodolo, line 4: This is a lot of hard-coded characters: can we have the localization exposed as part of the string?

views/partials/vpn-banner.hbs
```diff
@@ -0,0 +1,42 @@
...
+	</section>
+	<section class="vpn-banner-bottom">
+		<h2 class="vpn-banner-title">{{ getString "vpn-banner-title-1" }}
+			<span>{{ getString "vpn-banner-subtitle-1" }}</span>
+		</h2>
+		<div class="vpn-banner-lists">
+			<ul class="status-list">
+				<li>{{ getString "vpn-banner-status" }} <output class="vpn-banner-red">{{ getString "vpn-banner-not-protected"
+						}}</output></li>
+				<li>{{ getString "vpn-banner-ip-address" }} <output>{{ ip }}</output></li>
```
flodolo, line 17: Same here: can we expose the IP as a variable in the string?

locales/en/app.ftl
```diff
@@ -798,3 +798,21 @@ vpn-promo-cta = Get { -brand-mozilla-vpn }
...
+vpn-banner-status = Current status:
+vpn-banner-protected = Protected ✓
+vpn-banner-not-protected = Not protected ⚠
+vpn-banner-ip-address = IP address:
+vpn-banner-step-1 = Subscribe to { -brand-mozilla-vpn }
+vpn-banner-step-2 = Select a VPN location
+vpn-banner-step-3 = Activate VPN and browse securely
+vpn-banner-cta = Get { -brand-mozilla-vpn }
+vpn-banner-cta-expand = Expand
+vpn-banner-cta-close = Close
```
flodolo, line 818: Please keep an empty line at the end of these files.

flodolo (changes requested):

views/partials/vpn-banner.hbs
toufali: Do you mean `{{ shortLocation }}`?  It is derived from a 3rd party (Maxxmind).  Else, if you mean `&nbsp; • `, yes – that could likely be added to the .ftl if you prefer.

flodolo:
```
vpn-banner-location = Your location: <output>{ $location}</output>
```

BTW, why do you need the `&nbsp; •`? Are they all on the same line? I looked at Figma, and can't see anything similar. 

Do you have a screenshot of the result of this patch? The layout looks all but localization friendly.

locales/en/app.ftl
toufali: Thanks for the helpful link!  Assuming no objections I would like to add the documentation to this repo's README.

I'm trying to follow an example suggested in your doc by using this string that includes HTML: 
```
vpn-banner-protect-yourself = <output class="vpn-banner-red">Protect yourself</output> with { -brand-mozilla-vpn }.
```

Unfortunately this does not work as the tags are printed literally.  There are no additional steps listed in your doc – is there another resource or example that I can look at?  Or does is that functionality not supported for our setup?

flodolo: Unfortunately I'm not a dev, so I can only help up to a certain point. I thought this project was using Fluent React, so you'd need [React Overlays](https://github.com/projectfluent/fluent.js/wiki/React-Overlays), but I can't find references in packages.json

I see strings with HTML markup in the file, so it's definitely possible. Maybe previous devs on the project have pointers?

toufali: The way HTML is currently processed in this repo requires much additional code customized for each individual case.  I found a much simpler way using a different syntax in the templating engine.  This syntax `{{{ string }}}` requires no javascript to process the markup in the string.

Armed with this knowledge, I'll redo all strings as needed with your comments addressed.  Thanks for the help!

views/partials/vpn-banner.hbs
toufali: The `&nbsp; • ` is all on the same line – see top bar of Figma file after location.  I refactored this into the CSS instead; the resulting string and template should be cleaner now.

toufali:
@flodolo I refactored the code based on your comments.  

I spent some time trying to integrate the 3rd-party variables into the localization file but was unsuccessful.  Happy to keep working on this if you like.

cc @maxxcrawford in case he's passed 3rd-party vars through localization files successfully

controllers/ip-location.js
```diff
@@ -0,0 +1,41 @@
+/*eslint-disable no-process-env*/
+// using process.env here to keep the banner/VPN vars contained – for when the vpn-banner is eventually removed
+
+// IP location data includes GeoLite2 data created by MaxMind, available from https://www.maxmind.com.
+// Test IPs by comparing to the corresponding json, for example: https://github.com/maxmind/MaxMind-DB/blob/main/source-data/GeoLite2-City-Test.json
+
+"use strict";
+
+require("dotenv").config();
+const maxmindDb = process.env.GEOIP_GEOLITE2_PATH + process.env.GEOIP_GEOLITE2_CITY_FILENAME;
```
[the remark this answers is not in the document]
pdehaan: Or if we do the app-constants.js `optionalEnvVars` suggestion above, we might be able to use `AppConstants.GEOIP_GEOLITE2_PATH` here.
Or maybe tweak the maxmind env vars to pre-join `path.join(AppConstants.GEOIP_GEOLITE2_PATH, AppConstants.GEOIP_GEOLITE2_CITY_FILENAME)` in the app-constants.js file.

controllers/ip-location.js
```diff
@@ -0,0 +1,41 @@
...
+const Reader = require("@maxmind/geoip2-node").Reader;
+
+async function ipLocation(req, res) {
+  let reader, geoData, countryName, cityName, locationArr;
+  const clientIp = process.env.NODE_ENV === "dev" ? "216.160.83.56" : req.ip; // for dev, return an IP that exists in GeoLite2 test DB, e.g. 216.160.83.56
```
pdehaan, line 15: Should this use `AppConstants.NODE_ENV`?

controllers/ip-location.js
```diff
@@ -0,0 +1,41 @@
...
+
+  if (clientIp === req.session.locationData?.clientIp) return res.status(200).json(req.session.locationData);
+
+  try {
+    reader = await Reader.open(maxmindDb);
```
pdehaan, line 20: I'm pretty ignorant here (and in general), but what how does this line work?
Is it opening the maxmind DB on _each_ request? Or is this somehow cached?

controllers/ip-location.js
```diff
@@ -0,0 +1,41 @@
...
+    locationArr = [cityName, geoData.subdivisions?.[0].isoCode, countryName].filter(str => str); // [city name, state code, country code] with non-null items.
+  } catch (e) {
+    console.warn("Error reading location database:", e);
+    locationArr = [null]; // a null item will return empty strings for locationData below
+  }
+
+  req.session.locationData = {
+    clientIp,
+    shortLocation: locationArr.slice(0, 2).join(", "), // shows the first two location values from the ones available
+    fullLocation: locationArr.join(", "), // shows up to three location values from the ones available
```
pdehaan, line 33: If we error above and have `locationArr = [null]`, how do these work?

Ah, I guess they just return empty strings; which explains the comment above on L27 (although I see similar behavior w/ `[null]` vs `[]`). :+1:
```js
const locationArr = [null];
console.log({
  short: locationArr.slice(0, 2).join(", "),
  full: locationArr.join(", ")
});

// {short: "", full: ""}
```

public/js/monitor.js
```diff
@@ -413,47 +408,111 @@ function addWaitlistObservers() {
...
+  }
+
+  function getPageAttribution() {
+    let page = location.pathname;
+
+    if (page.startsWith("/")) page = page.slice(1);
+
+    if (page === "") page = "home";
+
+    return `&utm_content=${page}`;
```
pdehaan, line 501: Does this account for query strings that don't have a "?" or could we end up w/ URLs like `https://monitor.firefox.com&utm_content=home`?

public/js/monitor.js
```diff
@@ -523,17 +582,8 @@ function vpnBannerLogic() {
...
-  if (!preferredFirstLanguageIsTier1) {
-    return;
-  }
-
-  // Only show banner if users first language is English, Germand or French variant
-  if (["en", "de", "fr"].some(lang=>preferredLanguages[0].includes(lang))) {
-    vpnBannerLogic();
-  }
+  // Show vpn-banner for locales not in the blocked list
+  if (!["zh-CN"].some(lang => navigator.languages[0].includes(lang))) initVpnBanner();
```
pdehaan, line 586: Should the blocked list be in the AppConstants file or something?

public/js/monitor.js
```diff
@@ -413,47 +408,111 @@ function addWaitlistObservers() {
...
+    const json = await cache.match(req)
+      .then(res => res.json())
+      .catch(e => console.warn("Could not get cached response.", e.message));
+
+    return json;
+  }
+
+  async function fetchData(req) {
+    const abortController = new AbortController();
+    const timer = setTimeout(() => abortController.abort(), 4000); // abort a delayed response after 4s
```
pdehaan, line 482: Should this 4s timeout be configurable?

```js
async function fetchData(req, reqTimeoutMs=4000) {
  ...
}
```

locales/en/app.ftl
```diff
@@ -798,3 +798,23 @@ vpn-promo-cta = Get { -brand-mozilla-vpn }
 
 vpn-promo-headline-new = Save 50% with a full year subscription
 vpn-promo-copy-new = Protect your online data—and choose a VPN subscription plan that works for you.
+
+## VPN educational/promotional banner.  HTML tags should not be translated, e.g. `<em>`
```
flodolo, line 802: Leave an empty line after group comments.

locales/en/app.ftl
```diff
@@ -798,3 +798,23 @@ vpn-promo-cta = Get { -brand-mozilla-vpn }
...
+# user's location is populated dynamically by 3rd-party service, eg: "Your location: Los Angeles, CA".  The 3rd-party service will need to provide localization.
+vpn-banner-location = Your location:
```
flodolo, line 804: It is possible to expose the actual localization as a placeholder in the string? This basically comes down to [avoid hard-coding characters](https://mozilla-l10n.github.io/documentation/localization/dev_best_practices.html#dont-hardcode-characters).

Same applies to `vpn-banner-ip-address`.

locales/en/app.ftl
```diff
@@ -798,3 +798,23 @@ vpn-promo-cta = Get { -brand-mozilla-vpn }
...
+vpn-banner-subtitle-2 = Protect your location and browse securely in 3 steps
+vpn-banner-status-protected = Current status: <em>Protected ✓</em>
+vpn-banner-status-not-protected = Current status: <em>Not protected ⚠</em>
+# user's IP address is determined dynamically, eg: "IP address: 192.168.1.1"
+vpn-banner-ip-address = IP address:
+vpn-banner-step-1 = Subscribe to { -brand-mozilla-vpn }
+vpn-banner-step-2 = Select a VPN location
+vpn-banner-step-3 = Activate VPN and browse securely
+vpn-banner-cta = Get { -brand-mozilla-vpn }
+# button to expand or close panel
```
flodolo, line 818: This is current a single string comment, applying to the next string, it should be a group comment.

app-constants.js
```diff
@@ -43,22 +39,33 @@ const kEnvironmentVariables = [
...
-  "RECRUITMENT_BANNER_LINK",
-  "RECRUITMENT_BANNER_TEXT",
   "REDIS_URL",
   "SENTRY_DSN",
   "DELETE_UNVERIFIED_SUBSCRIBERS_TIMER",
   "EXPERIMENT_ACTIVE",
   "MAX_NUM_ADDRESSES",
 ];
 
+const optionalEnvVars = [
```
groovecoder, line 49: OMG thank you for this!!

controllers/ip-location.js
groovecoder: Yes please. Until we do a better fuller refactor, let's stick to use `AppConstants` consistently.

controllers/ip-location.js
groovecoder: Good idea. We could move this into something like the `LocaleUtils.loadLanguagesIntoApp`, `FormUtils.loadCountriesIntoApp`, `HIBP.loadBreachesIntoApp` so it loads the data into the app a single time during server startup. (they're only 20K and 18K)

public/js/monitor.js
```diff
@@ -413,47 +408,111 @@ function addWaitlistObservers() {
...
+
+  vpnBanner.cta = vpnBanner.querySelector("a.vpn-banner-cta");
+  vpnBanner.cta.setAttribute("href", vpnBanner.cta.getAttribute("href") + getPageAttribution());
+  vpnBanner.setAttribute("data-protected", Boolean(protectionData?.isProtected));
+  vpnBanner.addEventListener("click", handleClick);
+
+  if (cache && protectionData) cache.put(protectionDataReq, new Response(JSON.stringify(protectionData)));
+
+  async function initCache() {
+    const cacheAvailable = "caches" in self;
```
groovecoder, line 462: TIL [Cache API](https://developer.mozilla.org/en-US/docs/Web/API/Cache)

views/partials/top-level/security-tips.hbs
```diff
@@ -18,60 +18,61 @@
...
+              {{/if}}
+              {{> top-level/security-tips/tip-content }}
+              
+              {{> top-level/security-tips/dos-and-donts }}
+              {{/each}}
+              </div>
+              {{/each}}
+              </div>
+              </div>
+              </section>
```
groovecoder, line 77: This seems to have flattened out a bunch of the indentation? Is that on purpose?

views/partials/vpn-banner.hbs
```diff
@@ -0,0 +1,52 @@
...
+		<label class="vpn-banner-chevron" data-expand={{ getString "vpn-banner-cta-expand" }} data-close={{
+			getString "vpn-banner-cta-close" }}></label>
+	</section>
+	<section class="vpn-banner-bottom">
+		<h2 class="vpn-banner-title">
+			<span class="protected-txt">{{ getString "vpn-banner-title-1" }}</span>
+			<span class="protected-txt">
+				{{ getString "vpn-banner-title-2" }}
+				<span class="subheading">{{ getString "vpn-banner-subtitle-2" }}</span>
+				</span>
```
groovecoder, line 17: Another bit of weird indentation that doesn't seem to match where it should be?

controllers/ip-location.js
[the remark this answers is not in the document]
toufali: Thanks for the reminder – path added.

app-constants.js
```diff
@@ -43,22 +39,33 @@ const kEnvironmentVariables = [
...
+  "RECRUITMENT_BANNER_LINK",
+  "RECRUITMENT_BANNER_TEXT",
+];
```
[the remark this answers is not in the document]
toufali: Yes.  Funny I added the "optional" vars array for a separate case but didn't make use of it here.  GEOIP* vars added to "optional".

controllers/ip-location.js
toufali: done.

controllers/ip-location.js
toufali: @groovecoder I assume you're referring to the test MaxMind databases – they are actually stripped down for testing.  I imagine the actual databases are much larger...

@pdehaan I took this block to mean we're caching session data to redis: 
https://github.com/mozilla/blurts-server/blob/1a940301908ecd1fd0d2fb1a28dc8762d4609dfa/server.js#L195-L207

So, if a user's ip address matches an ip stored in session from an earlier successful lookup, we return that data instead of opening/reading the MaxMind database.

Admittedly this is my first work with Redis, so let me know if I've missed something!

controllers/ip-location.js
toufali: Yes, this was a remnant from my earlier build when the client was brittle and depended on these keys – empty or not.  With the env var addition discussed above, I took the opportunity to simplify and return a response with those keys removed altogether.

public/js/monitor.js
toufali: The link is hard-coded: 
`https://vpn.mozilla.org/?utm_source=firefox-monitor&utm_medium=banner&utm_campaign=ip`

The `utm_content` key is the only dynamic value.  If you agree, it might be more intuitive to add the entire query string in js, instead of splitting the static/dynamic parts between html/js.

public/js/monitor.js
toufali: Good suggestion – I added this, and also updated the recent email promo addition to use the same var/logic.

public/js/monitor.js
toufali: done.

locales/en/app.ftl
toufali: updated.

locales/en/app.ftl
toufali: done for both `vpn-banner-location` and `vpn-banner-ip-address`

locales/en/app.ftl
toufali: updated.

views/partials/top-level/security-tips.hbs
toufali: fixed.

views/partials/vpn-banner.hbs
toufali: fixed.

controllers/ip-location.js
groovecoder: That's good to optimize each session individually, but we should still load the maxmind DB into the application memory during server startup. Or else 1000 visitors will cause 1000 file open-and-read-into-memory operations when we may only need a single open-and-read-into-memory operation for all visitors.

groovecoder (changes requested):

controllers/ip-location.js
toufali: This has been updated – thanks for flagging.  I split the code out into a separate service from the controller.  See changes here: https://github.com/mozilla/blurts-server/pull/2284/commits/52c15b26fb4d04092e51e63459158ea8685f8c3c

ip-location-service.js
```diff
@@ -0,0 +1,55 @@
+"use strict";
+
+const path = require("path");
+const reader = require("@maxmind/geoip2-node").Reader;
+const AppConstants = require("./app-constants");
+
+let locationDb, timestamp;
```
groovecoder, line 7: `timestamp` seems like a risky variable name for something that's being declared in global scope. Change to `locationDBLastOpenedTimestamp` or something?

groovecoder (changes requested):
Code looks good except 1 small variable name change. Didn't spot-check yet though.

flodolo (approved):

ip-location-service.js
toufali: @pdehaan per my chat with @groovecoder:
`timestamp` is indeed scoped to that module:
`let locationDb, timestamp;` is short for
```
let locationDb;
let timestamp;
```
and both vars exist within a Node module which is scoped to that file alone.

.env-dist
```diff
@@ -77,3 +77,14 @@ MAX_NUM_ADDRESSES=5
 
 RECRUITMENT_BANNER_LINK=
 RECRUITMENT_BANNER_TEXT=
+
+# Locales blocked from viewing Mozilla VPN promos. Use CSV whithout whitespace.
```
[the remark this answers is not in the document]
toufali: I loathe typos committed by me – thanks for this ❤️

ip-location-service.js
```diff
@@ -0,0 +1,55 @@
...
+
+async function openLocationDb() {
+  if (locationDb && isFresh()) {
+    console.warn("Location database opened less than 24hrs: returning original instance.");
```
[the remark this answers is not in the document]
toufali: Per chat with @groovecoder , it's understood that opening the db is expensive, but if we open it just once it runs the risk of stale data as long as the server is up, right?  So this was the easy compromise.  It would be good to know how often the MaxMind db actually changes, to get a better interval – I bet we could go 48hrs without reload.  In fact, I think I read from the MaxMind site that the db is updated once per week...  

Happy to hear alternate suggestions!

ip-location-service.js
pdehaan: Yeah, I was thinking of https://eslint.org/docs/rules/one-var. But looking through the eslint source, I don't think that `one-var` is in the default `eslint:recommended` config. Carry on.

ip-location-service.js
```diff
@@ -0,0 +1,55 @@
...
+    return locationDb;
+  }
+
+  try {
+    const dbPath = path.join(AppConstants.GEOIP_GEOLITE2_PATH, AppConstants.GEOIP_GEOLITE2_CITY_FILENAME);
+    locationDb = await reader.open(dbPath);
+  } catch (e) {
+    return console.warn("Could not open location database:", e.message);
```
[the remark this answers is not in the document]
toufali: I'm enjoying insight into how others might read my code 😊 .  

Yes, returning the object in `openLocationDb()` is a "benign" remnant of an earlier attempt.  I see how it can be confusing – all returns could just have easily been `undefined`.  I can/might just remove the returned values.

My intention was to use locationDb as a private singleton.  I don't believe it's global.

ip-location-service.js
```diff
@@ -0,0 +1,55 @@
...
+  return locationDb;
+}
+
+async function readLocationData(ip, locales) {
+  let locationArr;
+
+  if (!isFresh()) await openLocationDb();
+
+  try {
+    const data = locationDb.city(ip);
```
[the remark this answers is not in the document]
toufali: In the case of `const data = locationDb.city(ip);` with undefined locationDb, the `catch` block kicks in.  I just confirmed 💪

ip-location-service.js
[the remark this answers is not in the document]
pdehaan: Yeah, maybe I'm missing the core concept here&hellip;

Say we have two MaxMind databases:

```sh
GEOIP_GEOLITE2_CITY_FILENAME=GeoLite2-City-Test.mmdb
GEOIP_GEOLITE2_COUNTRY_FILENAME=GeoLite2-Country-Test.mmdb
```

And we want to keep those fresh. How are those being... loaded/deployed/updated?

I see below (L17) we're using `await reader.open(dbPath)`. But do those files get updated dynamically while the server is deployed somehow? Or is the data always the same during a deployment and only gets updated manually or via bumping some npm dependency or redeployment?
I guess my ignorant concern was that we have the two databases, and we do a production deployment and the .mmdb data never changes in the 1-4 weeks between deployments, but we still re-read the same files every 24h, loading the same data.
But if we aren't refreshing the data _dynamically_ somehow, I figured we just read the data once and don't bother with a freshness check.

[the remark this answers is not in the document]
toufali: Yeah, this might be a good question for me to pose to SRE team since the database is managed within their k8 cluster.  Point taken that freshness might be overthinking this.

I confirmed that Geolite2 is updated once a week, but the paid GeoIP2 is updated twice weekly.  I will also check with SRE which they use.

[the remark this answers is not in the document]
pdehaan: OK, I'll retract my initial concern. From the sounds of the Slack conversation, OPs is updating the .mmdb files 2x/week and updating the container's file system. So if we poll for freshness from our side once every 24 hours we know the data can never be older than 48-72h.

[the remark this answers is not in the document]
toufali: @pdehaan I updated to reload the db upon client request every 3 days.  Let me know if you'd prefer reverting back to 24hrs 💪

groovecoder (approved):
Code LGTM and spot-checks work. Eager to see it working on dev/stage/prod!

PR #2308, opened 2021-09-29
Mh/kanary 029 sort order

heaversm:
sort removal results by status, then by blocked or unblocked, then alphabetically within each group
hides result risk if the status is complete

addresses #MPP-944

### DIFF
--- a/controllers/user.js
+++ b/controllers/user.js
@@ -19,7 +19,7 @@ const {
   getUTMContents,
   hasUserSignedUpForWaitlist,
 } = require("./utils");
-const { JS_CONSTANTS } = require("../js-constants");
+const { JS_CONSTANTS, REMOVAL_STATUS } = require("../js-constants");
 
 const FXA_MONITOR_SCOPE = "https://identity.mozilla.com/apps/monitor";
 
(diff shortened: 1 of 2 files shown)


controllers/user.js
```diff
@@ -689,6 +689,49 @@ async function getRemoveMoreTimePage(req, res) {
...
 
+function alphabetizeByBroker(data) {
+  return data.sort(function (a, b) {
+    if (a.broker < b.broker) {
+      return -1;
+    }
+    if (a.broker > b.broker) {
+      return 1;
+    }
+    return 0;
```
pdehaan, line 700: FWIW, I think you can do this w/ `return a.broker.localeCompare(b.broker)` (or close to it).

controllers/user.js
```diff
@@ -689,6 +689,49 @@ async function getRemoveMoreTimePage(req, res) {
...
+      removalItem.status !== REMOVAL_STATUS["COMPLETE"].id &&
+      removalItem.current_step !== JS_CONSTANTS.REMOVAL_STEP["BLOCKED"].code
+    );
+  });
+
+  alphabetizeByBroker(activeItems);
+
+  const sortedData = [...completeItems, ...activeItems, ...blockedItems];
+
+  console.log("sorted", sortedData);
```
pdehaan, line 730: nit: Not sure if we want to add an issue to remove the debugging `console.log` statements before going to production.

controllers/user.js
```diff
@@ -689,6 +689,49 @@ async function getRemoveMoreTimePage(req, res) {
...
+      removalItem.current_step === JS_CONSTANTS.REMOVAL_STEP["BLOCKED"].code
+    );
+  });
+
+  alphabetizeByBroker(blockedItems);
+
+  const activeItems = removalData.filter((removalItem) => {
+    return (
+      removalItem.status !== REMOVAL_STATUS["COMPLETE"].id &&
+      removalItem.current_step !== JS_CONSTANTS.REMOVAL_STEP["BLOCKED"].code
```
pdehaan, line 722: Given that we do this in a couple places, I wonder if it's ripe for some helper functions:

```js
function isComplete(status) {
  return status === REMOVAL_STATUS.COMPLETE.id;
}

function isBlocked(step) {
  return step === JS_CONSTANTS.REMOVAL_STEP.BLOCKED.code;
}

...

return (
  !isComplete(removalItem.status) &&
  !isBlocked(removalItem.current_step)
);
```

PR #2310, opened 2021-09-30
Migrate travis to circleci

toufali:


pdehaan:
Ah, I think they've been asking us to move off Travis for 10 months: https://github.com/mozilla/blurts-server/issues/2000

.circleci/config.yml
```diff
@@ -1,5 +1,32 @@
 version: 2.0
 jobs:
+    lint-js-css:
```
groovecoder, line 3: Can we break this up into separate `lint-js`, `lint-css`, and `lint-audit` jobs like Travis had? That's make it a bit easier to see exactly which job causes a PR to fail.

groovecoder (changes requested):
The `test` job in Circle seems to only run `npm run lint` and doesn't run `npm test`, which Travis does(?) by default when a job is named "test".

Since we're running all the `lint` commands as their own job, can we change the Circle `test` job to run `npm test` instead of `npm run lint` ?

.circleci/config.yml
toufali: Updated to break them up.  Took the opportunity to also refactor for latest CircleCI version, with Orbs for less boilerplate and faster builds.

toufali:
While splitting the lints up, I looked into the lint-audit job – as suspected it's likely not doing anything: The `npm-audit-ci-wrapper` package is deprecated, and there have been numerous audit alerts that have gone unflagged.  I'll open a new ticket to update this lint step.

PR #2311, opened 2021-09-30
Sort the REMOVAL_SITES

pdehaan:


js-constants.js
```diff
@@ -70,7 +71,6 @@ const JS_CONSTANTS = {
     "xlek.com",
     "yellowbook.com",
     "zabasearch.com",
-    "peoplebyname.com",
   ],
```
pdehaan, line 74: Alternatively, this is just an array of simple strings, so we can throw a `.sort()` at the end of this and not have to worry about the original sort order.

PR #2312, opened 2021-09-30
update package-lock and request npm 7 in "engines"

toufali:
A "benign" update to specify using NPM 7.  

NPM Version 7 is the current as of today – 6 is considered legacy.  Version 7 uses lockfile version 2, while previous versions use lockfile version 1.  This is not a big deal, as version 2 is backwards-compatible.  However, NPM often complains when lockfile versions do not match, and adds a few seconds to the install in order to sync things up.

If not using version 7, npm should simply issue a warning and let you go about your day.  But, the reminder might help convince you to upgrade one of these days!

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -71,11 +71,14 @@ jobs:
                 command: |
                     export NVM_DIR="/opt/circleci/.nvm"
                     [ -s "$NVM_DIR/nvm.sh" ] && \. "$NVM_DIR/nvm.sh"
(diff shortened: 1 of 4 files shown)


toufali:
Closing this because the docker image ships with an NPM version (6.x) tied to our Node version (14.x).  It's probably beter practice to keep these locked.  So, rather than specify an upgrade to NPM 7, I'll add a request in package.json for NPM 6 and downgrade my version.

PR #2313, opened 2021-09-30
Clean up package.json

toufali:
In preparation of vulnerability fixes, the following housekeeping has been done on `package.json`:

- **Specify npm version 6**
  To avoid lockfile warnings and adding more build-time, this update requests npm version 6 in package.json.
  
  I previously attempted to force npm version 7, which is current.  However, the docker image bundles version 6 to Node 14, which we recently upgraded to.  It's probably safer to use the pair as specified in the image.

- **Remove unused packages**
  Using depcheck in conjunction with full-repo searches, I identified a good amount of packages to remove.  This included cli tools, which in my opinion should be installed locally per dev as needed, rather than forced on every build.


toufali:
closing in favor of #2316

PR #2316, opened 2021-10-01
Cleanup and security

toufali:


package.json
```diff
@@ -58,30 +55,28 @@
...
     "stylelint-config-standard": "18.3.0",
     "stylelint-scss": "^3.19.0",
-    "uglify-js": "3.11.5",
     "wdio-docker-service": "2.4.0",
-    "wdio-image-comparison-service": "1.13.6",
+    "wdio-image-comparison-service": "^3.0.0",
     "wdio-video-reporter": "2.0.1"
   },
+  "peerDependencies": {
+    "sass": "^1.42.1"
```
pdehaan, line 75: Does Sass need to be a `dependency` _and_ a `peerDependency`?

toufali: I think so... It's definitely a peer dep, so it earned a spot in `peerDependencies`. But, `peerDependencies` are not automatically installed with `npm install` as of this writing.  So by placing it in `dependencies` it gets installed automatically, including in our automated builds.  Removing it from `peerDependencies` triggers an eslint error, and removes a nice-to-have context on why that package is installed.  Open to suggestions!

pdehaan: Weird, that's a real head scratcher. I don't think I'm reproducing the ESLint error locally when removing it as a peer dependency.

```sh
gh pr checkout 2316
nvm i 14.17
rm -rf node_modules
npm i
npm run lint
```

That succeeds, but if I remove the peerDependency and then delete the ./node_modules/ (and package-lock.json) and reinstall everything, it still works (macOS 11).

I think I was just confused because I didn't see any mentions of needing peer dependencies in https://github.com/dlmanning/gulp-sass docs.
But hey, if it works, probably best to leave it alone.

toufali: Looks like it was an error triggered by an eslint plugin, `eslint-plugin-node`.  This is the error I have in my history, tho admittedly I can't reproduce it now when removing `peerDependencies`:
https://github.com/mysticatea/eslint-plugin-node/blob/master/docs/rules/no-unpublished-require.md

I thought it was a head-scratcher indeed, but after adding `Sass` to peerDeps, that error went away from me.  Maybe it was a coincidence.

Regarding mentions in the gulp-sass docs, you're right – it's not specifically listed as a **_peer_** dependency, but requiring a 3rd party compiler is specifically mentioned in the _installation_ section: https://github.com/dlmanning/gulp-sass#installation

I took this to suggest a peer dependency, since it's not required by our repo.  

Reading into it further, it seems the right solution is to add Sass to the `peerDependency` section and not include it in `dependencies`.  For us, I don't think this will work unless we update our automated builds and add documentation specifying that a manual install is required.  From [the docs](https://docs.npmjs.com/cli/v7/configuring-npm/package-json#peerdependencies), peer dependencies are not auto-installed for our version of npm (6).  They are for version 7, which we should support next time we bump our NodeJS version.

pdehaan: Awesome, thanks for the explanation. One of these days I'll sit down and actually understand what that error means. It's managed to elude me for years and I usually end up just disabling it.
Plus I'm always confused by Node's stance on "peerDeps are bad. No wait, now peerDeps are OK again", but nice to see them coming back in npm@7.

maxxcrawford:
@toufali I upgraded SASS for Relay and [ran into an issue](https://github.com/mozilla/fx-private-relay/pull/1079#pullrequestreview-748277307) where I had to pin a specific version. Did this happen for you?

maxxcrawford (commented):
Local tests are failing for me. Not sure if its related – but wanted to call it out.

toufali:
> Local tests are failing for me. Not sure if its related – but wanted to call it out.

@maxxcrawford I'm not seeing a difference between running tests on `main` vs this branch – can you provide the error message?  

Disclaimer, after 14/14 passing test suites and 91/91 passing tests, my test has always given up on coverage tests (coveralls.js).  Something to look at in another PR, but unchanged behavior for me.

toufali:
> @toufali I upgraded SASS for Relay and [ran into an issue](https://github.com/mozilla/fx-private-relay/pull/1079#pullrequestreview-748277307) where I had to pin a specific version. Did this happen for you?

@maxxcrawford I didn't notice any issues with adding Sass, other than what was discussed above with @pdehaan ...

groovecoder:
Whoa cool! TIL `depcheck`. Is there a way we can run `depcheck` as a CI job and have it fail if we're installing un-used deps?

PR #2338, opened 2021-10-14
[DO NOT MERGE] Review FTL file in Kanary

flodolo:


flodolo:
Thanks, I really appreciate you working on this. I'll leave this one open as is for reference, and probably open another one to double check what's left.

PR #2383, opened 2021-11-22
Sync Localization to Main

toufali:
Changelog:
- IP Location Banner
- Pontoon translations


flodolo:
I shouldn't be blocking on merges from localization to main, just making it explicitly for this one ;-)

toufali:
@flodolo – you might see a bit more noise in your review requests since `main` now requires a PR to merge...

PR #2390, opened 2021-12-02
merge `main` into `localization`

toufali:


flodolo:
The diff for this one looks surprising? I wouldn't expect changes to the en file for example

toufali:
> The diff for this one looks surprising? I wouldn't expect changes to the en file for example

My thoughts exactly – I'm investigating, but it's possible this might have been a side effect of a rebase.  The diffs are similar to what was merged last week: https://github.com/mozilla/blurts-server/commit/1548f384c972a0355c0ce7d4c393a5731d558b24

flodolo:
Yeah, I just realized that I had similar issues in another project, where rebases cause very confusing diffs.

flodolo:
This diff looks definitely less scary https://github.com/mozilla/blurts-server/compare/localization...main

toufali:
> This diff looks definitely less scary [localization...main](https://github.com/mozilla/blurts-server/compare/localization...main)

Yes -- I believe that's the same diff as the PR we're on (main into localization)?  My plan was to merge this less-scary one first, then hopefully see a much less scary diff going from localization to main 🤞

PR #2391, opened 2021-12-07
fix undefined `supportedLocales` for prod

toufali:


template-helpers/hbs-helpers.js
```diff
@@ -21,8 +21,8 @@ function getSupportedLocales(args) {
 }
 
 function vpnPromoBlocked(args) {
-  const userLocale = args.data.root.req.supportedLocales[0];
-  return AppConstants.VPN_PROMO_BLOCKED_LOCALES?.some(blockedLocale => userLocale.includes(blockedLocale));
+  const userLocales = getSupportedLocales(args);
+  return AppConstants.VPN_PROMO_BLOCKED_LOCALES?.some(blockedLocale => userLocales[0].includes(blockedLocale));
```
pdehaan, line 25: Q: Is `userLocales` always guaranteed to be an array with at least 1 element?

toufali: From what I can tell, it's using negotiateLanguages from fluent-langneg.  From [the docs](https://projectfluent.org/fluent.js/langneg/modules.html#negotiatelanguages) should always return an array.  I had checked the original code with a single locale (en) to be sure – though admittedly not as thorough as the docs.

We can always add safety code if need be!  In this case, I was rushing a hotfix that did not present on stage.

PR #2399, opened 2021-12-15
Compare `merge/monitor-kanary-main-001` with `main`

toufali:


controllers/user.js
```diff
@@ -20,6 +20,14 @@ const {
 
 const FXA_MONITOR_SCOPE = "https://identity.mozilla.com/apps/monitor";
 
+//DATA REMOVAL SPECIFIC
+const { FormUtils } = require("../form-utils");
+const fetch = require("node-fetch");
```
toufali, line 25: This repo is already using "got" for Node requests: https://www.npmjs.com/package/got

I'm more used to node-fetch as well, but assuming "got" fits your needs, maybe we should stick to a single Node request library?  I think the usage is similar: 
`response = await got(url, reqOptions);`

controllers/user.js
```diff
@@ -594,6 +685,1369 @@ function logout(req, res) {
...
+
+  let lastAddedEmail = null;
+
+  req.session.user = await DB.setBreachesLastShownNow(user);
+  if (req.session.lastAddedEmail) {
+    lastAddedEmail = req.session.lastAddedEmail;
+    req.session["lastAddedEmail"] = null;
+  }
+
+  let removeData,
```
toufali, line 942: I'm seeing the "remove" namespace is pretty widespread.  Feel free to ignore my comment above if changing to "removal" is too tricky at this point.

controllers/user.js
```diff
@@ -594,6 +685,1369 @@ function logout(req, res) {
...
+    }
+  });
+
+  return emailDomainMatch;
+}
+
+async function getHashedWaitlist() {
+  const hashedWaitlistArray = await readFile("hashed-waitlist.txt", "binary");
+  return hashedWaitlistArray.toString().split("\n");
+}
```
toufali, line 1631: it looks like hashed-waitlist.txt is opened, read, and split on every user visit/login?  Perhaps it can be read and split once, upon server startup, and hashedWaitlistArray can be saved as a constant?

controllers/user.js
```diff
@@ -594,6 +685,1369 @@ function logout(req, res) {
...
+
+  const user = req.user;
+
+  if (!user.primary_email.includes("@mozilla.com")) {
+    console.error("non mozilla email");
+    return res.status(404).json({
+      error:
+        "You must be signed in with a mozilla.com email address to access this page",
+    });
+  }
```
toufali, line 2014: should fox-food-specific code be removed?

js-constants.js
```diff
@@ -0,0 +1,155 @@
...
+    "truthfinder.com",
+    "unmask.com",
+    "usphonebook.com",
+    "ussearch.com",
+    "veripages.com",
+    "whitepages.com",
+    "xlek.com",
+    "yellowbook.com",
+    "zabasearch.com",
+  ],
```
toufali, line 98: Is this info not provided by Kanary?  Not sure we want to be the stewards for the list of data brokers.

toufali (commented):
I got up to locale-utils.js in review round 1.  Will pickup from there for review round 2 😅

app-constants.js
```diff
@@ -3,9 +3,15 @@
 /* eslint-disable no-process-env */
 
 const path = require("path");
-require("dotenv").config({path: path.join(__dirname, ".env")});
+require("dotenv").config({ path: path.join(__dirname, ".env") });
+
+if (!process.env.SERVER_URL && process.env.NODE_ENV === "heroku") {
+  process.env.SERVER_URL = `https://${process.env.HEROKU_APP_NAME}.herokuapp.com`;
+}
```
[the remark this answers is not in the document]
heaversm: 👍🏽

app-constants.js
```diff
@@ -3,9 +3,15 @@
...
 
 const requiredEnvVars = [
+  "KANARY_ENDPOINT",
+  "KANARY_TOKEN",
```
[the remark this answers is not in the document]
heaversm: these are both necessary to call the API.

controllers/home.js
```diff
@@ -157,6 +163,16 @@ function notFound(req, res) {
...
 
+//DATA REMOVAL SPECIFIC
+
+function getRemoveFAQPage(req, res) {
+  console.log("get remove faq page");
+  return res.render("top-level-page", {
+    title: req.fluentFormat("remove-faq-title"),
+    whichPartial: "top-level/remove-faq",
+  });
+}
```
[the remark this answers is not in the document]
heaversm: Ayyyy...I know...I was trying to make things unbranded...in retrospect `kanary` might've been a better prefix for everything, but I came to the same conclusion. `remove` is super confusing. I'll try to rename from `remove` to `removal` where it's not too much trouble.

controllers/oauth.js
```diff
@@ -6,20 +6,25 @@ const crypto = require("crypto");
...
 const { FluentError } = require("../locale-utils");
 const HIBP = require("../hibp");
 const mozlog = require("../log");
 const sha1 = require("../sha1-utils");
 
+//DATA REMOVAL SPECIFIC
+const { JS_CONSTANTS } = require("../js-constants");
+const { checkIfOnRemovalPilotList } = require("./user");
+//END DATA REMOVAL SPECIFIC
+
```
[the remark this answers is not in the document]
heaversm: nod

controllers/user.js
heaversm: @toufali  - How'd you feel about adding this as a separate ticket? I tried a simple find and replace which didn't work, so I'll need to make sure I understand the syntax well and how it handles the various options we need in fetching data. 

Since this is the core functionality of the service, I'm hesitant to do this as part of the larger merge because I'm not sure what I'll be messing up in the process.

controllers/user.js
heaversm: Yeah - I did it on the controller function names since that was pretty straightforward - but I'll probably pause on everything else.

controllers/user.js
heaversm: Great catch. Didn't even think about that as an option. I'll add a ticket for this one too.

controllers/user.js
heaversm: This will actually be used throughout the pilot - but I'm going to try and arrange for Kanary to give us this data via an SQL query from their own DB rather than making an API request for all of this. I'll add a ticket to have this removed.

js-constants.js
heaversm: It's not provided via the API, but it is listed on their site. It's a good callout. I'll ask Kanary if there is a better way to either retrieve this info as an up-to-date list or deep link to the info on their site instead.

controllers/user.js
toufali: All good!  This is just a nice-to-have; not imperative.

controllers/user.js
toufali: I do a similar thing here, opening a db once on server start, and refreshing every 3 days (disregard comment saying it's every 24hrs – I need to change that):
https://github.com/mozilla/blurts-server/blob/99611af49f1ee60d3a5d92b316a86f5118b63b06/server.js#L99-L102

which calls this:
https://github.com/mozilla/blurts-server/blob/99611af49f1ee60d3a5d92b316a86f5118b63b06/ip-location-service.js#L9-L21

server.js
```diff
@@ -98,7 +124,7 @@ try {
 
 (async () => {
   // open location database once at server start. Service includes 24hr check to reload fresh database.
-  await IpLocationService.openLocationDb().catch(e => console.warn(e));
+  await IpLocationService.openLocationDb().catch((e) => console.warn(e));
```
toufali, line 127: Do you like this formatting rule? 😬 Much cleaner to me without the paren for a single arg.  Then again, I prefer single quotes and no semicolons, so this whole repo already fights with my syntax style 😅

views/layouts/default.hbs
```diff
@@ -1,7 +1,10 @@
 <!DOCTYPE html>
-<html lang="{{ locale }}">
+{{!-- MH: not sure locale is actually loading in properly - this is not data removal specific and should be fixed for prod monitor if broken --}}
+{{!-- MH TODO: verify this locale is actually coming through --}}
+<html lang="{{ getLocale }}">
```
toufali, line 4: This should be updated and working now in `main`.

toufali (commented):
I _think_ I tackled the majority of my review in round 1.  Only a couple very minor comments on the remaining files.  Of course it's likely I missed something, so if there's a particular section of code you'd like to me review, please let me know.  

Additionally, I think we should run Lighthouse or similar on the core `main` site and compare it to the kanary site to make sure users don't experience significant performance degradation.

server.js
```diff
@@ -127,6 +130,19 @@ try {
   await IpLocationService.openLocationDb().catch((e) => console.warn(e));
 })();
 
+//DATA REMOVAL SPECIFIC
+(async () => {
+  // open location database once at server start. Service includes 24hr check to reload fresh database.
```
toufali, line 135: Update or delete this comment since it pertains to a different service.

server.js
heaversm: hah - yeah, I still haven't gotten used to the double quotes vs single - end up doing half and half I think when I write code now...

I actually prefer the most verbose possible (annoyingly) - I like it when everything's written the same (vs some shorthand and some not, so what I would actually write would be:

```
catch((e) => {
  console.warn(e)
});
```

but I'm happy to use whatever you prefer - but we'd just need to change it in the linter / autoformatter level otherwise it'll keep getting changed.

views/layouts/default.hbs
heaversm: 👍🏽

toufali (commented):
I haven't tested the updated waitlist retrieval functionality IRL, but the code makes sense to me on "paper".

server.js
toufali: The most verbose! 😄 Yes, we are on the opposite sides of syntax spectrum.  I'm fine with your more verbose arrow function syntax.  But maybe we don't need to auto-correct all of mine - we can both live in harmony 😉

app-constants.js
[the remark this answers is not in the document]
toufali: I think this is still duplicated?

PR #2408, opened 2021-12-21
replace node-fetch library with got

heaversm:
`node-fetch` was providing duplicate functionality to `got`, which is already in use elsewhere in monitor
replace all usages in user controller
add a few status messages / codes on error

fixes https://mozilla-hub.atlassian.net/browse/FPRDS-5


toufali:
@heaversm This looks good, but I didn't notice an updated `package.json` – we should probably update this and associated lockfile?

PR #2410, opened 2021-12-22
adds the review screen as the default view on the removal form

heaversm:
...when someone already has an account, otherwise shows the form

prepopulates the review screen with account info if available
changes the back button from cancel and moves its location

figma design for back button (out of margin, no alt text, icon, larger than other items in row), would cause some responsive issues as well as accessibility hurdles, keeping it simple for now.

### DIFF
--- a/views/partials/dashboards/remove-form-confirm.hbs
+++ b/views/partials/dashboards/remove-form-confirm.hbs
@@ -1,15 +1,17 @@
 <div class="remove-dashboard-confirm">
   <div class="flx space-between remove-dashboard-confirm-review-container">
-    <h3 class="remove-dashboard-confirm-review-title">{{getRemoveString
-        "remove-dash-confirm-review"
-      }}</h3>
-    <div class="flx remove-dashboard-confirm-review-btns">
+    <div class="flx cntr">
       {{#if this.acctInfo}}
         <a
           href="/user/remove-data"
           class="remove-dashboard-confirm-cancel btn-grey btn-small"
-        >{{getRemoveString "remove-dash-confirm-cancel"}}</a>
+        >{{getRemoveString "remove-dash-confirm-back"}}</a>
       {{/if}}
+      <h3 class="remove-dashboard-confirm-review-title">{{getRemoveString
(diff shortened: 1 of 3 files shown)


views/partials/dashboards/remove-form-confirm.hbs
```diff
@@ -21,25 +23,36 @@
...
+          {{#if this.acctInfo.emails.0.email}}{{this.acctInfo.emails.0.email}}{{/if}}
+        </span>
       </li>
       <li class="remove-dashboard-confirm-item" data-id="location">
         <span class="remove-dashboard-confirm-item-label">{{getRemoveString
             "remove-dash-confirm-loc"
           }}</span>
-        <span class="remove-dashboard-confirm-item-entry"></span>
+        <span class="remove-dashboard-confirm-item-entry">
+          {{#if this.acctInfo.addresses.0.city}}{{this.acctInfo.addresses.0.city}}{{/if}},
```
pdehaan, line 45: If `this.acctInfo.addresses.0.city` is falsy, I think this will have an orphaned comma here. Not sure if the comma should be moved within the `{{#if }}..{{/if}}` block.

heaversm: Hi Peter! Nice one!

views/partials/dashboards/remove-form.hbs
```diff
@@ -144,7 +144,7 @@
         </div>
         {{/if}}
       </div>
-      {{> dashboards/remove-form-confirm}}
+      {{> dashboards/remove-form-confirm acctInfo=acctInfo}}
```
pdehaan, line 147: Q: Above (L3) this was `this.acctInfo`. Should this be `acctInfo=this.acctInfo`?

heaversm: I think the `this` is not needed here. Honestly, we could probably get rid of it in a lot of places. Inside of a loop or hbs helper function, `this` gives context to the loop / function, but otherwise, at top level, it can be omitted.

PR #2413, opened 2021-12-28
implement some lighthouse recommendations

heaversm:
set longer cache control on images and fonts
add apple touch icon
optimize removal images
remove unused font weights and ensure swap on all fonts
rel=noopener noreferrer on all removal external _blank links


public/img/removal/chevron-down-grey.svg
```diff
@@ -1,4 +1 @@
-<svg width="32" height="32" viewBox="0 0 32 32" fill="none" xmlns="http://www.w3.org/2000/svg">
-<title>Grey Down Chevron</title>
```
pdehaan, line 2: @heaversm FYI, looks like we lost all the `<title>` tags you added in https://github.com/mozilla/blurts-server/pull/2409/files.
Not sure how we minified the SVG files (svgo, or other), but if it was svgo, it looks like we might be able to tweak the `removeTitle` config and maybe preserve the `<title>` tags.

toufali: check out svgomg as well for a cool tool

PR #2415, opened 2021-12-29
fix the backend validation requirement for middle name to not be empty

heaversm:
plus

fix a layout issue on the modal trigger for birth year
move the "cancel service" button to the form confirm screen and out of the form itself.

### DIFF
--- a/views/partials/dashboards/remove-form.hbs
+++ b/views/partials/dashboards/remove-form.hbs
@@ -112,7 +112,7 @@
           <div class="input-group">
             <div class="flx flx-cntr">
               <label for="remove-dashboard-form-year" class="input-group-label required">{{ getRemoveString "remove-dash-form-birth-year-label"}}</label>
-              {{> removal/removal-question-trigger modal="modal-birth-year" alt="Why we require birth year"}}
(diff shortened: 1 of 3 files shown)


views/partials/dashboards/remove-form.hbs
```diff
@@ -110,9 +110,10 @@
             <div class="remove-dashboard-form-error --loc"></div>
           </div>
           <div class="input-group">
-            <label for="remove-dashboard-form-year" class="input-group-label field--full required">{{ getRemoveString "remove-dash-form-birth-year-label"}}
+            <div class="flx flx-cntr">
+              <label for="remove-dashboard-form-year" class="input-group-label required">{{ getRemoveString "remove-dash-form-birth-year-label"}}</label>
               {{> removal/removal-question-trigger modal="modal-birth-year" alt="Why we require birth year"}}
```
pdehaan, line 115: Should the `alt` be localized?

toufali (approved):
My local environment is not set up to test this in-browser, so I'll wait until after merge.

views/partials/dashboards/remove-form.hbs
heaversm: I'll put it in strings to be consistent - good catch! (none of the `getRemoveString` function calls actually get localized, but it keeps that possibility open).

PR #2417, opened 2021-12-30
DO NOT MERGE

heaversm:
Posting this here to review the DB seeding and make sure it is configured properly. The goal is that, upon developing locally or deploying this branch for the first time (and first time only), the `removal_pilot` table has one entry with the following content.

@groovecoder / @toufali   - Note the commit comment:

the knexfile does not specify a seed directory, and the only other seedfile is inside db/seeds, which is not the default. Should this file be there as well? Should the knexfile specify this db/seeds directory? Otherwise, the default is [root]/seeds/ (which is where I've placed this seed via the `knex seed:make` command ([reference](https://knexjs.org/#Seeds-CLI)).

If we need to change that config, the knexfile needs something like:

```
development: {
  client: ...,
  connection: { ... },
  seeds: {
      directory: './db/seeds'
  }
}
```

though I'm not sure what the exact values / environments would be here.

### DIFF
--- a/controllers/user.js
+++ b/controllers/user.js
@@ -58,7 +58,7 @@ async function resendEmail(req, res) {
     emailId
   );
 
-  const email = unverifiedEmailAddresxsRecord.email;
+  const email = unverifiedEmailAddressRecord.email;
   await EmailUtils.sendEmail(
     email,
     req.fluentFormat("email-subject-verify"),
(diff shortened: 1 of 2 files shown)


groovecoder (changes requested):
The existing file in `db/seeds` isn't really a knex seed file. ☹️ I.e. - we don't use it via `knex seed:run`. We use its data during `resetDB.js` while running tests. So it's more like a "test fixture" than a real knex seed file.

This new file feels more like a migration than a seed file though? From the docs:

> Seed files allow you to populate your database with test or seed data _independent of your migration files_.
(emphasis mine)

This seed requires migration `db/migrations/20211001135448_removal_pilot_table.js` before it can work, so it seems like this really needs to be `db/migrations/20220103_removal_pilot_first_data.js` ?

PR #2427, opened 2022-01-04
Add ability to opt out of pilot

heaversm:
*Note*: contains DB migrations which must be run in order to test, an…d upon merge / deployment to server

rename previous opt-out modal to manual modal, as it treats manual removal instructions for broker results
create new opt out modal that allows user to opt out of the removal pilot, thus hiding any part of the removal pilot experience from further view
implement this opt out modal trigger both on the enroll screen as well as in the initial data removal signup form (but only when signing up for the first time, not editing info).
handle setting the session variable which allows users to see the removal pilot navigation to false on optout
create a database migration which adds a removal_optout column to the subscribers table - this column will be checked as part of the `checkIfOnRemovalPilotList` function - if the user is on the list but has opted out, the check returns false. This allows us to prevent having to write to the pilot list and re-hash whenever someone opts out.
add an accessibility sr-only button label for all top right modal close buttons.

### DIFF
--- a/controllers/user.js
+++ b/controllers/user.js
@@ -761,6 +761,27 @@ async function getRemovalEnrolledPage(req, res) {
   });
 }
 
+async function handleRemovalOptout(req, res) {
+  const sessionUser = req.user;
+  const optoutRes = await DB.removalOptout(sessionUser);
+  if (!optoutRes || optoutRes !== 1) {
+    const localeError = LocaleUtils.formatRemoveString("remove-error-optout");
+    return res.status(400).json({
+      error: localeError,
+    });
+  }
+
+  if (!req.session || !req.session.kanary) {
+    const localeError = LocaleUtils.formatRemoveString("remove-error-optout");
+    return res.status(400).json({
+      error: localeError,
+    });
+  }
+
+  req.session.kanary.onRemovalPilotList = false; //this must be set so they no longer see the tab in the navigation
+  return res.redirect("/");
+}
+
 async function getRemovalEnrollFullPage(req, res) {
   const user = req.user;
   const utmOverrides = getUTMContents(req);
@@ -1276,6 +1297,19 @@ async function postRemovalKan(req, res) {
   res.redirect("/user/remove-delete-confirmation");
 }
 
+async function postRemovalKan(req, res) {
+  const sessionUser = req.user;
+  const deleteResponse = await removeKanaryAcct(sessionUser.kid);
+  if (!deleteResponse.id) {
+    const localeError = LocaleUtils.formatRemoveString("remove-error-no-id");
+    return res.status(400).json({
+      error: localeError,
+    });
+  }
+  await DB.removeKan(sessionUser);
+  res.redirect("/user/remove-delete-confirmation");
+}
+
 async function checkIfRemovalPilotFull(user) {
   const curPilot = await DB.getRemovalPilotByName(
     REMOVAL_CONSTANTS.REMOVAL_PILOT_GROUP
@@ -1381,7 +1415,19 @@ function checkIfRemoveDisplayMoreTime(user) {
 async function checkIfOnRemovalPilotList(user) {
   if (REMOVAL_CONSTANTS.REMOVE_CHECK_WAITLIST_ENABLED && user) {
     const hashMatch = await checkEmailHash(user.primary_email);
-    return hashMatch;
+    if (hashMatch) {
+      //user is on the list
+      const isOptedOut = await DB.getRemovalOptoutStatus(user);
+      if (isOptedOut) {
+        //have they opted out of the pilot?
+        return false;
+      } else {
+        //user is active in the pilot
+        return hashMatch;
+      }
+    } else {
+      return false;
+    }
   } else {
     console.log("pilot check not enabled");
     return false;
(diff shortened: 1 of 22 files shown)


controllers/user.js
```diff
@@ -761,6 +761,27 @@ async function getRemovalEnrolledPage(req, res) {
...
+    });
+  }
+
+  if (!req.session || !req.session.kanary) {
+    const localeError = LocaleUtils.formatRemoveString("remove-error-optout");
+    return res.status(400).json({
+      error: localeError,
+    });
+  }
+
```
pdehaan, line 780: Should these two `if` statements be combined?

controllers/user.js
```diff
@@ -1276,6 +1297,19 @@ async function postRemovalKan(req, res) {
   res.redirect("/user/remove-delete-confirmation");
 }
 
+async function postRemovalKan(req, res) {
+  const sessionUser = req.user;
+  const deleteResponse = await removeKanaryAcct(sessionUser.kid);
```
pdehaan, line 1302: Q: Will this method ever return `null` or `undefined`?
If so, do we need to add a null check before on L1303 before trying to access the `.id` property?

remove_en/remove.ftl
```diff
@@ -361,12 +362,17 @@ remove-modal-link-text = The following links are the web addresses where your in
...
 
-remove-modal-optout-title = Manual Removal Instructions
-remove-modal-optout-text = If your request is blocked, or if for any reason you wish to attempt to manually remove your information,
+remove-modal-manual-title = Manual Removal Instructions
+remove-modal-manual-text = If your request is blocked, or if for any reason you wish to attempt to manually remove your information,
  you can use the contact email or website provided for this result to do so.
 
+ # Data Removal Optout Modal
+
+remove-modal-optout-title = Confirm Opt-Out of Data Removal Pilot
```
pdehaan, line 373: nit: `Opt-Out` is hyphenated here (and L374 below), but not on L84 above.

views/partials/removal/removal-result.hbs
```diff
@@ -119,7 +119,7 @@
         <li class="remove-dash-results-details-item flx flx-col align-start">
           <p class="remove-dash-results-detail-title flx jst-cntr">
             {{getRemoveString "remove-result-opt-out"}}
-            {{> removal/removal-question-trigger modal="modal-remove-optout" alt="About Manual Removal"}}
+            {{> removal/removal-question-trigger modal="modal-remove-manual" alt="About Manual Removal"}}
```
pdehaan, line 122: Q: Should the `alt` be localized?

controllers/user.js
heaversm: ah - no, but good catch - the error response printed out in the second is incorrect.

controllers/user.js
heaversm: @pdehaan  - yes! It's possible if there's an error on Kanary's side. Good catch

remove_en/remove.ftl
heaversm: 👍🏽

views/partials/removal/removal-result.hbs
heaversm: on 🔥 ! Yes.

views/partials/dashboards/remove-enroll.hbs
```diff
@@ -27,9 +27,15 @@
         <input id="" class="js-enroll-submit remove-enroll-submit" type="submit" value="{{getRemoveString "remove-enroll-submit"}}"/>
       </form>
     </div>
+    <div class="flx jst-cntr">
+        <button
```
toufali, line 31: The default type for a `button` element is `submit`.  I _think_ you're safe here because the button is outside of the form element, but to be sure this button doesn't submit the form (on enter keypress, perhaps) you could add `type="button"`.

I've always found having to add `type="button"` to a `<button>` annoyingly redundant, but usually do it whenever a form is present on the page.

views/partials/dashboards/remove-form.hbs
```diff
@@ -129,14 +129,22 @@
...
           <div class="input-group-button inline">
             <input id="remove-dashboard-data-submit" class="js-remove-submit input-group-submit remove-dashboard-data-submit inline" type="submit" 
               value="{{ getRemoveString "remove-dash-change-submit" }}"/>
             {{> forms/loader }}
           </div>
-          <div class="flx flx-col flx-cntr">
-            <p class="alert txt-cntr remove-dashboard-form-rqd bold">{{ getRemoveString "remove-dash-form-required-helper"}}</p>
+          {{#unless this.acctInfo}}
+          <div class="flx jst-cntr">
+            <button
```
toufali, line 142: See above comment RE `type="button"`

toufali (approved):

controllers/user.js
```diff
@@ -1266,7 +1289,7 @@ async function getRemovalKan(req, res) {
 async function postRemovalKan(req, res) {
   const sessionUser = req.user;
   const deleteResponse = await removeKanaryAcct(sessionUser.kid);
-  if (!deleteResponse.id) {
+  if (!deleteResponse || !deleteResponse.id) {
```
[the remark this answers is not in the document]
heaversm: @toufali  - whoah! Never seen this syntax. Thanks!

views/partials/dashboards/remove-enroll.hbs
heaversm: Ah, that's good to know - that's part of the reason I did this stuff in a model is because I didn't want to deal with a button inside a form.

views/partials/dashboards/remove-form.hbs
heaversm: 👍🏽

PR #2430, opened 2022-01-06
Merge/mk 025 no manage info

heaversm:
adds a removal constant that toggles the ability to edit one's info.
implements a check on the info page for this constant:

- if the ability to edit is disabled users should be allowed to edit their info on initial form entry, but not on subsequent entries - in which case they will see an email address they should contact to change their info
- if the ability to edit is enabled, users should be able to edit their info regardless of whether it is their initial entry or a subsequent edit, and should no longer see the contact email address message.

If the user is editing their info, regardless of whether the ability to edit is enabled or disabled, the user should be able to go back via the "back" button on the manage info screen.

### DIFF
--- a/views/partials/dashboards/remove-form-confirm.hbs
+++ b/views/partials/dashboards/remove-form-confirm.hbs
@@ -12,9 +12,17 @@
           class="remove-dashboard-confirm-cancel btn-grey btn-small"
         >{{getRemoveString "remove-dash-confirm-back"}}</a>
       {{/if}}
-      <a href="#" class="js-confirm-edit-btn btn-blue btn-small">{{getRemoveString
+      {{#if this.acctInfo}}
+        {{#if this.allowEditRemovalInfo}}
+        <a href="#" class="js-confirm-edit-btn btn-blue btn-small">{{getRemoveString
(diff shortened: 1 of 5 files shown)


views/partials/dashboards/remove-form-confirm.hbs
```diff
@@ -58,11 +66,24 @@
...
+    {{#if this.acctInfo}}
+      {{#if this.allowEditRemovalInfo}}
+      <a
+        href="#"
+        class="js-remove-confirm remove-dashboard-confirm-submit jst-cntr flx btn-blue-primary"
+        role="button"
+      >{{getConfirmSubmitText}}</a>
+      
+      {{else}}
+      <p class="remove-dashboard-conirm-edit">{{getRemoveString "remove-form-change-contact-msg"}}: <a href="mailto:{{ getRemoveString "remove-about-support-email"}}">{{ getRemoveString "remove-about-support-email"}}</a></p>
```
pdehaan, line 78: typo? Should `remove-dashboard-conirm-edit` be `-confirm-`?

heaversm: Class should be removed. Thanks!

PR #2433, opened 2022-01-06
remove unused routes, css, templates and constants around:

heaversm:
- remove signup confirmation page (user is now sent straight to the dashboard after signup)
- removal willingness to pay screen (this feature was abandoned)


toufali (commented):
Looks good to me after that missing function lint is fixed 👍

PR #2435, opened 2022-01-06
prevents a user from signing up for the pilot again after account deletion

heaversm:
prevents a user from signing up for the pilot again by setting their optout status to true on kanary account deletion or fxm account deletion

### DIFF
--- a/controllers/user.js
+++ b/controllers/user.js
@@ -433,7 +433,9 @@ async function postRemoveFxm(req, res) {
     await DB.removeKan(sessionUser);
   }
   //END DATA REMOVAL SPECIFIC
-
+  if (req.session?.kanary) {
+    req.session.kanary.onRemovalPilotList = false;
+  }
   await DB.removeSubscriber(sessionUser);
   await FXA.revokeOAuthTokens(sessionUser);
 
@@ -771,7 +773,7 @@ async function handleRemovalOptout(req, res) {
     });
   }
 
-  if (!req.session || !req.session.kanary) {
+  if (!req.session?.kanary) {
     const localeError = LocaleUtils.formatRemoveString(
       "remove-error-no-session"
     );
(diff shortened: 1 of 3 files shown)


controllers/user.js
```diff
@@ -433,7 +433,9 @@ async function postRemoveFxm(req, res) {
     await DB.removeKan(sessionUser);
   }
   //END DATA REMOVAL SPECIFIC
-
+  if (req.session?.kanary) {
+    req.session.kanary.onRemovalPilotList = false;
```
pdehaan, line 437: Q: Is it possible that _`kanary`_ could be `undefined`? Or is every `req.session` object guaranteed to have a `.kanary` property (and therefore trying to access `.onRemovalPilotList` property wouldn't error out).

controllers/user.js
```diff
@@ -1246,6 +1248,9 @@ async function postRemovalKan(req, res) {
     });
   }
   await DB.removeKan(sessionUser);
+  if (req.session?.kanary) {
+    req.session.kanary.onRemovalPilotList = false;
```
pdehaan, line 1252: Same Q as above. I believe this is checking that `req.session` is non-null before checking the `.kanary` property, but do we similarly need to check if `req.session.kanary` is non-null before setting `.onRemovalPilotList`?

controllers/user.js
heaversm: @pdehaan  - just learned this from Amri, but that `req.session?.kanary` is an [optional chain](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining) - meaning if anything beyond the `req.session` object is undefined, that code doesn't execute. The code is an object, so setting `onRemovalPilotList` is fine (though maybe it'd be better to do `if req.session?.onRemovalPilotList`)

pdehaan: Oh interesting, I'll have to poke at that. I always thought you need to do something like `req.session?.kanary?`.
But yeah, makes sense. Rereading the code it will see if `req.session?.kanary` is defined/truthy, so if it was `undefined` the `if() {...}` block would fail and not try and set the `.onRemovalPilotList` property anyways.

heaversm: Screen Shot 2022-01-06 at 4 25 19 PM

PR #2436, opened 2022-01-07
Disable show=true parameter ability on stage / prod

heaversm:
modify the `show=true` parameter check  so that it can only be allowed if the `NODE_EV` parameter is for `dev` or `heroku`

### DIFF
--- a/controllers/user.js
+++ b/controllers/user.js
@@ -696,21 +696,20 @@ function logout(req, res) {
 async function getRemovalEnrollPage(req, res) {
   const user = req.user;
 
-  if (
-    checkIfRemovalEnrollmentEnded(user) &&
-    !FormUtils.canShowViaParams(req.query?.show)
-  ) {
+  const canShowViaParams = FormUtils.canShowViaParams(req.query?.show);
+
+  if (checkIfRemovalEnrollmentEnded(user) && !canShowViaParams) {
     //If the pilot enrollment period is not active
     return res.redirect("/user/remove-enroll-ended");
   }
 
   const isEnrolledInPilot = checkIfEnrolledInRemovalPilot(user);
-  if (isEnrolledInPilot && !FormUtils.canShowViaParams(req.query?.show)) {
+  if (isEnrolledInPilot && !canShowViaParams) {
     //if the user has already enrolled in the pilot, just send them to the remove data page
     return res.redirect("/user/remove-data");
   }
 
-  if ((await checkIfRemovalPilotFull(user)) && !FormUtils.canShowViaParams(req.query?.show)) {
+  if ((await checkIfRemovalPilotFull(user)) && !canShowViaParams) {
     //If we have already hit the enrollment limit:
     return res.redirect("/user/remove-enroll-full");
   }
@@ -750,7 +749,10 @@ async function getRemovalEnrolledPage(req, res) {
 
   const experimentFlags = getExperimentFlags(req, EXPERIMENTS_ENABLED);
 
-  if (!checkIfEnrolledInRemovalPilot(user) && !FormUtils.canShowViaParams(req.query?.show)) {
+  if (
+    !checkIfEnrolledInRemovalPilot(user) &&
+    !FormUtils.canShowViaParams(req.query?.show)
+  ) {
     return res.redirect("/user/remove-enroll");
   }
 
@@ -794,7 +796,10 @@ async function getRemovalEnrollFullPage(req, res) {
 
   const experimentFlags = getExperimentFlags(req, EXPERIMENTS_ENABLED);
 
-  if (!(await checkIfRemovalPilotFull(user)) && !FormUtils.canShowViaParams(req.query?.show)) {
+  if (
+    !(await checkIfRemovalPilotFull(user)) &&
+    !FormUtils.canShowViaParams(req.query?.show)
+  ) {
     return res.redirect("/user/remove-enroll");
   }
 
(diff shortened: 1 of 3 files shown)


controllers/user.js
```diff
@@ -696,18 +696,21 @@ function logout(req, res) {
 async function getRemovalEnrollPage(req, res) {
   const user = req.user;
 
-  if (checkIfRemovalEnrollmentEnded(user) && !req.query.show) {
+  if (
+    checkIfRemovalEnrollmentEnded(user) &&
+    !FormUtils.canShowViaParams(req.query?.show)
```
pdehaan, line 701: Q: Should we call `FormUtils.canShowViaParams(req.query?.show)` once and be caching the result?
Looks like we might be calling it multiple times in some methods (a total of 14&times; in this PR, but most of them seem to be in different methods though).

But the `getRemovalEnrollPage(req, res)` method here seems to call it 3&times;.

controllers/user.js
```diff
@@ -830,11 +833,11 @@ async function getRemovalEnrollEndedPage(req, res) {
 async function getRemovalPage(req, res) {
   const user = req.user;
 
-  if (!checkIfEnrolledInRemovalPilot(user) && !req.query.show) {
+  if (!checkIfEnrolledInRemovalPilot(user) && !FormUtils.canShowViaParams(req.query?.show)) {
     return res.redirect("/user/remove-enroll");
   }
 
-  if (checkIfRemovalPilotEnded(user) && !req.query.show) {
+  if (checkIfRemovalPilotEnded(user) && !FormUtils.canShowViaParams(req.query?.show)) {
```
pdehaan, line 840: Looks like we're checking `!FormUtils.canShowViaParams(req.query?.show)` 2&times; here.

Not sure if we should do something like:

```js
if (!FormUtils.canShowViaParams(req.query?.show)) {
  if (!checkIfEnrolledInRemovalPilot(user)) {
    return res.redirect("/user/remove-enroll");
  }
  if (checkIfRemovalPilotEnded(user)) {
    // if the pilot is over, redirect to the end screen
    return res.redirect("/user/remove-pilot-ended");
  }
}
```

controllers/user.js
```diff
@@ -936,11 +939,11 @@ async function getRemovalPage(req, res) {
 async function getRemovalConfirmationPage(req, res) {
   const user = req.user;
 
-  if (!user.kid && !checkIfEnrolledInRemovalPilot(user) && !req.query.show) {
+  if (!user.kid && !checkIfEnrolledInRemovalPilot(user) && !FormUtils.canShowViaParams(req.query?.show)) {
     return res.redirect("/user/remove-enroll");
   }
 
-  if (!user.kid && checkIfEnrolledInRemovalPilot(user) && !req.query.show) {
+  if (!user.kid && checkIfEnrolledInRemovalPilot(user) && !FormUtils.canShowViaParams(req.query?.show)) {
```
pdehaan, line 946: Meh... One more level of `if`, but DRYs it slightly. Not sure it's any better.
```js
  if (!user.kid && !FormUtils.canShowViaParams(req.query?.show)) {
    if (!checkIfEnrolledInRemovalPilot(user)) {
      return res.redirect("/user/remove-enroll");
    }
    if (checkIfEnrolledInRemovalPilot(user)) {
      return res.redirect("/user/remove-data");
    }
  }
```

controllers/user.js
```diff
@@ -992,11 +995,11 @@ async function getRemovalConfirmationPage(req, res) {
 async function getRemovalUpdateConfirmationPage(req, res) {
   const user = req.user;
 
-  if (!user.kid && !req.query.show) {
+  if (!user.kid && !FormUtils.canShowViaParams(req.query?.show)) {
     return res.redirect("/user/remove-enroll");
   }
 
-  if (user.kid && !req.query.show) {
+  if (user.kid && !FormUtils.canShowViaParams(req.query?.show)) {
```
pdehaan, line 1002: Again, not sure it's "better", or even more readable.
```js
  if (!FormUtils.canShowViaParams(req.query?.show)) {
    if (!user.kid) {
      return res.redirect("/user/remove-enroll");
    }
    return res.redirect("/user/remove-data");
  }
```

form-utils.js
```diff
@@ -93,6 +94,13 @@ const FormUtils = {
     )[0];
     return parseFloat(mode);
   },
+  canShowViaParams(doShow) {
+    if (["dev", "heroku"].includes(AppConstants.NODE_ENV) && doShow) {
+      return true;
+    } else {
+      return false;
+    }
```
pdehaan, line 102: Since you're just returning a boolean, you could one-liner this to:

```js
    return ["dev", "heroku"].includes(AppConstants.NODE_ENV) && doShow;
```

controllers/user.js
heaversm: Well - it checks each page's `show` URL parameter, so that can't be cached - the other piece of this is just a lookup of `AppConstants.NODE_ENV` so that is essentially required at the top of the file and thus essentially cached.

controllers/user.js
heaversm: I'm up for that.

controllers/user.js
pdehaan: Yeah, I think only 3 methods called it multiple times. Probably not worth over-optimizing and you can ignore.

form-utils.js
heaversm: 👍🏽

controllers/user.js
heaversm: @pdehaan  -Ah - sorry - I missed that we are calling it multiple times in the same function, sorry. Will implement your suggestion!

controllers/user.js
heaversm: Going to leave as is unless there's any strong objection.

PR #2444, opened 2022-01-11
add 3 data removal specific strings to main 

heaversm:
for localization so they don't have to be added each time we merge main into main-kanary


locales/en/app.ftl
```diff
@@ -827,3 +827,9 @@ vpn-banner-cta-close = Close
 relay-video-headline = How do you keep your email address a secret?
 # VPN video headline for placement on Monitor user's dashboard
 vpn-video-headline = Do you worry about safety on public Wi-Fi?
+
+# DATA REMOVAL SPECIFIC
+
+remove-header-exposures = Data Removal
+remove-footer-data-removal-faq = Data Removal FAQ
+remove-faq-title = Data Removal Pilot FAQ
```
flodolo, line 835: Please keep an empty line at the end of the file.

flodolo (commented):
Is there a reason not to have these in the `remove_en/remove.ftl`file in the kanary branch? 

If we add them, they'll get localized. I also wouldn't expect to not have code using them in `main`?

locales/en/app.ftl
heaversm: 👍🏽

heaversm:
> Is there a reason not to have these in the `remove_en/remove.ftl`file in the kanary branch?
> 
> If we add them, they'll get localized. I also wouldn't expect to not have code using them in `main`?

@flodolo  - yes - the reason is that there is a function that runs in a loop to execute the generation of the navigation links, and calls the `getLocaleString` function, which looks to the locale files for formatting of the strings. 

To not use this file, we have to write a rather kludgy conditional to make an exception for the data removal strings, versus just adding three strings for translation.

PR #2450, opened 2022-01-12
prevent localization of non-monitor footer and header nav links

heaversm:


toufali (approved):

locales/en/app.ftl
```diff
@@ -827,7 +827,3 @@ vpn-banner-cta-close = Close
 relay-video-headline = How do you keep your email address a secret?
 # VPN video headline for placement on Monitor user's dashboard
 vpn-video-headline = Do you worry about safety on public Wi-Fi?
-
```
toufali, line 830: Can we add an empty line here?  Or maybe it already exists and the diff doesn't show...

toufali (changes requested):

locales/en/app.ftl
toufali: ahh -- I just checked my pulled branch and there _does_ appear to be an empty line.  Dunno why the diff doesn't show, but I've experienced this in the past as well.

toufali (approved):

PR #2451, opened 2022-01-12
Add educational videos for VPN and Relay products

toufali:


views/partials/product-education-video.hbs
```diff
@@ -0,0 +1,6 @@
+<aside class="product-education {{ theme }}">
+  {{#with (productEducation type)}}
+  <h2>{{ headingTxt }}</h2>
+  <video src="{{ videoSrc }}" poster="{{ posterSrc }}" width=" 3840" height="2160" type="video/mp4" controls></video>
```
pdehaan, line 4: nit: space in the `width=` attribute (before 3840).

pdehaan: Those are some pretty big `width` and `height` attributes. How does that look on mobile?

toufali: Oooh, the space in the width value is super annoying, thank you for catching that.

RE sizes, these are just the intrinsic dimensions of the video (4K).  In reality, it's actually compressed to to 720p (1280 width).  Furthermore, the CSS caps the width to 960 which is qHD – (one-quarter of full HD).  This gives the video nice even divisions of each other, and still looks crisp when you go full-screen.  

It should take up the full width on mobile without causing overflow, and look crisp on higher res screens.

PR #2460, opened 2022-01-14
prelaunch adjustments

heaversm:
- new hashed waitlist
- rework the requireRemovalUser middleware to first check the session variable, and if it is not present, check the hashed waitlist
- update start times of pilot, with a temporary date for the first that allows us to test before launch
- rework oauth step so that the removalPilot check and the session variable is set appropriately for either new or existing users
- set kanary session variable to false when user removes their monitor account if present
- set the optout to true on opt out now that we're done testing that bit
- update terms and privacy start dates for pilot

TODO before launch: adjust the first pilot start date before launch and remove unnecessary test accounts from waitlist


controllers/user.js
```diff
@@ -787,6 +788,7 @@ async function handleRemovalOptout(req, res) {
   }
 
   req.session.kanary.onRemovalPilotList = false; //this must be set so they no longer see the tab in the navigation
+  req.session.destroy();
```
toufali, line 791: is line 790 setting var false redundant if we're just destroying the session next line?

removal-constants.js
```diff
@@ -53,13 +52,14 @@ const REMOVAL_CONSTANTS = {
     {
       id: 1,
       name: "group_01",
-      start_time: 1641916800, //wed, jan 11 8am PST
+      start_time: 1642197405, //MH TODO: remove after testing
```
toufali, line 55: Does this need to be removed?

toufali (approved):

controllers/oauth.js
```diff
@@ -101,6 +98,14 @@ async function confirmed(req, res, next, client = FxAOAuthClient) {
       fxaProfileData
     );
 
+    console.log("new subscriber", verifiedSubscriber);
```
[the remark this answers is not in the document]
heaversm: Nope - good catch. Will remove.

controllers/user.js
heaversm: Hah...good point. I think so. Will remove.

removal-constants.js
heaversm: I kept it in so QA could test - will change now for launch.

PR #2464, opened 2022-01-18
merge `main` into `main-kanary`

toufali:


heaversm (commented):
Looks good I think - essentially you're just introducing a default of 0 if for some reason the header can't be found or hasn't been rendered to the DOM yet? 

I wonder what was happening to give the DOM such high pixel values? For Setareh, it seemed like the issue was not that the header couldn't be found, it was that it was too big. 
Screen Shot 2022-01-18 at 11 43 22 AM

toufali:
I believe the issue was due to old CSS files cached on prod – it was only noticed on the new pages, and not reproduce-able after hard-refresh.  I noticed the similar layout issues earlier from prod.

Specifically, the new code sets a 0px default, and as well adds an additional height check/update on DOM content load.  Both should be benign and added for safety, in case the issue was something more than cached CSS.

PR #2465, opened 2022-01-19
Improve VPN banner location results/readability

toufali:


controllers/ip-location.js
```diff
@@ -7,7 +7,13 @@ const AppConstants = require("../app-constants");
...
 
 async function getIpLocation(req, res) {
-  const clientIp = AppConstants.NODE_ENV === "dev" ? "216.160.83.56" : req.ip; // for dev, return an IP that exists in GeoLite2 test DB, e.g. 216.160.83.56
+  let clientIp = req.ip;
+
+  if("test-ip" in req.headers){
+    clientIp = req.headers["test-ip"]; // allow testing IP addresses via custom header
+  }else if(AppConstants.NODE_ENV === "dev") {
+    clientIp = "216.160.83.56"; // local GeoLite2 test DB has a limited dataset – return an IP that exists, e.g. 216.160.83.56
+  }
```
pdehaan, line 16: To be honest, not sure I understand 100%.
What is `NODE_ENV === "dev"`? Is that only for localhost development?

I [ignorantly] was expecting more of:

```js
// Only allow IP override on Dev or Heroku (not stage or production).
if (["dev", "heroku"].includes(AppConstants.NODE_ENV)) {
  // Use `req.headers["test-ip"]` if truthy, otherwise fall back to default testing IP.
  clientIp = req.headers["test-ip"] || "216.160.83.56";
}
```

Tested the URL you gave me and it worked great for Playwright+screenshots. 

Not sure if we bother making it more flexible (if it only runs in dev+Heroku) by also checking `req.query.test_ip`, but not sure who else might use this that they'd want to tweak it that way (unless this is something localizers would be checking).

pdehaan: Posted in Slack, but I'll dump here for future reference too since I know I'll forget the repo URL:
https://github.com/pdehaan/playwright-blurts-vpn-test

Includes a list of random IPs/locations and their respective banner screenshots.

toufali: Yes, `NODE_ENV === "dev"` is for localhost.  I like your solution much better for multiple reasons.  `else if` rarely passes my sniff test, but I was in a rush 😅

toufali: Actually -- for Heroku I added a buildpack that opens up the much larger db.  It's only localhost that needs that static IP.  Might give this some more thought.

pdehaan (approved):
R+. 
lets test it on Heroku.

PR #2472, opened 2022-01-20
remove style nonce for CSP inline style rule

toufali:
CSP does not allow inline styles.  The style nonce code was added to safely bypass this, as well resolve related client errors.  Since we are not using inline styles, and I cannot replicate errors, this PR removes the associated code.

Removing the code removes the ability to bypass CSP for inline styles, and as well allows syncing between `main` and `main-kanary` to proceed without conflict.


heaversm (approved):
PR looks good, thanks for doing this. Didn't see any additional errors on my side.

Did a search for inline `style=""` references too - seems like only the svgs and email-related items make use of inline styles currently. The SVGs may throw a console error but I don't think it would affect anything.

PR #2476, opened 2022-01-21
Pilot/003 pilot mgmt

heaversm:
add a route that lets mozilla employees update a member's opt-in status by email address, since we don't have direct access to PSQL for the DB.

Adds middleware to do the admin email check so that we can add in different places.


middleware.js
```diff
@@ -298,6 +298,21 @@ async function requireNoOptOut(req, res, next) {
   }
 }
 
+async function requireMozAdmin(req, res, next) {
+  const user = req.user;
+
+  if (!user?.primary_email.includes("@mozilla.com")) {
+    console.error("non mozilla email");
+    return res.status(404).json({
```
toufali, line 306: Should this be error code 403 or 401?

toufali (approved):
Might want to reach out to @pdehaan or @groovecoder for a second pair of eyes, since this PR might expose us to more risk?

controllers/user.js
```diff
@@ -1937,6 +1919,46 @@ async function createRemovalHashWaitlist(req, res) {
...
+      return res.status(400).json({
+        error: "no email or optout param",
+      });
+    }
+  } else {
+    return res.status(400).json({
+      error: "no query",
+    });
+  }
+}
```
groovecoder, line 1960: I've started to treat `else` blocks as code "sniffs". I refactor them out by inverting their logic and returning from the function early. It avoid multiple levels of nested logic branches. I also treat `let` as a code "sniff" too. Make immutable `const` vars helps prevent future changes to this code from accidentally changing the variable value in a way that the original code didn't take into consideration.

This should be the same logic, in a more read-able form (not tested):

```suggestion
  if (!req.query || !Object.entries(req.query).length > 0) {
    return res.status(400).json({
      error: "no query",
    });
  }

  const { email, optout } = req.query;
  if (!email || !optout) {
    return res.status(400).json({
      error: "no email or optout param",
    });
  }

  if (!["true", "false"].includes(optout)) {
    return res.status(400).json({
      error: "no valid optout status",
    });
  }

  const doOptOut = (optout === "true");
  const removeResponse = await DB.handleRemovalOptOutByEmail(
    email,
    doOptOut
  );
  const removeSuccess = Boolean(removeResponse);
  if (!removeSuccess) {
    return res.status(400).json({
      error: "no record found",
    });
  }

  return res.status(200).json({
    msg: `${email} optout status set to ${doOptOut}`,
  });
```

middleware.js
```diff
@@ -298,6 +298,21 @@ async function requireNoOptOut(req, res, next) {
...
+  if (!user?.primary_email.includes("@mozilla.com")) {
+    console.error("non mozilla email");
+    return res.status(401).json({
+      error:
+        "You must be signed in with a mozilla.com email address to access this page",
+    });
+  } else {
+    next();
+    return;
+  }
```
groovecoder, line 313: This doesn't need an `else` block, because if the code flowed into the above `if` block, it will already call `return` and never execute the code in `else`. Somewhat simpler form:

```suggestion
  if (!user?.primary_email.includes("@mozilla.com")) {
    console.error("non mozilla email");
    return res.status(401).json({
      error:
        "You must be signed in with a mozilla.com email address to access this page",
    });
  }
  next();
  return;
```

groovecoder (changes requested):
I didn't spot-check, and the comments are mostly about code simplicity.

controllers/user.js
heaversm: Thanks! Definitely better structure.

middleware.js
heaversm: 👍🏽

middleware.js
```diff
@@ -298,6 +298,21 @@ async function requireNoOptOut(req, res, next) {
...
 }
 
+async function requireMozAdmin(req, res, next) {
+  const user = req.user;
+
+  if (!user?.primary_email.includes("@mozilla.com")) {
+    console.error("non mozilla email");
+    return res.status(401).json({
+      error:
+        "You must be signed in with a mozilla.com email address to access this page",
```
groovecoder, line 308: Let's not give away that someone could do this with an @mozilla.com email address. Make an attacker have to crawl thru our source code to figure that out.

groovecoder (changes requested):

middleware.js
```diff
@@ -302,10 +302,9 @@ async function requireMozAdmin(req, res, next) {
   const user = req.user;
 
   if (!user?.primary_email.includes("@mozilla.com")) {
```
pdehaan, line 304: Q: Could we use [`.endsWith()`](https://developer.mozilla.org/docs/Web/JavaScript/Reference/Global_Objects/String/endsWith) here instead of the more generic/risky `.includes()`?
Might be fractionally more secure.

heaversm:
The new approach exposes a new route:

`user/remove-pilot-mgmt`

Here, if you have an 'endswith' moz address, you can access the route, and reset a user's optin status, cancel an account by KID, or get a KID by email.

groovecoder (approved):

PR #2477, opened 2022-01-21
allow editing info via an edit=true query param

heaversm:
Allows users to edit their info by inclusion of an `edit=true` param. 

We are getting requests from support and research of users who input improper information - this will allow them to easily update their own info without us needing to make SRE requests or obtain personal information from our users. 

The edit info functionality itself was preserved from the foxfooding pilot, so this PR only requires that we set the feature flag by way of the param.


groovecoder:
Post-merge review looks good to me. I don't see any place we're displaying this user-passed `edit` value.

PR #2484, opened 2022-01-26
resolve postcss alert for stylelint

toufali:
Explanation of why this PR doesn't work:
https://github.com/stylelint-scss/stylelint-config-standard-scss/issues/5

The conflicting package is [gulp-sourcemaps](https://github.com/gulp-sourcemaps/gulp-sourcemaps) v3.0.0, which specifies an earlier version of PostCSS.

### DIFF
--- a/.stylelintrc
+++ b/.stylelintrc
@@ -1,9 +1,7 @@
 {
   "extends": [
-    "stylelint-config-standard"
-  ],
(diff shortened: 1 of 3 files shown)


toufali:
closing because Gulp has since been removed, with associated vulnerable packages

PR #2485, opened 2022-01-27
change limit and add a day for enrollment

heaversm:


removal-constants.js
```diff
@@ -62,7 +62,7 @@ const REMOVAL_CONSTANTS = {
       max_users: 1000, //max users represents the sum of all users from this and the preceding pilot groups (thus 700 in this cohort and 1000 total from both) - so that we can compare against a single DB value of max users
     },
   ],
-  REMOVAL_PILOT_ENROLLMENT_END_DAY: 11, //days from pilot start to when a user can no longer enroll in their pilot group
+  REMOVAL_PILOT_ENROLLMENT_END_DAY: 12, //days from pilot start to when a user can no longer enroll in their pilot group
```
toufali, line 65: This is probably right, but just double-checking it shouldn't be `13` in order to reach Monday?

removal-constants.js
```diff
@@ -53,7 +53,7 @@ const REMOVAL_CONSTANTS = {
       id: 1,
       name: "group_01",
       start_time: 1642492740, //monday, jan 18 12am PST
-      max_users: 300,
+      max_users: 400,
```
toufali, line 56: Was this supposed to be `500`?

PR #2488, opened 2022-01-28
Add breach logos for RedDoorz and BTC-Alpha

maxxcrawford:
Preview: 
BTCAlpha
RedDoorz


pdehaan:
@toufali Quick Q... For logos, should we be PR-ing those into **main-kanary** branch, or should those go into **main** directly (and then possibly remerged from main=>main-kanary).
Just want to make sure we don't lose some logos/commits once we end main-kanary experiment and return back to main branch.

pdehaan (approved):
Logos look good. R+
Awaiting confirmation from Amri on which branch these should live in.

toufali:
@pdehaan – per our recent chat you are correct: 
- All changes that are intended to live on past the pilot should go to `main` first.  
- Only changes that are specific to the pilot should go directly to `main-kanary`.

This particular PR change would have gone to the CDN regardless, however the logos would not appear locally post-pilot if this was only merged to `main-kanary`.

I've updated the target branch to `main` 💪 

cc @maxxcrawford

pdehaan:
Works for me! Thanks @toufali and @maxxcrawford.
Mergin'!

PR #2493, opened 2022-02-10
secure pilot mgmt admin interface

heaversm:
secure pilot mgmt admin interface by creating a hashed admin list and checking against it when accessing the mgmt interface

**note:** if you want to develop locally and generate a new list, you will need to first be on the list (myself, @groovecoder, or @toufali  right now). You will also need to either make a new list, or get the `removal-admins.txt` file from me.


controllers/user.js
```diff
@@ -1813,6 +1812,46 @@ async function createRemovalHashWaitlist(req, res) {
   });
 }
 
+async function createRemovalHashAdminList(req, res) {
```
groovecoder, line 1815: We can probabaly just get rid of this whole create-a-file step. We could add a single env var:
```
REMOVAL_ADMINS=26bb92cc8f4c7be7ea2adb4611fd713358eaf128,12eeb998fe0c7600b3a3aacf518aa760b1dbd775,df79383a705df15058f290aad1bb0f4c382a81ec
```

removal-waitlist.js
```diff
@@ -7,6 +7,12 @@ async function getHashedWaitlist() {
   return hashedWaitlistArray.toString().split("\n");
 }
 
+async function getHashedAdminList() {
+  const hashedAdminList = await readFile("hashed-admin-list.txt", "binary");
+  return hashedAdminList.toString().split("\n");
```
groovecoder, line 12: If we pull the admin hashes from an env var, this code can become:
```suggestion
  const hashedAdminList = AppConstants.REMOVAL_ADMINS;
  return hashedAdminList.toString().split(",");
```

groovecoder: Though now that I think about, if we put them list of admins in an env var, there's no need to hash them anymore either.

groovecoder (approved):
Approved, but now that I see it coded out, I realize it might be simpler (if the list is small) to just move the list of admins into an env var like `REMOVAL_ADMINS`, and then check that? It eliminates the need to create & maintain another hash list file.

PR #2502, opened 2022-02-17
remove pilot ending banner

heaversm:
to resolve issue with adding additional emails.

Requiring the `user` controller within `hbs-helpers` was causing some sort of load order issue, in which `const EmailUtils = require("../email-utils");` within the `user` controller was resolving as `undefined`. I'm not quite sure why that was happening, but for now, this pull request removes that banner altogether, as well as the code in the helper which was requiring the user controller.

We will need to come up with a method of showing a "pilot ending" banner sometime before April, but for now, pushing this PR to prevent further user issues with adding additional emails.

### DIFF
--- a/controllers/user.js
+++ b/controllers/user.js
@@ -1338,17 +1338,6 @@ function getPilotGroup(user) {
   return removalPilots[0];
 }
 
-function checkIfRemovalPilotEnding(user) {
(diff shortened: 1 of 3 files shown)


maxxcrawford (approved):
Works as expected for a normal (non-Kanary) account. Thanks!

PR #2508, opened 2022-03-04
Update breach logos on Kanary main branch

maxxcrawford:
Updated the /logos folder with what's on `main` for the Kanary branch.


pdehaan:
I admit that I'm a bit confused here... Many of these are going FROM 150x150 to larger and non-square sizes (which seems like the opposite direction).

pdehaan (changes requested):
Yeah, no.

PR #2513, opened 2022-03-17
Add CDEK logo

codemist:
Preview:
CDEK


pdehaan (approved):
Yep. It's a logo!
150&times;150px, and a svelte 1.8 kB.

Rebased onto **main** branch.

pdehaan:
Verified on prod.
Screen Shot 2022-03-17 at 1 36 25 PM

Breach is [`IsVerified=false`](https://haveibeenpwned.com/api/v3/breach/CDEK), which means I don't think it appears in https://monitor.firefox.com/breaches.

PR #2520, opened 2022-03-25
add kanary pilot offboarding flow

toufali:


heaversm:
Really simple, clean approach! Seems like then there will be no "activation date" for the banner, you'll just push this code live, and manually change the group from `1` to `2` in the `showRemovalEndingBanner` at the appropriate date?

I know Setareh had mentioned she'd ideally like this on all pages for pilot participants (most important would probably be the regular monitor dashboard), but maybe y'all decided otherwise since I've been out. 

As far as testing you could manually update the DB for the folks who need to test like Setareh - you can use any (imaginary) KID and set the `removal_enrolled_time` to a timestamp that works to either trigger (e.g. 2022-01-20 22:22:50.712+00) or not trigger (2022-01-20 22:22:50.712+00) the banner.

Let me know if you need the latest waitlist.txt file.

heaversm (approved):

PR #2524, opened 2022-04-05
add CSAT banner

toufali:


public/js/monitor.js
```diff
@@ -326,82 +326,82 @@ function recruitmentLogic() {
...
+//               "hitCallback": window.location.href = relayCta.href,
+//             });
+//           }
+//           return;
+//         }
+//         ga("send", "event", "Waitlist Test", "Engage", relayCta.dataset.analyticsLabel);
+//       });
+//     });
+//   }
+// }
```
groovecoder, line 404: question: Should we just go ahead an remove this code? Or stash it in a .js file that isn't used or referenced?

public/js/monitor.js
```diff
@@ -510,6 +510,35 @@ async function initVpnBanner() {
...
+
+  function handleEvent(e) {
+    const ttl = new Date();
+    ttl.setDate(ttl.getDate() + 90);
+    document.cookie = "csatHidden=1; path=/; sameSite=Lax; expires=" + ttl.toUTCString();
+
+    switch (e.target.name) {
+      case "csat-close-btn":
+        csatBanner.toggleAttribute("hidden", true);
+        csatBanner.removeEventListener("click", handleEvent);
```
groovecoder, line 528: question: could we add an event for `CSAT banner` `close` so we can keep track of how many `close` events we're getting compared to `submit` events?

template-helpers/hbs-helpers.js
```diff
@@ -54,6 +54,22 @@ function recruitmentBanner(args) {
...
 }
 
+function showCsatBanner(args) {
+  const signupDate = args.data.root.req.session.user?.created_at;
+
+  if (!signupDate) return; // don't show if user is not logged in or not signed up
+
+  if (args.data.root.req.cookies.csatHidden) return; // don't show if user closed banner
+
+  if (Date.now() - Date.parse(signupDate) < 604800000) return; // don't show if sign-up is less than 7 days old
```
groovecoder, line 64: nit: I find `7 * 24 * 60 * 60 * 1000` a bit more obvious for timestamp delta values.

views/partials/header/header.hbs
```diff
@@ -1,7 +1,8 @@
     {{> vpn-banner }}
     <header id="header" class="{{ addUpgradeClasses }}">
       {{{ recruitmentBanner }}}
-      {{{ microsurveyBanner }}}
+      {{!-- {{{ microsurveyBanner }}} --}}
```
groovecoder, line 4: IIRC we're replacing ALL the micro surveys with this single CSAT? We can probably just remove this line and the corresponding microsurvey logic completely?

groovecoder (approved):
post-merge approval with just some comments.

public/js/monitor.js
toufali: Yeah... I was saving it in order to follow the crumbs and delete all server-side portions of it post-data-removal pilot... I _think_ it might be tied in to a reused route in the pilot, and didn't want to deal with merge conflicts.  I could certainly add a "TODO"... 😬

public/js/monitor.js
toufali: This was added but then removed to avoid complicating the result set.  It's not a good measure of how many people dismiss the survey, because they can just scroll past.  Also, a naive approach would have tracked closing the banner _after_ form submission, as well as before – making the metric even less useful.

template-helpers/hbs-helpers.js
toufali: True.  I threw it in the comment, to be even _more_ obvious/annoying 😅 Don't have much of a preference either way

views/partials/header/header.hbs
toufali: yes.  Since microsurvey has nothing to do with the current pilot, this would be a good opportunity to de-clutter.  I'll add in a separate PR (possibly along with the signup code above)

PR #2526, opened 2022-04-06
remove micro-survey banner

toufali:


groovecoder (approved):
Looks good. Question: should we also remove the `micro-survey-*` messages from `locales/en/app.ftl` so new translators don't needlessly translate the response strings? I forget what the current best-practice is for removing un-used strings. @flodolo ?

toufali:
Yep -- I was going to ping @flodolo for a best practices there.  Was planning on removing more obsolete code first, but good to know as I clean house 🧹

flodolo:
Always remove unused strings.

In order to do so, simply remove them from the reference file (in this case `en/app.ftl`. Pontoon will remove them from localized files as soon as there's another update (new or changed translation).

toufali:
In addition to the micro-survey code, I wanted to delete unused signup/waitlist code.  Upon speaking with @heaversm we identified that code is still tied up with the current pilot.  Will do another cleanup after pilot concludes.

PR #2551, opened 2022-05-05
Add linter for reference strings

flodolo:
Not sure if this should be done against localization, or another branch.

P.S. this also checks for an empty line at the end of the files

### DIFF
--- a/.github/linter_config.yml
+++ b/.github/linter_config.yml
@@ -1,0 +1,35 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+# ID checks
+# ID01: check that identifiers only use lowercase and hyphens
+# ID02: check that identifiers have a minimum length
+#
+# Syntax checks
+#
+# Specify if syntax features are disabled. If the config is missing, it
+# will considered as enabled.
+#
+# Example:
+# SY01:
+#     disabled: true
+#
+# SY01: terms
+# SY02: message references
+# SY03: term references
+# SY04: variants
+# SY05: attributes
+
+---
+ID01:
+    enabled: true
+    exclusions:
+        messages: []
+        files: []
+ID02:
+    enabled: true
+    min_length: 6
+    exclusions:
+        messages: []
+        files: []
--- a/.github/requirements.txt
+++ b/.github/requirements.txt
@@ -1,0 +1,1 @@
+moz-fluent-linter==0.2.*
--- a/.github/workflows/reference_linter.yaml
+++ b/.github/workflows/reference_linter.yaml
@@ -1,0 +1,26 @@
+name: Lint Reference Files
+on:
+  push:
+    paths: ['en/**.ftl' ]
+    branches: [ main, main-kanary, localization ]
+  pull_request:
+    paths: ['en/**.ftl' ]
+    branches: [ main, main-kanary, localization ]
+  workflow_dispatch:
+jobs:
+  linter:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Clone repository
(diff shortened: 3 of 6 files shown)


flodolo:
This has a tone of noise because of trailing whitespaces, the actual changes are only around group comments :-( 

I think it's worth fixing that once and for all, but let me know if you want me to limit the changes.

toufali:
This is great, thanks!  I think we should merge into `main`.  
Questions:
- we already have l10n linting set up on CircleCI -- is that obsolete and removable?
https://github.com/mozilla/blurts-server/blob/bd63d53b948f9efa1f78de178c6b4284d80168c7/.circleci/config.yml#L43-L52
- I assume this is using Github actions? (I'm not yet familiar)  We may be migrating our testing from CircleCI to GHA in the next couple weeks tagging @mozrokafor for visibility.

flodolo:
> This is great, thanks! I think we should merge into `main`. Questions:
> 
>     * we already have l10n linting set up on CircleCI -- is that obsolete and removable?
>       https://github.com/mozilla/blurts-server/blob/bd63d53b948f9efa1f78de178c6b4284d80168c7/.circleci/config.yml#L43-L52

This runs a different type of checks on all locales locales (via compare-locales). It would basically flag syntax errors.

The code in this PR makes sure there are no issue with the content of the English file, and they're controls not available in compare-locales (e.g. typography).

>     * I assume this is using Github actions? (I'm not yet familiar)  We may be migrating our testing from CircleCI to GHA in the next couple weeks tagging @mozrokafor for visibility.

Correct. I saw that the kanary branch already uses GitHub action.

I'm not sure creating a PR against `main` is going to make things easier, because the FTL files diverge, and then this needs to be merged back into `localization`.

flodolo:
> I'm not sure creating a PR against `main` is going to make things easier, because the FTL files diverge, and then this needs to be merged back into `localization`.

Correction: the en-US file is pretty similar, but there are a ton of other changes because of localization commits.

toufali:
>I'm not sure creating a PR against main is going to make things easier, because the FTL files diverge, and then this needs to be merged back into localization.

@flodolo sorry, not following yet – As I understand there are currently two paths to `localization`:

1. a direct pull request to initiate translations, or 
2. merging `main` into `localization` for a sync of the latest build

Then, in order to get the latest translations, `localization` is merged into `main` periodically.  

It's possible that it doesn't matter which branch we merge this into, because the two will be sync'd eventually?

toufali (approved):

toufali:
@flodolo further to my prev comment, I _do_ agree that since this branch is based off of `localization` we should merge there as well.  Then, when we eventually merge `localization` into `main`, all the translation commits will show up as expected.

Also, FYI the broken integration test has been fixed with the latest `main` codebase.

flodolo:
> It's possible that it doesn't matter which branch we merge this into, because the two will be sync'd eventually?

Correct, the automation will be synced at some point when you merge the two branches.

> @flodolo further to my prev comment, I _do_ agree that since this branch is based off of `localization` we should merge there as well. Then, when we eventually merge `localization` into `main`, all the translation commits will show up as expected.

So, should we merge this now to localization, and then it will be merged to `main` later, or should I prepare a different PR based off `main` with my changes? The former seems easier.

flodolo:
> So, should we merge this now to localization, and then it will be merged to `main` later, or should I prepare a different PR based off `main` with my changes? The former seems easier.

In case, I have also another branch ready if we prefer to land against main
https://github.com/mozilla/blurts-server/compare/main...linter_main?expand=1

One other reason I picked `localization` is that, at least recently, we created PRs agains `localization`, and `main` is not the default branch (`main-kanary` is, and I assume it's the one currently deployed).

toufali:
agreed -- merging to `localization` first seems more sane.  We just deployed a new tag, including a localization sync, so now's a great time -- I'll push the button.

Thanks for this PR, will report back on its success!

PR #2555, opened 2022-05-10
Mntor 682/ad unit

toufali:
TODO:
- a better code formatter esp for CSS


toufali:
@groovecoder @pdehaan This should be a fairly safe, mostly front-end feature.  You might want to have a peek at `app-constants.js` and see if the synchronous function makes you cringe.  It's there to dynamically provide the total number of ad-units if/when we add or remove any.  I think it's fine because it's only read once on server start.

We're looking to push live tomorrow, so merging now to get the ball rolling.

PR #2558, opened 2022-05-11
Update ESLint to use "Standard" style guide

toufali:


toufali:
We should probably wait for the data-removal pilot to conclude before merging this.

redracer9:
Yes I think so
Thank you 



Sent from App for Gmail


Tuesday, May 10, 2022, 22:28 -0700 from ***@***.***  ***@***.***>:

PR #2614, opened 2022-08-17
MNTOR-697 - Add en strings for monthly unresolved breach email

toufali:


locales/en/email-strings.ftl
```diff
@@ -129,3 +129,26 @@ more-products-vpn = Protection for your whole device, on every device.
 more-products-cta-vpn = Get {-product-name-vpn}
 more-products-relay = Hide your real email address to help protect your identity.
 more-products-cta-relay = Get {-product-name-relay}
+
+## 2022 email template. HTML tags should not be translated, e.g. `<a>`
+
+email-2022-unsubscribe = You’re receiving this automated email as a subscriber of Firefox Monitor. <br>Feel free to change your email preferences at any time <a href="{ $unsubscribe-url }">here</a>.
```
flodolo, line 135:
```suggestion
email-2022-unsubscribe = You’re receiving this automated email as a subscriber of { -product-name }. <br>Feel free to change your email preferences at any time <a href="{ $unsubscribe-url }">here</a>.
```

flodolo: Do we really need to expose attributes (`href`, `rel`)? Previous strings didn't ([example](https://github.com/mozilla/blurts-server/blob/f8ceec0e35aa999ee81a4ac255035f9ca09e928c/locales/en/email-strings.ftl#L98))

locales/en/email-strings.ftl
```diff
@@ -129,3 +129,26 @@ more-products-vpn = Protection for your whole device, on every device.
...
+# Have I Been Pwned attribution
+email-2022-hibp-attribution = Breach data provided by <a href="{ $hibp-url }" rel="noopener">{-brand-HIBP}</a>
+
+## Monthly email for unresolved breaches. HTML tags should not be translated, e.g. `<br>`
+
+email-unresolved-heading = You have unresolved breaches
+email-unresolved-subhead = Your email has been exposed. <br>Fix it right away with {-product-name}.
```
flodolo, line 142:
```suggestion
email-unresolved-subhead = Your email has been exposed. <br>Fix it right away with { -product-name }.
```

locales/en/email-strings.ftl
```diff
@@ -129,3 +129,26 @@ more-products-vpn = Protection for your whole device, on every device.
...
+email-is-affected = Your email, { $email-address }, is affected by at least one data breach
+email-more-detail = Sign in to {-product-name} now to see more details about your breaches (including when they occured and what data was exposed), and learn what you should do when your email's been exposed in a data breach.
```
flodolo, line 144:
```suggestion
email-more-detail = Sign in to { -product-name } now to see more details about your breaches (including when they occurred and what data was exposed), and learn what you should do when your email’s been exposed in a data breach. 
```

locales/en/email-strings.ftl
toufali: > Do we really need to expose attributes (`href`, `rel`)? Previous strings didn't ([example](https://github.com/mozilla/blurts-server/blob/f8ceec0e35aa999ee81a4ac255035f9ca09e928c/locales/en/email-strings.ftl#L98))

@flodolo true, but the resulting HTML replacement function in JS didn't look so great to me:
```
  // replace placeholder anchor tag markup in first tout to make link
  // add UTM params which are passed to FxA for account creation
  const fxaTout1 = fxaTouts[0].paragraph
  const url = getPreFxaUtmParams(serverUrl, 'create-account-link', userEmail)
  if ((/<a>/).test(fxaTout1) && (/<\/a>/).test(fxaTout1)) {
    const openingAnchorTag = `<a class="pre-fxa-nested-link" href="${url}" style="color: #0060df; font-family: sans-serif; font-weight: 300; font-size: 15px; text-decoration: none;">`
    fxaTouts[0].paragraph = fxaTout1.replace('<a>', openingAnchorTag)
  }
```

To add similar code for each instance would not be ideal.  I could look into an alternate solution if it's a concern though.  I agree, putting the `rel` in the string made me wince.

flodolo: It's a concern because some locales will translate the variable, others will translate attributes. 

While we have checks, fixing them one by one scales poorly (150 locales x 30 projects), so we should try to avoid any potential cause of breakage.

Indeed, that piece of code seems overly complex. Maybe it's possible to have a simpler, generic helper that only takes URL and extra attributes?

locales/en/email-strings.ftl
```diff
@@ -129,3 +129,26 @@ more-products-vpn = Protection for your whole device, on every device.
 more-products-cta-vpn = Get {-product-name-vpn}
 more-products-relay = Hide your real email address to help protect your identity.
 more-products-cta-relay = Get {-product-name-relay}
+
+## 2022 email template. HTML tags should not be translated, e.g. `<a>`
+
+email-2022-unsubscribe = You’re receiving this automated email as a subscriber of { -product-name }. <br>Feel free to change your email preferences at any time <a href="{ $unsubscribe-url }">here</a>.
+# Have I Been Pwned attribution
+email-2022-hibp-attribution = Breach data provided by <a href="{ $hibp-url }" rel="noopener">{-brand-HIBP}</a>
```
flodolo, line 137:
```suggestion
email-2022-hibp-attribution = Breach data provided by <a href="{ $hibp-url }" rel="noopener">{ -brand-HIBP }</a>
```

Missed this one. It's not broken, but it would be nice to write the syntax consistently (leading and trailing space within {}).

flodolo (commented):
One nit but the strings look OK, pending agreement on the links.

locales/en/email-strings.ftl
toufali: @flodolo – updated strings to remove attributes from anchor tags 👍

flodolo (approved):

flodolo:
@toufali 
Could you provide a timeline on when the email will be available here for QA? https://fx-breach-alerts.herokuapp.com/email-l10n/

toufali:
@flodolo it should be posted in a few hours.  I'll send you the review link.

PR #2616, opened 2022-08-19
Deploy from CircleCI to Heroku

jwhitlock:
This is similar to https://github.com/mozilla/fx-private-relay/pull/2355, except I'm using the [deploy-from-git job](https://circleci.com/developer/orbs/orb/circleci/heroku#jobs-deploy-via-git) rather than the command, and there are two jobs, one for `main` and one for `localization`. I've added the environment variables to make these jobs work.

The [CircleCI heroku orb](https://circleci.com/developer/orbs/orb/circleci/heroku) does not support review apps. It looks possible using the [REST API](https://devcenter.heroku.com/articles/platform-api-reference#review-app), but not easy. The website still recommends the GitHub integration for reviews, and I don't think we're ready to enable that integration yet.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -20,6 +20,7 @@ executors:
 orbs:
     node: circleci/node@5.0.2
     python: circleci/python@1.4.0
+    heroku: circleci/heroku@1.2.6
 
 jobs:
     lint-js:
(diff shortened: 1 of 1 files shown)


Procfile
```diff
@@ -0,0 +1,2 @@
+release: npm run db:migrate
+web: npm start
```
toufali, line 2: Interesting – I had configured Heroku to run our app without a procfile and deleted it a little while ago.  

How/when is `release` triggered?  It _would_ be nice to have Heroku run migrations automatically... I wonder if there's a way to configure this _without_ the Heroku-only procfile?  Not opposed to reinstating it, just curious.

toufali (approved):

Procfile
jwhitlock: I also wanted to avoid `Procfile`, but I could not find documentation or a hook in https://github.com/heroku/heroku-buildpack-nodejs for running a command on release any other way 😢

PR #2618, opened 2022-08-22
Mntor 697/monthly email

toufali:
Monthly notification to users of pending/unresolved breaches.  Much of this PR is in support of QA testing, including mockup/preview page and ability to trigger email send behind auth.  

Assuming it's agreed, I will remove the preview page and email trigger functionality before going live.  Under separate ticket, I can incorporate this new email into the l10n page used for other email previews.

**Acceptance Criteria**
- trigger an email targeted towards all users with atleast 1 unresolved breach.
- send emails only to existing monitor users.
- localize emails to all monitor supported locales.
- email must include link to mozilla privacy policy and terms of service.
- have CTA in email to launch monitor in browser.
- email should also render correctly on mobile and on devices with dark theme.
- _TODO: Analytics_


flodolo:
Not exactly sure why I'm flagged as reviewer, this is really out of my area ;-)

middleware.js
```diff
@@ -165,6 +165,27 @@ async function requireSessionUser (req, res, next) {
...
+    return res.redirect(`/oauth/init?${queryParams}`)
+  }
+  const fxaProfileData = await FXA.getProfileData(user.fxa_access_token)
+  const admins = AppConstants.ADMINS?.split(',')
+  const isAdmin = admins.includes(JSON.parse(fxaProfileData).email)
+
+  if (!isAdmin || (fxaProfileData.hasOwnProperty('name') && fxaProfileData.name === 'HTTPError')) {
+    delete req.session.user
+    return res.sendStatus(401)
+  }
```
jwhitlock, line 181: I think if they are not an admin user, we should return a 401 but leave the user logged in. We can also address the [`no-prototype-builtins` eslint warning](https://eslint.org/docs/latest/rules/no-prototype-builtins)
```suggestion
  const hasFxaError = Object.prototype.hasOwnProperty.call(fxaProfileData, "name") && fxaProfileData.name
  if (hasFxaError) {
    delete req.session.user
  }
  if (!isAdmin || hasFxaError) {
    return res.sendStatus(401)
  }
```

views/layouts/email-2022.hbs
```diff
@@ -68,7 +82,8 @@
     <tr class="footer">
       <td style="border-top: 1px solid #ddd; padding: 24px 0">
         <p>{{{ getString "email-2022-unsubscribe" unsubscribe-link-attr=( concat "href=" this.SERVER_URL ) }}}</p>
-        <p>{{{ getString "email-2022-hibp-attribution" hibp-link-attr='href="https://haveibeenpwned.com/" rel="noopener"' }}}
+        <p>{{{ getString "email-2022-hibp-attribution" hibp-link-attr='href="https://haveibeenpwned.com/"
+          rel="noopener"' }}}
         </p>
         <img src="{{ SERVER_URL }}/img/email_images/mozilla-logo-bw-rgb.png" width="130px"
```
jwhitlock, line 88: Can you add an `alt="Mozilla"` to the logo image `mozilla-logo-bw-rgb.png` as well?

app-constants.js
```diff
@@ -55,7 +55,9 @@ const optionalEnvVars = [
   'GEOIP_GEOLITE2_COUNTRY_FILENAME',
   'VPN_PROMO_BLOCKED_LOCALES',
   'EDUCATION_VIDEO_URL_RELAY',
-  'EDUCATION_VIDEO_URL_VPN'
+  'EDUCATION_VIDEO_URL_VPN',
+  'ADMINS',
+  'MONTHLY_CRON_ENABLED'
```
jwhitlock, line 60: Please add placeholders to `.env-dist`

jwhitlock (approved):
Thanks @toufali! This worked well locally and in Heroku. I've got a few suggestions for this PR or the next. Feel free to re-request a review if you want me to check again.

app-constants.js
groovecoder: 👍 
```
CRITICAL fx-monitor.uncaughtException: TypeError: Cannot read properties of undefined (reading 'includes')
    at requireAdminUser (/Users/lcrouch/code/mozilla/blurts-server/middleware.js:176:26)
```

When `ADMINS` is left empty.

db/DB.js
```diff
@@ -419,6 +419,30 @@ const DB = {
...
+      })
+  },
+
+  async getBreachStats (id) {
+    const { breach_stats: breachStats } = await knex('subscribers')
+      .where({ id })
+      .first('breach_stats')
+
+    return breachStats
+  },
```
groovecoder, line 436: suggestion (non-blocking): we shouldn't need this DB query. In both of the places this is called, the same data is available at `req.user.breach_stats`. (`requireSessionUser` and `requireAdminUser` both set `req.user` to the result of `DB.getSubscriberById`)

controllers/email-l10n.js
```diff
@@ -108,7 +111,54 @@ function notFound (req, res) {
   })
 }
 
+async function previewEmail2022 (req, res) {
+  const breachStats = await DB.getBreachStats(req.user.id)
```
groovecoder, line 115:
```suggestion
  const breachStats = req.user.breach_stats
```

controllers/email-l10n.js
```diff
@@ -108,7 +111,54 @@ function notFound (req, res) {
...
+    whichPartial: 'email_partials/email-monthly-unresolved',
+    csrfToken: req.csrfToken(),
+    primaryEmail: req.user.primary_email,
+    breachStats
+  })
+}
+
+function sendTestEmail (data) {
+  return async function (req, res) {
+    const breachStats = await DB.getBreachStats(req.user.id)
```
groovecoder, line 128:
```suggestion
    const breachStats = req.user.breach_stats
```

db/DB.js
```diff
@@ -419,6 +419,30 @@ const DB = {
...
+
+  async getSubscribersWithUnresolvedBreaches () {
+    const res = await knex('subscribers')
+      .select('primary_email', 'breach_stats', 'signup_language')
+      .whereJsonPath('breach_stats', '$.numBreaches.numUnresolved', '>', 0)
```
groovecoder, line 441: risk: I worry about how this `WHERE breach_stats ...` query will perform at the scale of thousands or millions of records. (E.g., the [4.64M `en` subscribers](https://earthangel-b40313e5.influxcloud.net/d/dEpkGp4Wz/fx-monitor?orgId=1&viewPanel=45)) There's no index on the field, and I'm not 100% sure how indexes work on `jsonb` fields.

If this query only returns subscribers who have non-null value in `breach_stats` - i.e., subscribers who have visited Monitor and thereby populated their `breach_stats` field, then it seems fine For Now™ so long as we can monitor its DB performance to make sure it's not getting too wild as more and more subscribers are included in it.

email-utils.js
```diff
@@ -50,7 +50,8 @@ const EmailUtils = {
     }
 
     const emailContext = Object.assign({
-      SERVER_URL: AppConstants.SERVER_URL
+      SERVER_URL: AppConstants.SERVER_URL,
+      layout: aTemplate
```
groovecoder, line 54: 😆 : heh, I forgot about the `aVar` naming convention we inherited from Firefox JS code. I don't even remember what the `a` stands for anymore.

scripts/send-email-to-unresolved-breach-subscribers.js
```diff
@@ -0,0 +1,23 @@
+const DB = require('../db/DB')
+const EmailUtils = require('../email-utils')
+const { LocaleUtils } = require('../locale-utils')
+
+async function sendUnresolvedBreachEmails () {
+  const subscribers = await DB.getSubscribersWithUnresolvedBreaches()
```
groovecoder, line 6: risk: If we're planning to run this via `node scripts/send-email-to-unresolved-breach-subscribers.js`, then when this result set gets really big, we will want a way to send these emails in batches. If it's all done in a single command, that command process is going to take such a long time, and use up so much memory, that it will be hard to tell if it's actually working, or if it got hung up.

change (blocking): We should also include a way to track (a new DB field like `subscribers.needs_unresolved_breach_email`?) whether or not a subscriber has already been sent the email. (Are we sending these on a regular basis?) When there are errors during this process, we do NOT want to send repeat emails to people who have already received one.

scripts/send-email-to-unresolved-breach-subscribers.js
```diff
@@ -0,0 +1,23 @@
...
+
+  subscribers.forEach(async subscriber => {
+    const supportedLocales = [subscriber.signup_language, 'en'].filter(Boolean) // filter potential nullish signup_language
```
groovecoder, line 9: question (non-blocking): so we're sending this to everyone who has `en` in their signup language at all. Even if it's their 2nd or 3rd language?

server.js
```diff
@@ -34,6 +34,8 @@ const EmailUtils = require('./email-utils')
 const HBSHelpers = require('./template-helpers/')
 const HIBP = require('./hibp')
 const IpLocationService = require('./ip-location-service')
+const { initMonthlyCron } = require('./cron')
+const { sendUnresolvedBreachEmails } = require('./scripts/send-email-to-unresolved-breach-subscribers')
```
groovecoder, line 38: change (blocking): Oh wow, so this will run every month and it will run in the app server process?! That means it will slow the Monitor site down while it's sending tons of emails.

We should work with SRE to move this to a "real" cron job that is de-coupled from the app server process.

server.js
```diff
@@ -189,7 +191,7 @@ const hbs = exphbs.create({
   extname: '.hbs',
   layoutsDir: path.join(__dirname, '/views/layouts'),
   defaultLayout: 'default',
-  partialsDir: path.join(__dirname, '/views/partials'),
+  partialsDir: [path.join(__dirname, '/views/layouts'), path.join(__dirname, '/views/partials')],
```
groovecoder, line 194: question (non-blocking): why do we need the layouts in the `partialsDir`? (We already have it in `layoutsDir`) Is there something in `layouts/` that should be moved to `partials/`?

groovecoder (changes requested):

email-utils.js
pdehaan: Wonder no more!

a=argument

---

Per https://firefox-source-docs.mozilla.org/code-quality/coding-style/coding_style_js.html#prefixes:

> #### Prefixes
> - k=constant (e.g. `kNC_child`). Not all code uses this style; some uses `ALL_CAPS` for constants.
> - g=global (e.g. `gPrefService`)
> - a=argument (e.g. `aCount`)
> - JavaScript Specific Prefixes
>     - _=member (variable or function) (e.g. `_length` or `_setType(aType)`)
>    - k=enumeration value (e.g. `const kDisplayModeNormal = 0`)
>    - on=event handler (e.g. `function onLoad()`)
>    - Convenience constants for interface names should be prefixed with `nsI`:

scripts/send-email-to-unresolved-breach-subscribers.js
toufali: The way this is intended to work is to use the subscriber's signup language first.  If for some reason that's null (mine is locally), we still have `en` as the fallback.

server.js
toufali: This is primarily for QA testing.  The QA mockup (which includes email send form) was the layout and the monthly email layout was the partial in this case, with a 2-level-deep partial (Inception).  We could revert this if you like when we remove the QA mockup, though it's possible we may need to use this functionality:

The current l10n mockups are not truly loading the actual email layouts and associated styles – the html and style are actually duplicated/copied over and in some cases vary quite a bit from the actual emails.  It might be better to DRY it up and test the actual email.

routes/email-l10n.js
```diff
@@ -1,9 +1,22 @@
...
 const router = express.Router()
+const cspUnsafeInline = {
+  directives: {
+    defaultSrc: "'self'",
+    styleSrc: "'unsafe-inline'"
+  }
+}
 
 router.get('/', getEmailMockUps)
+router.get('/email-2022-mockup', csrfProtection, requireAdminUser, helmet.contentSecurityPolicy(cspUnsafeInline), previewEmail2022)
```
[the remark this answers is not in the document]
toufali: I was actually going to delete it altogether with the other QA artifacts.  I suppose we could leave it in only for dev, but it complicates the repo.  Maybe I'll leave it in until I have a chance to redo the l10n mockups as discussed above...

db/DB.js
toufali: great call!

db/DB.js
toufali: If we're punting the cron job to SRE, we could have them monitor it as well... 😄

server.js
groovecoder: Changing this to non-blocking so long as give ourselves fast-follow work to move this to a cron job with SRE support.

scripts/send-email-to-unresolved-breach-subscribers.js
toufali: @groovecoder this has been updated, please review latest dif!
https://github.com/mozilla/blurts-server/blob/ca674258389ceca63dda0ad8095865fb7452a8ed/scripts/send-email-to-unresolved-breach-subscribers.js#L21
https://github.com/mozilla/blurts-server/blob/ca674258389ceca63dda0ad8095865fb7452a8ed/db/DB.js#L430-L442

controllers/email-l10n.js
```diff
@@ -108,7 +110,59 @@ function notFound (req, res) {
   })
 }
 
+function previewEmail2022 (req, res) {
```
groovecoder, line 113: FWIW, I did this to spot-check in the view:
```suggestion
function previewEmail2022 (req, res) {
  const unsubscribeUrl = `${AppConstants.SERVER_URL}/user/unsubscribe-monthly?token=${req.user.primary_verification_token}`
```

controllers/email-l10n.js
```diff
@@ -108,7 +110,59 @@ function notFound (req, res) {
...
+  res.render('layouts/email-2022-mockup', {
+    layout: 'email-2022-mockup',
+    whichPartial: 'email_partials/email-monthly-unresolved',
+    supportedLocales: req.supportedLocales,
+    csrfToken: req.csrfToken(),
+    primaryEmail: req.user.primary_email,
+    breachStats: req.user.breach_stats
```
groovecoder, line 120: And this ...
```suggestion
    breachStats: req.user.breach_stats,
    unsubscribeUrl
```

groovecoder (approved):
Code and spot-check looks good. Thanks for the quick turn-around to restore an unsubscribe link! Agree we can address the unsubscribe url template/view in a fast-follow.

db/migrations/20220826220021_add-monthly-email-column.js
```diff
@@ -0,0 +1,13 @@
+'use strict'
+
+exports.up = function (knex) {
+  return knex.schema.table('subscribers', (table) => {
+    table.timestamp('monthly_email_at')
```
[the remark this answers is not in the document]
toufali: @groovecoder No, default is empty (null).  In the DB lookup I compare `created_at` with `monthly_email_at` to make sure neither are within the 30 day window.

controllers/email-l10n.js
toufali: Ah, I see – I tested actual emails and neglected the preview.  I will add this to avoid confusion 👍

PR #2623, opened 2022-08-25
Allow CircleCI deploy fail without failing build

jwhitlock:
Allow the CircleCI deploy to fail on the `git push` step but still show the job passing. A developer will need to check the job output to see if the deploy was successful, or check Heroku.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -134,6 +134,29 @@ jobs:
                   name: Sync static content to S3
                   command: ./.circleci/scripts/sync-static
 
+    deploy_heroku:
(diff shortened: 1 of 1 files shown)


toufali (approved):

jwhitlock:
When this merged, the Heroku push failed but did not fail the job: https://app.circleci.com/pipelines/github/mozilla/blurts-server/5105/workflows/5e8b8af2-b587-4bec-9051-54972e614cbc/jobs/30472?invite=true#step-104-17

We'll have to wait until we reset the Heroku app to see if a fast-forward push still works

PR #2645, opened 2022-09-08
CircleCI: Add tests for PostgreSQL versions 9.6.24 and 13.7

jwhitlock:
Run unit tests against specific PostgreSQL versions, using current CircleCI instructions for a PostgreSQL service container.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -17,6 +17,25 @@ executors:
         machine:
             image: ubuntu-2004:202107-02
 
+    node_with_postgres:
+        parameters:
+            postgres_version:
+                description: "Which cimg/postgres tag to use"
+                type: string
+        docker:
+            - image: cimg/node:16.15
+              auth:
+                username: $DOCKER_USER
+                password: $DOCKER_PASS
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -17,6 +17,25 @@ executors:
...
+                type: string
+        docker:
+            - image: cimg/node:16.15
+              auth:
+                username: $DOCKER_USER
+                password: $DOCKER_PASS
+            - image: cimg/postgres:<< parameters.postgres_version >>
+              environment:
+                  POSTGRES_USER: postgres
+                  POSTGRES_PASSWORD: 93e389316eaf4ea2be4010d526cc1468
```
toufali, line 33: safe to have this exposed?

toufali (approved):

.circleci/config.yml
jwhitlock: I think so. It's a UUID4 that I generated today, and it only exists for the duration of the test. It would be easy to move to a project secret if you prefer.

PR #2646, opened 2022-09-08
MNTOR-854: Add tests for `getSubscribersWithUnresolvedBreaches`, fix query for psql 9.6

jwhitlock:
Add a test subscriber and email `has_breaches`, with two tests for `getSubscribersWithUnresolvedBreaches`:

* The one subscriber is returned by `getSubscribersWithUnresolvedBreaches`
* No subscribers are returned when the `breach_stats` value is `null`

Update `getSubscribersWithUnresolvedBreaches` so that it works on PostgreSQL 9.6 and later.

### DIFF
--- a/db/DB.js
+++ b/db/DB.js
@@ -445,7 +445,6 @@ const DB = {
       .whereRaw('monthly_email_optout IS NOT TRUE')
       .whereRaw("greatest(created_at, monthly_email_at) < (now() - interval '30 days')")
       // .whereJsonPath('breach_stats', '$.numBreaches.numUnresolved', '>', 0)  // Requires psql 11
-      .whereNotNull('breach_stats')
       .whereRaw("(breach_stats #>> '{numBreaches, numUnresolved}')::int > 0")
 
     return res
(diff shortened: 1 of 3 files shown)


db/DB.js
```diff
@@ -444,7 +444,9 @@ const DB = {
       .select('primary_email', 'primary_verification_token', 'breach_stats', 'signup_language')
       .whereRaw('monthly_email_optout IS NOT TRUE')
       .whereRaw("greatest(created_at, monthly_email_at) < (now() - interval '30 days')")
-      .whereJsonPath('breach_stats', '$.numBreaches.numUnresolved', '>', 0)
+      // .whereJsonPath('breach_stats', '$.numBreaches.numUnresolved', '>', 0)  // Requires psql 11
+      .whereNotNull('breach_stats')
```
toufali, line 448: perhaps if the following line works without it, we can remove the null check in the next iteration?

toufali (approved):

db/DB.js
jwhitlock: Yep, that works!

PR #2653, opened 2022-09-14
[WIP] docs and cleanups

mansaj:
WIP:


jwhitlock:
The unit tests are passing, but failing when uploading to coveralls, because the `COVERALLS_REPO_TOKEN` is not available in forks. I would not consider these job failures a blocker for reviewing or merging.

I re-discovered this issue on my first PR (https://github.com/mozilla/blurts-server/pull/2612), but got repo write access and started pushing branches to the `mozilla/blurts-server` instead of my fork.

I'd like to skip attempting to upload coverage when it will fail. There's this note, but no existing issue (yet):

https://github.com/mozilla/blurts-server/blob/306a63430e61fa298b6bc1851736089709afb24d/README.md?plain=1#L157

PR #2671, opened 2022-10-10
Delete duplicate JobAndTalent logo

jwhitlock:
We have two copies of this logo:

* https://github.com/mozilla/blurts-server/blob/main/public/img/logos/Jobandtalent.png
* https://github.com/mozilla/blurts-server/blob/main/public/img/logos/JobAndTalent.png

The one on this page is `JobAndTalent.png`, so I'm deleting `Jobandtalent.png`:

https://monitor.firefox.com/breach-details/JobAndTalent

Additionally, the default macOS file system is case insensitive, so both files are not allowed at the same time.


pdehaan (approved):

pdehaan:
Good catch. 👍 
I used to have a script for catching unused logos. I'll have to see if I can find it and see if it would have caught this case.

PR #2674, opened 2022-10-13
Mntor 792/main template

toufali:
Initial effort to introduce Monitor V2 into old codebase, allowing us to switch versions using the MONITOR_V2 env var.  When MONITOR_V2 is 'true', `npm start` and `npm run dev` run in the V2 workspace, starting the V2 server.  Otherwise, the existing V1 server runs.

- V2 source resides in new `/src` dir
- V2 client files build into git-ignored `/dist` dir, from `/src/client`
- Handlebars replaced by native ES Template Literals
- RequireJS replaced by ES Modules
- Fluent functions refactored and updated.
- NPM Workspaces utilized to facilitate version switch

TODO: when V1 is retired, refactor/combine package.json files and remove need for NPM workspace
NOTE: the Squoosh library triggers noisy warnings: `MaxListenersExceededWarning`.  Monitoring [the issue](https://github.com/GoogleChromeLabs/squoosh/issues/1206)


package.json
```diff
@@ -77,8 +78,8 @@
     "url": "git+https://github.com/mozilla/blurts-server.git"
   },
   "scripts": {
-    "start": "npm run build & node server.js",
-    "dev": "npm run build & nodemon server.js",
+    "start": "source .env && if [[ $MONITOR_V2 == 'true' ]]; then npm start --workspace=src; else npm run build & node server.js; fi;",
```
mansaj, line 81: For these two commands, an option is to have two separate .env files (one for the old system, living in root, and another for the new version, in /src)

In that case, we will `source .env` separately based on $MONITOR_V2

toufali: A bit of chicken and the egg there – this code is sourcing .env in order to get `MONITOR_V2` in the first place 😅

mansaj: That's true, nvm

src/app.js
```diff
@@ -0,0 +1,35 @@
+import express from 'express'
+import accepts from 'accepts'
+
+import AppConstants from '../app-constants.js'
+import { initFluentBundles, updateAppLocale } from './utils/fluent.js'
+import indexRouter from './routes/index.js'
+
+const app = express()
+const port = AppConstants.NODE_ENV === 'dev' ? 3333 : null
```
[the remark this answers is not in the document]
toufali: The port is only set by us for dev.  Heroku and SRE use their own ports.  But yes, I can't think of a good reason for dev to have different v1/v2 ports – I'll update to keep same 👍

src/app.js
```diff
@@ -0,0 +1,35 @@
+import express from 'express'
+import accepts from 'accepts'
+
+import AppConstants from '../app-constants.js'
```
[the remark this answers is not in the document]
toufali: Yes, I think you're right – good call!

mansaj (approved):

package.json
```diff
@@ -77,8 +78,8 @@
...
+    "dev": "source .env && if [[ $MONITOR_V2 == 'true' ]]; then npm run dev --workspace=src; else nodemon server.js; fi;",
```
jwhitlock, line 82: This works OK in local development, and will work in Docker because `.env-dist` is copied to `.env` in the image. I think it will fail in Heroku\, where `.env` doesn't exist by default. I tried this:

```sh
> mv .env .env.back
> npm start

> blurts-server@1.0.0 start
> source .env && if [[ $MONITOR_V2 == 'true' ]]; then npm start --workspace=src; else npm run build & node server.js; fi;

/var/folders/_y/72__kkbs2tqd5ch4gwrgnwtr0000gn/T/start-204f5703.sh: line 1: .env: No such file or directory
```

On way would be to test existence first:

```suggestion
    "start": "if [[ -f .env ]]; then source .env; fi && if [[ $MONITOR_V2 == 'true' ]]; then npm start --workspace=src; else npm run build & node server.js; fi;",
    "dev": "if [[ -f .env ]]; then source .env; fi && if [[ $MONITOR_V2 == 'true' ]]; then npm run dev --workspace=src; else nodemon server.js; fi;",
```

Another way may be explicit commands:

```suggestion
    "start:v1": "npm run build & node server.js",
    "dev:v1": "npm run build & nodemon server.js",
    "start:v2": "npm start --workspace=src",
    "dev:v2": "npm run dev --workspace=src",
    "start": "if [[ $MONITOR_V2 == 'true' ]]; then npm run start:v2; else npm run start:v1; fi;",
    "dev": "if [[ $MONITOR_V2 == 'true' ]]; then npm run dev:v2; else npm run dev:v1; fi;",
```

Then v2 could be selected by environment variable `MONITOR_V2` or explictly with `npm run start:v2`.

src/views/layouts/main.js
```diff
@@ -0,0 +1,34 @@
...
+      </header>
+      <nav>
+      </nav>
+      <main>
+        ${data.partial}
+      </main>
+      <footer>
+        <a href='https://www.mozilla.org' target='_blank'><img src='images/moz-logo-1color-white-rgb-01.svg' width='100'></a>
+        <menu>
+          <li><a href='https://www.mozilla.org/privacy/firefox-monitor' target='_blank'>Terms & Privacy</a></li>
```
jwhitlock, line 28: Should `Terms & Privacy` be a translated string? Should it be `Terms &amp; Privacy?`

src/views/layouts/main.js
```diff
@@ -0,0 +1,34 @@
...
+      <link rel="preconnect" href="https://fonts.googleapis.com">
+      <link rel="preconnect" href="https://fonts.gstatic.com" crossorigin>
+      <link href="https://fonts.googleapis.com/css2?family=Inter&display=swap" rel="stylesheet"> 
+      <link href='css/index.css' type='text/css' rel='stylesheet'>
+      <script src='js/index.js' type='module'></script>
+    </head>
+    <body>
+      <header>
+        <a href='/'><img src='images/monitor-logo-transparent.png' width='220' height='40'></a>
+        <a href='/user/login' class='button'>Sign Up</a>
```
jwhitlock, line 18: Should `Sign Up` be a translated string?

src/views/partials/dashboard.js
```diff
@@ -0,0 +1,3 @@
+export default data => `
+<h1>This is the breach dashboard page</h1>
```
jwhitlock, line 2: The dashboard title is hidden behind the Logo / Sign Up header for me. Should the body have some padding at the top to make room for the header?

jwhitlock (approved):
Thanks @toufali, this looks good as a framework for the next version. I've identified some nits to fix this PR or a future one. My biggest concern is if it will deploy cleanly to Heroku, but we may want to delay that work until a PR that focuses on a Heroku deployment that runs v2 all the time.

src/views/partials/dashboard.js
toufali: Ah yes, this was a quick POC by Joe that will be updated in a future PR 👍

package.json
toufali: Thanks for flagging this!  

Testing existence in the first example seems verbose.  However, the second example could have confusing behavior locally for `npm start`,  i.e. a dev expects `npm start` to "just work", but they will always get v2 even if they set `$MONITOR_V2`.

Whatever we decide, we can probably just limit it to `npm start` since that's the command used by other environments, vs `npm run dev` which is local-only?

Agreed with getting Heroku fixed in a separate PR – in fact I had already created a Jira ticket this morning to get that going 👍

PR #2686, opened 2022-10-31
Mntor 962

mansaj:
Logout part 1:


src/controllers/auth.js
```diff
@@ -109,9 +109,24 @@ async function confirmed (req, res, next, client = FxAOAuthClient) {
...
-  res.redirect('/')
+/**
+ * Controller to trigger a logout for user
+ * @param {object} req contains session.user
+ * @param {object} res redirects to homepage
+ */
+async function logout (req, res) {
+  log.debug('logout', req.session)
+  const subscriber = req.session?.user
+  log.debug('logout', subscriber?.primary_email)
```
toufali, line 120: Do we want to remove or consolidate the debugs?  Are they silent in prod?

src/db/index.js
```diff
@@ -274,6 +274,34 @@ async function updateFxAData (subscriber, fxaAccessToken, fxaRefreshToken, fxaPr
...
+    })
+    .returning('*')
+  const updatedSubscriber = Array.isArray(updated) ? updated[0] : null
+  if (updatedSubscriber && subscriber.fxa_refresh_token) {
+    log.debug('removeFxAData.destroy-refresh-token', subscriber.fxa_refresh_token)
+    log.debug('removeFxAData.destroy-refresh-token', await destroyOAuthToken({ refresh_token: subscriber.fxa_refresh_token }))
+  }
+  if (updatedSubscriber && subscriber.fxa_access_token) {
+    log.debug('removeFxAData.destroy-access-token', subscriber.fxa_access_token)
+    log.debug('removeFxAData.destroy-access-token', await destroyOAuthToken({ token: subscriber.fxa_access_token }))
```
toufali, line 300: Is all the debug code still needed?  Should we break out the `destroyOAuthToken` functions so that they don't live as part of a debug call?

toufali (approved):
LGTM!  Only comments around debug code 👍

src/controllers/auth.js
mansaj: Yes the debugs are silent in prod. I will actually change the email one to info

PR #2687, opened 2022-11-01
Mntor 801: 

mansaj:


src/controllers/breaches.js
```diff
@@ -1,13 +1,319 @@
...
+ * Modify breach resolution for a user
+ * @param {object} req containing {user, body: {affectedEmail, recencyIndex, resolutionsChecked, isUnresolve}}
+ *
+ * recencyIndex: corresponds to the relevant breach from HIBP
+ *
+ * resolutionsChecked: has the following structure [DataTypes]
+ *
+ * @param {object} res JSON object containing the updated breach resolution
+ */
+async function putBreachResolution (req, res) {
```
mansaj, line 43: @toufali Ignore this function, I will create another PR addressing the PUT call. I want to get the review started since it's already a pretty huge PR

src/app.js
```diff
@@ -6,7 +6,7 @@ import accepts from 'accepts'
 import redis from 'redis'
 
 import AppConstants from './app-constants.js'
-import { initFluentBundles, updateAppLocale } from './utils/fluent.js'
+import { initFluentBundles, updateAppLocale, loadBreachesIntoApp } from './utils/index.js'
```
toufali, line 9: I think this answers my comment below re `/utils/index.js`... 

I miss the context I get from reading the actual module where each function comes from.  Also by linking directly, I can find the related file faster – using the index with wildcard loses the paper-trail for me, or has me open the index file only to find a dead-end? 😬

src/app.js
```diff
@@ -53,10 +53,20 @@ app.use(session({
...
 }))
 
+// Load breaches into namespaced cache
+await (async () => {
+  try {
+    await loadBreachesIntoApp(app)
+  } catch (error) {
+    console.error('Error loading breaches into app.locals', error)
+  }
+})()
```
toufali, line 63: I believe top-level `await` is available to us in this Node version, so you should be able to make this more readable?  e.g.
```
try {
  await loadBreachesIntoApp(app)
} catch (error) {
  console.error('Error loading breaches into app.locals', error)
}
```

toufali (commented):
Nice job wrangling all this logic together!  I appreciated the JSDoc annotations.  The controllers/breaches.js page still seems pretty complicated... I assume this is mostly copy/paste in effort to get something out – maybe we can simplify next year!

src/app.js
mansaj: Thanks for this catch!

src/db/migrations/20221026215921_add-breach-resolution-column.cjs
```diff
@@ -0,0 +1,13 @@
...
+  return knex.schema.table('subscribers', (table) => {
+    table.jsonb('breach_resolution').defaultTo(null)
+  })
+}
+
+exports.down = function (knex) {
+  return knex.schema.table('subscribers', (table) => {
+    table.dropColumn('breach_resolution')
+  })
+}
```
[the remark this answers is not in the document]
mansaj: I've actually reverted the migration folder back to the /db directory (package.json script `db:migration`), so we won't be using `/src/` for that. I've deleted all the duplicated migration `.js` files.. I tried using a different directory for the migration script but I think the CICD would need to be updated with a new `.env` file and that's way beyond the scope of this story. It'll be tracked as a tech debt for Q1 & beyond

src/db/migrations/20221026215921_add-breach-resolution-column.cjs
```diff
@@ -0,0 +1,13 @@
+'use strict'
```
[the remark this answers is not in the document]
mansaj: deleted, see comment above

src/controllers/breaches.js
```diff
@@ -1,13 +1,310 @@
...
+    for (const newBreachForEmail of newBreachesForEmail) {
+      newBreachForEmail.NewBreach = true // add "NewBreach" property to the new breach.
+      emailEntry.hasNewBreaches = newBreachesForEmail.length // add the number of new breaches to the email
+    }
+  }
+
+  return { verifiedEmails, unverifiedEmails }
+}
+
+function addRecencyIndex (foundBreaches) {
```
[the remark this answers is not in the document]
mansaj: So this is a tricky one. I also thought we could just delete it and sort by date BUT then I realized why it's unfortunately necessary.

The way this function was written before was less than conventional... So we would first assign these indices then filter at the end of the function. What that means is that the data can end up looking like:
```
[{
  recency_1: {}
}, {
  recency_3: {}
}, {
  recency_28: {}
}
  ...
]
```

Basically there will be "skips" after the filtering. Let's say there are 4 items in the array after filtering, the indices won't just be `0-3` if that makes sense.

To keep it backwards compatible, we have to follow the same logic for v2
@toufali if this is confusing, I can get on a call with you to explain more with an example

src/app.js
mansaj: The thing I like about having consolidated index imports is that we don't end up with extra lines per each file.
In terms of tracking files, more IDEs (VS code) usually has a way to get there easily, for example, by doing `cmd+click` while hovering over the function you are interested in tracing.

Does that work?

mansaj:
> Nice job wrangling all this logic together! I appreciated the JSDoc annotations. The controllers/breaches.js page still seems pretty complicated... I assume this is mostly copy/paste in effort to get something out – maybe we can simplify next year!

It is definitely a mess.. Since we have to maintain backwards compatibility, we have to keep most of the current logic intact (otherwise our production data will become corrupt and unreadable). The only time we can really start refactoring is when we deprecate the older version of this app.

src/controllers/breaches.js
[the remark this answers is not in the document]
toufali: I'd love to hear more about it!  Just hoping we have a way out of poor code after backwards compatibility is no longer needed.

src/app.js
toufali: Great tip about `cmd+click`!  In this case I had a hard time finding the file/function in GitHub.  I don't want to mess with your flow – I'm sure it's fine.  

Final thought though – doesn't the wildcard index you're using auto-export everything?  Even functions that are not used in the codebase?  I'm concerned about the eventual cruft that will get loaded even though it's not used.  In particular for front-end this would be unwanted behavior for tree-shaking, etc?

mansaj: I'm not sure I'm interpreting this correctly, but wouldn't that depend on what you import instead of export? 
If you are concerned about private functions being exported by mistake, the exports are bounded in the individual files and the wildcard in `index.js` cannot export more than what's allowed by the underlying files.

toufali (approved):

PR #2688, opened 2022-11-06
Mntor 797/landing page

toufali:


locales/en/app.ftl
```diff
@@ -3,11 +3,14 @@
 ## The following messages are brand and should be kept entirely in English
 ## unless otherwise indicated.
 
+-product-fx-monitor = Firefox Monitor
```
flodolo, line 6: We shouldn't add new terms if we already have them (this and Firefox).

If the concern is making sure they won't be removed when we clean up the file, then group them together with a clear warning.

locales/en/landing.ftl
```diff
@@ -0,0 +1,44 @@
...
+
+## Hero
+
+find-out-if-breached = Find out if you’ve been in a data breach
+stay-safe-with-tool = Stay safe with a simple yet powerful tool that tracks whether your personal information has been exposed in a data breach. We’ll help you understand what to do next and continuously monitor for any new breaches — all at no cost to you.
+sign-up-free = Sign up for free
+
+## Why use Firefox Monitor?
+
+why-use-monitor = Why use Firefox Monitor?
```
flodolo, line 11: Use brand for Monitor

locales/en/landing.ftl
```diff
@@ -0,0 +1,44 @@
...
+identifying-breaches = Identifying and resolving data breaches is an important step in ensuring your online privacy.
+protect-account = Protect your accounts
+protect-account-prevent-hackers = Prevent hackers from getting into your accounts with breached passwords.
+prevent-fraud = Prevent fraud & cybercrimes
```
flodolo, line 15: I think the content guidelines suggest to stay away from &
https://mozilla-hub.atlassian.net/wiki/spaces/FIREFOX/pages/11043269/Punctuation#Punctuation-Ampersand(%26)

locales/en/landing.ftl
```diff
@@ -0,0 +1,44 @@
...
+## Your privacy is safe with us
+
+safe-with-us = Your privacy is safe with us
+parent-company = At {-brand-Mozilla}, the parent company of {-brand-firefox} and {-product-fx-monitor}, we believe individuals’ security and privacy on the internet is fundamental and must not be treated as optional.
+our-mission = Our mission is to build a better internet — one where people can shape their own experiences, feel empowered and stay safe. We’re wholly owned by the {-brand-mozilla-foundation}, a non-profit, which enables us to put individual and public benefit over profit.
+learn-more-mission = Learn more about our mission
+
+## Top questions about Firefox Monitor
+
+top-questions-about-monitor = Top questions about Firefox Monitor
```
flodolo, line 39: Use brand for Monitor.

locales/en/landing.ftl
```diff
@@ -0,0 +1,44 @@
...
+see-all-faq = See all FAQs
+
+## See if you’ve been in a data breach
+
+see-if-data-breach = See if you’ve been in a data breach
```
flodolo, line 44: Please keep an empty line at the end of files.

locales/en/app.ftl
toufali: Thanks @flodolo – you're right on the concern.  Additionally the way the original variables were named is confusing/misleading (looks like you've also called this out RE "brand" below).

Question: I assume placement in the page does not matter?  Given your feedback, my plan is to move the new (duplicated) terms down into the "V2" section so that it's clear these are part of the new build.  Then, after cleanup/deletion of old terms, move them back to top-of-file with existing brand vars that remain.

flodolo: Yes, placement is not important, and it would be OK to move them around afterwards within the file.

mansaj (approved):
LGTM on the first pass, will look at it more later this evening

flodolo (approved):

flodolo:
This landed on main, so it's not exposed to localization yet. Is that expected, as in you still have more work to do on it?

toufali:
@flodolo correct - we had some copy changes since this was merged. I'm set to go through them and update Mon/Tue.

toufali:
(Thanks for checking though)

PR #2692, opened 2022-11-08
Mntor 921

mansaj:
Create a `PUT` endpoint to update the breach resolution state:
- endpoint: `api/v1/user/breaches`
- the body requires these attributes:
  - affectedEmail
  - recencyIndex
  - resolutionsChecked (array)

The call returns the updated breach resolution object


package.json
```diff
@@ -78,8 +78,8 @@
     "url": "git+https://github.com/mozilla/blurts-server.git"
   },
   "scripts": {
-    "start": "source .env && if [[ $MONITOR_V2 == 'true' ]]; then npm start --workspace=src; else npm run build & node server.js; fi;",
-    "dev": "source .env && if [[ $MONITOR_V2 == 'true' ]]; then npm run dev --workspace=src; else nodemon server.js; fi;",
+    "start": ". .env && if [[ $MONITOR_V2 == 'true' ]]; then npm start --workspace=src; else npm run build & node server.js; fi;",
```
toufali, line 81: Interesting!  Can you explain this change?

src/controllers/breaches.js
```diff
@@ -1,10 +1,11 @@
...
-import { setBreachResolution, setBreachesResolved, updateBreachStats } from '../db/tables/subscribers.js'
+import { setBreachResolution, updateBreachStats } from '../db/tables/subscribers.js'
 import { getUserEmails } from '../db/tables/email_addresses.js'
 import { getBreachesForEmail, filterBreaches } from '../utils/hibp.js'
 import { filterBreachDataTypes } from '../utils/breach-resolution.js'
 import { getSha1 } from '../utils/fxa.js'
+
 async function breachesPage (req, res) {
   // TODO: remove: to test out getBreaches call with JSON returns
   const breachesData = await getAllEmailsAndBreaches(req.user, req.app.locals.breaches)
```
toufali, line 11: Should `breachesData` now instead await `getBreaches()` below?

src/controllers/breaches.js
```diff
@@ -189,13 +176,13 @@ async function bundleVerifiedEmails (options) {
...
-    // add resolved status from v1: breach_resolved
+    // add resolved status from v1: breaches_resolved
     breach.IsResolved = !!resolvedBreachesV1.includes(breach.recencyIndex)
 
     // add resolved status from v2: breach_resolution
     if (breachResolutionV2[breach.recencyIndex] && !breach.IsResolved) {
-      const IsResolved = !breachResolutionV2[breach.recencyIndex].isActive
-      breach.IsResolved = breach.IsResolved || IsResolved
+      const IsResolved = breachResolutionV2[breach.recencyIndex].isResolved
+      breach.IsResolved = breach.IsResolved || IsResolved // if either v1 or v2 is marked as resolved, breach is resolved
```
toufali, line 185: Could this loop be optimized or more readable with [nullish coalescing operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing)?
i.e.
```
for (const breach of foundBreachesWithRecency) {
    // if either v1 or v2 is marked resolved, breach is resolved
    breach.IsResolved = !!resolvedBreachesV1.includes(breach.recencyIndex) ?? !!breachResolutionV2[breach.recencyIndex].isResolved
}
```

src/controllers/breaches.js
```diff
@@ -189,13 +176,13 @@ async function bundleVerifiedEmails (options) {
     : []
 
   for (const breach of foundBreachesWithRecency) {
-    // add resolved status from v1: breach_resolved
+    // add resolved status from v1: breaches_resolved
     breach.IsResolved = !!resolvedBreachesV1.includes(breach.recencyIndex)
```
toufali, line 180: Can/should we use lowercase "i", i.e. breach.isResolved?

src/views/layouts/main.js
```diff
@@ -11,7 +11,7 @@ const landingHeader = data => `
 
 const mainHeader = data => `
 <header>
-  <a href='/user/breaches'><img class='monitor-logo' src='images/monitor-logo-transparent.png' width='220' height='40'></a>
+  <a href='/user/breaches'><img class='monitor-logo' src='/images/monitor-logo-transparent.png' width='220' height='40'></a>
```
toufali, line 14: Ah – I made all these path changes in [my branch too](https://github.com/mozilla/blurts-server/pull/2688/files#diff-e0482d58197a5dd0faa98700a3ee1b5368f922040ac9e494029282089a3e91ca)
I also added additional code on the same line, so it's likely this will result in conflict.  

Might be best to remove these from your PR and keep in mine?

package.json
mansaj: `source` is bash, heroku will throw an error if you do this.. It'll help your heroku story

src/controllers/breaches.js
mansaj: No, because `getBreaches` is used for the API (returns a callback of JSON). What we want here is the underlying function. I'll have to refactor this function `getAllEmailsAndBreaches` and name it something else during the Q1 refactor

src/controllers/breaches.js
mansaj: I think we are keeping it consistent with the rest of the booleans from HIBP

src/views/layouts/main.js
mansaj: Done

src/controllers/breaches.js
mansaj: This is cool, however, without the if checks, I think we might see a null pointer exception if
`breachResolutionV2[breach.recencyIndex]` does not exist

mansaj: we might be able to do: `!!breachResolutionV2[breach.recencyIndex]?.isResolved`

mansaj: The issue I have with ^ is that it's actually not as readable to me. I think the traditional if/else statements are more readable for engineers of all levels.

mansaj: Updated with `||`

toufali (approved):

PR #2701, opened 2022-11-15
MNTOR-1009: heroku proxy fix

mansaj:
Because of heroku's proxy settings, request / cookies were not persisted between calls. Hence, after login, an error is thrown when `req.session` does not contain the state hash.

Setting the trust proxy to high and securing the cookie allowed the cookie to persist.
If cookie.secure is set as true, for nodejs behind proxy, "trust proxy" needs to be set.

### DIFF
--- a/src/app.js
+++ b/src/app.js
@@ -40,6 +40,14 @@ app.use((req, res, next) => {
   next()
 })
 
+// MNTOR-1009:
+// Because of heroku's proxy settings, request / cookies are not persisted between calls
+// Setting the trust proxy to high and securing the cookie allowed the cookie to persist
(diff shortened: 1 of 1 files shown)


mansaj:
> Is it possible that GCP Stage and Prod will also require this enabled? If so, we'd probably need something like 👇 ? `if (AppConstants.NODE_ENV !== 'dev') app.set('trust proxy', 1)`
> 
> But I'm happy to test without it first – once we push to stage, we can confirm the need or not (unless you know the answer already!)

I wouldn't blindly set it for prod/stage. When it comes to security practices, we want to abide by the least privilege principle, whether it's access control or "trust" in this case.

PR #2702, opened 2022-11-15
always force clean deploy for heroku

mansaj:
when main is merged, always force deploy to heroku, overwriting current heroku main


toufali:
LGTM.  Throwing @jwhitlock on for sanity check since he was last to fiddle with Heroku.

mansaj:
Merging this in now. @jwhitlock let me know if you see an issue with it.

PR #2703, opened 2022-11-18
Mntor-989: Error Handling

mansaj:
Added basic error handling:
* middleware to handle errors in one place
* 404 response for all the page routes
* 405 response for all the API routes
* extended error classes for better err throwing

### DIFF
--- a/src/app.js
+++ b/src/app.js
@@ -32,7 +32,6 @@ async function getRedisStore () {
 }
 
 // middleware
-app.use(express.json())
 app.use(helmet())
 app.use((req, res, next) => {
   if (!req.headers.accept.startsWith('text/html')) return next() // only update for html req
@@ -74,6 +73,7 @@ try {
 // routing
 app.use(express.static(staticPath))
 app.use('/', indexRouter)
+app.use(express.json())
 app.use(errorHandler)
 
 // start server
--- a/src/app.js
+++ b/src/app.js
@@ -6,6 +6,7 @@ import accepts from 'accepts'
 import redis from 'redis'
 
 import AppConstants from './app-constants.js'
+import { logErrors, errorHandler } from './middleware/error.js'
 import { initFluentBundles, updateAppLocale } from './utils/fluent.js'
 import { loadBreachesIntoApp } from './utils/hibp.js'
 import indexRouter from './routes/index.js'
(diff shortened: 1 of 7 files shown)


src/app.js
```diff
@@ -71,8 +72,9 @@ try {
 }
 
 // routing
-app.use('/', indexRouter)
 app.use(express.static(staticPath))
+app.use('/', indexRouter)
```
toufali, line 76: Good call ordering this after static files.  Should we also move `app.use(express.json())` after static?  Or should that middleware be called only specifically when needed per route?

src/esbuild.js
```diff
@@ -16,5 +16,6 @@ esbuild.build({
   minify: AppConstants.NODE_ENV !== 'dev',
   sourcemap: AppConstants.NODE_ENV !== 'dev',
   splitting: true,
-  treeShaking: true
+  treeShaking: true,
+  platform: 'node'
```
toufali, line 20: ESBuild is only bundling for the client, so I think leaving `platform` null is actually preferred, since it defaults to `'browser'`.  

That being said, the docs suggest we might use `platform: 'neutral'` instead of `'browser'` to use more modern ES features.  `'browser'` wraps everything in an iife, which we don't necessarily need/want.
https://esbuild.github.io/api/#platform

src/middleware/error.js
```diff
@@ -0,0 +1,47 @@
...
+  log.error('error', { stack: err.stack })
+  const errStatus = err.statusCode || 500
+  const errMsg = err.message || 'Empty error message'
+  res.status(errStatus).json({
+    success: false,
+    status: errStatus,
+    message: process.env.NODE_ENV !== 'production' ? errMsg : 'Something went wrong', // hide error message when in production
+    stack: process.env.NODE_ENV === 'dev' ? err.stack : {} // hide stack when not in dev
+  })
+}
```
toufali, line 28: code-style nit: should we prefer top-level function declarations within modules?  Especially when the functions include const, it reads better to me 🤷

toufali (approved):

src/app.js
mansaj: I think that's a good idea actually

src/esbuild.js
mansaj: I think I added this in order to get jest working. I'm still in the process of fixing jest to run tests within `/src`... it seems like a long process

src/middleware/error.js
mansaj: Sure.

PR #2705, opened 2022-11-18
Fix l10n linter workflow: update paths and config, resolve errors

flodolo:


.github/workflows/reference_linter.yaml
```diff
@@ -1,10 +1,10 @@
 name: Lint Reference Files
 on:
   push:
-    paths: ['en/**.ftl' ]
+    paths: ['locales/en/**.ftl' ]
     branches: [ main, main-kanary, localization ]
```
toufali, line 5: we can remove `main-kanary` now if you like, it won't be used again.

toufali (approved):

PR #2707, opened 2022-11-21
fix Mozilla footer link

toufali:
Fixes Mozilla logo link hit-area extends full width of footer.  Also makes minor style improvements and uses safe but less-strict default for referrer-policy.


mansaj (commented):
LGTM

PR #2710, opened 2022-11-21
Mntor-975: Breach Checklist

mansaj:
Add a `breachChecklist` field to the breach details JSON return:
* Added fluent file for the 13 data classes
* The checklist object should include the `body`, `header`, and `priority`.
* The breach checklist should contain the same data class as defined by the breach in `DataClasses` field


locales/en/breach-checklist.ftl
```diff
@@ -0,0 +1,57 @@
+# links
+-link-firefox-password-manager = <a href="https://www.mozilla.org/en-US/firefox/features/password-manager/">Firefox Password Manager</a>
```
flodolo, line 2: These URLs are not supposed to be translated, so these links should not be defined as terms.

Also, there is a difference between a term and a message, and both can be referenced in other message. This is not a good use for terms (brand name typically is).

Also, remove `en-US` from mozilla.org strings, you should let the browser ask for the right language.

locales/en/breach-checklist.ftl
```diff
@@ -0,0 +1,57 @@
...
+-link-mozilla-vpn = <a href="https://www.mozilla.org/en-US/products/vpn/">Mozilla VPN</a>
+-link-firefox-relay = <a href="https://relay.firefox.com/">Firefox Relay</a>
+-link-equifax = <a href="https://www.equifax.com/personal/credit-report-services/credit-freeze/">Equifax</a>
+-link-experian = <a href="https://www.experian.com/freeze/center.html">Experian</a>
+-link-transunion = <a href="https://www.transunion.com/credit-freeze">TransUnion</a>
+
+# password
```
flodolo, line 9: If these comments are supposed to apply to both strings, they should be group-level comments (2 dashes). Also, it would be nice to provide more context than one work.

locales/en/breach-checklist.ftl
```diff
@@ -0,0 +1,57 @@
...
+breach-checklist-pw-header = Go to {$companyLink} to change your password and enable two-factor authentication (2FA).
```
flodolo, line 10: When there is a variable, please provide a comment explaining what replaces it at run-time.

locales/en/breach-checklist.ftl
```diff
@@ -0,0 +1,57 @@
...
+-link-firefox-password-manager = <a href="https://www.mozilla.org/en-US/firefox/features/password-manager/">Firefox Password Manager</a>
+-link-mozilla-vpn = <a href="https://www.mozilla.org/en-US/products/vpn/">Mozilla VPN</a>
+-link-firefox-relay = <a href="https://relay.firefox.com/">Firefox Relay</a>
+-link-equifax = <a href="https://www.equifax.com/personal/credit-report-services/credit-freeze/">Equifax</a>
+-link-experian = <a href="https://www.experian.com/freeze/center.html">Experian</a>
+-link-transunion = <a href="https://www.transunion.com/credit-freeze">TransUnion</a>
+
+# password
+breach-checklist-pw-header = Go to {$companyLink} to change your password and enable two-factor authentication (2FA). 
+breach-checklist-pw-body = Make sure your password is unique and hard to guess. If this password is used on any other accounts, be sure to change it there too. {-link-firefox-password-manager} can help you securely keep track of all of your passwords.
```
flodolo, line 11:
```suggestion
breach-checklist-pw-body = Make sure your password is unique and hard to guess. If this password is used on any other accounts, be sure to change it there too. <a>{-brand-name-firefox} Password Manager</a> can help you securely keep track of all of your passwords.
```

This is an example. You want the actual Mozilla brand defined as a term, but not the whole link.

locales/en/breach-checklist.ftl
```diff
@@ -0,0 +1,57 @@
...
+
+# email
+breach-checklist-email-header = Protect your email with an email masking service like {-link-firefox-relay}.
+breach-checklist-email-body = This can hide your true email address while forwarding emails to your real inbox.
+
+# social security number
+breach-checklist-ssn-header = Monitor your credit report for accounts, loans, or credit cards you don't recognize.
+breach-checklist-ssn-body = You can also consider freezing your credit on {-link-equifax}, {-link-experian} and {-link-transunion} to stop scammers from opening new accounts in your name. It’s free and won’t affect your credit score.
```
flodolo, line 19: This group needs a clear explanation, as this is something that makes sense only for US citizens.

I wouldn't define terms for external brands.

locales/en/breach-checklist.ftl
```diff
@@ -0,0 +1,57 @@
...
+breach-checklist-cc-header = Report this breach to your credit card issuer and request a new card with a new number.
+breach-checklist-cc-body = You should also review your credit card statements for unrecognized charges.
+
+# bank account
+breach-checklist-bank-header = Notify your bank immediately that your account number has been compromised.
+breach-checklist-bank-body = Doing so faster could give you more legal protections to help you recover any losses. You’ll also want to check your accounts for any unrecognized charges.
+
+# pin
+breach-checklist-pin-header = Notify your card issuer and change your PIN immediately.
+breach-checklist-pin-body = Make sure your new PIN, or any other PIN, doesn't include easily guessed numbers such as your birth date or address.
```
flodolo, line 31: Please check the linter
https://github.com/mozilla/blurts-server/actions/runs/3516656994/jobs/5893493489

locales/en/breach-checklist.ftl
```diff
@@ -0,0 +1,57 @@
...
+# security questions
+breach-checklist-sq-header = Update your security questions on {$companyLink}. 
+breach-checklist-sq-body = Use long, random answers, and store them somewhere safe. Do this anywhere else you’ve used the same security questions.
+
+# historical passwords
+breach-checklist-hp-header = Create unique, strong passwords for any account where you’ve re-used passwords.
+breach-checklist-hp-body = A password manager like {-link-firefox-password-manager} (which is free and built-in to the Firefox browser) can help you keep track of all your passwords and access them securely from all your devices. 
+
+# general
+breach-checklist-general-header = Reach out to {$companyName} to inform them about this breach and ask for specific steps you can take.
```
flodolo, line 57: Keep an empty line at the end of the file.

locales/en/breach-checklist.ftl
mansaj: Done, thanks

locales/en/breach-checklist.ftl
mansaj: Is there a way to reference the same string in only this local file (local variable) in fluent?
Are you suggesting that we duplicate this string itself in multiple places in the file?

flodolo: You should put the text in the string, even if that means duplicating some content.

As mentioned, the URL should not be exposed in the string.

locales/en/breach-checklist.ftl
mansaj: Done, thanks!

locales/en/breach-checklist.ftl
mansaj: Done, thanks!

locales/en/breach-checklist.ftl
```diff
@@ -0,0 +1,67 @@
...
+# NOTE: { $companyLink } is a placeholder for the URL to the password change page of the company where the breach occurred 
+breach-checklist-pw-header = Go to <a>{$companyLink}</a> to change your password and enable two-factor authentication (2FA). 
+breach-checklist-pw-body = Make sure your password is unique and hard to guess. If this password is used on any other accounts, be sure to change it there too. <a>{-brand-name} Password Manager</a> can help you securely keep track of all of your passwords.
+
+## Prompts the user for changes when there is a breach detected of email
+
+breach-checklist-email-header = Protect your email with an email masking service like <a>{-brand-relay}</a>.
+breach-checklist-email-body = This can hide your true email address while forwarding emails to your real inbox.
+
+## Prompts the user for changes when there is a breach detected of social security number
```
flodolo, line 14: As mentioned, this would benefit from a short explanation on what a credit report is, what freezing credit means, otherwise we risk getting random, out-of-topic literal translations.

locales/en/breach-checklist.ftl
```diff
@@ -0,0 +1,67 @@
...
+breach-checklist-dob-header = Change any passwords or PINs that include your date of birth.  
+breach-checklist-dob-body = Birth dates are easy to find in public records, and people who find it could easily guess your PIN.
+
+## Prompts the user for changes when there is a breach detected of phone number
+
+breach-checklist-phone-header = Protect your phone number with a masking service like <a>{-brand-relay}</a>, which hides your true phone number. 
+
+## Prompts the user for changes when there is a breach detected of security questions
+
+# NOTE: { $companyLink } is a placeholder for the URL to the password change page of the company where the breach occurred
```
flodolo, line 55: It's not the link to the password page though? Should this be `$companyName` instead?

locales/en/breach-checklist.ftl
```diff
@@ -0,0 +1,67 @@
...
+## Prompts the user for changes when there is a breach detected of security questions
+
+# NOTE: { $companyLink } is a placeholder for the URL to the password change page of the company where the breach occurred 
+breach-checklist-sq-header = Update your security questions on <a>{$companyLink}</a>. 
+breach-checklist-sq-body = Use long, random answers, and store them somewhere safe. Do this anywhere else you’ve used the same security questions.
+
+## Prompts the user for changes when there is a breach detected of historical password
+
+breach-checklist-hp-header = Create unique, strong passwords for any account where you’ve re-used passwords.
+breach-checklist-hp-body = A password manager like <a>{-brand-name} Password Manager</a> (which is free and built-in to the Firefox browser) can help you keep track of all your passwords and access them securely from all your devices.
```
flodolo, line 62:
```suggestion
breach-checklist-hp-body = A password manager like <a>{-brand-name} Password Manager</a> (which is free and built-in to the {-brand-name} browser) can help you keep track of all your passwords and access them securely from all your devices. 
```

flodolo:
FYI, you can run the linter locally with Python, only need to install a package once

```
pip install -r .github/requirements.txt
moz-fluent-lint ./locales/en --config .github/linter_config.yml
```

locales/en/breach-checklist.ftl
mansaj: According to copy, this should be the website. Updated the var name and comment to better reflect intent

locales/en/breach-checklist.ftl
mansaj: Looping in content/copy for an appropriate explanation here.

locales/en/breach-checklist.ftl
mansaj: Looping in content/copy for an appropriate explanation here.

flodolo (approved):
FTL looks good now.

I wonder if we end up showing content for SSN to non-US breaches? Because that information is really relevant only for US.

mansaj:
> FTL looks good now.
> 
> I wonder if we end up showing content for SSN to non-US breaches? Because that information is really relevant only for US.

Good call, we will keep that in mind when implementing geo-filtering logic in future releases

locales/en/breach-checklist.ftl
```diff
@@ -0,0 +1,72 @@
+# Strings for the breach details checklists
+
+## Prompts the user for changes when there is a breach detected of password
+
+# NOTE: { $companyWebsite } is a placeholder for the URL to the website of the company where the breach occurred 
+breach-checklist-pw-header = Go to <a>{$companyWebsite}</a> to change your password and enable two-factor authentication (2FA).
```
toufali, line 6: Without a hook into the the anchor's attributes, adding the relevant link is difficult.  I've seen different forms of doing this throughout the old repo, but the link pattern I settled on is similar to this:

```
// locales/en/email-strings.ftl
email-2022-hibp-attribution = Breach data provided by <a { $hibp-link-attr }>{ -brand-HIBP }</a>
```

This way, we have a hook to provide the `href` attribute.  Using Handlebars, I did it like this:
```hbs
// views/partials/email_partials/email-footer.hbs
<p>{{{ getString "email-2022-hibp-attribution" hibp-link-attr='href="https://haveibeenpwned.com/"' }}}</p>
```

With our newest codebase using Template Literals it might be similar to 
```js
`<p>${ getMessage('email-2022-hibp-attribution', { hibp-link-attr: 'href="https://haveibeenpwned.com/"' }) }</p>`
```

Thinking about it further, a shorter naming convention might be better for the new repo:
Instead of ` { $hibp-link-attr }`, we could use an identifier prefix with an "href" sufix, i.e. ` { $hibp-href }`.

cc @flodolo

locales/en/breach-checklist.ftl
```diff
@@ -0,0 +1,72 @@
+# Strings for the breach details checklists
```
toufali, line 1: I've been breaking out the .ftl files to match the naming conventions used across the repo.  So in addition to `app.ftl` I recently added `landing.ftl` which includes all landing page copy.  Should we continue this pattern here?  This file would be named `breaches.ftl` corresponding with the breaches page, and any other related breaches copy (there shouldn't be much more) would also be added to this page...

locales/en/breach-checklist.ftl
```diff
@@ -0,0 +1,72 @@
...
+
+## Prompts the user for changes when there is a breach detected of password
+
+# NOTE: { $companyWebsite } is a placeholder for the URL to the website of the company where the breach occurred
```
toufali, line 5: Is `{ $companyWebsite }` too generic?  Would something like `{ $breachedCompanyUrl }` and `{ $breachedCompanyName }` be more appropriate when thinking about future features and translations?  For example, a data broker might also technically fall under the `{ $companyWebsite }` var.

package.json
```diff
@@ -87,6 +87,7 @@
     "lint": "npm run lint:css && npm run lint:js",
     "lint:js": "eslint .",
     "lint:css": "stylelint public/css/",
+    "lint:fluent": "pip3 install -r .github/requirements.txt && moz-fluent-lint ./locales/en --config .github/linter_config.yml",
```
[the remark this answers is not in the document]
flodolo: Good point. Not only it would try to install it every time, but it would not update it if already installed.

One possible alternative is to use a virtualenv, and call a bash script here ([example](https://github.com/flodolo/mozpm_stats/blob/main/firefox_stats/extract_stats.sh)). Then just put python and virtualenv as requirements in the README?

locales/en/breach-checklist.ftl
flodolo: For us, it would be better if we could have just the anchor without any attribute exposed. That way, there's no risk of the variable being altered, and you don't need to retranslate everything if you decide to add other attributes (id, class, etc.). In projects where we used the variable approach, the latter seems to happen a bit too often :-(

src/utils/breach-resolution.js
```diff
@@ -24,83 +26,121 @@ const BreachDataTypes = {
...
+ * @returns {*} void
+ */
+function appendBreachResolutionChecklist (userBreachData) {
+  const { verifiedEmails } = userBreachData
+  for (const { breaches } of verifiedEmails) {
+    breaches.forEach(b => {
+      const dataClasses = b.DataClasses
+      const args = {
+        companyName: b.Name,
+        companyWebsite: `https://${b.Domain}`
```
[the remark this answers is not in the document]
mansaj: I think it's probably a good idea! We probably shouldn't direct people to a HTTP site anymore as a security product

locales/en/breach-checklist.ftl
mansaj: Done thanks!

locales/en/breach-checklist.ftl
mansaj: Done, thanks

locales/en/breach-checklist.ftl
```diff
@@ -0,0 +1,72 @@
+# Strings for the breach details checklists
+
+## Prompts the user for changes when there is a breach detected of password
+
+# NOTE: { $companyWebsite } is a placeholder for the URL to the website of the company where the breach occurred 
+breach-checklist-pw-header = Go to <a>{$companyWebsite}</a> to change your password and enable two-factor authentication (2FA). 
+breach-checklist-pw-body = Make sure your password is unique and hard to guess. If this password is used on any other accounts, be sure to change it there too. <a>{-brand-name} Password Manager</a> can help you securely keep track of all of your passwords.
```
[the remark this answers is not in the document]
mansaj: We also have some uses of `{-brand-name}` in `app.ftl` and `bento.ftl`.. Would we want to also change those to `-brand-firefox` for consistency?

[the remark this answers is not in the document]
mansaj: Did it anyways

package.json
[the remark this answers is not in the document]
mansaj: Good point, guys. Separated it into readme + command

flodolo (changes requested):
Please undo the change to old strings for the term, we don't want to ask all locales to retranslate.

Also, any change of that type requires a new ID.

locales/en/breach-checklist.ftl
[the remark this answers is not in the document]
toufali: Sorry I didn't reply in time; looks like @flodolo caught this though.  We don't want/need to change the old stuff, it will be deprecated.  Until then, it's fine for the old site to refer to the old strings.  

We will create new .ftl files as needed for the new site, and things that are site-wide can go at the bottom of `app.ftl` with the other new strings.  After old site is retired, we'll delete the old strings from `app.ftl`.

locales/en/breach-checklist.ftl
toufali: @mansaj 
Per @flodolo, we can leave this as is, and in a future release add a util function to parse anchor tags adding `href` and any other attributes.  I'll make a note.

locales/en/breaches.ftl
```diff
@@ -0,0 +1,72 @@
...
+
+## Prompts the user for changes when there is a breach detected of historical password
+
+breach-checklist-hp-header = Create unique, strong passwords for any account where you’ve re-used passwords.
+breach-checklist-hp-body = A password manager like <a>{-brand-firefox} Password Manager</a> (which is free and built-in to the {-brand-firefox} browser) can help you keep track of all your passwords and access them securely from all your devices. 
+
+## Prompts the user for changes when there is a breach detected of other types
+
+# NOTE: { $companyName } is a placeholder for the name of the company where the breach occurred 
+breach-checklist-general-header = Reach out to {$companyName} to inform them about this breach and ask for specific steps you can take.
```
toufali, line 72: Should this be `{$breachedCompanyName}` to match `{$breachedCompanyUrl}` above?

locales/en/breach-checklist.ftl
mansaj: Sounds good

toufali (approved):

PR #2713, opened 2022-11-22
Update Fluent linter to check for brand names

flodolo:


.github/linter_config.yml
```diff
@@ -33,3 +33,19 @@ ID02:
...
+CO01:
+    enabled: true
+    brands:
+        - Firefox
+        - Mozilla
+        - Relay
+        - Monitor
+    exclusions:
+        messages:
+            - breach-checklist-ssn-header
```
flodolo, line 46: This one is for #2710

flodolo (commented):
Since Monitor is a common word, we might have false positive, but I think it's still worth having the check in place.

toufali (approved):

mansaj (approved):

PR #2717, opened 2022-11-29
MNTOR-974: force relogin after logout

mansaj:
Post login, the user will be forced to re-enter their password

Reference FxA fix: FXA-6155

### DIFF
--- a/src/controllers/auth.js
+++ b/src/controllers/auth.js
@@ -22,7 +22,8 @@ function init (req, res, next, client = FxAOAuthClient) {
   const fxaParams = new URL(req.url, SERVER_URL)
 
   req.session.utmContents = {}
-
+  url.searchParams.append('prompt', 'login')
+  url.searchParams.append('max_age', 0)
   url.searchParams.append('access_type', 'offline')
   url.searchParams.append('action', 'email')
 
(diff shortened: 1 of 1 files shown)


src/controllers/auth.js
```diff
@@ -64,7 +65,7 @@ async function confirmed (req, res, next, client = FxAOAuthClient) {
 
   // Check if user is signing up or signing in,
   // then add new users to db and send email.
-  if (!existingUser || existingUser.fxa_refresh_token === null) {
+  if (!existingUser) {
     // req.session.newUser determines whether or not we show "fxa_new_user_bar" in template
     req.session.newUser = true
     const signupLanguage = req.headers['accept-language']
```
toufali, line 71: I know this is beyond the scope of this PR, but for future we might consider using `req.appLocale` instead of `req.headers['accept-language']`.  `req.appLocale` is set via middleware in `app.js` and is supposed to intelligently negotiate a language that meets the request and that Monitor supports.  In contrast, the pure header as shown here might store multiple locales or even `*` or `null`.

Just wanted to share in case you or I run into this in the future!

toufali (approved):

src/controllers/auth.js
mansaj: Good to know, let me change this

PR #2719, opened 2022-11-30
Mntor-954: fluent error

mansaj:
* Added tests for utility files
* Fluent error


src/utils/fluent.test.js
```diff
@@ -0,0 +1,34 @@
+import { initFluentBundles, updateAppLocale, getMessage, getRawMessage, fluentError } from './fluent.js'
+
+describe('utils:fluent', () => {
+  beforeAll(async () => {
+    await initFluentBundles()
+    updateAppLocale('en')
```
toufali, line 6: Sorry, I just refactored this here: https://github.com/mozilla/blurts-server/pull/2720
The new function name is now `updateLocale()`, since "app" was misleading – locales are request-based not app-wide

src/utils/breach-resolution.test.js
```diff
@@ -0,0 +1,55 @@
+import { initFluentBundles, updateAppLocale } from './fluent.js'
+import { filterBreachDataTypes, appendBreachResolutionChecklist, BreachDataTypes } from './breach-resolution.js'
+
+describe('utils:breach-resolution', () => {
+  beforeAll(async () => {
+    await initFluentBundles()
+    updateAppLocale('en')
```
toufali, line 7: See comment below... sorry!

src/utils/breach-resolution.test.js
```diff
@@ -0,0 +1,55 @@
+import { initFluentBundles, updateAppLocale } from './fluent.js'
```
toufali, line 1: See comment below re `updateAppLocale`

src/package.json
```diff
@@ -10,6 +10,7 @@
     "build": "node esbuild & npm run build:png && npm run build:svg && npm run copy:root",
     "build:png": "squoosh-cli --oxipng '{level: 2}' --quant '{numColors: 255}' -d ../dist/images ./client/images/*.png",
     "build:svg": "svgo -f ./client/images/ -o ../dist/images",
+    "test": "node --experimental-vm-modules ../node_modules/jest/bin/jest.js --runInBand --coverage .",
```
toufali, line 13: Is it ok to move the test script below `copy:root`?  `copy:root` is grouped/run with the build scripts.  Perhaps it would be more obvious if I changed it instead to `build:root`, even though it's just a copy script 🤷

src/utils/fluent.js
```diff
@@ -57,7 +57,7 @@ function updateLocale (requestedLocales) {
 * @param {string} id - The Fluent message id.
 */
 function getRawMessage (id) {
-  const locale = localStorage.getStore().get('locale')
+  const locale = localStorage?.getStore()?.get('locale') || 'en'
```
toufali, line 60: Good catch – this should fallback locale when not coming from a server request, which is what Jest is presumably doing.

I think just a single `?` would suffice?  `localStorage` will always be imported in this module, but the store hasn't been initialized via `run()` in `app.js` – that's only done in a request phase.

I tested, but can you confirm this works?
`localStorage.getStore()?.get('locale') || 'en'`

src/utils/fluent.js
```diff
@@ -78,7 +78,7 @@ function getRawMessage (id) {
 * // Returns "Hello, Jane!"
 */
 function getMessage (id, args) {
-  const locale = localStorage.getStore().get('locale')
+  const locale = localStorage?.getStore()?.get('locale') || 'en'
```
toufali, line 81: see comment above

toufali (approved):
Looks good, thanks for this!  I added a minor request to remove likely-redundant optional chaining `?`

src/utils/fluent.js
toufali, line 60:
```suggestion
  const locale = localStorage.getStore()?.get('locale') || 'en'
```

src/utils/fluent.js
toufali, line 81:
```suggestion
  const locale = localStorage.getStore()?.get('locale') || 'en'
```

toufali (approved):

src/utils/fluent.js
mansaj: Makes sense! Works.

PR #2721, opened 2022-12-02

toufali:
It's been a long time since `main` has been synced to `localization`, partly because we were waiting on final copy to be confirmed after a late round of user research.


flodolo (approved):
At this point, what is the expected timeline for this content to go live? I will need to make sure to communicate it to localizers.

toufali:
> At this point, what is the expected timeline for this content to go live? I will need to make sure to communicate it to localizers.

@flodolo the timeline has shifted and the updated target is now Q1.

PR #2723, opened 2022-12-03
Mntor 1015/breach table view

toufali:
Formats the data provided to the Breaches page into rows of expandable components.

Outstanding items to-do under separate tickets:
- format HIBP data to match Monitor specified copy for data classes and breach description
- localization
- refactor to allow sorting between email addresses (currently all emails are shown grouped without separation)
- remove old JSON test code


mansaj (approved):
Once we figure out the testing framework, we should start adding more tests when we create these util functions

mansaj:
Outside of the scope of this feature, but reminder:
```
/css/partials/dataRemoval.css
/css/partials/settings.css
/css/partials/dashboard.css
```
These aren't set up correctly yet.

toufali:
> Outside of the scope of this feature, but reminder:
> 
> ```
> /css/partials/dataRemoval.css
> /css/partials/settings.css
> /css/partials/dashboard.css
> ```
> 
> These aren't set up correctly yet.

Yep.  I added extra logic to dynamically add/remove based on their existence, but I removed because it felt like I was over-planning/engineering.  We'll see a raw 404 now if something isn't set up right - (naming-convention, missing styles).

Good heads-up tho, if the case ever comes up, I'll add more logic to handle partial stylesheet that is not required.

PR #2726, opened 2022-12-06
Mntor 1043/upgrade node 18

toufali:
Upgrades Node from 16.17 to 18.12.

It doesn't appear that we can selectively apply Node 18 to the V2 build alone.  This PR upgrades the entire build including V1.

The major casualty appears to be Squoosh, which was configured to auto-compress pngs.  Squoosh CLI is built by Google, but apparently [no longer maintained](https://github.com/GoogleChromeLabs/squoosh/pull/1266#issuecomment-1208149979), and does not support Node 18.  This PR switches from png to webp.  Unlike png, the webp images will already be compressed within `src/...` and only need to be copied to `dist/...`.

TODO: update readme


mansaj:
I think we currently have 3 high severity vulnerabilities when `npm i`
Can be fixed potentially by `npm audit fix`

src/package.json
```diff
@@ -3,14 +3,19 @@
...
     "prestart": "npm run build",
     "start": "node app.js",
     "dev": "nodemon app.js",
-    "build": "node esbuild & npm run build:png && npm run build:svg && npm run copy:root",
-    "build:png": "squoosh-cli --oxipng '{level: 2}' --quant '{numColors: 255}' -d ../dist/images ./client/images/*.png",
+    "build": "node esbuild & npm run copy:root & npm run copy:webp & npm run build:svg",
     "build:svg": "svgo -f ./client/images/ -o ../dist/images",
+    "copy:webp": "rsync -d --include='*.webp' ./client/images/ ../dist/images/",
     "copy:root": "rsync -d ./client/ ../dist/",
+    "convert:webp": "sh ./utils/webp.sh",
```
mansaj, line 18: This is meant to be run locally? I understand that a encoder is needed to run the script

src/package.json
```diff
@@ -3,14 +3,19 @@
...
+  },
   "scripts": {
     "prestart": "npm run build",
     "start": "node app.js",
     "dev": "nodemon app.js",
-    "build": "node esbuild & npm run build:png && npm run build:svg && npm run copy:root",
-    "build:png": "squoosh-cli --oxipng '{level: 2}' --quant '{numColors: 255}' -d ../dist/images ./client/images/*.png",
+    "build": "node esbuild & npm run copy:root & npm run copy:webp & npm run build:svg",
     "build:svg": "svgo -f ./client/images/ -o ../dist/images",
+    "copy:webp": "rsync -d --include='*.webp' ./client/images/ ../dist/images/",
```
mansaj, line 16: is there a reason that `copy:webp` and `copy:root` cannot be combined into one command?

src/package.json
toufali: yes – if any .png files are added we'll want to run this to convert them to .webp.  It would obviously be me running this for the foreseeable future.

I think we need to start a readme for V2...

toufali:
> I think we currently have 3 high severity vulnerabilities when `npm i` Can be fixed potentially by `npm audit fix`

@mansaj I believe those vulns are unrelated to this PR -- they exist on `main` and are likely more related to V1.  We'll need a separate ticket to address next sprint.  I think one of them involves upgrading Express...

src/package.json
mansaj: I'm just wondering how much work it is to have it as a part of the CICD pipeline. 
We would have to add a step to install the cwebp...

Oh another thing I forgot to mention, let's not put `.sh` in util. I think it makes sense to have a `script` folder for all the scripts that are not a part of the application code

src/package.json
mansaj: Since we shouldn't have anything other than `.webp` in `./client/images/`, can we just remove `copy:webp`?

src/package.json
toufali: @mansaj ok, I will start a `script` folder and add that file.

Having it in the CICD pipeline is tricky in ways that might take too long to explain in writing.  For .png files it made sense since there's a high likelihood of them being copied into the folder uncompressed.  But .webp arrive in the folder already compressed – this is probably better handled manually as needed.

src/package.json
toufali: we have and will continue to have quite a few .svg in there as well...

mansaj: Ah gotcha thanks

toufali: Actually – I was considering adding a script in the future to handle .svg the same way as .webp to keep them consistent.  i.e. compress .svg locally, manually via script.  When that's done, we could copy the entire image folder as you suggest...

src/package.json
mansaj: Sounds good

mansaj (approved):
pending folder change

PR #2728, opened 2022-12-06
replace GOT with native Node Fetch

toufali:


src/utils/fxa.js
```diff
@@ -79,15 +82,13 @@ async function getProfileData (accessToken) {
 
 async function sendMetricsFlowPing (path) {
```
toufali, line 83: @mansaj note, I was not able to test the `sendMetricsFlowPing` function – not sure it's being used?

mansaj:
@toufali can we take out Got from `package.json` yet?

src/utils/hibp.js
```diff
@@ -18,49 +17,49 @@ function _addStandardOptions (options = {}) {
...
-    },
-    responseType: 'json'
+    }
   }
   return Object.assign(options, hibpOptions)
 }
 
-async function _throttledGot (url, reqOptions, tryCount = 1) {
-  let response
+async function _throttledFetch (url, reqOptions, tryCount = 1) {
```
toufali, line 25: @mansaj note this may improve/reduce the amount of triggered errors (handling certain http status codes before throwing).  However, I have not refactored this.  As discussed, this should be refactored to throttle more elegantly.

src/utils/fxa.js
```diff
@@ -67,10 +69,11 @@ async function revokeOAuthTokens (subscriber) {
 
 async function getProfileData (accessToken) {
```
toufali, line 70: @mansaj as discussed, this request gets called twice – unrelated to Fetch replacement.  We should look into fixing in case we can avoid a second redundant network request.

src/utils/fxa.js
mansaj: Not for v2, I'm still trying to figure out what it does

mansaj: I think it's a part of the analytics

src/utils/fxa.js
mansaj: created a bug ticket: MNTOR-1064

mansaj:
@toufali FYI res.status() now takes number instead of str (otherwise, will throw a warning). Added a commit to address that

PR #2730, opened 2022-12-07
Mntor 1058/breaches formatted copy

toufali:
Formats breach table data and summary to screenshot below:
Screen Shot 2022-12-07 at 10 42 26 AM


src/views/partials/breaches.js
```diff
@@ -1,26 +1,38 @@
...
 </details>
 `
 
 function createRows (data) {
   const allEmailBreaches = data.verifiedEmails.flatMap(item => item.breaches)
-  const html = allEmailBreaches.map(breach => rowHtml(breach)).join('')
-  return html
+
+  allEmailBreaches.forEach(breach => {
+    breach.dataClassesFormatted = breach.DataClasses.map(item => getMessage(item)).join(', ')
```
flodolo, line 19: This is against best practices:
* It assumes that every language is OK with a comma, and a space before the command.
* It might end up using incorrect capitalization (it does even for English).

Are we doing this somewhere else in the code (I see you copied existing code)? Is there any way to display this differently, e.g. as a bullet point lists?

toufali: Thanks @flodolo! 

This is a common JS pattern and it's interesting to know the potential bad practice.  Yes, this is being done in the existing site as well, for example in `template-helpers/breach-detail.js`
```js
  sortedDataClasses.lowerPriority = sortedDataClasses.lowerPriority.join(', ')
```
Screen Shot 2022-12-07 at 12 05 32 PM

A bulleted list would not work well due to readability and space-constraints.  Note this list is used in two locations for each breach, once in the collapsed summary with ellipsis as needed, and the expanded list unabridged:

Screen Shot 2022-12-07 at 12 07 39 PM

 Do you have any other suggestions?

flodolo: CLDR can help solve at least the concatenation with hard-coded text part
https://cldr.unicode.org/translation/miscellaneous-displaying-lists
https://unicode-org.github.io/cldr-staging/charts/latest/by_type/miscellaneous.displaying_lists.html

I assume you should have access to some package that allows for list formatting based on locale? If not, I guess we can stick to the current code (sadly, there's no easy way to fix the capitalization).

toufali: @flodolo thanks again for flagging – After some research I learned there is a native JS API that includes list localization: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/ListFormat/ListFormat

The code has been refactored to include this API.

flodolo (approved):

PR #2738, opened 2022-12-28
Mntor 802/breaches email cta

toufali:
Add localized count/total emails monitored to the header on breaches page

TO-DO:
- link CTA to add new email flow

### DIFF
--- a/locales/en/breaches.ftl
+++ b/locales/en/breaches.ftl
@@ -8,7 +8,12 @@
 breach-heading-email = Data breaches for { $email-select }
 
 # $count is the number of emails a user has added out of $total allowed
-emails-monitored = { $count } of { $total } emails monitored
+emails-monitored = 
+  {
+    $total ->
+      [one] { $count } of { $total } email monitored
+     *[other] { $count } of { $total } emails monitored
+  }
 
 ## Breaches table
 
--- a/locales/en/breaches.ftl
+++ b/locales/en/breaches.ftl
@@ -7,6 +7,9 @@
 # $email-select is an interactive <select> element displaying the current email address
 breach-heading-email = Data breaches for { $email-select }
 
+# $count is the number of emails a user has added out of $total allowed
+emails-monitored = { $count } of { $total } emails monitored
+
 ## Breaches table
 
 column-company = COMPANY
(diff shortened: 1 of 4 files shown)


locales/en/breaches.ftl
```diff
@@ -7,6 +7,9 @@
 # $email-select is an interactive <select> element displaying the current email address
 breach-heading-email = Data breaches for { $email-select }
 
+# $count is the number of emails a user has added out of $total allowed
+emails-monitored = { $count } of { $total } emails monitored
```
flodolo, line 11: We should make this a plural, even if English might not need it (it will trigger the plural UI in Pontoon)

```
emails-monitored = 
  {
    $total ->
      [one] { $count } of { $total } email monitored
     *[other] { $count } of { $total } emails monitored
  }
```

flodolo (changes requested):

src/controllers/breaches.js
```diff
@@ -7,11 +7,13 @@ import { filterBreachDataTypes, appendBreachResolutionChecklist } from '../utils
 import { getSha1 } from '../utils/fxa.js'
 
 async function breachesPage (req, res) {
+  const emailCount = 1 + (req.user.email_addresses?.length || 0)
```
mansaj, line 10: So if I'm understanding this correctly, we are adding 1 for the primary email?

src/app-constants.js
```diff
@@ -73,8 +73,4 @@ optionalEnvVars.forEach(key => {
   if (key in process.env) AppConstants[key] = process.env[key]
 })
 
-AppConstants.VPN_PROMO_BLOCKED_LOCALES = AppConstants.VPN_PROMO_BLOCKED_LOCALES?.split(',')
```
mansaj, line 76: We should also delete this in `.env-dist` for consistency?

src/controllers/breaches.js
toufali: @mansaj yes, correct.  I will add a comment to make this more clear.

src/app-constants.js
toufali: Good idea!  

Though thinking on it a bit more – technically this var is still used in the legacy build.  The `app-constants.js` file I'm targeting here is specific to the refresh build.  We should definitely clean up `.env` and `env-dist` after refresh launch...

locales/en/breaches.ftl
toufali: updated to your code.  Didn't know Pontoon could do that – thanks!

flodolo (approved):

mansaj (approved):

PR #2739, opened 2022-12-29
Mntor 1026

mansaj:
* Refactoring
* HIBP util
* csurf protection


toufali (approved):

src/package.json
```diff
@@ -37,6 +37,7 @@
     "esbuild": "^0.15.10",
     "helmet": "^6.0.0",
     "svgo": "^2.8.0",
```
rhelmer, line 39: Already discussed in slack, going to see how difficult https://github.com/Psifi-Solutions/csrf-sync would be to add.

src/utils/hibp.js
```diff
@@ -247,16 +233,35 @@ async function subscribeHash (sha1) {
   return await kAnonReq(path, options)
 }
 
+/**
+A range subscription can be deleted with the following request:
+DELETE /range/[hash prefix]
+
+There is one possible response code that can be returned:
+1. HTTP 200: Range subscription already exists
+
```
rhelmer, line 242: This is valid JS but I think for JSDoc you want something more like:

```suggestion
/**
 * A range subscription can be deleted with the following request:
 * DELETE /range/[hash prefix]
 *
 * There is one possible response code that can be returned:
 * 1. HTTP 200: Range subscription already exists
 *
```

rhelmer (commented):
This seems to work as advertised, just a minor style nit, and a note about the alternative csrf library we discussed in Slack.

src/package.json
mansaj: I think with csrf-sync, we will now have to manually generate token per session or per request. 
The backend will generate via the `generateToken` method and the frontend will have to take that and add it to the request header in order for the request to go through..

src/utils/hibp.js
mansaj: Fixed

src/package.json
mansaj: @rhelmer with doublecsrf I believe the same flow is still needed as `csrf-sync`... As in every request (POST PUT PATCH DELETE etc) needs the token in the request header from `generateToken`.

mansaj: But I think this is what's intended.. When the frontend sends requests via POST etc, we will have to be mindful to set the header `x-csrf-token`

rhelmer (approved):

PR #2740, opened 2022-12-29
Mntor 1061/resolved filter

toufali:
**Unresolved/resolved filter bar**

- display amount of resolved/unresolved breaches above table
- update the amount dynamically based on email or resolved status
- allow user to select either option and update the table with respective data

image

Not included:
- zero-breach results design state (under separate PR)

### DIFF
--- a/locales/en/breaches.ftl
+++ b/locales/en/breaches.ftl
@@ -15,6 +15,11 @@ emails-monitored =
      *[other] { $count } of { $total } emails monitored
   }
 
+## Breaches resolved filter
+
+label-unresolved = Unresolved breaches
+label-resolved = Resolved breaches
+
 ## Breaches table
 
 column-company = COMPANY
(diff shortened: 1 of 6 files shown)


locales/en/breaches.ftl
```diff
@@ -15,6 +15,11 @@ emails-monitored =
      *[other] { $count } of { $total } emails monitored
   }
 
+## Breaches resolved filter
+
+label-unresolved = Unresolved breaches
```
flodolo, line 20: Can we use more semantic IDs, also to discourage reuse in other contexts? e.g. ´filter-label-unresolved`.

flodolo (approved):

rhelmer (approved):

locales/en/breaches.ftl
mansaj: Since @toufali is out on PTO, I've addressed this

Issue #2744, opened 2023-01-02
Unhandled error when deleting a subscriber with other emails


jwhitlock:
When sending a `POST /user/unsubscribe` via a verification token, a SQL error is raised:

```
delete from "subscribers" where "primary_verification_token" = $1 and "primary_sha1" = $2 - update or delete on table "subscribers" violates foreign key constraint "email_addresses_subscriber_id_foreign" on table "email_addresses"
```

This appears to crash the process and require a restart.

Tracked in Sentry:

* Team link: https://sentry.io/organizations/mozilla/issues/3520992928
* Public link: https://sentry.io/share/issue/76a5620c1ed64ba8b35c63824e75f1c1/

jwhitlock:
This also happens when `removeEmail` is called:

https://github.com/mozilla/blurts-server/blob/b26e7d757dcf32af89c65ec4689a80174578f26e/db/DB.js#L352-L361

However, the comment suggests it is desired to instead raise an exception. Changes are needed if we're going to us `ON DELETE CASCADE`, as suggested in PR #2746.

jwhitlock:
This error is still occurring when the FxA account is deleted and the user has additional emails:

https://github.com/mozilla/blurts-server/blob/5983132b374eaa802e22e174ae8e1e0b9135677f/db/DB.js#L406-L408

I would assume the user's intention is to delete the additional emails.

jwhitlock:
After some discussion, the developers agree that the "right thing" is to delete the related emails when a subscriber is deleted, due to a hard bounce, a spam complaint, or deleting their FxA account. The SQL-side solution in https://github.com/mozilla/blurts-server/pull/2746 would do this.

EMMLynch:
Closing since we've redesigned the site and functionality since this was created. If you feel that this is still needed, please let me know.

PR #2745, opened 2023-01-02
When unsubscribing a user by token, delete emails

jwhitlock:
This is a hotfix for v15.3.1, currently in production, to address issue #2744.

Add a test to remove a subscription via the primary verification token and hash. Fix `removeSubscriberByToken` by first deleting the user's emails, to avoid violating the primary key constraint.

This could be better handled by a cascading delete, but that fix should be applied to the main branch.

### DIFF
--- a/db/DB.js
+++ b/db/DB.js
@@ -365,6 +365,12 @@ const DB = {
     if (!subscriber) {
       return false
     }
+
+    // Delete subscriber's emails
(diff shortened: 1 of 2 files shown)


jwhitlock:
Integration tests passed. Merging with self-review.

toufali:
Thanks @jwhitlock – I updated your initial description to reflect that this has gone live as `v15.3.2.1`.

PR #2746, opened 2023-01-02
When deleting a subscriber, also delete emails

jwhitlock:
Add a migration that adds `ON DELETE` and `ON UPDATE` clauses to the foreign key constraint, so that deleting a subscriber also deletes their emails.

Add a test to remove a subscription via the primary verification token and hash. This test fails without the migration.

This is an alternate to the hotfix in PR #2745, which deleted the emails manually, but couldn't use a migration.

### DIFF
--- a/db/migrations/20230102203608_delete-emails-with-subscriber.js
+++ b/db/migrations/20230102203608_delete-emails-with-subscriber.js
@@ -1,0 +1,32 @@
+// Issue 2744: Delete email_addresses when the subscriber is deleted
+// email_addresses.subscriber_id added in 20190328111900_add_email_addresses_table.js
+
+/**
+ * @param { import("knex").Knex } knex
+ * @returns { Promise<void> }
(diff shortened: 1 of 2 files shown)


jwhitlock:
There is code that expects an error to be raised when a subscriber with emails is deleted:

https://github.com/mozilla/blurts-server/blob/b26e7d757dcf32af89c65ec4689a80174578f26e/db/DB.js#L352-L361

This would break that code, deleting a subscriber if they mark any email as spam. Changes are needed if we want to cascade deletion of subscribers:

1. Go ahead and delete a subscriber if they mark any monitor email as spam
2. Check that a subscriber has no remaining emails before running delete

jwhitlock:
After sleeping on it, I think it would be dangerous to use `ON DELETE CASCADE`, since `removeEmail` relies on the deletion failing if the subscriber has other emails.

PR #2751, opened 2023-01-04
MNTOR-1025: csrf

mansaj:


rhelmer:
My patch in MNTOR-803/settings-v2 now fails with `invalid csrf token` by default for API calls, so that's great :)

I did a quick local patch to test using the docs:

1. added a new `/api/v1/csrf-token` route
2. fetched it before making the next API call
3. called `/api/v1/email` with `x-csrf-token` and the value from (2) above

That seems to work! Per the docs, I think we can also just generate the token when we build the server-side pages as well, to eliminate that extra API call just to get the token. I also see the `monitor.x-csrf-token` cookie is set (and I modified the `fetch` to include credentials with `credentials: include`)

rhelmer (approved):

mansaj:
> My patch in [MNTOR-803](https://mozilla-hub.atlassian.net/browse/MNTOR-803)/settings-v2 now fails with `invalid csrf token` by default for API calls, so that's great :)
> 
> I did a quick local patch to test using the docs:
> 
>     1. added a new `/api/v1/csrf-token` route
> 
>     2. fetched it before making the next API call
> 
>     3. called `/api/v1/email` with `x-csrf-token` and the value from (2) above
> 
> 
> That seems to work! Per the docs, I think we can also just generate the token when we build the server-side pages as well, to eliminate that extra API call just to get the token. I also see the `monitor.x-csrf-token` cookie is set (and I modified the `fetch` to include credentials with `credentials: include`)

@rhelmer just to clarify, the GET calls are ignored right? I think if not set, `GET,HEAD,OPTIONS` by default are ignored. You were testing with a PATCH?

rhelmer:
> > My patch in [MNTOR-803](https://mozilla-hub.atlassian.net/browse/MNTOR-803)/settings-v2 now fails with `invalid csrf token` by default for API calls, so that's great :)
> > I did a quick local patch to test using the docs:
> > ```
> > 1. added a new `/api/v1/csrf-token` route
> > 
> > 2. fetched it before making the next API call
> > 
> > 3. called `/api/v1/email` with `x-csrf-token` and the value from (2) above
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > That seems to work! Per the docs, I think we can also just generate the token when we build the server-side pages as well, to eliminate that extra API call just to get the token. I also see the `monitor.x-csrf-token` cookie is set (and I modified the `fetch` to include credentials with `credentials: include`)
> 
> @rhelmer just to clarify, the GET calls are ignored right? I think if not set, `GET,HEAD,OPTIONS` by default are ignored. You were testing with a PATCH?

I tested using the new POST endpoints in my patch, yes it works fine w/ GET/HEAD/OPTIONS AFAICT.

PR #2755, opened 2023-01-10
MNTOR-1046: Add user menu

flozia:
## Description

This PR adds a partial for the user menu that is visible for users that are logged in.

## Reference

- Jira [MNTOR-1046](https://mozilla-hub.atlassian.net/browse/MNTOR-1046): Create user menu for site

## Screenshots

image


locales/en/app.ftl
```diff
@@ -821,6 +821,16 @@ meta-desc = Find out if you’ve been part of a data breach with {-brand-fx-moni
 brand-fx-monitor = {-brand-fx-monitor}
 sign-in = Sign In
 
+## User menu
+
+menu-button-title = User menu
+menu-button-alt = Open user menu
+menu-item-fxa = Manage your {-brand-firefox} account
```
flodolo, line 828: This should use the term for Firefox account
```suggestion
menu-item-fxa = Manage your {-brand-fx-account}
```

You probably want to move the definition and comment around line 815.

flodolo (changes requested):
As mentioned in the previous review, you need to move the existing string
https://github.com/mozilla/blurts-server/blob/main/locales/en/app.ftl#L20-L24

You can't have the same string defined twice, and you need to "close" the group-level comment defined on line 807, because it doesn't apply to this term.

locales/en/app.ftl
```diff
@@ -807,11 +803,18 @@ ad-unit-6-before-you-complete = Before you complete that next signup, use an ema
 ## The following messages are brands and should be kept entirely in English
 
 -brand-firefox = Firefox
+-brand-fx-account = Firefox account
```
flodolo, line 806: Remove this

flodolo (approved):
String-wise this is OK now.

Could you confirm that the layout can cope gracefully with longer translations (no truncation)?

flozia:
> String-wise this is OK now.
> 
> Could you confirm that the layout can cope gracefully with longer translations (no truncation)?

Thank you for the reviews @flodolo. There is no truncation or similar treatment. Longer translations wrap and should be accounted for in the layout. Here is a quick example:

image

src/client/css/partials/user-menu.css
```diff
@@ -0,0 +1,120 @@
...
+  align-items: flex-start;
+  display: flex;
+  gap: var(--padding-xs);
+  flex-direction: column;
+  padding: var(--padding-md) calc(var(--padding-md) * 1.5);
+  position: relative;
+  text-decoration: none;
+}
+
+.user-menu-header::after {
```
toufali, line 68: nit/fun experiment: I _think_ you could have just used an `<hr>` here instead of pseudo?  It think `<hr>` is as semantic as it gets, and might only require a single line of CSS: `border-image: var(--monitor-gradient) 1;`  What you have is perfectly fine too – only explore `<hr>` if you like fun 😄

src/controllers/data-removal.js
```diff
@@ -1,9 +1,13 @@
...
 import { dataRemoval } from '../views/partials/data-removal.js'
+import { userMenu } from '../views/partials/user-menu.js'
 
 function dataRemovalPage (req, res) {
+  const fxaProfileData = req.user.fxa_profile_json
+
   const data = {
-    partial: dataRemoval
+    partial: dataRemoval,
+    userMenu: userMenu(fxaProfileData)
```
toufali, line 10: We should come up with a better way to pass this data.  I don't have the answer yet (it's getting late!) but let's have a think about this to avoid needing to pass this for every partial.

src/views/main.js
```diff
@@ -66,6 +67,7 @@ const mainHeader = data => `
       </button>
     </li>
   </menu>
+  ${data.userMenu}
```
toufali, line 70: To expand on my note above regarding having to pass user-menu data to every partial, I _think_ an answer might be to build the user menu into the main header.  Maybe this should be a pattern for persistent elements.  Still needs a think but I think we can improve our framework here.

src/client/css/partials/user-menu.css
```diff
@@ -0,0 +1,120 @@
+.user-menu-wrapper {
+  font: normal 14px/1.5 Inter, Inter-fallback, sans-serif;
```
toufali, line 2: I often waffle between px and rem... for fonts we're so far using rem throughout (I believe) – should we stay consistent?

toufali (commented):
Great work @flozia !  A few notes after playing with it locally:

- Per the last design review meeting, we removed the "Contact us" menu item.
- Since a user is already signed in, is it possible to go direct into FxA account rather than prompt for login, or is the additional login prompt required for security, etc?
- Should we open a new window for FxA account?
- Looks like the avatar image is being blocked by CSP (on my machine it loads alt-text).  Let me know if you'd like help working this out.
- The user menu breaks the layout of the mobile nav button.  I originally intended the `<menu>` element to contain both user menu and mobile nav buttons.  This might make the layout work better?  (I imagine you'd need to add `display: flex` to the `<menu>` element).  Let's chat about this because I think my use of `<menu>` might also be confusing with my note below suggesting the user menu use `<menu>` as well.  (Even just writing this out is confusing!)
- Do you have the StyleLint extension installed?  On my end I'm seeing a few lint errors – they're minor, but it would still be good for us all to stay consistent.  I'm not sure if using StyleLint (and ESLint) extensions is suggested in the README, but it probably should be!

src/views/partials/user-menu.js
```diff
@@ -0,0 +1,57 @@
...
+  aria-haspopup='true'
+  class='user-menu-button'
+  title='${getMessage('menu-button-title')}'
+>
+  <img src='${data.avatar}' alt='${getMessage('menu-button-alt')}' />
+</button>
+`
+
+const userMenuPopover = data => `
+  <ul class='user-menu-list user-menu-popover' role='menu' hidden>
```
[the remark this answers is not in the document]
flozia: That’s a good point. I’m not sure if it would be semantically correct to nest a `<menu>` inside a `<menu>`, but in that case I probably should also drop the `role` attribute.

src/client/css/partials/user-menu.css
flozia: Cool — just replaced the pseudo element! That must be the first time I’m utilizing `border-image`.

src/controllers/data-removal.js
flozia: I agree. I didn’t want to blow up the scope of this PR, but thought about how to handle passing multiple partials could be done best. I created PR with a suggestion and for discussing possible approaches:
- #2761

src/client/css/partials/user-menu.css
flozia: Yes, let’s stay consistent. I converted the values to `rem`.

flozia:
Thanks for the review. I addressed some of the actionable items and have some comments for others:

> * Per the last design review meeting, we removed the "Contact us" menu item.

Removed.

> * Since a user is already signed in, is it possible to go direct into FxA account rather than prompt for login, or is the additional login prompt required for security, etc?

I’ll have to check how we can directly link to the FxA account without having to go through the login again: Looking into it.

> * Should we open a new window for FxA account?

Yes — done.

> * Looks like the avatar image is being blocked by CSP (on my machine it loads alt-text).  Let me know if you'd like help working this out.

I [whitelisted](https://github.com/mozilla/blurts-server/pull/2755/files#diff-c72a907ac323cd2f334ed0e2bd07d15ab62581c4753660c8a0d1c681b30be4b6R49) the profile URLs for loading image sources, but they are still being blocked it seems?

> * The user menu breaks the layout of the mobile nav button.  I originally intended the `<menu>` element to contain both user menu and mobile nav buttons.  This might make the layout work better?  (I imagine you'd need to add `display: flex` to the `<menu>` element).  Let's chat about this because I think my use of `<menu>` might also be confusing with my note below suggesting the user menu use `<menu>` as well.  (Even just writing this out is confusing!)

Moved the user menu into `<menu>` for now.

> * Do you have the StyleLint extension installed?  On my end I'm seeing a few lint errors – they're minor, but it would still be good for us all to stay consistent.  I'm not sure if using StyleLint (and ESLint) extensions is suggested in the README, but it probably should be!

Yes, I have them installed. Just overlooked the linter highlights (my editor theme colors concealed them very well). Fixed!

src/views/main.js
```diff
@@ -52,6 +52,49 @@ const mainLayout = data => `
...
+  aria-expanded='false'
+  aria-haspopup='true'
+  class='user-menu-button'
+  title='${getMessage('menu-button-title')}'
+  >
+    <img src='${data.fxaProfile?.avatar}' alt='${getMessage('menu-button-alt')}' />
+  </button>
+  <ul class='user-menu-list user-menu-popover' role='menu' hidden>
+    <li role='menuitem' tabindex='1'>
+      <a href='${AppConstants.FXA_SETTINGS_URL}' target='_blank' class='user-menu-header'>
```
flozia, line 67: We were asked to login again before because I just linked to FxA prod.

flozia (commented):
I addressed the outstanding issues @toufali we discussed:
- The users FxA avatar can be fetched locally as well
- Following the link to FxA does not require reauthentication
- Moved the user menu from its own partial to the `main` layout

src/app.js
```diff
@@ -36,10 +36,26 @@ async function getRedisStore () {
...
+  "'self'"
+]
+
+if (AppConstants.FXA_ENABLED) {
+  const fxaSrc = new URL(AppConstants.OAUTH_PROFILE_URI).origin
+  imgSrc.push(fxaSrc)
+
+  app.use(
+    helmet.crossOriginEmbedderPolicy({
+      policy: 'credentialless'
```
toufali, line 49: I spent some time trying to figure this out.  `credentialless` is "experiemental" and [not supported by Firefox or Safari](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cross-Origin-Embedder-Policy).  Locally it _did_ work for me in Firefox, but Safari sure enough did not display the avatar.

I was wondering why the avatar works in the legacy site, but not the new version.  I believe it's because the new version loads ALL Helmet modules including crossOriginEmbedderPolicy which defaults to `require-corp`, which you overwrite to `credentialless`.  In the legacy site, we only opt-in to `contentSecurityPolicy` and `referrerPolicy`.  `crossOriginEmbedderPolicy` is unset and the avatar works on all browsers.

I believe @groovecoder set up Helmet on the legacy site and I'm curious on his take of what Helmet modules we should include and what we might omit – hoping `crossOriginEmbedderPolicy` is one we can do without for now.  If not, we'll have to see how this avatar behaves on Safari in production with `crossOriginEmbedderPolicy` enabled – maybe it's a non-issue, but I'm pessimistic on this.  We could also raise the issue with FxA to see if others have the avatar working with `crossOriginEmbedderPolicy` set to the supported default `require-corp`.

src/views/main.js
```diff
@@ -65,6 +108,7 @@ const mainHeader = data => `
         </svg>
       </button>
     </li>
+    ${userMenu(data)}
```
toufali, line 111: I originally used the `<menu>` element above to hold both the nav button and the user menu.  So, as intended, this would look more like this:
```js
  <menu>
    <li>
      <button class='nav-toggle'>
        <svg xmlns='http://www.w3.org/2000/svg' viewBox='0 0 10 8' width='20'>
          <path d='M1 1h8M1 4h8M1 7h8' stroke='#000' stroke-width='1' stroke-linecap='round'/>
        </svg>
      </button>
    </li>
    <li>${userMenu(data)}</li>
  </menu>
```
As we discussed though, I may have gotten too caught up in semantics; `<menu>` is probably be more appropriate for your user menu.  That might look a bit more like this (note that I _think_ by using `<menu>` you can avoid all the `role='menuitem'` attributes?):
```js
const userMenu = data => `
<div class='user-menu-wrapper' tabindex='-1'>
  <button
  aria-expanded='false'
  aria-haspopup='true'
  class='user-menu-button'
  title='${getMessage('menu-button-title')}'
  >
    <img src='${data.fxaProfile?.avatar}' alt='${getMessage('menu-button-alt')}' />
  </button>
  <menu class='user-menu-list user-menu-popover' hidden>
    <li tabindex='1'>...</li>
  </menu>
</div>

...

  <ul>
    <li>
      <button class='nav-toggle'>
        <svg xmlns='http://www.w3.org/2000/svg' viewBox='0 0 10 8' width='20'>
          <path d='M1 1h8M1 4h8M1 7h8' stroke='#000' stroke-width='1' stroke-linecap='round'/>
        </svg>
      </button>
    </li>
    <li>${userMenu(data)}</li>
  </ul>
```

And finally, we could probably simplify even further by removing the top level list altogether, since it's not really a list and the items are not very similar... Taking the above and modifying it one last time:

```js
  <div>
    <button class='nav-toggle'>
      <svg xmlns='http://www.w3.org/2000/svg' viewBox='0 0 10 8' width='20'>
        <path d='M1 1h8M1 4h8M1 7h8' stroke='#000' stroke-width='1' stroke-linecap='round'/>
      </svg>
    </button>
    ${userMenu(data)}
  </div>
```

Hope my thoughts are clear, happy to chat in realtime if you have additional thoughts!

src/client/css/user-menu.css
```diff
@@ -0,0 +1,109 @@
...
+  gap: var(--padding-sm);
+  padding: calc(var(--padding-sm) * 1.5) calc(var(--padding-md) * 1.5);
+  text-decoration: none;
+}
+
+.user-menu-subtitle {
+  padding: 0;
+}
+
+.user-menu-header:hover,
```
toufali, line 94: Per discussion with Zachery, did we decide to remove the hover over the account header link to remain consistent with Relay?  I think removing the hover over the account link also makes more sense because it is more clearly an external link 🤷

toufali (commented):
Almost there @flozia!  I think the only real blocker is Helmet's `crossOriginEmbedderPolicy`, which I believe is currently breaking the FxA avatar in Safari.  It would be nice to have @groovecoder's thoughts since he put in place the original, but I believe we can just omit `crossOriginEmbedderPolicy` until the directive we need is better supported.  To do this, we would specifically opt-in to the policies we want: https://github.com/helmetjs/helmet#how-it-works

cc @mansaj

src/views/main.js
flozia: I agree that the `menu` tag as a wrapper for both — the `nav` and `userMenu` — does not make too much sense. I applied the adjustments proposed by you and use `<menu>` for the user menu.

Just as an interesting note: `ul` vs `menu` only makes a difference syntactically and not semantically when interpreted by the browser ([MDN: `<menu>` Usage notes](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/menu#usage_notes)) anymore. The tag `menuitem` has been deprecated.

toufali (approved):
@flozia the updates work great – tested on Safari as well.

One item I forgot to mention in my last review is that the active state of the avatar button remains visible when I believe it should be hidden.  To replicate:
click the avatar button and then click outside the menu – the avatar button auto-highlights itself.  I believe it's related to tab-order, but not sure if it's intentional/desirable?

Lastly, there's a lot of literature about misuse of ARIA roles:
https://adrianroselli.com/2017/10/dont-use-aria-menu-roles-for-site-nav.html
https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/navigation_role

TL:DR from the articles above, we might consider following the [MDN example](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/navigation_role#examples), removing `role='menuitem'` from each `<li>` and instead using a single `role="navigation"` on the parent of the menu list and `aria-label` describing the navigation landmark as "account navigation" or similar.  As the article states, we could also consider using a second `<nav>` since this implies `role="navigation"` automatically.

Neither of the above are blockers and I've held this PR up enough, so I'm marking it approved and defer to you on any changes you might add relating to the above.  Thanks again for the stellar work you've done in such a short time!

flozia:
@toufali Thank you for reviewing the changes and your suggestions.

> One item I forgot to mention in my last review is that the active state of the avatar button remains visible when I believe it should be hidden.

That is intentional. Would you prefer setting the focus on another element or no specific element at all? I’m actively setting the focus on the trigger after the menu gets hidden `onBlur` to support orientation as suggested by [W3C Focus Order 2.4.3](https://www.w3.org/TR/UNDERSTANDING-WCAG20/navigation-mechanisms-focus-order.html): “[…] When the dialog is dismissed, focus returns to the button or the element following the button.”

> … and instead using a single role="navigation" on the parent of the menu list and aria-label describing the navigation landmark as "account navigation" or similar. As the article states, we could also consider using a second <nav> since this implies role="navigation" automatically.

That’s a good idea. I added the role `navigation` and will add the label description in a separate PR since this will require a new localized string and we already have approval for the ones that were introduced here.

toufali:
> That is intentional. Would you prefer setting the focus on another element or no specific element at all? I’m actively setting the focus on the trigger after the menu gets hidden `onBlur` to support orientation as suggested by [W3C Focus Order 2.4.3](https://www.w3.org/TR/UNDERSTANDING-WCAG20/navigation-mechanisms-focus-order.html): “[…] When the dialog is dismissed, focus returns to the button or the element following the button.”

Thanks for the reference @flozia !  I think no specific element focus would be less confusing for pointer/touch users, especially for the case where I click on another button ("Resolved breaches" for example) and see the user-menu focus outline light up at the same time!  

It's not so much the focus, but the highlight/outline that confuses me.  If you were to rework to maintain focus but lose the highlight on click-outside, I'd be happy with that too.  Really, if a user is clicking outside they're not using a keyboard anyway?

PR #2757, opened 2023-01-11
MNTOR-977: Migrate breach resolution data (step 1)

mansaj:
* A script that executes once, to convert all `subscriber.breaches_resolved` to `subscriber.breaches_resolution`.
* An additional PR will be needed to deprecate the column "breaches_resolved." That PR won't be worked on until V2 is rolled out and we are certain that the new column is sufficient and error free. The old column is essentially our fail-over strategy at the moment
* Some refactor of shared breaches util functions


src/scripts/convertBreachResolutions.js
```diff
@@ -0,0 +1,97 @@
...
+let subscribersArr = []
+
+// load all breaches for ref
+const allBreaches = await getAllBreachesFromDb()
+if (allBreaches && allBreaches.length > 0) console.log('breaches loaded successfully! ', allBreaches.length)
+
+// find all subscribers who resolved any breaches in the past, convert those
+// records into the new v2 format
+do {
+  console.log(`Converting breadches_resolved to breach_resolution - start: ${offset} limit: ${LIMIT}`)
```
toufali, line 26: nit:
```suggestion
  console.log(`Converting breaches_resolved to breach_resolution - start: ${offset} limit: ${LIMIT}`)
```

src/scripts/convertBreachResolutions.js
```diff
@@ -0,0 +1,97 @@
...
+    .whereNotNull('breaches_resolved')
+    .limit(LIMIT)
+    .offset(offset)
+
+  console.log(`Loaded # of subscribers: ${subscribersArr.length}`)
+
+  for (const subscriber of subscribersArr) {
+    const { breaches_resolved: v1, breach_resolution: v2 } = subscriber
+    console.debug({ v1 })
+    console.debug({ v2 })
```
toufali, line 39: Testing this locally, `v2` is coming up null for me...

src/scripts/convertBreachResolutions.js
```diff
@@ -0,0 +1,97 @@
...
+        console.debug({ subBreach })
+
+        if (!subBreach || !subBreach.DataClasses) {
+          console.warn(`SKIP: Cannot find subscribers breach and data types - recency: ${recencyIndex} email: ${email}`)
+          continue
+        }
+
+        // if email does not exist in v2, we need to add it to the object
+        // format: {email: { recencyIndex: { isResolved: true, resolutionsChecked: [DataTypes]}}}
+        if (!v2[email]) {
```
toufali, line 64: Because `v2` is null for me locally, the script fails at this point.  We could add optional chaining `?.` but I'm not sure that fixes the root issue?  Should `v2` first be instantiated as an object maybe?

toufali (commented):
This is complicated!  Thanks for diving in.  I tried to run the script locally but ran into issues – see notes below.  I can try again after notes resolved, or indication on what I might need to do on my end to fix/run.

Additional question – are we looking to get away from the "Recency Index" naming convention?  I think we deduced that this is essentially a breach ID – is there a better name we can use?

src/scripts/convertBreachResolutions.js
mansaj: v2 can be null, it's expected

src/scripts/convertBreachResolutions.js
mansaj: See comment above. We don't want to init, we should let it fail. I'll add another check here for `!v2 ||`

mansaj: Oh nvm I see what you are saying

mansaj:
> This is complicated! Thanks for diving in. I tried to run the script locally but ran into issues – see notes below. I can try again after notes resolved, or indication on what I might need to do on my end to fix/run.
> 
> Additional question – are we looking to get away from the "Recency Index" naming convention? I think we deduced that this is essentially a breach ID – is there a better name we can use?

Recency index isn't addressed in this PR. We are still using RI here. With https://github.com/mozilla/blurts-server/pull/2766 we will start to address the changes from RI to breach ID

toufali (approved):
With your updates, this now seems to work as expected for me locally.  Note my Slack question on storing breaches, which I understand could be covered under separate PR.

Also note that this code is complex and it would be great to have a 3rd set of eyes on it to test another local case.  Approving as to not block though!

PR #2759, opened 2023-01-11
MNTOR-1101 run V2 unit tests using Github Actions

rhelmer:
Run V2 unit tests using GitHub Actions.

### DIFF
--- a/.github/workflows/unittests.yaml
+++ b/.github/workflows/unittests.yaml
@@ -1,0 +1,20 @@
+name: Unit Tests
+
+on: [push]
+
+jobs:
+  build:
+
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v3
+      - name: Use Node.js
+        uses: actions/setup-node@v3
+        with:
+          node-version: '18.12.x'
+      - run: npm ci --workspace=src
+      - run: npm ci
+      - run: cp .env-dist .env
+      - run: npm run build --if-present --workspace=src
+      - run: npm test --workspace=src
--- a/package-lock.json
+++ b/package-lock.json
@@ -75,20 +75,22 @@
       }
     },
     "node_modules/@babel/code-frame": {
-      "version": "7.16.7",
+      "version": "7.18.6",
+      "resolved": "https://registry.npmjs.org/@babel/code-frame/-/code-frame-7.18.6.tgz",
+      "integrity": "sha512-TDCmlK5eOvH+eH7cdAFlNXeVJqWIQ7gW9tY1GJIpUtFb6CmjVyq2VM3u71bOyR8CRihcCgMUYoDNyLXao3+70Q==",
       "dev": true,
-      "license": "MIT",
       "dependencies": {
-        "@babel/highlight": "^7.16.7"
+        "@babel/highlight": "^7.18.6"
       },
       "engines": {
         "node": ">=6.9.0"
       }
     },
     "node_modules/@babel/compat-data": {
-      "version": "7.18.5",
+      "version": "7.20.10",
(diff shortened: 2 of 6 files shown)


rhelmer:
Here is my understanding of the Jest+ES situation, based on reviewing the docs and issue below:

Jest has some issues with ES modules, so this PR uses Babel to transpile. The alternative is to make some ugly modifications to the test code and enable experimental support in node:

- `https://jestjs.io/docs/ecmascript-modules`
- `https://github.com/facebook/jest/issues/9430`

I feel like using babel for now and then just removing it as node and jest stabilize would be easiest, but open to trying to go pure ES now if we want to. Note that it's not just enabling the experimental ES support in node, you also need to change the jest calls in the unit test code (per the docs above), e.g. `jest.unstable_mockModule`, and hoisting doesn't work yet with ES so you need to make that call first then do a dynamic `import (const { myThing } = await import('myModule')` - that's what I mean specifically by modifications to the test code.

Using Babel allows us to write the tests as ES and it gets transpiled to CJS which Jest understands, so you can write unit tests the normal way:

```js
import { myThing } from 'myModule'
jest.mock('myModule') // myThing is now automatically mocked
```

toufali (commented):
Thanks for explaining the problem, and agreed that the workaround using `jest.unstable_mockModule` seems awful!  I'm mostly on board with the Babel workaround, but a couple thoughts:

- Should we be concerned with Jest testing babel-transformed code, but in production the code is not transformed?  Wouldn't it be better to use a test-runner that tests the actual (un-transformed) code?
- [This blurb](https://jestjs.io/docs/ecmascript-modules#module-mocking-in-esm) seems to suggest there's a block for Jest mocks to ever work nicely with ESM.  It makes me wonder if Jest mocks will ever work without dynamic imports...  If so, simply removing Babel in the future _without_ code change may not be realistic?
- Is the ability to use Mocks the main reason for this work-around?  If so, maybe we can just avoid their use?  I've read mocks can sometimes suggest code-smell!
- With all the above, should we considered a solution like [esmock](https://www.npmjs.com/package/esmock)?  It could be used with Jest, or maybe better (as the examples show) using the native Node 18+ test runner?

rhelmer:
> > Do you have any references I can read?
> 
> Yep, this is what I got into yesterday: https://dev.to/kettanaito/when-should-i-not-use-mocks-in-testing-544e https://chemaclass.medium.com/to-mock-or-not-to-mock-af995072b22e https://medium.com/javascript-scene/mocking-is-a-code-smell-944a70c90a6a
> 
> Interesting to note the slight contradictions in the articles with regard to unit test mocks. I think the big take-away will always be "it depends".

Thanks. I think it's worth digging into this more, right now we have a bunch of existing test code, and our coverage is quite low right now - I think it's worth continuing with mocking but we should consider whether there are better approaches.

I think in particular we'd benefit from end-to-end testing using Playwright or similar, which i know @mozrokafor already did some work on. I don't think the existing test make excessive use of mocking, at least not the few that I have ported over so far (email-utils and the settings controller).

rhelmer:
@toufali OK here's a different approach, using the minimal test runner [Ava](https://github.com/avajs/ava) which is pure ES.

The API isn't quite mocha compatible but it's similar. I've ported over the tests that @mansaj added in `./src/*.test.js`, neither of these use mocks, so I don't think we need to worry about introducing `esmock` or similar quite yet.

toufali:
Oh wow, this update looks fantastic and runs super fast!  Awesome work – I appreciate the extra diligence here 🙌

toufali (approved):

mansaj (approved):
There were some more tests added into `/utils` yesterday just FYI

rhelmer:
> There were some more tests added into `/utils` yesterday just FYI

Thanks, I'll rebase and convert those over as well.

PR #2761, opened 2023-01-11
Pass multiple partials to main

flozia:
I was wondering how we would handle creating partials within our current architecture best. Let’s assume we would like to create partials that:
- are consumed by other partials.
- share data between them.
- have their own stylesheets.

What are your thoughts on this approach or is there already another way we can do this?


src/views/main.js
```diff
@@ -22,7 +23,13 @@ const mainLayout = data => `
     <link rel="preload" href="/fonts/Metropolis-Bold.woff2" as="font" type="font/woff2" crossorigin>
     <link rel="preload" href="/fonts/Inter-Regular-latin.woff2" as="font" type="font/woff2" crossorigin>
     <link rel='stylesheet' href='/css/index.css' type='text/css'>
-    <link rel='stylesheet' href='/css/partials/${data.partial.name}.css' type='text/css'>
+    ${hasPartials(data)
+      ? data.partials.map(partial => (
+          `<link rel='stylesheet' href='/css/partials/${partial.name}.css' type='text/css'>`
+        )).join()
+      : ''
+    }
```
flozia, line 31: Load the necessary stylesheets for the partials that we would like to use.

toufali:
Recapping our discussion:
- one can simply import additional partials or more atomic includes.  
- Partials might/should be thought of as a full page template.  Smaller templates can simply be imported into the larger template
- We might consider minor refactor to use a better folder structure/naming convention to make this more clear
- A mechanism for loading additional stylesheets might still be necessary.  Perhaps we might consider using @import within the full page template stylesheet?

src/views/main.js
rhelmer: Not blocking, but maybe we should consider doing this for JS too... Am I correct in understanding all the client-side JS gets loaded by all pages currently?

toufali: @rhelmer that is correct – I'm keeping an eye on JS bundle size and it's still quite low: 4.35 kB transfered (for non-dev environments).  Our CSS tends to be a bit heavier, so splitting those files per-page is currently providing a better benefit.  Will keep _monitoring_!

PR #2762, opened 2023-01-11

toufali:
@flodolo this PR syncs the latest Refresh site strings with the localization branch.  

Question: When would you like the Heroku localization server to reflect the new Refresh site?  Going forward, the far majority of strings will be for Refresh.  Or do you want to keep the legacy site up until we've got the Refresh site on prod?


flodolo:
We should switch the dev server too. The old website hasn't received updates in a while, and we should make sure the new strings can be tested.

With that said, the number of new strings seems very small? I was expecting more (e.g. there's nothing from app.ftl)

toufali:
I will make that server switch 👍 

RE amount of strings -- I had the same thought, but it looks like we did a sync before the holiday break: https://github.com/mozilla/blurts-server/pull/2721

flodolo:
> I will make that server switch 👍
> 
> RE amount of strings -- I had the same thought, but it looks like we did a sync before the holiday break: #2721

I realized I had 2 PRs reviewed but they were not merged
https://github.com/mozilla/blurts-server/pulls?q=is%3Apr+is%3Aopen+reviewed-by%3A%40me

Should we wait for them before merging?

flodolo (approved):

toufali:
> 

Yes, our test pipeline was broken but has just come back up, so I've gone ahead and merged the older of the 2 PRs you suggested.  We still have work on the other PR and will merge it likely next week.

toufali:
@flodolo the localization server should now reflect the WIP Refresh site.

PR #2743, opened 2023-01-01
MNTOR-1004, MNTOR-1004 - backend settings screen for v2 and refactor email utility

rhelmer:
This is still a work-in-progress, need to finish the following:

- [ ] add unit tests (depends on https://github.com/mozilla/blurts-server/pull/2759)
- [ ] resolve FIXME and TODOs
- [x] implement CSRF (depends on https://github.com/mozilla/blurts-server/pull/2751)

PR #2763, opened 2023-01-13
MNTOR-983, MNTOR-1057 settings v2 frontend and migrate verification email

rhelmer:
This covers the following issues:

- migrate email to new platform MNTOR-1057
- frontend work for v2 settings MNTOR-983

It depends on the backend changes in https://github.com/mozilla/blurts-server/pull/2743

### DIFF
--- a/src/views/email-2022.js
+++ b/src/views/email-2022.js
@@ -1,0 +1,172 @@
+import AppConstants from '../app-constants.js'
+import { getMessage } from '../utils/fluent.js'
+
+const emailHeader = (data) => `
+<tr class="logo">
+  <td height="100"
+    style="height: 100px; background: #F9F9FA url('${
+      AppConstants.SERVER_URL
+    }/img/email_images/monitor-logo-bg-f9f9fa.png') no-repeat 50%; background-size: 240px 50px;">
+  </td>
+</tr>
+<tr class="header">
+  <td>
+    <table style="width: 100%; height: 331px; background-color: #321C64; color: white; text-align: left;"
+      role="presentation" border="0" cellpadding="0" cellspacing="0">
+      <tr>
+        <td>
+          <h1>${getMessage(data.heading)}</h1>
+          <p>${getMessage(data.subhead)}</p>
+        </td>
+        <td class="header-image" style="width: 50%; background-color: #321C64; vertical-align: bottom;">
+          <img src="${
+            AppConstants.SERVER_URL
+          }/img/email_images/person-at-desk.png" width="476" height="331"
+            style="display: block; max-width: 100%; margin-left: auto; object-fit: cover; object-position: left;"
+            alt="">
+        </td>
+      </tr>
+    </table>
+  </td>
+</tr>
+`
+
+const emailFooter = (data) => `
+<tr class="footer">
+  <td style="border-top: 1px solid #ddd; padding: 24px 0; background: #F9F9FA;">
+    <p>${getEmailFooterCopy(data)}</p>
+    <p>${getMessage('email-2022-hibp-attribution',
+      { 'hibp-link-attr': 'href="https://haveibeenpwned.com/" rel="noopener"'
+    })}
+    </p>
+    <img src="${
+      AppConstants.SERVER_URL
+    }/img/email_images/mozilla-logo-bw-rgb.png" width="130px"
+      style="display: block; margin: 24px auto 0;" alt="Mozilla">
+    <p>
+      2 Harrison St. #175, San Francisco, California 94105 USA
+    </p>
+    <p>
+      <a
+        href="https://www.mozilla.org/about/legal?utm_source=fx-monitor&utm_medium=email&utm_campaign=${
+          data.utmCampaign
+        }&utm_content=email-footer-link">${getMessage('legal')}</a> • <a
+        href="https://www.mozilla.org/privacy/firefox-monitor?utm_source=fx-monitor&utm_medium=email&utm_campaign=${
+          data.utmCampaign
+        }&utm_content=email-footer-link">${getMessage('terms-and-privacy')}</a>
+    </p>
+  </td>
+</tr>
+`
+function getEmailFooterCopy (data) {
+  let faqLink = getMessage(
+    'frequently-asked-questions'
+  )
+  faqLink = `<a href="https://support.mozilla.org/kb/firefox-monitor-faq">${faqLink}</a>`
+
+  if (!(data.whichPartial === 'email_partials/email-monthly-unresolved')) {
+    return getMessage('email-verify-footer-copy', {
+      faqLink
+    })
+  }
+
+  const unsubUrl = data.unsubscribeUrl
+  const unsubLinkText = getMessage('email-unsub-link')
+  const unsubLink = `<a href="${unsubUrl}">${unsubLinkText}</a>`
+
+  const localizedFooterCopy = getMessage(
+    'email-footer-blurb',
+    {
+      unsubLink,
+      faqLink
+    }
+  )
+
+  return localizedFooterCopy
+}
+
+const verifyPartial = (data) => `
+<tr>
+  <td style="color: black; background: #F9F9FA; padding: 36px 0 24px;">
+    <p>${getMessage('email-verify-simply-click')}</p>
+    <a href="${data.ctaHref}"
+      style="display: inline-block; margin: auto; color: white; background-color: #0060DF; border-radius: 4px; padding: 12px 24px; margin: 24px 0">${getMessage(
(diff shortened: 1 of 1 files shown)


rhelmer:
@flozia a few things are missing in V2 that we need to implement this, that aren't explicitly covered in figma:

- error messages
  - we can either redirect to an error page (which is what V1 does), or we can show some kind of error message using JS: 
image
  - see [src/client/js/settings.js](https://github.com/mozilla/blurts-server/pull/2763/files#diff-790e0d9a5e177ba94933197ab52bc546ed75f9a5026f8e306572cc0b02dc3a0c) for where the client is `fetch`ing the API (and not handling the exceptions right now just logging them to the console)
- success messages
  - I didn't find this in figma, but I see that V1 (current production on monitor.firefox.com) shows this using JS: 
image
- image formats
  - I exported the images in use here from Figma using PNG, is SVG preferred?

locales/en/settings.ftl
```diff
@@ -0,0 +1,28 @@
...
+
+comm-opt-0 = Email me if one of my email addresses below appears in a data breach.
+comm-opt-1 = Send all breach alerts to { $primaryEmail }
+
+send-verification = Send Verification Link
+add-new-email = Add a new email address
+email-verification-required = Email verification required
+settings-title = { -product-short-name } Settings
+monitored-email-addresses = Monitored email addresses
+email-address-limit = Your account includes monitoring of up to { $limit } emails
```
flodolo, line 23: This should use the plural structure (like the first string in this file)

flodolo (changes requested):
See the linter results https://github.com/mozilla/blurts-server/pull/2763/checks

More important: what's the reasoning behind splitting these strings into a separate file? This will require retranslating them across all locales.

If we do this, we should relook at the IDs that we used, and use better ones (providing more context on how/where the string is used).

toufali:
> what's the reasoning behind splitting these strings into a separate file?

Hi @flodolo – as part of this Refresh release, we're re-architecting/refactoring.  The strings @rhelmer moved over are specific only to the settings page, and we have been creating a new .ftl for each page in the site (~5 total expected for Refresh).

The app.ftl file has become very large and hard to navigate – We're hoping to reserve this page for site-wide strings only. I was under the impression that .ftl files concat and that moving strings between files was ok, but it sounds like I was wrong!  

For this case is it ok to separate into a more logical file structure?  Monitor is expected to continue growing, and I'd like to set up good practice in the Refresh release, so future features can benefit.

flodolo:
> The app.ftl file has become very large and hard to navigate – We're hoping to reserve this page for site-wide strings only. I was under the impression that .ftl files concat and that moving strings between files was ok, but it sounds like I was wrong!

No concatenation ;-) 
https://pontoon.mozilla.org/it/firefox-monitor-website/

I'm OK with splitting things up in smaller files, as long as it's done purposefully. Pontoon will show a match in translation memory, so it doesn't need to be translated from scratch. But we should also take the opportunity to do a better job with IDs, since we're basically creating new strings.

I'm heading offline for the week, but I'll try to provide more comments in the next days of what I mean.

locales/en/settings.ftl
```diff
@@ -0,0 +1,28 @@
+# This string is shown beneath each of the user’s email addresses to indicate
+# how many known breaches that email address was found in.
+appears-in-x-breaches =
+  { $breachCount ->
+        [one] Appears in { $breachCount } known breach.
+       *[other] Appears in { $breachCount } known breaches.
+  }
+
+resend-verification = Resend verification email
```
flodolo, line 9: I understand these come from existing strings, but we didn't do a good job here, and we can use this work to improve things.

IDs should provide context for localizers, if there's no comment doing it. For example: is this a button, a link, something else? Putting more information, and using a more specific ID based on the context would help localizers, and discourage reuse in different contexts.

In this case I would go with `settings-resend-verification-link`.

This applies to all the strings in this file, in particular to very undescriptive ones like `menu`.
`email-verification-required`: is it a title, a message? These might require different style in different locales.

src/client/js/settings.js
```diff
@@ -0,0 +1,164 @@
...
+      method: 'POST',
+      body: JSON.stringify({ email })
+    })
+
+    if (!response || !response.ok) {
+      throw new Error(`sending verification email failed: ${response.error}`)
+    }
+
+    const addEmailDialogContent = document.getElementById('add-email-modal-content')
+    addEmailDialogContent.textContent = `Verify the link sent to ${email} to add it to Firefox Monitor. Manage all email addresses in Settings.`
```
pdehaan, line 86: Q: Does this need to be localized?

src/utils/email.test.js
```diff
@@ -0,0 +1,234 @@
+'use strict'
+
+import {
+  TEST_SUBSCRIBERS,
+  TEST_EMAIL_ADDRESSES
+} from '../db/seeds/test_subscribers'
+import { initFluentBundles } from './fluent.js'
+
+import AppConstants from '../app-constants'
+import EmailUtils from './email'
```
pdehaan, line 10: Should these have .js extensions for consistency?

src/views/email-2022.js
```diff
@@ -0,0 +1,172 @@
...
+  </td>
+</tr>
+`
+function getEmailFooterCopy (data) {
+  let faqLink = getMessage(
+    'frequently-asked-questions'
+  )
+  faqLink = `<a href="https://support.mozilla.org/kb/firefox-monitor-faq">${faqLink}</a>`
+
+  if (!(data.whichPartial === 'email_partials/email-monthly-unresolved')) {
```
pdehaan, line 67: Why not this?
```suggestion
  if (data.whichPartial !== 'email_partials/email-monthly-unresolved') {
```

src/views/partials/settings.js
```diff
@@ -1,5 +1,93 @@
...
+}
+
+export const settings = (data) => `
+<div id="settings" class="parent" data-csrf-token="${data.csrfToken}">
+  <div class="child-left">
+    <h2>${getMessage('settings-title')} </h2>
+  </div>
+  <div class="child-right">
+    <section>
+      <h3>Breach alert preferences</h3>
```
pdehaan, line 44: Q: Localized?

src/views/partials/settings.js
```diff
@@ -1,5 +1,93 @@
...
+        <div id="add-email-modal-controls">
+          <label>${getMessage('email-address')}<input id="email" type="text"></label>
+          <button id="send-verification" class="monitor-button">${getMessage('send-verification')}</button>
+        </div>
+      </dialog>
+    </section>
+    <hr>
+    <section>
+      <h3>${getMessage('deactivate-account-title')}</h3>
+      <div>${getMessage('deactivate-account')}.</div>
```
pdehaan, line 88: Q: Should the period be within the translated string?

locales/en/settings.ftl
```diff
@@ -0,0 +1,28 @@
...
+comm-opt-0 = Email me if one of my email addresses below appears in a data breach.
+comm-opt-1 = Send all breach alerts to { $primaryEmail }
+
+send-verification = Send Verification Link
+add-new-email = Add a new email address
+email-verification-required = Email verification required
+settings-title = { -product-short-name } Settings
+monitored-email-addresses = Monitored email addresses
+email-address-limit = Your account includes monitoring of up to { $limit } emails
+add-another-email = Add a new email address to see if it's been involved in a breach
```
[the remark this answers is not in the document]
flodolo: https://github.com/mozilla/blurts-server/pull/2763/checks

locales/en/settings.ftl
```diff
@@ -0,0 +1,28 @@
...
+email-address = Email Address
+deactivate-account-title = Deactivate account
+deactivate-account = You can deactivate Firefox Monitor by deleting your Firefox account
```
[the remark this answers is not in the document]
flodolo: Indeed, both Firefox Monitor and Firefox account should use the corresponding term.

src/client/js/settings.js
```diff
@@ -0,0 +1,164 @@
...
+      headers: {
+        'Content-Type': 'application/json',
+        'x-csrf-token': csrfToken
+      },
+      mode: 'same-origin',
+      method: 'POST',
+      body: JSON.stringify({ email })
+    })
+
+    if (!response || !response.ok) {
```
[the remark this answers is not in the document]
rhelmer: This is from the existing front-end code, but I agree it doesn't make much sense as-is right now, will correct this.

rhelmer:
This is still a draft PR, let's move this work over to https://github.com/mozilla/blurts-server/pull/2765 instead.

PR #2764, opened 2023-01-13
MNTOR-1045: Circle graph component

flozia:
# Description

This PR introduces a chart that we can use on the breaches dashboards. The `pieChart` function takes data with absolute `count` values and renders an SVG. It can variety of things:

- Take the shape of a pie or donut chart.
- Show the percentage of one specific item in the center of the chart.
- Scales responsively and can be styled via CSS.
- Show a list of labels.

**Schema**
```
{
    title: "string",
    data: [
      {
        key: "string",
        name: "string",
        count: "int",
        color: "string",
       // First item in `data` that has this value set to `true`
       // will be shown in the center of the chart
        showPercentage: "boolean"
      }
    ],
    isDonut: "boolean"
 }
```

# References 
- Jira: [MNTOR-1045](https://mozilla-hub.atlassian.net/browse/MNTOR-1045)
- Figma: [Breaches dashboard](https://www.figma.com/file/gvyyqS9egsGjsH0LUDqT1m/Breach-Resolution-Flow-Design?node-id=396%3A15305&t=2grz4fqOB4UX7sNg-4)

# Screenshot

pie-charts-pr


src/views/includes/pie-chart.js
```diff
@@ -0,0 +1,133 @@
...
+  const { isDonut } = chartData
+  const innerRadius = isDonut ? 0.85 : 0
+
+  const total = chartData.data.reduce((acc, curr) => acc + curr.count, 0)
+  let sliceOffset = 0
+
+  return `
+    ${styles}
+    <figure class='pie-chart'>
+      <svg viewBox='0 0 1 1'>
```
flozia, line 109: Figuring out that an `SVG` with the size of `1` just scales perfectly fine (almost feels like it was made for it) and that we than can just use `Math.PI` for the circumference makes me unreasonably happy.

toufali (commented):
This is excellent @flozia!  A few notes to move this forward: 

- The graphs work as expected on Firefox and Chrome, but on Safari the percentages don't display and some of the slices look wonky (see screenshot).
- Is it possible to repurpose this as a client-side component with an internal function that accepts a data object and can update the graph dynamically?  I ask because on the Breaches page the user can select between email addresses via fetch – the data is re-hydrated without re-rendering the entire page.  Happy to discuss further if you already had a plan in mind for this functionality
- with the previous note, it might be nice to include some sort of animation between data states?  I imagined the pie slices expanding/contracting to their final state, but I haven't thought about how feasible or complex that might be to accomplish.  A simple fade-in between states might serve the purpose just as well – I'll leave it to you to explore!

Safari screenshot:
Screen Shot 2023-01-18 at 9 57 44 PM

toufali:
I was thinking about my second bullet above - client-side web component.  Instead of an internal function that a user of the component triggers, you could also consider using `<slot>` for a more declarative instantiation.  Pseudo code example (shown with custom attribute "percent" which can be any other flag):

```html
<circle-graph percent>
  <h2 slot="title">Breach data:</h2>
  <ul slot="legend">
    <li>passwords</li>
    <li>emails</li>
    <li>address</li>
  </ul>
</circle-graph>
```

Just a thought!

EMMLynch:
@toufali, is this a suggestion to @flozia? I'm just trying to understand if this is a requirement or a nice to have. If it's not needed for this Jira ticket, I would propose you create a new ticket to capture this so we don't lose it and move forward as is.

toufali:
> is this a suggestion to @flozia?

@EMMLynch This is a suggestion on how one might implement a requirement I listed above.  The requirement is that the component be built client-side to be able to respond to user interaction (email account switching).

flozia:
Thank you @toufali!

> The graphs work as expected on Firefox and Chrome, but on Safari the percentages don't display and some of the slices look wonky (see screenshot).

I fixed the wonky slices for Safari — I’ll still need to find out why the percentage label won’t show correctly.

> Is it possible to repurpose this as a client-side component with an internal function that accepts a data object and can update the graph dynamically? I ask because on the Breaches page the user can select between email addresses via fetch – the data is re-hydrated without re-rendering the entire page. Happy to discuss further if you already had a plan in mind for this functionality

Yes, sure! I just ported the implementation over to a web component. I like the declarative example posted. Using slots for the labels would be nice. Right now I created the component to calculate and render everything that is needed from the `data-chart` attribute. I’m hesitant to put too much effort into making decisions for the component's API right now if we only need it for two use cases for now. Would it be OK if we would refactor the component to match our needs better when we apply it where needed?

> with the previous note, it might be nice to include some sort of animation between data states? I imagined the pie slices expanding/contracting to their final state, but I haven't thought about how feasible or complex that might be to accomplish. A simple fade-in between states might serve the purpose just as well – I'll leave it to you to explore!

I added a simple fade-in/fade-out for the entire chart. This could be extended easily to fun transitions on single elements of the chart with some more time for experimentation.

src/client/js/app.js
```diff
@@ -1 +1,26 @@
...
+          key: 'unresolved',
+          name: 'Unresolved',
+          count: 5,
+          color: '#321c64'
+        }
+      ],
+      isDonut: true
+    }))
+  })
+})
```
flozia, line 26: Example of how the chart would be updated with new data on the client side.

src/client/js/components/circle-chart.js
```diff
@@ -0,0 +1,223 @@
...
+  return parseFloat((value / total).toFixed(3, 10))
+}
+
+const html = () => `
+  ${styles}
+  <figure class='circle-chart'></figure>
+`
+
+customElements.define('circle-chart', class extends HTMLElement {
+  static get observedAttributes () { return ['data-chart'] }
```
flozia, line 86: Only take the attribute `data-chart` which contains all of the necessary info to render the charts for our current use cases. Happy to discuss a declarative and more intuitive API, but with regard to our time constraints I’d like to do this when we actually had the chance to use them on the actual pages. Would this approach work for you @toufali?

src/views/partials/landing.js
```diff
@@ -5,10 +5,11 @@ export const landing = data => `
...
   ${chartTestData.map(data => `
     <circle-chart
-      class='chart-test'
-      data-chart='${JSON.stringify(data)}'
-    >
-    </circle-chart>
+      data='${JSON.stringify(data.data)}'
+      is-donut='${JSON.stringify(data.isDonut)}'
+      show-percentage-for='${data.showPercentageFor}'
+      title='${data.title}'
```
flozia, line 11: Moved some properties into their own attributes as proposed.

src/client/js/components/circle-chart.js
flozia: Updated the attributes that the custom component takes as discussed.

src/views/partials/landing.js
```diff
@@ -1,6 +1,18 @@
...
+import chartTestData from '../../../src/client/js/components/circle-chart-test-data.js'
 
 export const landing = data => `
+<div style='display: grid; grid-template-columns: 1fr 1fr 1fr; gap: var(--padding-lg);'>
+  ${chartTestData.map(data => `
+    <circle-chart
+      class='chart-test'
+      data='${JSON.stringify(data.data)}'
+      is-donut='${JSON.stringify(data.isDonut)}'
+      show-percentage-for='${data.showPercentageFor}'
```
toufali, line 11: The `show-percentage-for` attribute doesn't render correctly without `is-donut="true"`.  In the UX discussion, there was only ever a request for 2 states: donut with percent, and pie-chart.  This makes sense to me: If percent is not needed, the graph seems more legible as a pie-chart.

Suggestion: I think we should remove the `isDonut` attribute altogether and include the donut style as part of `showPercentFor`.  This leaves 3 attributes: `title`, `data`, and `show-percentage-for`.

Nit: we might reduce wordiness by using `show-percent-for`.

src/client/js/components/circle-chart-test-data.js
```diff
@@ -0,0 +1,250 @@
...
+      {
+        key: 'unresolved',
+        name: 'Unresolved',
+        count: 10,
+        color: '#321c64'
+      }
+    ],
+    showPercentageFor: 'resolved',
+    isDonut: true
+  },
```
toufali, line 20: Presumably we will be deleting this file and the associated test changes on the landing page?  Assuming yes, it would still be nice to keep a schema example in comment form within the component.  A usage example might be nice too, specifically demonstrating the `JSON.stringify` aspect of the data.

src/client/js/components/circle-chart.js
```diff
@@ -0,0 +1,244 @@
...
+    }
+
+    if (!this.data) {
+      return
+    }
+
+    this.total = this.data.reduce((acc, curr) => acc + curr.count, 0)
+
+    this.chartElement.classList.add('updating')
+    this.updateTimeout = setTimeout(() => {
```
toufali, line 229: I'm confused at the need for a timeout?  Is there a better way?  Happy to help here if needed.

src/client/js/components/circle-chart.js
```diff
@@ -0,0 +1,244 @@
...
+
+  updateChart () {
+    this.svg.innerHTML = `
+      ${this.composeCircles()}
+      ${this.createCircleLabel()}
+    `
+    this.labels.innerHTML = this.createChartLabels()
+  }
+
+  createOrUpdateChart () {
```
toufali, line 217: The naming of this function confused me at first – is this just an "init" function?  Would the built-in lifecycle callback `connectedCallback` be more appropriate?  Perhaps it might also negate the need for timeout discussed below?

src/client/js/components/circle-chart.js
```diff
@@ -0,0 +1,244 @@
...
+    transition: opacity ${CHART_UPDATE_DURATION * 0.5}ms ease;
+  }
+
+  .circle-chart.updating {
+    opacity: 0
+  }
+
+  .circle-chart-title {
+    display: block;
+    font-family: Inter, Inter-fallback, sans-serif;
```
toufali, line 26: I had issues with the font sizes when I attempted to add the component to the breaches page: the labels were larger than the heading.  Styles in web components are tricky – we can finesse this when we add it to the breaches page.

toufali (approved):
@flozia – fantastic work done on a tricky component!  Well done 👏 

Two minor changes and a couple comments, most of which we can address under separate work.  I'm marking this as approved since changes are non-blocking.

src/client/js/components/circle-chart.js
flozia: Oh. Yes, let’s look into this when adding the charts to the view they are intended for.

src/client/js/components/circle-chart.js
flozia: This method handles the logic around when to init the circles or when they only need to be updated. When the chart is only being updated and not created from scratch on `data` change it’s being called from here](https://github.com/mozilla/blurts-server/pull/2764/files#diff-c6fd04d9ec6e829967576bae19b241f7f063a5e1d10bf3537675c0cb15d53027R119). Moving it to `connectedCallback` would only render the chart once and not on data updates. Like the example update that currently occurs when clicking on one of the charts.

In the event of an update, the chart is not being recreated from scratch. That would make adding transitions for the chart parts possible in the future. For now, I only opted for a simple fade-in/out which is why I added it. Without the timeout, we would not be able to have a CSS transition. Happy to discuss other approaches as we have the opportunity to make it better when we know how it behaves for our use cases.

src/views/partials/landing.js
flozia: Updated!

flozia:
Thank you @toufali!

I added an example and schema for how to use the chart to the component. I hope my comments regarding the `createOrUpdateChart` method and the usage of `setTimeout` make sense. I’d go ahead and merge the component and address any follow-ups and issues that might come up when we add it to the dashboard if that is OK with you.

PR #2765, opened 2023-01-17
MNTOR-983: Settings page frontend

flozia:
# References: 
Jira: [MNTOR-983](https://mozilla-hub.atlassian.net/browse/MNTOR-983)
Figma: [Monitor settings page](https://www.figma.com/file/gvyyqS9egsGjsH0LUDqT1m/Breach-Resolution-Flow-Design?node-id=214%3A20758&t=BXLEdrjwcip3gbdZ-0) (with changes to the [email list style](https://www.figma.com/file/gvyyqS9egsGjsH0LUDqT1m?node-id=214:20759#347392821))

# Description

⚠️ Depends on the backend changes introduced by PR #2773

This PR adds the frontend parts for the settings page.
# Screenshot

monitor-settings-page

# How to test


src/client/js/settings.js
```diff
@@ -0,0 +1,156 @@
...
+      body: JSON.stringify({ email })
+    })
+
+    if (!response?.ok) {
+      throw new Error(`Sending verification email failed: ${response?.error}`)
+    }
+
+    const addEmailDialogContent = document.getElementById('js-settings-modal-content')
+    // TODO: Localize string below
+    addEmailDialogContent.textContent = `Verify the link sent to ${email} to add it to Firefox Monitor. Manage all email addresses in Settings.`
```
flozia, line 82: How should we best handle localized strings that are being set via JS — should we include `fluent`?

src/controllers/settings.js
```diff
@@ -3,7 +3,36 @@ import { settings } from '../views/partials/settings.js'
 
 function settingsPage (req, res) {
   const data = {
-    partial: settings
+    partial: settings,
+    // TODO: The entries below are for testing only and should be removed.
```
flozia, line 7: This data will be added by PR #2743 which introduces the backend changes for the settings page.

src/views/email-2022.js
```diff
@@ -0,0 +1,172 @@
+import AppConstants from '../app-constants.js'
```
flozia, line 1: Email template taken from the previous implementation.

src/client/js/settings.js
rhelmer: I think so, @toufali might have a better idea.

toufali: The `getMessage()` function in utils/fluent.js accepts an optional argument that should help cases like this.  Example:

https://github.com/mozilla/blurts-server/blob/fe9e0c31aa18bb0968311abaa4791136fbaeb1eb/src/views/partials/breaches.js#L75

https://github.com/mozilla/blurts-server/blob/fe9e0c31aa18bb0968311abaa4791136fbaeb1eb/locales/en/breaches.ftl#L10-L16


Here's another (possibly less complex) example where the string is translated into a constant rather than template:
https://github.com/mozilla/blurts-server/blob/fe9e0c31aa18bb0968311abaa4791136fbaeb1eb/src/views/partials/breaches.js#L37-L42

https://github.com/mozilla/blurts-server/blob/fe9e0c31aa18bb0968311abaa4791136fbaeb1eb/locales/en/breaches.ftl#L31-L33

locales/en/settings.ftl
```diff
@@ -0,0 +1,40 @@
...
+
+settings-alert-preferences-title = Breach alert preferences
+settings-alert-preferences-option-one = Send breach alerts to the affected email address
+settings-alert-preferences-option-two = Send all breach alerts to the primary email address
+
+## Monitored email addresses
+
+settings-email-label-primary = {$email} (primary)
+settings-email-list-title = Monitored email addresses
+settings-email-limit-info = Your account includes monitoring of up to {$limit} emails.
```
flodolo, line 15: This still needs a plural structure (like the one on line 22-27)

locales/en/settings.ftl
```diff
@@ -0,0 +1,40 @@
...
+settings-email-verification-callout = Email verification required
+settings-resend-email-verification = Resend verification email
```
flodolo, line 17:
```suggestion
settings-resend-email-verification-link = Resend verification email
```

locales/en/settings.ftl
```diff
@@ -0,0 +1,40 @@
...
+  {
+    $breachCount ->
+        [one] Appears in {$breachCount} known breach.
+       *[other] Appears in {$breachCount} known breaches.
+  }
+
+## Deactivate account
+
+settings-deactivate-account-title = Deactivate account
+settings-deactivate-account-info = You can deactivate Firefox Monitor by deleting your Firefox account.
```
flodolo, line 32:
```suggestion
settings-deactivate-account-info = You can deactivate {-product-short-name} by deleting your {-brand-fx-account}.
```

locales/en/settings.ftl
```diff
@@ -0,0 +1,40 @@
...
+settings-fxa-link-label = Go to Firefox Settings
```
flodolo, line 33:
```suggestion
settings-fxa-link-label = Go to {-brand-name} Settings
```

The naming of that term was pretty poor :-( (which brand?)

locales/en/settings.ftl
```diff
@@ -0,0 +1,40 @@
...
+
+## Add email dialog
+
+settings-email-dialog-title = Add another email address
+settings-add-email-text = Add a new email address to see if it’s been involved in a breach.
+settings-email-input-label = Email address
+settings-send-email-verification = Send verification link
```
flodolo, line 40:
```suggestion
settings-send-email-verification-button = Send verification link
```

flodolo (changes requested):
Thanks, the FTL file looks a lot better now, still a few things to fix.

src/client/js/settings.js
flozia: @toufali Thanks for the examples. My question here specifically was more related to: If we set localized strings from `JS` we’d need to include `fluent.js` in the frontend resources or is there another way we could do this?

locales/en/settings.ftl
flozia: Just saw that we also have {[-brand-firefox](https://github.com/mozilla/blurts-server/blob/main/locales/en/app.ftl#L805)}. :)

src/client/js/settings.js
toufali: Ah I see.  

I'm not familiar with using fluent.js on the client – wouldn't that require special endpoints or localization server, and might that introduce lag in showing a proper translation?  Would be interested to hear more of this approach.

The legacy site appears to creates the modal with translations on the server, and then pass additional translation strings as needed via XHR response after user interaction:

https://github.com/mozilla/blurts-server/blob/1d59e3383aed8d672dd2b130f1e30b940536f12f/public/js/resolve-breaches.js#L116-L157

Other options:
- use a data-attribute on the relevant tag (if it exists) on initial render
- include the translations in an SSR `<script>` tag (though I believe this pollutes global scope, so likely not a first choice without additional thought) :
```js
<script>
var l10n = {
         addEmailDialogContent: 'a translated string'
      }
</script>
```

- you could use an SSR `<template>` tag to store strings for use on client:
```html
<template>
  <p data-dialog-txt='translated string'></p>
  <!-- or -->
  <p data-dialog-txt>translated string</p>
</template>
```

- expanding on the previous, you could just use a template element for the entire structure you intend to render dynamically, not just translations...  This might play nicely with the lightbox/dialog web component forthcoming this sprint – i.e. the client web component is a shell/wrapper that loads the SSR template content 🤷

src/views/partials/settings.js
```diff
@@ -1,93 +1,178 @@
...
+  </li>
+`
+
+// Sort first by presence of `primary` key, then by email address.
+const getSortedEmails = emails => [...emails].sort((a, b) => (
+  a.primary
+    ? -1
+    : b.primary
+      ? 1
+      : 0 || a.email.localeCompare(b.email)
```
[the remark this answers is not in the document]
flozia: You are totally right — cleaned that up!

flodolo (approved):

toufali:
I see a few references to https://github.com/mozilla/blurts-server/pull/2743 but that PR is closed?  Is there an updated PR, and should I review that one first?

rhelmer:
> I see a few references to #2743 but that PR is closed? Is there an updated PR, and should I review that one first?

Sorry, I split this out to https://github.com/mozilla/blurts-server/pull/2773 (which is done except for the older tests that I am porting over and reviewing, we should land that and https://github.com/mozilla/blurts-server/pull/2772. I'll leave a comment in #2743 for posterity as well.

I'd suggest landing those first and then this current one (plus any small changes we need to properly pass the email verification template).

toufali:
> I split this out to #2773 (which is done except for the older tests that I am porting over and reviewing), to support this PR we need to land that and #2772. I'll leave a comment in #2743 for posterity as well.
> 
> I'd suggest landing those first and then this current one (plus any small changes we need to properly pass the email verification template).

Thanks @rhelmer!  #2772 is approved and I'll keep my eye out for #2773 when it's ready for review.  After both are merged I can test this PR on top of them 👍

src/views/partials/settings.js
```diff
@@ -1,5 +1,183 @@
...
+      </button>
+    </div>
+  </dialog>
+`
+
+export const settings = data => {
+  const { breachCounts, csrfToken, emails, limit } = data
+
+  return `
+    <div id='js-settings' class='settings' data-csrf-token='${csrfToken}'>
```
toufali, line 132: nit: since element IDs create global vars, should we avoid their use – at least on non-form elements?  If so, is this one avoidable?  The selector `.settings[data-csrf-token]` (or a similar variation) seems unique enough to me.

In the merged branch that I'm reviewing locally (prepared by @rhelmer) this ID is simply "settings", which is more of a concern as a global var.

src/views/partials/settings.js
```diff
@@ -1,5 +1,183 @@
...
+        <!-- Monitored email addresses -->
+        <section>
+          <h3 class='settings-section-title'>
+            ${getMessage('settings-email-list-title')}
+          </h3>
+          <p>${getMessage('settings-email-limit-info', { limit })}</p>
+
+          ${createEmailList(emails, breachCounts)}
+          <button
+            id='js-settings-add-email-opener'
```
toufali, line 155: See note about ID attribute global vars and if we should avoid for non-form elements.

src/views/partials/settings.js
```diff
@@ -1,5 +1,183 @@
...
+  <ul class='settings-email-list'>
+    ${getSortedEmails(emails)
+      .map(email => createEmailItem(email, breachCounts))
+      .join('')}
+  </ul>
+`
+
+const optionInput = (csrfToken, { isChecked, option }) => `
+  <input
+    checked='${isChecked}'
```
toufali, line 68: Be careful here, I believe `checked` == `checked="true"` == `checked="false"`.  So, having `checked="false"` is actually specifying the checkbox _is_ checked.

You probably need to use a conditional here:
```suggestion
    ${isChecked ? 'checked' : ''}
```

https://html.spec.whatwg.org/multipage/common-microsyntaxes.html#boolean-attributes

src/client/css/partials/settings.css
```diff
@@ -0,0 +1,158 @@
...
+
+.settings-section-title {
+  margin: var(--padding-sm) 0 var(--padding-md) 0;
+}
+
+.settings-content {
+  color: var(--gray-40);
+  font-size: 0.875rem;
+  flex: 2;
+}
```
toufali, line 24: Suggestion: make this page responsive across all viewports rather than just for mobile media-query.  The following style will wrap the settings page sooner, rather than waiting for the breakpoint:

```suggestion
.settings {
  display: flex;
  flex-wrap: wrap;
  gap: var(--padding-md) var(--padding-xl);
  padding: var(--padding-lg);
}

.settings-title,
.settings-section-title {
  color: var(--main-dark);
}

.settings-section-title {
  margin: var(--padding-sm) 0 var(--padding-md) 0;
}

.settings-content {
  color: var(--gray-40);
  font-size: 0.875rem;
  flex: 1 1 66%;
}
```

src/client/css/partials/settings.css
```diff
@@ -0,0 +1,158 @@
...
+  font-size: 2em;
+  left: var(--padding-sm);
+  line-height: 0.35em;
+  top: 50%;
+  transform: translateY(-50%);
+}
+
+.mobile .settings {
+  flex-direction: column;
+}
```
toufali, line 158: Per responsive comment above, this can be deleted if accepting suggestion:
```suggestion
```

toufali (changes requested):
Good first pass!  I'm rushed for time in this review, but I wanted to post thoughts early since it's overdue:

I think a major blocker is that the Breach alert preferences radio buttons don't reflect the state I set them at – i.e. if I check the first radio labeled "Send breach alerts to the affected email address" and then reload, it still displays the second radio as selected ("Send all breach alerts to primary").  This could be just a result of `checkbox="false"` issue (see inline comment below) – I haven't gone deeper into the user db settings than this.

A minor blocker is responsiveness across viewports between mobile and desktop.  See screenshot and suggestion inline.

Screen Shot 2023-01-28 at 4 47 59 PM 

_Page at this viewport looks broken, and user is unable to scroll horizontally to see content._

locales/en/settings.ftl
```diff
@@ -0,0 +1,45 @@
+# Settings page
```
flodolo, line 1: I haven't been strict about FTL files in this repository, but since this is a new file, I think it's worth cleaning up the formatting.

This is how the file would be formatted in mozilla-central:
* License header
* Spaces around message/term/variable references
* Mandatory comments for variables

Could you update this file accordingly? I'm attaching the updated FTL (renamed to .txt to be able to attach it here).

[settings.txt](https://github.com/mozilla/blurts-server/files/10529118/settings.txt)

I'd  like to enforce this style going forward for new strings.

src/client/js/settings.js
```diff
@@ -0,0 +1,160 @@
...
+        }
+      } catch (err) {
+        console.error(`Error: ${err}`)
+      }
+    })
+  }
+}
+
+const settingsResendEmailLinks = document.getElementsByClassName('js-settings-resend-email')
+if (settingsResendEmailLinks?.length) {
```
toufali, line 131: Minor note, non-blocking: instead of using a conditional to check each interactive element on the page, you could simply verify that this is the settings partial.  Each "main" template is created with the partial name as a data attribute, so you could do:
```js
const settingsPartial = document.querySelector("[data-partial='settings']")
```
followed by a single conditional check similar to:
```js
if (settingsPartial) init()
```
Where `init()` is a function to set up listeners, run-once code, etc.

In addition to feeling confident the correct partial and elements are loaded, you have access to the `<main>` wrapper around your partial structure, and can query within it (`settingsPartial.getElementsByClassName('js-settings-resend-email'`) instead of the entire document.  This might allow less verbose names?

If interested, see https://github.com/mozilla/blurts-server/blob/main/src/client/js/partials/breaches.js for example.

locales/en/settings.ftl
flozia: Thank you @flodolo for putting together example FTL — updated and noted for any new additions.

src/client/css/variables.css
```diff
@@ -12,12 +12,17 @@
   --h2-font: 700 1.75rem/1 metropolis, sans-serif;
   --h3-font: 700 1.125rem/1 metropolis, sans-serif;
   --h4-font: 700 1rem/1 metropolis, sans-serif;
+  --main-dark: #2e2c34;
```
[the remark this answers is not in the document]
flozia: Good callout. That probably sneaked it’s way in from the older styles. `Main Dark` [is being used in Figma](https://www.figma.com/file/gvyyqS9egsGjsH0LUDqT1m/Breach-Resolution-Flow-Design?node-id=272%3A17536&t=JKv9eigvfBIXIObU-4), but I’ll sync with UX and see if we can replace this.

src/client/js/settings.js
```diff
@@ -0,0 +1,160 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+const settingsAlertOptionsInputs = document.getElementsByClassName('js-settings-alert-options-input')
```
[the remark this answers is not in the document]
flozia: Adding  `js` communicates clearly that this class is not used for styling, but as a target for scripting — separating those reduces the chances that the classname is used for other purposes than originally intended. Also I like the extra context that descriptive classnames give. If we have another convention or would like to introduce one I’m happy to adopt it.

src/views/partials/settings.js
flozia: I agree on trying to avoid using `id`s. For entire pages I think it might be helpful to have something more descriptive than `.settings[data-csrf-token]`. I’ll add an identifier to the class list.

src/client/css/partials/settings.css
flozia: Thanks for finding the layout issue and providing a great fix with this suggestion! :)

src/client/js/settings.js
flozia: I like that suggestion — it also helps to grasp which dynamic elements are expected to be present easily.

flozia:
> I think a major blocker is that the Breach alert preferences radio buttons don't reflect the state I set them at […]

Fixed by commit 8f98008 — I didn’t pass the users communication preferences to the partial at all.

rhelmer (approved):

src/client/js/settings.js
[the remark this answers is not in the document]
toufali: Thanks for the context – I wasn't aware of that convention (which doesn't help either of our cases 😄 ).  

I've heard of using data-attributes for cases where an element needs to be selected but not styled, but I'm not sure if that's any better.  I usually resort to a specific-enough query selector on a stable parent and deal with any consequences later!

src/client/css/global.css
```diff
@@ -173,6 +180,12 @@ button.primary,
   background: var(--blue-50);
   border: 1px solid var(--blue-50);
   padding: var(--padding-md);
+  transition: background 150ms ease;
```
toufali, line 183: nit: I believe `ease` is the default value – we could likely delete this 🤷

src/client/css/partials/settings.css
```diff
@@ -0,0 +1,151 @@
+.settings {
+  display: flex;
+  flex-wrap: wrap;
+  gap: var(--padding-md) var(--padding-xl);
+  padding: var(--padding-lg);
+}
+
+.settings-title,
+.settings-section-title {
+  color: var(--main-dark);
```
toufali, line 10: I think this var doesn't exist any more?

toufali (approved):
This looks good!  Approving with a couple notes to follow up on:
- [I commented](https://github.com/mozilla/blurts-server/pull/2773#discussion_r1091442592) in @rhelmer 's PR that we need to update the images and associated paths for the verification email.  Not sure if we want to merge his PR and handle the update in yours since it's more on the "frontend" side, but I'll let y'all work that out.
- one last thought on all the DOM query + conditionals: I _think_ this settings page will make all those DOM queries for every page load, because all client scripts are bundled.  e.g. even though the Breaches page only queries the DOM once to ensure the right partial, it will also incur the cost of every DOM query and conditional made by Settings.  It might be a micro-optimization, but something to keep in mind – maybe we create a refactor ticket to optimize when we've got more time?

flozia:
> This looks good! Approving with a couple notes to follow up on:

Thank you @toufali — I addressed your inline comments and regarding your other points:

- **Confirmation email images**
Just opened to update the images https://github.com/mozilla/blurts-server/pull/2799
- **Optimize DOM queries**
That’s a very good point and a desirable optimization. I created [MNTOR-1139](https://mozilla-hub.atlassian.net/browse/MNTOR-1139) for a follow-up.

PR #2771, opened 2023-01-18
adr-0001-native-templating

toufali:
# Description
Proposal on upgrading the template system used by Monitor


rhelmer:
FYI the FxA team wrote this ADR a while back, the context was email templates specifically:
https://github.com/mozilla/fxa/blob/main/docs/adr/0024-upgrade-templating-toolset-of-auth-server-emails.md

toufali:
Thanks for sharing @rhelmer – this is helpful context.  Especially interesting was the use of Storybook for email previews – I'd like to explore this further!

rhelmer (approved):
lgtm and it's the way we've been going already, I don't know if you want to tackle it in here but besides storybook, another thing we should consider addressing is the ability to send a text version of the email as well.

IIRC FxA uses a tool to extract text from the HTML template (EJS), which I like because there's less redundancy, but I worry that the text version could end up with weird artifacts from this translation.

mansaj (approved):

PR #2772, opened 2023-01-19
MNTOR-1038 refactor email utility for V2

rhelmer:
# References: 
Jira: MNTOR-1038


src/utils/email.js
```diff
@@ -0,0 +1,124 @@
...
+  getVerificationUrl (subscriber) {
+    if (!subscriber.verification_token) throw new Error('subscriber has no verification_token')
+    let url = new URL(`${AppConstants.SERVER_URL}/api/v1/user/verify-email`)
+    url = this.appendUtmParams(url, 'verified-subscribers', 'account-verification-email')
+    url.searchParams.append('token', encodeURIComponent(subscriber.verification_token))
+    return url
+  },
+
+  getUnsubscribeUrl (subscriber, emailType) {
+    // TODO: email unsubscribe is broken for most emails
```
rhelmer, line 102: Do we want to try to tackle this now? I'd like more details on the problem.

src/utils/email.js
```diff
@@ -0,0 +1,124 @@
...
+
+import AppConstants from '../app-constants.js'
+
+import { createTransport } from 'nodemailer'
+import mozlog from './log.js'
+
+// TODO
+// import { getTemplate, verifyPartial } from '../views/email-2022.js'
+
+const log = mozlog('email-utils')
```
rhelmer, line 13: Is this still preferred over `console.*`? I noticed there's a `log.js` in `./utils` so I assumed so, but I see we're using the former in some places in `./src`.

src/utils/email.js
```diff
@@ -0,0 +1,124 @@
...
+   * @param {object} data
+   * @returns <Promise>
+   */
+  sendEmail (recipient, subject, template, data) {
+    if (!gTransporter) {
+      return Promise.reject(new Error('SMTP transport not initialized'))
+    }
+
+    // const html = getTemplate(verifyPartial, data)
+    // TODO implement when email template is ready
```
rhelmer, line 47: https://github.com/mozilla/blurts-server/pull/2765

src/utils/email.js
```diff
@@ -0,0 +1,124 @@
...
+// TODO
+// import { getTemplate, verifyPartial } from '../views/email-2022.js'
+
+const log = mozlog('email-utils')
+
+// The SMTP transport object. This is initialized to a nodemailer transport
+// object while reading SMTP credentials, or to a dummy function in debug mode.
+let gTransporter
+
+const EmailUtils = {
```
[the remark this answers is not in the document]
rhelmer: Is this something we can have tooling do? eslint or prettier etc?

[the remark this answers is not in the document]
rhelmer: Oh OK I misread at first that it was just styling :) Yes I think this could use named exports of functions vs. exporting the whole object.

src/utils/sha1.js
```diff
@@ -0,0 +1,10 @@
+// TODO: to be deprecated
```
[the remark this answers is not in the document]
rhelmer: Oh OK I didn't see it was already exported from there - that's fine for now, it does seem a little unrelated to FxA proper but not a big deal, I'd rather file an issue for later if we want to do this.

src/utils/email.js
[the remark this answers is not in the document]
rhelmer: @mansaj oh, hm so the reason I did this is that the current design of `EmailUtils` is a singleton, so `init` is called once and then the other functions use `this.*` as appropriate.

I'm not actually sure this is necessary (or desirable) though, I'm going to make this change but leave this comment unresolved for the moment while I look into it.

mansaj (approved):
LGTM, pending unsub discussion

package.json
```diff
@@ -49,7 +49,8 @@
     "nodemon": "^2.0.20",
     "redis-mock": "^0.56.3",
     "stylelint": "^14.9.1",
-    "stylelint-config-standard": "^26.0.0"
+    "stylelint-config-standard": "^26.0.0",
+    "testdouble": "^3.16.8"
```
toufali, line 53: Do we want to include this as part of `src/package.json` or is intended to live in the root (legacy) package?  (We _will_ have to reconcile both files, but it might be nice for new stuff to live in the new package?)

src/package.json
```diff
@@ -43,6 +43,7 @@
   "devDependencies": {
     "ava": "^5.1.0",
     "c8": "^7.12.0",
+    "quibble": "^0.6.15",
```
toufali, line 46: I see `quibble` added but not used – is it safe to remove?

toufali (commented):
Thanks for this refactor @rhelmer!  Tests work as expected locally.  
- There's a readme section (Trigger Breach Alert Email) that involves this utility – does it need updating as part of this PR?
- It might be nice to align on code style – it could simplify reviews and make git blame work better!

toufali:
RE Unsubscribe – it's currently implemented for the monthly emails only, and done in a "quick and dirty" way.  I had a discussion with PM that we should have a more established "subscriptions" section in our settings page, but that has been de-prioritized for now, so we can stick with what we've got!

src/package.json
rhelmer: Hm, that's a dependency of `testdouble` (`quibble` is a module loader, it's a lot like `esmock`, but these aren't very useful without a mocking library like `testdouble` or similar).

I don't think it needs to be in the `package.json` though, this is probably a remnant of when I was exploring using `esmock` and then tried `quibble` directly, and then realized `testdouble` was closer to what we needed to replace `jest`. I'll see if I can remove it from here and just have it as a dep of td in `package-lock.json`

rhelmer:
> Thanks for this refactor @rhelmer! Tests work as expected locally.
> 
>     * There's a readme section (Trigger Breach Alert Email) that involves this utility – does it need updating as part of this PR?

Thanks for pointing this out - I think it will need updating, but two things make me thing we're not ready yet:

1. I'm not sure we've implemented enough in V2 to be able to complete the test in that section (have we ported the breach alerts template yet?)
2. I feel like we should start explicitly defaulting to V2, otherwise it's hard to tell which version the readme refers to

I think this is important though, could we file follow-up tickets? I feel like we should have one to convert the breach alerts template (if we haven't done that already) and we should explicitly flip V2 on by default (blocked by the current V2 staging work).

I didn't want to tackle converting all the templates over in this PR because I have still need to understand better how to structure these, and I want to unblock the settings back-end and front-end PRs in the meantime.

Also as an aside, on re-reading this section I just noticed the mention in the readme that Mailinator makes it easier to test images unlike the way I've been testing, that makes thing much easier for me!

>     * It might be nice to align on code auto-formatting – it could simplify reviews and make git blame work better!

Yes you mentioned this before and I meant to follow up before now, sorry about that! I just took a few minutes to look at my setup, and I think I was assuming we were using a shared config already. Is it possible to take the settings in the readme and put them into `./vscode/settings.json`?

BTW I'm happy to file the issues if that sounds correct and you agree that these are OK to break out as followup tasks ^

rhelmer:
> > ```
> > * It might be nice to align on code auto-formatting – it could simplify reviews and make git blame work better!
> > ```
> 
> Yes you mentioned this before and I meant to follow up before now, sorry about that! I just took a few minutes to look at my setup, and I think I was assuming we were using a shared config already. Is it possible to take the settings in the readme and put them into `./vscode/settings.json`?

OK I went ahead and fixed up my local environment, and I opened this WIP PR as an example of how we could share: https://github.com/mozilla/blurts-server/pull/2777. Need to file a Jira ticket if we decide we want to do this.

toufali:
@rhelmer thanks for your replies – all sounds good to me 👍

toufali (approved):

rhelmer:
> > * There's a readme section (Trigger Breach Alert Email) that involves this utility – does it need updating as part of this PR?
> 
> Thanks for pointing this out - I think it will need updating, but two things make me think we're not ready yet:
> 
>     1. I'm not sure we've implemented enough in V2 to be able to complete the test in that section (have we ported the breach alerts template yet?)

Filed https://mozilla-hub.atlassian.net/browse/MNTOR-1111

>     2. I feel like we should start explicitly defaulting to V2, otherwise it's hard to tell which version the readme refers to

Filed https://mozilla-hub.atlassian.net/browse/MNTOR-1112

PR #2773, opened 2023-01-19
backend work for v2 settings MNTOR-1004

rhelmer:
# References: 
Jira: MNTOR-1004
Figma: 

# Description

Refactor existing settings (formerly "user" in V1) controller and API.

# Screenshot (if applicable)

Not applicable.

# How to test

API can be tested directly, or using front-end code in https://github.com/mozilla/blurts-server/pull/2765


src/utils/csrf.js
```diff
@@ -7,7 +7,7 @@ import * as AppConstants from '../app-constants.js'
 
 // defaults to ignore GET, HEAD, OPTIONS
 const { doubleCsrfProtection, generateToken } = doubleCsrf({
-  getSecret: (req) => req.secret || AppConstants.CSRF_SECRET,
+  getSecret: () => AppConstants.CSRF_SECRET,
```
rhelmer, line 10: We discussed this in eng meeting today, need to make sure we're using this correctly.

rhelmer:
This is just about ready for review, except for finishing porting over unit tests, and reconciling with https://github.com/mozilla/blurts-server/pull/2786 - we might want to land that one first so things are easier to understand.

src/client/css/variables.css
```diff
@@ -1,15 +1,8 @@
-/* This Source Code Form is subject to the terms of the Mozilla Public
- * License, v. 2.0. If a copy of the MPL was not distributed with this
- * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
```
flozia, line 3: Do we not need to keep this header?

src/client/css/variables.css
```diff
@@ -1,15 +1,8 @@
...
-
 :root {
   --multiplier: 1;
   --max-width: 1200px;
   --min-width: 320px;
-  --mobile-breakpoint: 480px;
   --border-radius: 4px;
-  --checkbox-w: 20px;
-  --padding-xs: calc(4px * var(--multiplier));
```
flozia, line 12: This one is being used for the user menu as well.

src/client/css/variables.css
```diff
@@ -1,15 +1,8 @@
-/* This Source Code Form is subject to the terms of the Mozilla Public
- * License, v. 2.0. If a copy of the MPL was not distributed with this
- * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
-
 :root {
   --multiplier: 1;
   --max-width: 1200px;
   --min-width: 320px;
-  --mobile-breakpoint: 480px;
```
flozia, line 9: This variable is being used in src/client/css/user-menu.css.

src/client/css/variables.css
```diff
@@ -19,16 +12,17 @@
   --h2-font: 700 1.75rem/1 metropolis, sans-serif;
   --h3-font: 700 1.125rem/1 metropolis, sans-serif;
   --h4-font: 700 1rem/1 metropolis, sans-serif;
+  --main-dark: #2e2c34;
   --gray-5: #f9f9fa;
   --gray-10: #e7e7e7;
   --gray-20: #cececf;
   --gray-40: #6d6d6e;
-  --gray-50: #3d3d3d;
```
flozia, line 26: Actually, all of the removed ones in this PR are used — I think.

src/controllers/settings.js
```diff
@@ -2,16 +2,191 @@
...
+  const allBreaches = req.app.locals.breaches
+  for (const email of emails) {
+    const breaches = await getBreachesForEmail(
+      email.sha1,
+      allBreaches,
+      true,
+      false
+    )
+    breachCounts.set(email.email, breaches.length)
+  }
```
flozia, line 49: Suggestion (non-blocking): I’m wondering if it would be good to make the breach count part of the email data and pass it altogether.

src/controllers/settings.test.js
```diff
@@ -0,0 +1,416 @@
+import test from 'ava'
+import * as td from 'testdouble'
+
+import { createResponse, createRequest } from 'node-mocks-http'
```
flozia, line 4: Praise: Awesome that you got to make the tests work!

src/controllers/settings.test.js
```diff
@@ -0,0 +1,416 @@
+import test from 'ava'
```
flozia, line 1: Question: Do we need a license header comment here?

src/controllers/settings.js
```diff
@@ -2,16 +2,191 @@
...
+
+import { setAllEmailsToPrimary } from '../db/tables/subscribers.js'
+
+import { fluentError, getMessage } from '../utils/fluent.js'
+import { sendEmail, getVerificationUrl, getUnsubscribeUrl } from '../utils/email.js'
 import { mainLayout } from '../views/main.js'
 import { settings } from '../views/partials/settings.js'
+import AppConstants from '../app-constants.js'
+import { getBreachesForEmail } from '../utils/hibp.js'
+import { generateToken } from '../utils/csrf.js'
```
flozia, line 22: Suggestion (non-blocking): Let’s group the utils and partial imports.

flozia (commented):
This is looking good! All of the data is being populated and the emails are sent on my end. 

Some notes:
1. There is one error I noticed when navigating to the “Settings” page: `fx-monitor.middleware.error: Error: Page not found! GET /user/undefined`
2. There are a few CSS variable were removed as part of this PR, which are used in other places.
3. Would you like to address the commented code in a follow-up? If that’s the case, we should consider still removing it from this PR — what do you think?

src/controllers/settings.js
```diff
@@ -2,16 +2,191 @@
...
+
+  const allBreaches = req.app.locals.breaches
+  for (const email of emails) {
+    const breaches = await getBreachesForEmail(
+      email.sha1,
+      allBreaches,
+      true,
+      false
+    )
+    breachCounts.set(email.email, breaches.length)
```
mansaj, line 48:
```suggestion
    breachCounts.set(email.email, breaches?.length || 0)
```
And avoid having 2 loops (remove ln 38)?

src/controllers/settings.js
```diff
@@ -2,16 +2,191 @@
...
+
+  const unverifiedSubscriber = await addSubscriberUnverifiedEmailHash(
+    req.session.user,
+    email
+  )
+
+  await _sendVerificationEmail(unverifiedSubscriber.id)
+
+  // TODO: what should we return to the client?
+  return res.json('Resent the email')
```
mansaj, line 87: Resent? Judging by the name of the function `addEmail`, it's an universal add no?

As discussed in eng meeting, we want to get to a generic format, something like. This refactor work can certainly be in a different PR/jira task though!
```
{
  success: true,
  status: 200,
  message: '',
  stack (only dev/debug): {}
}
```

src/controllers/settings.js
```diff
@@ -2,16 +2,191 @@
...
+    const allEmailsToPrimary = Number(req.body.communicationOption) === 1
+    const updatedSubscriber = await setAllEmailsToPrimary(
+      sessionUser,
+      allEmailsToPrimary
+    )
+    req.session.user = updatedSubscriber
+
+    return res.json('Comm options updated')
+  } catch (ex) {
+    return res.json('Error updating comm options')
```
mansaj, line 181: might want to log the `ex`

src/controllers/settings.js
```diff
@@ -2,16 +2,191 @@
...
+    email
+  )
+
+  await _sendVerificationEmail(unverifiedSubscriber.id)
+
+  // TODO: what should we return to the client?
+  return res.json('Resent the email')
+}
+
+function _checkForDuplicateEmail (sessionUser, email) {
```
mansaj, line 90: Styling (non-blocking): do we want to adopt _ for all private functions?

src/client/css/variables.css
rhelmer: Sorry, I think I merged this the wrong way -_- I'll fix.

src/controllers/settings.test.js
rhelmer: Thanks! If you browse through them you can see I've disabled things here and there, but I think we can preserve a lot of the good work that was already done here.

src/controllers/settings.js
rhelmer: Ah good catch, this is a holdover from the [V1 user controller](https://github.com/mozilla/blurts-server/blob/96fae648cd5428372e2401c4f964fa19ea857be6/controllers/user.js#L68-L69).

I'll go over the various responses and make them consistent for now, we discussed centralizing this and using schema validation etc. but I will file a follow-up for that.

src/controllers/settings.js
rhelmer: That's what I've seen most commonly, I know there's a [private syntax for classes](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Private_class_fields) now, but I haven't seen that used outside of classes.

src/client/css/variables.css
rhelmer: Ah, these came from me merging the MNTOR-983-settings-v2-frontend-additions branch, I'll go ahead and revert. Thanks!

src/controllers/settings.js
rhelmer: (note that the old `user` controller duplicated the function for sending and re-sending email verification, I consolidated that since I didn't see any important differences, but I ended up keeping the API response from the latter apparently)

rhelmer: The error handler function (which should be triggered if we throw any uncaught exceptions) looks like it matches this format: https://github.com/mozilla/blurts-server/blob/8d941c5572bd07ae28121d457ff39dd3321b0c3f/src/middleware/error.js#L22-L32

So I think it's really the success cases we need to handle, I'll do it for now. We probably don't need `stack` for success cases I don't think.

src/controllers/settings.js
rhelmer: Per above, I think if we don't try to catch here, the error handler function registered in Express will do the right thing. I think this is a good pattern overall, it's better not to catch exceptions if we aren't sure we can handle them, and in this case the catch-all handler does exactly what you're proposing (log the message and return a standardized format JSON response)

rhelmer:
>     1. There is one error I noticed when navigating to the “Settings” page: `fx-monitor.middleware.error: Error: Page not found! GET /user/undefined`

I haven't tried too hard to track this down yet, but I think the front-end might be an image load that's going wrong from the front-end somehow, which is probably something I introduced before handing that off. I'll dig into this today and confirm.

flozia:
> > ```
> > 1. There is one error I noticed when navigating to the “Settings” page: `fx-monitor.middleware.error: Error: Page not found! GET /user/undefined`
> > ```
> 
> I haven't tried too hard to track this down yet, but I think the front-end might be an image load that's going wrong from the front-end somehow, which is probably something I introduced before handing that off. I'll dig into this today and confirm.

You mentioning an image you made me check the user menu and that one is currently not working on the settings page. Readding this line should resolve the issue: https://github.com/mozilla/blurts-server/pull/2773/files#diff-b0bffe56a4a49387046eefb9c7d501a83487e1bcf11f968f85a65cbfd1e3ef4aL10. 🙂

src/controllers/settings.js
```diff
@@ -2,16 +2,199 @@
...
+  const recipientEmail = unverifiedEmailAddressRecord.email
+  const data = {
+    recipientEmail,
+    ctaHref: getVerificationUrl(unverifiedEmailAddressRecord),
+    utmCampaign: 'email_verify',
+    unsubscribeUrl: getUnsubscribeUrl(
+      unverifiedEmailAddressRecord,
+      'account-verification-email'
+    ),
+    whichPartial: 'email_partials/email_verify',
```
flozia, line 156: This can be removed, we don’t use it anymore.

```suggestion
```

src/controllers/settings.js
```diff
@@ -2,16 +2,199 @@
...
+
+  return res.json({
+    success: true,
+    status: 200,
+    message: 'Sent the verification email'
+  })
+}
+
+function _checkForDuplicateEmail (sessionUser, email) {
+  email = email.toLowerCase()
```
flozia, line 96: Suggestion (non-blocking): Let’s try to avoid mutating passed args.

```suggestion
  const emailLowerCase = email.toLowerCase()
```

flozia (approved):

src/routes/api/v1/user.js
```diff
@@ -2,16 +2,22 @@
...
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
-import { Router } from 'express'
+import { Router, json } from 'express'
 
 import { asyncMiddleware } from '../../../middleware/util.js'
 import { requireSessionUser } from '../../../middleware/auth.js'
 import { methodNotAllowed } from '../../../middleware/error.js'
 import { putBreachResolution, getBreaches } from '../../../controllers/breaches.js'
+import { addEmail, resendEmail, removeEmail, verifyEmail, updateCommunicationOptions } from '../../../controllers/settings.js'
```
flozia, line 11: Suggestion (formatting): It’s getting a bit looong here. :)
```suggestion
import {
  addEmail,
  resendEmail,
  removeEmail,
  verifyEmail,
  updateCommunicationOptions
} from '../../../controllers/settings.js'
```

toufali (commented):
I want to review this more thoroughly but ran out of time looking at the frontend PR first.  I'll continue this review later, but wanted to post this note early:

Is the "add email" flow covered under this PR, or separate?  I think it was originally a separate ticket, but in case it was intended to work here, here's an error I encountered after clicking the "Send verification link" button:
```
ERROR fx-monitor.middleware.error: error: insert into "email_addresses" ("email", "sha1", "subscriber_id", "verification_token", "verified") values ($1, $2, $3, $4, $5) returning * - duplicate key value violates unique constraint "email_addresses_pkey"
    at Parser.parseErrorMessage (/Users/atoufali/Dev/blurts-server/node_modules/pg-protocol/dist/parser.js:287:98)
    at Parser.handlePacket (/Users/atoufali/Dev/blurts-server/node_modules/pg-protocol/dist/parser.js:126:29)
    at Parser.parse (/Users/atoufali/Dev/blurts-server/node_modules/pg-protocol/dist/parser.js:39:38)
    at Socket.<anonymous> (/Users/atoufali/Dev/blurts-server/node_modules/pg-protocol/dist/index.js:11:42)
    at Socket.emit (node:events:513:28)
    at addChunk (node:internal/streams/readable:324:12)
    at readableAddChunk (node:internal/streams/readable:297:9)
    at Readable.push (node:internal/streams/readable:234:10)
    at TCP.onStreamRead (node:internal/stream_base_commons:190:23)
    at TCP.callbackTrampoline (node:internal/async_hooks:130:17)
```

toufali:
Update on the "Add new email" flow: after another attempt, I was able to trigger the verification email successfully, but upon pushing the "Verify email" button in the message, I got this message:

```
ERROR fx-monitor.hibp._throttledFetch: {"err":"Error: bad response: 400"}
ERROR fx-monitor.middleware.error: Error: Error connecting to Have I Been Pwned.
    at fluentError (file:///Users/atoufali/Dev/blurts-server/src/utils/fluent.js:102:10)
    at _throttledFetch (file:///Users/atoufali/Dev/blurts-server/src/utils/hibp.js:53:11)
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
    at async kAnonReq (file:///Users/atoufali/Dev/blurts-server/src/utils/hibp.js:67:10)
    at async subscribeHash (file:///Users/atoufali/Dev/blurts-server/src/utils/hibp.js:238:10)
    at async _verifyNewEmail (file:///Users/atoufali/Dev/blurts-server/src/db/tables/email_addresses.js:175:3)
    at async verifyEmailHash (file:///Users/atoufali/Dev/blurts-server/src/db/tables/email_addresses.js:72:25)
    at async verifyEmail (file:///Users/atoufali/Dev/blurts-server/src/controllers/settings.js:170:3)
```

Again, maybe this work is/should be handled under separate PR – I might be testing prematurely?

rhelmer:
> Update on the "Add new email" flow: after another attempt, I was able to trigger the verification email successfully, but upon pushing the "Verify email" button in the message, I got this message:
> 
> ```
> ERROR fx-monitor.hibp._throttledFetch: {"err":"Error: bad response: 400"}
> ERROR fx-monitor.middleware.error: Error: Error connecting to Have I Been Pwned.
>     at fluentError (file:///Users/atoufali/Dev/blurts-server/src/utils/fluent.js:102:10)
>     at _throttledFetch (file:///Users/atoufali/Dev/blurts-server/src/utils/hibp.js:53:11)
>     at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
>     at async kAnonReq (file:///Users/atoufali/Dev/blurts-server/src/utils/hibp.js:67:10)
>     at async subscribeHash (file:///Users/atoufali/Dev/blurts-server/src/utils/hibp.js:238:10)
>     at async _verifyNewEmail (file:///Users/atoufali/Dev/blurts-server/src/db/tables/email_addresses.js:175:3)
>     at async verifyEmailHash (file:///Users/atoufali/Dev/blurts-server/src/db/tables/email_addresses.js:72:25)
>     at async verifyEmail (file:///Users/atoufali/Dev/blurts-server/src/controllers/settings.js:170:3)
> ```
> 
> Again, maybe this work is/should be handled under separate PR – I might be testing prematurely?
> 
> The "Resend verification email" CTA appears and functions correctly, but I encounter the same error clicking from a resent email. I confirmed HIBP is not down.
> 
> Perhaps this is related to @mansaj DB updates? If so, we might need to update the code here.
> 
> Deleting the unverified email works as expected.

Hm, I am seeing this now too, is it possible that we need to update the HIBP key in our local envs? This part of the code is largely the same as V1 (it was in `DB.js` in V1 and now moved to `rc/db/tables/email_addresses.js` in V2)

src/controllers/settings.js
rhelmer: I'll file a followup to come back to this one, I think it makes sense to do this processing in the controller vs. the template, but I'd like to take some time to think a bit more about the datastructures we want to use.

rhelmer: https://mozilla-hub.atlassian.net/browse/MNTOR-1135

src/controllers/settings.js
```diff
@@ -2,16 +2,198 @@
...
+
+  const allBreaches = req.app.locals.breaches
+  for (const email of emails) {
+    const breaches = await getBreachesForEmail(
+      email.sha1,
+      allBreaches,
+      true,
+      false
+    )
+    breachCounts.set(email.email, breaches?.length || 0)
```
toufali, line 49: Is this correct usage of optional chaining?  [From MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining#description):

>Optional chaining cannot be used on a non-declared root object, but can be used with a root object with value undefined.

```js
undeclaredVar?.prop; // ReferenceError: undeclaredVar is not defined
```

src/controllers/settings.js
```diff
@@ -2,16 +2,198 @@
...
+    limit: AppConstants.MAX_NUM_ADDRESSES,
+    csrfToken: generateToken(res)
   }
 
   res.send(mainLayout(data))
 }
 
-export { settingsPage }
+async function addEmail (req, res) {
+  const sessionUser = await req.user
```
toufali, line 65: I don't understand why we need to await the `user` prop?

toufali (commented):
Haven't been able to test locally yet (per Slack discussion), but code-wise this looks good, thanks @rhelmer!  No major blockers, but a few minor things we should consider changing before merging.  💪

src/controllers/settings.js
rhelmer: Yeah that's a good point, I think exported functions are clearly public, it still can be useful to communicate to internal consumers which functions/methods should not be used outside of a particular module or library, some language make this distinction using `private`/`public`/`protected`.

JS isn't really one of those languages though so probably not worth worrying a whole lot about! I'm fine removing it for now.

src/emails/templates/email-2022.js
```diff
@@ -0,0 +1,238 @@
...
+          subhead: 'email-verify-subhead'
+        })}
+        ${partial}
+        ${emailFooter(data)}
+      </table>
+    </body>
+  </html>
+`
+
+export { getTemplate }
```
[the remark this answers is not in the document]
rhelmer: Yes I made this before everything was settled, I'll adapt this.

src/routes/api/v1/user.js
```diff
@@ -2,16 +2,28 @@
...
 
 const router = Router()
 // breaches
 router.put('/breaches', requireSessionUser, asyncMiddleware(putBreachResolution))
 router.get('/breaches', requireSessionUser, asyncMiddleware(getBreaches))
+router.post('/email', requireSessionUser, asyncMiddleware(addEmail))
+router.post('/resend-email', requireSessionUser, asyncMiddleware(resendEmail))
+router.post('/remove-email', requireSessionUser, asyncMiddleware(removeEmail))
+router.get('/verify-email', asyncMiddleware(verifyEmail))
+router.post('/update-comm-option', json(), requireSessionUser, asyncMiddleware(updateCommunicationOptions))
```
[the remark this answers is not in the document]
rhelmer: Ah good catch thanks!

src/controllers/settings.js
rhelmer: Hm this came along from the bits of the user controller in V1 and I agree it's likely redundant. This is the kind of thing TS would warn about, if it knew that the type of `req` (pretty sure we could use JSDoc or other optional annotations without going full TS and get the benefits too)

src/controllers/settings.js
rhelmer: Hm, wouldn't `breaches` always be declared right above this? From a quick look at `getBreachesForEmail()` it looks like it might return `undefined`, so this would just avoid throwing in that case:

```js
> const test = undefined
> test?.length
undefined
> test.length
Uncaught TypeError: Cannot read properties of undefined (reading 'length')
```

rhelmer:
As I mentioned in standup today, I think we should defer porting some of the old `user` controller tests over, since they use a mix of mocks and local DB (I'd prefer to just use one or the other, and mocks are probably more appropriate for these particular tests).

Using a local dev database can be OK in some instances, I'd rather do that in the context of integration/end-to-end tests generally though and not unit tests.

Filed https://mozilla-hub.atlassian.net/browse/MNTOR-1138

rhelmer:
BTW here are a few small examples of how mocks in the unit tests helped find bugs in the course of writing this. The unit tests are very quick to run and I've tried to keep them very targeted on testing the smallest unit of code possible (ideally a particular exported function).

1. This code is not correct:
```js
  if (!email || !email.test(emailRegex)) {
    throw fluentError('user-add-invalid-email')
  }
```
It should be:
```js
 if (!email || !emailRegex.test(email)) {
    throw fluentError('user-add-invalid-email')
  }
```
The test mocked everything except the core logic that this function performs, which includes checking for valid-looking email addresses.

2. When I moved the email verification template, I forgot to change the relative import to `fluent`. The unit tests started failing right away.

These were both far faster than running the server and doing manual or automated QA, it's fast enough that you can configure VSCode to just run them on save or just depend on the github status.

src/controllers/settings.js
toufali: Yes, that makes sense and fits the exception given by MDN that I didn't read clearly!  Marking this resolved 👍

src/views/email-2022.js
```diff
@@ -0,0 +1,238 @@
...
+  legal: `https://www.mozilla.org/about/legal?utm_source=fx-monitor&utm_medium=email&utm_campaign=${data.utmCampaign}&utm_content=email-footer-link`,
+  termsAndPrivacy: `https://www.mozilla.org/privacy/firefox-monitor?utm_source=fx-monitor&utm_medium=email&utm_campaign=${data.utmCampaign}&utm_content=email-footer-link`
+})
+
+const images = {
+  header: `${AppConstants.SERVER_URL}/img/email_images/person-at-desk.png`,
+  footer: `${AppConstants.SERVER_URL}/img/email_images/mozilla-logo-bw-rgb.png`,
+  logoDark: `${AppConstants.SERVER_URL}/img/email_images/monitor-logo-transparent-dark-mode.png`,
+  logoLight: `${AppConstants.SERVER_URL}/img/email_images/monitor-logo-bg-f9f9fa.png`
+}
```
toufali, line 21: Woops, almost missed this.  I think these paths refer to the old site.  We need to duplicate these images into the `src/client/images` folder.  And another twist – our new repo is using .webp and we should avoid .png (the build phase is not set to handle .png files)

I believe @flozia knows how to use the `convert:webp` script in `src/package.json`, and perhaps this is all already handled in the frontend PR for this feature?  I'll take a look next.

toufali (approved):
Looking good!  Approving this so that I'm not blocking, but there _is_ one blocker that needs to be resolved, possibly under @flozia 's frontend PR.  See note about images and their paths pointing to the old site structure.

toufali:
> 
>     2. When I moved the email verification template, I forgot to change the relative import to `fluent`. The unit tests started failing right away.
> 

My version of VS Code updates imports automagically 😉
But point taken, thanks for the examples.

src/views/email-2022.js
flozia: @toufali That’s my bad, I only learned about our use of `webp` after introducing this part of the email and did not update respectively. @rhelmer Opened PR #2799 which adds the converted images.

rhelmer: Thanks for catching this! @flozia took care of this for us in https://github.com/mozilla/blurts-server/pull/2799

PR #2789, opened 2023-01-25
fix webp script and convert png

toufali:
# References: 
Jira: MNTOR-1120

# Description
Fixes broken webp convert script and checks in a converted png file so that it works in non-dev environments (heroku, GCP)

# How to test
- As documented in the script, `npm run convert:webp` should convert all png files in `client/images/` to webp and delete the source png

- The temp circle-graph image on the breaches page should not appear broken when running `npm start`


flozia (approved):
Works great!

PR #2790, opened 2023-01-25
Adds email preview for email verification and breach notification

flozia:
# References: 
Jira:
- [MNTOR-1056](https://mozilla-hub.atlassian.net/browse/MNTOR-1056)
- [MNTOR-1107](https://mozilla-hub.atlassian.net/browse/MNTOR-1107)

# Description

This PR migrates the breach alert email and adds a preview page from where we can also send test emails.

# Screenshot

image

# How to test

1. Set the `.env` variable `ADMINS=` so you can navigate to `/admin/emails`
2. Set `EMAIL_TEST_RECIPIENT=localmonitor20200827@mailinator.com` — this is the same public Mailinator inbox address we already use. For more info see [README](https://github.com/mozilla/blurts-server#trigger-breach-alert-email).
3. You can toggle between email preview templates using the select input and send test emails to your primary or the Mailinator email address mentioned in the previous step.


src/client/js/partials/breaches.js
```diff
@@ -104,4 +104,30 @@ function render () {
...
+    })
+
+    if (!res.ok) throw new Error('Bad fetch response')
+
+    console.log('Sent breach alert email')
+  } catch (e) {
+    console.error('Could not send breach alert email:', e)
+  }
+}
+
```
flozia, line 132: This is for the “Send test breach alert” button on the “Breaches page” and is only meant for testing this PR.

src/emails/partials/email-breach-alert.js
```diff
@@ -0,0 +1,139 @@
...
+                  ${data.breachAlert.Title}
+                </td>
+              </tr>
+              <tr>
+                <td style='padding: 24px;'>
+                  <p
+                    class='text-light'
+                    style='${breachAlertLabelStyle}'
+                  >
+                    ${'data.breachAlert.localizedStrings.breachAdded'}
```
flozia, line 97: I’ll have to look into how we get this …

src/emails/partials/email-breach-alert.js
```diff
@@ -0,0 +1,139 @@
...
+                    ${data.breachAlert.AddedDate}
+                  </p>
+
+                  ${data.breachAlert.DataClasses
+                    ? `
+                        <p
+                          class='text-light'
+                          style='${breachAlertLabelStyle}'
+                        >
+                          ${'data.breachAlert.localizedStrings.compromisedData'}
```
flozia, line 112: … and this localized string in V1 a bit more.

src/views/partials/breaches.js
```diff
@@ -75,6 +75,9 @@ export const breaches = data => `
 <section>
   <header class='breaches-header'>
     <h1>${getMessage('breach-heading-email', { 'email-select': `<custom-select name='email-account'>${createEmailOptions(data.breachesData)}</custom-select>` })}</h1>
+    <button id='sendAlertTestButton' class='primary'>
+      Send test breach alert
+    </button>
```
flozia, line 80: Only for testing the PR.

src/controllers/hibp.js
```diff
@@ -2,9 +2,27 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
+import { acceptedLanguages, negotiateLanguages } from 'fluent-langneg'
```
toufali, line 5: hmm... don't we already store the user's language preference in db?

toufali (commented):
Yikes there's a lot going on here and I'll need fresh eyes to take a look.  Added just a single comment that is legitimately a question – it's likely what your doing is correct, I'm just wondering out loud.

src/controllers/hibp.js
flozia: Yes, you are right!

src/client/js/partials/breaches.js
```diff
@@ -104,4 +107,28 @@ function render () {
...
+        hashPrefix: '365050',
+        hashSuffixes: ['53cbb89874fc738c0512daf12bc4d91765']
+      })
+    })
+
+    if (!res.ok) throw new Error('Bad fetch response')
+
+    console.log('Sent breach alert email')
+  } catch (e) {
+    console.error('Could not send breach alert email:', e)
```
rhelmer, line 130: Do you intend on landing this test function? It might be better not to catch here, that should make the top-level error handler throw a generic JSON error message.

flozia: Not exactly in this form, but I’ll intend to have a similar request that is being handled by the controller. I think your comment still applies though!

src/views/partials/breaches.js
flozia: The env `HIBP_NOTIFY_TOKEN` variable needs to be filled in [here](https://github.com/mozilla/blurts-server/pull/2790/files#diff-7b24ff39ac683ac7411b6527c5a510cfc036a9244fa613535460dbfcd734bc3dR115) for this to currently work. I’m just about to push a change that is handling this logic on the backend and should work out of the box.

.env-dist
```diff
@@ -6,6 +6,8 @@ LOGOS_ORIGIN=
 COOKIE_SECRET=3895d33b5f9730f5eb2a2067fe0a690e298f55f5e382c032fd3656863412
 CSRF_SECRET=
 SESSION_DURATION_HOURS=48
+EMAIL_PREVIEW_ENABLED=
+EMAIL_RECIPIENT_DUMMY=localmonitor20200827@mailinator.com
```
flozia, line 10: This is the same public Mailinator address we are already using for our [README](https://github.com/mozilla/blurts-server#trigger-breach-alert-email).

src/controllers/email-preview.js
```diff
@@ -0,0 +1,113 @@
...
+
+async function sendTestNotification (req, res) {
+  // The test breach notification can be viewed in the public Mailinator inbox
+  // as documented in the README:
+  // https://github.com/mozilla/blurts-server#trigger-breach-alert-email
+  const breachNotificationData = {
+    breachName: 'Adobe',
+    // Hash for dummy email `localmonitor20200827@mailinator.com`
+    hashPrefix: '365050',
+    hashSuffixes: ['53cbb89874fc738c0512daf12bc4d91765']
```
flozia, line 71: This is also public info we use in our `README`: [Trigger Breach Alert Email](https://github.com/mozilla/blurts-server#trigger-breach-alert-email)

src/views/partials/email-breach-alert.js
```diff
@@ -0,0 +1,134 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
```
flozia, line 1: The email templates are migrated from `hbs` and I tried to only extract the inline style to make the file more readable. We decided to not make any fundamental changes to the established code for the initial migration.

flozia (commented):
I still have to address the test failure and update the README if we decide to stick with the current `.env` variables. Other than that this PR is ready for review — let me know if there is anything that you come across that blocks you from testing any part of this PR.

src/routes/user.js
```diff
@@ -23,10 +24,20 @@ router.get('/breaches', requireSessionUser, breachesPage)
...
 
 // sign the user out
 router.get('/logout', asyncMiddleware(logout))
 
+// send test email
+router.post(
+  '/send-test-email',
+  requireSessionUser,
+  asyncMiddleware(sendTestEmail)
+)
```
Vinnl, line 41: Do we have anything that allows us to distinguish between local and other environments? I think we might be setting ourselves up for a DoS attack if we leave this open on public environments?

src/utils/hibp.js
```diff
@@ -146,6 +146,37 @@ async function loadBreachesIntoApp (app) {
   }
   log.info('done-loading-breaches', 'great success 👍')
 }
+
+/**
+Get addresses and language from either subscribers or email_addresses fields:
+TODO: Add route
```
Vinnl, line 152: It might just be that I don't have enough clarity on the codebase and project timelines yet, but possibly this comment could use some clarification on what actually needs to happen?

src/controllers/hibp.js
```diff
@@ -42,74 +59,125 @@ async function notify (req, res) {
...
+
+  if (doNotSendEmail) {
+    // Get a list of the failed condition `logId`s
+    const conditionLogIds = unsatisfiedConditions
+      .map(condition => condition.logId)
+      .join(', ')
+
+    log.info('Breach alert email was not sent.', {
+      name: breachAlert.Name,
+      reason: `The following conditions were not satisfied: ${conditionLogIds}.`
```
Vinnl, line 100: Nice job logging this!

src/utils/pretty-date.js
```diff
@@ -0,0 +1,13 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+function prettyDate (date, locales) {
```
Vinnl, line 5: Super nit, but (especially without TypeScript) I find it a helpful convention to start function names with a verb, to make them jump out when they're not called - so this could e.g. be `formatDate` or similar.

src/views/partials/email-breach-alert.js
Vinnl: Oh nce, I was going to leave a comment about how picky email clients are, but the styles are still actually inline - good job! (It's probably still a good idea to test in actual mail clients, e.g. new lines might still cause issues, but any issues are probably relatively easily solvable then.)

Vinnl (approved):
With the caveats that I'm not too familiar with the codebase yet, and am not terribly experienced in back-end work, no big things jumped out to me. Nice job!

src/utils/hibp.js
flozia: Oh, that one is a WIP comment that I’ve missed — thx for catching it. It’s not relevant anymore.

src/views/partials/email-breach-alert.js
flozia: That’s a good point! I we see any strange behavior we could easily strip the new lines.

src/routes/user.js
flozia: Right! @rhelmer added rate limiting for sending email verifications in PR #2807 — could we adopt this mechanic to work for this use-case easily?

rhelmer: We may want to consider restricting this further - I noticed there's this entry in the `.env-dist`:
```
# Email addresses that are allowed to test and send emails
ADMINS=
```

rhelmer: > We may want to consider restricting this further - I noticed there's this entry in the `.env-dist`:
> 
> ```
> # Email addresses that are allowed to test and send emails
> ADMINS=
> ```

Would it be possible to restrict this API to only be used by authenticated users that are on that list?

src/app-constants.js
```diff
@@ -16,6 +16,8 @@ const requiredEnvVars = [
   'SESSION_DURATION_HOURS',
   'SMTP_URL',
   'EMAIL_FROM',
+  'EMAIL_PREVIEW_ENABLED',
+  'EMAIL_RECIPIENT_DUMMY',
```
toufali, line 20: Should these env vars be moved to "optional"?  i.e. should we block a dev from running the site if they're not set?

src/client/js/index.js
```diff
@@ -9,5 +9,6 @@ import './components/custom-select.js'
 import './nav.js'
 import './user-menu.js'
 import './partials/breaches.js'
+import './partials/email-preview.js'
```
toufali, line 12: can we import this separately from the index bundle, so that non-dev users aren't affected?

src/controllers/hibp.js
toufali: Ah, maybe I was thinking of the server store: maybe the `getLocale ()` function in `utils/fluent.js` already does this work for you?  

It's been a while, but I _think_ we detect and negotiate lang preference on request already, and then store it to avoid running these functions again.  I could be mistaken, more investigation needed!

src/routes/user.js
toufali: Yes, the send functionality should be restricted to admins (dev and QA) only.  For context, this is the reason @flodolo requested a separate version without auth, so that volunteers could preview without needing admin/send access

src/utils/format-date.js
```diff
@@ -0,0 +1,13 @@
...
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+function formatDate (date, locales) {
+  const jsDate = new Date(date)
+  const options = { year: 'numeric', month: 'long', day: 'numeric' }
+  const intlDateTimeFormatter = new Intl.DateTimeFormat(locales, options)
+
+  return intlDateTimeFormatter.format(jsDate)
+}
```
toufali, line 11: No change needed here, but for context I created a similar util function but then resolved it was just as simple if not clearer to just include the date format in context with the associated code, for example in `views/partials/breaches.js`:

```js
function createBreachRows (data) {
  const locale = getLocale()
  const shortDate = new Intl.DateTimeFormat(locale, { year: 'numeric', month: '2-digit', day: '2-digit', timeZone: 'UTC' })
  const shortList = new Intl.ListFormat(locale, { style: 'narrow' })
  const longDate = new Intl.DateTimeFormat(locale, { dateStyle: 'long', timeZone: 'UTC' })
  const longList = new Intl.ListFormat(locale, { style: 'long' })
  const breachRowsHTML = data.verifiedEmails.flatMap(account => { ......
```
Setting the above constants allowed me to reuse the same custom format for different dates.  But, if it's more helpful for us to have this util, it doesn't hurt and I'm all for it 👍

.env-dist
toufali: For the previous version, QA mentioned how helpful it was to be able to send emails to specific addresses, not locked down to one (mailanator).  This easily allows testing on multiple email clients.  I wonder if it's possible to reintroduce this behavior?

If it's too difficult, we might at least default to sending to the current logged-in email.

Both of these options come with the auth caveat mentioned below.

toufali (commented):
This looks fantastic!  Great improvement over the previous version – I love the ability to preview in the context of our site.

Biggest notes from me are to consider auth and the ability to send emails to specific addresses.  Also, it would be nice to decouple this from any public facing loads, i.e. users should not have to download or run code for this feature.  You might check out dynamic imports – they might be helpful if you run into difficulty decoupling.

Thanks again for taking this on!  💪

src/routes/user.js
mansaj: yes so in the v1 code, we have a middleware`requireAdminUser` which is used on all email testing endpoints

src/controllers/email-preview.js
```diff
@@ -0,0 +1,113 @@
...
+  getNotifictionDummyData,
+  getVerificationDummyData,
+  sendEmail
+} from '../utils/email.js'
+
+const { EMAIL_RECIPIENT_DUMMY } = AppConstants
+
+function emailsPage (req, res) {
+  if (!AppConstants.EMAIL_PREVIEW_ENABLED) {
+    throw new UnauthorizedError(`${req.method} ${req.originalUrl}`)
```
mansaj, line 27: Since this isn't auth related, perhaps a 404 (not found error) would be more appropriate

src/controllers/email-preview.js
```diff
@@ -0,0 +1,113 @@
...
+  }
+
+  await notify(notifyReq, res)
+}
+
+async function sendTestEmail (req, res) {
+  const { emailId } = req.body
+
+  switch (emailId) {
+    case 'verification': {
```
mansaj, line 87: It might be a good idea to define email IDs as enums to avoid misspelling. I think it's also reference on L31

src/controllers/email-preview.js
```diff
@@ -0,0 +1,113 @@
...
+        verifyPartial
+      )
+      await sendEmail(
+        EMAIL_RECIPIENT_DUMMY,
+        getMessage('email-subject-verify'),
+        emailTemplate
+      )
+      break
+    }
+    case 'notification': {
```
mansaj, line 100: same as above re enum

.env-dist
flozia: I added support for sending test emails to the primary email and Mailinator. It would not be hard to add additional emails. If OK I would like to break out additional enhancement into follow-up PRs so we can start testing the changes and see if they are working as expected.

rhelmer (commented):
This is a very nice feature! Just testing it out, I see a few things, are these expected?

- The `send-test-email` API does not always seem to return a response to the browser, but I see the verification email was received and it logged `Sent test email: verification`
- If I try to test the breach alert email, I see `Sent test email: notification` logged on the console but I haven't received the email

src/controllers/email-preview.js
```diff
@@ -0,0 +1,116 @@
...
+      // Send test breach notification email
+      await sendTestNotification(req, res)
+      break
+    }
+    default: {
+      throw new Error(`No test email found for ${emailId}`)
+    }
+  }
+
+  console.info(`Sent test email: ${emailId}`)
```
rhelmer, line 113: Should there be a `return res.status(200).json({...})` here? I noticed the browser doesn't seem to get a response in the success case right now.

rhelmer (changes requested):
This looks great, other than the issues I pointed out from testing, in the settings PR we tried to settle on a common response object for success conditions, which matches the one we use for error conditions:

```
  return res.json({
    success: true,
    status: 200,
    message: ...
  })
}
```
It feels a bit redundant to put the `status` and `success` in there when the HTTP response exists (which is `200` by default I believe), it would be nice to figure out a way to have a shared format here (maybe a class or utility function).

src/controllers/email-preview.js
flozia: Yes, ty for catching that! I created [MNTOR-1176](https://mozilla-hub.atlassian.net/browse/MNTOR-1176) for creating a class or util and consolidating API responses.

flozia:
> * The `send-test-email` API does not always seem to return a response to the browser, but I see the verification email was received and it logged `Sent test email: verification`

Added the missing response as suggested!

> * If I try to test the breach alert email, I see `Sent test email: notification` logged on the console but I haven't received the email

If the env variable `EMAIL_TEST_RECIPIENT=localmonitor20200827@mailinator.com` is set that is not expected.

Vinnl (approved):
Florian accidentally dismissed my previous review, so re-approving, but also repeating my earlier caveat:

> With the caveats that I'm not too familiar with the codebase yet, and am not terribly experienced in back-end work, no big things jumped out to me.

src/utils/format-date.js
flozia: I find having utils for formatting dates in one place helpful. Only having to change them in one place helps with being mindful of adding formats.

rhelmer (approved):

mansaj (approved):

src/app-constants.js
```diff
@@ -8,59 +8,61 @@ import * as dotenv from 'dotenv'
...
-  'VPN_PROMO_BLOCKED_LOCALES',
-  'EDUCATION_VIDEO_URL_RELAY',
+  'GEOIP_GEOLITE2_PATH',
   'MONITOR_V2',
-  'SENTRY_DSN_LEGACY'
+  'PORT',
+  'RECRUITMENT_BANNER_LINK',
+  'RECRUITMENT_BANNER_TEXT',
+  'SENTRY_DSN_LEGACY',
+  'VPN_PROMO_BLOCKED_LOCALES'
```
toufali, line 65: praise: I love your alphabetizing skills!

src/client/js/partials/email-preview.js
```diff
@@ -0,0 +1,62 @@
...
+      method: 'POST',
+      body: JSON.stringify({
+        emailId: emailTemplateSelect.value,
+        recipient: selectedRecipient
+      })
+    })
+
+    if (response?.redirected) {
+      throw response.error
+    }
```
toufali, line 56: suggestion (non-blocking): should we let the user know if/when email was sent successfully?  A simple browser alert could suffice:
```suggestion
    }
    window.alert('Email sent!')
```

src/views/partials/email-preview.js
```diff
@@ -0,0 +1,66 @@
...
+
+export const emailPreview = data => {
+  const {
+    template: currentTemplateKey,
+    data: emailTemplates,
+    recipients
+  } = data.email
+  const selectedPreviewTemplate = emailTemplates[currentTemplateKey]?.template
+
+  return `
```
toufali, line 46: Per comment above, could we simply add the static script tag here?  It's `type="module"` so it should still be deferred and loaded as expected?

```suggestion
  return `
  <script src="/js/partials/email-preview.js" type="module"></script>
```

UPDATE: suggested change works right now via `npm run dev`, but to work for `npm start` depends on this new code in the add-email PR:
https://github.com/mozilla/blurts-server/blob/11ec0ba60e03dabff53b64d089d7f932778540bc/src/esbuild.js#L18-L21

(which will be merged shortly...)

src/views/main.js
```diff
@@ -36,6 +36,12 @@ const mainLayout = data => `
     <link rel='apple-touch-icon' href='/images/apple-touch-icon.webp' sizes='180x180'>
 
     <script src='/js/index.js' type='module'></script>
+    ${
+      data.partial.name === 'emailPreview'
+        ? '<script src="/js/partials/email-preview.js" type="module"></script>'
+        : ''
+    }
+
```
toufali, line 44: Instead of adding a conditional block into the main template for this script, why not statically add it directly in the partial?  See suggestion below.

toufali (approved):
Great work on a large endeavor @flozia!

src/views/partials/email-preview.js
toufali: Update, ☝️ this code was merged, so pulling `main` into your branch should enable the correct functionality if you decide to implement.

src/views/main.js
flozia: Yes — let’s do it like that.

src/views/partials/email-preview.js
flozia: Addressed in [8bd7bab](https://github.com/mozilla/blurts-server/pull/2790/commits/8bd7babdb6fd7baa1678823bd18d5a9cbc9f0145).

src/app-constants.js
flozia: I must admit that there might be a plugin helping me out here.

src/client/js/partials/email-preview.js
flozia: That’s a good idea until we have time to make things more user-friendly — [c8f3527](https://github.com/mozilla/blurts-server/pull/2790/commits/c8f3527e75a27a7a58636de5a0cb019f0c3a9aa0).

PR #2793, opened 2023-01-28
reset status filter on reload

toufali:
# References: 
Jira: MNTOR-1132

# Description
Firefox (and potentially other browsers) preserves form inputs on reload.  If a user has the "Resolved breaches" filter activated (a radio input), it will persist in this state through browser reload.  This can be confusing, especially since the email select input is reset to primary email.

This PR invokes the `form.reset()` function on initial load to show expected defaults on reload.

### DIFF
--- a/src/client/js/partials/breaches.js
+++ b/src/client/js/partials/breaches.js
@@ -27,6 +27,7 @@ function init () {
   resolvedCountOutput = statusFilter.querySelector("label[for='breaches-resolved'] output")
   unresolvedCountOutput = statusFilter.querySelector("label[for='breaches-unresolved'] output")
 
+  statusFilter.querySelector('form').reset() // resets status to "unresolved" on reload
   state.selectedEmail = emailSelect.value // triggers render
 
   emailSelect.addEventListener('change', handleEvent)
--- a/src/client/js/partials/breaches.js
+++ b/src/client/js/partials/breaches.js
@@ -27,7 +27,6 @@ function init () {
   resolvedCountOutput = statusFilter.querySelector("label[for='breaches-resolved'] output")
   unresolvedCountOutput = statusFilter.querySelector("label[for='breaches-unresolved'] output")
 
-  statusFilter.querySelector('form').reset() // resets status to "unresolved" on reload
   state.selectedEmail = emailSelect.value // triggers render
 
   emailSelect.addEventListener('change', handleEvent)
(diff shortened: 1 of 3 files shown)


src/views/partials/breaches.js
```diff
@@ -93,10 +93,12 @@ export const breaches = data => `
   </header>
 </section>
 <section class='breaches-filter'>
-  <input id='breaches-unresolved' type='radio' name='breaches-status' value='unresolved' checked>
-  <label for='breaches-unresolved'><output>&nbsp;</output>${getMessage('filter-label-unresolved')}</label>
-  <input id='breaches-resolved' type='radio' name='breaches-status' value='resolved'>
-  <label for='breaches-resolved'><output>&nbsp;</output>${getMessage('filter-label-resolved')}</label>
+  <form>
+    <input id='breaches-unresolved' type='radio' name='breaches-status' value='unresolved' checked>
```
flozia, line 97:
```suggestion
    <input id='breaches-unresolved' type='radio' name='breaches-status' value='unresolved' checked autocomplete='off'>
```

src/views/partials/breaches.js
```diff
@@ -93,10 +93,12 @@ export const breaches = data => `
...
+    <label for='breaches-unresolved'><output>&nbsp;</output>${getMessage('filter-label-unresolved')}</label>
+    <input id='breaches-resolved' type='radio' name='breaches-status' value='resolved'>
```
flozia, line 99:
```suggestion
    <input id='breaches-resolved' type='radio' name='breaches-status' value='resolved' autocomplete='off'>
```

src/client/js/partials/breaches.js
```diff
@@ -27,6 +27,7 @@ function init () {
   resolvedCountOutput = statusFilter.querySelector("label[for='breaches-resolved'] output")
   unresolvedCountOutput = statusFilter.querySelector("label[for='breaches-unresolved'] output")
 
+  statusFilter.querySelector('form').reset() // resets status to "unresolved" on reload
```
flozia, line 30:
```suggestion
```

flozia (approved):
That’s a pretty neat feature, but would definitely be more useful if this behavior would be the same across browsers. I checked for Chrome and Safari and they reset the state by default. Instead of resetting the form via JS we could also add `autocomplete='off'` to the input elements.

toufali:
> Instead of resetting the form via JS we could also add `autocomplete='off'` to the input elements, which prevents Firefox from preserving its state in the first place.

Good tip!  I saw that solution in my research, but initially dismissed it because I thought it didn't apply or didn't make logical sense.  But, after testing and thinking on it more, I now think your suggestion is the way to go 👍

PR #2800, opened 2023-01-31
MNTOR-983 - pass CSRF attribute to API calls

rhelmer:
# References: 
Jira: MNTOR-983

# Description

Followup to pass CSRF tokens from settings front-end for API calls.


flozia (approved):
Works great!

toufali (commented):
Isn't the token the same for the entire page?  If so, wouldn't we only need to load it once, perhaps from a parent element that we know exists on the page?  

This might also relate to the refactor ticket about detecting the parent partial once (`main[data-partial="settings"]`) vs detecting each dynamic element.

PR #2803, opened 2023-02-01
MNTOR-1115: add sentry to v2

mansaj:
# References: 
Jira: MNTOR-1115
Figma: n/a

# Description
* Adding sentry and modifying errors (by utilizing extended classes) to control noises in Sentry & provide better feedback for the frontend
* New Sentry project created, moving the old Sentry DSN to its own env var (legacy site has own sentry DSN now)
* Env vars are updated on the deployment infra side
* Going forward, we should
  * Use `throw new Error` only if it's a 500+, these errors will be seen as server side issues by Sentry and recorded
  * Any responses with statusCode 300s and 400s (using a class in `/utils/errors.js`) will NOT be picked up by Sentry. These are usually errors that the frontend would want to see

# How to test
1) In `src/controllers/landing.js`, L12, add an error: 
```
  throw new Error('Test: very serious error - server side')
```

Load the website, see the error triggered and recorded in Sentry

2) Change the error to
```
  throw new UserInputError('Test: less serious input error - client side')
```

and see that the error is not recorded in Sentry


package.json
```diff
@@ -9,7 +9,8 @@
     "@fluent/bundle": "^0.17.1",
     "@fluent/langneg": "^0.6.2",
     "@maxmind/geoip2-node": "^3.1.0",
-    "@sentry/node": "7.12.0",
+    "@sentry/node": "7.36.0",
+    "@sentry/tracing": "^7.36.0",
```
toufali, line 13: should these two be installed in the `src/package.json` file instead (or in addition)?

src/app-constants.js
```diff
@@ -43,6 +43,7 @@ const requiredEnvVars = [
   'PRODUCT_PROMOS_ENABLED',
   'REDIS_URL',
   'SENTRY_DSN',
+  'SENTRY_DSN_LEGACY',
```
toufali, line 46: Is this actually a required var, or would it work as "optional"?  

I think a missing "required" var break/blocks the site, but optional allows it to run, in the event a dev might not have the key.

If this is for the legacy site, do we even need to add it to the `src` constants?

app-constants.js
```diff
@@ -41,6 +41,7 @@ const requiredEnvVars = [
   'PRODUCT_PROMOS_ENABLED',
   'REDIS_URL',
   'SENTRY_DSN',
+  'SENTRY_DSN_LEGACY',
```
toufali, line 44: See other comment about required vs optional.  I created the "optional" category after many of these vars were already established, so many are in the required column but maybe are more optional 🤷

toufali (approved):

rhelmer (approved):

toufali:
Question RE 
> Any responses with statusCode 300s and 400s (using a class in /utils/errors.js) will NOT be picked up by Sentry

Would it be worthwhile for Sentry to still pick these up and store them so that we have something to refer to?  For example, if we wanted to know the top 404 generating url. Are we able to simply hide the noise in Sentry via GUI view settings?

rhelmer:
> Question RE
> 
> > Any responses with statusCode 300s and 400s (using a class in /utils/errors.js) will NOT be picked up by Sentry
> 
> Would it be worthwhile for Sentry to still pick these up and store them so that we have something to refer to? For example, if we wanted to know the top 404 generating url. Are we able to simply hide the noise in Sentry via GUI view settings?

Hm it would be nice to see [HTTP 429](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/429) which the rate limiting will throw, to see if users are encountering those very often in the wild.

Are there any plans to use Sentry from the front-end? That would be an alternative to catching these on the server.

mansaj:
> > Question RE
> > > Any responses with statusCode 300s and 400s (using a class in /utils/errors.js) will NOT be picked up by Sentry
> > 
> > 
> > Would it be worthwhile for Sentry to still pick these up and store them so that we have something to refer to? For example, if we wanted to know the top 404 generating url. Are we able to simply hide the noise in Sentry via GUI view settings?
> 
> Hm it would be nice to see [HTTP 429](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/429) which the rate limiting will throw, to see if users are encountering those very often in the wild.
> 
> Are there any plans to use Sentry from the front-end? That would be an alternative to catching these on the server.

@toufali @rhelmer We can whitelist specific 400 errors (429 in this case) on the server side, though I think it would be interesting to think about utilizing sentry on the frontend.. It might seem more intentional if the user errors come from the frontend, we can potentially be more selective?

package.json
mansaj: So the reason I'm putting sentry here is because V1 also needs sentry (and also should be upgraded). We can merge the two package.json when we launch v2

app-constants.js
mansaj: Fixed, thanks!

src/app.js
```diff
@@ -135,6 +162,15 @@ app.use('/api', apiLimiter)
 
 // routing
 app.use('/', indexRouter)
+
+// sentry error handler
+app.use(Sentry.Handlers.errorHandler({
+  shouldHandleError (error) {
+    if (error instanceof RateLimitError) return true
```
mansaj, line 169: FYI @rhelmer, 429 enabled in Sentry

PR #2804, opened 2023-02-01
MNTOR-1153: Prevent users from attempting to add more emails than limit

flozia:
# References: 
Jira: [MNTOR-1153](https://mozilla-hub.atlassian.net/browse/MNTOR-1153)

# Description

Disable the “Add email” button when users reached the limit for observed emails.

# Screenshot (if applicable)

image

# How to test

Reach the limit of allowed observed emails and try to add an email via the “Add email button” on the settings page

### DIFF
--- a/src/views/partials/settings.js
+++ b/src/views/partials/settings.js
@@ -151,7 +151,10 @@ export const settings = data => {
           <p>${getMessage('settings-email-limit-info', { limit })}</p>
 
           ${createEmailList(emails, breachCounts)}
-          <button class='settings-add-email-button primary js-settings-add-email-opener'>
+          <button
+            class='settings-add-email-button primary js-settings-add-email-opener'
+            ${emails.length >= AppConstants.MAX_NUM_ADDRESSES ? 'disabled' : ''}
+          >
             ${getMessage('settings-add-email-button')}
           </button>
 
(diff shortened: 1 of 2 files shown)


src/views/partials/settings.js
```diff
@@ -151,7 +151,10 @@ export const settings = data => {
           <p>${getMessage('settings-email-limit-info', { limit })}</p>
 
           ${createEmailList(emails, breachCounts)}
-          <button class='settings-add-email-button primary js-settings-add-email-opener'>
+          <button
+            class='settings-add-email-button primary js-settings-add-email-opener'
+            ${emails.length >= AppConstants.MAX_NUM_ADDRESSES ? 'disabled' : ''}
```
rhelmer, line 156: Hm, I think this isn't really possible to do via the API purely client-side, this makes sense for now. I'd like to make it possible to see how many email addresses the user has left via the API (and also make the error messages from the API more actionable) but I'll file that separately). This is simple and works fine.

rhelmer (approved):

src/views/partials/settings.js
toufali: This works!  But if you do revisit, you could also completely hide the CTA and any relevant bits by using a function similar to the one used in partials/breaches.js:
https://github.com/mozilla/blurts-server/blob/5ba2cbc4be4664e064433690967f6fc17cfae4e1/src/views/partials/breaches.js#L15-L22

And usage in the template could be less verbose, similar to: `${createEmailCTA(data.emailCount)}`

PR #2806, opened 2023-02-02
Sitewide dialog feature

toufali:
# References: 
Jira: MNTOR-1019

# Description

# Screenshot (if applicable)

Not applicable.

# How to test
Navigate to "Breaches" page and click the "Add email address" CTA in the header (assuming you have not used up all available email addresses)


src/views/partials/add-email.js
```diff
@@ -0,0 +1,13 @@
...
+  <button class='close'></button>
+</header>
+<section>
+  <p>Your account includes monitoring of up to 5 email addresses. Add a new email address to see if it’s been involved in a breach.</p>
+  <form>
+    <label>Email address<input type='email'></label>
+    <button class='primary' type='submit'>Send verification link</button>
+  </form>
+</section>
+`
```
flozia, line 13: Praise: That’s cool that we’ll have that level of flexibility with content.

src/client/css/global.css
```diff
@@ -189,23 +187,58 @@ button.primary,
...
+  color: var(--blue-50);
+  border: 1px solid;
+  padding: var(--padding-md);
+}
+
+button.primary:active,
+.button.primary:active,
+button.secondary:active,
+.button.secondary:active {
+  box-shadow: inset 0 0 64px #80808011;
```
flozia, line 201: Question (non-blocking): Should we create CSS variables for shadows so we can try to reuse them as much as possible instead of creating new ones in the future?

src/client/js/dialog-handler.js
```diff
@@ -0,0 +1,62 @@
...
+      closeDialog()
+      break
+  }
+}
+
+async function openDialog (partialName) {
+  dialogEl.showModal() // provide immediate UI response by showing ::backdrop regardless of dialog content
+  dialogEl.addEventListener('click', handleEvent)
+
+  const content = contentMap.get(partialName) ?? await fetchContent(partialName)
```
flozia, line 32: Question (non-blocking): Let’s assume the user is on a very slow or flacky connection they just would not get any form of feedback — does a loading state for dynamic content would make sense here?

flozia (commented):
This is already working really well. There are two things related to dismissing the dialog that I found:

1. Clicking the dialog element dismisses it. Only clicking the backdrop and “close” button should trigger that.
2. Closing the dialog using the `esc` results in duplicate content (see screenshot below) when reopening it.

image

src/client/css/global.css
toufali: Yes!  But, for this case it's quite unique -- I'm just trying to cover the inside of the button with an overlay, not so much a shadow.  This "shadow" is unlikely to be reused.

src/client/js/dialog-handler.js
toufali: Yes, a loading state for dynamic content makes sense.  We can/should add that – I'm not sure if we have any loading animation assets from UX yet.  As you mention, this would only be visible for very slow connections, since we're displaying the HTML immediately – images and other media assets would load afterwords.

That being said, the current logic _should_ already be built to provide immediate feedback as mentioned in the comment.  (Let me know if that comment isn't clear enough, or maybe I don't understand.)  As currently programmed, the backdrop (an 80% opacity black scrim) will appear immediately regardless of available content.  If/when content is loaded, that content will fade in over the backdrop.  By "loaded", I mean just the HTML, which should be fast.  Images would load subsequently.  If content fails to load, the scrim is expected to disappear to avoid blocking the user.

toufali:
> This is already working really well. There are two things related to dismissing the dialog that I found:
> 
>     1. Clicking the dialog element dismisses it. Only clicking the backdrop and “close” button should trigger that.
> 
>     2. Closing the dialog using the `esc` results in duplicate content (see screenshot below) when reopening it.
> 

Wow these are great finds, thank you!  The backdrop click worked initially when I tested, but my tests must have included clicks on actual modal content.  It looks like if you click non-content (in between content or in padding) the modal is dismissed.  The escape button find I had completely missed – way to go!

I pushed a fix for both of these issues:
- I removed click-outside dismissal functionality.  Click-outside is likely unwanted behavior anyway.  The main reason I initially included it was because the backdrop is displayed immediately and a touchscreen user (without escape button) might not have a way to dismiss if the content doesn't load.  I've since added error handling though, which will auto-close the modal if content doesn't load.
- pushing escape and loading the modal again should now be handled properly – thanks again!

src/client/js/dialog-handler.js
flozia: That comment was really just directed toward dynamic content within the modal — thanks for elaborating. In that case we should be good.

flozia (approved):

src/views/partials/add-email.js
```diff
@@ -0,0 +1,13 @@
+export const addEmail = data => `
```
flozia, line 1:
```suggestion
/* This Source Code Form is subject to the terms of the Mozilla Public
 * License, v. 2.0. If a copy of the MPL was not distributed with this
 * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
 export const addEmail = data => `
```

PR #2807, opened 2023-02-02
MNTOR-1143 - limit how often users can verify email addresses to 5 minutes, with generic rate limiting for all APIs

rhelmer:
This uses a postgres-only knex raw query in order to compare timestamps on the database side.

# References: 
Jira: MNTOR-1143

# Description

Limit how often users can cause verification to be sent to a particular email address to 5 minutes.

# Screenshot (if applicable)

Not applicable.

# How to test

1. Adding new secondary email addresses should immediately send a new verification email
2. Click "Resend verification email" within 5 minutes and confirm that API returns error
3. Click "Resend verification email" again after 5 minutes, should send a new verification email

### DIFF
--- a/src/controllers/settings.js
+++ b/src/controllers/settings.js
@@ -20,6 +20,7 @@ import { sendEmail, getVerificationUrl, getUnsubscribeUrl } from '../utils/email
 
 import { getBreachesForEmail } from '../utils/hibp.js'
 import { generateToken } from '../utils/csrf.js'
+import { RateLimitError } from '../utils/error.js'
 
 import { mainLayout } from '../views/main.js'
 import { settings } from '../views/partials/settings.js'
@@ -148,27 +149,35 @@ async function resendEmail (req, res) {
 }
 
 async function sendVerificationEmail (emailId) {
-  const unverifiedEmailAddressRecord = await resetUnverifiedEmailAddress(
-    emailId
-  )
-  const recipientEmail = unverifiedEmailAddressRecord.email
-  const data = {
-    recipientEmail,
-    ctaHref: getVerificationUrl(unverifiedEmailAddressRecord),
-    utmCampaign: 'email_verify',
-    unsubscribeUrl: getUnsubscribeUrl(
-      unverifiedEmailAddressRecord,
-      'account-verification-email'
-    ),
-    heading: getMessage('email-verify-heading'),
-    subheading: getMessage('email-verify-subhead'),
-    partial: { name: 'verify' }
+  try {
+    const unverifiedEmailAddressRecord = await resetUnverifiedEmailAddress(
+      emailId
+    )
+    const recipientEmail = unverifiedEmailAddressRecord.email
+    const data = {
+      recipientEmail,
+      ctaHref: getVerificationUrl(unverifiedEmailAddressRecord),
+      utmCampaign: 'email_verify',
+      unsubscribeUrl: getUnsubscribeUrl(
+        unverifiedEmailAddressRecord,
+        'account-verification-email'
+      ),
+      heading: getMessage('email-verify-heading'),
+      subheading: getMessage('email-verify-subhead'),
+      partial: { name: 'verify' }
+    }
+    await sendEmail(
+      recipientEmail,
+      getMessage('email-subject-verify'),
+      getTemplate(data, verifyPartial(data))
+    )
+  } catch (err) {
+    if (err.message === 'error-email-validation-pending') {
+      throw new RateLimitError('Verification email recently sent, try again later')
+    } else {
+      throw err
+    }
   }
-  await sendEmail(
-    recipientEmail,
-    getMessage('email-subject-verify'),
-    getTemplate(data, verifyPartial(data))
-  )
 }
 
 async function verifyEmail (req, res) {
--- a/src/db/tables/email_addresses.js
+++ b/src/db/tables/email_addresses.js
@@ -54,6 +54,20 @@ async function addSubscriberUnverifiedEmailHash (user, email) {
 
 async function resetUnverifiedEmailAddress (emailAddressId) {
   const newVerificationToken = uuidv4()
+
+  // Time in ms to require between verification reset.
+  const verificationWait = 5 * 60 * 1000 // 5 minutes
+
+  const verificationRecentlyUpdated = await knex('email_addresses')
+    .select('id')
+    .whereRaw('"updated_at" > NOW() - INTERVAL \'1 MILLISECOND\' * ?', verificationWait)
+    .andWhere('id', emailAddressId)
+    .first()
+
+  if (verificationRecentlyUpdated?.id === parseInt(emailAddressId)) {
+    throw fluentError('error-email-validation-pending')
+  }
+
   const res = await knex('email_addresses')
     .update({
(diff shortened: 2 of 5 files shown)


src/db/tables/email_addresses.js
```diff
@@ -54,6 +54,20 @@ async function addSubscriberUnverifiedEmailHash (user, email) {
...
+  // Time in ms to require between verification reset.
+  const verificationWait = 5 * 60 * 1000 // 5 minutes
+
+  const verificationRecentlyUpdated = await knex('email_addresses')
+    .select('id')
+    .whereRaw(`"updated_at" > NOW() - INTERVAL '${verificationWait} MILLISECONDS'`)
+    .andWhere('id', emailAddressId)
+    .first()
+
+  if (verificationRecentlyUpdated?.id === parseInt(emailAddressId)) {
```
rhelmer, line 67: It would be nice if knex supported `SELECT EXISTS` for this kind of thing, but we'd have to use a completely raw query and I'm trying to minimize that.

src/db/tables/email_addresses.js
```diff
@@ -54,6 +54,20 @@ async function addSubscriberUnverifiedEmailHash (user, email) {
 
 async function resetUnverifiedEmailAddress (emailAddressId) {
   const newVerificationToken = uuidv4()
+
+  // Time in ms to require between verification reset.
+  const verificationWait = 5 * 60 * 1000 // 5 minutes
+
+  const verificationRecentlyUpdated = await knex('email_addresses')
+    .select('id')
+    .whereRaw('"updated_at" > NOW() - INTERVAL \'1 MILLISECOND\' * ?', verificationWait)
```
rhelmer, line 63: It's hard to find a `where` clause here that:
- looks nice
- is cross-database (doesn't use `raw`)
- handles parameterized queries
- compares only timestamps from the DB, not the app server

I ended up trading off the "looks nice" and "cross-database" in favor of the other points, which are more about safety and correctness. We obviously control `verificationWait` in this patch, but using template strings here just to make it look a little nicer seems like it could set a trap for someone thinking it'd be safe to make that variable somehow be user-controlled.

rhelmer:
As discussed, I also added generic rate limiting for all `/api` calls as well, the targeted fix here prevents users from accidentally or purposely pressing the "re-send verification" email over and over, but without the generic limiting there is nothing stopping someone from scripting add/remove/re-add via the API to do the same thing.

We want to give some leeway for users to make mistakes and re-do things, but have a way to prevent excessive use of the APIs. I've used the suggested defaults in the docs for the `express-rate-limiter` package: 100 API calls per 15 minutes per IP.

Let's keep an eye on API usage and adjust this as necessary.

src/controllers/settings.js
```diff
@@ -138,13 +151,26 @@ async function resendEmail (req, res) {
...
   }
 
-  await sendVerificationEmail(emailId)
-
-  return res.json({
-    success: true,
-    status: 200,
-    message: 'Sent the verification email'
-  })
+  try {
```
mansaj, line 154: Is there a way to not dup the same code as line 93?

src/db/tables/email_addresses.js
mansaj: Makes sense!

src/db/tables/email_addresses.js
```diff
@@ -54,6 +54,20 @@ async function addSubscriberUnverifiedEmailHash (user, email) {
 
 async function resetUnverifiedEmailAddress (emailAddressId) {
   const newVerificationToken = uuidv4()
+
+  // Time in ms to require between verification reset.
+  const verificationWait = 5 * 60 * 1000 // 5 minutes
```
mansaj, line 59: Interesting tradeoff to consider, what if we make this into a string

`${5 * 50 * 1000} MILLISECOND`

and pass that directly into the whereRaw below, is that cleaner? 2x calculations vs. 1 here?

rhelmer: Hm yeah I actually did something like that originally (I think I put `5 MINUTES` in the constant, Postgres has a very flexible parser for intervals.

I thought we might end up moving that constant out of the function, or making it an argument etc, so I worried about a future JS developer seeing something that isn't a `Number` representing milliseconds in a constant and getting confused. Maybe I'm overthinking.

rhelmer: I mean one thing we could do here is just forget about using a parameter and just hardcode `INTERVAL 15 MINUTES`? Unless we plan to pull the constant out of this function then it probably is clearer anyway.

src/controllers/settings.js
rhelmer: Yeah this is what I was complaining about re: the way we're doing error responses right now... I'll figure out some way to do so.

src/db/tables/email_addresses.js
mansaj: Got it.. hmm I think the current way is better than `INTERVAL 15 MINUTES` :D

src/controllers/settings.js
mansaj: Is it worth it to wrap L154-173 into a function?

rhelmer: Hm actually in the `addEmail` case this can't happen because the record won't be present, I'll try a couple different options and see which is the least bad.

src/controllers/settings.js
```diff
@@ -138,39 +132,51 @@ async function resendEmail (req, res) {
     throw fluentError('user-verify-token-error')
   }
 
-  await sendVerificationEmail(emailId)
+  await sendVerificationEmail(res, emailId)
```
rhelmer, line 135: OK I made this less repetitive, but this approach makes `sendVerificationEmail` take `res` as an argument and it's now responsible for reporting success or failure... it seems like we must be able to have a better pattern than this though.

mansaj: yes.. not sure if this is better or worse...

mansaj (approved):
Up to you

src/controllers/settings.js
```diff
@@ -138,39 +132,51 @@ async function resendEmail (req, res) {
...
+      partial: { name: 'verify' }
+    }
+    await sendEmail(
+      recipientEmail,
+      getMessage('email-subject-verify'),
+      getTemplate(data, verifyPartial(data))
+    )
+  } catch (err) {
+    if (err.message === 'error-email-validation-pending') {
+      res.status(429)
```
mansaj, line 163: Here you can use `RateLimitError` class and it'll set the json automatically

rhelmer: Oh this is great, thanks!

src/controllers/settings.js
rhelmer: Thanks, I'll research this a bit on Monday... there's probably a better way to do this that I'm not thinking of.

rhelmer: Oh, after you pointed out the existence of that `RateLimitError` class, I realized we could use that here as well - if the controller doesn't catch the exception from `sendVerificationEmail()`, then the default error handler does the right thing.

I put just the success JSON response back in. I bet we could simplify these too somehow, but I like the idea of the API route functions being in charge of the `req` and `res` objects and not delegating that to a function.

PR #2818, opened 2023-02-07
MNTOR-1165: fix primary email case

mansaj:
# References: 
Jira: MNTOR-1165

# Description
On the Breaches page, with my primary email selected, when marking a resolution step complete, the PUT request fails. We were not accounting for the primary email case

# How to test
1) create a new Monitor account with primary email that has known breaches
2) visit the breaches page and expand a breach
3) tick or untick a resolution step

### DIFF
--- a/src/controllers/breaches.js
+++ b/src/controllers/breaches.js
@@ -56,19 +56,24 @@ async function putBreachResolution (req, res) {
   const sessionUser = req.user
   const { affectedEmail, breachId, resolutionsChecked } = req.body
   const breachIdNumber = Number(breachId)
-  const affectedEmailIsSubscriberRecord = sessionUser.primary_email === affectedEmail
-  const affectedEmailInEmailAddresses = sessionUser.email_addresses.filter(ea => ea.email === affectedEmail)
+  const affectedEmailAsSubscriber = sessionUser.primary_email === affectedEmail ? sessionUser.primary_email : false
+  const affectedEmailInEmailAddresses = sessionUser.email_addresses.filter(ea => ea.email === affectedEmail)?.[0]?.email || false
 
   // check if current user's emails array contain affectedEmail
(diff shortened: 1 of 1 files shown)


src/controllers/breaches.js
```diff
@@ -56,19 +56,24 @@ async function putBreachResolution (req, res) {
   const sessionUser = req.user
   const { affectedEmail, breachId, resolutionsChecked } = req.body
   const breachIdNumber = Number(breachId)
-  const affectedEmailIsSubscriberRecord = sessionUser.primary_email === affectedEmail
-  const affectedEmailInEmailAddresses = sessionUser.email_addresses.filter(ea => ea.email === affectedEmail)
+  const affectedEmailAsSubscriber = sessionUser.primary_email === affectedEmail ? sessionUser.primary_email : false
```
toufali, line 59: Was the const name supposed to be `affectedEmailIsSubscriber` (is instead of as?)

src/controllers/breaches.js
```diff
@@ -56,19 +56,24 @@ async function putBreachResolution (req, res) {
...
+  const affectedEmailInEmailAddresses = sessionUser.email_addresses.filter(ea => ea.email === affectedEmail)?.[0]?.email || false
```
toufali, line 60: Would `Array.find()` be more appropriate here?
```suggestion
  const affectedEmailInEmailAddresses = sessionUser.email_addresses.find(ea => ea.email === affectedEmail)?.email || false
```

toufali (approved):
A couple non-blocking comments.  Thanks for the quick turnaround!

src/controllers/breaches.js
mansaj: So I changed the purpose of this var from a `boolean` type to a `(string | boolean)`. If email exists, then it becomes the email. This way we save a step later when referencing the email

src/controllers/breaches.js
mansaj: Awesome

PR #2821, opened 2023-02-09
Add empty state for the breach list

Vinnl:
# References: 
Jira: MNTOR-1086
Figma: https://www.figma.com/file/gvyyqS9egsGjsH0LUDqT1m/Breach-Resolution-Flow-Design?node-id=654%3A16313&t=I1S7G3TnYG1IGh0z-4

# Description

This adds an empty state for the list of breaches. The acceptance criteria also mentioned telemetry, but I assume that that is depending on @rhelmer's work?

# Screenshot (if applicable)

image

Screen Shot 2023-02-09 at 15 05 49

image

Screen Shot 2023-02-09 at 15 05 37

# How to test

1. Open the breaches page for an email address with zero breaches - it should show the "No breaches found" message and hide the "Unresolved / Resolved" tabs.
2. Open the breaches page for an email address with more than zero unresolved breaches - it should just list those breaches.
3. Open the breaches page for an email address with more than zero resolved, and no unesolved breaches - it should show the "All breaches resolved" under the "Unresolved" tab, and list all breaches under the "Resolved" tab.
4. 1 and 3 above should show an "Add email address" button if the user has not reached their email address quota. Once they do, it should be hidden.

- [ ] I've added a unit test to test for potential regressions of this bug. - Don't know how to - do we have unit tests for UI?
- [ ] All acceptance criteria are met. - No telemetry, strings haven't been localised yet, and a couple of open questions inline.


src/views/partials/breaches.js
```diff
@@ -76,6 +76,26 @@ function createResolveSteps (breach) {
...
+function createAddEmailButton (data, status) {
+  if (data.emailCount >= AppConstants.MAX_NUM_ADDRESSES) {
+    return ''
+  }
+
+  return `
+    <p>
+      ${getMessage(`breaches-${status}-cta-blurb`)}
+    </p>
+    <button class="primary" data-dialog='add-email'>
```
Vinnl, line 93: I just reused the `primary` class, but in the designs, it uses the Metropolis font and a font weight of 600. Should I update the `primary` class, or change it just for this instance?

src/client/css/partials/breaches.css
```diff
@@ -252,3 +256,41 @@
...
+.no-breaches-message p,
+.all-breaches-resolved-message p {
+  font: var(--h3-font);
+  font-weight: 500;
+  max-width: 700px;
+}
+
+.no-breaches-message p b,
+.all-breaches-resolved-message p b {
+  font-weight: 700;
```
Vinnl, line 295: These and the font settings for the `p`s above don't seem to have any effect. I expect that the (Metropolis) fonts aren't initialised for these weights? Was that a conscious decision?

src/views/partials/breaches.js
```diff
@@ -118,5 +138,35 @@ export const breaches = data => `
...
   ${createBreachRows(data.breachesData)}
+  <div class="no-unresolved-breaches-message">
+    <div class="no-breaches-message">
+      <img src='/images/breaches-none.svg' alt='' width="136" height="102" />
+      <h2>
+        ${getMessage('breaches-none-headline')}
+      </h2>
+      <p>
+        ${
+          data.breachesData.verifiedEmails.map(account => {
```
Vinnl, line 149: Note that I'm just including the same string here multiple times - one for every of the user's email address. This is because I need to be able to toggle between the different strings on the client side, when the user selects a different email address, but of course becomes pretty cumbersome (or in some cases, even impossible) if we have to do this for every variable that changes on the client side. Do we have better solutions for this?

Vinnl (commented):
Added some of my open questions inline. Other open question is whether I need to add telemetry already and if so, how?

locales/en/breaches.ftl
```diff
@@ -33,6 +33,20 @@ column-status-badge-resolved = Resolved
...
+breaches-none-copy = Good news! No known breaches were reported for { $email }. We’ll keep monitoring this email and will let you know if any new breaches occur.
+breaches-none-cta-blurb = Would you like to monitor another email?
+breaches-none-cta = Add email address
+
+breaches-all-resolved-headline = All breaches resolved
+# Variables:
+#   $email (String) - An email address for which all breaches have been resolved, e.g. `someone@example.com`
+breaches-all-resolved-copy = Nicely done! You’ve resolved all breaches for { $email }. We’ll keep monitoring this email and will let you know if any new breaches occur.
+breaches-all-resolved-cta-blurb = Would you like to monitor another email?
+breaches-all-resolved-cta = Add email address
```
flodolo, line 48:
```suggestion
breaches-all-resolved-cta-button = Add email address
```

It's nice to know from the ID where the string is used. Same for `breaches-none-cta`.

flodolo (approved):

src/views/partials/breaches.js
```diff
@@ -118,5 +138,35 @@ export const breaches = data => `
...
+            return `<span data-email="${account.email}" ${account.primary ? '' : 'hidden'}>${getMessage('breaches-none-copy', { email: `<b>${account.email}</b>` })}</span>`
```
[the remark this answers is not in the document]
flozia: Do you mean another way than directly putting them in the l10n strings or like the `<a></a>` tag for the string with ID `breach-checklist-pw-header` in `locales/en/breaches.ftl`?

src/views/partials/breaches.js
flozia: We probably want to stay consistent with the buttons. I’d advocate for reusing the `primary` class for now and we can leave a comment for UX to see if they can adjust the design accordingly or otherwise change it respectively.

src/client/css/partials/breaches.css
flozia: Only Metropolis Bold `700` is initialized in `src/client/css/fonts.css`. I guess that’s probably just due to the zero state using it for the first time?

src/client/css/partials/breaches.css
```diff
@@ -252,3 +256,41 @@
...
+.breaches-table .no-unresolved-breaches-message > .no-breaches-message,
+.breaches-table .no-unresolved-breaches-message > .all-breaches-resolved-message {
+  display: none;
+}
+
+.breaches-table[data-nr-unresolved='0'][data-nr-resolved='0'] .no-unresolved-breaches-message .no-breaches-message {
+  display: flex;
+}
+
+.breaches-filter:not([data-selected='resolved']) + .breaches-table[data-nr-unresolved='0']:not([data-nr-resolved='0']) .no-unresolved-breaches-message .all-breaches-resolved-message {
```
flozia, line 269: Nit (non-blocking): Would be nice to have a comment on what state we are targeting here.

flozia (commented):
This is working great!

I replied to your inline comments. The only thing I think should be addressed before landing this PR is adding the missing Metropolis weights and possible changes depending on what decision we land on regarding the multiple strings.

src/views/partials/breaches.js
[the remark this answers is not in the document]
Vinnl: Ah yes, I guess you can include them directly like that `<a>`, but (for other use cases) I'd like to be able to set its attributes, potentially dynamically. In Fluent's React bindings, for example, you can pass an `elems` argument that looks somewhat like this:

```jsx
elems: {
  a: <a href={`https://some-dyamic.link/${userId}`} className="some-class"/>,
},
```

which results in `<a>` tags in the source string get replaced by the given element.

src/client/css/partials/breaches.css
Vinnl: Ah gotcha. I added them here, of course at the cost of two additional resources to fetch: 97351dbf

locales/en/breaches.ftl
Vinnl: Heh, I actually went back and forth on that :sweat_smile: Changed in 731b7a05.

src/client/css/fonts.css
```diff
@@ -197,6 +197,26 @@ https://fonts.googleapis.com/css2?family=Inter:wght@400;600&display=swap
...
+@font-face {
+  font-family: Metropolis;
+  font-style: normal;
+  font-weight: 600;
+  font-display: optional;
+  src:
+    local('Metropolis-SemiBold'),
+    url('../fonts/Metropolis-SemiBold.woff2') format('woff2');
+}
+
```
toufali, line 219: Sorry for the lack of context RE fonts:
It was discussed that we try to keep Refresh fonts to a minimum for simplicity and page-load considerations. Inter should have a regular and bold weight, Metropolis (which coincidentally is no longer supported) should have a bold weight. Please try to conform to these options and let UX know during review. If there is UX concern we can add new fonts.

src/client/css/partials/breaches.css
toufali: Ah, I was late to this party.  Yes it was a conscious decision – we agreed with UX that we would try to conform to a limited set of fonts to keep it simple and avoid extra client loads.

Same with heading and paragraph sizes – let's conform to the closest established variable and get UX sign-off.

For this case since it was already merged, I can chat with UX and post a follow-up PR with any adjustments.

src/views/partials/breaches.js
toufali: Yes, as my other comments have trended – instead of creating too many new unique styles, default to using our established vars for consistency and get UX sign-off during review.

src/views/partials/breaches.js
toufali: Yeah, I see the difficulty.  I posted a follow up and tagged you for review: https://github.com/mozilla/blurts-server/pull/2828

I think using fetch for the dynamic content as @flozia suggested would have worked too, but in this case we were already tracking the selected email via module state, so it wasn't much more of a stretch to inject dynamically.

The client/partials/breaches.js module needed touch up anyway (and will need more after [MNTOR-1160/add-new-email](https://github.com/mozilla/blurts-server/tree/MNTOR-1160/add-new-email) lands).  Using `<template>` was not required, but I opted for it instead of dealing with toggling between div `hidden` states.

src/client/css/fonts.css
Vinnl: Addressed in https://github.com/mozilla/blurts-server/pull/2825#discussion_r1104363695.

PR #2822, opened 2023-02-09
Use relative unit for font size

Vinnl:
See https://github.com/mozilla/blurts-server/pull/2817#discussion_r1100549633.

### DIFF
--- a/src/client/css/partials/breaches.css
+++ b/src/client/css/partials/breaches.css
@@ -186,7 +186,7 @@
      so that everything remains vertically-aligned, and the badge still
      disappears when getting flex-wrapped on small screens.
   */
-  font-size: 0.75rem;
+  font-size: 0.6875rem;
 }
 
 .resolution-badge.is-active {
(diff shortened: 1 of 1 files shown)


src/client/css/partials/breaches.css
```diff
@@ -186,7 +186,7 @@
      so that everything remains vertically-aligned, and the badge still
      disappears when getting flex-wrapped on small screens.
   */
-  font-size: 11px;
+  font-size: 0.75rem;
```
toufali, line 189: It's an ugly number, but should this be `0.6875rem` to "equal" 11px?  Alternatively, I could ask UX if they might stick to even-number font-sizes...

toufali, line 189: It's an ugly number, but should this be 0.6875rem to "equal" 11px? 

```suggestion
  font-size: 0.6875rem;
```

Alternatively, I could ask UX if they might stick to even-number font-sizes...

Vinnl: That works for me, though honestly, I feel like it small deviations like this should be within our purview. After all, the entire point of the relative units is that they might _not_ resolve to exactly a certain number of pixels :grin:

toufali (approved):

PR #2823, opened 2023-02-09
Use Volta to align toolchains

Vinnl:
This ensures that anyone with [Volta](https://volta.sh/) installed is running the same versions of Node and npm.

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,7 +20,7 @@ the "what" and "why" of data breach alerts.
 
 ### Requirements
 
-* [Node](https://nodejs.org/) (with NPM): we recommend using [Node Version Manager](https://github.com/nvm-sh/nvm) to install and manage Node/NPM.
+* [Volta](https://volta.sh/) (installs the correct version of Node and npm)
 * [Postgres](https://www.postgresql.org/)
 
 ### Code style
--- a/package.json
+++ b/package.json
@@ -119,5 +119,9 @@
     ],
     "ext": "js,css,hbs,json,ftl"
   },
-  "supportedLocales": "cak,cs,cy,da,de,el,en,en-CA,en-GB,es-AR,es-CL,es-ES,es-MX,fi,fr,fy-NL,gn,hu,kab,ia,id,it,ja,nb-NO,nl,nn-NO,pt-BR,pt-PT,ro,ru,sk,sl,sq,sv-SE,tr,uk,vi,zh-CN,zh-TW"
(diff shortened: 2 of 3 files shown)


flozia (approved):
That’s neat!

mansaj (approved):
FYI perf study on Volta vs. NVM:
https://duncanlew.medium.com/a-better-node-version-manager-volta-vs-nvm-5917c1edd0a1

README.md
```diff
@@ -20,7 +20,7 @@ the "what" and "why" of data breach alerts.
 
 ### Requirements
 
-* [Node](https://nodejs.org/) (with NPM): we recommend using [Node Version Manager](https://github.com/nvm-sh/nvm) to install and manage Node/NPM.
+* [Volta](https://volta.sh/) (installs the correct version of Node and npm)
```
toufali, line 23: suggestion: should this be added under a new "Recommended tools" section?  Something like Postgres is required, but Volta is not.  Also, I have some other recommendations to add, should such a section exist 😉

toufali (approved):

package.json
```diff
@@ -119,5 +119,9 @@
     ],
     "ext": "js,css,hbs,json,ftl"
   },
-  "supportedLocales": "cak,cs,cy,da,de,el,en,en-CA,en-GB,es-AR,es-CL,es-ES,es-MX,fi,fr,fy-NL,gn,hu,kab,ia,id,it,ja,nb-NO,nl,nn-NO,pt-BR,pt-PT,ro,ru,sk,sl,sq,sv-SE,tr,uk,vi,zh-CN,zh-TW"
+  "supportedLocales": "cak,cs,cy,da,de,el,en,en-CA,en-GB,es-AR,es-CL,es-ES,es-MX,fi,fr,fy-NL,gn,hu,kab,ia,id,it,ja,nb-NO,nl,nn-NO,pt-BR,pt-PT,ro,ru,sk,sl,sq,sv-SE,tr,uk,vi,zh-CN,zh-TW",
+  "volta": {
+    "node": "18.12.1",
+    "npm": "8.19.3"
+  }
```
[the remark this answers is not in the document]
Vinnl: I haven't seen any mention of that, but the two reasons I would expect are:

1. `engines` [doesn't necessarily resolve to a specific version](https://docs.npmjs.com/cli/v6/configuring-npm/package-json#engines), i.e. you can specify a range of versions that the project is compatible with.
2. They serve different purposes, i.e. as a hint about compatibility for downstreams vs. aligning contributors on a specific version for consistent reproducible experiences.

I do agree that it would be nice to not have to update these in multiple places. My suggestion therefore would be to remove the `engines` field, since that's more aimed at downstreams - which we don't have, since we're not a library. What do you think?

README.md
Vinnl: Yeah, that sounds like a good idea - Volta might not be required, but using the same version of Node and particularly npm is. Would you mind submitting a separate PR with those recommended tools? Then we can iterate on that list in that PR, and just replace this with something like `the correct version of Node and npm - we [recommend](#recommended-tools) installing Volta to take care of this`?

PR #2825, opened 2023-02-10
Align navigation with design, remove unused links

Vinnl:
# References: 
Jira: MNTOR-1169
Figma: https://www.figma.com/file/gvyyqS9egsGjsH0LUDqT1m/Breach-Resolution-Flow-Design?node-id=214%3A9127&t=sEAfwuHATkggJONz-0

# Description

This splits the left-hand side navigation into a primary and secondary part, and removes unused links. It also adds an icon to the "Resolve data breaches" link, and otherwise updates the styling to more closely match the design.

I encountered another missing font, so I copied that over from Relay and added that as yet another resource to load.

# Screenshot (if applicable)

image

Screen Shot 2023-02-10 at 18 25 27

# How to test

View the navigation. Resize your screen to make sure it looks good on the relevant screen sizes.

- [ ] I've added a unit test to test for potential regressions of this bug. - N/A, layout changes


src/client/js/nav.js
```diff
@@ -18,7 +18,6 @@ function handleEvent (e) {
       nav.toggleAttribute('hidden')
       break
     case e.target.matches('.nav-item'):
-      e.target.style.setProperty('color', 'var(--purple-70)')
```
Vinnl, line 21: I'm not sure why we wanted to set the colour of a link to purple when you click it, but now that the hover style has a purple background, that would result in purple-in-purple, so I removed it.

flodolo (approved):

src/views/main.js
```diff
@@ -118,11 +118,27 @@ const mainHeader = data => `
...
+      <svg width="24" height="24" viewBox="0 0 24 24" fill="none" xmlns="http://www.w3.org/2000/svg">
+        <path fill-rule="evenodd" clip-rule="evenodd" d="M10.5942 20.049C9.87439 21.3816 10.8394 22.9996 12.3539 22.9996H19.657C21.1692 22.9996 22.1344 21.3862 21.4193 20.0538L17.7796 13.2724C17.0264 11.8689 15.0148 11.8662 14.2577 13.2676L10.5942 20.049Z" fill="white" stroke="currentcolor" stroke-width="2" stroke-linecap="round" stroke-linejoin="round"/>
+        <path fill-rule="evenodd" clip-rule="evenodd" d="M16 21C16.5523 21 17 20.5523 17 20C17 19.4477 16.5523 19 16 19C15.4477 19 15 19.4477 15 20C15 20.5523 15.4477 21 16 21Z" fill="currentcolor"/>
+        <path d="M16 17V16" stroke="currentcolor" stroke-width="2" stroke-linecap="round"/>
+        <path d="M7 22H5C3.89543 22 3 21.1046 3 20V11C3 9.89543 3.89543 9 5 9H19C20.1046 9 21 9.89543 21 11V13" stroke="currentcolor" stroke-width="2" stroke-linecap="round" stroke-linejoin="round"/>
+        <path d="M7 9V7C7 4.23858 9.23858 2 12 2C14.7614 2 17 4.23858 17 7V9" stroke="currentcolor" stroke-width="2" stroke-linecap="round" stroke-linejoin="round"/>
+      </svg>
```
toufali, line 130: Would it be better to load this as an image to get a cache benefit – in particular since this nav is loaded with every page view?

src/client/css/nav.css
```diff
@@ -2,40 +2,63 @@
...
-nav > a {
-  padding: var(--padding-sm) var(--padding-lg);
-  text-decoration: none;
+nav.site-nav .pages-nav,
+nav.site-nav .meta-nav {
+  display: flex;
+  flex-direction: column;
+  align-items: stretch;
+  justify-content: space-evenly;
+  gap: var(--padding-sm);
```
flozia, line 29: Question: What do you think of not using `gap` to space the link items and instead giving `nav-item` a higher vertical padding? That would increase the clickable area and visually align the hover highlight with the height of the “active indicator” (the stripe purple stripe on the left side).

src/views/main.js
```diff
@@ -118,11 +118,27 @@ const mainHeader = data => `
...
+      ${getMessage('site-nav-breaches-link')}
+    </a>
+  </div>
+  <div class='meta-nav'>
+    <a href='/user/settings' class='nav-item ${data.partial.name === 'settings' ? 'current' : ''}'>
+      ${getMessage('site-nav-settings-link')}
+    </a>
+    <a target="_blank" rel="noopener noreferrer" href='https://support.mozilla.org/kb/firefox-monitor' class='nav-item'>
```
toufali, line 138: `rel="noopener noreferrer"` is already handled via CSP – I think this attribute can be safely removed for all cases.

flozia (commented):
This is already looking good! There seems to be an issue when following external links. When going back the nav looses it’s active page indicator.

https://user-images.githubusercontent.com/13835474/218166341-ea9ddedd-136e-4ede-bf68-31a881c02a9e.mov

src/views/main.js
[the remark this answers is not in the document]
flozia: Seems like we’ve been reviewing simultaneously @toufali. Just saw that this is a [duplicate comment](https://github.com/mozilla/blurts-server/pull/2825#discussion_r1103063121) — resolving so we do have a conversation in two places.

toufali (commented):
This is looking much better than before, thanks!

On the responsive front, it's looking like the nav takes up more space than needed:
Screen Shot 2023-02-10 at 10 38 58 AM

toufali:
> This is already looking good! There seems to be an issue when following external links. When going back the nav looses it’s active page indicator.
> nav-pr-2825.mov

Good catch @flozia – this is likely a shortcoming of the original build; it did not account for external links.  Since this ticket was more about style, feel free to create a separate bug ticket and assign me if you prefer @Vinnl

src/client/js/nav.js
Vinnl: Links should not only be distinguishable by colour so that people who have trouble distinguishing similar colours are able to recognise them. Additionally, the background colour makes it easier to recognise the clickable area, which helps users with motor impairments.

The Figma indeed did not include designs for the hover styles, so similar to when it does not include responsive designs, I came up with something myself. I will call it out in the desk check.

src/client/css/nav.css
```diff
@@ -2,40 +2,63 @@
...
+nav.site-nav a.current {
   color: var(--purple-70);
   animation: nav-indicator 0.2s ease-out forwards;
 }
 
-.mobile nav {
+nav.site-nav a:not(.current):hover {
+  background-color: var(--purple-70);
+  color: white;
+}
```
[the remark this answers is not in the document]
Vinnl: See https://github.com/mozilla/blurts-server/pull/2825#discussion_r1104362890.

locales/en/app.ftl
```diff
@@ -823,6 +823,12 @@ meta-desc = Find out if you’ve been part of a data breach with {-brand-fx-moni
 brand-fx-monitor = {-brand-fx-monitor}
 sign-in = Sign In
 
+## Site navigation
+
+site-nav-breaches-link = Resolve Data Breaches
+site-nav-settings-link = Settings
+site-nav-help-link = Help & Support
```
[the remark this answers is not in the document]
Vinnl: Thanks both, fixed in f0b1dd7c.

src/client/css/fonts.css
```diff
@@ -105,6 +105,14 @@ https://fonts.googleapis.com/css2?family=Inter:wght@400;600&display=swap
...
 }
 
+@font-face {
+  font-family: Inter;
+  font-style: normal;
+  font-weight: 500;
+  font-display: optional;
+  src: url('../fonts/Inter-Medium.woff2') format('woff2');
+}
+
```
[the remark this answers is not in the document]
Vinnl: Oh, are there UX review meetings? If so, could you invite me to those? I haven't really interacted with UX yet, so it would be great to be able to ask questions about things like this.

I removed both this font and the ones for the breaches empty state in af1df92b. Note that this does mean that the user's email address is no longer called out (i.e. bolded), and it's looking quite a bit heavier:

image

src/views/main.js
Vinnl: It's not a particularly heavy load, so I think the cost of the additional HTTP request, in combination with there not being that many pages (and thus navigations) in the first place, might outweigh it? Either way I expect the difference to be marginal and thus not worth measuring?

src/views/main.js
Vinnl: Looks like in general the measure is to protect users of older browsers, since [`target="_blank"` implies `rel="noopener"` on modern browsers](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a#browser_compatibility). I expect those old browsers to have issues in general and this indeed not being needed, so that's good to know. But then, it can't really hurt either :)

src/client/css/nav.css
Vinnl: Good idea, fixed in 513347c5.

Vinnl:
I fixed @flozia's link indicator issue in 3f7d9154, though I'm not quite sure why that code is needed in the first place - see the commit message. @toufali, you might still want to submit a followup PR if you feel like more is needed there.

flodolo (approved):

src/views/main.js
Vinnl: Oh! I also forgot to mention I went with inline in the first place: that allows me to recolour the icon to be purple when the page is currently active (see references to `currentcolor` in the SVG).

toufali: Agreed it's not a concern now and we can merge this as is.  

Our nav items _will_ increase though (expected soon), and with it the html filesize loaded for every page. Since we're SSR it would be nice to keep our html as lean as possible.

Recoloring could be accomplished by using a purple icon for active and `filter: grayscale(1) opacity(...)` for inactive.  

Or a maybe [a mask](https://css-tricks.com/change-color-of-svg-on-hover/#aa-use-a-mask-instead-of-a-background-image)?  Neither of these options are as simple as svg `currentColor`, but the trade-off might be worth the lighter HTML and decluttered code.

src/client/css/fonts.css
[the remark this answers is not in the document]
flozia: Yeah, I think we would at least need one other weight for Metropolis in order to use it typographically as intended or fallback use Inter in this instance. I agree that we should be cautions of resources being loaded if we only need them in single instances. Cutting back too far on design decisions is not necessarily the best solution in my opinion. Let’s consolidate with UX and make sure we understand implications and come to a decision together.

toufali (commented):
Thanks for the updates!  Further to my responsive comment last week, it looks like a quick fix might be to remove nav-item inline padding so that they align with the line-edges as in the design:

Screen Shot 2023-02-13 at 2 14 08 PM

I imagine there might still be more to it than that, but it's a start!

flozia (approved):
The changes are working great! Depending on your preferences we should address two things before landing this PR or as a follow-up:
- Consolidate with UX regarding the font styles we’d like to use
- Enhancing the responsive layout for smaller desktop viewports as [mentioned in this comment](https://github.com/mozilla/blurts-server/pull/2825#pullrequestreview-1293797849) by @toufali. This might be also a good candidate to address when we have all of the layouts in place.

src/views/main.js
toufali:
```suggestion
    <a target="_blank" href='https://support.mozilla.org/kb/firefox-monitor' class='nav-item'>
```

Vinnl: I think that CSP only prevents a full Referer header when linking to a non-HTTPS site, rather than protecting against [this security issue](https://developer.chrome.com/docs/lighthouse/best-practices/external-anchors-use-rel-noopener/) that `noopener` does. Additionally, older browsers still vulnerable to that might not even support respecting the Content Security Policy.

But if you prefer, since those older browsers probably won't work anyway and are pretty rare, I'll remove it :)

src/client/js/nav.js
```diff
@@ -17,8 +17,7 @@ function handleEvent (e) {
     case e.target.matches('.nav-toggle'):
       nav.toggleAttribute('hidden')
       break
-    case e.target.matches('.nav-item'):
-      e.target.style.setProperty('color', 'var(--purple-70)')
+    case e.target.matches('.nav-item') && e.target.getAttribute('target') !== '_blank':
```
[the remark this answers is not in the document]
Vinnl: Yep, nice one!

Vinnl:
I also removed the inadvertent left padding, but honestly, I think the breakpoint for when to switch to the overlay menu is just too low - there's no way it's all going to fit on one page and be readable at the current breakpoint.

I'll rebase this and resolve the conflicts, assuming that the font changes look as desired after https://github.com/mozilla/blurts-server/pull/2828, and then merge - the responsive issue is not a regression, so we can address that separately if desired.

src/client/css/fonts.css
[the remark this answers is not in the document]
toufali: Just a follow up on this particular case -- I reviewed it with UX and it was updated via https://github.com/mozilla/blurts-server/pull/2828

The solution agreed with UX was to use the paragraph text styles we already had in place (16px non-bold inter).  Apparently this should have already been in the designs, but content created this section instead of design, and so there were discrepencies. 

Currently the repo is set to use Metropolis for bolded headings only, and inter for paragraph with all styles allowed (bold/italic).  If we notice design going against this pattern, let's flag and discuss!

src/views/main.js
toufali: Thank you!

src/client/js/nav.js
toufali: Sounds good.  We remove the color and highlight in order to provide the user immediate feedback for the seconds before the new page load.

src/client/css/fonts.css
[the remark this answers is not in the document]
Vinnl: Ah OK, that's good to know, thanks for sharing!

PR #2828, opened 2023-02-12
use template and dynamic email for zero-state

toufali:
# References: 
Jira: MNTOR-1086
Figma: https://www.figma.com/file/gvyyqS9egsGjsH0LUDqT1m/Breach-Resolution-Flow-Design?node-id=654%3A16313&t=I1S7G3TnYG1IGh0z-4

# Description
Refactor Breaches page zero-states to use a single template with dynamic email updated with state.  

Per @Vinnl [comment](https://github.com/mozilla/blurts-server/pull/2821#discussion_r1101504633), creating a separate zero state for each condition under each email account is not ideal.  This refactor attempts to simplify it by using `selectedEmail` stored in state to inject into a single zero-state, created as needed.

# Screenshot (if applicable)

Not applicable.

# How to test

1. Open the breaches page for an email address with zero breaches - it should show the "No breaches found" message and hide the "Unresolved / Resolved" tabs.
2. Open the breaches page for an email address with more than zero unresolved breaches - it should just list those breaches.
3. Open the breaches page for an email address with more than zero resolved, and no unresolved breaches - it should show the "All breaches resolved" under the "Unresolved" tab, and list all breaches under the "Resolved" tab.
4. 1 and 3 above should show an "Add email address" button if the user has not reached their email address quota. Once they do, it should be hidden.

### DIFF
--- a/src/client/js/partials/breaches.js
+++ b/src/client/js/partials/breaches.js
@@ -5,14 +5,17 @@
 const breachesPartial = document.querySelector("[data-partial='breaches']")
 const state = new Proxy({
   selectedEmail: null,
-  selectedStatus: 'unresolved'
+  selectedStatus: 'unresolved',
+  resolvedCount: null,
+  unresolvedCount: null,
+  emailCount: null,
+  emailTotal: null
 }, {
   set (target, key, value) {
-    if (target[key] !== value) {
-      target[key] = value
-      render()
-    }
+    if (target[key] === value) return true
 
+    target[key] = value
+    if (key === 'selectedEmail' || key === 'selectedStatus') render()
     return true
   }
 })
@@ -27,6 +30,8 @@ function init () {
   resolvedCountOutput = statusFilter.querySelector("label[for='breaches-resolved'] output")
   unresolvedCountOutput = statusFilter.querySelector("label[for='breaches-unresolved'] output")
 
+  state.emailCount = Number(breachesPartial.querySelector('.email-stats').dataset.count)
+  state.emailTotal = Number(breachesPartial.querySelector('.email-stats').dataset.total)
   state.selectedEmail = emailSelect.value // triggers render
 
   emailSelect.addEventListener('change', handleEvent)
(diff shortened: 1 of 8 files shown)


src/client/js/partials/breaches.js
```diff
@@ -80,11 +85,10 @@ async function updateBreachStatus (input) {
 }
 
 function renderResolvedCounts () {
-  const nrResolved = breachesPartial.querySelectorAll(`[data-status='resolved'][data-email='${state.selectedEmail}']`).length
-  const nrUnresolved = breachesPartial.querySelectorAll(`[data-status='unresolved'][data-email='${state.selectedEmail}']`).length
-  resolvedCountOutput.textContent = nrResolved
-  unresolvedCountOutput.textContent = nrUnresolved
-  statusFilter.toggleAttribute('hidden', nrResolved === 0 && nrUnresolved === 0)
+  state.resolvedCount = breachesPartial.querySelectorAll(`[data-status='resolved'][data-email='${state.selectedEmail}']`).length
```
Vinnl, line 88: Nice, I always struggle with `nrOf` names, but `Count` is a good suffix to use :clap:

Vinnl (approved):
Works as intended for me. And I guess that means our solution for variables in strings that change on the client side is to wrap those variables in a tag that we can target with a selector.

src/views/partials/breaches.js
```diff
@@ -138,35 +118,27 @@ export const breaches = data => `
...
-        ${getMessage('breaches-none-headline')}
-      </h2>
-      <p>
-        ${
-          data.breachesData.verifiedEmails.map(account => {
-            return `<span data-email="${account.email}" ${account.primary ? '' : 'hidden'}>${getMessage('breaches-none-copy', { email: `<b>${account.email}</b>` })}</span>`
-          }).join('')
-        }
+      <h2>${getMessage('breaches-none-headline')}</h2>
+      <p>${getMessage('breaches-none-copy', { email: '<b class="current-email"></b>' })}</p>
```
flozia, line 125: Haven’t thought of doing it this way. In case we adopt this as a pattern in multiple places we should think of a higher order util that we can use for injecting string dynamically.

flozia (approved):
This is working well!

toufali:
@Vinnl as discussed I updated style per UX and used disable rather than hide for zero-state filter per PM.

Merging now, but please review for sanity-check and in case it causes conflict with your nav PR.

PR #2830, opened 2023-02-12
Mntor 1160/add new email

toufali:
# References: 
Jira: MNTOR-1160
Figma: https://www.figma.com/file/gvyyqS9egsGjsH0LUDqT1m/Breach-Resolution-Flow-Design?node-id=214%3A20759&t=qgRO530h7MeDnEBP-0

# Description
TODO: add description

# How to test
TODO


flodolo (approved):
Please fix the linter error ;-)

src/client/js/partials/add-email.js
```diff
@@ -0,0 +1,53 @@
...
+        'x-csrf-token': form.elements['csrf-token'].value
+      },
+      mode: 'same-origin',
+      method: 'POST',
+      body: JSON.stringify({
+        email: form.elements['email-address'].value
+      })
+    })
+
+    if (!res.ok) throw new Error('Bad fetch response')
```
flozia, line 29: Question (non-blocking): Should we move and consolidate new errors in `src/utils/error.js`?

src/views/partials/settings.js
```diff
@@ -151,12 +151,7 @@ export const settings = data => {
           <p>${getMessage('settings-email-limit-info', { limit })}</p>
 
           ${createEmailList(emails, breachCounts)}
-          <button
-            class='settings-add-email-button primary js-settings-add-email-opener'
-            ${emails.length >= AppConstants.MAX_NUM_ADDRESSES ? 'disabled' : ''}
-          >
-            ${getMessage('settings-add-email-button')}
-          </button>
+          <a class='button primary settings-add-email-button' href='dialog/add-email'>${getMessage('settings-add-email-button')}</a>
```
flozia, line 154: Suggestion (blocking): We need to re-add the `disabled` attribute when the user reached the email limit.

src/views/partials/breaches.js
```diff
@@ -17,7 +17,7 @@ function createEmailCTA (count) {
 
   if (count >= total) return '' // don't show CTA if additional emails are not available for monitor
 
-  return `<button class='anchor' data-dialog='add-email'>${getMessage('add-email-link')}</button>`
+  return `<a href='dialog/add-email'>${getMessage('add-email-link')}</a>`
```
flozia, line 20: Question (non-blocking): From the perspective of a user the behavior is more like a button. Should we add the attribute `role="button"`?

src/views/partials/add-email.js
```diff
@@ -2,16 +2,24 @@
...
+  <p>${getMessage('add-email-your-account-includes', { total: data.emailLimit })}</p>
+  <input type='hidden' name='csrf-token' value='${data.csrfToken}'>
+  <label class='text-field'>
+    <span>${getMessage('add-email-address-input-label')}</span>
+    <input type='email' name='email-address' required>
+  </label>
+  <button class='primary' type='submit' name='email-submit'>${getMessage('add-email-send-verification-button')}</button>
+</form>
+<template data-success>
+  <p class='add-email-success'>${getMessage('add-email-verify-the-link', { email: '<b class="current-email"></b>', 'settings-href': 'href="/user/settings"' })}</p>
```
flozia, line 23: Nit (non-blocking): This line is getting a bit long.

src/client/js/dialog.js
```diff
@@ -0,0 +1,75 @@
...
+  }
+}
+
+async function openDialog (path) {
+  const partialName = path.substring(path.lastIndexOf('/') + 1)
+
+  dialogEl.showModal() // provide immediate UI response by showing ::backdrop regardless of content load
+  dialogEl.setAttribute('data-partial', partialName) // allow selector access, e.g. dialog[data-partial='add-email']
+  dialogEl.addEventListener('click', handleEvent)
+  dialogEl.addEventListener('close', resetDialog)
```
flozia, line 45: Suggestion: I’d expect clicking the backdrop of the dialog to dismiss it a well.

src/client/js/dialog.js
```diff
@@ -0,0 +1,75 @@
...
+
+function resetDialog () {
+  dialogEl.removeEventListener('click', handleEvent)
+  dialogEl.removeEventListener('close', resetDialog)
+  dialogEl.removeAttribute('data-partial')
+  dialogEl.replaceChildren()
+}
+
+if (triggerLinks.length) init(triggerLinks) // adds event listeners for dialog links already in DOM
+observer.observe(main, { attributes: false, childList: true, subtree: true }) // watches for new dialog links dynamically added to DOM
```
flozia, line 75: Praise: Awesome that you’re already anticipating this!

flozia (commented):
The new dialog component is awesome and the add new email flow is already working really well. I’ve left a few comments inline — other than these it would be good if we could disable page scrolling when the dialog is open.

src/client/js/partials/add-email.js
toufali: Yes, I might need a ticket and a tech discussion on this to align – I out of the loop on the patterns we've established!

src/views/partials/settings.js
toufali: Great catch – I totally missed that added functionality.  Fixing this was funny – `disabled` is not allowed on `<a>`, so I ended up needing to update the dialog handler to allow `<button>` as well.  A worthy addition.

Should be fixed via https://github.com/mozilla/blurts-server/pull/2830/commits/9efba5d249d99479124743a60d6ea6001fa48e17

src/views/partials/breaches.js
toufali: Fair point.  I changed these to buttons in https://github.com/mozilla/blurts-server/pull/2830/commits/9efba5d249d99479124743a60d6ea6001fa48e17

src/views/partials/add-email.js
toufali: fixed via https://github.com/mozilla/blurts-server/pull/2830/commits/11ec0ba60e03dabff53b64d089d7f932778540bc

src/views/partials/settings.js
flozia: Thank you for doing that and also adding the “add email flow” to the settings page.

src/client/js/dialog.js
toufali: I've always been a click-backdrop-to-dismiss person too.  The research I did left me on the fence. Using the native dialog element makes it non-trivial to add (with `::backdrop`).  In fact, I think you were the one who pointed out a bug where clicking on the dialog parent incorrectly closed it!  

Rather than throw more code at the feature I just removed it.  I support adding click-backdrop-to-dismiss functionality under separate ticket. 👍

src/client/js/dialog.js
toufali: Thanks! This was a rushed midnight addition, but so far seems to be holding up 🤞

flozia (approved):
Thanks for addressing my comments!

There are two follow-ups that would be nice to address:
- It would be nice if dismissing the dialog could be dismissed on blur as discussed inline.
- After adding a new email to be monitored and then closing the dialog does not update the counter. Only after reloading the page the UI reflects the actual email count.

toufali:
> Thanks for addressing my comments!
> 
> There are two follow-ups that would be nice to address:
> 
>     * It would be nice if dismissing the dialog could be dismissed on blur as discussed inline.
> 
>     * After adding a new email to be monitored and then closing the dialog does not update the counter. Only after reloading the page the UI reflects the actual email count.

Yes, agreed with both follow-ups.  Will create tickets now – the second item I flagged in today's and is high-priority.

Vinnl (commented):
Since it's already reviewed and merged, I'm not going to peruse through the rest of the code anymore, but just submitting my in-progress comments, and these notes I made while trying it out locally:

- The input in modal the modal is not focused when opening it
- Submit errors are not handled, e.g. when the user enters an existing email address (this might be a separate ticket already?)
- The "add new email" button is still visible in the settings screen when the user has reached the limit
- Opening the dialog link in a new tab results in only seeing the dialog contents
- Compared to the Figma some content was missing in the verification email, but I'm not sure if you touched that in this PR
- borders around addresses in settings

locales/en/add-email.ftl
```diff
@@ -0,0 +1,21 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+### Dialog window that allows a user to add a new email address to be monitored
+
+add-email-add-another-heading = Add another email address
+
+# $total is the number of emails a user is allowed to add
```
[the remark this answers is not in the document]
flodolo: We haven't been able very strict around formatting but yes, that would be the preferred format for comments.

PR #2832, opened 2023-02-14
Add a 404 error page

Vinnl:
# References: 
Jira: MNTOR-1167
Figma: https://www.figma.com/file/gvyyqS9egsGjsH0LUDqT1m/Breach-Resolution-Flow-Design?node-id=654%3A16492&t=NYtMRtj552xcpQeO-4

# Description

(Also adding @mansaj as a reviewer, because I'm not sure if/how I should add logging to this, with the removal of the `NotFoundError`.)

Adds a 404 error page. Note that I kept the header; this allowed me to avoid another branch for the layout, and I imagine it's still useful when you're on the wrong page. ~~I'll also verify with UX whether that's OK.~~ [UX is OK with this.](https://mozilla-hub.atlassian.net/browse/MNTOR-1167?focusedCommentId=648427)

I also didn't have ideas on how to vertically center (classic...) the error message without pushing the header and footer out of the page or reworking the entire site layout mechanisms, so open to suggestions there.

# Screenshot (if applicable)

image

Screen Shot 2023-02-14 at 15 09 33

# How to test

Visit a page that does not exist.

- [ ] I've added a unit test to test for potential regressions of this bug. - Not sure where the best place for this is.
- [ ] All acceptance criteria are met. - This is only the 404 page, since feedback here probably applies to all error pages.


locales/en/app.ftl
```diff
@@ -838,3 +838,11 @@ menu-item-logout = Sign out
 mozilla = {-brand-mozilla}
 terms-and-privacy = Terms & Privacy
 github = {-brand-github}
+
+## Error page
+
+# Variables:
+#   $errorCode (number) - "404"
+error-page-error-404-title = { $errorCode } Page not found
```
flodolo, line 846: Page not found is only 404. What's the benefit of using a variable here?

flodolo (approved):

src/middleware/error.js
```diff
@@ -31,21 +31,11 @@ function errorHandler (err, req, res, next) {
...
-/**
- * Used as a 404 default for routes
- */
-function notFound (req) {
-  // TODO: when there's a 404 page to be rendered here,
-  // replace the json return with html
-  throw new NotFoundError(`Page not found! ${req.method} ${req.originalUrl}`)
-}
-
 function methodNotAllowed (req) {
```
mansaj, line 34: For logging, what if you kept the `notFound` function, and instead of throwing an error, just log the error in there and then renders the page?

mansaj (approved):

src/client/js/partials/notFound.js
```diff
@@ -0,0 +1,19 @@
+if (window.history.length > 1) {
```
flozia, line 1: The license header is missing.

```suggestion
/* This Source Code Form is subject to the terms of the Mozilla Public
 * License, v. 2.0. If a copy of the MPL was not distributed with this
 * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
 if (window.history.length > 1) {
```

src/client/css/partials/notFound.css
```diff
@@ -0,0 +1,24 @@
...
+  gap: var(--padding-sm);
+}
+
+.error-code {
+  font: var(--h3-font);
+  font-weight: bold;
+  font-size: 10rem;
+  color: var(--purple-60);
+  text-shadow: 0.75rem 0 0 var(--purple-20);
+}
```
flozia, line 24: “No newline at end of file”. :)

src/client/css/variables.css
```diff
@@ -29,6 +29,8 @@
   --blue-60: #0250bb;
   --blue-90: #09204d;
   --purple-5: #e7dfff;
+  --purple-20: #cb9eff;
+    --purple-60: #7542e5;
```
flozia, line 33: Suggestion: Identation

```suggestion
  --purple-60: #7542e5;
```

src/client/js/partials/notFound.js
```diff
@@ -0,0 +1,19 @@
...
+    // The call to `window.history.back()` might do nothing; for example, if the
+    // user has clicked Back to get to the 404 page: while there might not be
+    // another page to go back to, window.history.length will still be > 1
+    // because of the page the user can go forward to.
+    // Unfortunately there's no way to check whether it's possible to go back,
+    // or to check whether `.back()` succeeded, so we just use the fallback
+    // after half a second as passed:
+    setTimeout(() => {
+      document.location = '/'
+    }, 500)
```
flozia, line 15: Question (non-blocking): Would it be possible to pass the prev path from the controller and append it to `#back-cta` via a `data-*` attribute?

flozia (approved):
The page is looking great! Approving with no blocking remarks, but some nits that should be addressed before merging.

src/client/js/partials/notFound.js
Vinnl: Fixed in 2f5a84c9a417b67709f2dc007c429f1926b4aac3

BUT it's a waste of your and my time to manually be checking this in reviews, so I also submitted https://github.com/mozilla/blurts-server/pull/2836 as well.

src/client/css/variables.css
Vinnl: Oh sorry, forgot to call this out - I cannot get VSCode to _not_ add this indentation back. I expect it's something in our shared VSCode settings?

Vinnl: Actually, looks like it will let me now: af46b139

src/client/css/partials/notFound.css
Vinnl: I think this was the same story of VSCode automatically removing it, but it works now: af46b139

src/client/js/partials/notFound.js
Vinnl: Oof. Ehm, maybe, but then we'd be tracking the user's navigation state on the server-side, which I think is a bit out of scope for this PR...

src/middleware/error.js
Vinnl: Yeah that works for me. I'm not entirely sure about the format of our log messages though. What do you think about 500e23c5?

src/client/js/partials/notFound.js
flozia: Totally out of scope — just left this as a thought in case we need it in the future.

src/client/js/partials/notFound.js
flozia: Awesome — ty!

src/middleware/error.js
mansaj: Works for me

mansaj (approved):

src/views/main.js
```diff
@@ -47,7 +47,7 @@ const mainLayout = data => `
     <!-- End Google Tag Manager -->
     </head>
   <body>
-    ${data.partial.name === 'landing' ? landingHeader(data) : mainHeader(data)}
+    ${['landing', 'notFound'].includes(data.partial.name) ? landingHeader(data) : mainHeader(data)}
```
toufali, line 50: With the addition of this error page we should probably create a separate layout to handle these two partials and the others to come that don't require UI for logged-in user.  I created the following task to track this:
https://mozilla-hub.atlassian.net/browse/MNTOR-1254

PR #2834, opened 2023-02-14
make the GA env var optional

toufali:
# Description
Allows the app to build without requiring GA env var.


mansaj (approved):
Approving to let other features through for QA

PR #2836, opened 2023-02-15
Add ESLint plugin to enforce the licence header

Vinnl:
In response to https://github.com/mozilla/blurts-server/pull/2832#discussion_r1106318187.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -12,10 +12,19 @@
         "ecmaVersion": "latest",
         "sourceType": "module"
     },
+    "plugins": [
+        "header"
+    ],
     "rules": {
         "no-prototype-builtins": "warn",
         "no-undef": "warn",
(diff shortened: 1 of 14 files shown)


flozia (approved):
That’s great — thank you for adding this rule.

mansaj (approved):

toufali (commented):
Thanks for adding this!  

I've been annoyed by a weird indentation resulting from this header: the first line I add after this header includes a 2-space indent.  I don't imagine this PR would fix it, but I'll be on the lookout.  Presumably just closing this comment on a new line would fix  🤷

Vinnl:
Hmm, I haven't seen that myself yet, but if a newline will fix it, then this might indeed take care of it - the `2` in the config indicates that the comment should be followed by two newlines.

PR #2837, opened 2023-02-15
Incorporate UX feedback for navigation design update

Vinnl:
image


flozia (approved):
Looks as expected for me!

PR #2841, opened 2023-02-16
Always enable source maps

Vinnl:
We're an open source project, so no need to make it harder to read our source code. By generating source maps as separate files, they also don't add to the page weight.

Besides being more transparent, this also makes it easier for us debug issues that don't arise in our dev environments.

### DIFF
--- a/src/esbuild.js
+++ b/src/esbuild.js
@@ -31,7 +31,7 @@ esbuild.build({
   outdir: '../dist',
   format: 'esm',
   minify: AppConstants.NODE_ENV !== 'dev',
-  sourcemap: AppConstants.NODE_ENV !== 'dev',
+  sourcemap: 'linked',
   splitting: true,
   treeShaking: true,
   platform: 'neutral'


src/esbuild.js
```diff
@@ -31,7 +31,7 @@ esbuild.build({
   outdir: '../dist',
   format: 'esm',
   minify: AppConstants.NODE_ENV !== 'dev',
-  sourcemap: AppConstants.NODE_ENV !== 'dev',
+  sourcemap: 'linked',
```
toufali, line 34: The [docs are confusing here](https://esbuild.github.io/api/#sourcemap), but the code example suggests this value should be `true`, as originally coded.  I imagine `'linked'` might work as well, though I haven't tested.

toufali (commented):
This should already be enabled for all environments except our local env:
`sourcemap: AppConstants.NODE_ENV !== 'dev'` implies "true" for heroku, stage, and production.  

For 'dev' (our local env) we don't need sourcemaps because we import modules directly (unbundled) via `npm run dev`, and the modules are not minified.  `npm run dev` doesn't trigger ESBuild anyway, so this setting doesn't run.  It might only have an effect if we were testing/developing using `npm start`, but even in this case the files are not minified, so I'm not sure what benefit we'd achieve at the cost of increasing build time?

Vinnl:
Oh haha, that's what I get for making assumptions rather than actually reading the code :sweat_smile: Closing this PR then.

PR #2842, opened 2023-02-16
Don't call methods on objects that are `null`

Vinnl:
Well well well, if it isn't my code without TypeScript's guard rails :sweat_smile: 

The JS for the 404 page assumes the DOM elements of the 404 page, which will obviously fail if it's not on the 404 page.

### DIFF
--- a/src/client/js/partials/notFound.js
+++ b/src/client/js/partials/notFound.js
@@ -4,7 +4,7 @@
 
 if (window.history.length > 1) {
   const backCta = document.getElementById('back-cta')
-  backCta.addEventListener('click', () => {
+  backCta?.addEventListener('click', () => {
     window.history.back()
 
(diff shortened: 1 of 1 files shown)


flozia (approved):

toufali (commented):
This should be partially safeguarded in the future with the [code-splitting task](https://mozilla-hub.atlassian.net/browse/MNTOR-1171) that should only load client js related to the the active partial.

In the meantime, a pattern I've been using is to use the `data-partial` attribute present on the `<main>` element to initialize code only if necessary.  For example:

```js
const notFoundPartial = document.querySelector("main[data-partial='notFound']")

function init(){
  // do stuff without worrying if we're in the right partial
}

if (notFoundPartial) init()
```

PR #2848, opened 2023-02-17
Mntor 1188/breaches email count

toufali:
# References: 
Jira: MNTOR-1188

# Description
With this PR:
- the "monitored email" count in the header panel should only display verified email count
- the “Add email address” link in the header panel should open the “Settings” page instead of the “Add email” dialog
- the “Add email address” buttons in the Breaches table “zero-state” should continue to open the “Add email” dialog
- the “Add email address” buttons in the Breaches table “zero-state” should dynamically disappear after the user has reached their limit

TODO: Waiting for UX decision on whether the header link should read "manage emails" instead of "add email address".

# Screenshot (if applicable)

Screen Shot 2023-02-16 at 9 15 54 PM

# How to test
Visit the breaches page and test the acceptance criteria above.

### DIFF
--- a/src/controllers/breaches.js
+++ b/src/controllers/breaches.js
@@ -10,14 +10,16 @@ import { generateToken } from '../utils/csrf.js'
 import { getAllEmailsAndBreaches } from '../utils/breaches.js'
 
 async function breachesPage (req, res) {
-  const emailCount = 1 + (req.user.email_addresses?.length || 0) // +1 because user.email_addresses does not include primary
   // TODO: remove: to test out getBreaches call with JSON returns
   const breachesData = await getAllEmailsAndBreaches(req.user, req.app.locals.breaches)
+  const emailVerifiedCount = breachesData.verifiedEmails?.length ?? 0
+  const emailTotalCount = emailVerifiedCount + (breachesData.unverifiedEmails?.length ?? 0)
   appendBreachResolutionChecklist(breachesData)
 
   const data = {
     breachesData,
-    emailCount,
+    emailVerifiedCount,
+    emailTotalCount,
     partial: breaches,
     csrfToken: generateToken(res),
     fxaProfile: req.user.fxa_profile_json,
--- a/locales/en/breaches.ftl
+++ b/locales/en/breaches.ftl
@@ -15,7 +15,8 @@ emails-monitored =
      *[other] { $count } of { $total } emails monitored
   }
 
-add-email-link = Add email address
+# link to Settings page where user can add/remove emails and set message preferences
+manage-emails-link = Manage emails
 
 ## Breaches resolved filter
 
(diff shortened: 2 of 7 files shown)


toufali:
tagging @rhelmer for last person who worked on settings
tagging @mansaj for insight on error handling audit

Vinnl (approved):
I think related to your suggestion to use "Manage emails", but the link to settings wasn't visible at all when I had reached my limit, so I had to know where to go to remove an email or resend a verification email:

image

I was also momentarily confused by the "4 out of 5" - why weren't all five addresses monitored? Having a link to click through and then see that one was unverified will hopefully help there.

But the code does what it should do, and nothing in particular jumps out to me, so nice job :+1:

src/controllers/breaches.js
```diff
@@ -10,14 +10,16 @@ import { generateToken } from '../utils/csrf.js'
 import { getAllEmailsAndBreaches } from '../utils/breaches.js'
 
 async function breachesPage (req, res) {
-  const emailCount = 1 + (req.user.email_addresses?.length || 0) // +1 because user.email_addresses does not include primary
   // TODO: remove: to test out getBreaches call with JSON returns
   const breachesData = await getAllEmailsAndBreaches(req.user, req.app.locals.breaches)
+  const emailVerifiedCount = breachesData.verifiedEmails?.length ?? 0
```
flozia, line 15: Now we decided to not include unverified emails towards the limit we’d also need to pass the `emailVerifiedCount` to the settings page layout. Currently, we use the [total count for disabling the “Add email address” button](https://github.com/mozilla/blurts-server/blob/MNTOR-1188/breaches-email-count/src/views/partials/settings.js#L130) in the list of monitored emails.

toufali:
> ...but the link to settings wasn't visible at all when I had reached my limit

Agreed @Vinnl – I think we need more clarity from UX/PM.  I will add you to the Slack thread where I have been groaning on about this.  Specific to your point, I think we should change the copy to "Manage emails" and keep it on screen at all times.  (We currently hide "Add email" as you noted, because a user has already reached their limit.)

src/controllers/breaches.js
toufali: I might have lost a thread (or just confused) @flozia – the CTAs for this page behave as though a user has a hard limit regardless of verified/unverified.  For example, if a user has 3 verified emails and 2 unverified emails, they would not be allowed to add another email, and the add-email CTAs are hidden.  So at least in regards to this code, unverified emails _do_ count towards the limit, and I believe this is in line with Settings page?

The difference is with the "3 of 5 emails monitored" copy – this only includes verified emails.  And the CTA below it seems wrongly worded – I'll add you to the UX/PM conversation as well.

flozia (approved):

flodolo (approved):

PR #2838, opened 2023-02-15
Fixes email breaches count on settings page

flozia:
# References: 
Jira: [MNTOR-1173](https://mozilla-hub.atlassian.net/browse/MNTOR-1173)

# Description

Makes sure that the count of breaches shown on the settings page for a monitored email aligns with the count that is shown in the breaches table. Before this PR the list of breaches was not filtered for breaches that are not valid.

# Screenshot

pr-1173-screenshot

# How to test

1. Have an email to be monitored added that has associated breaches.
2. Navigate to the settings page.
3. Check the number of breaches mentioned in the subheader under the email card.

# Checklist (Definition of Done)
- [x] Localization strings (if needed) have been added.
- [x] Commits in this PR are minimal and [have descriptive commit messages](https://chris.beams.io/posts/git-commit/).
- [x] I've added or updated the relevant sections in readme and/or code comments
- [x] I've added a unit test to test for potential regressions of this bug.
- [x] Product Owner accepted the User Story (demo of functionality completed) or waived the privilege.
- [x] All acceptance criteria are met.
- [x] Jira ticket has been updated (if needed) to match changes made during the development process.
- [x] Jira ticket has been updated (if needed) with suggestions for QA when this PR is deployed to stage.

PR #2849, opened 2023-02-17
Add eslint-plugin-jsdoc

flozia:
# References: 

Putting this up as a draft in response to https://github.com/mozilla/blurts-server/pull/2838#discussion_r1108802218 for discussion.

# Description

This enables all the “recommended” rules by `eslint-plugin-jsdoc`. Are we all in favor of adding these rules, just a subset — or not at all?

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -6,14 +6,16 @@
         "jest": true
     },
     "extends": [
-        "standard"
+        "standard",
+        "plugin:jsdoc/recommended"
     ],
     "parserOptions": {
         "ecmaVersion": "latest",
         "sourceType": "module"
     },
     "plugins": [
-        "header"
+        "header",
+        "jsdoc"
     ],
     "rules": {
         "no-prototype-builtins": "warn",
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -27,6 +27,8 @@
             "block",
             " This Source Code Form is subject to the terms of the Mozilla Public\n * License, v. 2.0. If a copy of the MPL was not distributed with this\n * file, You can obtain one at http://mozilla.org/MPL/2.0/. ",
             2
-        ]
+        ],
+        "jsdoc/require-jsdoc": "off",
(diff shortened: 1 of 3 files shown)


Vinnl:
Looks like you forgot to commit the updated the lockfile. But yes, I am generally in favour of avoiding tedious discussions about individual linting rules, and just adopting recommended sets with general principles. (So for example, on Relay we enable rules that help with catching potential bugs, but leave styling issues to Prettier so we don't have to manually fix those.)

toufali (approved):
Not familiar with this plugin, but in favor of it, assuming it kicks in when we've opted in to a JSDoc block (and doesn't mandate we add one for every function 😅). 

I would appreciate an error if I've specifically used JSDoc syntax that is broken, and simple warnings for suggested-but-not-required practice.  (Maybe these are already "sensible" defaults?)

rhelmer (approved):

flozia:
> […] , but leave styling issues to Prettier so we don't have to manually fix those.)

@Vinnl Seems like styling-related lining is not included in the [set of recommend rules](https://github.com/gajus/eslint-plugin-jsdoc#configuration).

> Not familiar with this plugin, but in favor of it, assuming it kicks in when we've opted into a JSDoc block (and doesn't mandate we add one for every function 😅).
>
> I would appreciate an error if I've specifically used JSDoc syntax that is broken, and simple warnings for suggested-but-not-required practice. (Maybe these are already "sensible" defaults?)

@toufali Good point. I disabled the rule `require-jsdoc` so that the plugin only kicks in if we opt for a comment. I did not find a “sensible defaults” option and it looks like we’d have to set more restrictive rules ourselves. Let’s do that for individual rules as we encounter them.

Vinnl:
> Seems like styling-related lining is not included in the [set of recommend rules](https://github.com/gajus/eslint-plugin-jsdoc#configuration).

I didn't mean that as a specific suggestion, just an example of what "general principles" may look like. So this is perfect :)

PR #2851, opened 2023-02-17
Add documentation for emails

flozia:
# References: 
Jira: [MNTOR-1256](https://mozilla-hub.atlassian.net/browse/MNTOR-1256)

# Description

Documentation for the emails we are migrating for the refresh as well as the previewer.

## TODO

- [x] Email types
- [x] Previewer
- [x] Sending emails


docs/monitor-emails.md
```diff
@@ -0,0 +1,56 @@
+# Monitor emails
+
+## Email types
```
Vinnl, line 3: A short note about where to find these templates in the repository could be helpful, for developers?

Vinnl (approved):
Looking good, thanks for writing this up!

docs/monitor-emails.md
flozia: That’s a good point. One reason why I did not add a note that they did not have a dedicated place to point to. I think they should move into their own directory and not live between the other partials: PR #2858.

PR #2862, opened 2023-02-22
Mntor 1073/details pie chart

toufali:
# References: 
Jira: MNTOR-1073
Figma: https://www.figma.com/file/gvyyqS9egsGjsH0LUDqT1m/Breach-Resolution-Flow-Design?node-id=214%3A9127&t=f1s0mVIutb3dKOa7-0

# Description
Add pie chart data visualization to the Breaches page

# Screenshot (if applicable)

Screen Shot 2023-02-22 at 1 26 27 PM

# How to test
Visit the breaches page and toggle between email accounts, confirming the pie chart slices are sensible and the data classes/labels are in order of most-to-least frequent.


src/client/js/partials/breaches.js
```diff
@@ -3,6 +3,7 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
 const breachesPartial = document.querySelector("[data-partial='breaches']")
+const chartColors = ['#321C64', '#AB71FF', '#952BB9', '#D74CF0', '#9e9e9e']
```
Vinnl, line 6: `#AB71FF` and `#D74CF0` don't provide enough context at this font size, making them hard to read for people with e.g. lots of backlight or low-quality displays. Perhaps we could ask Design for darker colours?

image

image

src/client/js/partials/breaches.js
```diff
@@ -141,12 +143,55 @@ function renderZeroState () {
...
+        color: chartColors[3]
+      }
+      // falls through
+    default:
+      classesTop3.forEach((name, i) => {
+        chartData[i] = {
+          key: name,
+          name,
+          count: classesMap.get(name),
+          color: chartColors[i]
```
Vinnl, line 180: I'm guessing the order will generally stay the same, so in practice this might not be relevant, but I was wondering if it might not be helpful if the same type of data would always get the same colour? Otherwise, every time the user switches between different email addresses, they'll have to consult the legend again.

Vinnl (approved):
Obligatory [save the pies for dessert](https://www.perceptualedge.com/articles/visual_business_intelligence/save_the_pies_for_dessert.pdf) :P

But yeah, works well, does what it should do, nice job!

flozia (approved):

flodolo (commented):
I have an overall problem with this, and that's the use of data class names, together with "Other" and "None".

English used uppercase in `data-classes.ftl`, but these are names displayed in a list, and that would be incorrect in many languages (with the exception of German, as far as I know). In fact, both Italian and French use lowercase.

English: 
Screenshot 2023-02-23 alle 14 01 58

Italian: 
Screenshot 2023-02-23 alle 14 04 22

In this case, the data class is stand-alone, so it should be uppercase. It's not ideal, but could we apply `toLocaleUpperCase` to the name before displaying it in the legend?

src/client/js/partials/breaches.js
toufali: Was thinking the exact thing during dev – I'll flag with design 👍

toufali:
> I have an overall problem with this, and that's the use of data class names, together with "Other" and "None".
> 
> In this case, the data class is stand-alone, so it should be uppercase. It's not ideal, but could we apply `toLocaleUpperCase` to the name before displaying it in the legend?

Thanks for your review @flodolo.  Rereading it, I think I understand that the change you're requesting is to use lowercase "other" and "none" and to add `toLocaleUpperCase` as needed – no changes to the `data-classes.ftl` required?

I think it's acceptable to use lowercase "other" and "none" for English too – they're not titles – therefore avoiding `toLocaleUpperCase`.  If this works for you I will get confirmation from Content team.

flodolo:
> Thanks for your review @flodolo. Rereading it, I think I understand that the change you're requesting is to use lowercase "other" and "none" and to add `toLocaleUpperCase` as needed – no changes to the `data-classes.ftl` required?
> 
> I think it's acceptable to use lowercase "other" and "none" for English too – they're not titles – therefore avoiding `toLocaleUpperCase`. If this works for you I will get confirmation from Content team.

No, all data classes for English start with uppercase, so `None` and `Other` should stay uppercase for consistency.
https://github.com/mozilla/blurts-server/blob/main/locales/en/data-classes.ftl

What I'm saying is to apply `toLocaleUpperCase()` on names when used in the legend, so that locales like French and Italian will look a bit better (it will be a no-op for English).

flodolo:
>     * `toLocaleUpperCase()` converts all characters to uppercase, I assume you meant `toLocaleLowerCase()`? (JS does not have a native "TitleCase" method)

Sorry, I meant to apply `toLocaleUpperCase()` to the first character.

```
var capitalize = function(s) { return s.charAt(0).toUpperCase() + s.slice(1);}

capitalize("Credit cards") 
"Carte di credito"

capitalize("carte di credito") 
"Credit cards"

capitalize("账户余额")
"账户余额" 

capitalize("קבוצות גיל")

"קבוצות גיל"
```

The only locale changing would be Italian in the examples.

Based on conversations for a similar issue in Firefox, it's not ideal, but it should be safe (and safer than using CSS)

> What if we move "Other" and "None" to [data-classes.ftl](https://github.com/mozilla/blurts-server/blob/main/locales/en/data-classes.ftl)? Would this make it more clear that the words should be lowercase for consistency with the rest of the file (for Italian, French, etc) ?

It would help making sure that they're consistent, but:
* "None" and "Other" are not used in lists, only in the graph (as far as I understand).
* In the graph legend, it would be better to have them with the first letter capitalized, because they're not list items, they're stand-alone.

toufali:
@flodolo thanks for explaining in detail – previously I thought you _wanted_ Italian and French to be lowercase, but your last message clarifies that your concern is of languages that have no concept of an uppercase character.  Added via https://github.com/mozilla/blurts-server/pull/2862/commits/3362d26954ce73ff52892912e2a02a76cdf02aba

Regarding your note:
> "None" and "Other" are not used in lists, only in the graph (as far as I understand)
The graph is expected to be reused in a forthcoming dashboard, including "other" and "none".  Knowing this, would it now make more sense to include the two in [data-classes.ftl](https://github.com/mozilla/blurts-server/blob/main/locales/en/data-classes.ftl)?  Otherwise, we'll need to duplicate those strings into to a new "dashboard.ftl".

flodolo:
> @flodolo thanks for explaining in detail – previously I thought you _wanted_ Italian and French to be lowercase, but your last message clarifies that your concern is of languages that have no concept of an uppercase character. Added via [3362d26](https://github.com/mozilla/blurts-server/commit/3362d26954ce73ff52892912e2a02a76cdf02aba)
> 
> Regarding your note:
> 
> > "None" and "Other" are not used in lists, only in the graph (as far as I understand)
> 
> The graph is expected to be reused in a forthcoming dashboard, including "other" and "none". Knowing this, would it now make more sense to include the two in [data-classes.ftl](https://github.com/mozilla/blurts-server/blob/main/locales/en/data-classes.ftl)? Otherwise, we'll need to duplicate those strings into to a new "dashboard.ftl".

In that case, yes, it would make sense. And we should apply the same logic (uppercase first character) when using them in the graph legend (stand-alone) vs list items.

flodolo (approved):

src/client/js/partials/breaches.js
toufali: Fixed via https://github.com/mozilla/blurts-server/pull/2862/commits/fb71affa7c554763a84e804f597cd550e387ed9c - now using dark gray for text.

PR #2865, opened 2023-02-23
Download and show favicons for breached companies

Vinnl:
# References: 
Jira: MNTOR-1089
Figma: https://www.figma.com/file/gvyyqS9egsGjsH0LUDqT1m/Breach-Resolution-Flow-Design?node-id=718%3A17861&t=dJpMD2tw90Y3CHVS-4

# Description

This downloads favicons for breached companies when we load the breaches, and renders them in the breach resolution UI, if available.

Open questions:

- The fallback icon can look pretty busy - I'll call that out to Design.
- I'm not sure if it makes sense to link back to the breach domain, since that domain may no longer work - for example, PHPfreaks.com no longer resolves.
- I wanted to download the favicons and cache them, to avoid sending DDG heavy load, and to avoid leaking data to third parties. Not sure if the current place is the best place to do it though - happy to hear suggestions, and/or other people to tag for their input.

# Screenshot (if applicable)

image

With fallback icons:

image

# How to test

Open the breach resolution page with an email that has breaches.


src/views/partials/breaches.js
```diff
@@ -31,11 +31,17 @@ function createBreachRows (data) {
...
       return `
       <details class='breach-row' data-status=${status} data-email=${account.email} ${isHidden ? 'hidden' : ''}>
         <summary>
-          <span>${breach.Title}</span>
+          <span>
+            <a href='https://${breach.Domain}' target='_blank' class='breach-company'>
+              <img src='${logoPath}' alt='' class='breach-logo' height='32' />
+              ${breach.Title}
+            </a>
+          </span>
```
toufali, line 44: It looks like wrapping the breach title (company name) in `<a>` broke the ellipses.  As mentioned, I think we should remove the link altogether – hopefully that resolves the issue.

before:
Screen Shot 2023-02-23 at 3 06 52 PM
 
after:
Screen Shot 2023-02-23 at 2 58 44 PM

src/controllers/breaches.js
```diff
@@ -18,6 +22,7 @@ async function breachesPage (req, res) {
 
   const data = {
     breachesData,
+    breachLogos: await getBreachLogos(req.app.locals.breaches),
```
toufali, line 25: Would this potentially add delay to each render? 

Could we front-load this work into the "get all breaches" function in `/utils/hibp.js`, which occurs much less frequently?  Then, perhaps it can be derived as part of the `breachesData` object?  See other comment with more discussion.

cc @mansaj for having created `breachesData`

src/utils/hibp.js
```diff
@@ -147,8 +155,32 @@ async function loadBreachesIntoApp (app) {
...
+    get(`https://icons.duckduckgo.com/ip3/${breachDomain}.ico`, (response) => {
+      if (response.statusCode !== 200) {
+        return
+      }
+
+      const file = createWriteStream(resolve(logoFolder, breachDomain.toLowerCase() + '.ico'))
+      response.pipe(file)
+      file.on('finish', () => {
+        file.close()
+      })
```
toufali, line 176: Is it possible to add this path to the `breaches` object under the associated breach, to avoid having to call it dynamically for each page render in the controller?

Within `breaches` there exists `breach.LogoPath` which is I believe is unused (and potentially includes an incorrect path) – maybe we can repurpose it?

cc @mansaj for having worked on this last

src/utils/hibp.js
```diff
@@ -147,8 +155,32 @@ async function loadBreachesIntoApp (app) {
   log.info('done-loading-breaches', 'great success 👍')
 }
 
+async function downloadBreachIcons (breaches) {
+  const breachDomains = breaches.map(breach => breach.Domain)
+  const logoFolder = resolve(dirname(fileURLToPath(import.meta.url)), '../client/images/logo_cache/')
+  try {
+    await mkdir(logoFolder)
```
toufali, line 162: Are there any Kubernetes implications here, since we run multiple instances of the server?

toufali (commented):
This feature looks great; the logos add much needed flair!  

I appreciate the approach you took loading these server side – I left some questions/suggestions for improvement, assuming I understand what's going on...

- agreed that we should remove links.
- cc'd @mansaj who last worked on associated code and had thoughts on long-term solutions

src/views/partials/breaches.js
Vinnl: Ah yes, wasn't aware that we did that, but removing the links should resolve that. If we should keep the links, I'll make sure to transplant the ellipsis.

src/utils/hibp.js
Vinnl: Yeah, I was contemplating doing that, but with the `breaches` object coming directly from HIBP as I understand it, I didn't want to mess with that object. I'm not entirely opposed to that method though.

Another approach would be to still generate a breach-logo mapping, but do it at breach-loading time rather than in the controller, avoiding the delay there. I'll add that tomorrow.

src/utils/hibp.js
Vinnl: @mansaj? I don't know much about Kubernetes, though I'd expect the logo cache to just be recreated on every server?

src/utils/hibp.js
Vinnl: Well, it's past midnight here, so officially tomorrow: 80d8f1e7

src/controllers/breaches.js
Vinnl: Let's centralise this thread here: https://github.com/mozilla/blurts-server/pull/2865#discussion_r1116373574

src/utils/hibp.js
pdehaan: drive by 👀 , but I'm a bit confused as to when/how this gets called. In the current iteration, we merge some PRs and it gets pushed to a CDN.
But how frequently is the server polling for changes to the upstream HIBP feed and potentially fetch these logos via DDG?
Currently [HIBP](https://haveibeenpwned.com/api/v3/breaches) has 656 breaches. Is Monitor syncing w/ HIBP every 15 minutes? 60 minutes? The .env-dist seems to have a default of 10 minute polling, but not sure what you're using in prod.
https://github.com/mozilla/blurts-server/blob/56f08119dd546f880cefd18c4de8eebc47c34964/.env-dist#L48-L49

I didn't see if the code below was checking if the cached image is already stored locally, versus looping over `breachDomains.map(...)` and using `https.get()` and write streams to cache the .ico. So I guess wondering if we're fetching 650 .ico files every 10 minutes in production, that means roughly 3900 images per hour (per instance if we're caching locally and not on a shared CDN).
Or maybe add a quick check to see if we have a cached logo before refetching.

src/views/partials/breaches.js
```diff
@@ -31,11 +31,17 @@ function createBreachRows (data) {
...
         dataClasses: longList.format(dataClassesTranslated)
       })
+      const logoPath = logos.get(breach.Domain) ?? '/images/fallback-logo.svg'
 
       return `
       <details class='breach-row' data-status=${status} data-email=${account.email} ${isHidden ? 'hidden' : ''}>
         <summary>
-          <span>${breach.Title}</span>
+          <span>
+            <a href='https://${breach.Domain}' target='_blank' class='breach-company'>
```
pdehaan, line 40: nit: maybe a `rel="noopener noreferrer nofollow"` on this link (if it needs to be linked) since some of these sites are sometimes a bit sketchy.
I've never been convinced the domain needs to be linked. I think last time I checked, it failed like 40% of the time if the site was no longer a site, or wasn't https:// compatible.

pdehaan: Also, does this need to do a check for `breach.Domain`? I think all records have them, but if I do a quick Ctrl+F on `Domain: ""` I get about 26 results (unless those results have already been filtered out). Although I do see a nullish check on `breach.Domain` above on L34, so&hellip; /suspicious_eyes

Vinnl: Well, now I'm getting conflicting input :sweat_smile: https://github.com/mozilla/blurts-server/pull/2825#discussion_r1105046801

But hopefully PM agrees that the link can go away and it won't be an issue. If not, then yes, this probably needs such a check, good callout.

src/utils/hibp.js
Vinnl: Good callout, here's a check to only download it if we don't have an icon cached locally: ba354c12

src/views/partials/breaches.js
toufali: Y'all gonna make me cite sources tonight eh? 😄 

Regarding `rel="noreferrer"`:
>It indicates that no referrer information is to be leaked when following the link _**and also implies the `noopener` keyword behavior under the same conditions**_.
https://html.spec.whatwg.org/multipage/links.html#link-type-noreferrer

>...and creates a top-level browsing context as if `noopener` were also set
https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/rel#attr-noreferrer

>Sick of typing `<a rel="noopener noreferrer">`?
`<a rel="noreferrer">` implies noopener.
Specifying both is redundant 🤪
https://twitter.com/_developit/status/1054422343609655296

Regarding [our CSP](https://github.com/mozilla/blurts-server/blob/a27170725ff114f5859a4bedc2efd72c4ed93b34/src/app.js#L119-L123) which is actually stricter than `no-referrer`! (`no-referrer` is our fallback):
>Using Referrer-Policy: no-referrer is effectively setting `rel=noreferrer` on every link and protects the most against information leakage.
https://kevinhakanson.com/2018-02-07-referrer-policy-http-header

Regarding `rel="nofollow"`:
This seems more SEO-related but probably worthy of inclusion, particularly for links like these!

pdehaan: Interesting, I didn't know this: https://mathiasbynens.github.io/rel-noopener/

> _"2021 update: Browsers now implicitly set `rel=noopener` for any `target=_blank` link, [following a spec change](https://github.com/whatwg/html/issues/4078). If the demo on this page no longer seems scary, congratulations — you’re using a modern browser!"_

TL:DR: Remove the link and lets never discuss this again. /disappears into fog

src/utils/hibp.js
```diff
@@ -157,7 +157,9 @@ async function loadBreachesIntoApp (app) {
 }
 
 async function downloadBreachIcons (breaches) {
-  const breachDomains = breaches.map(breach => breach.Domain)
+  const breachDomains = breaches
+    .map(breach => breach.Domain)
```
pdehaan, line 161: free tip: you could almost add `.toLowerCase()` here and avoid using it 3-4 places below on L171-186.

src/views/partials/breaches.js
Vinnl: Links removed in c56c084b.

src/utils/hibp.js
Vinnl: Yeah I suppose so. I think I leaned this way so as not to manipulate input data (i.e. received from HIBP's API), but only data derived from it (i.e. the icon's filename). But of course, with an intermediate value like this, the line between input and output gets blurry :sweat_smile:

Vinnl:
I discussed with UI that we could show a unique letter-based "icon" instead of the fallback icon, which I implemented in e636c53f and looks like this:

image

Holding off for a review by @mansaj, who implied that there would probably be problems with the approach taken in this PR (and hopefully can suggest a way to solve that :smile: ).

src/utils/hibp.js
mansaj: @Vinnl so I'm not sure if this directory would work when our service is running in kube. In our build steps we transform the `client` folder into `dist` folder. When this function gets called (on server startup), it will download the logos into the `client` folder, which isn't going to be used anymore.

src/utils/hibp.js
```diff
@@ -147,8 +156,48 @@ async function loadBreachesIntoApp (app) {
   log.info('done-loading-breaches', 'great success 👍')
 }
 
+async function downloadBreachIcons (breaches) {
+  const breachDomains = breaches
+    .map(breach => breach.Domain)
+    .filter(breachDomain => breachDomain.length > 0)
+  const logoFolder = pathResolve(dirname(fileURLToPath(import.meta.url)), '../client/images/logo_cache/')
```
mansaj, line 163: Use `../../dist` instead of `../client` here. See Slack thread

src/views/partials/breaches.js
```diff
@@ -32,10 +32,17 @@ function createBreachRows (data) {
         dataClasses: longList.format(dataClassesTranslated)
       })
 
+      const logo = logos.has(breach.Domain)
+        ? `<img src='${logos.get(breach.Domain)}' alt='' class='breach-logo' height='32' />`
+        : `<span role="img" aria-hidden='true' class='breach-logo' style='background-color: var(${getColorForName(breach.Domain)});'>${breach.Domain.substring(0, 1)}</span>`
```
pdehaan, line 37: Not sure if I'm tracking all the code properly here, but does `breach.Domain` here account for potentially empty breach domains? Currently 26 breaches have an empty domain, per:

```sh
curl https://monitor.firefox.com/hibp/breaches | jq '.[] | [{name:.Name, domain:.Domain}]' | grep -B1 '"domain": ""'
```

it almost looks like we check if the logos Map has the current `breach.Domain`. Presumably if the logo is NOT in that map, there was an empty `Domain` or DDG just didn't have a hit. If there WAS a hit, we return the found logo. If not, we might be trying to calcuate a color via `getColorForName("")` on an empty string and trying to do `breach.Domain.substring(0, 1)` on an empty string (which would return an empty string). Not sure if we need to maybe fall back to `breach.Name` or something if Domain is an empty string.

src/utils/hibp.js
Vinnl: OK, I've added a `LIVE_RELOAD` environment variable that is set to `'true'` when we're running with `npm run dev`, and then determine which directory to use based on that: 6e6e2ec9

src/views/partials/breaches.js
Vinnl: Oh, thanks for sharing the link to the data structure! And yes, good point, apparently the domains are empty strings if they're not set... Looking through the `breach.Name`s in the data set, those do appear to be properly set for all breaches, so I just used that to determine the colour now - the actual colour doesn't matter that much, as long as it's deterministic for every breach. de467e27

rhelmer (commented):
Not sure if it's related to this change, but I'm using the `redis-mock` and getting this on server startup now:
```
CRITICAL fx-monitor.uncaughtException: TypeError: Class constructor RedisStore cannot be invoked without 'new'
    at getRedisStore (file:///Users/rhelmer/src/blurts-server/src/app.js:57:33)
    at file:///Users/rhelmer/src/blurts-server/src/app.js:156:18
```

rhelmer:
> Not sure if it's related to this change, but I'm using the `redis-mock` and getting this on server startup now:
> 
> ```
> CRITICAL fx-monitor.uncaughtException: TypeError: Class constructor RedisStore cannot be invoked without 'new'
>     at getRedisStore (file:///Users/rhelmer/src/blurts-server/src/app.js:57:33)
>     at file:///Users/rhelmer/src/blurts-server/src/app.js:156:18
> ```

I rebased against `main` and this went away

Vinnl:
@rhelmer Great. I'll rebase when this is approved - don't want to mess up people's local checkouts of this branch.

rhelmer (approved):
lgtm in local dev env. Could you file a followup ticket in the backlog to revisit our caching strategy? As we discussed, we have a CDN that we use for some things and not others, it might be nice to unify the strategy here.

PR #2867, opened 2023-02-23
Remember the user's last-selected email

Vinnl:
# References: 
Jira: MNTOR-1155
Figma: N/A

# Description

We don't want to leave traces of the user's known email addresses after their session expires, so we only store the index of the email. This does mean that if two different users share the same device, one user's selected email might affect the preselected one for the other. That does not seem like a major deal.

# Screenshot (if applicable)

Not applicable.

# How to test

Select a different email address to view the breaches of. Refresh the page. That email address should still be selected.


src/client/js/partials/breaches.js
```diff
@@ -45,6 +45,7 @@ function handleEvent (e) {
     case e.target.matches('custom-select[name="email-account"]'):
       state.selectedEmail = e.target.value
       breachesTable.querySelectorAll('span[data-email]').forEach(message => message.toggleAttribute('hidden', message.dataset.email !== e.target.value))
+      document.cookie = `monitor.selected-email-index=${Array.from(e.target.shadowRoot.querySelectorAll('option')).findIndex(optionEl => optionEl.value === e.target.value)}`
```
Vinnl, line 48: I'm not a big fan of using the DOM as the source of truth for data. Another option would be to add the list of email addresses to the state object - let me know if you prefer that.

Vinnl: Also, this line is looong. In absence of an automated code formatter, happy to take suggestions; I would've extracted the index into a separate value, but ESLint wouldn't let me do that in `case` statements.

toufali: Hope you accept contributions!  I added code via https://github.com/mozilla/blurts-server/pull/2867/commits/fc71963f31c97e9b3d136dde638587f32464c770 to expose `selectedIndex` in `<custom-select>`.  (This property is native to a standard `<select>`.)  

This significantly shortens your line 😉

toufali (approved):
Looks and works well!  I added some code per your comment.

The only flag I found is that when logging in under a different account, the index still persists and displays an email default that I did not set.  Might need to brainstorm a fix for that... clear cookie prop on sign-out?

mansaj:
In the edge case where a user has multiple FxA accounts, it might be important to separate out the settings for the different accounts. In order to do so, we need an identifier for each account.

Some thoughts re not storing plain text email for privacy reasons:
From @flozia, we can use last x digits of fxa user id as the identifier
From @mansaj, we can use hash of the email as an identifier

@Vinnl

src/client/js/partials/breaches.js
Vinnl: That works, thanks!

PR #2869, opened 2023-02-25
chore: ensure all required dependencies for v2 are in src npm workspace

rhelmer:
I noticed that we don't have all required dependencies in the `./src` workspace - I did my best to separate the runtime and dev dependencies.

I tested this by running unit tests and the server itself. It's possible that something is missing if it's imported dynamically, but I don't see any place the current code is doing that outside of tests.


toufali (approved):
I hope you don't mind my contribution – it was faster than commenting all the things 😅 

I think this is a great first step, thanks for getting it started.  I tested by deleting node_modules and package-lock.json, and deleting all deps and dev-deps from the root package.json, then running unit tests and server.

Only flag I noticed was `git-rev-sync` in legacy code but not in new src.  I think it's being used in legacy to create a docker package.  This is probably still ongoing and we'll need to port this functionality over to src.  There might be other cases like this that we'll need to address, but for now, I think this PR pushes us in the right direction.

pdehaan:
> _"I think it's being used in legacy to create a docker package. This is probably still ongoing and we'll need to port this functionality over to src."_

https://github.com/mozilla/blurts-server/blob/3032742847c38d3f63f460940d5f58c0bc2717a9/controllers/dockerflow.js#L19
Generally we used it to populate the deployed SHA in the [`/__version__`](https://monitor.firefox.com/__version__) endpoint so QE/SRE would know what was deployed. But it looks like the staging server's endpoint (per the foxfooding draft document) `/__version__` is a 404'n. Maybe a bug, maybe not, maybe a conversation between you, your SRE, and your priest.

toufali:
Thanks @pdehaan for the context! I think it's not working on Stage because we have yet to port that code to the new codebase.  We already ported heartbeat etc because we learned the hard way that Stage breaks without it.

PR #2874, opened 2023-02-28
remove V1/V2 switch functionality

toufali:
# References: 
Jira: MNTOR-1272

# Description
A variable and scripts were created to easily allow switching between Monitor V1 and V2.  This PR removes that functionality and starts V2 every time.


toufali:
@mansaj @rhelmer merging this now to get it on stage asap.  Should be safe in that it either "works" or the site won't load 😆

PR #2888, opened 2023-03-08
Mntor 1254/update layouts

toufali:
# References: 
Jira: MNTOR-1254

# Description
The Landing, error, email preview, and all other pages that should be displayed using guest/anonymous layout – no nav, avatar, user menu, etc.  Currently they overload the "main" (signed-in user) layout using conditional and irrelevant code.  With a separate and specific layout we can separate concerns, de-clutter code, and help avoid confusion.

# Screenshot (if applicable)

Not applicable.

# How to test
- visit guest/anonymous pages to confirm they display correctly and do not display nav, user menu, etc
- visit logged-in-user pages to confirm they show nav, user menu, etc.


toufali:
@flozia would it make sense to move the email layout out of `/emails` and into the parent `/views` folder?  Does it behave similar to the other layouts?  Perhaps at a minimum we might rename the email layout module with the `Layout` suffix?

rhelmer (approved):
I was just going over this code for the /breaches pages, this makes sense to me. I don't know if we want to keep it but note that the current /breaches and /breach-details/* pages do detect if the user is logged in, but we may not keep that feature (especially if these pages are moving off of our domain). In this case, I think they'd still want to use this "guest" style layout and not show the sidebar etc., but I'm not sure. It's not worth worrying about yet I don't think.

mansaj:
> I was just going over this code for the /breaches pages, this makes sense to me. I don't know if we want to keep it but note that the current /breaches and /breach-details/* pages do detect if the user is logged in, but we may not keep that feature (especially if these pages are moving off of our domain). In this case, I think they'd still want to use this "guest" style layout and not show the sidebar etc., but I'm not sure. It's not worth worrying about yet I don't think.

Agreed. With v2, we should keep `/user` path auth'd, but we can have the same `/breaches-details` route as before (for SEO?) but keep that public?

rhelmer:
> @flozia would it make sense to move the email layout out of `/emails` and into the parent `/views` folder? Does it behave similar to the other layouts? Perhaps at a minimum we might rename the email layout module with the `Layout` suffix?

I think keeping the email templates separate from the website templates probably makes sense (unless we want to try to blend and share elements, which I'd argue we don't, since email clients tend to limit HTML and especially JS).

Maybe mirroring it like this would help?
```
src/views/
  web/
    partials/
  email/
    partials/
```

I'd suggest filing a separate ticket for that, unless you want to try to decide in this one.

toufali:
Thanks @mansaj and @rhelmer for your comments and reviews.  Makes sense RE keeping emails separate.  In that case, I'll merge this as is.

PR #2891, opened 2023-03-10
Localise "Resolve this breach" heading

Vinnl:
# References: 
Jira: MNTOR-1319
Figma: N/A

I don't have to do anything else to make a string available for localisation, correct?


flodolo:
This is fine, but next time, please wait for the review. There's a reason why the review is triggered automatically via CODEOWNERS ;-)

PR #2892, opened 2023-03-10

Vinnl:
Specifically this one: https://github.com/mozilla/blurts-server/pull/2891

(Do we want to batch new l10n string updates? Because I expect MNTOR-1322 to result in a couple more strings as well.)


toufali:
@Vinnl yes. Often I batch but more out of concern that "locked" copy might change - coming up with new string keys is annoying 😅

@flodolo might comment on whether batching is more helpful for localizers

toufali:
Oh, one last note - it probably doesn't matter, but I had been using a naming convention for syncs like this in case we ever want to find or ignore similar PRs:

Maybe it doesn't matter if this tradition ends 🤷‍♂️

toufali (approved):
See note about PR title, in case we want to continue naming convention - approved either way

flodolo:
In general, I prefer stacking changes, but I don't know if we have anything
else planned in the short term.

flodolo:
> In general, I prefer stacking changes, but I don't know if we have anything else planned in the short term.

Speaking of the devil, #2897

Vinnl:
OK, got a bunch of strings in, merging now!

PR #2898, opened 2023-03-13
Actually mark a 404 page as a 404

### LINKED ISSUE #2895: Staging server 404 page returns status code 200
`<staging_url>/404-page-yo` returns an HTTP 200 which makes it difficult to check if a logo is missing.
Note that the same URL on production, https://monitor.firefox.com/404-page-yo, _does_ return a 404 status code.

Vinnl:
Fixes #2895.


flozia (approved):
Works as expected!

PR #2904, opened 2023-03-15
Prevent non-unsubscribe pages from breaking

Vinnl:
The query selector from the unsubscribe script failed on other pages (where the script gets loaded as well), leading to an attempt to call a method on `undefined`, halting all JS execution.


toufali (commented):
@Vinnl funny I recently merged an [almost identical PR](https://github.com/mozilla/blurts-server/pull/2908) before seeing you had already posted this.  I think we can close this PR, but nice fix anyway!

PR #2905, opened 2023-03-15
Preserve the user's locale with form submissions

Vinnl:
# References: 
Jira: MNTOR-1316
Figma: N/A

# Description

The user's locale would only get stored for requests for text/html documents. This meant that other requests, such as a POST form submission when adding an additional email address, would not have the user's locale available, e.g. when sending an email to confirm an additional email address.

Only tagging @toufali for review, since he [initially landed it](https://github.com/mozilla/blurts-server/commit/8c40a0e120700970eae277b3835e110c8474e084#diff-c72a907ac323cd2f334ed0e2bd07d15ab62581c4753660c8a0d1c681b30be4b6), and I'm not 100% confident about my changes here.

# How to test

Set your browser Accept-Language to a localised one (e.g. `fr-FR`). Create a new account, and check that the email you receive is in your chosen language. Then go to the Settings page and add another email address. The email you then receive on that other email address should also be in your chosen language, rather than in English.

- [ ] I've added a unit test to test for potential regressions of this bug. - Seems like a bit of a hassle, but could give it a shot.


src/app.js
```diff
@@ -130,8 +130,6 @@ app.use(
 // When a text/html request is received, negotiate and store the requested language
 // Using asyncLocalStorage avoids having to pass req context down through every function (e.g. getMessage())
 app.use((req, res, next) => {
-  if (!req.headers.accept?.startsWith('text/html')) return next()
```
Vinnl, line 133: This is the actual fix of this PR. The changes in the other file don't change the behaviour, but fix a potential issue where we relied on implicit casting, described in more detail in 4ee5ec74.

toufali (approved):
@Vinnl Regarding localePreferences array – great find/fix, thank you!

Regarding the actual fix to send localized emails, your fix works, but it removes a filter that was put in place to avoid unnecessarily negotiating language and updating locale for every resource on a page.  For example, this is the list of resources requested for the landing page:

```
text/html,application/xhtml+xml,application/xml;q=0.9,image/avif,image/webp,*/*;q=0.8
*/*
application/font-woff2;q=1.0,application/font-woff;q=0.9,*/*;q=0.8
application/font-woff2;q=1.0,application/font-woff;q=0.9,*/*;q=0.8
text/css,*/*;q=0.1
text/css,*/*;q=0.1
*/*
*/*
*/*
*/*
*/*
*/*
*/*
*/*
*/*
*/*
*/*
*/*
*/*
text/css,*/*;q=0.1
text/css,*/*;q=0.1
text/css,*/*;q=0.1
text/css,*/*;q=0.1
text/css,*/*;q=0.1
image/avif,image/webp,*/*
image/avif,image/webp,*/*
image/avif,image/webp,*/*
image/avif,image/webp,*/*
image/avif,image/webp,*/*
image/avif,image/webp,*/*
image/avif,image/webp,*/*
image/avif,image/webp,*/*
image/avif,image/webp,*/*
image/avif,image/webp,*/*
image/avif,image/webp,*/*
```

Since we only care about localizing the first request with `text/html`, bypassing the filter means we call a couple fluent functions unnecessarily for each additional request (~35 redundant calls!)

Since it was not possible to use GitHub "suggest" for an alternate fix, I committed a fix, preserving the resource filter.  Feel free to roll it back if you think of something better.

I will say that I don't think my fix is ideal, because a dev will need to remember to set the Accept header when they want a localized response.  I'm pretty sure this will be neglected and we will run into this issue again.  Maybe we just need to update the README 🤷 

An alternative to setting Accept would be to relax the filter so that the localization functions run only on startsWith `text/html` AND `*/*`, the latter of which is sent when no Accept is set.  With this solution, a dev can forget/ignore setting Accept, and localization will still update.  However, it will also update unnecessarily for other requests that are not explicitly set (about 50% of requests listed above) – but not as bad as EVERY request like images.

Do you have any other ideas or thoughts?  Apologies for the word-wall, I wanted to talk about this in real-time but given our timezones figured I'd write it all down.

toufali:
@Vinnl - yeah, it's surprisingly hard to filter out requests since they all seem to fallback on `*/*`!  

In keeping with your suggestion to not require setting Accept, and potential implementation differences using `req.headers.accept?.startsWith()` I [updated the code](https://github.com/mozilla/blurts-server/pull/2905/commits/6b315f7517c68362807c4645385570caf0acf70f) to handle locale using `accepts(req).types()[0]` which is expected to return [types in order of preference](https://github.com/jshttp/accepts#types).

So, in creating a safety/fallback for contributors that don't set Accept, we're calling a few locale functions unnecessarily, but at least it's not for obviously irrelevant requests like images.  From my research, `*/*` will largely be scripts, which one might expect to use localization.

PR #2906, opened 2023-03-15
Update fluent linter, fix placeables and variable comments

flodolo:
This switches to v0.4.* of the linter, which includes checks for missing variable comments, and style checks when using placeables (variables, term and message references)


flodolo:
@Vinnl Since you're the mind behind the placeables check ;-)

Vinnl (approved):
Wow that was quick, thanks @flodolo!

PR #2910, opened 2023-03-16
Don't pause server start for icon download

Vinnl:
Instead of waiting for the icon downloads to finish (which can take a while, even if we already have a cache of the images - disk access is slow), we just continue starting the server, and any sessions that load before the icon download has finished will show the fallback letter-based icons. A page refresh should then result in the actual icons being shown.

### DIFF
--- a/src/utils/hibp.js
+++ b/src/utils/hibp.js
@@ -146,8 +146,13 @@ async function loadBreachesIntoApp (app) {
       // sync the "breaches" table with the latest from HIBP
       await upsertBreaches(breaches)
     }
-    const breachLogoMap = await downloadBreachIcons(breaches)
-    app.locals.breachLogoMap = breachLogoMap
+    // This will be replaced by a map with the breach logos when
+    // `downloadBreachIcons` resolves, but by setting it to an empty Map first,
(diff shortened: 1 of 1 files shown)


src/utils/hibp.js
```diff
@@ -146,8 +146,13 @@ async function loadBreachesIntoApp (app) {
...
     }
-    const breachLogoMap = await downloadBreachIcons(breaches)
-    app.locals.breachLogoMap = breachLogoMap
+    downloadBreachIcons(breaches).then(breachLogoMap => {
+      app.locals.breachLogoMap = breachLogoMap
+    })
+    // This will be replaced by a map with the breach logos when
+    // `downloadBreachIcons` resolves, but by setting it to an empty Map first,
+    // we don't delay the server start - we just won't have breach logos yet.
+    app.locals.breachLogoMap = new Map()
```
mansaj, line 155: Would it be better to put this line above 149, to avoid the potential of race condition 149 already executed (chance is low yes) and `breachLogoMap` is already set and now it's reset

mansaj (approved):

PR #2916, opened 2023-03-21
MNTOR-1440 switch from GTM to gtag js

rhelmer:
# References: 
Jira: MNTOR-1440

# Description

Switch from Google Tag Manager to `gtag.js` and instrument explicitly in the code.


toufali (commented):
Thanks for taking a stab on this – it's quite involved!  Overall, I couldn't tell if gtag provides additional details by default, like the calling context.  For example, an "add email" event – does it automatically include details on if it's called from the Breaches page, or are more args needed in the call?

Happy to work with you further on this – might require an "audit" similar to the toast error messages!

src/client/js/dialog.js
```diff
@@ -29,11 +29,15 @@ function handleEvent (e) {
     case e.target.matches('a[href*="dialog/"]'):
       e.preventDefault()
       openDialog(e.target.href)
+      gtag('event', 'Verify Email')
       break
     case e.target.matches('button[data-dialog]'):
       openDialog(`dialog/${e.target.dataset.dialog}`)
+      gtag('event', 'Verify Email')
```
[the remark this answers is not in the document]
rhelmer: Ah right thanks, sorry I meant to come back and make that more generic...

.eslintrc.json
```diff
@@ -9,6 +9,9 @@
         "standard",
         "plugin:jsdoc/recommended"
     ],
+    "globals": {
+        "gtag": true
+    },
```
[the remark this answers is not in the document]
rhelmer: Oh yeah that's a good point, we already have a `window` global might as well just use that.

src/client/js/partials/notFound.js
```diff
@@ -7,6 +7,8 @@ if (window.history.length > 1) {
   backCta?.addEventListener('click', () => {
     window.history.back()
 
+    gtag('event', 'Go Back')
```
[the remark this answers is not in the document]
rhelmer: Thanks! Yes after reading the existing analytics docs and reading a UA->GA4 conversion guide, I have a better idea of how to put metadata in here in a way that should show up better in GA.

src/client/js/components/toast-alert.js
```diff
@@ -178,9 +178,11 @@ customElements.define('toast-alert', class extends HTMLElement {
     switch (true) {
       case e.target.matches('button'):
         this.remove()
+        gtag('event', 'Dismiss Toast Alert')
         break
       case e.animationName === 'fade-out':
         this.remove()
+        gtag('event', 'Faded toast alert')
```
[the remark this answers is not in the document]
rhelmer: OK I made this a little nicer too I think (the event category is `Toast Alert`, and it has what happened as an `action` parameter)

src/client/js/dialog.js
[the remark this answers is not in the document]
rhelmer: Sorry about that, I fixed this up to just be category `Dialog` with an open/close action param

src/views/mainLayout.js
```diff
@@ -35,14 +35,27 @@ const mainLayout = data => `
...
-    <script nonce='${data.nonce}'>(function(w,d,s,l,i){w[l]=w[l]||[];w[l].push({'gtm.start':
-    new Date().getTime(),event:'gtm.js'});var f=d.getElementsByTagName(s)[0],
-    j=d.createElement(s),dl=l!='dataLayer'?'&l='+l:'';j.async=true;j.src=
-    'https://www.googletagmanager.com/gtm.js?id='+i+dl;var n=d.querySelector('[nonce]');
-    n&&j.setAttribute('nonce',n.nonce||n.getAttribute('nonce'));f.parentNode.insertBefore(j,f);
-    })(window,document,'script','dataLayer','${AppConstants.GA4_MEASUREMENT_ID}');</script>
-    <!-- End Google Tag Manager -->
+    <!-- Google tag (gtag.js) -->
+    <script nonce='${data.nonce}' type='module'>
+      import { dntEnabled } from '/js/analytics_dnt-helper.js'
```
toufali: This import works in dev because the modules are not bundled/minified/etc.  However, this doesn't work for production (`npm start`).  To make a module available for import in prod with current setup, we need to decide if it's specific to a partial, or site-wide.  
- If specific to a partial, we could add to `/js/partials/`, ESBuild will process/minify it, and you include the script tag in the partial HTML.
- If site-wide (which I would say this is), we would simply import like the other scripts it in `/js/index.js`.

HOWEVER, the way you're using this in a separate script means that importing it into index.js is not going to be a slam dunk.  You could actually do this in index.js:
`export { dntEnabled } from './analytics_dnt-helper.js'`
Then, you could import it in your separate script tag like so:  `import { dntEnabled } from '/js/index.js'`

But as I investigated all this looking for an easier, more current way of accomplishing our goal, I think it would be more performant and less complex to simply test for `navigator.doNotTrack`, and remove the entire `analytics_dnt-helper.js` module.  I suggest removing this module for a couple reasons:
- it's old and possibly outdated
- we don't support internet explorer
- it's mostly based on `navigator.doNotTrack` anyway, but tries to make it friendly to other browsers, however `navigator.doNotTrack` [has been deprecated](https://developer.mozilla.org/en-US/docs/Web/API/Navigator/doNotTrack), so really this module seems much ado about nothing.

Sorry for the word-wall.  I pushed up my suggestion here: https://github.com/mozilla/blurts-server/pull/2916/commits/89012aff6ac237b8e0737cde51cc2211201933d2  Happy to continue the discussion or rollback if you need – the commit was easier than writing out suggested changes to the various files.

src/client/js/dialog.js
```diff
@@ -29,9 +29,11 @@ function handleEvent (e) {
     case e.target.matches('a[href*="dialog/"]'):
       e.preventDefault()
       openDialog(e.target.href)
+      window.gtag('event', 'Dialog', { action: 'open', page_location: location.href })
       break
     case e.target.matches('button[data-dialog]'):
       openDialog(`dialog/${e.target.dataset.dialog}`)
+      window.gtag('event', 'Dialog', { action: 'open', page_location: location.href })
```
toufali, line 36: Should we just put this call once in the `openDialog()` function below, rather than twice in the switch?  

Additionally, `openDialog()` is called with a path that might be helpful to include – or even `partialName` derived from that path.  This way we can know exactly what dialog was opened, instead of the containing page which could have multiple dialogs.

If you _do_ opt for this suggestion, the gtag should be called somewhere after the line `if (!res.ok) throw new Error('Bad fetch response')`, which signifies the modal was open successfully.  You could still add a gtag before there, or specifically in the event of an error 🤷

src/client/js/partials/settings.js
```diff
@@ -12,7 +12,10 @@ function handleEvent (e) {
   switch (true) {
     case e.type === 'email-added':
       document.querySelector('dialog[data-partial="add-email"]')
-        .addEventListener('close', () => window.location.reload(), { once: true })
+        .addEventListener('close', () => {
+          window.gtag('event', 'Add Email', { action: 'click', page_location: location.href })
+          window.location.reload()
+        }, { once: true })
```
toufali, line 18: This is actually listening for the close event of the dialog box, so I think this would fire off an incorrect event.

If you want to track that a user has clicked the "add email" button, that might be elsewhere – but I'd imagine button clicks are already tracked by default?  If so, we could probably just delete this gtag addition.

src/client/js/partials/settings.js
```diff
@@ -44,6 +47,7 @@ if (settingsAlertOptionsInputs?.length) {
         throw new Error(`Updating communication option failed: ${err}`)
       }
       event.preventDefault()
+      window.gtag('event', 'Change Email Preference', { action: 'click', page_location: location.href })
```
toufali, line 50: Question: Are button clicks not auto-tracked for free in GA4?

toufali (commented):
Great updates -- we're almost there!  Please pardon my verbosity – I got excited in the heat of investigation

src/client/js/analytics_dnt-helper.js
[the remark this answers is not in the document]
toufali: This comment is redundant if my commit above works.

toufali:
Sorry to tack this on, but I noticed that the GTM code (on the landing page) is the largest file-size by far at 225kB (transferred size is a smaller ~80 kB, but still largest).  I did a quick search and came up with this, but don't know what to make of it or how it applies, other than suggesting we might be loading an oversized container?

https://web.dev/tag-best-practices/#keep-an-eye-on-container-size

src/client/js/partials/settings.js
rhelmer: Hm, I think only outbound clicks are tracked by default, we have "enhanced event measurment" enabled: https://support.google.com/analytics/answer/9216061

But yes that's a typo... let me double-check this.

src/client/js/partials/settings.js
rhelmer: After testing, I don't think so... it looks like only outbound links are tracked by default with our current config. With GTM this is a bit different. We could have a generic `click` handler we always install, but I thought we should try to measure things more explicitly. We also don't want the generic `click` event since you'd have to drill down in the reports to see where it came from.

src/client/js/partials/settings.js
rhelmer: Hm, yeah I actually already moved this to the `add-email` module, so I think this can just be removed.

src/client/js/dialog.js
rhelmer: Yeah after reading the docs and the UA implementation in the previous version of Monitor, I think that setting success/fail parameters in the events is probably useful anyway. Some of Tony's ask around this included that so we might as well try to be consistent.

rhelmer: Well, hm. Actually this wouldn't fail for any reason that is the user's fault... I think just checking for dialog open is probably fine. The line between GA and Sentry can be strange :)

rhelmer:
> Sorry to tack this on, but I noticed that the GTM code (on the landing page) is the largest file-size by far at 225kB (transferred size is a smaller ~80 kB, but still largest). I did a quick search and came up with this, but don't know what to make of it or how it applies, other than suggesting we might be loading an oversized container?
> 
> https://web.dev/tag-best-practices/#keep-an-eye-on-container-size

Hm, we're not using tag manager (at least not directly) but I'll look into this anyway and see if we can minimize this.

src/client/js/dialog.js
```diff
@@ -29,9 +29,11 @@ function handleEvent (e) {
     case e.target.matches('a[href*="dialog/"]'):
       e.preventDefault()
       openDialog(e.target.href)
+      window.gtag('event', 'Dialog', { action: 'open', page_location: location.href })
```
toufali, line 32:
```suggestion
```
I think this event is now covered by the one in `openDialog()`?

src/client/js/dialog.js
toufali, line 36: I think both these gtag calls are now covered by the one in `openDialog()`?

```suggestion
      break
    case e.target.matches('button[data-dialog]'):
      openDialog(`dialog/${e.target.dataset.dialog}`)
```

src/client/js/dialog.js
```diff
@@ -67,6 +70,7 @@ async function openDialog (path) {
     }
   } catch (e) {
     dialogEl.close()
+    window.gtag('event', 'Dialog', { action: 'open', result: 'failed', page_location: location.href })
```
toufali, line 73: Do we want to add a ref to the actual dialog content being called or nah?  I don't mind either way, but just future-casting when a page_location has multiple dialogs and GA can't differentiate.  🤷 

e.g.: `window.gtag('event', 'Dialog', { action: 'open', result: 'failed', page_location: location.href, dialog_content: partialName })`

toufali (approved):
Looks good @rhelmer !  I left one suggestion – I think we added the dialog open event calls to the function, but left them in the switch, which I think results in duplicate calls.  See below!

src/client/js/dialog.js
rhelmer: Thanks!

rhelmer:
@toufali awesome, thanks for all the help with this one! Adding more parameters is fine, just as an FYI I was reading the docs for how these work in GA4 and there are some limits but I don't think we need to worry too much about hitting them: https://support.google.com/analytics/answer/11202874

Event (and user) parameters don't show up in many reports by default, but we can create metrics and dimensions for them which also have limits: https://support.google.com/analytics/answer/10075209?hl=en

So I think we'll probably have an ongoing refactoring of these as we adjust things on the server, since there are limits we'll want to make sure we're not using more than we need. As I mentioned above though we've got a long way to go before we start bumping against these.

PR #2918, opened 2023-03-22
Don't pass unsanitised user input into HTML

Vinnl:
# References: 
Jira: MNTOR-1438
Figma: N/A

# Description

It should now no longer be possible to pass arbitrary HTML (e.g. including malicious JS, though our CSP blocks that) into the query parameters and have that show up on the unsubscribe page. Additionally, the query parameters get validated, and only the relevant parameters are sent to the back-end.

Note: I don't actually know how to get to the unsubscribe page with valid parameters, so I haven't been able to verify that I'm not actually breaking unsubscription - would be good to get some input there.

# How to test

Visit `/user/unsubscribe?'"><marquee><h1>In ur HTML</h1></marquee>` - you should just see an unmodified "Unsubscribe" button.

Also, visit `/user/unsubscribe?token={"__proto__":"{"test":"ouch"}}&hash=hi`, and verify that the token is passed as a verbatim string to the body of the HTTP request sent when you click the "Unsubscribe" button.

- [ ] I've added a unit test to test for potential regressions of this bug. - Lack of UI testing infra.

### DIFF
--- a/src/controllers/unsubscribe.js
+++ b/src/controllers/unsubscribe.js
@@ -13,8 +13,7 @@ import { unsubscribeFromMonthlyReport } from '../utils/email.js'
 function unsubscribePage (req, res) {
   const data = {
     csrfToken: generateToken(res),
-    partial: unsubscribe,
-    queryParams: req.query
+    partial: unsubscribe
   }
 
   res.send(guestLayout(data))
@@ -30,8 +29,7 @@ async function unsubscribeMonthlyPage (req, res) {
 
   const data = {
     csrfToken: generateToken(res),
-    partial: unsubscribeMonthly,
-    queryParams: req.query
+    partial: unsubscribeMonthly
   }
 
   res.send(guestLayout(data))
--- a/src/client/js/partials/unsubscribe.js
+++ b/src/client/js/partials/unsubscribe.js
@@ -9,7 +9,7 @@ function init () {
 }
 
 async function handleEvent (event) {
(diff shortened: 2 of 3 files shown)


flozia:
To your question:
> Note: I don't actually know how to get to the unsubscribe page with valid parameters, so I haven't been able to verify that I'm not actually breaking unsubscription - would be good to get some input there.

Migrating the unsubscribe views over was done in preparation for the refresh. We do not use `/user/unsubscribe` for now, though. That behavior is broken on the current site and the ability to actually unsubscribe users was pushed out to be implemented in V2 for the emails that do not necessarily need the “unsubscribe” functionality. For unsubscribing from the monthly breaches emails (that we decided to drop for the refresh) we use the same existing pattern and users can unsubscribe following the link from the email — e.g: `/user/unsubscribe-monthly?token=test_123`.

src/controllers/unsubscribe.js
```diff
@@ -13,8 +13,7 @@ import { unsubscribeFromMonthlyReport } from '../utils/email.js'
 function unsubscribePage (req, res) {
   const data = {
     csrfToken: generateToken(res),
-    partial: unsubscribe,
-    queryParams: req.query
+    partial: unsubscribe
```
flozia, line 16: I was just looking at this one as well and about to leave a comment.

Vinnl:
OK cool, so we're not currently sending emails that people need to unsubscribe from, so we can make sure it all works when we do. Merging this then!

PR #2920, opened 2023-03-22
Prevent search engines from indexing non-prod envs

Vinnl:
# References: 
Jira: MNTOR-1351
Figma: N/A

# Description

Previously we provided a robots.txt, telling search engines and other bots not to proactively visit our stage and dev sites. That's good for reducing the load, but won't tell them not to add our dev and stage sites to their search indexes. The X-Robots-Tag: noindex header does.

Note that this also required removing the robots.txt, since otherwise that would tell search engines not to crawl those headers, apparently.

See also:
https://developers.google.com/search/docs/crawling-indexing/block-indexing

# How to test

Visit a resource locally, and verify that it has the ` X-Robots-Tag: noindex` header. That doesn't test all the way until seeing that search engines actually do not index our content, but I don't know of a way to do that.

- [ ] All acceptance criteria are met. - Acceptance criteria mentioned adding this _also_, on top of robots.txt, but Google's instructions say that that's wrong.

### DIFF
--- a/src/routes/index.js
+++ b/src/routes/index.js
@@ -17,12 +17,10 @@ import { dialog } from '../controllers/dialog.js'
 import { landingPage } from '../controllers/landing.js'
 import { notFoundPage } from '../controllers/notFound.js'
 import { notFound } from '../middleware/error.js'
-import { robotsTxt } from '../middleware/robots.js'
 
 const router = express.Router()
 
 router.get('/', landingPage)
-router.get('/robots.txt', robotsTxt)
 router.get('*/dialog/:name', dialog)
 
 router.use('/', dockerFlowRoutes)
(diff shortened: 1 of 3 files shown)


src/routes/index.js
```diff
@@ -17,12 +17,10 @@ import { dialog } from '../controllers/dialog.js'
 import { landingPage } from '../controllers/landing.js'
 import { notFoundPage } from '../controllers/notFound.js'
 import { notFound } from '../middleware/error.js'
-import { robotsTxt } from '../middleware/robots.js'
```
Vinnl, line 20: To reiterate from the PR description:

> Note that this also required removing the robots.txt, since otherwise that would tell search engines not to crawl those headers, apparently.
> 
> See also:
> https://developers.google.com/search/docs/crawling-indexing/block-indexing

toufali (approved):
Thanks for the research and implementation!

non-blocking nit: the name `noindex.js` could be confusing since we have a bunch of `index.js` files and one might thing they're related, or that maybe this is a script for when there's no HTML index?  😅 Naming can be fickle, but to me something like `seoIndex.js` would have been more clear.

 Works as expected, though, nice!

PR #2925, opened 2023-03-23
MNTOR-1440 fixup: allow third-party GA cookie for Firefox standard tracking protection

rhelmer:
# References: 
Jira: MNTOR-1440

# Description

`gtag.js` appears to load fine in Firefox, but fails when trying to submit data:

`
Cookie “_ga_[...]” does not have a proper “SameSite” attribute value. Soon, cookies without the “SameSite” attribute or with an invalid value will be treated as “Lax”. This means that the cookie will no longer be sent in third-party contexts. If your application depends on this cookie being available in such contexts, please add the “SameSite=None“ attribute to it. To know more about the “SameSite“ attribute, read https://developer.mozilla.org/docs/Web/HTTP/Headers/Set-Cookie/SameSite
`

Followed by:
`
Cookie “_ga_[...]” has been rejected for invalid domain.
`

# How to test

Use the site with Firefox network devtools open:

## Expected

HTTP `POST` to `https://www.google-analytics.com/g/collect` succeeds with HTTP status `200`

## Actual

HTTP `POST` to `https://www.google-analytics.com/g/collect` fails and devtools displays `Blocked` in the "Status" column and `NS_ERROR_ABORT` in the "Transferred" column.

### DIFF
--- a/src/views/guestLayout.js
+++ b/src/views/guestLayout.js
@@ -41,7 +41,7 @@ const guestLayout = data => `
     <script src='/js/index.js' type='module'></script>
 
     <!-- Google tag (gtag.js) -->
-    <script nonce='${data.nonce}' type='module'>
+    <script nonce='${data.nonce}'>
       if (navigator.doNotTrack !== '1') {
         (function(w,d,s,l,i){w[l]=w[l]||[];w[l].push({'gtm.start':
         new Date().getTime(),event:'gtm.js'});var f=d.getElementsByTagName(s)[0],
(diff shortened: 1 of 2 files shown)


rhelmer:
As an aside - this `<script>` tag for GS is kind of a monster and it doesn't need to be any different across pages (we may want to set a user property if users are logged in, but that would be just a call to `gtag('config')` I think).

I haven't come up with a clever way to share this between the guest/main layouts, I'll file a follow issue for this.

rhelmer:
OK I'm seeing this on (new!) prod with a latest release Firefox (111.0.1 (64-bit)) profile on macOS... I also raised this in Slack with browser folks, as it looks like a warning but I think it's actually enforced.

Vinnl (approved):
I'm assuming this is mostly copy-pasted from the GA docs? It looks OK, but I'm not sure what it's supposed to look like :)

rhelmer:
> I'm assuming this is mostly copy-pasted from the GA docs? It looks OK, but I'm not sure what it's supposed to look like :)

It's... complicated :) `SameSite` cookies are explained in https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie/SameSite and GA4 has documentation about config but it's spread all over the place...

PR #2929, opened 2023-03-23
MNTOR-1395: add async script for recencyToBreachId

mansaj:
# References: 
Jira: MNTOR-1395
Figma: 

# Description
Switch recencyToBreachId to async mode so it can be ran as a kube job

# Screenshot (if applicable)

Not applicable.

# How to test

### DIFF
--- a/src/scripts/kube-jobs/recencyToBreachId.js
+++ b/src/scripts/kube-jobs/recencyToBreachId.js
@@ -87,7 +87,7 @@ console.log(`Start time is: ${startTime}`)
 // load all breaches for ref
 const allBreaches = await getAllBreachesFromDb()
 if (allBreaches && allBreaches.length > 0) console.log('breaches loaded successfully! ', allBreaches.length)
-console.log(JSON.stringify(allBreaches[0]))
+// console.log(JSON.stringify(allBreaches[0]))
 
 // find all subscribers who resolved any breaches in the past,
 // replace recency index with breach id
@@ -146,7 +146,7 @@ for (const subscriber of subscribersArr) {
 
   // check if v2 is changed, if so, upsert the new v2
   newResolutions.useBreachId = true
-  console.log(JSON.stringify(newResolutions))
+  // console.log(JSON.stringify(newResolutions))
   updateCollection.push({ user: subscriber, updatedBreachesResolution: newResolutions })
 }
 
--- a/src/scripts/kube-jobs/recencyToBreachId.js
+++ b/src/scripts/kube-jobs/recencyToBreachId.js
@@ -21,11 +21,12 @@ let subscribersArr = []
 
 const selectAndLockResolutions = async () => {
   const trx = await knex.transaction()
-  let subscribersArr = []
+  let subscribers = []
   try {
-    subscribersArr = await knex.select('id', 'primary_email', 'breach_resolution')
+    subscribers = await knex.select('id', 'primary_email', 'breach_resolution')
       .from('subscribers')
       .whereNotNull('breach_resolution')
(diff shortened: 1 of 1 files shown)


src/scripts/kube-jobs/recencyToBreachId.js
```diff
@@ -0,0 +1,160 @@
...
+        .update({
+          db_migration_2: true
+        })
+        .transacting(trx)
+    }))
+
+    await trx.commit()
+  } catch (error) {
+    await trx.rollback()
+    console.log('select & mark rows failed!! first row:')
```
pdehaan, line 49: Should this be `console.log()` or `console.error()`?
Minor nit, but noticed that you used `console.error()` below on L77:

https://github.com/mozilla/blurts-server/blob/28c393374ca69d9733d8605c0f7a5d56669287a1/src/scripts/kube-jobs/recencyToBreachId.js#L77

src/scripts/kube-jobs/recencyToBreachId.js
```diff
@@ -0,0 +1,160 @@
...
+// find all subscribers who resolved any breaches in the past,
+// replace recency index with breach id
+
+let failedToSelect = true
+while (failedToSelect) {
+  try {
+    subscribersArr = await selectAndLockResolutions()
+    failedToSelect = false
+  } catch (e) {
+    failedToSelect = true
```
pdehaan, line 101: I think this is already set to `true` above on L95.
But does this need some sort of sleep or backoff or total number of attempts before launching itself into the sun?
Not sure how possible it is that `selectAndLockResolutions()` might fail and put this in an infinite loop.

src/scripts/kube-jobs/recencyToBreachId.js
```diff
@@ -0,0 +1,160 @@
...
+import { getAllBreachesFromDb } from '../../utils/hibp.js'
+import { getAllEmailsAndBreaches } from '../../utils/breaches.js'
+const knex = Knex(knexConfig)
+
+const LIMIT = 1000 // with millions of records, we have to load a few at a time
+let subscribersArr = []
+
+const selectAndLockResolutions = async () => {
+  const trx = await knex.transaction()
+  let subscribersArr = []
```
pdehaan, line 24: nit: not sure if you want to rename this w/ an underscore or something to disambiguate it from L20 above w/ the same name. Took me a second to figure out where the L20 variable was getting set from. Spoiler alert: L98

https://github.com/mozilla/blurts-server/blob/28c393374ca69d9733d8605c0f7a5d56669287a1/src/scripts/kube-jobs/recencyToBreachId.js#L98

mansaj: oh good catch, there should just be one

src/scripts/kube-jobs/recencyToBreachId.js
mansaj: Fixed, thanks

src/scripts/kube-jobs/recencyToBreachId.js
mansaj: Fixed, thanks

mansaj: The select is expected to return without error even if we run out of records. The only way it should error out is if the update (lock marking) failed due to race condition. We haven't seen this happen during our tests due to these kubernetes pods being spun up with a slight delay

src/scripts/kube-jobs/recencyToBreachId.js
mansaj: Fixed. Thanks!

PR #2932, opened 2023-03-24
add more allowed domains to CSP imgSrc

toufali:
# References: 
Jira: MNTOR-1463

# Description
Filed via FoxFooding:

> Refused to load the image 'https://mozillausercontent.com/8fbe1aab6a132ff9daa597de1cca6d91' because it violates the following Content Security Policy directive: "img-src 'self' https://profile.stage.mozaws.net/ [www.googletagmanager.com](http://www.googletagmanager.com/)". Not sure if this is why my DevTools network tab is showing a 404 for https://stage.firefoxmonitor.nonprod.cloudops.mozgcp.net/user/undefined URL. (note the fishy fishy `undefined` in the URL)

# How to test
Summon @pdehaan


src/app.js
```diff
@@ -87,17 +87,18 @@ app.use(
 )
 
 const imgSrc = [
-  "'self'"
+  "'self'",
+  'https://www.googletagmanager.com', // Support GA4 per https://developers.google.com/tag-platform/tag-manager/web/csp
+  'https://firefoxusercontent.com',
+  'https://mozillausercontent.com/',
+  'https://monitor.cdn.mozilla.net/'
```
pdehaan, line 94: Y'all still using the CDN?
I thought all the logos were local now, like: `https://stage.<sshhhh>/images/logo_cache/kickstarter.com.ico`

These were the only CDN results I found for a quick grep:
```sh
git grep -n "https://monitor.cdn.mozilla.net/" | cat

.env-dist:98:EDUCATION_VIDEO_URL_RELAY=https://monitor.cdn.mozilla.net/videos/FF_Relay_version_02.mp4
.env-dist:99:EDUCATION_VIDEO_URL_VPN=https://monitor.cdn.mozilla.net/videos/Mozilla_VPN.mp4
routes/email-l10n.js:15:    imgSrc: ["'self'", 'https://monitor.cdn.mozilla.net/'],
server.js:135:  'https://monitor.cdn.mozilla.net/'
server.js:168:        'https://monitor.cdn.mozilla.net/'
src/app.js:94:  'https://monitor.cdn.mozilla.net/'
```

&hellip; but wasn't sure if you're still using those education MP4s and the only mention in src/ seems to be this file. No big deal, but fewer exceptions the better, I s'pose.

pdehaan: But if it fixes this, I'm a happy lad:

Screenshot 2023-03-23 at 6 37 12 PM

toufali: Ah -- actually that's a different issue.  FxA expires after 24hrs (don't quote me on the interval) and we don't handle it properly.  We have multiple tickets filed against this 😅

toufali: signing out and signing back in should result in a "defined" session?

PR #2934, opened 2023-03-24
Mntor 1470/fix broken seo images

toufali:
# References: 
Jira: MNTOR-1470

# Description
- ~Fixes path for images displayed on breach details pages~
- Fixes SVGO cli recursive flag not working as expected:
https://github.com/svg/svgo#cli-usage (suggests -f for recursive)
https://github.com/svg/svgo/issues/1269 (suggests -f _and_ -r)

# Screenshot (if applicable)

Screen Shot 2023-03-24 at 8 59 12 AM

# How to test
visit a breach detail page and confirm images next to recommendations display as expected (see screenshot above)

### DIFF
--- a/src/package.json
+++ b/src/package.json
@@ -12,7 +12,7 @@
     "start": "node app.js",
     "dev": "nodemon app.js",
     "build": "node esbuild & npm run copy:root & npm run copy:webp & npm run copy:png & npm run build:svg",
-    "build:svg": "svgo -f ./client/images/ -o ../dist/images",
+    "build:svg": "svgo -f ./client/images/ -r -o ../dist/images",
     "copy:root": "mkdir -p ../dist/ && cp ./client/*.* ../dist/",
     "copy:webp": "mkdir -p ../dist/images/ && cp -r ./client/images/*.webp ../dist/images/",
     "copy:png": "mkdir -p ../dist/images/email/ && cp -r ./client/images/email/*.png ../dist/images/email/",
--- a/src/client/css/partials/breachDetails.css
+++ b/src/client/css/partials/breachDetails.css
@@ -154,79 +154,79 @@
 }
 
 .breach-detail-recommendation.rec-pw-1 {
-  background-image: url('../../images/recommendation-icons/change-password.svg');
+  background-image: url('/images/recommendation-icons/change-password.svg');
 }
 
 .breach-detail-recommendation.rec-pw-2 {
-  background-image: url('../../images/recommendation-icons/update-passwords.svg');
+  background-image: url('/images/recommendation-icons/update-passwords.svg');
 }
 
 .breach-detail-recommendation.rec-pw-3 {
-  background-image: url('../../images/recommendation-icons/store-safe-place.svg');
+  background-image: url('/images/recommendation-icons/store-safe-place.svg');
 }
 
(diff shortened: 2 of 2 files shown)


src/package.json
```diff
@@ -12,7 +12,7 @@
     "start": "node app.js",
     "dev": "nodemon app.js",
     "build": "node esbuild & npm run copy:root & npm run copy:webp & npm run copy:png & npm run build:svg",
-    "build:svg": "svgo -f ./client/images/ -o ../dist/images",
+    "build:svg": "svgo -f ./client/images/ -r -o ../dist/images",
```
Vinnl, line 15: Pet peeve of mine, but could we use non-abbreviated parameters here, to make it easier to understand for contributors? (I assume this would be `--recursive`?)

Vinnl (approved):
I wonder why root-relative URLs didn't work? I was going to suggest setting a [`<base>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/base) link, but I think that only applies if there _hadn't_ been a `/` in front?

mansaj (approved):

src/package.json
toufali: I don't actually even know if `--recursive` would work.  `-f` was supposed to do the trick until I read https://github.com/svg/svgo/issues/1269

I haven't had time to read the entire readme, but as far as I know `-r` is a secret flag.  I haven't looked into the CLI's own help docs – maybe more current info there

rhelmer (approved):

src/package.json
Vinnl: Secret flags? Well that's awesome :P

pdehaan: Secret? It's in `--help`:
```sh
npx svgo --version # 3.0.2
npx svgo --help
Usage: svgo [options] [INPUT...]

Nodejs-based tool for optimizing SVG vector graphics files

Arguments:
  INPUT                      Alias to --input

Options:
  -v, --version              output the version number
  -i, --input <INPUT...>     Input files, "-" for STDIN
  -s, --string <STRING>      Input SVG data string
  -f, --folder <FOLDER>      Input folder, optimize and rewrite all *.svg files
  -o, --output <OUTPUT...>   Output file or folder (by default the same as the input), "-" for STDOUT
  -p, --precision <INTEGER>  Set number of digits in the fractional part, overrides plugins params
  --config <CONFIG>          Custom config file, only .js is supported
  --datauri <FORMAT>         Output as Data URI string (base64), URI encoded (enc) or unencoded (unenc)
  --multipass                Pass over SVGs multiple times to ensure all optimizations are applied
  --pretty                   Make SVG pretty printed
  --indent <INTEGER>         Indent number when pretty printing SVGs
  --eol <EOL>                Line break to use when outputting SVG: lf, crlf. If unspecified, uses platform default.
  --final-newline            Ensure SVG ends with a line break
  -r, --recursive            Use with '--folder'. Optimizes *.svg files in folders recursively.
  --exclude <PATTERN...>     Use with '--folder'. Exclude files matching regular expression pattern.
  -q, --quiet                Only output error messages, not regular status messages
  --show-plugins             Show available plugins and exit
  --no-color                 Output plain text without color
  -h, --help                 display help for comman
```

PR #2945, opened 2023-03-27
MNTOR-1437 Instrument CTAs in authenticated pages for GA

rhelmer:
# References: 
Jira: MNTOR-1437

### DIFF
--- a/src/views/guestLayout.js
+++ b/src/views/guestLayout.js
@@ -53,6 +53,11 @@ const guestLayout = data => `
         }
         window.gtag = gtag
       }
+      document.querySelectorAll('[data-cta-id]').forEach(a =>
+        a.addEventListener('click', e => {
+          gtag('event', 'cta_click', { cta_id: a.dataset.ctaId })
+        })
+      )
       </script>
     <!-- End Google tag (gtag.js) -->
   </head>
@@ -62,7 +67,7 @@ const guestLayout = data => `
         <img class='monitor-logo' srcset='/images/monitor-logo-transparent.webp 213w, /images/monitor-logo-transparent@2x.webp 425w' width='213' height='33' alt='${getMessage('brand-fx-monitor')}'>
       </a>
       <menu>
-        <li><a href='/user/breaches' class='button secondary'>${getMessage('sign-in')}</a></li>
+        <li><a href='/user/breaches' data-cta-id='sign-in-1' class='button secondary'>${getMessage('sign-in')}</a></li>
       </menu>
     </header>
     <main data-partial='${data.partial.name}'>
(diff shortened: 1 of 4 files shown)


rhelmer:
> This should work for these locations, but it won't work if a button or link is added after page load - might be worth adding a Mutation Observer to add/remove event listeners as well? But also fine with considering that out of scope for this.

Oh that's a good idea... this is for the guest pages which I know are all static, but I'll file a followup to do this, it feels more future-proof.

rhelmer:
> > This should work for these locations, but it won't work if a button or link is added after page load - might be worth adding a Mutation Observer to add/remove event listeners as well? But also fine with considering that out of scope for this.
> 
> Oh that's a good idea... this is for the guest pages which I know are all static, but I'll file a followup to do this, it feels more future-proof.

https://mozilla-hub.atlassian.net/browse/MNTOR-1478

toufali (commented):
Thanks @rhelmer!  A few thoughts:

- This PR's title was confusing compared to the code changes, is it intended to be _non-auth pages_?

- I see this code added to the guest layout, but a trigger is included in `breaches.js` which uses the auth'd `main.js` layout, and I don't see the code added there?  Was the intent to add the trigger to `all-breaches.js` instead, which uses the non-auth guest layout?  Otherwise, do we need to add the event listeners via main layout as well?

- The event listeners seem tucked away and could potentially be added to many elements.  A couple thoughts might be to use [event delegation](https://developer.mozilla.org/en-US/docs/Learn/JavaScript/Building_blocks/Events#event_delegation).  Also, there's enough GA code now between both guest and main layouts that it might be worth combining these into a `ga.js` module or similar?

rhelmer:
> 



> Thanks @rhelmer! A few thoughts:
> 
>     * This PR's title was confusing compared to the code changes, is it intended to be _non-auth pages_?
> 
>     * I see this code added to the guest layout, but a trigger is included in `breaches.js` which uses the auth'd `main.js` layout, and I don't see the code added there?  Was the intent to add the trigger to `all-breaches.js` instead, which uses the non-auth guest layout?  Otherwise, do we need to add the event listeners via main layout as well?

Yes the intent was for `all-breaches` thanks.

> 
>     * The event listeners seem tucked away and could potentially be added to many elements.  A couple thoughts might be to use [event delegation](https://developer.mozilla.org/en-US/docs/Learn/JavaScript/Building_blocks/Events#event_delegation).  Also, there's enough GA code now between both guest and main layouts that it might be worth combining these into a `ga.js` module or similar?

Hm well it's using a query selector that should only select if the `[data-cta-id]` data attribute is set, do you think this is likely to matter? Happy to change it if so

PR #2966, opened 2023-04-03
MNTOR-1509 remove doNotTrack handling

rhelmer:
# References: 
Jira: MNTOR-1509

# Description

We are currently trying to read `navigator.doNotTrack`, but this is deprecated and not supported in Safari https://developer.mozilla.org/en-US/docs/Web/API/Navigator/doNotTrack - browsers seem to be moving away from DNT and towards using tracking protection in the client anyway.

I think we should just remove it for now.

### DIFF
--- a/src/views/guestLayout.js
+++ b/src/views/guestLayout.js
@@ -42,28 +42,23 @@ const guestLayout = data => `
 
     <!-- Google tag (gtag.js) -->
     <script nonce='${data.nonce}'>
-      if (navigator.doNotTrack !== '1') {
-        (function(w,d,s,l,i){w[l]=w[l]||[];w[l].push({'gtm.start':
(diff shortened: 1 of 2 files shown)


rhelmer:
Per https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/DNT the DNT header is also no longer recommended:

> Deprecated: This feature is no longer recommended. Though some browsers might still support it, it may have already been removed from the relevant web standards, may be in the process of being dropped, or may only be kept for compatibility purposes. Avoid using it, and update existing code if possible; see the [compatibility table](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/DNT#browser_compatibility) at the bottom of this page to guide your decision. Be aware that this feature may cease to work at any time.

mansaj (approved):
LGTM, just deprecating the navigator.doNotTrack

PR #2967, opened 2023-04-04
MNTOR-1512 - backwards-compatibility with breach query

rhelmer:
# References: 
Jira: MNTOR-1512

# Description

Google Search is linking to Monitor as the top result for "monitor twitter breach" and it returnsa two links:
1. `/?breach=[...]` which is unhandled redirects to `/`
2. `/breach-details/[...]`which works as expected

# Screenshot (if applicable)

image

# How to test

### Expected:

URLs with `/?breach=[...]` redirect to `/breach-details/[...]` automatically.

### Actual:

URLs with `/?breach=[...]` redirect to `/`

### DIFF
--- a/src/controllers/landing.js
+++ b/src/controllers/landing.js
@@ -11,6 +11,12 @@ function landingPage (req, res) {
     nonce: res.locals.nonce
   }
 
+  // Backward-compatibility with Monitor V1, for SEO.
+  if (req.query.breach) {
+    const breach = encodeURIComponent(req.query.breach)
+    return res.redirect(`/breach-details/${breach}`)
+  }
+
   res.send(guestLayout(data))
 }
 


mansaj (approved):

pdehaan:
Hmm, yeah, I think that was the old (somewhat confusing) "See if you were in this breach" page via the homepage: https://web.archive.org/web/20200220040842/https://monitor.firefox.com/

Q: Should this be a default `res.redirect()` or should we try and do some [HTTP 301 "Moved permanently"](https://en.wikipedia.org/wiki/HTTP_301) style redirect?
https://expressjs.com/en/4x/api.html#res.redirect seems to imply we can just add the desired status code (default: [302](https://en.wikipedia.org/wiki/List_of_HTTP_status_codes)) like this, `res.redirect(301, 'http://example.com')`.

Also, how did we find this? Was there a big spike in 404s or redirects somewhere in GCP/AWS? Or does our auto-redirect to the homepage on unmatched routes hide all of our 404s?

rhelmer:
> Hmm, yeah, I think that was the old (somewhat confusing) "See if you were in this breach" page via the homepage: https://web.archive.org/web/20200220040842/https://monitor.firefox.com/
> 
> Q: Should this be a default `res.redirect()` or should we try and do some [HTTP 301 "Moved permanently"](https://en.wikipedia.org/wiki/HTTP_301) style redirect? https://expressjs.com/en/4x/api.html#res.redirect seems to imply we can just add the desired status code (default: [302](https://en.wikipedia.org/wiki/List_of_HTTP_status_codes)) like this, `res.redirect(301, 'http://example.com')`.

I was wondering about this... I figured temp redirect was the safer thing to do, permanent redirect should cause search engines to permanently change their link, so we should make sure this is what we want to do and there there aren't any negative consequences. We've discussed moving these to a mozilla.org domain in the next quarter or two for example.

> Also, how did we find this? Was there a big spike in 404s or redirects somewhere in GCP/AWS? Or does our auto-redirect to the homepage on unmatched routes hide all of our 404s?

I found this because I've been investigating the large number of direct, unattributed traffic in https://mozilla-hub.atlassian.net/browse/MNTOR-1509 so I've been trying to reconcile our server logs with what GA4 is reporting.

I've landed a few PRs that fix browser-specific issues, this is one where I think Google is sending a fair amount of traffic but we issue a redirect before GA runs so it isn't seeing a referrer so we're not sure where the traffic is from.

rhelmer:
> I found this because I've been investigating the large number of direct, unattributed traffic in https://mozilla-hub.atlassian.net/browse/MNTOR-1509 so I've been trying to reconcile our server logs with what GA4 is reporting.

Note that this isn't really new looking at the UA dashboard or at other S&P products data, but it seems very high to me so I suspect there's some measurement error going on. Since it's a brand-new setup seemed like a good time to dig in before we start really relying on the data.

rhelmer:
We should also consider handling UTM codes when we do redirects like this, I think that the redirect runs before GA or any other page would be loaded, so these will be lost. I'll do this in a separate PR.

PR #2968, opened 2023-04-05
Allow scanning for exposures without logging in

Vinnl:
# References: 
Jira: MNTOR-1309, MNTOR-1381, MNTOR-1382 and MNTOR-1383.
Figma: https://www.figma.com/file/iJFOikscJeplRBOuKOtYEW/Onboarding?node-id=259-23583&t=c89tLRxJ801kMwlG-4

# Description

This adds the ability for users ~~located in the US (still verifying if that's a requirement - it's not, e6ecf8265)~~ to scan for breaches and sign up afterwards.

# Screenshot (if applicable)

image

image

image

image

# How to test

~~Set your browser language to `en-US`, then~~ enter an email address on the homepage. Make sure to test with an unbreached email, an email with <6 breaches, and an email with more than six breaches.

- [x] Localization strings (if needed) have been added. (But not submitted for localisation.)
- [x] All acceptance criteria are met. ~~(Except for available globally - this is still unclear.)~~


src/controllers/request-breach-scan.js
```diff
@@ -0,0 +1,72 @@
...
+          .replace('<email>', `<span class="breach-result-email" title="${email}">`)
+          .replace('</email>', '</span>')
+          .replace('<count>', '<span class="breach-result-count">')
+          .replace('</count>', '</span>'),
+      // This is sent in the API response because we can't call `getBreachLogo`
+      // client side, where it would expose AppConstants:
+      logos: breaches.map(breach => getBreachLogo(breach, req.app.locals.breachLogoMap)),
+      // This is sent in the API response because we don't have Fluent on the
+      // client side, and thus can't dynamically localise breached data classes:
+      dataClassStrings: breaches.map(breach => breach.DataClasses.map(/** @param {string} dataClass */ dataClass => getMessage(dataClass)))
```
Vinnl, line 63: This is a kludge, but I don't know of a better way, given that Fluent is not available locally. We really don't have a good store w.r.t. server-side/client-side code sharing :(

src/utils/fluent.js
```diff
@@ -41,6 +41,23 @@ async function initFluentBundles () {
 
   await Promise.allSettled(promises)
 
+  // Fluent files in this directory are not currently localized, but might be
+  // moved to `/locales/` later to submit them for localization:
+  const unlocalizedDirname = resolve('../unlocalized/en')
```
Vinnl, line 46: I added the strings for this feature in an `unlocalized` dir, since this feature is (currently) only available to US users.

src/e2e/pages/landingPage.js
```diff
@@ -28,12 +28,14 @@ export class LandingPage {
...
       this.page.waitForNavigation(),
-      this.signUpButton.click()
+      this.signUpButton.click(),
+      this.page.waitForURL('**/oauth/**')
     ])
   }
 
   async goToSignIn () {
     this.signInButton.click()
+    await this.page.waitForURL('**/oauth/**')
```
Vinnl, line 38: Without this line, Playwright would immediate start finding an `input[name="email"]`, and now it would actually immediately find it, since such an input was added to the landing page. However, it _should_ match the one on the Firefox Accounts page :)

(Sorry for the force pushes, took a couple of tries to get the Playwright test right.)

locales/en/exposure-scan.ftl
```diff
@@ -0,0 +1,34 @@
...
+
+exposure-landing-result-footer-attribution = Breach data provided by <hibp-link>{ -brand-HIBP }</hibp-link>
+
+exposure-landing-result-overflow-hero-lead = Sign in to get clear steps on how to resolve these breaches, view all breaches, and get continuous monitoring for any new known breaches.
+exposure-landing-result-overflow-hero-cta-label = Sign in to resolve breaches
+exposure-landing-result-overflow-footer-cta-label = Sign in to view all
+
+exposure-landing-result-some-hero-lead = Sign in to get clear steps on how to resolve these breaches, view all breaches, and get continuous monitoring for any new known breaches.
+exposure-landing-result-some-hero-cta-label = Sign in to resolve breaches
+exposure-landing-result-some-footer-cta-label = Sign up for alerts
```
flodolo, line 30: I only see this as a button when there are zero alerts. Where is the hero text in the two strings above displayed? If the context is exactly the same as the previous group, we shouldn't duplicate the strings.

src/client/js/partials/exposure-scan.js
```diff
@@ -0,0 +1,109 @@
...
+  init()
+}
+
+async function init () {
+  const urlParams = new URLSearchParams(document.location.search)
+  const emailFromUrl = urlParams.get('email')
+  /** @type {HTMLTemplateElement} */
+  const dataEl = exposureScanPartial.querySelector('#data')
+  dataEl.dataset.email = emailFromUrl
+  const sanitisedEmail = dataEl.dataset.email.trim()
```
rhelmer, line 19: -1 for Commonwealth spelling (sanitise vs. sanitize :P)

src/controllers/request-breach-scan.js
```diff
@@ -0,0 +1,72 @@
...
+async function requestBreachScan (req, res, next) {
+  if (req.method !== 'POST' || typeof req.body?.email !== 'string') {
+    return next()
+  }
+
+  // TODO could share this validation logic with add-email API?
+  const email = req.body.email
+  // Use the same regex as HTML5 email input type
+  // https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/email#basic_validation
+  const emailRegex = /^[a-zA-Z0-9.!#$%&'*+/=?^_`{|}~-]+@[a-zA-Z0-9](?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?(?:\.[a-zA-Z0-9](?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?)*$/
```
rhelmer, line 27: I think this is a good common-sense check, but I don't think it's sanitized enough to make it safe to append to the DOM as a string.

We might want to try to not duplicate it and have a utility function also.

src/controllers/request-breach-scan.js
```diff
@@ -0,0 +1,72 @@
...
+        // on the variables, specifically the count, and we don't run Fluent on
+        // the client side):
+        getMessage(
+          'exposure-landing-result-hero-heading',
+          {
+            email,
+            count: breaches.length
+          }
+        )
+          .replace('<email>', `<span class="breach-result-email" title="${email}">`)
```
rhelmer, line 54: Per above, I'm not confident that this is sanitized that something couldn't sneak in here. Email addresses allow quite a range of characters, depending on sanitizing alone may not give us what we want.

One approach to handling potentially unsafe strings is to apply it using client-side DOM, there are a number of "safe sinks" https://cheatsheetseries.owasp.org/cheatsheets/Cross_Site_Scripting_Prevention_Cheat_Sheet.html#safe-sinks

If you want to stick to doing this on server-side, I'd at least run `encodeURIComponent(email)` but I'd suggest running it by security folks.

src/custom-types.d.ts
```diff
@@ -10,3 +10,11 @@ interface HTMLElement {
...
   shadowRoot: ShadowRoot;
 }
+
+// This lint rule does not apply to type definitions:
+// eslint-disable-next-line no-unused-vars
+type ViewPartial<ViewPartialParams = object> = (data: ViewPartialParams) => string;
+type ViewPartialData<ViewPartialParams = object> = {
+    partial: ViewPartial<ViewPartialParams>,
+    nonce: string
+  } & ViewPartialParams;
```
rhelmer, line 20: nice!

rhelmer (changes requested):
lgtm overall, thanks for adding the extra typescript annotations!

Requesting changes because I'm not totally confident that the `email` query param is safe to use to build the page server-side as-is, I left a few notes and a suggestion to run it by security folks.

src/client/js/partials/exposure-scan.js
mansaj: We should have a linter / autofix for this :D

flozia (commented):
This is looking awesome!

I left minor remarks inline that I’d consider rather nits. The only thing that is holding this back from my point of view would be the points that @rhelmer touched on regarding adding more safeguards for the email — let us at least run `encodeURIComponent` and sync with security.

Additional observations:

1. **Browser back button**
When navigating back with the native browser back button after signing in using the “Sign in to resolve breaches” button, `/scan` returns a `403`.
pr-2968-back

2. **Ellipsis color**
In cases where an email address is overflowing, we show an ellipsis. I like it in the same color as the email better. For that to work we would need to wrap it in a `div` instead of a `span` though — up to you (or UX) to decide.

| Ellipis black | Ellipsis purple |
| --- | --- |
| pr-2968-ellipsis-black | pr-2968-ellipsis-purple |

3. **No breaches found layout**
Having three breaches not found cards seems to be a bit redundant. At first glance, they might be mistaken for loading skeletons. For smaller viewports, they increase the height of the page significantly a
pr-2968-no-breaches
nd also push down the “sign in to resolve breaches” button.

src/utils/fluent.js
Vinnl: Removed again, since this will be availably to everyone ( :tada: ).

locales/en/exposure-scan.ftl
Vinnl: Sorry, I messed these up while juggling the six different CTAs: bab6a2f90269db32e430599808c7701ea2f72fb3

The strings above (`hero-lead` and `hero-cta-label`) are these:

image

The lead text is the same for >6 and <=6 detected breaches, but different from 0 detected breaches. The other strings differ for all three cases, as far as I'm aware, except for the header, which is the same for each (and uses the same string).

So... Not sure, when does something count as the same context? I can deduplicate `exposure-landing-result-overflow-hero-lead` and `exposure-landing-result-some-hero-lead`, but not `exposure-landing-result-none-hero-lead`, even though they're all used in the same position on the page.

src/client/js/partials/exposure-scan.js
Vinnl: Oops sorry, but who came up with the idea of having so many variants of what is already my second language? :P 0dabfdd48

(Actually, a spell checker for variable names isn't a [terrible idea](https://en.wikipedia.org/wiki/HTTP_referer#Etymology).)

src/controllers/request-breach-scan.js
Vinnl: Oh crap yes, nice catch. I didn't intend to inject it server-side, but then of couse I can into the issues with trying to inject the email address into a Fluent variable. !@#$%^&*(

I added bf99ddc08 to do it client-side and reuse the email that has been passed through the safe sink.

src/controllers/request-breach-scan.js
Vinnl: Dammit this was your code :P But yes, moved this into a separate validation function and reused it in both places: e00e5079c

As for not using it to inject into the DOM, see https://github.com/mozilla/blurts-server/pull/2968#discussion_r1159668495.

src/client/js/partials/exposure-scan.js
```diff
@@ -0,0 +1,109 @@
...
+      Accept: 'application/json'
+    },
+    mode: 'same-origin',
+    method: 'POST',
+    body: JSON.stringify({
+      email: sanitisedEmail
+    })
+  })
+
+  if (!res.ok) {
```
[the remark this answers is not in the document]
Vinnl: Yes, good one, and I also updated it to more graciously recover from failures on invalid emails: 8ed8b0ca1

@rhelmer That does mean invalid emails will no longer show up in our logs as errors (though maybe the requests will still be logged?).

src/controllers/request-breach-scan.js
```diff
@@ -0,0 +1,72 @@
...
+  }
+
+  try {
+    const allBreaches = req.app.locals.breaches
+    const breaches = await getBreachesForEmail(getSha1(email), allBreaches, false)
+
+    /** @type {RequestBreachScanSuccessResponse} */
+    const successResponse = {
+      success: true,
+      breaches: breaches.slice(0, 6),
```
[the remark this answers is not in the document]
Vinnl: Yeah, unfortunately one of these runs server-side and one of these runs client-side, so they don't have a shared environment, so I think that fussing around to pass them from server to client is probably more hassle than updating both of them in the unlikely even that we'll change this number in the far future.

src/client/js/partials/exposure-scan.js
```diff
@@ -0,0 +1,109 @@
...
+  exposureScanPartial.querySelector('#exposure-scan-loading').hidden = true
+  exposureScanPartial.querySelector('#exposure-scan-results-none').hidden = false
+}
+
+function getBreachCard (breach, logo, dataClassStrings) {
+  const newCard = document.getElementById('breach-template').content.cloneNode(true)
+  newCard.querySelector('.exposure-scan-breach-company-logo').innerHTML = logo
+  newCard.querySelector('.exposure-scan-breach-company-name').textContent = breach.Title
+  newCard.querySelector('.exposure-scan-breach-added dd').textContent = new Date(breach.AddedDate).toLocaleString(navigator.languages, { year: 'numeric', month: 'long', day: 'numeric' })
+  newCard.querySelector('.exposure-scan-breach-data dd').textContent = formatList(dataClassStrings)
```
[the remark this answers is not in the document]
Vinnl: Added comments to the empty elements in 01ec944c4.

Also happy to add more classes, but the class names had been getting so long already, and this one was already specific enough to be confident that it would not match other `<dd>` elements, that I figured it was probably less visual noise to avoid them.

(Underappreciated benefit of JSX: removing this layer of indirection.)

src/client/js/partials/exposure-scan.js
```diff
@@ -0,0 +1,109 @@
...
+  if (!responseBody.success) {
+    return
+  }
+
+  if (responseBody.total > 6) {
+    showOverflowingBreachResults(responseBody)
+  } else if (responseBody.total > 0) {
+    showSomeBreachResults(responseBody)
+  } else {
+    showNoBreachesResult(responseBody)
```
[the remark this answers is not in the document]
Vinnl: [Asked on Slack.](https://mozilla.slack.com/archives/CEV7MGSQH/p1680778460081159?thread_ts=1680720134.369249&cid=CEV7MGSQH)

src/controllers/request-breach-scan.js
```diff
@@ -0,0 +1,72 @@
...
+ * @typedef {RequestBreachScanErrorResponse | RequestBreachScanSuccessResponse} RequestBreachScanResponse
+ */
+
+/** @type {import('express').RequestHandler<unknown, RequestBreachScanResponse>} */
+async function requestBreachScan (req, res, next) {
+  if (req.method !== 'POST' || typeof req.body?.email !== 'string') {
+    return next()
+  }
+
+  // TODO could share this validation logic with add-email API?
```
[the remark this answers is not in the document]
Vinnl: See https://github.com/mozilla/blurts-server/pull/2968#discussion_r1159704204.

flodolo (approved):
Probably OK to have some duplication, especially if we end up making changes to only one of the cases.

src/controllers/request-breach-scan.js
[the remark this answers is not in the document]
flozia: I see. No worries in that case!

pdehaan:
I searched myself on the [HIBP homepage](https://haveibeenpwned.com/) and did appreciate their wording on sensitive breaches, since I don't think those appear unless you do have an account. I think we're the same, but [sadly?] am not in any sensitive breaches to confirm that.
> Pwned in 12 [data breaches](https://haveibeenpwned.com/FAQs#DataSource) and found no [pastes](https://haveibeenpwned.com/FAQs#Pastes) ([subscribe](https://haveibeenpwned.com/NotifyMe) to search sensitive breaches)

Not sure if we want to update our wording to a more generic "we didn't find your email any public breaches, but there might be some that are hidden unless you subscribe." (but obviously written better since my wording is trash)

Although, checking that same email on Monitor show I've been in 9 breaches, so not sure what the other 3 on HIBP are that aren't currently reported via Monitor.
> 9 Unresolved breaches
> 0 Resolved breaches

flozia:
> @flozia Thanks for that hint about the ellipsis colour - I'd never have figured that out by myself! Took me way too long to get one to show up in the first place. Fixed in [87d7802](https://github.com/mozilla/blurts-server/commit/87d780220fd85ac784db1aa6820b21bb6b432af7).
> 
> I've shared your "No breaches layout" concern with UX, although I'll note that on `.mobile`, I already removed the second and third card.
> 
> As for the back button issue, I can't reproduce that. Mind sharing a screen recording?

@Vinnl Thanks for addressing all the feedback and syncing with UX. Looking at the requirements, I don’t see anything blocking this PR.

If the browser back button issue is something that is reproducible on stage we can address the issue as a follow-up. Here is a clipped screen recording with only the upper part of my browser's window. Let me know if you need any more info or would like to pair for debugging:

https://user-images.githubusercontent.com/13835474/230445950-20554af2-5d3c-4cf9-9dcc-66bc4183954e.mov

rhelmer (changes requested):
While testing this on the dev server, I noticed there's one unfortunate consequence of using a query string here - it's getting into the GA4 event (`page_location` param).

I think what we want to do here is override it for this page, something like:
```
gtag('set', 'page_location', 'https://example.com/page1');
```
This is from the GA4 event docs: https://developers.google.com/analytics/devguides/collection/ga4/reference/config

The GA4 blob in main/guest layout is getting pretty ungainly, we should figure out a way to share that and also to do per-page overrides. 

Sorry I don't want to blow up the scope on this :( but we don't want to get PII into GA if we can help it.

rhelmer:
> While testing this on the dev server, I noticed there's one unfortunate consequence of using a query string here - it's getting into the GA4 event (`page_location` param).
> 
> I think what we want to do here is override it for this page, something like:
> 
> ```
> gtag('set', 'page_location', 'https://example.com/page1');
> ```
> 
> This is from the GA4 event docs: https://developers.google.com/analytics/devguides/collection/ga4/reference/config
> 
> The GA4 blob in main/guest layout is getting pretty ungainly, we should figure out a way to share that and also to do per-page overrides.
> 
> Sorry I don't want to blow up the scope on this :( but we don't want to get PII into GA and email addresses count.

I think just to get this done quickly, you could exclude query params from GA by modifying the config in the guest and main layouts:

```
    'page_location': location.protocol + '//' + location.host + location.pathname,
```

I'll add a commit for this, I think the way the config is done now isn't right (there should be just one big `gtag('config', ...)` call)

src/client/js/partials/exposure-scan.js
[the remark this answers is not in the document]
Vinnl: Added bfe6b0602 to only [show one, without a box shadow and background](https://mozilla.slack.com/archives/CEV7MGSQH/p1680788941552139?thread_ts=1680720134.369249&cid=CEV7MGSQH).

Vinnl:
@rhelmer It's a good callout, though I'm not sure if not sending query params to GA is the solution - I feel like that might trip us up in the future.

Instead (or, well, I didn't revert your commit yet, but I suggest doing that - what do you think?), I added fe2dad16d9b79bb4b45b8eada1359cfae50023b9 to use the hash fragment identifier to pass the email address to the next page. Browsers never send these to the server, so that it doesn't show up in our server logs either, and presumably GA doesn't track them either.

src/client/js/partials/exposure-scan.js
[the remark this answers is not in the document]
rhelmer: Yes I think we could figure it out from server logs if someone was abusing this endpoint for instance, there's also generic rate limiting so that should show up in the sentry logs too.

rhelmer:
> @rhelmer It's a good callout, though I'm not sure if not sending query params to GA is the solution - I feel like that might trip us up in the future.
> 
> Instead (or, well, I didn't revert your commit yet, but I suggest doing that - what do you think?), I added [fe2dad1](https://github.com/mozilla/blurts-server/commit/fe2dad16d9b79bb4b45b8eada1359cfae50023b9) to use the hash fragment identifier to pass the email address to the next page. Browsers never send these to the server, so that it doesn't show up in our server logs either, and presumably GA doesn't track them either.
> 
> @flozia OK, I managed to reproduce your issue (which was annoying - it only occurs if pages are cached, where my network monitor disabled the cache 😭 ). So what happens is that the CSRF token will still be cached, and by the time you get back to that page, it's invalid. I wonder if we should add a cache-control header to every page we have that does HTTP requests to tell the browser not to cache it?

Oh that's a good idea, sure feel free to revert my commit. There's only one other place I know of that we use query params, for backwards compat with the old `?breach=${breachName}` and that just redirects to `/breach-details/${breachName}`, but agreed this could be confusing in the future.

rhelmer (approved):

flozia:
> @flozia OK, I managed to reproduce your issue (which was annoying - it only occurs if pages are cached, where my network monitor disabled the cache 😭 ). So what happens is that the CSRF token will still be cached, and by the time you get back to that page, it's invalid. I wonder if we should add a cache-control header to every page we have that does HTTP requests to tell the browser not to cache it?

@Vinnl Nice find! Since this has the potential to break standard user navigation this sounds like something we should do — or at least create an issue for.

Vinnl:
@flozia Done: https://github.com/mozilla/blurts-server/issues/2979

Will fix conflicts now, clean up the commits, and then merge.

PR #2970, opened 2023-04-06
MNTOR-1509 - re-implement DNT check so it works with Safari

rhelmer:
# References: 
Jira: MNTOR-1509

# Description

The Do Not Track (DNT) header is deprecated and not implemented in all browsers (notably Safari), but until we have a better opt-out mechanism we should keep this in place to be consistent with other Mozilla sites.

### DIFF
--- a/src/views/guestLayout.js
+++ b/src/views/guestLayout.js
@@ -42,28 +42,32 @@ const guestLayout = data => `
 
     <!-- Google tag (gtag.js) -->
     <script nonce='${data.nonce}'>
-      (function(w,d,s,l,i){w[l]=w[l]||[];w[l].push({'gtm.start':
-      new Date().getTime(),event:'gtm.js'});var f=d.getElementsByTagName(s)[0],
-      j=d.createElement(s),dl=l!='dataLayer'?'&l='+l:'';j.async=true;j.src=
-      'https://www.googletagmanager.com/gtag/js?id='+i+dl;var n=d.querySelector('[nonce]');
-      n&&j.setAttribute('nonce',n.nonce||n.getAttribute('nonce'));f.parentNode.insertBefore(j,f);
-      })(window,document,'script','dataLayer','${AppConstants.GA4_MEASUREMENT_ID}');
-      function gtag(){dataLayer.push(arguments);}
-      ${AppConstants.GA4_DEBUG_MODE
-        ? `gtag('config', '${AppConstants.GA4_MEASUREMENT_ID}', { 'debug_mode': true })`
-        : ''}
-      gtag('js', new Date());
-      gtag('config', '${AppConstants.GA4_MEASUREMENT_ID}',
-        { cookie_domain: window.location.hostname, cookie_flags: "SameSite=None;Secure" }
-      );
-      window.gtag = gtag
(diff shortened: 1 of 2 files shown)


src/views/guestLayout.js
```diff
@@ -42,28 +42,33 @@ const guestLayout = data => `
...
-      );
-      window.gtag = gtag
+      if (!navigator.doNotTrack || navigator.doNotTrack !== '1')
+        (function(w,d,s,l,i){w[l]=w[l]||[];w[l].push({'gtm.start':
+        new Date().getTime(),event:'gtm.js'});var f=d.getElementsByTagName(s)[0],
+        j=d.createElement(s),dl=l!='dataLayer'?'&l='+l:'';j.async=true;j.src=
+        'https://www.googletagmanager.com/gtag/js?id='+i+dl;var n=d.querySelector('[nonce]');
+        n&&j.setAttribute('nonce',n.nonce||n.getAttribute('nonce'));f.parentNode.insertBefore(j,f);
+        })(window,document,'script','dataLayer','${AppConstants.GA4_MEASUREMENT_ID}');
+        function gtag(){dataLayer.push(arguments);}
```
rhelmer, line 52: This big blob of code is what's recommended on https://developers.google.com/tag-platform/tag-manager/web/csp to work with script nonce FWIW :)

mansaj: So is this a revert of the previous PR or a little bit more?

mansaj (approved):
Looks to me like a revert of the previous PR

src/views/guestLayout.js
rhelmer: It's slightly different, it checks for `navigator.doNotTrack` explicitly now. Safari doesn't have that API anymore, and the way it was written before it disabled GA erroneously.

rhelmer:
> Looks to me like a revert of the previous PR

It is but I also corrected the check so it works correctly if `navigator.doNotTrack` is undefined (Safari doesn't implement it), and cleaned up the config section as well.

src/views/guestLayout.js
```diff
@@ -42,28 +42,32 @@ const guestLayout = data => `
...
-      function gtag(){dataLayer.push(arguments);}
-      ${AppConstants.GA4_DEBUG_MODE
-        ? `gtag('config', '${AppConstants.GA4_MEASUREMENT_ID}', { 'debug_mode': true })`
-        : ''}
-      gtag('js', new Date());
-      gtag('config', '${AppConstants.GA4_MEASUREMENT_ID}',
-        { cookie_domain: window.location.hostname, cookie_flags: "SameSite=None;Secure" }
-      );
-      window.gtag = gtag
+      if (!navigator.doNotTrack || navigator.doNotTrack !== '1') {
```
toufali, line 45: Shouldn't `navigator.doNotTrack !== '1'` also cover the case when navigator.doNotTrack is undefined?

toufali, line 45: Shouldn't navigator.doNotTrack !== '1' also cover the case when navigator.doNotTrack is undefined?

rhelmer: Hm, yeah for some reason I thought it would throw there, but you're right. I'll clean this up separately, thanks!

PR #2972, opened 2023-04-07
MNTOR-1509 fix for broken GA JS on guest page

rhelmer:
# References: 
Jira: MNTOR-1509
Figma: 

# Description

Testing on dev, I noticed that the GA code in the guest layout was not working correctly. It's hard to test for syntax errors here, and it intentionally doesn't break the page so the e2e test doesn't catch it.

I think we should try to share more of this analytics code between the main and guest layout, maybe as a partial that they import? Anyway, for now this PR fixes up the script and also some small issues like a redundant DNT check.

### DIFF
--- a/src/views/guestLayout.js
+++ b/src/views/guestLayout.js
@@ -42,7 +42,7 @@ const guestLayout = data => `
 
     <!-- Google tag (gtag.js) -->
     <script nonce='${data.nonce}'>
-      if (!navigator.doNotTrack || navigator.doNotTrack !== '1') {
+      if (navigator.doNotTrack !== '1') {
         (function(w,d,s,l,i){w[l]=w[l]||[];w[l].push({'gtm.start':
         new Date().getTime(),event:'gtm.js'});var f=d.getElementsByTagName(s)[0],
         j=d.createElement(s),dl=l!='dataLayer'?'&l='+l:'';j.async=true;j.src=
@@ -56,19 +56,19 @@ const guestLayout = data => `
           cookie_flags: "SameSite=None;Secure",
           debug_mode: ${Boolean(AppConstants.GA4_DEBUG_MODE)}
         });
-        window.gtag = gtag
 
         // Detect CTA clicks on public pages.
         document.querySelectorAll('[data-cta-id]').forEach(a =>
           a.addEventListener('click', e => {
             gtag('event', 'clicked_cta', { cta_id: a.dataset.ctaId })
-          })
-        )
+          }))
       } else {
         function gtag() {
           console.debug("Google Analytics disbled by DNT")
+        }
       }
-      </script>
+      window.gtag = gtag
+    </script>
     <!-- End Google tag (gtag.js) -->
 
   </head>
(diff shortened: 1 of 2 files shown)


src/views/mainLayout.js
```diff
@@ -57,7 +57,7 @@ const mainLayout = data => `
           debug_mode: ${Boolean(AppConstants.GA4_DEBUG_MODE)}
         });
       } else {
-        function gtag() {q
```
rhelmer, line 60: This typo apparently sneaked in here, this else branch is only taken if DNT is disabled, I'll test that as well... we don't have a great way to test this in an automated way (yet, we might be able to get playwright to notice this)

src/views/guestLayout.js
```diff
@@ -42,7 +42,7 @@ const guestLayout = data => `
 
     <!-- Google tag (gtag.js) -->
     <script nonce='${data.nonce}'>
-      if (!navigator.doNotTrack || navigator.doNotTrack !== '1') {
+      if (navigator.doNotTrack !== '1') {
```
rhelmer, line 45: Huh! So I am thoroughly testing for both guest and main layout with:

- Firefox (with and without DNT)
- Chrome (with and without DNT)
- Safari (no DNT)

In the console, Safari evaluates:
```
> navigator.doNotTrack !== '1'
< true
```

But it seems to have the opposite effect when I load the test page, it's very odd.

rhelmer: As an aside, this is super tedious to test. I am sure we can come up with some way to signal whether this DNT switch is working in a playwright test, I'll file a ticket in the backlog.

src/views/partials/analytics.js
```diff
@@ -0,0 +1,39 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import AppConstants from '../../app-constants.js'
+
+export const analytics = `
```
rhelmer, line 7: OK I just moved analytics to a partial so there isn't the duplication between guest/main, it means we'll listen for CTAs on the non-guest pages, but maybe we want to do that anyway?

rhelmer:
> @rhelmer I think your new approach is excellent and I like how GTM isn't loaded at all if DNT is enabled.
> 
> As discussed, I pushed a commit – still a WIP – for discussion. Rather than use a partial which is commonly associated with HTML, I think we're better suited by using an imported module on the client-side, since that's where GA operates anyway and we're working with JS alone, no HTML.
> 

Oh I couldn't get it working as an imported script like this, I did try that first. It was before I did a bunch of refactoring :) 

> If my refactored approach works for you, we still need to discuss 2 things:
> 
>     * nonce

Does it work if we set this on the `<script src='/js/index.js' type='module'></script>` script tag?

It might be best to add a second `<script src='/js/analytics.js' type='module' nonce='${data.nonce}'></script>`


> 
>     * env vars

We only need a few, they could be globals set by the page? Do we have an example of this sort of thing elsewhere?

toufali:
@rhelmer per discussion see updates which remove nonce and add pattern for passing server vars to client, including fallback when running dev mode (no build-phase).

toufali (approved):

PR #2976, opened 2023-04-10
ADR: Queue Services Proposal

rhelmer:
# References: 
Jira: MNTOR-1152

# Description

This is a proposal to prevent the kinds of errors we've seen under heavy load when we see:

1. a very large spike on HIBP (e.g. the Twitter200M breach in January), which causes
2. HIBP to send out many thousands of requests, which in turn
3. causes Monitor to process millions of records and send millions of emails

There are two main parts:

1. use Postgres read replicas for all services that do not need to write, and
2. introduce queues for handling tasks async

For example, for the `api/v1/hibp` endpoint, instead of immediately starting database processing and sending emails etc., we would immediately queue the request and then have background workers doing any computation + database changes separate from the main Monitor app that serves users. This gives us the ability to do flow control and re-try errors when we get exceptionally large spikes.


mansaj:
Moving internal discussions here:
* Re emulator for GCP CloudTasks: there's no official emulator, but there is an open source one that's being considered: https://github.com/aertje/cloud-tasks-emulator
  * The downside of using an unofficial emulator is that it can become out of sync and no longer maintained at any time. 
* Would it be make sense to move our emailing service from AWS to GCP as a part of this work (out of scope for now) given that CloudTasks might integrate better with a GCP emailing service? We are slowly sun-setting AWS anyways.

csadilek (commented):
Thank you for sharing! This makes perfect sense to me. After some research I also landed at GCP Cloud tasks as a viable option.

I just had some comments/questions below which would be great to clarify more as part of this ADR.

docs/adr/0003-use-queue-for-backend-services.md
```diff
@@ -0,0 +1,126 @@
...
+
+Monitor has had occasional service problems when traffic spikes, because of the way it and its
+external service dependencies work. Monitor relies on PostgreSQL both to store information about breaches,
+and it is also necessary for the site to function. Monitor also uses Redis as a cache layer, but if
+Postgres is down or too slow to respond, the Monitor site will start displaying error pages.
+
+* When users add an email address to Monitor, the user is registered with the breach provider (HIBP)
+  * NOTE: this uses ranges of sha256 hash sums to provide k-anonymity, so PII like raw email addresses is not sent to HIBP
+* HIBP calls Monitor's `/hibp` API for each hash range, which (for large breaches) can cause substantial database activity
+* Monitor sends email to user to notify them of breaches, which causes more users to click through to the site, causing more database activity
```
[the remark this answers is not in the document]
rhelmer: Thanks for the review! We found out more about this as a result of investigating it together with SRE - the problem we've run into most often is that the breach data from HIBP won't be handled correctly for some reason (a bug in monitor, a change in the data format, etc) and that unhandled exception causes app servers to report as unhealthy, and k8s starts restarting things.

Since HIBP breach data comes in as a large number of separate connections, the restarts don't really resolve anything and just make the availability problem worse. We're confident that we fixed the specific bug(s) in question, but we're not as sure that we have a general solution for this problem.

rhelmer:
@mansaj and I were discussing this, and realized we could probably get the immediate benefits by just splitting this notification endpoint out to a separate app server / cloud function, and incrementally introduce the queue (since HIBP currently backs-off and re-tries on failure, but I think we shouldn't count on every vendor providing this).

I'll update the proposal to introduce this over a few phases, so we can reduce risk and get the benefit of improving user availability sooner.

rhelmer:
Note that we've started to separate back-end from front-end services, we explored with SRE using cloud functions but it might be too big a lift to do yet, so we're going to keep those running in k8s for now. It still sets us up to be able to introduce queues in the near future though, cloud functions aren't strictly required just nice to have.

rhelmer:
FYI - this is on hold until we get some infrastructure issues figured out - leaving in draft mode for now.

rhelmer:
I've updated the ADR to reflect a few changes we decided on in the meantime:

1. use PubSub instead of CloudTask
2. use an hourly cron job as worker for now (switching to HTTP endpoints for workers later)

This is acceptable for less latency-sensitive tasks like breach alerts, we do want to replace (2) so we can use this for more latency-sensitive tasks such as user-facing features in the near future.

mansaj (approved):
Thanks

docs/adr/0003-use-queue-for-backend-services.md
```diff
@@ -0,0 +1,140 @@
+# Use queues for more resilient backend services
+
+- Status: proposed
```
pdehaan, line 3: Q: is this actually proposed? Sounds like you settled on "GCP Pub/Sub" below.

rhelmer: Yes we did and it's being tested on stage, I'll update this section thanks! I'd like to bring it up in the tech meeting tomorrow and just make sure everyone is aligned now that it's working on stage - we should ship to prod soon if all goes well.

docs/adr/0003-use-queue-for-backend-services.md
```diff
@@ -0,0 +1,140 @@
...
+  - Supports option of using cloud functions vs. having to host workers in k8s
+  - Relatively simple to set up and use compared to other options
+  - Mozilla is already using this cloud vendor
+- Cons:
+  - Locked in to GCP ecosystem
+  - Hard to test locally/offline
+    - No official emulator, unofficial open-source emulators exist
+  - Might be more difficult for external users of open-source
+  - No support for strict ordering
+    - https://cloud.google.com/tasks/docs/common-pitfalls
```
bbangert, line 109: Is this a big deal? Most of the 3rd party notification systems I've used don't claim to send them in order to begin with.

docs/adr/0003-use-queue-for-backend-services.md
```diff
@@ -0,0 +1,140 @@
...
+maintain backwards compatibility during migrations.
+
+## Pros and Cons of the Options
+
+### [GCP Cloud Tasks](https://cloud.google.com/tasks)
+
+- Pros:
+  - Supports option of using cloud functions vs. having to host workers in k8s
+  - Relatively simple to set up and use compared to other options
+  - Mozilla is already using this cloud vendor
```
bbangert, line 102: I think there's two important pros missing here, if the processor is having a bug handling it, one has to hope that it was logged sufficiently to see what the event was. With Cloud Tasks, this capability is part of the service:
> Access and management of individual tasks in a queue

The other important feature, especially in regards to the original issue here of a traffic spike, is the ability to throttle the rate. Cloud Tasks has explicit rate controls while Pub/Sub requires using a Polling worker (vs. HTTP Push) to implement flow control.

rhelmer: Right good points (we are using polling workers, and have started trying parallel k8s jobs which seems to be working well). I'll mention these.

docs/adr/0003-use-queue-for-backend-services.md
rhelmer: Not for Monitor, no.

rhelmer:
Thanks for all the feedback - we've had this running in production for a few weeks, and we've found the [k8s job controller](https://kubernetes.io/docs/concepts/workloads/controllers/job/) to be flexible enough to work as a job queue and allow for flow control. We've been watching the system as breach alerts come in making a list of metrics that need to be monitored.

There will be an opportunity to investigate moving to more managed services such as cloud functions and tasks in the near future as well, which we will continue to discuss.

PR #2977, opened 2023-04-10
MNTOR-24: Feature Flags ADR

mansaj:
# References: 
Jira: MNTOR-24

# Description
We need a robust way to A/B testing features, manage when/what to deploy, and standardize feature flag practices to potentially service multiple code bases / projects in the future


docs/adr/0005-feature-flags.md
```diff
@@ -0,0 +1,145 @@
...
+
+A good naming **convention** I found is as follows:
+```
+section_featurepurpose_service
+```
+
+The feature name example that follows has three parts:
+1. section: we present the name of the section the feature is gating. 
+2. purpose: indicates what the feature does
+3. layer: where in the stack the feature is located
```
pdehaan, line 54: Not sure if these align with the example above.
Should it be this?
1. `section`: &hellip;
2. `featurepurpose`: &hellip;
3. `service`: &hellip;

Or, should `section_featurepurpose_service` change? (if we do change it, "featurepurpose" feels a bit of a mouthful).

mansaj: Thanks Peter! Changed

docs/adr/0005-feature-flags.md
```diff
@@ -0,0 +1,149 @@
...
+### Naming Flags
+Defining a pattern or naming convention is generally a good practice for keeping things organized. Engineers can quickly identify from the name what the feature is about and be able to recognize what areas of the application for which the feature is being used. I'll list a few useful examples of naming structures and demonstrate the practical benefits.
+
+A good naming **convention** I found is as follows:
+```
+section_purpose_service
+```
+
+The feature name example that follows has three parts:
+1. section (optional): we present the name of the section of the app the feature is gating (free/premium).
```
Vinnl, line 52: Could you clarify whether this is an exhaustive list? I'm always struggling figuring out what a "section" is, e.g. is it the settings page? Or a specific section of the settings page, such as Premium settings? What if a feature is available to both premium and free subscribers?

Alternatively, I'd also be fine with just skipping `section`, to avoid [smurf naming](https://blog.codinghorror.com/new-programming-jargon/#21).

docs/adr/0005-feature-flags.md
```diff
@@ -0,0 +1,149 @@
...
+
+An **example** of such names would be:
+```
+premium_subplat_email
+```
+
+In this example, the feature is gating functionality in the "settings" section. It is a part of the "premium" feature. The main area of concern is the "email"
+service.
+
+### Retiring Flags / Zombies
```
Vinnl, line 64: Yes! Thank you for including this :clap: One other radical suggestion I've seen is to disallow feature flags enabling something for everyone, i.e. you can roll out to 95% of users, but to roll out to everyone, you need to remove the flag — but that's probably a bit too radical. :sweat_smile:

One thing that's been helpful on Relay is to define a union of strings [enumerating all active feature flags](https://github.com/mozilla/fx-private-relay/blob/57dde8ff1c5794330232e5ee899870e79facbab1/frontend/src/hooks/api/runtimeData.ts#L6-L16) in TypeScript. Then we can just remove a flag from that type, and TS will tell us which bits of code will need to be updated - because it's often not entire if/else branches you can just delete.

docs/adr/0005-feature-flags.md
```diff
@@ -0,0 +1,149 @@
...
+
+### Flag Tagging
+Attaching information like expiration dates and ownership to a flag are specific examples of a more general capability: the ability to tag flags.
+
+Tagging refers to the ability to attach arbitrary key:value pairs to a feature flag. This concept is employed in infrastructure systems like Kubernetes and AWS. It provides a highly adaptable method for users to associate semantic metadata with entities within the system. Tags can be used to track various aspects of a flag, such as its creation date, expiration date, life-cycle status, flag type, ownership, authorized modifier etc.
+
+## Considered Options
+* Environment Variables
+* Unleashed Open Source Feature Flag (self host)
+* Unleased (managed cloud)
```
Vinnl, line 86:
```suggestion
* [Unleash](https://www.getunleash.io/) Open Source Feature Flag (self host)
* Unleash (managed cloud)
```

Vinnl (approved):
Looking good!

docs/adr/0005-feature-flags.md
```diff
@@ -0,0 +1,149 @@
...
+For more drastic measure, there are teams that opt to place a limit on the number of active flags they have under management. This incentivizes the removal of old flags to make room for a new flag.
+
+It's much easier to spot the zombie flags if our code is being maintained properly after feature complete. The if/elses that usually associate with feature flags should be deleted upon completition of a feature, and with a simple search, we should be able to quickly spot stale feature flags
+
+### Ownership
+Looking ahead, there may come a time when we need a system that enables the ability to assign ownership of a flag
+
+Feature flags offer a powerful way to directly modify the behavior of production systems, acting as a kill switch. The configuration of flags should undergo the same level of change control as the deployment of production code. Associating explicit ownership with a flag allows for the application of change control policies. For instance, individuals can be restricted from modifying flags they do not own, or approval can be required before making feature-flag configuration changes in the production environment.
+
+### Flag Tagging
```
flozia, line 78: Having the ability to tag would be really nice.

docs/adr/0005-feature-flags.md
```diff
@@ -0,0 +1,149 @@
...
+Attaching information like expiration dates and ownership to a flag are specific examples of a more general capability: the ability to tag flags.
+
+Tagging refers to the ability to attach arbitrary key:value pairs to a feature flag. This concept is employed in infrastructure systems like Kubernetes and AWS. It provides a highly adaptable method for users to associate semantic metadata with entities within the system. Tags can be used to track various aspects of a flag, such as its creation date, expiration date, life-cycle status, flag type, ownership, authorized modifier etc.
+
+## Considered Options
```
flozia, line 83: Another option could be having a JSON file with a data structure that we define. When it comes to infrastructure that option would be similar to an `.env`, but it could give us still the ability to add more context like tagging, sub-feature flags and  dependencies on other features. Something like:
```
{
  "flags": [
    {
      "name": "Party mode",
      "key": "party-mode",
      "party-mode": {
        "dev": true,
        "prod": false
      },
      "dependencies": ["confetti"]
    },
    {
      "name": "Confetti",
      "key": "confetti",
      "party-mode": {
        "dev": true,
        "prod": false
      },
      "dependencies": []
    }
  ]
}
```

flozia (approved):

docs/adr/0005-feature-flags.md
```diff
@@ -0,0 +1,149 @@
...
+  ? '<script src="confetti.js"></script>'
+  : "";
+```
+
+* Good, because it's the easiest/quickest way to implement feature flags
+* Good, because the changes can be easily tested in separate environments, programmatically
+* Good, because the changes can be scalable (by simply adding and subtracting from the env var array)
+* Good, because it's easy to maintain and it's centralized
+* Good, because it's easy to learn and understand
+* Bad, because we will become dependent on SRE to make any changes in stage and production environment. However, it's not too bad because we will likely have to involve SRE when releasing to production anyways, and during release is when we will be deciding and modifying the environment variables.
```
maxxcrawford, line 114: thought: This may also be a good security point, but limiting access to who can flip flags.

docs/adr/0005-feature-flags.md
```diff
@@ -0,0 +1,149 @@
...
+
+* Good, because it's the easiest/quickest way to implement feature flags
+* Good, because the changes can be easily tested in separate environments, programmatically
+* Good, because the changes can be scalable (by simply adding and subtracting from the env var array)
+* Good, because it's easy to maintain and it's centralized
+* Good, because it's easy to learn and understand
+* Bad, because we will become dependent on SRE to make any changes in stage and production environment. However, it's not too bad because we will likely have to involve SRE when releasing to production anyways, and during release is when we will be deciding and modifying the environment variables. 
+* Bad, because the "implementation" is limited to the current repo. In a world where we have multiple repos serving different parts of the app, there's no easy way to share
+
+### Unleash (Cloud)
```
maxxcrawford, line 117: Question(s): Are other projects using this service? Has this gone through a vendor security review?

docs/adr/0005-feature-flags.md
```diff
@@ -0,0 +1,149 @@
...
+  * feature flag owner
+  * current flag state
+  * group feature flags
+  * change history
+  * expiration date and cleanup
+* Good, because set up is minimal
+* Good, because it is manged so we have support when issues occur, relatively worry-free
+* Good, because new features get added without costing us valuable engineering time
+* Bad, because we have to pay and probably sign a contract (involving legal)
+* Bad, because we adds an additonal dependency. If Unleash goes down, it will affect our service's availability
```
maxxcrawford, line 128: Question: What would a "Unleash" outage look like? Do all the fences go down and the dinosaurs escape?

docs/adr/0005-feature-flags.md
```diff
@@ -0,0 +1,149 @@
...
+  * feature flag owner
+  * current flag state
+  * group feature flags
+  * change history
+  * expiration date and cleanup
+* Good, because we have visibility into the entire stack
+* Good, because it is more flexible, we can pick and choose what we need
+* Good, because it's open source and we can fork / modify features
+* Bad, because we need more support from SRE
+* Bad, because we have to keep up with open source development
```
maxxcrawford, line 142: Question: What does upkeep look like? Making sure to install security patches/updates?

docs/adr/0005-feature-flags.md
```diff
@@ -0,0 +1,149 @@
...
+* Bad, because we won't have the support that a paying customer would get, have to rely on Github issues mainly for support
+
+## Links <!-- optional -->
+
+* [Unleash: Open source](https://github.com/Unleash/unleash)
+* [Unleash: Hosted](https://www.getunleash.io/)
+* [Feature Flags in Next.js with iron-session](https://doist.dev/posts/feature-flags-iron-session-nextjs)
```
maxxcrawford, line 149: Question: Would this power Opt 1 / Environment Variable?

maxxcrawford (commented):
Mostly questions from my end. Great work, @mansaj!

rhelmer (approved):

docs/adr/0005-feature-flags.md
mansaj: When naming something, I always think about future proofing, and namespaces. To namespace, I think we always want to think from large concepts to smaller concepts.

I don't know if "section" is the best way to word to describe what I was thinking there.. but I was thinking more about the different "versions" an app can exist in: free version, premium version, admin version (potentially)

docs/adr/0005-feature-flags.md
mansaj: I agree, this is also a good DIY option. A lot more extensible. I'll make a separate option for this

mansaj: Added `JSON` as an option. Thanks @flozia

docs/adr/0005-feature-flags.md
mansaj: Agreed

docs/adr/0005-feature-flags.md
```diff
@@ -0,0 +1,184 @@
...
+* Good, because it's the easiest/quickest way to implement feature flags
+* Good, because the changes can be easily tested in separate environments, programmatically
+* Good, because the changes can be scalable (by simply adding and subtracting from the env var array)
+* Good, because it's easy to maintain and it's centralized
+* Good, because it's easy to learn and understand
+* Good, having SRE involved adds another level of sanity check
+* Bad, because we will become dependent on SRE to make any changes in stage and production environment. However, it's not too bad because we will likely have to involve SRE when releasing to production anyways, and during release is when we will be deciding and modifying the environment variables. 
+* Bad, because the "implementation" is limited to the current repo. In a world where we have multiple repos serving different parts of the app, there's no easy way to share
+
+### JSON
```
Vinnl, line 124: What's unclear to me from this description is where this JSON would live? In an env var, or in a file in the repo? If the latter, how easy would it be to have different flags per environment? And then I'd also list as a downside that flipping a flag (e.g. if it introduces a bug that we quickly want to back out) would require a new hotfix release and deploy, which is probably riskier than just redeploying the same code with a different environment variable.

Vinnl: Oh, I suppose the `environments` property adapts it to different environments. Shooting from the hip: maybe we could even add an `envVar` property that looks for a particular value in a feature-flag-specific env var, and then enable (or disable?) it when that value is present? Then we'd have the best of both worlds.

docs/adr/0005-feature-flags.md
mansaj: Not within Mozilla, but they have a list on their website. 
After tech meeting, closing this discussion as we will no longer consider this option

docs/adr/0005-feature-flags.md
mansaj: Yes, all dinosaurs would escape. I'd say part of our website will likely be down also. Not great

docs/adr/0005-feature-flags.md
mansaj: Probably keeping an eye on it from time to time :( 
Pay more attention for major version upgrades

docs/adr/0005-feature-flags.md
mansaj: No longer considering iron-session, but it was a good article

docs/adr/0005-feature-flags.md
mansaj: Hmm good question. I think if it's JSON, the assumption is that the JSON would live in the code.

We definitely need to think about the pros and cons of these two approaches a bit more...
cc @rhelmer @maxxcrawford @flozia @codemist 

1. `env var` approach seems pretty straightforward, but it's actually not in practice:
    1. we actually have a hard time as devs (even SREs) to have a holistic picture of what are the env vars that are currently in place for stage and production environment. These are values that are buried deep inside of config maps and yaml files in moz services repo, which I don't know how many of our team has access
    2. changing `stage` and `prod` env vars require SRE approval / restart the server
2. let's think about the deployment of `json` approach
    1. changes to `stage` can be done by devs via our normal deployment process
    2. changes to `prod` requires SRE approval / restart
    3. I agree with you that the changes would be riskier since it'll be code changes -- but I guess if we do feature flags correctly, feature flags are meant for reducing this exact risk?
    4. we (at least as devs) will have a better view of what's in dev/stage/prod, re the first point under `env var`
    5. Not to mention all the other config benefits from having a json structure already in the doc
    

Separately, I'm thinking maybe the below format makes more sense, a bit more dup in data, but faster retrieval..:
```json
{
  "dev": {
    "name": "Premium Subplat Email", 
    "description": "description",
    "key": "premium_subplat_email",
    "createdAt": "",
    "expiresAt": "",
    ...
  }
  "stage":
  "prod":
}
```

Vinnl: Regarding

> we actually have a hard time as devs (even SREs) to have a holistic picture of what are the env vars that are currently in place for stage and production environment. 

on Relay, we just expose the feature flag's env var's value for inspection ([at `/runtime_data/`](https://relay.firefox.com/api/v1/runtime_data)).

As for

> I agree with you that the changes would be riskier since it'll be code changes -- but I guess if we do feature flags correctly, feature flags are meant for reducing this exact risk?

I guess the question is if we're OK with an always-deployable `main` (in which case we could just push the latest `main` through to production with the feature flag flipped), or if we still want to make sure that commits since the latest deployments are not going to prod. If the former, the risk is indeed not increased, but if the latter, I'd at least mention it as a downside, though I wouldn't necessarily consider it reason not to go ahead with it.

Agreed with the rest of your analysis and your proposal.

mansaj: @Vinnl Good points. 

Some of these feature flag discussions inevitably have to turn into "deployment process" discussions. For Monitor, so far, we haven't gone beyond the "always deployable main" model yet. We did have some tech meetings about potentially adopting a release branch strategy.. which escalates to having a dedicated "release engineer" or someone who is going to be in charge of cherry picking at the time of release.. which escalates to longer release cycles. I think for the size that we are at, the "always deployable main" strategy still seems to work, though I do have to acknowledge that the more people we have, the more chaotic `main` becomes

Vinnl: Oh heh, I'd much prefer the always-deployable `main`, I just wasn't sure if we were there yet :sweat_smile:

docs/adr/0005-feature-flags.md
```diff
@@ -0,0 +1,187 @@
...
+ENABLED_FEATURE_FLAGS = "premium_subplat_email, free_breaches_hibp"
+
+function isFlagEnabled(flag: string): boolean {
+  return (process.env.ENABLED_FEATURE_FLAGS ?? '').split(',').includes(flag);
+}
+
+const confettiScript = isFlagEnabled('party-mode')
+  ? '<script src="confetti.js"></script>'
+  : "";
+```
```
pdehaan, line 116: Q: Does this allow us to do %-based rollouts?
Recently in FxA we did a 15% rollout of some new React work before rolling out to 100% of users.

PR #2980, opened 2023-04-11
MNTOR-981: Fxa Events

mansaj:
# References: 
Jira: MNTOR-981

# Description
Set up FxA event callback URI endpoint to receive and handle events from FxA. Some events we can expect are:
* User password changes
* User profile change (pending FxA investigation, currently not getting any profile change events)
* User deletion
* User subscription changes

This PR contains the logic for handling the first 3. We will wait for the SubPlat work to be done before handling the subscription changes

# How to test
To test the code, we need to take over heroku server, and push this change there. The testing needs to be done on dev because of the URI where the callback is pointing to.
* watch the logs of the heroku server
* create a new account in monitor
* change the FxA password in FxA settings
* receive a `password-change` events, see it in the logs
* delete the account
* receive a `delete` event
* check the database row to make sure that the user is deleted from both `subscriber` and `email_address` tables

# Screenshots

Screenshot 2023-05-04 at 1 01 27 PM

Screenshot 2023-05-04 at 1 01 39 PM


mansaj:
This code is untested. To test it, we'd have to wait for this PR (https://github.com/mozilla-services/cloudops-infra/pull/4817) to go through so we can start receiving events. The testing needs to be done on dev because of the URI where the callback is pointing to.

src/controllers/fxaRpEvents.js
```diff
@@ -0,0 +1,167 @@
...
+const log = mozlog('controllers.fxa-rp-events')
+
+const FXA_PROFILE_CHANGE_EVENT = 'https://schemas.accounts.firefox.com/event/profile-change'
+const FXA_SUBSCRIPTION_CHANGE_EVENT = 'https://schemas.accounts.firefox.com/event/subscription-state-change'
+const FXA_DELETE_USER_EVENT = 'https://schemas.accounts.firefox.com/event/delete-user'
+
+/**
+ * Fetch FxA JWT Public for verification
+ *
+ * @returns {Array} keys an array of FxA JWT keys
```
rhelmer, line 21:
```suggestion
 * @returns {Promise<Array<jwt.JwtPayload> | undefined>} keys an array of FxA JWT keys
```

src/controllers/fxaRpEvents.js
```diff
@@ -0,0 +1,167 @@
...
+ */
+const getJwtPubKey = async () => {
+  jwtKeyUri = `${appConstants.OAUTH_ACCOUNT_URI}/jwt`
+  try {
+    const res = await fetch(jwtKeyUri, {
+      headers: {
+        'Content-Type': 'application/json'
+      }
+    })
+    const { keys } = res
```
rhelmer, line 31: I don't think there is a `keys` property in the [fetch Response](https://developer.mozilla.org/en-US/docs/Web/API/Response)?

src/controllers/fxaRpEvents.js
```diff
@@ -0,0 +1,167 @@
...
+const FXA_SUBSCRIPTION_CHANGE_EVENT = 'https://schemas.accounts.firefox.com/event/subscription-state-change'
+const FXA_DELETE_USER_EVENT = 'https://schemas.accounts.firefox.com/event/delete-user'
+
+/**
+ * Fetch FxA JWT Public for verification
+ *
+ * @returns {Array} keys an array of FxA JWT keys
+ */
+const getJwtPubKey = async () => {
+  jwtKeyUri = `${appConstants.OAUTH_ACCOUNT_URI}/jwt`
```
rhelmer, line 24:
```suggestion
  const jwtKeyUri = `${appConstants.OAUTH_ACCOUNT_URI}/jwt`
```

src/controllers/fxaRpEvents.js
```diff
@@ -0,0 +1,167 @@
...
+  } catch (e) {
+    captureException('Could not get JWT public key', jwtKeyUri)
+  }
+}
+
+/**
+ * Authenticate FxA JWT for FxA relay event requests
+ *
+ * @param {*} req
+ * @returns {object} decoded JWT data, which should contain FxA events
```
rhelmer, line 43:
```suggestion
 * @param {import('express').Request} req
 * @returns {Promise<jwt.JwtPayload>} decoded JWT data, which should contain FxA events
```

src/controllers/fxaRpEvents.js
```diff
@@ -0,0 +1,167 @@
...
+ */
+const authenticateFxaJWT = async (req) => {
+  // Assuming this is how you retrieve your auth header.
+  const authHeader = req?.headers?.authorization
+
+  // Require an auth header
+  if (!authHeader) {
+    captureException('No auth header found', req?.headers)
+    throw UnauthorizedError('No auth header found')
```
rhelmer, line 52:
```suggestion
    throw new UnauthorizedError('No auth header found')
```

src/controllers/fxaRpEvents.js
```diff
@@ -0,0 +1,167 @@
...
+  }
+
+  // Extract the first portion which should be 'Bearer'
+  const headerType = authHeader.substr(0, authHeader.indexOf(' '))
+  if (headerType !== 'Bearer') {
+    throw UnauthorizedError('Invalid auth type')
```
rhelmer, line 58:
```suggestion
    throw new UnauthorizedError('Invalid auth type')
```

src/controllers/fxaRpEvents.js
```diff
@@ -0,0 +1,167 @@
...
+  }
+
+  // The remaining portion, which should be the token
+  const headerToken = authHeader.substr(authHeader.indexOf(' ') + 1)
+
+  // Decode the token, require it to come out ok as an object
+  const token = jwt.decode(headerToken, { complete: true })
+  if (!token || typeof token === 'string') {
+    throw UnauthorizedError('Invalid token type')
```
rhelmer, line 67:
```suggestion
    throw new UnauthorizedError('Invalid token type')
```

src/controllers/fxaRpEvents.js
```diff
@@ -0,0 +1,167 @@
...
+  }
+
+  // Verify we have a key for this kid, this assumes that you have fetched
+  // the publicJwks from FxA and put both them in an Array.
+  const publicJwks = await getJwtPubKey()
+  const jwk = publicJwks.find(j => j.kid === token.header.kid)
+  if (!jwk) {
+    throw UnauthorizedError('No jwk found for this kid: ' + token.header.kid)
```
rhelmer, line 75:
```suggestion
    throw new UnauthorizedError('No jwk found for this kid: ' + token.header.kid)
```

src/controllers/fxaRpEvents.js
```diff
@@ -0,0 +1,167 @@
...
+  // This is the JWT data itself.
+  return decoded
+}
+
+/**
+ * Handler for FxA events, used by FxA as a callback URI endpoint
+ * Example events include FxA user deletion, profile changes, and subscription changes
+ *
+ * @param {*} req
+ * @param {*} res
```
rhelmer, line 92:
```suggestion
 * @param {import('express').Request} req
 * @param {import('express').Response} res
```

src/controllers/fxaRpEvents.js
```diff
@@ -0,0 +1,167 @@
...
+  const authHeader = req?.headers?.authorization
+
+  // Require an auth header
+  if (!authHeader) {
+    captureException('No auth header found', req?.headers)
+    throw UnauthorizedError('No auth header found')
+  }
+
+  // Extract the first portion which should be 'Bearer'
+  const headerType = authHeader.substr(0, authHeader.indexOf(' '))
```
rhelmer, line 56: I think `substr` is deprecated in favor of `substring`, ref: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/substr

src/controllers/fxaRpEvents.js
```diff
@@ -0,0 +1,167 @@
...
+        break
+      case FXA_PROFILE_CHANGE_EVENT: {
+        const updatedProfileFromEvent = decodedJWT.events[event]
+        log.debug('fxa_profile_update', {
+          event,
+          updatedProfileFromEvent
+        })
+
+        // get current profiledata
+        const { fxa_profile_json: currentFxAProfile } = subscriber
```
rhelmer, line 140:
```suggestion
        const { fxa_profile_json: currentFxAProfile } = await subscriber
```

src/controllers/fxaRpEvents.js
```diff
@@ -0,0 +1,167 @@
...
+
+        // merge new event into existing profile data
+        for (const key in updatedProfileFromEvent) {
+          if (currentFxAProfile[key]) currentFxAProfile[key] = updatedProfileFromEvent[key]
+        }
+
+        // update fxa profile data
+        await updateFxAProfileData(subscriber, currentFxAProfile)
```
rhelmer, line 148: `tsc` complains about this `await` but that's because the JSDoc on the callee is wrong :)

src/controllers/fxaRpEvents.js
```diff
@@ -0,0 +1,167 @@
...
+    const res = await fetch(jwtKeyUri, {
+      headers: {
+        'Content-Type': 'application/json'
+      }
+    })
+    const { keys } = res
+    log.info('getJwtPubKey', `fetched jwt public keys from: ${jwtKeyUri} - ${keys}`)
+    return keys
+  } catch (e) {
+    captureException('Could not get JWT public key', jwtKeyUri)
```
rhelmer, line 35: I think this method is intended for exceptions not strings: https://docs.sentry.io/platforms/javascript/usage/

src/controllers/fxaRpEvents.js
```diff
@@ -0,0 +1,167 @@
...
+ * @param {*} req
+ * @returns {object} decoded JWT data, which should contain FxA events
+ */
+const authenticateFxaJWT = async (req) => {
+  // Assuming this is how you retrieve your auth header.
+  const authHeader = req?.headers?.authorization
+
+  // Require an auth header
+  if (!authHeader) {
+    captureException('No auth header found', req?.headers)
```
rhelmer, line 51: Same re: above about this API, unless you're passing the exception I think you want `captureMessage`

src/controllers/fxaRpEvents.js
```diff
@@ -0,0 +1,167 @@
...
+ * @param {*} res
+ * @returns
+ */
+const fxaRpEvents = async (req, res) => {
+  let decodedJWT
+  try {
+    decodedJWT = await authenticateFxaJWT(req)
+  } catch (e) {
+    log.error('fxaRpEvents', e)
+    captureException(e)
```
rhelmer, line 101: I think this is the proper expected use of `captureException` :) And same below.

src/controllers/fxaRpEvents.js
```diff
@@ -0,0 +1,167 @@
...
+  }
+
+  // Verify we have a key for this kid, this assumes that you have fetched
+  // the publicJwks from FxA and put both them in an Array.
+  const publicJwks = await getJwtPubKey()
+  const jwk = publicJwks.find(j => j.kid === token.header.kid)
+  if (!jwk) {
+    throw UnauthorizedError('No jwk found for this kid: ' + token.header.kid)
+  }
+  const jwkPem = jwkToPem(jwk)
```
rhelmer, line 77: I'm not sure this is doing what you want... doesn't `getJwtPubKey()` return FxA's public key in JWT format? It's stored here as `publicJwks` which seems to be assumed to be an array of JWKs, feels like maybe there's a step missing. Or I'm just not understanding the flow here, also possible!

rhelmer (changes requested):
OK I went over the `tsc` results and I think these are the most important parts, but I definitely suggest that we enable this so CI can check it for you (there's a `npm run lint:ts` you can use locally too)

src/controllers/fxaRpEvents.js
mansaj: I wonder if it sends an event to Sentry the same way that captureExcpetion would... Another option is to wrap the message in an `new Error`. I'll test it out.

src/controllers/fxaRpEvents.js
mansaj: thanks,good catch

mansaj:
> This code is untested. To test it, we'd have to wait for this PR (https://github.com/mozilla-services/cloudops-infra/pull/4817) to go through so we can start receiving events. The testing needs to be done on dev because of the URI where the callback is pointing to.

I've tested this code in dev environment and the events seem to come through nicely

src/controllers/fxaRpEvents.js
mansaj: Yes, we take the jwt pubkey provided by fxa and use that for verification

rhelmer (approved):
lgtm!

PR #2983, opened 2023-04-12
MNTOR-971: Landing page section transitions

flozia:
# References: 
Jira: [MNTOR-971](https://mozilla-hub.atlassian.net/browse/MNTOR-971)
Design spec: https://ff-landing-page.webflow.io/#color

# Description

Adds simple appearance transitions to sections on the landing page.

# Screenshot

https://user-images.githubusercontent.com/13835474/231775154-df01d5ce-6a2a-45ec-956b-fe6da7d46671.mov

# How to test

1. Navigate to the landing page: `/`
2. Scroll through the landing page


src/client/js/section-observer.js
```diff
@@ -0,0 +1,74 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+const classNameToObserve = 'section-transition'
+const classNameEntered = `${classNameToObserve}-entered`
+const cueIntervalDuration = 150
+const sectionThreshold = 0.1
```
flozia, line 8: Percentage of the section that has to be in the viewport in order to trigger the appearance transition.

src/client/js/section-observer.js
```diff
@@ -0,0 +1,74 @@
...
+
+let observers
+let cueInterval
+const entryCue = []
```
flozia, line 12: Holds the section entries so they are appearing one after another.

src/client/js/section-observer.js
```diff
@@ -0,0 +1,74 @@
...
+}
+
+if (!observers) {
+  const mediaQuery = window.matchMedia('(prefers-reduced-motion: no-preference)')
+  const allowMotion = mediaQuery && mediaQuery.matches
+
+  const sections = document.getElementsByClassName(classNameToObserve)
+  mediaQuery.addEventListener('change', () => {
+    [...sections].forEach(section => section.classList.add(classNameEntered))
+  })
```
flozia, line 69: Handle the following edge case: A user has the Monitor landing page open and is setting their preferences.

src/client/js/section-observer.js
```diff
@@ -0,0 +1,74 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+const classNameToObserve = 'section-transition'
+const classNameEntered = `${classNameToObserve}-entered`
+const cueIntervalDuration = 150
```
flozia, line 7: That’s a bit shorter than what is shown in the design prototype but feels more responsive. Alignment on this with UX is TBD.

Vinnl (approved):
Have you considered leaving most of your inline comments in this PR as code comments? :grin: 

Other than that, one thing that might be annoying is that, if the JS doesn't load, or loads slowly (or is blocked, or broken), content won't be visible by default. But on the other hand, if we let JS hide the content initially, that could result in a quick flash of content before it animates in, which also isn't great. So I guess I don't have a better solution there :shrug:

src/client/js/section-observer.js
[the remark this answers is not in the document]
flozia: I most definitely did mean queue: [81c85d3](https://github.com/mozilla/blurts-server/pull/2983/commits/81c85d342275adefbc4dfc77188d70dc14e3b1f0).

flozia:
> Have you considered leaving most of your inline comments in this PR as code comments? 😁

I did but now I’ve added them as well: [fb036ad](https://github.com/mozilla/blurts-server/pull/2983/commits/fb036ad587069bad9d59393c804c7703d2b042fa). :)

> Other than that, one thing that might be annoying is that, if the JS doesn't load, or loads slowly (or is blocked, or broken), content won't be visible by default. But on the other hand, if we let JS hide the content initially, that could result in a quick flash of content before it animates in, which also isn't great. So I guess I don't have a better solution there 🤷

Good point. I’m not really happy about this only working with JS as well. The flashing you describe would somewhat defeat the purpose of an appearance transition and in that case, I’d prefer no transition at all. Looking forward to [animation-timeline](https://developer.mozilla.org/en-US/docs/Web/CSS/animation-timeline) evolving out of experimental mode! We could at least take care of instances where JS is not allowed — what do you think about this approach: [e774f14](https://github.com/mozilla/blurts-server/pull/2983/commits/e774f146f1b6a5317e2001e67726d1863479510c)?

src/client/js/section-observer.js
```diff
@@ -0,0 +1,79 @@
...
+  observers = [...sections].map(section => {
+    observer.observe(section)
+    return observer
+  })
+
+  setQueueInterval()
+}
+
+if (!observers) {
+  const mediaQuery = window.matchMedia('(prefers-reduced-motion: no-preference)')
```
toufali, line 66: Is this really necessary if we've already got `@media screen and (prefers-reduced-motion: no-preference)` in the CSS?  Or maybe the CSS at-rule is not necessary?

src/client/js/section-observer.js
toufali: Not sure I understand this edge case 😊

toufali (commented):
This is looking good @flozia! 

It works for me and I had only one "blocker-ish" thought -- I wonder if we should remove the transition from the Hero section?  I don't know if it impacts CLS, but the hero _does_ have a button and forthcoming form.  Perhaps more a UX decision, but I wonder if it would be best to only apply the fly-in effect to sections that are just coming into view – not those already in view?  Or at least not the very first element?

Additional thoughts more about implementation – no blockers just potential simplifiers/optimizers:

- I notice you're tracking when a section is entered/transition has been run.  Have you already tried using `animation` instead of `transition`?  Using `animation` is often well-suited for run-once type behavior.
- For the no-script users, `animation` might also benefit – it works using only a "to" property, which can keep sections visible unless JS is available to inject the proper initial state.  Pseudo-ish code:
 
CSS:
```css
[data-animate] {
  /* available by default to all with no-script defaults */
  opacity: var(--section-animation-opacity, 1);
  transform: translateY(var(--section-animation-y, 0));
}

[data-animate='visible'] {
  /* triggered via script when element in viewport */
  animation: section-animation 1s forwards;
}

@keyframes section-animation {
  to {
    opacity: 1;
  }
}
```
JS:
```js
function init(){
  // we know script is available so we can populate initial state into vars
  // we can add these once in top level in `<html>` and it will apply to all associated elements
  document.documentElement.style.setProperty('--section-animation-opacity', '0')
  document.documentElement.style.setProperty('--section-animation-y', '24px')
}

function handleInteresectionEvent(){
  if (entry.isIntersecting) {
    element.dataset.animate = 'visible'
  }
}
```
HTML:
```html
<section data-animate>...</section>
```

Happy to chat further or pair on this if you like!

src/client/js/section-observer.js
toufali: +1 on speedy transitions! 👍

src/views/partials/landing.js
```diff
@@ -4,8 +4,15 @@
...
-export const landing = data => `
-<section class='hero'>
+export const landing = () => `
+<noscript>
+  <style>
+    .section-transition {
+      opacity: 1;
+    }
+  </style>
+</noscript>
```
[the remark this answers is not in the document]
Vinnl: This works; considering animation sounds good too. It doesn't necessarily deal with slow-loading JS; we could potentially also add a delayed animation in CSS that changes the opacity to 1 after a certain delay, and if the JS loads in time it can disable that and replace it with the actual animation trigger. But also, let's maybe not overthink it, as it's probably pretty much an edge case.

src/client/js/section-observer.js
flozia: The edge case is:
- a user loaded the Monitor site with their motion preferences set to “reduce”
- has the Monitor site open and sets the preferences back to “no-preference”
- in that case the page would be blank until the next reload

Instead of setting the class names I’m setting the CSS variables now.

https://user-images.githubusercontent.com/13835474/232039492-b1a91af7-62bf-4cb7-a754-3f08ec0d0456.mov

src/client/js/section-observer.js
flozia: See comment below: https://github.com/mozilla/blurts-server/pull/2983#discussion_r1166754169.

src/client/css/partials/landing.css
```diff
@@ -236,3 +236,16 @@
...
+  .section-transition {
+    opacity: 0;
+    transition: opacity 350ms ease-out, transform 350ms ease-out;
+    transform: translateY(var(--padding-lg));
+  }
+
+  .section-transition-entered {
+    opacity: 1;
+    transform: translateY(0);
+  }
```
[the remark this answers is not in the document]
flozia: Good idea! I’m using animations instead of transitions now.

src/client/js/section-observer.js
```diff
@@ -0,0 +1,79 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+const classNameToObserve = 'section-transition'
+const classNameEntered = `${classNameToObserve}-entered`
```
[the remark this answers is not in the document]
flozia: We need a way to check which sections we need to add to the queue. Checking for the data attribute `enter-transition` now.

src/views/partials/landing.js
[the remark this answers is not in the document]
flozia: I’ve added a more generic way to add entering transitions for the site and am using CSS variables as suggested.

flozia:
Thanks for the suggestions @toufali! I switched to using animations and I am using `data-attributes` as well as CSS variables instead.

Good point about potential issues we run into with hiding the hero section — I’ll raise that with UX.

src/views/partials/landing.js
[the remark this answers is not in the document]
flozia: … and [9453676](https://github.com/mozilla/blurts-server/pull/2983/commits/94536765ef7ac373b17ed4929b59263335125000) should take care of blocked or JS that is loading too slow.

PR #2984, opened 2023-04-12
Add routes for exposure scan signup and overview

Vinnl:
This should make it easier to work on both in parallel while minimising conflicts.

# References: 
Jira: MNTOR-1544
Figma: https://www.figma.com/file/iJFOikscJeplRBOuKOtYEW/Onboarding?node-id=297%3A26021&t=AExoU4tTaoOmlEt9-4

# Description

This adds a `/user/exposures` route, which should eventually show the exposure scan signup if the user hasn't gone through that signup before, and shows the exposures dashboard otherwise.

Note that I added all files to the `tsconfig.json`; if you need help with defining type definitions, either let me know, or remove the file you're working on from `tsconfig.json`. (Try it out by running `npm run lint:ts` in the project root.)

The new route isn't linked to, so it should not block production deployments.

# Screenshot (if applicable)

image

image

# How to test

Visit `/user/exposures`. You should be prompted to log in, and see the partial for exposure signup if you are logged in. You can flip a boolean in code to see the exposure dashboard partial.

### DIFF
--- a/src/custom-types.d.ts
+++ b/src/custom-types.d.ts
@@ -13,8 +13,23 @@ interface HTMLElement {
 
 // This lint rule does not apply to type definitions:
 // eslint-disable-next-line no-unused-vars
-type ViewPartial<ViewPartialParams = object> = (data: ViewPartialParams) => string;
-type ViewPartialData<ViewPartialParams = object> = {
-    partial: ViewPartial<ViewPartialParams>,
-    nonce: string
+type ViewPartial<ViewPartialParams = object> = (data: ViewPartialParams & { partial: { name: string } }) => string;
+type GuestViewPartialData<ViewPartialParams = object> = {
+    partial: ViewPartial<ViewPartialParams>;
+    nonce: string;
   } & ViewPartialParams;
+type MainViewPartialData<ViewPartialParams = object> = {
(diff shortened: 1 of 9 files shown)


src/custom-types.d.ts
```diff
@@ -13,8 +13,23 @@ interface HTMLElement {
...
+    nonce: string;
   } & ViewPartialParams;
+type MainViewPartialData<ViewPartialParams = object> = {
+    fxaProfile: NonNullable<import('express').Request['user']>['fxa_profile_json'];
+  } & GuestViewPartialData<ViewPartialParams>;
+
+declare namespace Express {
+  export interface Request {
+    user?: {
+      // TODO: Finish the type definition of the user object
```
flozia, line 28: Just FYI: I believe this is the schema for the fxa response: https://github.com/mozilla/fxa/blob/main/packages/fxa-profile-server/lib/routes/profile.js#L63-L77.

flozia (approved):
Looks good — thanks for setting this up!

src/custom-types.d.ts
Vinnl: Ah thanks, added in 0045fdf48.

PR #2987, opened 2023-04-13
MNTOR-1171: Code Splitting

toufali:
# References: 
Jira: MNTOR-1171

# Description
Currently, client-side javascript files are all bundled and loaded for every page regardless of need.  

Example: on Landing page, we currently load a bundle that includes Landing page-specific script and site-wide scripts, as well as unnecessary scripts from Breaches page and Settings page.  

This PR aims to separate global modules used for every page from page-specific modules that should be loaded as-needed.

TODO:
- [x] ~~use dynamic import to~~ auto-load page-specific modules
- [x] rename script and stylesheet files to camelCase, in order to integrate with auto-load
- [x] refactor/simplify dialog module to work with camelCase filenames
- [x] add a linter to enforce camelCase filenames
- [ ] explore adding new "pages" folder to distinguish between full page html and smaller partial snippets
- [ ] explore if any benefit to adding new "dialogs" folder to hold full page html that lives within dialog layout
- [ ] update readme

# How to test
Regression testing on every page


src/client/css/index.css
```diff
@@ -2,8 +2,10 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
+/* Import global styles that apply to every page.
+ * Partial-specific stylesheets are auto-loaded based on filename and should not be added here */
```
Vinnl, line 6: Good callout!

src/views/guestLayout.js
```diff
@@ -42,6 +42,7 @@ const guestLayout = data => `
     <link rel='apple-touch-icon' href='/images/apple-touch-icon.webp' sizes='180x180'>
 
     <script src='/js/index.js' type='module'></script>
+    ${data.skipPartialModule ? '' : `<script src='/js/partials/${data.partial.name}.js' type='module'></script>`}
```
Vinnl, line 45: What's the failure mode if someone forgets to set `skipPartialModule` but doesn't have client-side JS for that partial? Is it just a 404 in the console, or does anything break? If the former, then this looks fine to me, but if the latter, I wonder if we could replace it with a check that the file exists when the controller renders? Or maybe even better, checking which scripts are available when the server starts, to avoid disk access on every request.

But again, if it's just an extra 404, then that's no biggie :)

Vinnl (approved):
The changes look good, and I'm looking forward to that darned error being prevented from now on!

I wonder, while we're in conflictopalooza, and if it's not too much work, if you could also adjust the controller file paths to align with their routes? So that have e.g. `/src/controllers/api/v1/requestBreachScan.js` instead of `/src/controllers/requestBreachScan.js`. I have some minor but consistent struggles to find the right controller, esp. distinguishing between backend and frontend routes. But if that gets you too deep in the weeds, then never mind, we can deal with it some other time.

groovecoder:
I've noticed that Monitor sends `cache-control: public, max-age=0` on CSS files so there's no browser caching at all. Claim: Code-splitting the CSS into more separate network requests without fixing CSS caching will actually make performance WORSE. It will save bandwidth, but increase the total latency needed to serve monitor pages. 😢

flozia (approved):
Looking good!

mansaj (approved):
LGTM!

src/views/guestLayout.js
toufali: it's just an extra 404 🤗

toufali: I did initially have code in the module to fetch the url first and handle `!res.ok` before dynamically importing the module, but it seemed overkill.

toufali:
> I've noticed that Monitor sends `cache-control: public, max-age=0` on CSS files so there's no browser caching at all. Claim: Code-splitting the CSS into more separate network requests without fixing CSS caching will actually make performance WORSE. It will save bandwidth, but increase the total latency needed to serve monitor pages. 😢

Thanks @groovecoder for pointing this out – I've added this as a separate ticket with high-priority.

src/views/mainLayout.js
```diff
@@ -42,6 +42,7 @@ const mainLayout = data => `
     <link rel='apple-touch-icon' href='/images/apple-touch-icon.webp' sizes='180x180'>
 
     <script src='/js/index.js' type='module'></script>
+    ${data.skipPartialModule ? '' : `<script src='/js/partials/${data.partial.name}.js' type='module'></script>`}
```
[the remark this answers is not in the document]
toufali: @flozia do you mean accommodating no-script users?  I think it could be helpful for promotion pages (Landing for ex), but if script is disabled for landing, the newly added form simply doesn't work – the entire rest of the page should still function as a promo page sharing content.  Perhaps we could audit the site with script turned off to see what the impact is – at a certain point (past the landing) our site is just not usable without script...

toufali:
>also adjust the controller file paths to align with their routes? So that have e.g. /src/controllers/api/v1/requestBreachScan.js instead of /src/controllers/requestBreachScan.js

@Vinnl I love this idea.  I think @mansaj has been primarily adding API files, I wonder if there's anything additional to consider?

In effort to get this PR in before new devs start working on Monday, I'll merge this now and create a separate ticket for the API path matching and cc you.

PR #2989, opened 2023-04-14
Add API to store data to scan for exposures

Vinnl:
# References: 
Jira: MNTOR-1543
Figma: https://www.figma.com/file/iJFOikscJeplRBOuKOtYEW/Onboarding?node-id=297%3A26021&t=AExoU4tTaoOmlEt9-4

# Description

This adds an API at `/api/v1/user/exposures/` that allows passing in the data to scan for to find exposures. And adds a button at `/user/exposures` to send mock data to that API, to test it out.

# How to test

Make sure to run the migrations first and add the sandbox `ONEREP_API_KEY` to your env vars (ask me or Amri), then visit http://localhost:6060/user/exposures. Open your browser's network console before you click the button. After clicking it, your user should have a OneRep profile ID, and no longer be shown the button.

- [ ] Commits in this PR are minimal and [have descriptive commit messages](https://chris.beams.io/posts/git-commit/). - Will fixup before merging.
- [ ] I've added a unit test to test for potential regressions of this bug. - The API layer is pretty thing, so not sure how much value it adds, but willing to give it a go if desired.

### DIFF
--- a/src/controllers/exposures.js
+++ b/src/controllers/exposures.js
@@ -6,12 +6,18 @@ import { mainLayout } from '../views/mainLayout.js'
 import { generateToken } from '../utils/csrf.js'
 import { exposuresSetup } from '../views/partials/exposuresSetup.js'
 import { exposuresList } from '../views/partials/exposuresList.js'
+import { listScanResults } from '../external/onerep.js'
 
 /**
  * @type {import('express').RequestHandler}
  */
-async function exposuresPage (req, res) {
-  const showDashboard = await hasSetUpExposureScanning(req.user)
+async function exposuresPage (req, res, next) {
+  if (!req.user) {
+    next()
+    return
+  }
+  const user = req.user
+  const showDashboard = hasSetUpExposureScanning(user)
 
   if (!showDashboard) {
     /**
@@ -21,12 +27,13 @@ async function exposuresPage (req, res) {
       partial: exposuresSetup,
       csrfToken: generateToken(res, req),
       nonce: res.locals.nonce,
-      fxaProfile: req.user?.fxa_profile_json
+      fxaProfile: user.fxa_profile_json
     }
     res.send(mainLayout(data))
     return
   }
 
+  const scanResults = await listScanResults(user.onerep_profile_id)
   /**
    * @type {MainViewPartialData<import('../views/partials/exposuresList').PartialParameters>}
    */
@@ -34,20 +41,19 @@ async function exposuresPage (req, res) {
     partial: exposuresList,
     csrfToken: generateToken(res, req),
     nonce: res.locals.nonce,
-    fxaProfile: req.user?.fxa_profile_json
+    fxaProfile: user.fxa_profile_json,
+    scanResults
   }
 
   res.send(mainLayout(data))
 }
 
 /**
- * @param {import('express').Request['user']} user
- * @returns {Promise<boolean>} Whether the user has set up exposure scanning already
+ * @param {NonNullable<import('express').Request['user']>} user
+ * @returns {user is import('express').Request['user'] & { onerep_profile_id: number }} Whether the user has set up exposure scanning already
  */
-async function hasSetUpExposureScanning (user) {
-  // TODO: Once the back-end supports storing a user's exposure scan data,
-  //       check whether it has been entered:
-  return false
+function hasSetUpExposureScanning (user) {
+  return typeof user.onerep_profile_id === 'number'
 }
 
 export { exposuresPage }
--- a/src/db/tables/subscribers.js
+++ b/src/db/tables/subscribers.js
@@ -124,6 +124,18 @@ async function removeFxAData (subscriber) {
   return updatedSubscriber
 }
 
+/**
+ * @param {import('./subscribers_types').SubscriberRow} subscriber
+ * @param {number} onerepProfileId
+ */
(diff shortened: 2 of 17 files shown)


src/controllers/store-exposure-scan-data.js
```diff
@@ -0,0 +1,46 @@
...
+  }
+  const exposureScanInput = parseExposureScanData(req.body)
+  if (exposureScanInput === null) {
+    res.status(400).send({ success: false })
+    return
+  }
+
+  try {
+    const onerepProfileId = await createProfile(exposureScanInput)
+    await activateProfile(onerepProfileId)
```
Vinnl, line 32: When a user enters their data but is not a Premium user yet, we're supposed to do a single scan, without monitoring yet. The API docs mention that the first profile activation triggers a scan (with `initial` as the `reason`), but it's not clear to me if that also activates monitoring and automatic opt-out. If so, we probably don't want to do this here, but I'm not sure if we can then trigger a one-time scan.

src/client/js/partials/exposures-setup.js
```diff
@@ -0,0 +1,58 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+/** @type {HTMLDivElement} */
+// @ts-ignore: We guard against a null value by not calling init():
```
rhelmer, line 6: Hm, are you sure this is needed? The `lint:ts` target and VSCode both don't complain if I remove this comment.

src/db/tables/subscribers.js
```diff
@@ -124,6 +124,18 @@ async function removeFxAData (subscriber) {
   return updatedSubscriber
 }
 
+/**
+ * @param {import('./subscribers_types').SubscriberRow} subscriber
+ * @param {number} onerepProfileId
+ */
+async function setOnerepProfileId (subscriber, onerepProfileId) {
+  await knex('subscribers')
+    .where('id', '=', subscriber.id)
```
rhelmer, line 133: Nit: KNEX also [supports](https://knexjs.org/guide/query-builder.html#where):
```suggestion
    .where('id', subscriber.id)
```

src/db/tables/subscribers_types.d.ts
```diff
@@ -0,0 +1,10 @@
+// This file contains types for subscribers.js until it's fully typed
```
rhelmer, line 1: Thanks for adding type definitions!

src/controllers/exposures.js
```diff
@@ -21,33 +27,33 @@ async function exposuresPage (req, res) {
...
+  const scanResults = await listScanResults(user.onerep_profile_id)
   /**
    * @type {MainViewPartialData<import('../views/partials/exposures-list').PartialParameters>}
    */
   const data = {
     partial: exposuresList,
     csrfToken: generateToken(res, req),
     nonce: res.locals.nonce,
-    fxaProfile: req.user?.fxa_profile_json
+    fxaProfile: req.user?.fxa_profile_json,
```
rhelmer, line 44: Nit: should this use `user` variable as above? I know you didn't really modify this other than appending `,` but still :)

rhelmer (approved):
lgtm, just a few nits you can take or leave. Thanks!

src/client/js/partials/exposures-setup.js
Vinnl: Yeah, that's because I did not finish the typings in this file and thus did not add it to the `tsconfig.json`. I did so now in 254764a1b, though for our client-side partials, with all cajoling going on to obtain elements and string from a DOM that is disconnected from the script, and thus lots of non-null assertions being needed, I'm not sure how much value typing provides there.

src/db/tables/subscribers.js
Vinnl: Ah, good to know. I hadn't used Knex before, so I think I just got this from the docs: 97139c2ea

src/controllers/exposures.js
Vinnl: Oh heh, yeah that's more consistent: 305a61afe

src/controllers/store-exposure-scan-data.js
Vinnl: Updated, because I found the [Scan Pathway](https://docs.onerep.com/#section/Scan-Pathway) docs, which provided the answer (you can scan without activating the profile).

src/client/js/partials/exposures-setup.js
Vinnl: Oh that's right, I didn't add it to `tsconfig.json` because it references the `request-breach-scan` controller module, which references a couple of untyped modules and thus causes errors. 4ecc4ca1c

cadeyrn:
@Vinnl I see that you use the OneRep API. But OneRep does not seem to work for people outside of the USA. Will this feature be US only or are there any plans for a similar feature for people in other parts of the world, like Europe?

Vinnl:
Hi @cadeyrn, AFAIK we'd always like to make features as widely available as possible, technical, legal and pricing constraints allowing, but try not to make promises about that before we're sure that we can keep them - especially not in off-hand comments on GitHub :) I expect [the Distilled blog](https://blog.mozilla.org/en/category/privacy-security/) to make announcements when we have any to make, so I'd recommend following that.

PR #2996, opened 2023-04-17
MNTOR-1130: Animations preview

flozia:
# References: 
Jira: [MNTOR-1130](https://mozilla-hub.atlassian.net/browse/MNTOR-1130)

# Description

Adds a test preview for Lottie animations.

# Screenshot

mntor-1130-lottie-preview

# How to test

1. Navigate to `/preview/animations` — the `/preview` route should be only available for `dev` and `staging`
2. Select and play/pause animations


src/client/js/partials/animationPreview.js
```diff
@@ -0,0 +1,55 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import '../../lib/lottie.min.js'
```
flozia, line 5: I would really like to use the node module `lottie-web` here.

rhelmer (changes requested):
preview lgtm overall, question about whether (and how) we could support `prefers-reduced-motion`.

toufali (commented):
Thanks for the initial exploration here!  A few thoughts:
- as written, this wouldn't work in prod because `npm start` bundles code in `client/js/`.  We'd probably want to move `lib/` into `client/js/`, and then have ESBuild watch that directory as we do for the partials directory.  That being said, I was not able to get this module running after processing with ESBuild – ESBuild (or was it ESLint) issued warnings for this package, suggesting that lottie isn't well suited for module imports (also note the use of `eval`):
`[WARNING] The CommonJS "module" variable is treated as a global variable in an ECMAScript module and may not work as expected [commonjs-variable-in-esm]`
`[WARNING] Using direct eval with a bundler is not recommended and may cause problems`
TL:DR we might consider a script tag to avoid processing with ESBuild, and further we could consider use one of the CDN links provided (adjusting CSP as needed)

- Loading the minified version of the lottie package was still very heavy at ~300kB. To me, it suggests this package might be overkill for the scope of our product? Especially since the animations I previewed from UX appeared to simple transformations (rotation, translate, etc).  What I'm trying to say is, I wonder if we might push for a simpler alternative since our product does not depend on complex SVG animations enabled by Lottie?  A discussion to be had with PM/UX.

- Have you worked with SVG animations before?  I wonder if the scope of this spike still has time to explore this as an alternative, particularly for simple transformations like the ones UX previewed?  Or perhaps the AE Lottie plugin might export to native SVG without need for the large package load – assuming simple animations only?

PR #2999, opened 2023-04-18
Add eslint-plugin-check-file

jswinarton:
# Description

Add the eslint-plugin-check-file package to the dev requirements. eslint complains about this being missing if it's not there.


Vinnl:
I think this package is already listed in `/src/package.json`; @toufali's work should unify those two.

jswinarton:
You are absolutely correct and it turns out my branch was a couple commits behind main. Sorry for the premature PR!

PR #2961, opened 2023-03-31
Add links to company websites for breach resolution

flozia:
# References: 
Jira: [MNTOR-1504](https://mozilla-hub.atlassian.net/browse/MNTOR-1504)

# Description

Adds a link to the breach resolution when passwords or security questions were involved.

# Screenshot

MNTOR-1504

# How to test

1. Login with an account or add an email address that was exposed in a breach.
2. Visit `/user/breaches`
3. If the breach includes passwords or security questions there should be a link to the company website in the associated resolution step.

# Checklist (Definition of Done)
- [x] Localization strings (if needed) have been added.
- [x] Commits in this PR are minimal and [have descriptive commit messages](https://chris.beams.io/posts/git-commit/).
- [x] I've added or updated the relevant sections in readme and/or code comments
- [x] I've added a unit test to test for potential regressions of this bug.
- [x] Product Owner accepted the User Story (demo of functionality completed) or waived the privilege.
- [x] All acceptance criteria are met.
- [x] Jira ticket has been updated (if needed) to match changes made during the development process.
- [x] Jira ticket has been updated (if needed) with suggestions for QA when this PR is deployed to stage.

PR #3000, opened 2023-04-19
Additional tests for blocklisted breach resolution links

flozia:
Additional tests for the blocked company domain links in the breach resolution as requested and promised in https://github.com/mozilla/blurts-server/pull/2961#discussion_r1170343857.

### DIFF
--- a/src/utils/breachResolution.test.js
+++ b/src/utils/breachResolution.test.js
@@ -4,6 +4,7 @@
 
 import test from 'ava'
 
+import AppConstants from '../appConstants.js'
 import { initFluentBundles, updateLocale } from './fluent.js'
 import { filterBreachDataTypes, appendBreachResolutionChecklist, BreachDataTypes } from './breachResolution.js'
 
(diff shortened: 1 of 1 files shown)


src/utils/breachResolution.test.js
```diff
@@ -158,15 +159,61 @@ test('appendBreachResolutionChecklist: data class with a resolution referring to
...
+        DataClasses: [BreachDataTypes.Passwords, BreachDataTypes.SecurityQuestions],
+        Name: 'blockedCompanyName',
+        Domain: 'blockedCompanyName.com'
+      }]
+    }],
+    unverifiedEmails: []
+  }
+
+  appendBreachResolutionChecklist(userBreachData)
+
```
Vinnl, line 206: I assume this needn't be called twice?

```suggestion
```

Vinnl (approved):
Looking good! Does `t.is` do partial matches? (Since you were able to add the part about the password manager without updating the code.)

src/utils/breachResolution.test.js
flozia: No, it doesn’t. Leftover from when I had both test cases combined in one.

flozia:
> Looking good! Does `t.is` do partial matches? (Since you were able to add the part about the password manager without updating the code.)

The change originally was introduced in another PR by https://github.com/mozilla/blurts-server/pull/2961/commits/94152f52e3da20813118966c0006e0ea8021a13b. During cherry-picking I must have messed something up and it resulted in another commit hash: https://github.com/mozilla/blurts-server/pull/3000/commits/83f1fb95d935318b1305f52f0fe590db1788fd14.

PR #3003, opened 2023-04-20
Proposal: caching static generated assets

Vinnl:
(Draft, so that we can discuss whether this is the right approach.)

# References: 
Jira: MNTOR-1552
Figma: N/A

This adds a hash of our client-side CSS and JS files to their filenames, and adds a function `getAssetPath` that returns that hashed file name given an original filename. I don't like how this requires us to remember to always use that function when referring to static assets, but I'm also not aware of a better way, as long as we're manually writing `<script>` and `<link>` tags.

Open questions:
- Do we want this? If not, what _do_ we want?
- ~~[How do we distinguish between built- and un-built environments?](https://mozilla.slack.com/archives/CEV7MGSQH/p1681999187269279) `NODE_ENV` still seems to be `dev` even when running `npm start`.~~ Fixed [using `npm_lifecycle_event`](https://mozilla.slack.com/archives/CEV7MGSQH/p1682006006342409?thread_ts=1681999187.269279&cid=CEV7MGSQH): a1d3a97b7
- How to deal with assets not handled by ESBuild, e.g. images?
- This still doesn't work; it looks like some script is referencing something under `/oath` and I'm not sure what yet, so I haven't been able to test the logged-in site yet.

So don't merge this, but do give input on the above :)

### DIFF
--- a/src/app.js
+++ b/src/app.js
@@ -170,7 +170,16 @@ try {
 }
 
 app.use(noSearchEngineIndex)
-app.use(express.static(staticPath))
+app.use(express.static(staticPath, {
+  setHeaders: (res, path, stat) => {
+    if (path.substring(0, path.lastIndexOf('.')).endsWith('.hashed')) {
+      // Files with extensions like `.hashed.js` have a content hash in their
+      // file name, and hence will have a new filename if their content changes.
+      // Thus, it can be cached basically forever:
+      res.set('Cache-Control', 'max-age=31536000, immutable')
+    }
+  }
+}))
 app.use(express.json())
 app.use(cookieParser(AppConstants.COOKIE_SECRET))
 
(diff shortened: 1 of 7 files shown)


toufali (commented):
Thank you for exploring what would be needed to go hashed/immutable route!  Agreed that our setup makes hashed more complicated.  I think the additional complexity is overkill for the payoff...

I spent too much time researching this and formed the opinion that we should simply set Cache-Control to something that makes sense and be done!  e.g.
```
app.use(express.static(staticPath, {
  setHeaders: res => res.set('Cache-Control', 'max-age=86400, stale-while-revalidate=86400')
}))
```
Since we're SSR, we have a bonus with this approach in that we shouldn't even need to worry about HTML files.  And as you pointed out, ETags are already set up and should play nice here.  

The simple approach seems plenty for a repeat user, without added complexity and potential bugs to our codebase – happy to continue discussion and get your/other thoughts.

Vinnl:
OK, closing this; this'll probably get handled automatically if we execute on https://github.com/mozilla/blurts-server/pull/3019 anyway.

PR #3010, opened 2023-04-26
MNTOR-1271: Extend errors classes

flozia:
# References: 
Jira: [MNTOR-1271](https://mozilla-hub.atlassian.net/browse/MNTOR-1271)

# Description

Update `Errors` module to be central store for serving unified error messages.

# How to test

Since this PR is replacing error classes across the codebase we just should make sure, that the error types that are being replaced are the ones we would expect to be thrown.

For testing the client-side error that is showing a toast notification optionally one could invert the condition [in this line](https://github.com/mozilla/blurts-server/pull/3010/files#diff-135402cfb34bfc8e8f3df69241e60d14eec552b4edfdbbf87996209784f60c4aR59) and trigger a modal.


src/client/js/dialog.js
```diff
@@ -54,7 +56,11 @@ async function openDialog (path) {
       }
     })
 
-    if (!res.ok) throw new Error('Bad fetch response')
+    if (!res.ok) {
+      throw new ClientError('Something went wrong. Please try again or come back later.', {
+        action: 'toast'
```
pdehaan, line 61: nit: We define an enum in src/client/js/error.js:
https://github.com/mozilla/blurts-server/blob/a80f61a40fd6338ebcd3392fe5de52cec3c8365a/src/client/js/error.js#L6-L9

Not sure if we should be exporting that, importing it here, and using `ErrorActionTypes.Toast`.

src/client/js/error.js
```diff
@@ -0,0 +1,50 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+const genericErrorMessage = 'Something went wrong. Please try again or come back later.'
+const ErrorActionTypes = {
+  None: 'none',
+  Toast: 'toast'
+}
```
pdehaan, line 9: Mentioned briefly (and probably poorly) in previous comment, but should this be exported so other file(s) can use the error type consts?

src/client/js/error.js
```diff
@@ -0,0 +1,50 @@
...
+class ClientError extends Error {
+  /**
+   * @param {string} message
+   * @param {Array<{ action: string, targetHref: string }>} config
+   */
+  constructor (message, config = defaultConfig) {
+    super(message, config)
+
+    this.config = config
+    this.message = message || genericErrorMessage
```
pdehaan, line 21: q: why fall back to `genericErrorMessage` here, versus as a constructor default (like `defaultConfig`)?
Probably doesn't matter, just personally curious I guess. And whether it makes a difference since we call `super(message, config)` with a potentially falsy `message`.

Vinnl (commented):
Intuitively this seems fine to me (pending Peter's questions), but since observability is primarily a backend question, I'd like to defer to @mansaj's expertise for approval.

src/utils/email.js
```diff
@@ -133,7 +133,7 @@ function getUnsubscribeCtaHref ({ subscriber, isMonthlyEmail }) {
   }
 
   if (isMonthlyEmail && !subscriber.primary_verification_token) {
-    throw new Error('subscriber has no primary verification_token')
+    throw new MethodNotAllowedError('subscriber has no primary verification_token')
```
mansaj, line 136: Err 405 "indicates that the web server has recognized a request from a web browser to access the website but rejects the specific HTTP method"

This seems like a 400 or 401

src/utils/error.js
```diff
@@ -2,49 +2,123 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
-class UserInputError extends Error {
+import { getMessage } from './fluent.js'
+
+const genericErrorMessage = 'Something went wrong. Please try again or come back later.'
+const permissionErrorMessage = 'You don’t have permission to access this content.'
+const tooManyRequestsErrorMessage = 'Too many requests received. Please try again later. '
```
mansaj, line 9: Are these messages supposed to be localized? ie a part of a fluent file?

src/utils/hibp.js
```diff
@@ -44,18 +44,18 @@ async function _throttledFetch (url, reqOptions, tryCount = 1) {
       case 429:
         log.info('_throttledFetch', { err: 'Error 429, tryCount: ' + tryCount })
         if (tryCount >= HIBP_THROTTLE_MAX_TRIES) {
-          throw fluentError('error-hibp-throttled')
+          throw FluentError('error-hibp-throttled')
```
mansaj, line 47: Is this InternalServerError?

src/utils/hibp.js
```diff
@@ -156,7 +156,7 @@ async function loadBreachesIntoApp (app) {
     app.locals.breaches = breaches
     app.locals.breachesLoadedDateTime = Date.now()
   } catch (error) {
-    throw fluentError('error-hibp-load-breaches')
+    throw FluentError('error-hibp-load-breaches')
```
mansaj, line 159: Is this also internal server error?
I think we can get rid of the concept of "FluentError" and just write out `getMessage()` when necessary. `FluentError` isn't doing too much as a wrapper imo

Hopefully when we get more of these messages standardized and translated, most if not all errors will be fluent errors

42-star:
I get a lot of those error codes all the time.

Sent from Outlook for Android<https://aka.ms/AAb9ysg>
________________________________

src/client/js/dialog.js
```diff
@@ -54,7 +56,11 @@ async function openDialog (path) {
       }
     })
 
-    if (!res.ok) throw new Error('Bad fetch response')
+    if (!res.ok) {
+      throw new ClientError('Something went wrong. Please try again or come back later.', {
```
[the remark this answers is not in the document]
flozia: Unfortunately, we are currently not able to localize client-side strings and would need to pass them from the backend. We deprioritized adding this functionality in light of upcoming changes that would resolve the issue.

I’m using a default message inside of the error class now so we can just pass `null` — thanks for the suggestion!

src/client/js/dialog.js
flozia: Done in https://github.com/mozilla/blurts-server/pull/3010/commits/18a61ed85714578ca0a05edf669d2bdb2080828c.

src/utils/hibp.js
```diff
@@ -44,18 +44,18 @@ async function _throttledFetch (url, reqOptions, tryCount = 1) {
...
           return await _throttledFetch(url, reqOptions, tryCount)
         }
       default:
-        throw new Error(`bad response: ${response.status}`)
+        throw new InternalServerError(`bad response: ${response.status}`)
     }
   } catch (err) {
     log.error('_throttledFetch', { err })
-    throw fluentError('error-hibp-connect')
+    throw FluentError('error-hibp-connect')
```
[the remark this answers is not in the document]
flozia: Good observation! Added `throw new` to the missing instances: https://github.com/mozilla/blurts-server/pull/3010/commits/375109eb27d3cb69637d0f75e1f8d129eeecc154.

src/utils/error.js
```diff
@@ -2,49 +2,123 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
-class UserInputError extends Error {
+import { getMessage } from './fluent.js'
+
+const genericErrorMessage = 'Something went wrong. Please try again or come back later.'
```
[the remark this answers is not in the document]
flozia: I was planning on adding localized strings after a first pass or as a follow-up for server-side strings. Just to make sure there are no resulting string changes.

src/client/js/error.js
flozia: The thinking was: Passing `null` the arg would not fall back to a default value. Only no value or `undefined` would make that happen and personally, I’d like the option of being able to pass `null` like now [here](https://github.com/mozilla/blurts-server/pull/3010/files#diff-135402cfb34bfc8e8f3df69241e60d14eec552b4edfdbbf87996209784f60c4aR60). Adding `message` to super is definitely redundant in this case, though. Addressed this in commit https://github.com/mozilla/blurts-server/pull/3010/commits/e62334fb364d7e4de8386c6838fcfb43d7896d66 — WDYT?

src/utils/error.js
flozia: We could localize them, but we currently don’t show them anywhere user-facing. It might make sense to only localize them as soon we surface them in case we need them to change.

src/utils/hibp.js
flozia: Good idea — I removed `FluentError` in [21ef539](https://github.com/mozilla/blurts-server/pull/3010/commits/21ef539c1c3ebdbfafb41d37853f42680c42cbd9).

mansaj (approved):
LGTM thanks

src/utils/error.js
rhelmer: My preference would also be to use a consistent language for error messages internally (this will be sent to Sentry or the console), and localize as late in the process as possible, like when we format the final output for users.

src/utils/error.test.js
```diff
@@ -0,0 +1,63 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
```
rhelmer, line 1: Thanks for adding tests!

rhelmer (approved):

PR #3011, opened 2023-04-26
Add custom type definition for breach data object

maxxcrawford:
Note: This PR doesn't actually type anything. It's for future usage. :)

### DIFF
--- a/src/customTypes.d.ts
+++ b/src/customTypes.d.ts
@@ -68,3 +68,22 @@ declare module 'mozlog' {
 interface Window {
   gtag: any
 }
+
+declare namespace breach {
+  let Name: string;
+  let Title: string;
+  let Domain: string;
+  let BreachDate: string;
+  let AddedDate: string;
+  let ModifiedDate: string;
+  let PwnCount: number;
+  let Description: string;
+  let LogoPath: string;
+  let DataClasses: Array<string>[];
+  let IsVerified: boolean;
+  let IsFabricated: boolean;
+  let IsSensitive: boolean;
+  let IsRetired: boolean;
+  let IsSpamList: boolean;
+  let IsMalware: boolean;
+}
--- a/src/customTypes.d.ts
+++ b/src/customTypes.d.ts
@@ -70,20 +70,20 @@ interface Window {
 }
 
 declare namespace breach {
-  let Name: string;
-  let Title: string;
-  let Domain: string;
-  let BreachDate: string;
-  let AddedDate: string;
-  let ModifiedDate: string;
-  let PwnCount: number;
-  let Description: string;
-  let LogoPath: string;
-  let DataClasses: Array<string>[];
-  let IsVerified: boolean;
-  let IsFabricated: boolean;
-  let IsSensitive: boolean;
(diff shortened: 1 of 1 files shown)


src/customTypes.d.ts
```diff
@@ -63,3 +63,22 @@ declare module 'mozlog' {
 interface Window {
   gtag: any
 }
+
+declare namespace breach {
```
maxxcrawford, line 67: Sample data: 

```
{
  "Name": "000webhost",
  "Title": "000webhost",
  "Domain": "000webhost.com",
  "BreachDate": "2015-03-01",
  "AddedDate": "2015-10-26T23:35:45Z",
  "ModifiedDate": "2017-12-10T21:44:27Z",
  "PwnCount": 14936670,
  "Description": "In approximately March 2015, the free web hosting provider <a href=\"http://www.troyhunt.com/2015/10/breaches-traders-plain-text-passwords.html\" target=\"_blank\" rel=\"noopener\">000webhost suffered a major data breach</a> that exposed almost 15 million customer records. The data was sold and traded before 000webhost was alerted in October. The breach included names, email addresses and plain text passwords.",
  "LogoPath": "https://haveibeenpwned.com/Content/Images/PwnedLogos/000webhost.png",
  "DataClasses": [
    "Email addresses",
    "IP addresses",
    "Names",
    "Passwords"
  ],
  "IsVerified": true,
  "IsFabricated": false,
  "IsSensitive": false,
  "IsRetired": false,
  "IsSpamList": false,
  "IsMalware": false
},
```

pdehaan: Verified w/ the above namespace schema and Joi. Doesn't look like any breaches have any more/fewer properties.

```js
import Joi from "joi";

const { array, boolean, number, string } = Joi.types();
const schema = Joi.object({
  Name: string,
  Title: string,
  Domain: string.allow(''),
  BreachDate: string,
  AddedDate: string,
  ModifiedDate: string,
  PwnCount: number,
  Description: string,
  LogoPath: string,
  DataClasses: array.items(string).min(1),
  IsVerified: boolean,
  IsFabricated: boolean,
  IsSensitive: boolean,
  IsRetired: boolean,
  IsSpamList: boolean,
  IsMalware: boolean,
});

try {
  const breaches = await fetchBreaches();
  // delete breaches[5].DataClasses; // Throws an error due to now missing property.
  const opts = { abortEarly: false, presence: 'required' };
  await array.items(schema).validateAsync(breaches, opts);
} catch (err) {
  console.error(err?.details || err);
}

async function fetchBreaches() {
  const res = await fetch("https://haveibeenpwned.com/api/v3/breaches");
  return res.json();
}
```

src/customTypes.d.ts
```diff
@@ -68,3 +68,22 @@ declare module 'mozlog' {
...
```
Vinnl, line 72: If the types agree with the schema that sounds good, but maybe change it to be an actual type, instead of a namespace:

```ts
type Breach = {
  Name: string;
  Title: string;
  // ...
}
```

Other than that this looks good to go!

Vinnl (approved):

PR #3004, opened 2023-04-20
Mntor 1505/legacy cleanup

toufali:
# References: 
Jira: MNTOR-1505

# Description
Legacy file/folder cleanup.  
Still left to do (so far):
- ~~fix tests~~
- update readme [MNTOR-1565](https://mozilla-hub.atlassian.net/browse/MNTOR-1565)
- unify scripts folder [MNTOR-1561](https://mozilla-hub.atlassian.net/browse/MNTOR-1561)
- clean up locale strings [MNTOR-1562](https://mozilla-hub.atlassian.net/browse/MNTOR-1562)

# How to test
- Full site regression tests
- Email functionality tests

PR #3024, opened 2023-05-05
MNTOR-1591: fix circle ci config

mansaj:
# References: 
Jira: MNTOR-1591

# Description
Tagged deploy isn't working, causing staged deploy from not being able to be kicked off

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -149,7 +149,7 @@ workflows:
             - deploy:
                 filters:
                     tags:
-                        only: /*/
+                        only: /.*/
                     branches:
                         only: main
 
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -147,11 +147,9 @@ workflows:
             - lint-css
             - lint-l10n
             - deploy:
-                requires:
-                    - lint-js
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -147,8 +147,6 @@ workflows:
             - lint-css
             - lint-l10n
             - deploy:
-                requires:
-                    - lint-js
```
Vinnl, line 151: @mansaj Could you clarify why this was necessary? Shouldn't this task have succeeded?

mansaj: @Vinnl Good question, so this took me a while to figure out 

The reason we need to get rid of this requirement is that lint-js actually does NOT run in tagged workflow. In CircleCI, tags only run when they are specified (see deploy step below). Because lint-js wasn't running, the deploy step did not run either, resulting in our circleCI skipping deploy completely, thus skipping our Jenkins pipeline.

Vinnl: Ah gotcha, and so the reason it broke recently is because #3004 introduced 212fe07, which added this dependency. Thanks!

PR #3027, opened 2023-05-08
MNTOR-1594: Add OptinMonster embed code

flozia:
# References: 
Jira: [MNTOR-1594](https://mozilla-hub.atlassian.net/browse/MNTOR-1594)

# Description

This PR adds the OptinMonster embed code. The S&P Marketing team I would like to be able to contact people about Firefox Monitor Premium with their consent so we can grow Monitor Premium.

# Screenshot

mntor-1594-optinmonster-embed

# How to test

There are a couple of floating banners and popups that are being prepared. For testing purposes, there is only one banner enabled for now. Rules that determine for which pages and other display preferences can be adjusted remotely via OptinMonster.

- Set the env variables: `OPTINMONSTER_ACCOUNT` and `OPTINMONSTER_USER`
- Visit `/breaches` or `/user/breaches`
- Wait for 5 seconds
- A OptionMonster banner should show up


src/app.js
```diff
@@ -105,15 +109,33 @@ app.use((_req, res, _next) => {
...
+        "'self'",
+        "'unsafe-inline'",
+        // Support OptinMonster Campaign Embed Code:
+        // https://optinmonster.com/docs/how-to-manually-add-an-after-post-or-inline-optin/#Campaign_Embed_Code
+        'https://*.omappapi.com',
+        'http://fonts.googleapis.com'
+      ],
+      fontSrc: [
+        "'self'",
+        'http://fonts.gstatic.com'
```
flozia, line 138: Potentially blocking: What’s really not optimal is that OptionMonster is loading font resources over `http` instead of `https`.

rhelmer: Hm, yeah this is going to lead to mixed-content warnings I think (if it loads at all) - do any other sites at Mozilla use this services? I'd like to see what kind of CSP they came up with.

rhelmer (changes requested):
Do you have credentials you can share with me privately for testing purposes? I'd like to see if we can't tighten up this CSP. Thanks!

src/app.js
Vinnl: Yeah, I feel this might be a blocker. In addition to the warnings, I'd like to avoid the ability for ISPs, VPNs, open networks, etc. to MITM these, even if they're just fonts. And I'm not a big fan of adding more fonts to the page load size just for this one component (whatever it is - I don't have the context here).

But perhaps more importantly, needing this makes me suspicious of the development practices of OptinMonster: unnecessarily adding more third-party connections, and over an unencrypted connection at that, does not bode well...

flozia: Quick update on the unencrypted connection: After checking locally again it seems like — presumably after disabling the OptinMonster logo on the banner — there is no attempt to load resources from http://fonts.gstatic.com anymore. Even though the banner itself does not have any custom fonts used CSS is still being loaded from https://fonts.googleapis.com/css?family=Open+Sans, though.

src/app.js
```diff
@@ -90,7 +90,11 @@ const imgSrc = [
   'https://*.googletagmanager.com',
   'https://firefoxusercontent.com',
   'https://mozillausercontent.com/',
-  'https://monitor.cdn.mozilla.net/'
+  'https://monitor.cdn.mozilla.net/',
+  // Support OptinMonster Campaign Embed Code:
+  // https://optinmonster.com/docs/how-to-manually-add-an-after-post-or-inline-optin/#Campaign_Embed_Code
+  ' data:',
```
rhelmer, line 96: Hm, unfortunate that we need this but I don't think there's a way to restrict `data:` further unfortunately.

rhelmer (approved):

src/app.js
Vinnl: Oh I'd missed this one, but this feels like a fairly significant loosening of restrictons, opening us up to embedding of images that abuse vulnerabilities in image formats, or e.g. SVGs that include JavaScript?

Not sure how big of a risk that is, assuming that we trust all scripts that we do allow to run on this page :shrug: 

(I think theoretically

PR #3032, opened 2023-05-11
MNTOR-1548 docs - add release process matching Relay

rhelmer:
# References: 
Jira: MNTOR-1548


rhelmer:
@Vinnl @maxxcrawford @mansaj I modeled this closely after the Relay version of the doc - I think it's roughly compatible with our current docs (see the Jira ticket for link), but I wanted to make sure this is what we want to do?

If we want to keep Relay and Monitor release processes more tightly in sync, I'd suggest we have a standard S&P template (or similar) as a next step.

It'd be good to take a look at what other folks are doing at Mozilla, there might be an opportunity to standardize at least across our org.

docs/release_process.md
```diff
@@ -0,0 +1,319 @@
...
+environment. To do this, we first release code to [Dev][dev] and
+[Stage][stage].
+
+## Release to Dev
+Every commit to `main` is automatically deployed to the [Dev][dev] server, as
+long as it can be done with a fast-forward push. Since the
+[Great GitHub Heroku Incident of 2022][github-heroku-incident], this is
+done from CircleCI using a [service account][service-account].
+
+To push a different branch, you need to add the Heroku app as a remote:
```
Vinnl, line 69: Maybe we could also ask to also give a quick heads-up on Slack that you're taking over `dev`? Just to let others know if/when they can take it again.

docs/release_process.md
```diff
@@ -0,0 +1,319 @@
...
+
+## Release to Stage
+Every tag pushed to GitHub is automatically deployed to the [Stage][stage]
+server. The standard practice is to create a tag from `main` every Tuesday at
+the end of the day, and to name the tag with `YYYY-MM-DD` [CalVer][calver]
+syntax. This tag will include only the changes that have been merged to `main`.
+E.g.,
+
+1. `git tag 2022.08.02`
+2. `git push --tags`
```
Vinnl, line 91: I'm not sure that it's better, but just offering up for consideration: I tend to push a specific tag, just in case I have inadvertently created other tags locally. On the other hand, pushing all tags every time might be a better way of finding out that you've done that.

```suggestion
2. `git push origin 2022.08.02`
```

docs/release_process.md
```diff
@@ -0,0 +1,319 @@
...
+```
+
+### Create Release Notes on GitHub
+After you push the tag to GitHub, you should also
+[make a pre-release on GitHub][github-new-release] for the tag.
+
+1. Choose the tag you just pushed (e.g., `2022.08.02`)
+2. Type the same tag name for the releae title (e.g., `2022.08.02`)
+3. Click "Previous tag:" and choose the tag currently on production.
+   * You can find this at [the `__version__` endpoint][prod-version].
```
Vinnl, line 124: We should restore/add this some time!

docs/release_process.md
```diff
@@ -0,0 +1,319 @@
...
+deploy the tag to production:
+
+1. File an [SRE ticket][sre-board] to deploy the tag to [Prod][prod].
+   * Include a link to the GitHub Release
+   * You can assign it directly to our primary SRE for the day
+2. When SRE starts the deploy, "cloudops-jenkins" will send status messages
+   into the #fx-monitor-engineering channel.
+3. When you see `PROMOTE PROD COMPLETE`, do some checks on prod:
+   * Check sentry prod project for a spike in any new issues
+   * Check grafana dashboard for any unexpected spike in ops
```
Vinnl, line 142: Could you add links for these? (I don't have them :/)

docs/release_process.md
```diff
@@ -0,0 +1,319 @@
...
+   * Spot-check the site for basic functionality
+   * Ping SDET to run end-to-end tests on prod
+4. Update the GitHub Release from "pre-release" to a full release and reference the production deploy SRE Jira ticket.
```
Vinnl, line 145: Hmm, who is that for? Since GH Releases are publicly-viewable, whereas I assume the SRE ticket isn't.

Or is it the other way around, link to the release in the SRE ticket?

docs/release_process.md
```diff
@@ -0,0 +1,319 @@
...
+       commit
+       commit
+       checkout main
+       branch stage-fix-2022.08.02
+       commit
+       checkout main
+       merge stage-fix-2022.08.02 tag: "2022.08.02.01"
+```
+
+### "Dirty `main`" flow
```
Vinnl, line 195: Since this flow is fairly similar to the "clean `main`" flow, and will have the exact same end result if `main` is clean, _and_ given that it's easy to miss the requirement that `main` is clean, maybe we should consider making this the only flow? Would also make the docs shorter :)

Vinnl: Oh, I guess the "Future" flow section below references it. That said, IMHO documentation that doesn't describe the _current_ state of things isn't super useful anyway, so I'd propose removing that section as well.

docs/release_process.md
```diff
@@ -0,0 +1,319 @@
...
+[feature flags][feature-flags] docs for more.
+
+When we are confident that `main` can always be released, we may get rid of
+release tags completely, and move to something more like a GitLab Flow where we
+merge from `main` to long-running branches for `dev`, `stage`, `pre-prod`, and
+`prod` environments.
+
+![Future release process with long-running
+branches](release-process-future-long-branches.png "Future release process with
+long-running branches")
```
Vinnl, line 305: This image doesn't exist :)

Vinnl (approved):
Looks good, no blockers from me, just some suggestions to consider. Thanks for this!

docs/release_process.md
```diff
@@ -0,0 +1,319 @@
...
+[dev]: https://fx-breach-alerts.herokuapp.com/
+[readme]: https://github.com/mozilla/blurts-server/blob/main/README.md
+[docs]: https://github.com/mozilla/blurts-server/tree/main/docs
+[github-flow]: https://docs.github.com/en/get-started/quickstart/github-flow
+[github-heroku-incident]: https://blog.heroku.com/april-2022-incident-review
+[service-account]: https://mana.mozilla.org/wiki/display/TS/List+of+Heroku+service+accounts
+[calver]: https://calver.org/
+[sre-board]: https://mozilla-hub.atlassian.net/jira/software/c/projects/SVCSE/boards/316
+[github-new-release]: https://github.com/mozilla/blurts-server/releases/new
+[prod-version]: https://relay.firefox.com/__version__
```
mansaj, line 319: This links to relay's version endpoint, we might consider adding our own

docs/release_process.md
```diff
@@ -0,0 +1,319 @@
...
+       commit
+       checkout main
+```
+
+### Create Release Notes on GitHub
+After you push the tag to GitHub, you should also
+[make a pre-release on GitHub][github-new-release] for the tag.
+
+1. Choose the tag you just pushed (e.g., `2022.08.02`)
+2. Type the same tag name for the releae title (e.g., `2022.08.02`)
```
pdehaan, line 122:
```suggestion
2. Type the same tag name for the release title (e.g., `2022.08.02`)
```

docs/release_process.md
pdehaan: Do we have a Jira ticket for this? I didn't find anything obvious in GitHub or Jira.

rhelmer: This is on the `nextjs` branch now! https://github.com/mozilla/blurts-server/commit/374eab444d9490cc61a443a2c7babea013e6af0e

docs/release_process.md
rhelmer: I guess for us, or anyone that wants to have an authoritative place for folks to check to see what the latest release is, I find it convenient when looking at code and trying to figure out what's shipped and what's not.

docs/release_process.md
rhelmer: I don't want to hold up this PR anymore :) But I agree we should either move to the future state or just document the current state, let's do that in a followup if you don't mind.

docs/release_process.md
Vinnl: Just in case it's useful: if you `git blame` a line and view that commit in GitHub ([example](https://github.com/mozilla/blurts-server/commit/db9b115a82312a0b5a8602a7fc7a3b080cf77802)), then it will list which tags include that commit:

image

I've often used this (combined with the `__version__` endpoint) to figure out where something has or has not shipped yet.

PR #3051, opened 2023-05-17
Wire up Next.js to FxA using iron-session

Vinnl:
# References: 
Jira: MNTOR-1624

This basically copies over the controllers from https://github.com/mozilla/blurts-server/blob/main/src/controllers/auth.js over to Next.js route handlers, replacing references to `req.session` with [iron-session](https://github.com/vvo/iron-session).

Two things to call out about iron-session:
- it stores session data client-side in an encrypted cookie, as opposed to in Redis like we do with express-session
- it has [a rewrite going on](https://github.com/vvo/iron-session/issues/586) with support for Next.js's app directory, which we are using - but I think we can already make that work with the current version (edit: actually, not sure yet how to use this in components in the app dir - but we can always fall back to the old way)

To test, add a random string of length >32 to `.env` named `IRON_SESSION_PASSPHRASE`, then visit `/oauth/init`. You can also add a `console.log(session)` there to inspect the session data. If you then visit that page again after completing the sign-in, you should see that it has a `.user` property with user data from FxA.

### DIFF
--- a/src/app/functions/server/getSession.ts
+++ b/src/app/functions/server/getSession.ts
@@ -1,0 +1,27 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import { getIronSession, IronSession } from "iron-session";
+import { NextResponse } from "next/server";
+import { UserData } from "../../transitionTypes";
+
+export type SessionData = Partial<{
+  oauthState: string;
+  user: UserData;
+}>;
+
+const sessionConfig = {
+  cookieName: "monitor-iron-session",
+  password: process.env.IRON_SESSION_PASSPHRASE!,
+  cookieOptions: {
+    secure: process.env.NODE_ENV !== "development",
+  },
+};
+
+export async function getSession(
+  req: Request,
+  res: NextResponse
+): Promise<IronSession & SessionData> {
+  return getIronSession(req, res, sessionConfig);
+}
--- a/src/app/oauth/confirmed/route.ts
+++ b/src/app/oauth/confirmed/route.ts
@@ -1,0 +1,119 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import { NextResponse } from "next/server";
+import { FxAOAuthClient, getProfileData, getSha1 } from "../../../utils/fxa.js";
+import {
+  getSubscriberByEmail,
+  updateFxAData,
+} from "../../../db/tables/subscribers.js";
+import { addSubscriber } from "../../../db/tables/emailAddresses.js";
+import { getSession } from "../../functions/server/getSession";
+import { getEmailCtaHref, sendEmail } from "../../../utils/email.js";
+import mozlog from "../../../utils/log.js";
+import { UnauthorizedError } from "../../../utils/error.js";
+import { getTemplate } from "../../../views/emails/email2022.js";
+import { signupReportEmailPartial } from "../../../views/emails/emailSignupReport.js";
+import { getL10n, getLocale } from "../../functions/server/l10n";
+import {
+  getBreachIcons,
+  getBreaches,
+} from "../../functions/server/getBreaches";
+import { getBreachesForEmail } from "../../../utils/hibp.js";
+
+const log = mozlog("controllers.auth");
+
+export async function GET(request: Request) {
+  const returnURL = new URL("user/breaches", process.env.SERVER_URL);
+  const response = new NextResponse("", {
+    status: 302,
+    headers: { Location: returnURL.pathname + returnURL.search },
+  });
+  const session = await getSession(request, response);
+  const oauthState = session.oauthState;
+  const l10n = getL10n();
+  if (!oauthState) {
+    log.error("oauth-invalid-session", "req.session.state missing");
+    throw new UnauthorizedError(l10n.getString("oauth-invalid-session"));
+  }
+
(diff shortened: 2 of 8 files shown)


src/app/functions/server/getSession.ts
```diff
@@ -0,0 +1,27 @@
...
+import { UserData } from "../../transitionTypes";
+
+export type SessionData = Partial<{
+  oauthState: string;
+  user: UserData;
+}>;
+
+const sessionConfig = {
+  cookieName: "monitor-iron-session",
+  password: process.env.IRON_SESSION_PASSPHRASE!,
```
Vinnl, line 16: AIUI this is used to encrypt the client-side cookie. I wonder if we could re-use the existing env var `COOKIE_SECRET` for this?

mansaj: I was thinking about the same thing.. Yes I don't see a reason why not

rhelmer:
This seems OK but what's the advantage of using this vs. some more standard approach like JWT? This section addresses it but doesn't really answer the question IMHO :) https://github.com/vvo/iron-session#how-is-this-different-from-jwt:

rhelmer (changes requested):
This seems OK but what's the advantage of using this vs. some more standard approach like JWT? This section addresses it but doesn't really answer the question IMHO :) https://github.com/vvo/iron-session#how-is-this-different-from-jwt:

> How is this different from [JWT](https://jwt.io/)?
> Not so much:
    JWT is a standard, it stores metadata in the JWT token themselves to ensure communication between different systems is flawless.
    JWT tokens are not encrypted, the payload is visible by customers if they manage to inspect the seal. You would have to use [JWE](https://tools.ietf.org/html/rfc7516) to achieve the same.
    @hapi/iron mechanism is not a standard, it's a way to sign and encrypt data into seals

> Depending on your own needs and preferences, iron-session may or may not fit you.

I'm a bit worried about doing something different than the standard if we're not getting specific advantages... in this case I'd be happy to use JWT, I think it's a better approach for scaling our services, but it does have the same trade-off that it's more difficult to revoke vs. using a session cookie that's completely managed by the backend like Monitor does today.

rhelmer:
> This seems OK but what's the advantage of using this vs. some more standard approach like JWT? This section addresses it but doesn't really answer the question IMHO :) https://github.com/vvo/iron-session#how-is-this-different-from-jwt:
> 
> > How is this different from [JWT](https://jwt.io/)?
> > Not so much:
> > JWT is a standard, it stores metadata in the JWT token themselves to ensure communication between different systems is flawless.
> > JWT tokens are not encrypted, the payload is visible by customers if they manage to inspect the seal. You would have to use [JWE](https://tools.ietf.org/html/rfc7516) to achieve the same.
> > @hapi/iron mechanism is not a standard, it's a way to sign and encrypt data into seals
> 
> > Depending on your own needs and preferences, iron-session may or may not fit you.
> 
> I'm a bit worried about doing something different than the standard if we're not getting specific advantages... in this case I'd be happy to use JWT, I think it's a better approach for scaling our services, but it does have the same trade-off that it's more difficult to revoke vs. using a session cookie that's completely managed by the backend like Monitor does today.

I'm also not sure why the JWT payload not being encrypted is a problem, in general the JWT is signed by the auth server (which can be checked) and it can make claims (e.g. user is an admin, or has a particular ID, etc). The signature is checked by the service to ensure that it is valid, but I don't know that the payload itself needs to be encrypted (it can't be faked).

src/app/oauth/confirmed/route.ts
```diff
@@ -0,0 +1,119 @@
...
+import { getTemplate } from "../../../views/emails/email2022.js";
+import { signupReportEmailPartial } from "../../../views/emails/emailSignupReport.js";
+import { getL10n, getLocale } from "../../functions/server/l10n";
+import {
+  getBreachIcons,
+  getBreaches,
+} from "../../functions/server/getBreaches";
+import { getBreachesForEmail } from "../../../utils/hibp.js";
+
+const log = mozlog("controllers.auth");
```
pdehaan, line 25: Q: should this be `"controllers.auth"` or something more specific? Looks like we're using that string 3&times; in 3 different files:

```diff
 git grep -n " mozlog(" | grep -v "://"

 src/app/functions/server/getBreaches.ts:14:const log = mozlog('hibp')
+src/app/oauth/confirmed/route.ts:25:const log = mozlog("controllers.auth");
+src/app/oauth/logout/route.ts:10:const log = mozlog("controllers.auth");
+src/controllers/auth.js:30:const log = mozlog('controllers.auth')
 src/controllers/hibp.js:27:const log = mozlog('controllers.hibp')
 src/db/tables/breaches.js:9:const log = mozlog('DB.breaches')
 src/db/tables/emailAddresses.js:14:const log = mozlog('DB.email_addresses')
 src/db/tables/subscribers.js:12:const log = mozlog('DB.subscribers')
 src/external/onerep.js:9:const log = mozlog('external.onerep')
 src/middleware/error.js:9:const log = mozlog('middleware')
 src/scripts/migrationCleanup.js:16:const log = mozlog('script.migrationCleanup')
 src/utils/email.js:14:const log = mozlog('email-utils')
 src/utils/hibp.js:24:const log = mozlog('hibp')
```

Also looks like we're using "hibp" in 2 places.

Vinnl: Yeah, that's because they're currently all in the same file and therefore have that same label, and I didn't want to introduce discrepancies in our logs.

That said, logging is disabled at the moment since mozlog (and in particular, its main dependency that hasn't been updated in six years) appears to not work well with modern module systems (first guess). So we'll have to either fix that or at least reproduce its logging format (see https://mozilla-hub.atlassian.net/browse/MNTOR-1643).

mansaj:
@Vinnl @rhelmer If we are set on not using iron-session, shall we close this? Sorry @Vinnl

Vinnl:
Superceded by #3060.

PR #3053, opened 2023-05-17
Merge `localization` into v15.9.3

maxxcrawford:
This release merges all string translations from the `localization` branch into a copy of [v15.9.3](https://github.com/mozilla/blurts-server/releases/tag/v15.9.3)


Vinnl (approved):
Generally we even just merge to and from the `localization` branch without reviews, so feel free to merge :+1:

PR #3063, opened 2023-05-22
MNTOR-1748: fix dup primary email edge case

mansaj:
# References: 
Jira: MNTOR-1748
Jira: MNTOR-1749

# Description
Fix two QA bugs found during testing for FxA events. Both cases occur when there are multiple accounts in:
* FxA 
* Monitor

# Screenshot (if applicable)

Not applicable.

# How to test
See separate tickets for test flow


src/db/tables/subscribers.js
```diff
@@ -254,11 +277,22 @@ async function deleteUnverifiedSubscribers () {
...
+ * @param {object} sub subscriber object
  */
-async function deleteSubscriberByFxAUID (fxaUID) {
-  const subscriber = await knex('subscribers').returning('id').where('fxa_uid', fxaUID).del()
-  if (subscriber && subscriber[0]) { await knex('email_addresses').where({ subscriber_id: subscriber[0].id }).del() }
+async function deleteSubscriber (sub) {
+  const trx = await knex.transaction()
+  log.debug('deleteSubscriber', JSON.stringify(sub))
+
+  try {
```
mansaj, line 286: Fixing the delete here:
* When we have a secondary that's linked to the primary in `email_addresses` table, we will get a foreign key constraint if we don't delete it first.

src/db/tables/subscribers.js
```diff
@@ -64,13 +64,36 @@ async function getSubscriberByEmail (email) {
...
+    // update subscriber primary email to updated email
+    subscriberTableUpdated = await knex('subscribers')
+      .where('id', '=', subscriber.id)
+      .update({
+        primary_email: updatedEmail
+      })
+      .returning('*')
+      .transacting(trx)
+
+    // if email_addresses table has updatedEmail as a secondary in Monitor
```
mansaj, line 79: Adding fix for an edge case here:
* When we have two emails, A & B. If we swap them in FxA, we need to swap them in Monitor as well. Prior to this, we'd get a dup because only the `subscribers` table email was changed

src/controllers/fxaRpEvents.js
```diff
@@ -96,21 +96,21 @@ const fxaRpEvents = async (req, res) => {
...
-    res.status(202)
+    res.status(400).send('Bad Request')
   }
 
   if (!decodedJWT?.events) {
     // capture an exception in Sentry only. Throwing error will trigger FXA retry
     log.error('fxaRpEvents', decodedJWT)
     captureMessage('fxaRpEvents: decodedJWT is missing attribute "events"', decodedJWT)
-    res.status(202)
+    res.status(400).send('Bad Request')
```
rhelmer, line 106: Hm, would it be better to throw `UserInputError`?

rhelmer (approved):
SQL changes lgtm, wondering if we should throw the exception that will cause the error handler to run and return `400` vs. doing it this way (just to get the consistent error message, and I think it's the norm elsewhere) but it's fine as-is too.

src/controllers/fxaRpEvents.js
mansaj: Fixed. thanks!

PR #3065, opened 2023-05-23
Session data

Vinnl:
# References: 
Jira: MNTOR-1624

# Description

This adds a `/test/` page that is clearly not finished yet; however, it does take an initial stab at an initial scaffolding for the proper React pages, and more importantly, it shows how we can access user account data in the pages. It also adds a `pendingTranslations.ftl` file, like on Relay.

Also note that this hardcodes a couple of numeric values - I'll bring that up at today's frontend pairing.

# How to test

Visit `/test/` and verify that there's a button to sign in. After signing in, it shows your email address.


src/next-auth.d.ts
```diff
@@ -0,0 +1,47 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import NextAuth, { DefaultSession } from "next-auth";
+import { JWT } from "next-auth/jwt";
+
+declare module "next-auth" {
+  /** The OAuth profile extracted from Firefox Accounts */
```
flozia, line 9: Praise for adding not only the types but also comprehensive comments.

src/app/components/server/Icons.tsx
```diff
@@ -0,0 +1,54 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import { SVGProps } from "react";
+import styles from "./Icons.module.scss";
+
+// Note: ideally, the Nebula icons are made available in a repository somewhere,
+//       then added to react-icons: https://react-icons.github.io/react-icons/.
+//       These manually-created components are a workaround until that is done.
```
flozia, line 10: +1

src/app/(proper_react)/components/ControlsWrapper.module.scss
```diff
@@ -0,0 +1,142 @@
...
+  height: 100%;
+}
+
+.mainMenu {
+  flex: 0 0 250px;
+  box-shadow: 0px 4px 12px 0px $color-grey-10;
+  display: flex;
+  flex-direction: column;
+  gap: 16px;
+  padding: 1px;
```
flozia, line 14: Nit: The padding prevents the sidebar from collapsing entirely and has still a width of `2px` when closed.

flozia (approved):
This is working well for me and is looking good so far.

src/app/(proper_react)/components/ControlsWrapper.module.scss
Vinnl: Oh, good point. I just encountered a different Figma where the sidebar isn't even collapsible in the first place, so I'd propose not spending too much effort refining the styling until we have actual designs for it.

pdehaan:
> This adds a `/test/` page

Is this just a temp page that we plan on deleting before launch? Or do we have some code to prevent this page from appearing in production env? (assuming we don;t want this page deployed to production and it's only for local/dev/stage testing)

Vinnl:
> Is this just a temp page that we plan on deleting before launch? Or do we have some code to prevent this page from appearing in production env? (assuming we don;t want this page deployed to production and it's only for local/dev/stage testing)

We'll probably delete it before launch, but even if we don't, since there are no links pointing there and there are no obscenities on the page, even if we overlook it I don't think there's much risk.

PR #3066, opened 2023-05-23
Make mozlog work with Next.js

Vinnl:
# References: 
Jira: MNTOR-1643
Figma: N/A

# Description

Unfortunately, this required patching the `intel` package in a `postinstall` script. That said, since that package hasn't been updated in six years, this should be relatively safe.

The problem is that `intel` was trying to dynamically determine which modules to load based on which files were present in its directory. However, since Next.js moves (and presumably bundles) Node modules into the `.next` folder, it was unable to find the modules that `mozlog` was expecting to use.

The patch fixes this by simply explicitly importing those four modules.

# How to test

You should not encounter mozlog-related errors anymore when running the application.

### DIFF
--- a/src/utils/log.js
+++ b/src/utils/log.js
@@ -2,17 +2,14 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
-// import mozlog from 'mozlog'
+import mozlog from 'mozlog'
 
 import AppConstants from '../appConstants.js'
 
-// mozlog temporarily disabled because its module system appears to not work well
-// with Next.js (uncomment this and comment out line 16 to see the errors)
-// const log = mozlog({
-//   app: 'fx-monitor',
-//   level: AppConstants.MOZLOG_LEVEL,
-//   fmt: AppConstants.MOZLOG_FMT
-// })
-const log = () => console
+const log = mozlog({
+  app: 'fx-monitor',
+  level: AppConstants.MOZLOG_LEVEL,
+  fmt: AppConstants.MOZLOG_FMT
+})
 
 export default log
--- a/package-lock.json
+++ b/package-lock.json
@@ -7,6 +7,7 @@
     "": {
       "name": "monitor",
       "version": "1.0.0",
+      "hasInstallScript": true,
       "license": "MPL-2.0",
       "dependencies": {
         "@fluent/bundle": "^0.17.1",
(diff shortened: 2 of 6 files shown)


Vinnl:
Some more details about the problem, for those interested. Feel free to skip if not.

Let’s look at the error message we used to get:

```
- error Error [TypeError]: Cannot read properties of undefined (reading 'prototype')
    at eval (webpack-internal:///(sc_server)/./node_modules/mozlog/lib/handler.js:7:26
```

Looking at [that line](https://github.com/mozilla/mozlog/blob/1b9cf8dd3a2a681c0d4aefbe76053c43ed52760e/lib/handler.js#L8-L9), it looks like `handlers.Stream` isn’t defined in the `intel` package. Why is that?

Well, `intel` is trying to automatically detect the handlers it ships (all four of them) instead of enumerating them, and [does so by](https://github.com/seanmonstar/intel/blob/65e77bc379924fef0dda1f75adc611325028381d/lib/handlers/index.js#L7) reading the contents of `__dirname`.

However, since the code gets copied over (and presumably bundled) into `.next`, the `__dirname` [is modified](https://github.com/vercel/next.js/discussions/14341#discussioncomment-28242) and ends up being the path to the page that gets loaded - which contains no handlers, thus resulting in `undefined`.

src/utils/log.js
```diff
@@ -2,17 +2,14 @@
...
-// })
-const log = () => console
+const log = mozlog({
+  app: 'fx-monitor',
+  level: AppConstants.MOZLOG_LEVEL,
+  fmt: AppConstants.MOZLOG_FMT
+})
 
-export default log
+export default log
```
pdehaan, line 15: nit: looks like a missing newline at the end of the file. 

https://eslint.org/docs/latest/rules/eol-last

rhelmer (approved):
Clever workaround, could you also file a bug on mozlog to remove or fix this as well? I'd prefer not to have to keep this around forever, seems brittle.

mansaj:
> Clever workaround, could you also file a bug on mozlog to remove or fix this as well? I'd prefer not to have to keep this around forever, seems brittle.

Second that... There's always the possibility of us moving to only `console.`.. @toufali wanted to do that since day 1 since mozlog doesn't seem to be maintained anymore

pdehaan:
Re: https://github.com/mozilla/mozlog,
Let me know after you file a mozlog ticket and I might talk to the fine folks of FxA and see if they have any interest in adopting that repo (looks like a bunch of ex-FxA/Mozilla employees up in that contributors list). Or we can work w/ the -moz-GitHub admins if one of y'all want to try taking over the repo and maintaining it.

src/utils/log.js
Vinnl: Fixed while rebasing.

Vinnl:
What do you know, PR approved and merged in no time - maybe we'll actually see a new release with it, and be able to revert this PR :)

PR #3074, opened 2023-05-24
Prevent merging unresolved conflicts

Vinnl:
This adds a CI job that checks that there are no conflict markers checked in. Direct reason: I fixed some conflicts in https://github.com/mozilla/blurts-server/pull/3073, but [missed one](https://github.com/mozilla/blurts-server/commit/f307c8752e02ae41efa83626467ce0984659dad6).

### DIFF
--- a/.github/workflows/conflicts.yml
+++ b/.github/workflows/conflicts.yml
@@ -1,0 +1,17 @@
+name: No unresolved conflicts
+on:
+  push:
+    branches: [ main ]
+  pull_request:
+    branches: [ main, localization ]
+jobs:
+  detect-unresolved-conflicts:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - name: List files with merge conflict markers
+        run: git --no-pager grep "<<<<<<<" ":(exclude).github/" || true
+      - name: Fail or succeed job if any files with merge conflict markers have been checked in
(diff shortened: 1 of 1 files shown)


.github/workflows/conflicts.yml
```diff
@@ -0,0 +1,17 @@
...
+    branches: [ main ]
+  pull_request:
+    branches: [ main, localization ]
+jobs:
+  detect-unresolved-conflicts:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - name: List files with merge conflict markers
+        run: git --no-pager grep "<<<<<<<" ":(exclude).github/" || true
```
maxxcrawford, line 13: Praise: I love the simple logic here!

maxxcrawford (approved):
I see no _conflicts_ here. 😹

mansaj (approved):
Love this improvement.. do not love git pulling unresolved conflicts 😆

.github/workflows/conflicts.yml
Vinnl: I was almost to make it simpler with the `--quiet` parameter to `git grep`, but unfortunately that fails if it finds _no_ matches. I'm sure there's a clever Bash trick to invert the exist code, but :shrug:

PR #3075, opened 2023-05-24
Add sentry to nextJS branch

rhelmer:
# Description

Adds Sentry to NextJS branch, for both client and server.

See https://docs.sentry.io/platforms/javascript/guides/nextjs/ for more information.

# How to test

Performance results are sent automatically (Web Vitals for example). To test error handling, an unhandled exception can be thrown, or see https://docs.sentry.io/platforms/javascript/guides/nextjs/usage/ for explicitly sending an exception.


sentry.edge.config.ts
```diff
@@ -0,0 +1,20 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
```
rhelmer, line 1: I'm not sure if we're planning on using Vercel Edge Runtime, if so this might be useful.

sentry.client.config.ts
```diff
@@ -0,0 +1,34 @@
...
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+// This file configures the initialization of Sentry on the client.
+// The config you add here will be used whenever a users loads a page in their browser.
+// https://docs.sentry.io/platforms/javascript/guides/nextjs/
+
+import * as Sentry from "@sentry/nextjs";
+
+Sentry.init({
+  dsn: "",
```
rhelmer, line 12: Note that I didn't include the DSN, I'm pretty sure it's safe to just land (e.g. if clients are sending errors to the same env), but we could put it in an env var and we probably want to tunnel through our own server-side route anyway, so we don't need to hardcode it here in any case. For the moment you do need to fill it in to test though.

sentry.edge.config.ts
rhelmer: [Preview Deployments](https://vercel.com/docs/concepts/deployments/preview-deployments) the Vercel thing I thought we might find useful, but I don't know enough about it to know if this is needed, happy to remove it for now if we're not sure.

sentry.client.config.ts
Vinnl: Looking at [our current config](https://github.com/mozilla/blurts-server/blob/326a24c225ac94fd419d085fa86cb0acd7cd3a36/src/app.js#L30-L44), we already have an env var `SENTRY_DSN` that we could use?

(Although if this script gets run on the client-side, it might [need to be prefixed `NEXT_PUBLIC`](https://nextjs.org/docs/app/building-your-application/configuring/environment-variables#exposing-environment-variables-to-the-browser).)

sentry.client.config.ts
```diff
@@ -0,0 +1,34 @@
...
+
+  // Adjust this value in production, or use tracesSampler for greater control
+  tracesSampleRate: 1,
+
+  // Setting this option to true will print useful information to the console while you're setting up Sentry.
+  debug: true,
```
Vinnl, line 18: Perhaps this should be something like this? (Same for the other config files.)

```suggestion
  debug: process.env.NODE_ENV === "development",
```

Though maybe it'll result in too much noise in our console, in which case maybe it should just be false?

sentry.client.config.ts
```diff
@@ -0,0 +1,34 @@
...
+  // in development and sample at a lower rate in production
+  replaysSessionSampleRate: 0.1,
+
+  // You can remove this option if you're not planning to use the Sentry Session Replay feature:
+  integrations: [
+    new Sentry.Replay({
+      // Additional Replay configuration goes in here, for example:
+      maskAllText: true,
+      blockAllMedia: true,
+    }),
```
Vinnl, line 32: Hmm, this would definitely be very useful, and [the privacy measures](https://docs.sentry.io/platforms/javascript/guides/nextjs/session-replay/privacy/) look pretty good, but I think I'd prefer to be one the safe side and not enable it at first, so we can get some more input on whether we actually should enable it.

sentry.edge.config.ts
Vinnl: I don't think we'll use it since we already have a Mozilla Netlify account, and we'd need to hook it up to our own database anyway, so I'd lean towards removing it. But having it doesn't do much harm either, so I don't mind much either way :)

sentry.server.config.ts
```diff
@@ -0,0 +1,19 @@
...
+import * as Sentry from "@sentry/nextjs";
+
+Sentry.init({
+  dsn: "",
+
+  // Adjust this value in production, or use tracesSampler for greater control
+  tracesSampleRate: 1,
+
+  // Setting this option to true will print useful information to the console while you're setting up Sentry.
+  debug: false,
```
Vinnl, line 18: This might or might not be useful here too?

```suggestion
  debug: process.env.NODE_ENV === "development",
```

Vinnl (approved):
Overall looking good, though I haven't been able to try it myself yet since I don't have the required Sentry config data - I'll reach out on Slack about that.

sentry.edge.config.ts
pdehaan: Q: has Vercel been vetted and approved by IT/security or whomever does those things for external things?

Vinnl: Probably not, so _if_ we were to use it (which I don't expect we will), we'll have to look into that.

sentry.client.config.ts
```diff
@@ -0,0 +1,34 @@
...
+  tracesSampleRate: 1,
+
+  // Setting this option to true will print useful information to the console while you're setting up Sentry.
+  debug: true,
+
+  replaysOnErrorSampleRate: 1.0,
+
+  // This sets the sample rate to be 10%. You may want this to be 100% while
+  // in development and sample at a lower rate in production
+  replaysSessionSampleRate: 0.1,
```
pdehaan, line 24: re: development-vs-production, should these be set in ENV vars instead of hardcoded so people don't accidentally commit development debug settings into **main** and we end up collecting 100% production data?

sentry.client.config.ts
pdehaan: Q: Is it "development" or "dev"?

```sh
git --no-pager grep -n "\.NODE_ENV "

esbuild.js:31:  minify: AppConstants.NODE_ENV !== 'dev',
esbuild.js:32:  sourcemap: AppConstants.NODE_ENV !== 'dev',
next.config.js:19:              `script-src 'self' ${process.env.NODE_ENV === 'development' ? "'unsafe-eval' 'unsafe-inline'" : ''} https://*.googletagmanager.com`,
next.config.js:21:              `connect-src 'self' ${process.env.NODE_ENV === 'development' ? 'webpack://*' : ''} https://*.google-analytics.com https://*.analytics.google.com https://*.googletagmanager.com`,
next.config.js:24:              `style-src 'self' ${process.env.NODE_ENV === 'development' ? "'unsafe-inline'" : ''}`,
scripts/send-email-to-unresolved-breach-subscribers.js:125:  if (AppConstants.NODE_ENV !== 'tests') {
src/app.js:27:const isDev = AppConstants.NODE_ENV === 'dev'
src/appConstants.js:60:if (!process.env.SERVER_URL && process.env.NODE_ENV === 'heroku') {
src/appConstants.js:77:export default AppConstants.NODE_ENV === 'test'
src/client/js/analytics.js:7:const debugMode = AppConstants.NODE_ENV !== 'production'
src/middleware/error.js:41:    message: process.env.NODE_ENV !== 'production' ? errMsg : 'Something went wrong', // hide error message when in production
src/middleware/error.js:42:    stack: process.env.NODE_ENV === 'dev' ? err.stack : {} // hide stack when not in dev
src/routes/api/v1/user.js:26:if (AppConstants.NODE_ENV === 'dev') {
src/routes/index.js:45:if (AppConstants.NODE_ENV !== 'production') {
src/routes/user.js:35:if (AppConstants.NODE_ENV === 'dev') {
```

```sh
git --no-pager grep -n "'development'"

next.config.js:19:              `script-src 'self' ${process.env.NODE_ENV === 'development' ? "'unsafe-eval' 'unsafe-inline'" : ''} https://*.googletagmanager.com`,
next.config.js:21:              `connect-src 'self' ${process.env.NODE_ENV === 'development' ? 'webpack://*' : ''} https://*.google-analytics.com https://*.analytics.google.com https://*.googletagmanager.com`,
next.config.js:24:              `style-src 'self' ${process.env.NODE_ENV === 'development' ? "'unsafe-inline'" : ''}`,
next.config.js:80:    const noindexEnvs = ['dev', 'development', 'heroku', 'stage']
```

sentry.edge.config.ts
rhelmer: OK cool let's just remove it for now to avoid confusion then, thanks! It's easy to add back if we need it.

sentry.client.config.ts
rhelmer: Yes I noticed this too, the old app called it `dev` and the nextjs calls it `development`.

sentry.client.config.ts
rhelmer: We need to be careful here, we don't want to accidentally collect PII. Why don't we gate this on `development` for now? Then we can't accidentally collect PII from production where it matters most.

rhelmer: Oh right this is client-side, hm.

sentry.client.config.ts
rhelmer: Oh @Vinnl just commented earlier, TIL we can do it this way: https://nextjs.org/docs/app/building-your-application/configuring/environment-variables#exposing-environment-variables-to-the-browser

sentry.client.config.ts
rhelmer: OK using `NEXT_PUBLIC` variables seems to work fine here, thanks!

rhelmer:
@Vinnl I've sent you a link on where to find the DSN in the Sentry web app. As you point out it's not secret (DSN must be exposed to the client) but we might as well treat it like we do any other config.

sentry.client.config.ts
Vinnl: Yes, dev only sounds good! And yeah, `NODE_ENV` is available everywhere just like `NEXT_PUBLIC_*` variables, IIRC.

PR #3076, opened 2023-05-24
MNTOR-1659 - instrument breach card open/close for GA

rhelmer:
# References: 
Jira: MNTOR-1659

# Description

Instrument the user opening/closing the breach details card for GA.

# How to test

1. open and close the breach details card
2. look for a ping to GA with a payload containing the following:
```
en: toggled_breach_card
ep.debug_mode: true
ep.action: opened
ep.company_name: Gravatar
```

You can also confirm using the "DebugView" feature in the GA dashboard that pings are coming in and contain the correct data.

### DIFF
--- a/src/client/js/partials/breaches.js
+++ b/src/client/js/partials/breaches.js
@@ -43,6 +43,8 @@ function init () {
   statusFilter.addEventListener('change', handleEvent)
   breachesTable.addEventListener('change', handleEvent)
   document.body.addEventListener('email-added', handleEvent)
+
+  breachRows.forEach(elem => elem.addEventListener('toggle', handleEvent))
 }
 
 function handleEvent (e) {
@@ -67,6 +69,13 @@ function handleEvent (e) {
     case e.type === 'email-added':
       state.emailCount = e.detail.newEmailCount
       renderZeroState()
+      break
+    case e.type === 'toggle':
+      window.gtag('event', 'toggled_breach_card', {
+        action: e.target.open ? 'opened' : 'closed',
+        page_location: (window.location.origin + window.location.pathname),
+        company_name: e.target.dataset.companyName
+      })
       break
   }
 }
--- a/src/views/partials/breaches.js
+++ b/src/views/partials/breaches.js
@@ -36,7 +36,7 @@ function createBreachRows (data, logos) {
       const logo = getBreachLogo(breach, logos)
 
       return `
-      <details class='breach-row' data-status=${status} data-email=${account.email} data-classes='${dataClassesTranslated}' ${isHidden ? 'hidden' : ''}>
+      <details class='breach-row' data-company-name=${breach.Name} data-status=${status} data-email=${account.email} data-classes='${dataClassesTranslated}' ${isHidden ? 'hidden' : ''}>
         <summary>
           <span class='breach-company'>${logo} ${breach.Title}</span>
           <span>${shortList.format(dataClassesTranslated)}</span>


src/client/js/partials/breaches.js
```diff
@@ -68,6 +70,13 @@ function handleEvent (e) {
       state.emailCount = e.detail.newEmailCount
       renderZeroState()
       break
+    case e.type === 'toggle':
```
rhelmer, line 73: The way the switch statement here is doing `switch (true)` and then matching on different kinds of things (some use the element name, some filter on the type) makes me a bit uncomfortable, I feel like it could lead to confusing bug. But, since we're not keeping this long-term, I felt like staying consistent with having a single handler here was more important.

src/client/js/partials/breaches.js
```diff
@@ -43,6 +43,8 @@ function init () {
   statusFilter.addEventListener('change', handleEvent)
   breachesTable.addEventListener('change', handleEvent)
   document.body.addEventListener('email-added', handleEvent)
+
+  breachRows.forEach(elem => elem.addEventListener('toggle', handleEvent))
```
rhelmer, line 47: @maxxcrawford would it be better to select an element higher in the DOM and add one listener there, vs. doing it this way?

Vinnl: Seems fine to me, I don't expect major performance issues from just this - otherwise we can adjust.

Vinnl (approved):
Verified to work locally. And just in time before we start porting this page to Next.js (otherwise we'd have to make the same change in the `nextjs` branch on the new files).

src/client/js/partials/breaches.js
```diff
@@ -68,6 +70,13 @@ function handleEvent (e) {
       state.emailCount = e.detail.newEmailCount
       renderZeroState()
       break
+    case e.type === 'toggle':
+      window.gtag('event', 'toggled_breach_card', {
+        action: e.target.open ? 'opened' : 'closed',
+        page_location: location.href,
```
pdehaan, line 76: Q: _If_ the `location.href` includes query string data, is there any risk that this would be logging any unwanted PII that we may want to sanitize?

src/views/partials/breaches.js
```diff
@@ -36,7 +36,7 @@ function createBreachRows (data, logos) {
       const logo = getBreachLogo(breach, logos)
 
       return `
-      <details class='breach-row' data-status=${status} data-email=${account.email} data-classes='${dataClassesTranslated}' ${isHidden ? 'hidden' : ''}>
+      <details class='breach-row' data-company-name=${breach.Title} data-status=${status} data-email=${account.email} data-classes='${dataClassesTranslated}' ${isHidden ? 'hidden' : ''}>
```
pdehaan, line 39: I'm a bit unclear on `data-company-name` vs `breach.Title` here.
Since this isn't really user-facing, do we want to use `breach.Title` (with potential spaces+punctuation) versus `breach.Name`?

`breach.Title` should always be unique, but not always "English" (if that matters in the pipeline):

```sh
…
{ Name: 'Zhenai', Title: 'Zhenai.com' }
{ Name: 'Zoosk', Title: 'Zoosk (2011)' }
{ Name: 'Zoosk2020', Title: 'Zoosk (2020)' }
{ Name: 'Parapa', Title: 'Пара Па' }
{ Name: 'SprashivaiRu', Title: 'Спрашивай.ру' }
{ Name: 'DecoratingTheHouse', Title: '집꾸미기' }
```

src/client/js/partials/breaches.js
rhelmer: This is a good point, it came up in the context of the free scan. We determined at the time that this was the only page that did this and avoided it, but I'll confirm these details. Thanks!

src/views/partials/breaches.js
rhelmer: Hm, I think you're right, but I'm a little on the fence since I think it'll be harder for the folks will need to figure out the mapping between the two, it won't match what's in the UI. I guess looking at the HTML will be a reliable way to do that on a one-off basis but that's not a great experience.

rhelmer: I'm imagining product folks and data science folks being the primary consumer of the analytics and reports, to be specific. However I think the formatting consistency of the `Name` is a really good point.

src/client/js/partials/breaches.js
pdehaan: Yeah, not sure if passing it through `new URL(location.href).pathname` sanitizes it enough/too much, but 🤷 &hellip;

```js
location.href
// 'https://github.com/mozilla/blurts-server/pull/3076#discussion_r1205879038?ok=computer'

new URL(location.href).pathname
// '/mozilla/blurts-server/pull/3076'
```

src/views/partials/breaches.js
pdehaan: For the record, I don't know anything about anything&hellip; but&hellip; here's a free theory for you (since I saw the Zoosk and Zoosk2020 above). 

Back in 2011 Zoosk was breached. It wasn't reported/added to HIBP until 2017. Suspiciously, according to HIBP, that record was modified sometime in mid-2020 (3.5 years after it was originally added to the site?!).

<details>
<summary><a href="https://haveibeenpwned.com/api/v3/breach/zoosk">haveibeenpwned.com/api/v3/breach/zoosk</a></summary>

```js
Name: "Zoosk",
Title: "Zoosk (2011)",
Domain: "zoosk.com",
BreachDate: "2011-01-01",
AddedDate: "2017-02-08T07:59:39Z",
ModifiedDate: "2020-08-06T21:54:13Z",
```
</details>

My theory is that zoosk's `Title` was renamed from "Zoosk" to "Zoosk (2011)" when the new "Zoosk (2020)" breach was added to the site to help disambiguate the two breaches.

<details>
<summary><a href="https://haveibeenpwned.com/api/v3/breach/zoosk2020">haveibeenpwned.com/api/v3/breach/zoosk2020</a></summary>

```js
Name: "Zoosk2020",
Title: "Zoosk (2020)",
Domain: "zoosk.com",
BreachDate: "2020-01-12",
AddedDate: "2020-08-07T03:46:38Z",
ModifiedDate: "2020-08-07T03:46:38Z",
```
</details>

Not sure if any of that nonsense means anything, but it _might_ imply that `Name`s are immutable but `Title`s aren't. Unclear on whether that would impact product/data science folks and their queries, and all speculation on my part, but something to maybe consider (or mock).

We might be able to see if that's a valid theory by scraping all breaches and seeing how many map to the same `Domain` and then checking the `ModifiedDate`s to see if the theory is maybe correct.

src/client/js/partials/breaches.js
rhelmer: Hm, so I think we do want the hostname here (traffic comes in to the same property from multiple sites), I might go with this instead: `(window.location.origin + window.location.pathname)`

PR #3081, opened 2023-05-25
MNTOR-1634: Port Settings page to React

maxxcrawford:
WIP

### DIFF
--- a/src/app/(nextjs_migration)/(authenticated)/user/settings/page.tsx
+++ b/src/app/(nextjs_migration)/(authenticated)/user/settings/page.tsx
@@ -3,9 +3,11 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
 import { getL10n } from "../../../../functions/server/l10n";
-import { getServerSession } from "next-auth";
-import { authOptions } from "../../../../api/auth/[...nextauth]/route";
+// import { getServerSession } from "next-auth";
+// import { authOptions } from "../../../../api/auth/[...nextauth]/route";
+import ImageIconDelete from "../../../../../client/images/icon-delete.svg";
 import "../../../../../client/css/partials/settings.css";
+import React from "react";
 
 // TODO: Set EmailObject to match what comes from db/tables/emailAddresses.js
 type EmailObject = {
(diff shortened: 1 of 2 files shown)


src/app/(nextjs_migration)/(authenticated)/user/settings/page.tsx
```diff
@@ -0,0 +1,288 @@
...
+  //   meta: {
+  //     title: getMessage("settings-meta-title"),
+  //   },
+  //   pathname: req._parsedOriginalUrl?.pathname,
+  // };
+
+  // res.send(mainLayout(data));
+
+  const limit = process.env.MAX_NUM_ADDRESSES;
+  const fxaUrl = process.env.FXA_SETTINGS_URL;
```
pdehaan, line 237: nit: Should these be `process.env.*` or `AppConstants.*`? (although possibly more of a global question)

```sh
git --no-pager grep -n "AppConstants\." | wc -l # 80
git --no-pager grep -n "process\.env\." | wc -l # 33
```

Vinnl: We're kinda leaning towards just using `process.env` from now on - we could possibily later add a separate script that checks for required env vars on startup if we really want to.

pdehaan: Related: https://github.com/mozilla/blurts-server/issues/275, and I'd still probably recommend **convict** for managing configs and env vars, etc.

Unrelated, but we should probably also re-triage the 300 open issues and close the majority of them since they're no longer relevant and see which ones we might want to do at some point during the Next migration, or future React migration, or future future TypeScript migration.

maxxcrawford: @pdehaan That's a great call out about looking at current issues on Monitor. I'll put together a issue-triage party for next week.

mansaj:
@maxxcrawford would you want to incorporate my API PR into this PR for integration? https://github.com/mozilla/blurts-server/pull/3095

maxxcrawford:
> @maxxcrawford would you want to incorporate my API PR into this PR for integration? #3095

Yes! My goal right now is to just land the UI with dummied data. @Vinnl has some blocking stuff too.

maxxcrawford:
Closing this branch as it was integrated with #3105

PR #3009, opened 2023-04-25
Add unique page titles

Vinnl:
# References: 
Jira: MNTOR-1486
Figma: N/A

# Description

Adds unique page titles and social media descriptions to every page. I still have an open question with Mela that Florian is CC'd on, and Florian will take on getting this PR over the finish line since I'll be out.

# How to test

Visit the different page, and check that their metadata and tab titles are unique.

# Checklist (Definition of Done)
- [x] Localization strings (if needed) have been added.
- [x] Commits in this PR are minimal and [have descriptive commit messages](https://chris.beams.io/posts/git-commit/).
- [x] I've added or updated the relevant sections in readme and/or code comments
- [ ] I've added a unit test to test for potential regressions of this bug. - N/A
- [ ] Product Owner accepted the User Story (demo of functionality completed) or waived the privilege.
- [ ] All acceptance criteria are met.
- [x] Jira ticket has been updated (if needed) to match changes made during the development process.
- [ ] Jira ticket has been updated (if needed) with suggestions for QA when this PR is deployed to stage.

PR #3082, opened 2023-05-26
MNTOR-1765 - set title, favicon, and meta tag correctly for nextjs app

rhelmer:
# Description

The favicon and meta title/description are the defaults set by create-next-app, these should be the same as Firefox Monitor in production.


src/app/layout.tsx
```diff
@@ -13,8 +13,9 @@ import { authOptions } from "./api/auth/[...nextauth]/route";
 const inter = Inter({ subsets: ["latin"] });
 
 export const metadata = {
-  title: "Create Next App",
-  description: "Generated by create next app",
+  title: "Firefox Monitor",
+  description:
+    "Find out if you’ve been part of a data breach with Firefox Monitor. We’ll help you understand what to do next and continuously monitor for any new breaches.",
```
pdehaan, line 18: Matches the current values on prod, but will need to confirm w/ a deployed Next.js version to see if we're setting all the `twitter:*` and `og:*` social bits as well. Plus I guess overriding these on the breach detail pages, ie: https://monitor.firefox.com/breach-details/linkedin.

```html
<head>
  <title>Firefox Monitor - LinkedIn Data Breach</title>
  <meta charset="utf-8">
  <meta name="viewport" content="width=320, initial-scale=1">
  <meta name="description" content="Use Firefox Monitor to find out if your personal information was exposed in this breach, and understand what to do next.">
  <meta name="twitter:title" content="Were you affected by the LinkedIn Data Breach?">
  …
</head>
```
Although I _think_ we're currently translating these values in the current pre-Next.js site:
https://github.com/mozilla/blurts-server/blob/83a32d11b4a95abbde24088fa9dcdcba2c5fc00a/src/views/guestLayout.js#L12-L15

rhelmer: Good points, thanks... yes once it's deployed to dev/stage this will be easy to check. As an aside, we should really start having unique titles per page, but for now I think it's OK. Also yes good point on localization, I'll add that to this PR.

rhelmer: Here are the [docs](https://nextjs.org/docs/app/api-reference/functions/generate-metadata), I needed to switch this from using `exprt const metadata ...` to `export async function generateMetadata()` so the `l10n` module can get the headers.

I took a look at what the current Monitor is returning for its meta tags and returned the equivalent above ^ I'll also review our current meta tags against the recommended. Having different titles for different pages is something that was suggested for better SEO, but I'm not as sure about the meta tags, I'll find out.

src/app/layout.tsx
```diff
@@ -9,14 +9,31 @@ import { getServerSession } from "next-auth";
...
+export async function generateMetadata() {
+  const l10n = getL10n();
+  return {
+    title: l10n.getString("brand-fx-monitor"),
+    description: l10n.getString("meta-desc-2"),
+    twitter: {
+      card: "summary_large_image",
+      title: l10n.getString("brand-fx-monitor"),
+      description: l10n.getString("meta-desc-2"),
+      images: ["/images/og-image.webp"],
```
pdehaan, line 25: I can confirm these w/ the Twitter+OpenGraph linters once we get this branch on a dev/staging server, but I'm pretty sure these images need to be fully qualified (if they don't magically get converted somewhere in the pipeline already).
Here's the old version (L31 and L37 below):
https://github.com/mozilla/blurts-server/blob/d4cb77862d91a61e0debf1032301428c2019ee04/src/views/guestLayout.js#L31-L37

src/app/layout.tsx
```diff
@@ -9,14 +9,31 @@ import { getServerSession } from "next-auth";
...
-    "Find out if you’ve been part of a data breach with Firefox Monitor. We’ll help you understand what to do next and continuously monitor for any new breaches.",
-};
+export async function generateMetadata() {
+  const l10n = getL10n();
+  return {
+    title: l10n.getString("brand-fx-monitor"),
+    description: l10n.getString("meta-desc-2"),
+    twitter: {
+      card: "summary_large_image",
+      title: l10n.getString("brand-fx-monitor"),
```
pdehaan, line 23: I think before we were overriding the social titles in some cases. Not sure if it matters, but 🤷 .
https://github.com/mozilla/blurts-server/blob/d4cb77862d91a61e0debf1032301428c2019ee04/src/views/guestLayout.js#L13-L16

src/app/layout.tsx
```diff
@@ -9,14 +9,31 @@ import { getServerSession } from "next-auth";
...
+      description: l10n.getString("meta-desc-2"),
+      images: ["/images/og-image.webp"],
+    },
+    openGraph: {
+      title: l10n.getString("brand-fx-monitor"),
+      description: l10n.getString("meta-desc-2"),
+      siteName: l10n.getString("brand-fx-monitor"),
+      type: "website",
+      url: "https://monitor.firefox.com",
```
pdehaan, line 32: Q: Does this need to be dynamic so every shared page doesn't link to ye olde [production] homepage? (although I'll ignorantly assume that dev/stage have some "noindex" robots action happening somewheres to avoid those getting SEO'ed)

src/app/layout.tsx
rhelmer: > Good points, thanks... yes once it's deployed to dev/stage this will be easy to check. As an aside, we should really start having unique titles per page, but for now I think it's OK. Also yes good point on localization, I'll add that to this PR.

Oh, I missed that this was fixed in the old app in https://github.com/mozilla/blurts-server/pull/3009 - I will do the same here as well then.

src/app/layout.tsx
rhelmer: It does indeed seem to be magically fully qualified, I get `http://localhost:6060/...` for the `image` properties.

src/app/layout.tsx
rhelmer: Thanks, I am reading about how to override these per-page and merge: https://nextjs.org/docs/app/building-your-application/optimizing/metadata

src/app/layout.tsx
pdehaan: Ah, here's a grep that I think answered at least parts of my ignorant assumptions:

```sh
curl -v --silent https://stage.firefoxmonitor.nonprod.cloudops.mozgcp.net/ 2>&1 | grep "robots"

< x-robots-tag: noindex
```

So it doesn't look like we're using a `<meta name="robots" content="noindex" />` tag, but we are setting `x-robots-tag: noindex` using headers for staging (but not production).

```sh
curl -v --silent https://monitor.firefox.com/ 2>&1 | grep "robots"
# no output
```

src/app/layout.tsx
rhelmer: > Thanks, I am reading about how to override these per-page and merge: https://nextjs.org/docs/app/building-your-application/optimizing/metadata

OK so per this doc, the problem is that you can override single keys, but it will not attempt to merge e.g. `openGraph` or `twitter`, so if this is set at the top-level:

```
  return {
    title: l10n.getString("brand-fx-monitor"),
    description: l10n.getString("meta-desc-2"),
    twitter: {
      card: "summary_large_image",
      title: l10n.getString("brand-fx-monitor"),
      description: l10n.getString("meta-desc-2"),
      images: ["/images/og-image.webp"],
    },
    openGraph: {
      title: l10n.getString("brand-fx-monitor"),
      description: l10n.getString("meta-desc-2"),
      siteName: l10n.getString("brand-fx-monitor"),
      type: "website",
      url: process.env.SERVER_URL,
      images: ["/images/og-image.webp"],
    },
  };
```

Then when a page wants to just override a small part of this:

```
  return {
    title: l10n.getString("..."),
    twitter: {
      title: l10n.getString("..."),
      description: l10n.getString("..."),
    },
    openGraph: {
      title: l10n.getString("..."),
      description: l10n.getString("..."),
    },
  };
```

This works fine for `title` (it leaves `description` defined), but the entire `twitter` and `openGraph` objects get overwritten.

There is a suggested way to deal with it, which is a bit trickier for us since we need to localize... for the moment I just duplicated the info on overriding since we don't have very many unique pages, but I'd like to figure out a cleaner solution.

This deep merge approach makes sense to me, we should see if it works for us: https://github.com/vercel/next.js/issues/46899#issuecomment-1472048414

rhelmer:
OK I ended up just going all the way to MNTOR-1765 and trying to make this equivalent to what we have on production Monitor now. The only thing I'm not totally happy with is https://github.com/mozilla/blurts-server/pull/3082#discussion_r1214677917 but I'd be happy to do that in a followup (or continue here, if folks have strong feelings).

rhelmer:
Also, sorry about the unrelated formatting in the diff, the default VSCode settings we have checked in do format-on-save and specify prettier. I can do this in a separate PR or commit if it's too hard to read.

src/app/layout.tsx
pdehaan: https://nextjs.org/docs/app/api-reference/file-conventions/metadata/opengraph-image also has a neat bit about custom opengraph and twitter images (although no mentions of them supporting .webp).

> Add an `opengraph-image.(jpg|jpeg|png|gif)` image file to any route segment.
> Add a `twitter-image.(jpg|jpeg|png|gif)` image file to any route segment.

And it will add a bunch of the meta tags automatically:
```html
<meta property="og:image" content="<generated>" />
<meta property="og:image:type" content="<generated>" />
<meta property="og:image:width" content="<generated>" />
<meta property="og:image:height" content="<generated>" />

<meta name="twitter:image" content="<generated>" />
<meta name="twitter:image:type" content="<generated>" />
<meta name="twitter:image:width" content="<generated>" />
<meta name="twitter:image:height" content="<generated>" />
```

Can't see mentions of webp specifically on [Open Graph protocol](https://ogp.me/) specs, and this looks possibly too outdated to be definitive: https://stackoverflow.com/questions/57955612/can-we-use-webp-extension-image-in-ogimage-meta-tag.

---

In fact, I see we're using .webp already in production, so if I go to the Meta Sharing Debugger at https://developers.facebook.com/tools/debug/?q=https%3A%2F%2Fmonitor.firefox.com it seems to throw a vague warning:

> ### ⚠️ Warnings That Should Be Fixed
> **Image Unavailable**
> Provided `og:image`, https://monitor.firefox.com could not be downloaded. This can happen due to several different reasons such as your server using unsupported content-encoding. The crawler accepts deflate and gzip content encodings.

src/app/layout.tsx
pdehaan: https://nextjs.org/docs/app/api-reference/functions/generate-metadata also has some good info on `default`s, `template`s and using `parent` metadata stuff:

```ts
import { Metadata, ResolvingMetadata } from 'next';
 
type Props = { ... };

export async function generateMetadata(
  { params, searchParams }: Props,
  parent: ResolvingMetadata,
): Promise<Metadata> {
  ...
}
```

The previously mentioned https://nextjs.org/docs/app/building-your-application/optimizing/metadata#merging also has a good section on merged metadata and overwriting `openGraph`, `twitter`, and `robots`:

> ### Merging
> Following the [evaluation order](https://nextjs.org/docs/app/building-your-application/optimizing/metadata#ordering), Metadata objects exported from multiple segments in the same route are shallowly merged together to form the final metadata output of a route. Duplicate keys are replaced based on their ordering.
>
> <i>**This means metadata with nested fields such as [`openGraph`](https://nextjs.org/docs/app/api-reference/functions/generate-metadata#opengraph) and [`robots`](https://nextjs.org/docs/app/api-reference/functions/generate-metadata#robots) that are defined in an earlier segment are overwritten by the last segment to define them.**
> &hellip;
> **If you'd like to share some nested fields between segments while overwriting others, you can pull them out into a separate variable: &hellip;**</i>

Unrelated, but interesting read, but here's one of the Next.js source files for metadata stuff: https://github.com/vercel/next.js/blob/canary/packages/next/src/lib/metadata/types/metadata-interface.ts

src/app/layout.tsx
Vinnl: Hmm OK, so looks like Google doesn't localise (i.e. it doesn't send `Accept-Language` headers, so it will always get English).

I can't find anything for Twitter cards, but it looks like we'll need to enumerate supported locales and switch locales according to the `fb_locale` query parameter (so much for OpenGraph being independent) to have Facebook actually use the localised versions: https://stackoverflow.com/a/20827962

So given that we don't currently do that on production already, I wouldn't worry too much about localisation.

src/app/layout.tsx
Vinnl: I see no problem with copy-pasting. One other workaround could be to export an object with the defaults and destructure that elsewhere, but especially with the low number of pages we have, I'd lean towards the indirection being more harmful than the duplication.

Vinnl (approved):
Thanks for taking care of all of them! You'll need to actually place the image in the right folder as well, but otherwise this look good.

PR #3086, opened 2023-05-30
Add the app shell for the React-based website

Vinnl:
# References: 
Jira: MNTOR-1626
Figma: https://www.figma.com/file/CaEKIhvSJqf6KNIMzSkt40/Concepts-for-Monitor-MVP-Redesign?type=design&node-id=565-48454&t=slHSP8Hx6UygIBTO-0

# Description

This adds the scaffolding for the redesigned sidebar / mobile menu implemented in React. This doesn't yet implement the user and app menu, and we'll probably want to iterate on this more once more pieces slot into place, but at least this can give some guidance on where to start when working on new pages. I'm submitting this mostly because I want to pause this work for a bit to help finish the porting process.

# Screenshot (if applicable)

image

image

# How to test

Visit /redesign/user/dashboard. (We can probably figure out something to serve the right thing on the same URL based on a feature flag, but for now, this seems fine - it's relatively trivial to removal the `/redesign/` part later.)


src/app/(proper_react)/MobileShell.tsx
```diff
@@ -0,0 +1,92 @@
...
+            />
+          </Link>
+        </div>
+        <div className={styles.headerEnd}>
+          {/* For the app and user menus */}
+        </div>
+      </header>
+      <div className={styles.nonHeader}>
+        <nav className={styles.mainMenuLayer}>
+          <div className={styles.mainMenu}>
```
pdehaan, line 67: nit: Do we need this `<div>` or can we move the `className={styles.mainMenu}` onto the following child `<ul>` and remove a level of nesting?

src/app/(proper_react)/MobileShell.tsx
```diff
@@ -0,0 +1,92 @@
...
+                  href="https://support.mozilla.org/kb/firefox-monitor-faq"
+                  title={l10n.getString("main-nav-link-faq-tooltip")}
+                >
+                  {l10n.getString("main-nav-link-faq-label")}
+                </a>
+              </li>
+            </ul>
+          </div>
+        </nav>
+        <div className={styles.content}>{props.children}</div>
```
pdehaan, line 88: Not sure I understand the entire DOM layout tree&hellip; but should this be a `<div>` or `<main>`?
Or are `<main>` elements nested elsewhere in child views?

src/app/(proper_react)/PageLink.tsx
```diff
@@ -0,0 +1,23 @@
...
+export type Props = ComponentProps<typeof Link> & {
+  activeClassName?: HTMLAttributes<HTMLAnchorElement>["className"];
+};
+
+export const PageLink = (props: Props) => {
+  const pathName = usePathname();
+  const activeClassSuffix =
+    pathName === props.href ? props.activeClassName : "";
+
+  const className = (props.className ?? "") + " " + activeClassSuffix;
```
pdehaan, line 20: Meh, not convinced this is any better. Feel free to mock and launch this into the sun...
```suggestion
  const className = `${props.className ?? ""} ${activeClassSuffix}`.trim();
```

src/app/(proper_react)/layout.tsx
```diff
@@ -19,18 +22,74 @@ export default async function Layout({ children }: { children: ReactNode }) {
...
+                className={styles.mozillaLink}
+                target="_blank"
+              >
+                <Image
+                  src={mozillaLogo}
+                  width={100}
+                  alt={l10n.getString("mozilla")}
+                />
+              </a>
+              <ul className={styles.externalLinks}>
```
pdehaan, line 69: Q: Should these be in a `<nav>` wrapper? Or is it not valid to have multiple sets `<nav>` tags in headers and footers?

src/app/components/client/UserMenu.tsx
```diff
@@ -0,0 +1,23 @@
...
+import { signIn, signOut } from "next-auth/react";
+
+export type Props = {
+  session: Session | null;
+};
+
+export const UserMenu = (props: Props) => {
+  // Placeholder to enable login/logout until we create an actual menu
+  if (!props.session) {
+    return <button onClick={() => signIn("fxa")}>Sign in</button>;
```
pdehaan, line 17: nit: all feels very placeholdery, but not sure if we want to localize these buttons now or not. 🤷

pendingTranslations.ftl
```diff
@@ -5,3 +5,9 @@ main-nav-button-collapse-tooltip = Collapse menu
 main-nav-button-expand-label = Expand menu
 main-nav-button-expand-tooltip = Expand menu
 main-nav-link-home-label = Home
+main-nav-link-dashboard-label = Dashboard
+main-nav-link-faq-label = FAQs
+main-nav-link-faq-tooltip = Frequently asked questions
+
+footer-external-link-faq-label = FAQs
+footer-external-link-faq-tooltip = Frequently asked questions
```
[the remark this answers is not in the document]
flodolo: Yes: different place = different available space. So, localizers can choose a different translation if they want or need.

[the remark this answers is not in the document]
Vinnl: Yeah, additionally, according to Peiying it's relatively easy for localisers to re-use a string that's been used elsewhere, so she taught me to err on the side of duplication :)

src/app/(proper_react)/MobileShell.tsx
```diff
@@ -0,0 +1,92 @@
...
+};
+
+export const MobileShell = (props: Props) => {
+  const l10n = useL10n();
+  const [isExpanded, setIsExpanded] = useState(false);
+
+  return (
+    <div
+      className={`${styles.wrapper} ${
+        isExpanded ? styles.hasOpenMenu : styles.hasClosedMenu
```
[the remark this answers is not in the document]
Vinnl: It's indeed unused at this time, but I used it a bit in some initial experiments, and figured it's not unlikely we'll use it in the future, so I didn't remove it.

For the consistency I'm not sure, since obviously I'm not a native speaker, but I was under the impression that they're still in the same category: "the menu is open" vs. "the menu is closed"? i.e. adjective instead of past tense? But again, not a native speaker :sweat_smile:

src/app/(proper_react)/MobileShell.tsx
Vinnl: Yeah, it grated my sensibilities as well. But unfortunately, I couldn't come up with a way to do away with one. `.mainMenuLayer` is `position: sticky` to overlap the main content, but also has `height: 0` to make sure the content does not get pushed down when the menu opens. `.mainMenu` then again has a regular height that breaks out of `mainMenuLayer` but still overlaps the content.

I did not want to use the `<ul>` for `.mainMenu`, because the menu will later contain more elements than just the list of links (e.g. an "Upgrade" button).

src/app/(proper_react)/MobileShell.tsx
Vinnl: Yeah, so two reasons it doesn't go here:

1. `.content` also includes the sidebar menu, which shouldn't be in `<main>`. (Unfortunately, I expect the sidebar and drop-down menu to be different enough that duplication is the only sane way to deal with them.)
2. Specific pages might still have non-main content inside the layout - e.g. the dashboard will also include a tab bar, the app menu, and the user menu at the top.

But good call out - the risk is that we won't mark up any content as the main content. I'm hoping to deal with this by running Axe scans on the pages during unit tests, like we do on Relay - that already caught a couple of instances where content was not in any landmark elements there.

src/app/(proper_react)/PageLink.tsx
Vinnl: :shrug: works for me! cbd8876e1

src/app/(proper_react)/layout.tsx
Vinnl: Multple `<nav>` tags [are valid](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/nav#usage_notes), as I understand it:

> A document may have several [<nav>](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/nav) elements, for example, one for site navigation and one for intra-page navigation. [aria-labelledby](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-labelledby) can be used in such case to promote accessibility, see [example](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/Heading_Elements#labeling_section_content).

But [supposedly](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/nav#usage_notes) not in footers:

> It's not necessary for all links to be contained in a <nav> element. <nav> is intended only for a major block of navigation links; typically the [<footer>](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/footer) element often has a list of links that don't need to be in a [<nav>](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/nav) element.

flozia (approved):
Looks good so far!

src/app/components/client/RedirectToSignIn.tsx
```diff
@@ -2,6 +2,15 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
-export default function TestPage() {
-  return <span>Hoi</span>;
-}
+"use client";
```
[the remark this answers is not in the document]
Vinnl: daa31356f

PR #3088, opened 2023-05-31
MNTOR-1773/move sentry error handler earlier in startup

rhelmer:
# References: 
Jira: MNTOR-1773

# Description

We saw some exceptions and error messages that indicated that exceptions were not being loaded into app local storage, but we only actually got Sentry exceptions (with stack traces etc) from the settings code trying to read it.

This is because the Sentry error handler is loaded (as Express middleware) pretty late in startup, this PR moves it to load as early as possible.

EDIT: see also https://github.com/mozilla/blurts-server/pull/3088#issuecomment-1569539983 - the function that is throwing the exception we want is actually swallowing the exception and throwing a more generic message, but we want that stack trace! I've added a commit to address this as well.

# How to test

I did a positive test by making sure that the app functions correctly and passes tests, and a negative test by modifying `app.js` to throw an exception in startup right after the Sentry middleware is loaded, and confirmed that it appears in the `dev` environment in Sentry.

### DIFF
--- a/src/app.js
+++ b/src/app.js
@@ -43,6 +43,13 @@ Sentry.init({
   }
 })
 
+// sentry error handler
+app.use(Sentry.Handlers.errorHandler({
+  shouldHandleError (error) {
+    if (error instanceof RateLimitError) return true
+  }
+}))
+
 // Determine from where to serve client code/assets:
 // Build script is triggered for `npm start` and assets are served from /dist.
 // Build script is NOT run for `npm run dev`, assets are served from /src, and nodemon restarts server without build (faster dev).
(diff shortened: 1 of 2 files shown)


rhelmer:
Hm, it looks like there's an attempt to catch and log the exception here:

https://github.com/mozilla/blurts-server/blob/83a32d11b4a95abbde24088fa9dcdcba2c5fc00a/src/app.js#L169

But the exception we really want is being swallowed here:

https://github.com/mozilla/blurts-server/blob/eb8ac5acaf91f71f490c64b3a85fcc43fe6889f1/src/utils/hibp.js#LL159C1-L159C1

This isn't being returned to the user, I'm not sure it's at all valuable to use `fluentError` here. I think `loadBreachesIntoApp` should just not attempt to handle this exception and let it throw so it can be caught and logged by `app.js`.

mansaj (approved):
Try and get this hotfix in today?

rhelmer:
> Try and get this hotfix in today?

OK I am going to rebase it against `main`, I had created this against the latest release but it'd be better to ride along with the release.

PR #3093, opened 2023-05-31
Re-enable analytic events

flozia:
# References: 
Jira: [MNTOR-1643](https://mozilla-hub.atlassian.net/browse/MNTOR-1642)

# Description

Re-enables analytics for the ported Next.js site.

# How to test

- Visit `/users/dashboard`
- Toggle the user menu
- `opened_closed_user_menu` event should be fired

### DIFF
--- a/public/nextjs_migration/client/js/analytics.js
+++ b/public/nextjs_migration/client/js/analytics.js
@@ -1,0 +1,41 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+const bodyDataset = document.body.dataset
+
+async function init({ debugMode, ga4MeasurementId }) {
+  if (navigator.doNotTrack === '1') {
+    window.gtag = function () {
+      console.debug('Analytics disabled by DNT')
+    }
+    window.gtag()
+  } else {
+    await import(`https://www.googletagmanager.com/gtag/js?id=${ga4MeasurementId}`)
+
+    window.dataLayer = window.dataLayer || []
+
+    window.gtag = function () {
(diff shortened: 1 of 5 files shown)


public/nextjs_migration/client/js/analytics.js
```diff
@@ -0,0 +1,41 @@
...
+    // Instrument CTA clicks for analytics.
+    document.querySelectorAll('[data-cta-id]').forEach(cta =>
+      cta.addEventListener('click', () => window.gtag('event', 'clicked_cta', { cta_id: cta.dataset.ctaId })))
+  }
+}
+
+if (bodyDataset) {
+  const analyticsConfig = Object.freeze({
+    debugMode: bodyDataset.nodeEnv !== 'production',
+    ga4MeasurementId: bodyDataset.ga4MeasurementId || 'G-CXG8K4KW4P'
```
flozia, line 37: The `GA4_MEASUREMENT_ID` is not a secret and we already have it in the public `AppConstants`: https://github.com/mozilla/blurts-server/blob/main/src/client/js/appConstants.js.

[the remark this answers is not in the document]
flozia: Yes — that is just a fallback value to replicate the behavior from here: https://github.com/mozilla/blurts-server/blob/main/src/client/js/appConstants.js#LL23C13-L23C13. This might not even be necessary since we won’t be building with ESbuild and the value will be picked up from the `.env` file regardless of the environment, though @rhelmer?

rhelmer: Yes that's true it's not secret, but it can be helpful to have it configurable. I'm not too concerned about having it as a fallback though.

rhelmer (approved):
This lgtm, I gave it a try and I'm seeing pings go out in the browser console, with debug mode set correctly. Nice work!

PR #3095, opened 2023-05-31
MNTOR-1631: email APIs

mansaj:
# References: 
Jira: MNTOR-1631

# Description
Porting email APIs to NextJS

# Screenshot (if applicable)

Not applicable.

# How to test


src/app/api/v1/user/email/route.ts
```diff
@@ -0,0 +1,137 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import { getToken } from "next-auth/jwt";
+import { NextRequest, NextResponse } from "next/server";
+import AppConstants from "../../../../../appConstants";
```
pdehaan, line 7: semi-related but definitely non-blocking: I wonder if we could use something like [path mapping](https://www.typescriptlang.org/docs/handbook/module-resolution.html#path-mapping) (or https://www.typescriptlang.org/tsconfig/#paths or https://medium.com/slackernoon/use-typescript-aliases-to-clean-up-your-import-statements-7210b7ec2af1) to shorten these imports and remove the deeply nested "../../" paths.

src/app/api/util/email.ts
```diff
@@ -0,0 +1,31 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
```
pdehaan, line 1: Should this folder be "util" singular or "utils" plural?
I noticed we have a different folder pluralized to "utils" below (L6) for an import from `"../../../utils/email"`.

src/app/api/v1/user/resend-email/route.ts
```diff
@@ -0,0 +1,65 @@
...
+          },
+          { status: 500 }
+        );
+      }
+
+      await sendVerificationEmail(subscriber, emailId);
+
+      return NextResponse.json({
+        success: true,
+        message: "Sent the verification email",
```
pdehaan, line 45: Q: Why are some responses (L36 above) translated and others aren't?

src/app/api/v1/user/update-comm-option/route.ts
```diff
@@ -0,0 +1,38 @@
...
+} from "../../../../../db/tables/subscribers";
+
+export async function POST(req: NextRequest) {
+  const token = await getToken({ req });
+
+  if (token) {
+    try {
+      const body = await req.json();
+      const subscriber = await getSubscriberByEmail(token.email);
+      // 0 = Send breach alerts to the email address found in brew breach.
```
pdehaan, line 26: Q: What's "brew breach"?

src/app/api/v1/user/resend-email/route.ts
mansaj: Good question. It's because some of these responses aren't meant to be seen by users. They are more for our internal consumption. We should be translating most if not all user-facing messages

src/app/api/util/email.ts
mansaj: Fixed. Thanks

src/app/api/v1/user/email/route.ts
rhelmer: @pdehaan that looks really cool - we should have a separate PR to introduce it and convert everything over though. I'd like input from front-end folks like @flozia and @Vinnl on whether they think it's a good thing for us to use. I'm not really familiar with it but sounds promising, thanks!

rhelmer: I would like to do something about our import statements for sure, it's kind of a papercut when refactoring.

rhelmer (changes requested):
Code lgtm, but when I try to test with a new user I get:
```
src/utils/breaches.js (21:51) @ id

  19 |  */
  20 | async function getAllEmailsAndBreaches (user, allBreaches) {
> 21 |   const monitoredEmails = await getUserEmails(user.id)
     |                                                   ^
  22 |   const verifiedEmails = []
  23 |   const unverifiedEmails = []
  24 |   verifiedEmails.push(await bundleVerifiedEmails({ user, email: user.primary_email, recordId: user.id, recordVerified: user.primary_verified, allBreaches }))
```

I remember discussing that earlier this week, is this just not implemented yet? Is there a ticket for it?

Also I'm wondering what the plan for unit tests is - IIRC at least some of this existing code is covered, right? Have we decided what framework etc. we'd like to use for NextJS?

Vinnl (commented):
The API changes look good on first look, though I've yet to rebase the settings page work on this to see it work.

I haven't looked at the lint-related changes yet, but maybe those could go in a separate PR? There's a couple of things I'm wary about (primarily `type: "module"`, which I believe [is not supported yet](https://github.com/vercel/next.js/issues/9607#issuecomment-944156493)), and things I don't think we should do (removing stylelint), but I don't think those should be blockers for the rest of this PR.

src/app/api/v1/user/email/route.ts
maxxcrawford: > as more of our code is TypeScript, editors can more easily suggest the right auto-imports, as well as update them when you move files around.

This auto-magic™ is really helpful. I don't stress about paths bc of them.

mansaj: I agree with @Vinnl's points above. I think the issue will resolve itself after our Nextjs migration, and overtime as we become more disciplined with our types.

mansaj:
> The API changes look good on first look, though I've yet to rebase the settings page work on this to see it work.
> 
> I haven't looked at the lint-related changes yet, but maybe those could go in a separate PR? There's a couple of things I'm wary about (primarily `type: "module"`, which I believe [is not supported yet](https://github.com/vercel/next.js/issues/9607#issuecomment-944156493)), and things I don't think we should do (removing stylelint), but I don't think those should be blockers for the rest of this PR.

hmm interesting. This branch still seem to run fine with `type:module` set. Let's move the stylelint discussion to slack. I'd like to learn more about it

src/app/api/v1/user/email/route.ts
```diff
@@ -0,0 +1,142 @@
...
+  addSubscriberUnverifiedEmailHash,
+  removeOneSecondaryEmail,
+  getEmailById,
+} from "../../../../../db/tables/emailAddresses.js";
+
+import { sendVerificationEmail } from "../../../utils/email";
+
+import { validateEmailAddress } from "../../../../../utils/emailAddress";
+import { getL10n } from "../../../../functions/server/l10n";
+const l10n = getL10n();
```
Vinnl, line 23: `getL10n` processes the request headers to determine which locale to load a string from, so it can't be called from the top level - it needs to be inside a request context, e.g. inside `POST()`.

From there, you can pass it on to function that need it - I added `getStringLookup` that takes it as an argument and then returns a replacement for `getMessage`, to ease the transition.

src/app/api/v1/user/email/route.ts
```diff
@@ -0,0 +1,142 @@
...
+          });
+        }
+      }
+
+      const unverifiedSubscriber = await addSubscriberUnverifiedEmailHash(
+        subscriber,
+        validatedEmail.email
+      );
+
+      await sendVerificationEmail(subscriber, unverifiedSubscriber.id);
```
Vinnl, line 80: Before being able to send an email, I had to add a call to `initEmail`, otherwise `gTransporter` in `src/utils/email.js` is undefined (in the Express app, this would be done on startup).

src/app/api/v1/user/email/route.ts
```diff
@@ -0,0 +1,142 @@
...
+    // Not Signed in, redirect to home
+    return NextResponse.redirect(AppConstants.SERVER_URL, 301);
+  }
+}
+
+interface EmailDeleteRequest {
+  emailId: string;
+}
+
+export async function DELETE(req: NextRequest) {
```
Vinnl, line 111: While `DELETE` makes more sense, the actual API the frontend was using here was a `POST` call to `/remove-email` :unamused:

src/app/api/v1/user/verify-email/route.ts
```diff
@@ -0,0 +1,17 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import { NextRequest, NextResponse } from "next/server";
+import { verifyEmailHash } from "../../../../../db/tables/emailAddresses.js";
+export async function POST(req: NextRequest) {
```
Vinnl, line 7: This is called as a `GET` request.

Vinnl (commented):
Just pointing out some changes I had to make in https://github.com/mozilla/blurts-server/pull/3105, which might be helpful for future work.

src/app/api/v1/user/email/route.ts
mansaj: Right, I was going to make that change on the frontend side (to use DEL) but up to you

Vinnl: Oh yeah that works too. I was trying to modify as little as possible now, but let's at least reconsider this when turning this into proper React. Should be a fairly minimal change anyway.

src/app/api/v1/user/email/route.ts
```diff
@@ -0,0 +1,142 @@
...
+  if (token) {
+    try {
+      const { emailId }: EmailDeleteRequest = await req.json();
+      const subscriber = await getSubscriberByEmail(token.email);
+      const existingEmail = await getEmailById(emailId);
+
+      if (existingEmail?.subscriber_id !== subscriber.id) {
+        return NextResponse.json({
+          success: false,
+          status: 400,
```
pdehaan, line 123: Same Q here re: `{ status: 400 }` as a second arg to `NextResponse.json({…}, { status: 400 })`.

mansaj:
Thanks @pdehaan, but I'm closing this PR. The work here has been absorbed into https://github.com/mozilla/blurts-server/pull/3105

PR #3098, opened 2023-06-03
MNTOR-949 - subscription platform integration for nextjs app

rhelmer:
# References: 
Jira: MNTOR-949

# Description

This provides basic subplat integration. Currently it demonstrates:

1. add new `subscription` `Array<string>` to `session.user.fxa` object 
3. example "subscribe" button for staging subplat
4. example welcome screen and badge in user menu

This PR is based on a recent spike, and still needs the following to be truly ready for review:

- [x] move configuration to env vars
- [x] add flag (disabled by default)
- [x] add example UI (also behind flag)

Per the [next-auth docs](https://next-auth.js.org/getting-started/client#updating-the-session), we should be able to update the session without requiring the user to log in again.

# How to test

1. sign in using unsubscribed account
2. click "Subscribe" button on user dashboard page
3. complete purchase using [stripe test card](https://stripe.com/docs/testing)
5. ensure that:
 a. redirected to the dashboard after login
 b. "Subscribe" button on user dashboard does not appear
6. Navigate to Firefox account settings and unsubscribe
7. Log out/in and ensure that premium features are still available for 1 month
8. TODO figure out how to test that cancellation worked without having to wait 1 month


src/app/(nextjs_migration)/(authenticated)/user/breaches/page.tsx
```diff
@@ -134,6 +134,11 @@ export default async function UserBreaches() {
       <Script src="/nextjs_migration/client/js/circleChart.js" />
       <Script src="/nextjs_migration/client/js/breaches.js" />
       <main data-partial="breaches">
+        <section>
+          <a href="https://accounts.stage.mozaws.net/subscriptions/products/prod_NErZh679W62lai?plan=price_1MUNq0Kb9q6OnNsL4BoJgepf">
+            <button className="button primary">Subscribe to premium</button>
```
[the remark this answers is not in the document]
rhelmer: Hm that's a good point! This isn't the final design just a working example :) But I'm curious now too, will find out.

src/app/(nextjs_migration)/components/client/UserMenu.tsx
```diff
@@ -56,6 +56,11 @@ export const UserMenu = ({ session, fxaSettingsUrl }: Props) => {
           </a>
           <hr />
         </li>
+        <li>
+          <p className="user-menu-header">
```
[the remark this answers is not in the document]
rhelmer: Yes temp UI, no it probably doesn't.

rhelmer:
BTW this conflicts with https://github.com/mozilla/blurts-server/pull/3101 so I'm going to change this one to be consistent with it - the word `subscriber` is unfortunately overloaded (Monitor calls existing users "subscribers", e.g. the DB table and in the code), and FxA also includes a `subscription` claim in the JWT. So we'll want to be really consistent in how we use these.

src/app/api/auth/[...nextauth]/route.ts
```diff
@@ -85,6 +87,7 @@ export const authOptions: AuthOptions = {
       session.user.metricsEnabled = token.metricsEnabled;
       session.user.avatar = token.avatar;
       session.user.avatarDefault = token.avatarDefault;
+      session.user.subscriber = token.subscriptions?.includes("monitor");
```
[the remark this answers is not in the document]
Comment (rhelmer) [reply src/app/api/auth/[...nextauth]/route.ts]: I pulled in PR 3101 and this changed completely - `subscriber`/`subscription` is a bit overloaded unfortunately. For the moment I've just exposed the list of subscriptions on the `session.fxa` object (which can be undefined)

src/app/(nextjs_migration)/(authenticated)/user/premium-welcome/page.tsx
```diff
@@ -0,0 +1,17 @@
...
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+"use client";
+import { useSession } from "next-auth/react";
+
+export default async function PremiumWelcome() {
+  const { update } = useSession();
+
+  const session = await update();
```
rhelmer, line 11: @Vinnl hey per [the docs](https://next-auth.js.org/getting-started/client#updating-the-session), we should be able to call `update()` to refresh the JWT from FxA, but I get this:

```
[next-auth][error][CLIENT_FETCH_ERROR] 
https://next-auth.js.org/errors#client_fetch_error Unexpected token o in JSON at position 1 {
  error: {
    message: 'Unexpected token o in JSON at position 1',
    stack: 'SyntaxError: Unexpected token o in JSON at position 1\n' +
      '    at JSON.parse (<anonymous>)\n' +
      '    at Response.json (node:internal/deps/undici/undici:6160:23)\n' +
      '    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)',
    name: 'SyntaxError'
  },
  url: 'http://localhost:6060/api/auth/session',
  message: 'Unexpected token o in JSON at position 1'
}
```

Am I using it wrong or is there something wrong with it? The endpoint above seems to return proper JSON.

Vinnl: Hmm, I don't get that error. The way it is written now does result in the function getting called every time the component re-renders, which can be often. Usually, you will want to wrap side effects like this into [a `useEffect` call](https://react.dev/learn/synchronizing-with-effects), and since this should run only once when the page first loads, with an empty `dependencies` parameter `[]`. Not sure if that'll solve your issue though, since I can't reproduce.

rhelmer: Ah right, that actually did seem to solve the problem. I'll update this to use `useEffect` properly, thanks!

rhelmer: OK well that worked, *but* when I test with a new account, it looks like `update` doesn't actually refresh the JWT from FxA unfortunately.

Requiring the user to sign in again works, but I think we should be able to do something like https://next-auth.js.org/v3/tutorials/refresh-token-rotation to do this without having to bother the user.

We could leave this for the front-end work, wdyt @Vinnl?

Vinnl: @rhelmer Ah, so as I understand it we need to distinguish two things here. The first are the refresh and access tokens we get from FxA, that verify that the user indeed controls that account. We store those in our database, in the `subscribers` table.

Then there's the JWT that _we_ generate and sign using our secret, in which we state that we have done the FxA verification. When you call `update`, _that_'s the token that gets refreshed.

And if I understand it correctly, you actually want to refresh the profile data we get from FxA. From what I can see, I don't think Next-Auth currently has an officially supported method to do that. However, I think that using the access token that we store in the database, you might be able to replicate [this call](https://github.com/mozilla/blurts-server/blob/fa4f784c789afb1201bffbc696ceef5f5637bf7d/src/app/api/auth/%5B...nextauth%5D/route.ts#L70-L75) to re-fetch the profile data? And if you are able to do so in Next-Auth's [`jwt` callback](https://next-auth.js.org/configuration/callbacks#jwt-callback), you could even persist the new data in the token right away.

Vinnl: Oh but yes, we could also defer looking at that until when we implement the UI.

rhelmer:
OK @Vinnl as discussed, I've removed the sample UI code and just left an API handler for the subplat redirect with some notes about what needs to be done. We could have this redirect to wherever we like, but I figured an API would be the most flexible.

I've left the issue around making the user re-authenticate (or trying to do it for them via a refresh token or other approach) to the upcoming front-end PR :)

Vinnl (commented):
Hey Rob, could you add some guidance on what we should primarily look at? `NEXT_PUBLIC_SUBSCRIBE_PREMIUM` doesn't seem to be used anywhere, I don't see the route `/api/v1/premium/`  set anywhere as a redirect URL, and I don't see the `subscriptions` property in my FxA profile data (do we need an additional scope or something?). I guess I'm mostly wondering what/how to test this locally.

rhelmer:
Sorry for the noise - the `nextjs` branch landed, so I rebased this PR against `main`. I temporarily put it back in draft mode, while I re-implement a basic UI and flag so it's easier to test.

rhelmer:
> Hey Rob, could you add some guidance on what we should primarily look at? `NEXT_PUBLIC_SUBSCRIBE_PREMIUM` doesn't seem to be used anywhere, I don't see the route `/api/v1/premium/` set anywhere as a redirect URL, and I don't see the `subscriptions` property in my FxA profile data (do we need an additional scope or something?). I guess I'm mostly wondering what/how to test this locally.

You need to set two env vars:
```
PREMIUM_ENABLED=true
SUBSCRIBE_PREMIUM_URL=https://accounts.stage.mozaws.net/subscriptions/products/prod_NErZh679W62lai?plan=price_1MUNq0Kb9q6OnNsL4BoJgepf
```

I've added very simple test UI that's gated by `PREMIUM_ENABLED`, it prompts you to sign out and then redirects you to the dashboard, which prompts you to sign back in. @Vinnl I suggest we land it like this, disabled by default, and investigate ways to do the re-authentication more seamlessly when the real UI is implemented. wdyt?

src/app/(nextjs_migration)/(authenticated)/user/breaches/page.tsx
```diff
@@ -157,6 +157,11 @@ export default async function UserBreaches() {
       <Script type="module" src="/nextjs_migration/client/js/breaches.js" />
       <Script type="module" src="/nextjs_migration/client/js/dialog.js" />
       <main data-partial="breaches">
+        <section>
+          {(process.env.PREMIUM_ENABLED === "true" && !session?.user.fxa?.subscriptions?.includes("monitor")) ?
+            <a className="button primary" href={process.env.SUBSCRIBE_PREMIUM_URL}>Subscribe to Premium</a>
+            : ''}
```
rhelmer, line 163: Maybe this should be a component (`SignupButton`?), so it can be displayed or not based on some state such as whether the user is already subscribed, are not in US, and so on?

src/app/(nextjs_migration)/(authenticated)/user/breaches/page.tsx
```diff
@@ -157,6 +157,11 @@ export default async function UserBreaches() {
       <Script type="module" src="/nextjs_migration/client/js/breaches.js" />
       <Script type="module" src="/nextjs_migration/client/js/dialog.js" />
       <main data-partial="breaches">
+        <section>
+          {(process.env.PREMIUM_ENABLED === "true" && !session?.user.fxa?.subscriptions?.includes("monitor")) ?
+            <a className="button primary" href={process.env.SUBSCRIBE_PREMIUM_URL}>Subscribe to Premium</a>
```
pdehaan, line 162: Q: should this be localized?

src/app/api/v1/premium/route.ts
```diff
@@ -0,0 +1,30 @@
...
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import { NextResponse } from "next/server";
+import type { NextRequest } from "next/server";
+
+export async function GET(req: NextRequest) {
+  try {
+    const { searchParams } = new URL(req.url);
+    const email = searchParams.get("email");
+    console.debug("Subscription redirect for:", email);
```
pdehaan, line 12: Q: since this is an API route, where does this get logged? Terminal? Log file? And is this some PII risk, or was it just meant for local debuggery?

src/app/(nextjs_migration)/(authenticated)/user/breaches/page.tsx
Vinnl: Let's leave that for when/where we actually implement upgrades.

src/app/(nextjs_migration)/(authenticated)/user/breaches/page.tsx
Vinnl: This is just for our own debugging purposes, so I wouldn't bother our localisers with it.

src/app/api/auth/[...nextauth]/route.ts
```diff
@@ -38,6 +38,7 @@ interface FxaProfile {
   /** URL to an avatar image for the current user */
   avatar: string;
   avatarDefault: boolean;
+  subscriptions: Array<string>;
```
Comment (Vinnl) [inline src/app/api/auth/[...nextauth]/route.ts:41]: When the user doesn't have any subscriptions, FxA doesn't add this property:

```suggestion
  subscriptions?: Array<string>;
```

For the other properties we'll either have to provide a fallback (`?? []`), or update those types as well to mark this one as optional.

Vinnl (approved):
It works, but playing around with it a bit, I've gotten more worried that we won't be able to update the profile data in the token without signing the user out. I was able to fetch updated profile data in the `jwt` callback using

```ts
      if (headers().get('x-invoke-path') === "/api/v1/premium" && !profile) {
        const fxaProvider = authOptions.providers[0] as any;
        const userInfo = await fxaProvider.userinfo.request({ tokens: { access_token: token.subscriber?.fxa_access_token } })
        profile = await fxaProvider.profile(userInfo);
      }
```

However, that did not lead to the token being updated in subsequent requests. Maybe it's just that we can't update a cookie in a redirect?

src/app/(nextjs_migration)/(authenticated)/user/breaches/page.tsx
rhelmer: I think this feature is going to be English an US-only, at least at first, so we probably want to avoid bothering localizers.

src/app/api/v1/premium/route.ts
rhelmer: Local just for demonstrative purposes, I'll just remove it, good catch thanks!

rhelmer:
> It works, but playing around with it a bit, I've gotten more worried that we won't be able to update the profile data in the token without signing the user out. I was able to fetch updated profile data in the `jwt` callback using
> 
> ```ts
>       if (headers().get('x-invoke-path') === "/api/v1/premium" && !profile) {
>         const fxaProvider = authOptions.providers[0] as any;
>         const userInfo = await fxaProvider.userinfo.request({ tokens: { access_token: token.subscriber?.fxa_access_token } })
>         profile = await fxaProvider.profile(userInfo);
>       }
> ```
> 
> However, that did not lead to the token being updated in subsequent requests. Maybe it's just that we can't update a cookie in a redirect?

Hm, OK well I'm going to land it as-is, I think there's a way to do this where we redirect users to FxA, but I'm hoping there's an easy way to update the server-side session with the latest JWT from FxA without having to do any kind of redirect. Let's tackle that as part of implementing the real subscribe experience.

PR #3114, opened 2023-06-09
Nextjs premerge

Vinnl:
A couple of changes that were necessary to make the built+deployed version work:

- Add `type="module"` to `<script>` tag for the ported client-side scripts. We already had this on the old website, and it gives each module its own scope - which is good, because there are overlapping identifiers in them (e.g. `init`).
- Fix the loading of the user's avatar - we used Next.js's `<Image>` component, but the avatar is an SVG, which it doesn't load.
- Use `<BreachLogo>` to display a company's logo on the dashboard as well. We already used it on the public breach pages as a replacement for the previous function that generated an HTML string.
- Allow Next.js's dynamically-injected scripts in our Content Security Policy.

This branch is currently deployed to Heroku, where you can see it in action.

Note that it builds on https://github.com/mozilla/blurts-server/pull/3112, so I set that as the PR target. As that gets merged, this PR's target should update to `nextjs` as well.

### DIFF
--- a/next.config.js
+++ b/next.config.js
@@ -52,7 +52,8 @@ const nextConfig = {
               `script-src 'self' ${
                 process.env.NODE_ENV === "development"
                   ? "'unsafe-eval' 'unsafe-inline'"
-                  : ""
+                  : // See https://github.com/vercel/next.js/discussions/51039
+                    "'unsafe-inline'"
               } https://*.googletagmanager.com`,
               "script-src-attr 'none'",
               `connect-src 'self' ${
@@ -64,9 +65,7 @@ const nextConfig = {
                   : ""
               }`,
               "child-src 'self'",
-              `style-src 'self' ${
-                process.env.NODE_ENV === "development" ? "'unsafe-inline'" : ""
-              }`,
+              "style-src 'self' 'unsafe-inline'",
               "font-src 'self'",
               "form-action 'self'",
               "frame-ancestors 'self'",
--- a/src/app/(nextjs_migration)/(authenticated)/user/breaches/page.tsx
+++ b/src/app/(nextjs_migration)/(authenticated)/user/breaches/page.tsx
@@ -154,10 +154,10 @@ export default async function UserBreaches() {
 
   return (
     <>
-      <Script src="/nextjs_migration/client/js/customSelect.js" />
-      <Script src="/nextjs_migration/client/js/circleChart.js" />
-      <Script src="/nextjs_migration/client/js/breaches.js" />
-      <Script src="/nextjs_migration/client/js/dialog.js" />
+      <Script type="module" src="/nextjs_migration/client/js/customSelect.js" />
+      <Script type="module" src="/nextjs_migration/client/js/circleChart.js" />
+      <Script type="module" src="/nextjs_migration/client/js/breaches.js" />
+      <Script type="module" src="/nextjs_migration/client/js/dialog.js" />
       <main data-partial="breaches">
         <section>
           <header className="breaches-header">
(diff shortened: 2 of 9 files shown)


src/app/(nextjs_migration)/(authenticated)/user/breaches/page.tsx
```diff
@@ -233,11 +232,7 @@ export default async function UserBreaches() {
             <span></span>
             <span>{l10n.getString("column-detected")}</span>
           </header>
-          <div
-            dangerouslySetInnerHTML={{
-              __html: createBreachRows(userBreachesData),
-            }}
-          />
+          <div>{createBreachRows(userBreachesData)}</div>
```
maxxcrawford, line 235: praise: _I also like to live dangerously._

src/app/(nextjs_migration)/components/client/UserMenu.tsx
```diff
@@ -28,19 +28,20 @@ export const UserMenu = ({ session, fxaSettingsUrl }: Props) => {
...
-      <Script src="/nextjs_migration/client/js/userMenu.js" />
+      <Script type="module" src="/nextjs_migration/client/js/userMenu.js" />
       <button
         aria-expanded="false"
         aria-haspopup="true"
         className="user-menu-button"
         title={l10n.getString("menu-button-title")}
       >
-        <Image
+        {/* The avatar is an SVG, which next/image doesn't process: https://nextjs.org/docs/pages/api-reference/components/image#dangerouslyallowsvg */}
```
maxxcrawford, line 38: question: should we refactor this tag to display an SVG file directly, instead of within an image/src?

next.config.js
```diff
@@ -64,9 +65,7 @@ const nextConfig = {
                   : ""
               }`,
               "child-src 'self'",
-              `style-src 'self' ${
-                process.env.NODE_ENV === "development" ? "'unsafe-inline'" : ""
-              }`,
+              "style-src 'self' 'unsafe-inline'",
```
maxxcrawford, line 68: question: Is this something that is for Heroku, or something we'll have to solve on prod too?

maxxcrawford (approved):
Solid work debugging Heroku, @Vinnl! I had a few questions.

src/app/(nextjs_migration)/components/client/UserMenu.tsx
Vinnl: The challenge is that the image is hosted by FxA, so we don't really know what it's going to look like.

next.config.js
rhelmer: Huh you're right, I wonder where that's being set? I don't see it in https://github.com/mozilla/blurts-server/blob/62ba006b4505d2eaf322b26bd946bc27f5d468aa/src/app.js#L109 or elsewhere (it can be overridden per-route as needed I think, I just don't see where it's happening now)

rhelmer (approved):
lgtm overall, although I'm wondering how `unsafe-inline` is getting set for `style-src` in the current app... is it possible this is being done on e.g. nginx config and not by the app?

mansaj (approved):

next.config.js
Vinnl: Oh yeah, that's surprising! It looks like [helmet's setting that by default](https://www.npmjs.com/package/helmet#content-security-policy):

> `Content-Security-Policy`
> Default
>     Content-Security-Policy: default-src 'self';base-uri 'self';font-src 'self' https: data:;form-action 'self';frame-ancestors 'self';img-src 'self' data:;object-src 'none';script-src 'self';script-src-attr 'none';style-src 'self' https: 'unsafe-inline';upgrade-insecure-requests

The style-src bit: `style-src 'self' https: 'unsafe-inline'`

I do expect that it's riskier for `script-src` than for `style-src`, so hopefully Next.js [will allow a more restrictive policy](https://github.com/vercel/next.js/discussions/51039) soon.

PR #3123, opened 2023-06-13
add build step in dockerfile

mansaj:
# Description
Fix stage build by adding build step in docker

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -14,6 +14,7 @@ COPY package-lock.json package-lock.json
 COPY --chown=app:app . /app
 
 RUN npm ci --audit=false && rm -rf ~app/.npm /tmp/*
+RUN npm run build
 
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -14,6 +14,7 @@ COPY package-lock.json package-lock.json
 COPY --chown=app:app . /app
 
 RUN npm ci --audit=false && rm -rf ~app/.npm /tmp/*
+RUN npm run build
```
maxxcrawford, line 17: Is this the correct time to run? (Before setting ENVs)

maxxcrawford (approved):
Conditionally approving to try deployment on stage.

Dockerfile
mansaj: Actually you are right, fixed

rhelmer (approved):

PR #3124, opened 2023-06-14
Set up l10n and optimised fonts for Storybook

Vinnl:
This updates our Storybook configuration to:

- support loading localised strings (c5e491abf1db6807293ffdd05d6182dc1f818435)
- support Next.js's font optimisation for Metropolis and Inter, and adds them to the tokens file (5911326ac1467e554795e7a543abe3d23ac2fe59)
- ensure our CSS reset is loaded in Storybook as well, for production-like styling (68e263f0b98ce7000251301a18976e98e932bf82)
- splits the `(proper_react)` root layout into a data-fetching part and a rendering part, and showcases the latter in Storybook (bb5f417bf95d0501b731adbbd1963b13e88c4d45)

That last one is probably the least interesting (no actual changes, just moving code around and creating a new story), but it's the one making this PR look biggest - but it also helps demonstrating the other changes. So for code review, it might help to look at just the other commits.


.storybook/main.ts
```diff
@@ -13,5 +13,22 @@ const config: StorybookConfig = {
...
+  async webpackFinal(config) {
+    config.module ??= {};
+    config.module.rules ??= [];
+    config.module.rules.push({
+      test: /\.ftl/,
+      type: "asset/source",
+    });
+
+    return config;
+  },
```
Vinnl, line 32: This replicates https://github.com/mozilla/blurts-server/blob/fae90e0dbbfe38c0b92c413b0c46b34d7de3a455/next.config.js#L153-L156

It allows us to `require` `.ftl` files.

.storybook/preview.tsx
Vinnl, line 1: This was renamed from `preview.ts` because it adds JSX in the `AppDecorator`. The two actual changes, in that decorator, are:

- Add `<L10nProvider>` to that components rendered by Storybook can call `useL10n`.
- Apply the `inter` and `metropolis` classes so that components rendered by Storybook default to the Inter font (`inter.className`) and can reference the Inter and Metropolis CSS variables (`inter.variable` and `metropolis.variable`).

src/app/(proper_react)/Shell.tsx
Vinnl, line 1: This file is just a verbatim copy of the children of `<L10nProvider>` in `./layout.tsx` - there are no further changes, other than data coming in through props.

src/app/functions/server/l10n.ts
Vinnl, line 1: When running the Next.js app, the Fluent bundles are collected on the server based on the language set in the request headers. In Storybook, we don't have a server-side component, so we don't read request headers (instead using `navigator.languages`), and we don't need JSDOM to render tags inside l10n strings (because the actual DOM is available).

src/app/layout.tsx
```diff
@@ -46,7 +47,7 @@ export default async function RootLayout({
   return (
     <html lang={currentLocale}>
       <body
-        className={inter.className}
+        className={`${inter.className} ${inter.variable} ${metropolis.variable}`}
```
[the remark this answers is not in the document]
Vinnl: Depends on what we want, really. `inter.variable` just makes it available to the tokens, `inter.className` also sets it as the default font for all children of `<body>`. I'm happy to remove it, I think you have a better idea about what we'd like to do there.

.storybook/preview.tsx
```diff
@@ -0,0 +1,38 @@
+import React from "react";
+import { Inter } from "next/font/google";
+import type { Preview } from "@storybook/react";
+import "../src/app/globals.css";
+import { L10nProvider } from "../src/contextProviders/localization";
+import { getL10nBundles } from "../src/app/functions/server/l10n";
+import { metropolis } from "../src/app/fonts/Metropolis/metropolis";
+
+const inter = Inter({ subsets: ["latin"], variable: "--font-inter" });
```
maxxcrawford, line 9: Question/Suggestion (for another PR?): Should we bring in the fonts the same way, instead of importing one from Google and one locally? My sense is to have the fonts all serve the same way/location for reliability.

src/app/tokens.scss
```diff
@@ -192,18 +190,18 @@ $layout-2xl: 192px;
 
 // Typeface
 // Title copy
-$text-title-lg: 700 56px / 56px Metropolis, sans-serif;
-$text-title-md: 700 48px / 48px Metropolis, sans-serif;
-$text-title-sm: 700 38px / 40px Metropolis, sans-serif;
-$text-title-xs: 700 28px / 30px Metropolis, sans-serif;
-$text-title-2xs: 700 24px / 26px Metropolis, sans-serif;
-$text-title-3xs: 700 18px / 20px Metropolis, sans-serif;
+$text-title-lg: 700 56px / 56px var(--font-metropolis), sans-serif;
```
maxxcrawford, line 193: Q: Should `var(--font-metropolis)` be a SCSS variable: `$font-metropolis`, etc. ?

maxxcrawford (approved):
A few questions. Thanks for the fix!

.storybook/preview.tsx
Vinnl: Yeah good question. I don't have a strong opinion; I like the consistency, but I also like automatically getting the latest/most complete version, and not vendoring dependencies where possible. But I'd approve a PR that changes it to a local font :)

src/app/tokens.scss
Vinnl: The variable [is defined by](https://github.com/mozilla/blurts-server/blob/c450f4535d096d0e1ed94c7d07e87d95a5e29af4/src/app/fonts/Metropolis/metropolis.ts#L25) Next.js's `localFont`, and is what allows us to use it in an SCSS variable while keeping its optimisations.

We could create a separate `$font-metropolis: var(--font-metropolis)` variable, but for that I think the question is whether we want other parts of the code to use that font independently, or if it should use all the other font settings that are included in the `$text-title-*` etc variables.

PR #3129, opened 2023-06-15
Add accessibility add-on for Storybook

Vinnl:
This will make it easier to detect and debug accessibility issues inside Storybook. It adds an accessibility tab to the UI underneath a story, outlining potential issues. AFAIK this also uses Axe, so it should be helpful in visually debugging test failures once we add axe failures to our unit test suite.

Try it out here: https://deploy-preview-3129--fx-monitor-storybook.netlify.app


flozia (approved):

Vinnl:
Sounded like Maxx has already seen this too, so merging :)

PR #3130, opened 2023-06-15
MNTOR-1820: feature flags

mansaj:
# References: 
Jira: MNTOR-1820

# Description
Feature Flags setup
* Check for feature flags 
* Output feature flags in `runtime-data` endpoint
* Update DoD in github template

# Screenshot (if applicable)

Not applicable.

# How to test
* Set feature flags in `.env`, example: `FEATURE_FLAGS={"test": true}`
* Go to `/api/v1/runtime-data` to see it populated

### DIFF
--- a/src/app/api/utils/featureFlags.ts
+++ b/src/app/api/utils/featureFlags.ts
@@ -4,21 +4,30 @@
 
 import appConstants from "../../../appConstants"
 
-type FeatureFlags = Record<"test" | "otherFlag" | "otherFlag", FeatureFlag | boolean>;
+type FeatureFlags = Record<
+  "test" |
+  "otherFlag" |
+  "otherFlag2",
+  FeatureFlag | null>;
 
 type FeatureFlag = {
-  name?: keyof FeatureFlags;
+  name: keyof FeatureFlags;
+  enabled: boolean;
   description?: string;
   dependencies?: Array<keyof FeatureFlags>;
 }
 
 const featureFlags = JSON.parse(appConstants.FEATURE_FLAGS) as FeatureFlags
 
-export function checkFeatureFlag(flag: keyof FeatureFlags) : (FeatureFlag | boolean) {
-  if (featureFlags[flag]) return featureFlags[flag]
-  return false
+export function isFeatureEnabled(flag: keyof FeatureFlags): boolean {
+  return featureFlags[flag]?.enabled || false
 }
 
-export function allFeatureFlags() : FeatureFlags {
+export function checkFeatureFlag(flag: keyof FeatureFlags): (FeatureFlag | null) {
+  if (featureFlags[flag]) return featureFlags[flag]
+  return null
+}
+
+export function allFeatureFlags(): FeatureFlags {
   return featureFlags
 }
--- a/src/app/api/utils/featureFlags.ts
+++ b/src/app/api/utils/featureFlags.ts
@@ -17,7 +17,10 @@ type FeatureFlag = {
   dependencies?: Array<keyof FeatureFlags>;
 }
 
-const featureFlags = JSON.parse(appConstants.FEATURE_FLAGS) as FeatureFlags
+let featureFlags: FeatureFlags
+try {
+  featureFlags = JSON.parse(appConstants.FEATURE_FLAGS) as FeatureFlags
+} catch (e) { console.error(e) }
 
 export function isFeatureEnabled(flag: keyof FeatureFlags): boolean {
   return featureFlags[flag]?.enabled || false
--- a/src/app/api/utils/featureFlags.ts
+++ b/src/app/api/utils/featureFlags.ts
@@ -1,0 +1,26 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import appConstants from "../../../appConstants"
+
+type FeatureFlags = {
+  [key: string]: FeatureFlag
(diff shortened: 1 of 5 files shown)


src/app/api/utils/featureFlags.ts
```diff
@@ -0,0 +1,26 @@
...
+
+type FeatureFlag = {
+  name?: string
+  description?: string
+  dependencies?: [string]
+}
+
+const featureFlags = JSON.parse(appConstants.FEATURE_FLAGS) as FeatureFlags
+
+export function checkFeatureFlag(flag: string) : (FeatureFlag | boolean) {
```
Vinnl, line 19: Hmm, could we rename this to `isFeatureFlagEnabled` and make it return a boolean? We could always add a `getFeatureFlagDetails` if we need all that.

(Also, maybe the `FeatureFlag` type could also add an `enabled` boolean property? So that we can (temporarily) toggle it off without removing the reference to it.)

Vinnl, line 19:
```suggestion
type FeatureFlags = Record<"test" | "otherFlag" | "otherFlag", FeatureFlag | boolean>;

type FeatureFlag = {
  name?: keyof FeatureFlags;
  description?: string;
  dependencies?: Array<keyof FeatureFlags>;
}

const featureFlags = JSON.parse(appConstants.FEATURE_FLAGS) as FeatureFlags

export function checkFeatureFlag(flag: keyof FeatureFlags) : (FeatureFlag | boolean) {
```

If we make the available flags an explicit union type of literal strings, it'll be easier to remove/change them later on than if they're arbitrary strings.

Vinnl (approved):
Cool that we're getting this! I added a small suggestion to explicitly incorporate the available flags in our type, to make it easier to find references to the flags later, but given that we probably won't have thousands right away, we can do that later as well.

src/app/api/utils/featureFlags.ts
mansaj: @Vinnl The feature flags will be set via SRE in env vars (source of truth is whatever is set in the helm chart). The function here is purely to read that stringified JSON. I'm afraid that setting these restrictions would make it a lot more work than necessary (since everytime we change something in env var, we will need to remember to add that here)

I do like the keyof suggestion though

Vinnl: > since everytime we change something in env var, we will need to remember to add that here

Only if we add a new feature flag, right? And only if we actually reference it in our code, so then we're modifying our code anyway. Or am I misunderstanding what you mean?

(I'd generally advise against enums, as they're a language-level rather than a type-level feature, and are incompatible with the proposed JS enums.)

mansaj: It doesn't matter if the flag is actually used in our code, as long as it exists in env var, it will need to be ref'd here since the key is strict. Yes should only apply when adding new feature flag (in env var)

I also think it gives the illusion that once you change it here, it's somehow automatically enabled. The real "change" needs to still happen with SRE

Vinnl: Type checking only happens at build time, so our code wouldn't suddenly break if a new key gets added to the environment variable. You just wouldn't be able to write code that references that key without also updating the type.

The illusion is a good point, though the worst that will result in is that the feature flag wouldn't be enabled yet.

mansaj: Right.. hm I think another issue is there needs to be a tool that generates the json string for `.env` updates. I can already see this being a pretty annoying manual process. Anyways, I will commit this change

src/app/api/utils/featureFlags.ts
```diff
@@ -0,0 +1,24 @@
...
+
+type FeatureFlags = Record<"test" | "otherFlag" | "otherFlag", FeatureFlag | boolean>;
+
+type FeatureFlag = {
+  name?: keyof FeatureFlags;
+  description?: string;
+  dependencies?: Array<keyof FeatureFlags>;
+}
+
+const featureFlags = JSON.parse(appConstants.FEATURE_FLAGS) as FeatureFlags
```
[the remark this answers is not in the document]
mansaj: Thanks @pdehaan good catch

src/app/api/utils/featureFlags.ts
```diff
@@ -0,0 +1,36 @@
...
+  name: keyof FeatureFlags;
+  enabled: boolean;
+  description?: string;
+  dependencies?: Array<keyof FeatureFlags>;
+}
+
+let featureFlags: FeatureFlags
+try {
+  featureFlags = JSON.parse(appConstants.FEATURE_FLAGS) as FeatureFlags
+} catch (e) { console.error(e) }
```
pdehaan, line 23: I think this has a subtle bug here.
Default value of `featureFlags` is `undefined`.
Based on my quick testing, this will throw a `TypeError` if you try accessing a nested property on the `featureFlags` object (like the `isFeatureEnabled()` method below on L25-27.

```js
let featureFlags;
console.log(featureFlags); // undefined
featureFlags['someExperiment']?.enabled
// VM165:1 Uncaught TypeError: Cannot read properties of undefined (reading 'someExperiment')
//    at <anonymous>:1:13
```

I think the workaround is either default `featureFlags` to an empty object `let featureFlags: FeatureFlags = {}` when creating the variable, or setting the fallback empty `featureFlags` object in the `catch` block.

Other workaround might be using one more optional chaining on the Boolean `featureFlags[flag]?.enabled` lookup on L26 below:

```js
return featureFlags?.[flag]?.enabled || false
```

mansaj:
We decided to go with another approach. https://github.com/mozilla/blurts-server/pull/3149

PR #3132, opened 2023-06-15
Don't block responses while downloading images

Vinnl:
OK, so I haven't been able to properly set up performance monitoring myself yet, which means I'm shooting in the dark, which I don't like. Also, this code keeps getting iffier and iffier.

That said, [this trace](https://mozilla.sentry.io/performance/summary/?project=4504612374052864&query=&referrer=performance-transaction-summary&statsPeriod=14d&transaction=Page+Server+Component+%28%2Fbreaches%29&unselectedSeries=p100%28%29) seems to suggest that fetching icons can take a looong time, _and_ is delaying returning a response. So in this PR, I wrapped the whole thing in an async function, and then simply return an empty map while that function is executing. Once it's done, future function calls can get a non-empty map with initialised icons - but until then, at least the server should respond to requests.

### DIFF
--- a/src/app/functions/server/getBreaches.ts
+++ b/src/app/functions/server/getBreaches.ts
@@ -61,57 +61,64 @@ export async function getBreachIcons(breaches: Breach[]): Promise<LogoMap> {
     return new Map();
   }
   isFetchingIcons = true;
-  const breachDomains = breaches
-    .map((breach) => breach.Domain)
-    .filter((breachDomain) => breachDomain.length > 0);
-  const logoFolder = pathResolve(
-    dirname(fileURLToPath(import.meta.url)),
-    "../../../../public/logo_cache/"
-  );
-  try {
-    await mkdir(logoFolder);
-  } catch {
-    // Do nothing; if the directory already exists, that's fine.
-  }
-  const existingLogos = await readdir(logoFolder);
-  // TODO: Batch to limit memory use?
-  const logoMapElems = (await Promise.all(
-    breachDomains.map((breachDomain) => {
-      return new Promise((resolve, reject) => {
-        const logoFilename = breachDomain.toLowerCase() + ".ico";
-        const logoPath = pathResolve(logoFolder, logoFilename);
-        if (existingLogos.includes(logoFilename)) {
-          resolve([breachDomain, `/logo_cache/${logoFilename}`]);
-          return;
-        }
-        get(
-          `https://icons.duckduckgo.com/ip3/${breachDomain}.ico`,
-          (response) => {
-            if (response.statusCode !== 200) {
-              resolve(null);
-              return;
-            }
-
-            const file = createWriteStream(logoPath);
-            response.pipe(file);
-            file.on("finish", () => {
-              file.close();
-              resolve([
-                breachDomain,
-                `/logo_cache/${breachDomain.toLowerCase()}.ico`,
-              ]);
-            });
-            file.on("error", (error) => reject(error));
+  async function fetchBreachIcons() {
+    const breachDomains = breaches
+      .map((breach) => breach.Domain)
+      .filter((breachDomain) => breachDomain.length > 0);
+    const logoFolder = pathResolve(
+      dirname(fileURLToPath(import.meta.url)),
+      "../../../../public/logo_cache/"
+    );
+    try {
+      await mkdir(logoFolder);
+    } catch {
+      // Do nothing; if the directory already exists, that's fine.
+    }
+    const existingLogos = await readdir(logoFolder);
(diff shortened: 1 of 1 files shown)


src/app/functions/server/getBreaches.ts
```diff
@@ -51,46 +51,53 @@ export async function getBreachIcons (breaches: Breach[]): Promise<LogoMap> {
...
-      const logoFilename = breachDomain.toLowerCase() + '.ico'
-      const logoPath = pathResolve(logoFolder, logoFilename)
-      if (existingLogos.includes(logoFilename)) {
-        resolve([breachDomain, `/logo_cache/${logoFilename}`])
-        return
-      }
-      get(`https://icons.duckduckgo.com/ip3/${breachDomain}.ico`, (response) => {
-        if (response.statusCode !== 200) {
-          resolve(null)
+  async function fetchBreachIcons () {
```
Vinnl, line 54: This function's contents is exactly the same as before, just indented. The part outside the braces is new.

src/app/functions/server/getBreaches.ts
```diff
@@ -51,46 +51,53 @@ export async function getBreachIcons (breaches: Breach[]): Promise<LogoMap> {
...
+    const breachDomains = breaches
+      .map(breach => breach.Domain)
+      .filter(breachDomain => breachDomain.length > 0)
```
pdehaan, line 57: Unrelated, but possibly interesting aside&hellip;
Several (15?) Domains have been breached 2+ times. Not sure if that means it's possible we'll request some site logos multiple times, or if the `async`/`await` Promises would resolve the first request and skip the other requests for the same logo.

```sh
curl --silent https://haveibeenpwned.com/api/v3/breaches | npx prettier --parser=json | grep '"Domain":' | sort | uniq -c | grep -Ev "\s+1\s+" | sort

   2     "Domain": "abandonia.com",
   2     "Domain": "adultfriendfinder.com",
   2     "Domain": "bell.ca",
   2     "Domain": "cardmafia.cc",
   2     "Domain": "cprewritten.net",
   2     "Domain": "data4marketers.com",
   2     "Domain": "flashflashrevolution.com",
   2     "Domain": "forum.btcsec.com",
   2     "Domain": "linkedin.com",
   2     "Domain": "mgmresorts.com",
   2     "Domain": "r2games.com",
   2     "Domain": "twitter.com",
   2     "Domain": "wishbone.io",
   2     "Domain": "zoosk.com",
   4     "Domain": "ogusers.com",
  27     "Domain": "",
```

Again, possibly not interesting, or a blocker, but probably more of a nit/microoptimization.

src/app/functions/server/getBreaches.ts
```diff
@@ -51,46 +51,53 @@ export async function getBreachIcons (breaches: Breach[]): Promise<LogoMap> {
...
+    const logoFolder = pathResolve(dirname(fileURLToPath(import.meta.url)), '../../../../public/logo_cache/')
+    try {
+      await mkdir(logoFolder)
```
pdehaan, line 60: Looking at https://nodejs.org/docs/latest-v18.x/api/fs.html#fspromisesmkdirpath-options I'm guessing we don't need to set `{ recursive: true }` since presumably the `/public/` folder is guaranteed here.
Any potential issues with a default `mode` of 777?

src/app/functions/server/getBreaches.ts
```diff
@@ -51,46 +51,53 @@ export async function getBreachIcons (breaches: Breach[]): Promise<LogoMap> {
...
-        const file = createWriteStream(logoPath)
-        response.pipe(file)
-        file.on('finish', () => {
-          file.close()
-          resolve([breachDomain, `/logo_cache/${breachDomain.toLowerCase()}.ico`])
+          const file = createWriteStream(logoPath)
+          response.pipe(file)
+          file.on('finish', () => {
+            file.close()
+            resolve([breachDomain, `/logo_cache/${breachDomain.toLowerCase()}.ico`])
```
pdehaan, line 84: Re: `/logo_cache/${breachDomain.toLowerCase()}.ico`

Not sure if we can use `logoFilename` if it's still in scope from L68 above:

```suggestion
            resolve([breachDomain, `/logo_cache/${logoFilename}`])
```

Don't feel too strongly (and I know this is just tweaking of existing code) but one less place we're hardcoding filenames+extensions.

rhelmer (approved):

src/app/functions/server/getBreaches.ts
Vinnl: Hmm yeah, we probably could, but again - avoiding yak shaving, unfortunately...

PR #3134, opened 2023-06-15
A couple of small fixes

Vinnl:
This fixes a couple of small things:

- Makes the HIBP link on the breach detail page clickable again ([example of bug](https://stage.firefoxmonitor.nonprod.cloudops.mozgcp.net/breach-details/JDGroup))
- Starts the built server on port 6060 as well (which aligns with the redirect URLs configured for FxA)
- Makes sure the breach icons are also vertically aligned on the dashboard, by applying styles that were no longer applied by using the new icon component.

### DIFF
--- a/package.json
+++ b/package.json
@@ -10,7 +10,7 @@
   "scripts": {
     "dev": "next dev --port=6060",
     "build": "next build",
-    "start": "next start",
+    "start": "next start --port=6060",
     "lint": "stylelint '**/*.scss' && prettier --check './src' && next lint",
     "test": "NODE_OPTIONS=--loader=testdouble c8 ava",
     "e2e": "playwright test src/e2e/",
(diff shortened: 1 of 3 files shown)


pdehaan:
We will also need to fix the invalid `og:image` and `twitter:image` URLs before we push to prod (since it breaks social sharing).
Although I can file a separate Jira ticket for that if we need to track it properly.

```sh
curl --silent https://stage.firefoxmonitor.nonprod.cloudops.mozgcp.net/ | npx prettier --parser=html  | grep -A1 -E '(property|name)="(og|twitter):image"'

      property="og:image"
      content="http://localhost:8000/images/og-image.webp"
--
      name="twitter:image"
      content="http://localhost:8000/images/og-image.webp"
```

flozia (approved):
Looks good!

PR #3135, opened 2023-06-15
MNTOR-1385 - request subscription scope, as some of our RPs are not yet allowed

rhelmer:
Request subscription scope when authenticating, as some of our RPs are not yet allowed.
This will fail early vs. just silently not working.

# References: 
Jira: MNTOR-1385

# How to test

Whether or not this works depends on the configuration of the environment on the FxA side, on investigating with SRE we found several client IDs registered for Fxa staging, most of which we are no longer using on the Monitor team.

SRE is adding the scope to our existing client IDs now, and we will work separately to deprecate any unused client IDs so there isn't confusing in the future. Additionally, SRE will enable this for production before we ship this feature (which is behind a flag).

### DIFF
--- a/src/app/api/auth/[...nextauth]/route.ts
+++ b/src/app/api/auth/[...nextauth]/route.ts
@@ -59,7 +59,7 @@ export const authOptions: AuthOptions = {
       authorization: {
         url: AppConstants.OAUTH_AUTHORIZATION_URI,
         params: {
-          scope: "profile",
+          scope: "profile https://identity.mozilla.com/account/subscriptions",
           access_type: "offline",
           action: "email",
           prompt: "login",


src/app/api/auth/[...nextauth]/route.ts
```diff
@@ -59,7 +59,7 @@ export const authOptions: AuthOptions = {
       authorization: {
         url: AppConstants.OAUTH_AUTHORIZATION_URI,
         params: {
-          scope: "profile",
+          scope: "profile https://identity.mozilla.com/account/subscriptions",
```
Comment (pdehaan) [inline src/app/api/auth/[...nextauth]/route.ts:62]: Q: Should this be hardcoded URL, or should it be from an env var so we can toggle between stage/prod based on deployment?

Comment (rhelmer) [reply src/app/api/auth/[...nextauth]/route.ts:62]: It's actually the name of the scope (similar to `profile`) which is admittedly somewhat confusing :) I think it's being used as a URI not URL in this case.

Comment (rhelmer) [reply src/app/api/auth/[...nextauth]/route.ts:62]: https://mozilla.github.io/ecosystem-platform/api#tag/OAuth-Server-API-Overview/operation/postIntrospect for reference

rhelmer:
BTW failing playwright test is expected as we're awaiting the `subscribers` scope to be allowed on the server-side:

image

Vinnl (approved):
I'm assuming the config will be added to FxA shortly? Otherwise it could be annoying that the Playwright tests will keep failing.

rhelmer:
> I'm assuming the config will be added to FxA shortly? Otherwise it could be annoying that the Playwright tests will keep failing.

Yes, SRE is working on it now - I will hold off on landing this until that's in-place. Thanks!

rhelmer:
SRE has enabled this for our stage client IDs, and the Playwright tests pass now (which makes sense because they use staging FxA) so it should presumably work from our dev/stage servers now too.

PR #3138, opened 2023-06-16
config: send more debug info from dev/heroku envs

rhelmer:
# Description

This is a small config change for the Heroku dev env to send in more performance and log debug data for Sentry. It causes Heroku (and local development, if enabled with the `SENTRY_DSN` and `NEXT_PUBLIC_SENTRY_DSN` env vars) to log to the console and also set the traces sample rate to 100%.

Production (and any other environments such as stage) will have the trace sample rate set to 1% and disabled debug logging.

### DIFF
--- a/sentry.client.config.ts
+++ b/sentry.client.config.ts
@@ -12,17 +12,22 @@ Sentry.init({
   dsn: process.env.NEXT_PUBLIC_SENTRY_DSN,
 
   // Adjust this value in production, or use tracesSampler for greater control
-  tracesSampleRate: 1,
+  tracesSampleRate: ["development", "heroku"].includes(process.env.NODE_ENV)
+    ? 1.0
+    : 0.1,
 
   // Setting this option to true will print useful information to the console while you're setting up Sentry.
-  debug: true,
+  debug: ["development", "heroku"].includes(process.env.NODE_ENV),
 
   replaysOnErrorSampleRate: 1.0,
 
   // This sets the sample rate to be 10%. You may want this to be 100% while
   // in development and sample at a lower rate in production
-  replaysSessionSampleRate:
-    process.env.NEXT_PUBLIC_NODE_ENV === "development" ? 1.0 : 0.1,
+  replaysSessionSampleRate: ["development", "heroku"].includes(
+    process.env.NODE_ENV
+  )
+    ? 1.0
+    : 0.1,
 
   // You can remove this option if you're not planning to use the Sentry Session Replay feature:
   integrations: [
--- a/sentry.client.config.ts
+++ b/sentry.client.config.ts
@@ -24,7 +24,7 @@ Sentry.init({
   // This sets the sample rate to be 10%. You may want this to be 100% while
   // in development and sample at a lower rate in production
   replaysSessionSampleRate: ["development", "heroku"].includes(
-    process.env.NODE_ENV
+    process.env.NEXT_PUBLIC_NODE_ENV
   )
     ? 1.0
     : 0.1,
--- a/sentry.client.config.ts
+++ b/sentry.client.config.ts
@@ -24,7 +24,7 @@ Sentry.init({
   // This sets the sample rate to be 10%. You may want this to be 100% while
   // in development and sample at a lower rate in production
   replaysSessionSampleRate: ["development", "heroku"].includes(
-    process.env.NEXT_PUBLIC_NODE_ENV
+    process.env.NODE_ENV
   )
     ? 1.0
     : 0.1,
@@ -33,8 +33,8 @@ Sentry.init({
   integrations: [
     new Sentry.Replay({
       // Additional Replay configuration goes in here, for example:
-      maskAllText: process.env.NEXT_PUBLIC_NODE_ENV === "development",
-      blockAllMedia: process.env.NEXT_PUBLIC_NODE_ENV === "development",
+      maskAllText: process.env.NODE_ENV === "development",
+      blockAllMedia: process.env.NODE_ENV === "development",
     }),
   ],
 });
--- a/sentry.client.config.ts
+++ b/sentry.client.config.ts
@@ -17,7 +17,7 @@ Sentry.init({
     : 0.1,
 
   // Setting this option to true will print useful information to the console while you're setting up Sentry.
-  debug: ["development", "heroku"].includes(process.env.NODE_ENV),
+  debug: false,
 
   replaysOnErrorSampleRate: 1.0,
 
--- a/sentry.server.config.ts
+++ b/sentry.server.config.ts
@@ -12,8 +12,10 @@ Sentry.init({
   dsn: process.env.SENTRY_DSN,
 
   // Adjust this value in production, or use tracesSampler for greater control
-  tracesSampleRate: 1,
+  tracesSampleRate: ["development", "heroku"].includes(process.env.NODE_ENV)
+    ? 1.0
+    : 0.1,
 
   // Setting this option to true will print useful information to the console while you're setting up Sentry.
-  debug: process.env.NODE_ENV === "development",
+  debug: ["development", "heroku"].includes(process.env.NODE_ENV),
 });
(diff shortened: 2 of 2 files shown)


sentry.client.config.ts
```diff
@@ -12,17 +12,22 @@ Sentry.init({
   dsn: process.env.NEXT_PUBLIC_SENTRY_DSN,
 
   // Adjust this value in production, or use tracesSampler for greater control
-  tracesSampleRate: 1,
+  tracesSampleRate: ["development", "heroku"].includes(process.env.NODE_ENV)
```
pdehaan, line 15: Q: Should we extract this to some shared `isDev()`-esque utility function since we're using the pattern at least 5&times;:

```sh
git grep -nP "\bheroku\b" -- ':!*.md' ':!.circleci/*' ':!.env*' ':!src/db/migrations/*' | cat

next.config.js:124:    const noindexEnvs = ["dev", "development", "heroku", "stage"];
sentry.client.config.ts:15:  tracesSampleRate: ["development", "heroku"].includes(process.env.NODE_ENV)
sentry.client.config.ts:20:  debug: ["development", "heroku"].includes(process.env.NODE_ENV),
sentry.client.config.ts:26:  replaysSessionSampleRate: ["development", "heroku"].includes(
sentry.server.config.ts:15:  tracesSampleRate: ["development", "heroku"].includes(process.env.NODE_ENV)
sentry.server.config.ts:20:  debug: ["development", "heroku"].includes(process.env.NODE_ENV),
src/appConstants.js:68:// @ts-ignore TS thinks NODE_ENV can't be "heroku". Shut up TS, you're drunk!
src/appConstants.js:69:if (!process.env.SERVER_URL && (process.env.NODE_ENV) === 'heroku') {
src/controllers/dockerflow.js:26:   if (AppConstants.NODE_ENV === 'heroku') {
src/db/knexfile.js:10:if (NODE_ENV === 'heroku') {
src/middleware/noSearchEngineIndex.js:8:const noindexEnvs = ['dev', 'heroku', 'stage']
```

sentry.client.config.ts
```diff
@@ -12,17 +12,22 @@ Sentry.init({
...
+  debug: ["development", "heroku"].includes(process.env.NODE_ENV),
 
   replaysOnErrorSampleRate: 1.0,
 
   // This sets the sample rate to be 10%. You may want this to be 100% while
   // in development and sample at a lower rate in production
-  replaysSessionSampleRate:
-    process.env.NEXT_PUBLIC_NODE_ENV === "development" ? 1.0 : 0.1,
+  replaysSessionSampleRate: ["development", "heroku"].includes(
+    process.env.NODE_ENV
```
pdehaan, line 27: I assume this is OK converting from `NEXT_PUBLIC_NODE_ENV` to `NODE_ENV`?

rhelmer: No that's not right, thanks for catching.

sentry.client.config.ts
rhelmer: Yes I think it's worth considering, but I would note that in the above examples:

- we sometimes want to include stage, but not always
- we're moving from `dev` to the more standard `development`
- most (but not all) of the code in `./src/` will be replaced by the NextJS app in `./src/app`

NextJS suggests this for environment names:
https://nextjs.org/docs/messages/non-standard-node-env

> The greater React ecosystem treats NODE_ENV as a convention, only permitting three (3) values:
>
>   production: When your application is built with next build
>   development: When your application is run with next dev
>   test: When your application is being tested (e.g. jest)

Regarding staging:

> If you need the concept of different environments in your application, e.g. staging, you should use a different environment variable name like APP_ENV.

So I would prefer to clean this up a bit and come back to it, I'll file an issue in our backlog.

rhelmer:
After trying this locally for a bit, I think the Sentry debug logs are too noisy to have on normally - on second though I think we should disable it generally.

sentry.client.config.ts
rhelmer: After trying this for a bit, this makes the logs way too noisy even in dev. I'm going to just set it to `false` by default, thanks for bringing this up though! Will follow up in the other tickets.

pdehaan, line 15: Q: Should this be `process.env.NODE_ENV` Or should it be `process.env.NEXT_PUBLIC_NODE_ENV` like below on L27?

sentry.server.config.ts
```diff
@@ -12,8 +12,10 @@ Sentry.init({
   dsn: process.env.SENTRY_DSN,
 
   // Adjust this value in production, or use tracesSampler for greater control
-  tracesSampleRate: 1,
+  tracesSampleRate: ["development", "heroku"].includes(process.env.NODE_ENV)
```
pdehaan, line 15: I guess same comment as above where now we're juggling `NODE_ENV` _and_ `NEXT_PUBLIC_NODE_ENV`, but I'm not 100% clear on whether one inherently sets the other or if both need to be explicitly set.

Or is the logic that client-side code can use the Node-esque `process.env.NEXT_PUBLIC_NODE_ENV` and Next.js will cleverly shim the value into the client code, but the server code can just use `process.env.*` like it always has since it has access to env vars.

Vinnl (approved):

sentry.client.config.ts
pdehaan: https://nextjs.org/docs/pages/building-your-application/configuring/environment-variables#bundling-environment-variables-for-the-browser

<i>

> **Non-`NEXT_PUBLIC_` environment variables are only available in the Node.js environment, meaning they aren't accessible to the browser (the client runs in a different environment).**
>
> In order to make the value of an environment variable accessible in the browser, Next.js can "inline" a value, at build time, into the js bundle that is delivered to the client, replacing all references to `process.env.[variable]` with a hard-coded value. To tell it to do this, you just have to prefix the variable with `NEXT_PUBLIC_`.
>
> ```ini
> NEXT_PUBLIC_ANALYTICS_ID=abcdefghijk
> ```
>
> This will tell Next.js to replace all references to `process.env.NEXT_PUBLIC_ANALYTICS_ID` in the Node.js environment with the value from the environment in which you run next build, allowing you to use it anywhere in your code. It will be inlined into any JavaScript sent to the browser.

</i>

Only current instances of `.NEXT_PUBLIC_` I could find currently were all in this file:
```sh
git grep -n "\.NEXT_PUBLIC_" | cat

sentry.client.config.ts:12:  dsn: process.env.NEXT_PUBLIC_SENTRY_DSN,
sentry.client.config.ts:27:    process.env.NEXT_PUBLIC_NODE_ENV
sentry.client.config.ts:36:      maskAllText: process.env.NEXT_PUBLIC_NODE_ENV === "development",
sentry.client.config.ts:37:      blockAllMedia: process.env.NEXT_PUBLIC_NODE_ENV === "development",
```
&hellip; and this looked like the only instance of an env var that WASN'T `NEXT_PUBLIC_` prefixed:
```sh
git grep -n "process\.env\." sentry.client.config.ts | grep -v "\.NEXT_PUBLIC_"

sentry.client.config.ts:15:  tracesSampleRate: ["development", "heroku"].includes(process.env.NODE_ENV)
```

sentry.server.config.ts
rhelmer: I'm a little confused by this too honestly... https://nextjs.org/docs/pages/building-your-application/configuring/environment-variables explains it but I'm not clear on why this would work.

rhelmer: Well, so I do know that the idea is that `NEXT_PUBLIC_*` is a signal to the bundler that it's OK to include this var in front-end code (in other words, it does not contains secrets and can be public in the web page source).

However I mean I'm not clear on why `process.env.NODE_ENV` is an exception to this rule, is that what you mean in https://github.com/mozilla/blurts-server/pull/3138#discussion_r1238304742 @Vinnl ?

Vinnl: Yeah, for some reason I can't find it in the docs, but I do recall `NODE_ENV` being the one exception to the `NEXT_PUBLIC_`. Which makes sense, since the prefix is there to avoid accidentally leaking secrets - but if you've got secrets in `NODE_ENV`, you'd have bigger problems :sweat_smile:

PR #3139, opened 2023-06-16
MNTOR-1689 - Chart component

codemist:
# References: 
Jira: MNTOR-1689

# Screenshot (if applicable)
image

# How to test

Check the Chart component on Storybook by running `npm run storybook`


src/app/components/client/stories/Chart.stories.ts
```diff
@@ -0,0 +1,41 @@
...
+const data: any[] = [
+  ['Home address', 11],
+  ['Family members', 12],
+  ['Contact Info', 1],
+  ['Full name', 5],
+  ['Other', 2]
+];
+
+const labels: string[] = data.map(([label]) => label);
+const numbers: number[] = data.map(([, number]) => number);
```
pdehaan, line 32: Since the above `data` structure is basically a `Map()`, you could also do something like this; although not convinced it's any better seems like the same amount of loops/variables/complexity:

```js
const data = new Map([
  ['Home address', 11],
  ['Family members', 12],
  ['Contact Info', 1],
  ['Full name', 5],
  ['Other', 2]
]);
const labels = [...data.keys()] // ['Home address', 'Family members', 'Contact Info', 'Full name', 'Other']
const numbers = [...data.values()]; // [11, 12, 1, 5, 2]
```

src/app/components/client/stories/Chart.stories.ts
```diff
@@ -0,0 +1,41 @@
...
+
+export const FixedExposures: Story = {
+  args: {
+    labels: labels,
+    data: numbers,
+    backgroundColors: purpleColors.slice(0, data.length),
```
pdehaan, line 38: Q: How would this work if we had like 20 fixed exposures and `purpleColors.length` is only 8? Does it loop the colors?

src/app/components/client/stories/Chart.stories.ts
```diff
@@ -0,0 +1,41 @@
...
+    totalExposures: 309,
+  },
+};
```
pdehaan, line 41: nit: add blank line

Vinnl (commented):
Looking at [this code example](https://react-spectrum.adobe.com/react-aria/useProgressBar.html#circular-example) (not because I think `useProgressBar` would make sense here, but because it's drawing a similar element), it looks like it might not be too hard to draw a donut chart using SVG ourselves, with `<circle>` elements with `stroke-dasharray` and `stroke-dashoffset`. So I'm wondering if that might not be preferable to adding two new dependencies with their separate maintainers to our supply chain, and 66+ kB to our bundle size. What do you think? I'd also be happy to prototype that myself if you prefer.

One other thing that struck my eye is that the entire chart is currently exposed to the accessibility tree as a single image with no text alternative - I think we should be able to provide the numbers as text?

Oh, and I just noticed that clicking the elements in the legend removes/re-adds those numbers to the donut chart, without feedback that that's what it's doing, which is a bit confusing.

Vinnl:
@codemist OK, I prototyped a proposal to replace chart.js with an inline-generated SVG: https://github.com/mozilla/blurts-server/pull/3179 That also ensures that the chart contents is communicated properly to screen readers (at least, as indicated by [the Accessibility Inspector](https://firefox-source-docs.mozilla.org/devtools-user/accessibility_inspector/index.html)), by turning the legend (which contains all the data) into a table, and adding a label and role for the SVG. Let me know what you think.

Vinnl (approved):

Vinnl:
@codemist Any objections if I merge this?

PR #3137, opened 2023-06-16
Do not execute scripts that render relevant DOM elements only after hydration

flozia:
# References

Jira: [MNTOR-1861](https://mozilla-hub.atlassian.net/browse/MNTOR-1861)

# Description

I have already attempted to address this issue for the landing page with https://github.com/mozilla/blurts-server/pull/3126, which seems to be error-prone. I suspect this happens because the preloaded scripts are causing hydration errors that are thrown by Next, since DOM elements have been altered when hydration happens. Due to the way we served HTML string literal templates and created additional elements client-side with JS this method is conflicting with the site ported (mostly as is) from Express to Next. This PR would be more of a band-aid until we place the components with proper React components, but due to the changes really going against best practices, we can also decide to accept the delayed rendering of UI elements or discuss another possible solution.

# Screenshot

**Before**

https://github.com/mozilla/blurts-server/assets/13835474/f7d9db16-35c2-4c39-9e4d-c0cddb2ebb1d

**After**

https://github.com/mozilla/blurts-server/assets/13835474/e92b155c-5127-4bc0-b136-1e14464ceab9

# How to test

- Visit `/`
- Visit `/users/breaches`

PR #3142, opened 2023-06-19
Add API for fetching modal component as HTML

flozia:
# References

Jira: [MNTOR-1862](https://mozilla-hub.atlassian.net/browse/MNTOR-1862)

# Description

This PR adds an API that only serves the relevant modal content. Right now we are fetching the `addEmail` modal content as a page including scripts, styles, and fonts which are not needed.

| Content size before | Content size after |
| --- | --- |
| ~1 MB | ~0.001 MB |

# How to test

- Visit `/user/settings`
- Click “Add email address”


public/nextjs_migration/client/css/addEmail.css
```diff
@@ -0,0 +1,28 @@
+[data-partial="addEmail"] {
```
pdehaan, line 1: nit: Does this file need to be .css or can it be a .scss file?
The nesting in scss would be nice here since everything seems to be scoped under `[data-partial="addEmail"]`.

```sh
git ls-files *.scss

src/app/(proper_react)/MobileShell.module.scss
src/app/(proper_react)/Shell.module.scss
src/app/(proper_react)/redesign/(authenticated)/user/dashboard/page.module.scss
src/app/components/client/ExposureCard.module.scss
src/app/components/server/Icons.module.scss
src/app/components/server/StatusPill.module.scss
src/app/components/server/button.module.scss
src/app/not-found.module.scss
src/app/tokens.scss
```

src/app/(nextjs_migration)/components/server/AddEmailDialog.tsx
```diff
@@ -2,23 +2,25 @@
...
 
   return (
     <>
+      {/* Styles need to be included manually when fetching the component via API */}
+      {/* eslint-disable-next-line @next/next/no-css-tags */}
+      <link rel="stylesheet" href="/nextjs_migration/client/css/addEmail.css" />
       <header>
         <button className="close"></button>
-        <Image src={cloudImage} alt="" />
+        {/* We can’t use next/image when we featch the component via API */}
```
pdehaan, line 21: typo
```suggestion
        {/* We can’t use next/image when we fetch the component via API */}
```

src/app/api/v1/[...dialog]/route.ts
```diff
@@ -0,0 +1,35 @@
...
+  let dialogContentString = ''
+
+  switch (dialogType) {
+    case 'addEmail':
+      dialogContentString = await getComponentAsString({
+        component: AddEmailDialog()
+      });
+      break;
+    default:
+      throw new Error(`No dialog found for: ${dialogUrl}`);
```
Comment (pdehaan) [inline src/app/api/v1/[...dialog]/route.ts:23]: Q: who/where does this throw to? I'm assuming the Next routing system, but wasn't sure if this whole block should be in the `try..catch` block around L30-34 so we catch locally and use `NextResponse` w/ status 500.

pdehaan:
Mysterious Playwright Test Action log, or does Playwright not work currently w/ Next.js?

```
[WebServer] Failed to compile.

[WebServer] src/app/api/auth/[...nextauth]/route.ts
Type error: Route "src/app/api/auth/[...nextauth]/route.ts" does not match the required types of a Next.js Route.
  "authOptions" is not a valid Route export field.

[WebServer] - warn You are using a non-standard "NODE_ENV" value in your environment. This creates inconsistencies in the project and is strongly advised against. Read more: https://nextjs.org/docs/messages/non-standard-node-env
[WebServer] - warn You are using a non-standard "NODE_ENV" value in your environment. This creates inconsistencies in the project and is strongly advised against. Read more: https://nextjs.org/docs/messages/non-standard-node-env
[WebServer] Error: Could not find a production build in the '/home/runner/work/blurts-server/blurts-server/.next' directory. Try building your app with 'next build' before starting the production server. https://nextjs.org/docs/messages/production-start-no-build-id
    at NextNodeServer.getBuildId (/home/runner/work/blurts-server/blurts-server/node_modules/next/dist/server/next-server.js:345:23)
    at new Server (/home/runner/work/blurts-server/blurts-server/node_modules/next/dist/server/base-server.js:151:29)
    at new NextNodeServer (/home/runner/work/blurts-server/blurts-server/node_modules/next/dist/server/next-server.js:172:9)
    at NextServer.createServer (/home/runner/work/blurts-server/blurts-server/node_modules/next/dist/server/next.js:1[79](https://github.com/mozilla/blurts-server/actions/runs/5312336095/jobs/9617613496?pr=3142#step:9:80):24)
    at async /home/runner/work/blurts-server/blurts-server/node_modules/next/dist/server/next.js:200:31
    at async NextServer.prepare (/home/runner/work/blurts-server/blurts-server/node_modules/next/dist/server/next.js:160:24)
    at async Server.<anonymous> (/home/runner/work/blurts-server/blurts-server/node_modules/next/dist/server/lib/render-server.js:123:17) {
  type: 'Error'
}
Error: Process completed with exit code 1.
```

public/nextjs_migration/client/css/addEmail.css
flozia: Unfortunately, this has to be loaded in plain CSS for now: https://github.com/mozilla/blurts-server/pull/3142/files#diff-0ab67d22b51933639a76489e45aea51c2a6881b9deb66458b49513df23a52e50R16-R18.

src/app/api/v1/[...dialog]/route.ts
Comment (flozia) [reply src/app/api/v1/[...dialog]/route.ts]: Yes, you are right. That’s a good point — done in [ab8c815](https://github.com/mozilla/blurts-server/pull/3142/commits/ab8c8152fbe1f6a82ab6ccb8e916a1bcecb24060).

src/app/api/auth/[...nextauth]/route.ts
```diff
@@ -2,208 +2,8 @@
...
-    signIn(message) {
-      log.debug("fxa-confirmed-profile-data", message.user);
-    },
-    signOut(message) {
-      log.debug("logout", message.token.email ?? undefined);
-    },
-  },
-};
+import NextAuth from "next-auth";
+import { authOptions } from "../../utils/auth"
```
Comment (flozia) [inline src/app/api/auth/[...nextauth]/route.ts:6]: So it seems like the test failure and error just really are what they say:
> Type error: Route "src/app/api/auth/[...nextauth]/route.ts" does not match the required types of a Next.js Route. "authOptions" is not a valid Route export field.

Moving `authOptions` into a separate file and not exporting it directly from the route resolves the issue. Not sure why this only started to be an issue now and not in recent PRs.

public/nextjs_migration/client/js/dialog.js
```diff
@@ -31,7 +31,7 @@ function handleEvent (e) {
       openDialog(e.target.href)
       break
     case e.target.matches('button[data-dialog]'):
-      openDialog(`dialog/${e.target.dataset.dialog}`)
+      openDialog(`/api/v1/dialog/${e.target.dataset.dialog}`)
```
Vinnl, line 34: Can we keep it at the present location? I think back-end will want to scope `/api/v1/` to APIs they have clear API contracts on, so we might be getting in their way if we pollute their namespace. (Also, it's not an API in the commonly-understood sense (i.e. returning JSON) anyway.)

src/app/api/auth/[...nextauth]/route.ts
Comment (Vinnl) [reply src/app/api/auth/[...nextauth]/route.ts:6]: Really weird that that didn't turn up before! /cc @rhelmer for potential merge conflicts in case his auth work also modifies `authOptions`.

src/app/functions/server/getComponentAsString.ts
Vinnl, line 1: Same comments as in https://github.com/mozilla/blurts-server/pull/3137/files/c8ea7ea42ffa2b27570bc64852bc05630637c391#diff-db5e3699e57ca72943289060791005d1f178ad82189ef0a714419a3a11645b55.

src/app/api/v1/dialog/[...type]/route.ts
Comment (Vinnl) [inline src/app/api/v1/dialog/[...type]/route.ts:1]: So I'd keep this page in the old location, just renaming `page.tsx` to `route.ts`.

Also, depending on the contents of `<AddEmailDialog>`, we could also just directly return that as a string, rather than running it through React? But then again, if it works now, why bother.

Vinnl (approved):
Another excellent find - I wonder why I didn't go the API route in the first place. I'm assuming there was some snag I hit that you did manage to deal with, so nice job :+1:

public/nextjs_migration/client/js/dialog.js
rhelmer: I don't think it'll be in the way, it is a little odd as an API though I agree (there are some HTML-over-the-wire frameworks out there fwiw)

rhelmer (approved):

src/app/functions/server/getComponentAsString.ts
flozia: Addressed the comment there and also here: https://github.com/mozilla/blurts-server/pull/3142/commits/16babedd6a8b49baf83259176c9a89aee6e095a8.

src/app/api/v1/dialog/[...type]/route.ts
Comment (flozia) [reply src/app/api/v1/dialog/[...type]/route.ts:1]: Keeping the page in the same location as a route is a good idea — moved it back in https://github.com/mozilla/blurts-server/pull/3142/commits/2784e2d1cf93da98ce035c9c67aa18181da56867.

For this component, we actually could have it as a string literal. Since this is not the only instance we are using `getComponentAsString` (we are using it for <BreachLogo> [here](https://github.com/mozilla/blurts-server/blob/main/src/app/(nextjs_migration)/(authenticated)/user/breaches/page.tsx#L168)) I’d like to keep it the way it is if that is OK.

public/nextjs_migration/client/js/dialog.js
flozia: > Can we keep it at the present location? […]

See https://github.com/mozilla/blurts-server/pull/3142#discussion_r1235999490.

PR #3144, opened 2023-06-19
Port and include script for mobile nav button

flozia:
# References: 
Jira: [MNTOR-1855](https://mozilla-hub.atlassian.net/browse/MNTOR-1855)

# Description

Fixes the mobile nav button.


Vinnl (approved):
Oh crap, we should probably do a pass over all the files imported from `src/client/index.js` to see if they're included in the new site. Thanks for fixing this one.

PR #3145, opened 2023-06-19
Make sidebar navigation scrollable

flozia:
# References: 
Jira: [MNTOR-1856](https://mozilla-hub.atlassian.net/browse/MNTOR-1856)


src/client/css/nav.css
```diff
@@ -19,6 +19,7 @@ nav.site-nav {
       #0000 97%
     )
     1;
+  overflow: scroll;
```
Vinnl, line 22: Juuust in case we inadvertently get horizontal overflow as well.

```suggestion
  overflow-y: scroll;
```

Vinnl (approved):
Heh, this is an issue on prod already too. But I assume that this works (haven't tested it myself, sorry - getting through my backlog).

PR #3146, opened 2023-06-20
Do not ignore formatting in src/app

flozia:
# Description

~There are a couple of formatting inconsistencies that we do not lint and enforce. These rules in this PR are just a proposal and the intention is to start a conversation around additional rules we would like to follow.~

Edit: There are a couple of formatting inconsistencies that we do not enforce for `src/app`.

### DIFF
--- a/.git-blame-ignore-revs
+++ b/.git-blame-ignore-revs
@@ -7,3 +7,5 @@ ed4ecc5a57bf248075af82b8238baf121d513b79
 # conform to stylelint formatting: kebab-case, single-quotes, modern color syntax, etc
 3d862f0d4c43f0e97354ef785b44061928e7e1e7
 8767bbdc0b9059027489152e03571f58a5b7a8b0
+# fix: Apply prettier for src/app
+9cf79b645181f0e37e1acb066e3d8c4ee0ad2fb8
--- a/src/app/(nextjs_migration)/(guest)/page.tsx
+++ b/src/app/(nextjs_migration)/(guest)/page.tsx
@@ -18,7 +18,11 @@ export default async function Home() {
 
   return (
     <div data-partial="landing">
-      <script type="module" src="/nextjs_migration/client/js/transitionObserver.js" async></script>
+      <script
+        type="module"
+        src="/nextjs_migration/client/js/transitionObserver.js"
+        async
+      ></script>
       <Script type="module" src="/nextjs_migration/client/js/landing.js" />
       <section className="hero">
         <div>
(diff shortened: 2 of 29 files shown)


Vinnl:
Formatting for newer code is done by Prettier, which should take care of these settings. So I expect that where you're seeing this, we're simply not applying Prettier for now? Which files did you encounter this in - should we enable Prettier there too?

flozia:
I see. What led me to add those was that I don’t see them applied for `src/app/components/client/*`. Just saw `.prettierignore` and this makes sense now.

Vinnl:
Oh oops yep, I did not expect to have that listed there. I don't mind a bulk formatting change _too_ much, as long as we try to apply it in a single commit, and we [add that commit to `.git-blame-ignore-revs`](https://docs.github.com/en/repositories/working-with-files/using-files/viewing-a-file#ignore-commits-in-the-blame-view).

.git-blame-ignore-revs
```diff
@@ -6,4 +6,6 @@ ed4ecc5a57bf248075af82b8238baf121d513b79
 7a16c1810755a6dafeed3ca83dd5c7071829966e
 # conform to stylelint formatting: kebab-case, single-quotes, modern color syntax, etc
 3d862f0d4c43f0e97354ef785b44061928e7e1e7
-8767bbdc0b9059027489152e03571f58a5b7a8b0
\ No newline at end of file
+8767bbdc0b9059027489152e03571f58a5b7a8b0
+# fix: Apply prettier for scr/app
```
pdehaan, line 10:
```suggestion
# fix: Apply prettier for src/app
```

src/app/(nextjs_migration)/(guest)/page.tsx
```diff
@@ -18,7 +18,11 @@ export default async function Home() {
 
   return (
     <div data-partial="landing">
-      <script type="module" src="/nextjs_migration/client/js/transitionObserver.js" async></script>
+      <script
+        type="module"
+        src="/nextjs_migration/client/js/transitionObserver.js"
+        async
+      ></script>
```
pdehaan, line 25: Personal curiosity, why use `<script>` here versus Next's `<Script>` like we do below on L26?
Or is it due to the `async` prop? And if so, could we use `<Script strategy="lazyOnload".../>` to do the same thing? https://nextjs.org/docs/app/api-reference/components/script#strategy

flozia: This particular `<script>` tag was converted back to `<Script>`. With our current setup for the in-between state in which we mostly ported the site to be served by Next.js instead of Express as a first step. Otherwise, some of the scripts that manipulate DOM in a non-React way are executing a bit late — only after hydration. More context here: https://github.com/mozilla/blurts-server/pull/3137.

rhelmer (approved):

PR #3147, opened 2023-06-20
MNTOR-949/check subscriptions on dashboard

rhelmer:
# References: 
Jira: MNTOR-949

# Description

Adds/removes subscription button based on users current subscription status.

The current approach of checking the FxA token for `subscription` claims seems to work fine locally, but not on our dev server. The current approach also requires refreshing the token, where this does not.

# How to test

This is the same set of steps to test as MNTOR-949, but with the added benefit that it does not require the user to log out/in again, and it should work on our dev/stage servers and not just locally.

1. Create new FxA user and log in to Monitor
2. Verify that "Subscribe to Premium" button appears on dashboard
3. Click "Subscribe to Premium" and subscribe
4. Verify that "Subscribe to Premium" button does *not* appear on dashboard

### DIFF
--- a/.env-dist
+++ b/.env-dist
@@ -38,12 +38,15 @@ SES_NOTIFICATION_LOG_ONLY=
 # leave FXA_ENABLED empty to disable FXA
 FXA_ENABLED=
 FXA_SETTINGS_URL=https://accounts.stage.mozaws.net/settings
+FXA_SUBSCRIPTIONS_URL=https://accounts.stage.mozaws.net/subscriptions/products/prod_NErZh679W62lai?plan=price_1MUNq0Kb9q6OnNsL4BoJgepf
+
 OAUTH_CLIENT_ID=edd29a80019d61a1
 OAUTH_CLIENT_SECRET=get-this-from-groovecoder-or-fxmonitor-engineering
 OAUTH_AUTHORIZATION_URI=https://oauth.stage.mozaws.net/v1/authorization
 OAUTH_PROFILE_URI=https://profile.stage.mozaws.net/v1/profile
 OAUTH_TOKEN_URI=https://oauth.stage.mozaws.net/v1/token
 OAUTH_ACCOUNT_URI = "https://oauth.accounts.firefox.com/v1"
+OAUTH_API_URI="https://api-accounts.stage.mozaws.net/v1"
 
 # HIBP API for breach data
 # How many seconds to wait before refreshing upstream breach data from HIBP
@@ -120,4 +123,5 @@ E2E_TEST_ACCOUNT_PASSWORD=
 # Monitor Premium features
 PREMIUM_ENABLED=
 # Link to start user on the subscription process. PREMIUM_ENABLED must be set to `true`.
-SUBSCRIBE_PREMIUM_URL=
+PREMIUM_PRODUCT_ID=prod_NErZh679W62lai
+PREMIUM_PLAN_ID_US=price_1MUNq0Kb9q6OnNsL4BoJgepf
--- a/.env-dist
+++ b/.env-dist
@@ -38,7 +38,7 @@ SES_NOTIFICATION_LOG_ONLY=
 # leave FXA_ENABLED empty to disable FXA
 FXA_ENABLED=
 FXA_SETTINGS_URL=https://accounts.stage.mozaws.net/settings
-FXA_SUBSCRIPTIONS_URL=https://accounts.stage.mozaws.net/subscriptions/products/prod_NErZh679W62lai?plan=price_1MUNq0Kb9q6OnNsL4BoJgepf
+FXA_SUBSCRIPTIONS_URL=https://accounts.stage.mozaws.net/subscriptions
 
 OAUTH_CLIENT_ID=edd29a80019d61a1
 OAUTH_CLIENT_SECRET=get-this-from-groovecoder-or-fxmonitor-engineering
--- a/src/app/(nextjs_migration)/(authenticated)/user/breaches/page.tsx
+++ b/src/app/(nextjs_migration)/(authenticated)/user/breaches/page.tsx
@@ -71,6 +71,50 @@ export default async function UserBreaches() {
     user: session!.user as any,
   });
 
+  type FxaSubscriptionResponse = {
+    subscriptions: Array<{
+      product_id: string;
+      plan_id: string;
+      status: "active";
+    }>;
+  };
+
+  // Fetch list of subscriptions.
+  let subscriptions;
+  const bearerToken = session?.user.subscriber?.fxa_access_token;
+  if (bearerToken) {
+    const result = await fetch(
+      `${process.env.OAUTH_API_URI}/oauth/mozilla-subscriptions/customer/billing-and-subscriptions`,
+      {
+        headers: {
+          "Content-Type": "application/json",
+          authorization: `Bearer ${bearerToken}`,
+        },
+      }
+    );
+    if (result.ok) {
+      subscriptions = await result.json();
+    }
+  } else {
+    console.error("User has no bearer token");
+  }
+
+  const isUserSubscribed = (subscriptions: FxaSubscriptionResponse) => {
+    if (!subscriptions) {
+      return false;
+    }
+    for (const subscription of subscriptions.subscriptions) {
+      if (
+        subscription.product_id === process.env.PREMIUM_PRODUCT_ID &&
+        subscription.plan_id === process.env.PREMIUM_PLAN_ID_US &&
+        subscription.status === "active"
+      ) {
+        return true;
+      }
+    }
+    return false;
+  };
+
   return (
     <>
       {/* These scripts predate the use of React and thus shouldn’t wait for
@@ -96,11 +140,11 @@ export default async function UserBreaches() {
 
       <main data-partial="breaches">
         {process.env.PREMIUM_ENABLED === "true" &&
-        !session?.user.fxa?.subscriptions?.includes("monitor") ? (
+        !isUserSubscribed(subscriptions) ? (
           <section>
             <a
               className="button primary"
-              href={process.env.SUBSCRIBE_PREMIUM_URL}
+              href={`${process.env.FXA_SUBSCRIPTIONS_URL}/products/${process.env.PREMIUM_PRODUCT_ID}?plan=${process.env.PREMIUM_PLAN_ID_US}`}
             >
               Subscribe to Premium
             </a>
--- a/src/app/api/v1/premium/route.ts
+++ b/src/app/api/v1/premium/route.ts
@@ -10,20 +10,13 @@ export async function GET(req: NextRequest) {
     // NOTE: the email address passed here cannot
     // be trusted, we must also check that the
     // `subscriptions` claim on the FxA token contains
-    // "monitor".
+    // "monitor", or call the auth server API to confirm.
 
     const { searchParams } = new URL(req.url);
     const _email = searchParams.get("email");
 
-    // TODO: The user either needs to be signed in again, or the
-    // JWT from FxA refreshed, in order to read the latest
-    // subscriptions.
-    //
-    // For now, redirect to the built-in sign-out, which will then
-    // send the user to the dashboard.
-
     return NextResponse.redirect(
-      `${process.env.SERVER_URL}/api/auth/signout?callbackUrl=/user/breaches`,
+      `${process.env.SERVER_URL}/user/breaches`,
       302
     );
   } catch (e) {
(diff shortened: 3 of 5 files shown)


src/app/(nextjs_migration)/(authenticated)/user/breaches/page.tsx
```diff
@@ -150,6 +150,35 @@ export default async function UserBreaches() {
     return breachRowsHTML;
   }
 
+  // Fetch list of subscriptions.
+  const result = await fetch(
+    `${process.env.OAUTH_API_URI}/oauth/mozilla-subscriptions/customer/billing-and-subscriptions`,
+    {
+      headers: {
+        "Content-Type": "application/json",
+        "authorization": `Bearer ${session?.user.subscriber?.fxa_access_token}`,
```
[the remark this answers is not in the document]
rhelmer: Yes, per the typescript definition for `session`... I think in practice we should always have the `fxa_access_token` if the user is logged in, but I can check this and throw earlier just in case.

src/app/(nextjs_migration)/(authenticated)/user/breaches/page.tsx
```diff
@@ -158,7 +187,7 @@ export default async function UserBreaches() {
       <Script type="module" src="/nextjs_migration/client/js/dialog.js" />
       <main data-partial="breaches">
         <section>
-          {(process.env.PREMIUM_ENABLED === "true" && !session?.user.fxa?.subscriptions?.includes("monitor")) ?
+          {(process.env.PREMIUM_ENABLED === "true" && !userSubscribed(subscriptions)) ?
```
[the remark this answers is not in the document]
rhelmer: I'll let that one land first and then rebase this one.

src/app/api/v1/premium/route.ts
```diff
@@ -10,19 +10,12 @@ export async function GET(req: NextRequest) {
...
     const _email = searchParams.get("email");
 
-    // TODO: The user either needs to be signed in again, or the
-    // JWT from FxA refreshed, in order to read the latest
-    // subscriptions.
-    //
-    // For now, redirect to the built-in sign-out, which will then
-    // send the user to the dashboard.
-
-    return NextResponse.redirect(`${process.env.SERVER_URL}/api/auth/signout?callbackUrl=/user/breaches`, 302);
```
rhelmer, line 25: Hm, we could actually move the subscription check here and add it to the session if it doesn't already exist. In the meantime I'll get to the bottom of why the `subscription` claim doesn't seem to exist on the dev server - it is using a different client ID than we use for local development which could be why.

src/app/(nextjs_migration)/(authenticated)/user/breaches/page.tsx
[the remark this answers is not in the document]
flozia: PR https://github.com/mozilla/blurts-server/pull/3143 has been merged.

src/app/api/v1/premium/route.ts
rhelmer: I took a shot at this but if I `getSession()` and then modify it, I don't see my modification later when the page is rendered and I'm not sure why... I think it's not too important yet, there's only one place we currently check for subscription status, but I think checking in the callback from subplat and adding it to the session is probably a better place.

.env-dist
```diff
@@ -120,4 +121,6 @@ E2E_TEST_ACCOUNT_PASSWORD=
 # Monitor Premium features
 PREMIUM_ENABLED=
 # Link to start user on the subscription process. PREMIUM_ENABLED must be set to `true`.
+SUBSCRIBE_PREMIUM_PRODUCT=
+SUBSCRIBE_PREMIUM_PLAN=
```
Vinnl, line 125: In case we expand to other countries later, it might be good to distinguish this from their plans:

```suggestion
SUBSCRIBE_PREMIUM_PLAN_US=
```

And we might even want to [align with Relay](https://github.com/mozilla/fx-private-relay/blob/2a589b32818ec68d7660a2b1c5606c874cdb9b7e/privaterelay/settings.py#L344-L350) and name these `PREMIUM_PRODUCT_ID` and `PREMIUM_PLAN_ID_US`, while we're at it.

src/app/(nextjs_migration)/(authenticated)/user/breaches/page.tsx
```diff
@@ -161,6 +161,45 @@ export default async function UserBreaches() {
...
+        authorization: `Bearer ${bearerToken}`,
+      },
+    }
+  );
+  let subscriptions;
+  if (result.ok) {
+    subscriptions = await result.json();
+  } else {
+    console.error("Could not refresh subscriptions:", result);
+  }
```
Vinnl, line 183: Might it make sense to turn this into a function that can also be called from elsewhere? Also, I'd really appreciate type definitions for the return type, even if they only cover what we actually use for now :) e.g.

```ts
type FxaSubscriptionResponse = {
  subscriptions: Array<{
    product_id: string;
    plan_id: string;
    status: "active";
  }>;
};
```

.env-dist
Vinnl: Also, I believe these aren't secret (I think they're exposed in the URL when subscribing), so we could also add working values here.

src/app/(nextjs_migration)/(authenticated)/user/breaches/page.tsx
```diff
@@ -161,6 +161,45 @@ export default async function UserBreaches() {
...
+
+  function userSubscribed(subscriptions: any) {
```
Vinnl, line 185: Nit, but I often find it useful if functions start with a verb, and functions that return a boolean to start with `is`, `has`, `contains`, or similar. So this could be e.g. `isUserSubscriber` or `includesPremiumSubscription`.

Vinnl (approved):
It's not great to be calling this API on every page load, but I assume this at least works for now. Added a couple of nits, feel free to ignore.

flozia (approved):
> I took a shot at this but if I getSession() and then modify it, I don't see my modification later when the page is rendered and I'm not sure why...

Just out of interest: How and from where did you modify the session?

src/app/(nextjs_migration)/(authenticated)/user/breaches/page.tsx
```diff
@@ -161,6 +161,45 @@ export default async function UserBreaches() {
...
+  const bearerToken = session?.user.subscriber?.fxa_access_token;
+  if (!bearerToken) {
+    console.error("User has no bearer token");
+  }
+  const result = await fetch(
+    `${process.env.OAUTH_API_URI}/oauth/mozilla-subscriptions/customer/billing-and-subscriptions`,
+    {
+      headers: {
+        "Content-Type": "application/json",
+        authorization: `Bearer ${bearerToken}`,
```
pdehaan, line 174: If this receives `Bearer undefined` will it always fail the fetch request? And if so, should we skip the network request entirely?

src/app/(nextjs_migration)/(authenticated)/user/breaches/page.tsx
```diff
@@ -161,6 +161,45 @@ export default async function UserBreaches() {
     return breachRowsHTML;
   }
 
+  // Fetch list of subscriptions.
+  const bearerToken = session?.user.subscriber?.fxa_access_token;
+  if (!bearerToken) {
+    console.error("User has no bearer token");
```
pdehaan, line 167: I think this relates to a previous comment, but if there is no bearer token, will the `fetch()` work, or should we just add some code here to abort early?

Although, I guess if the `bearerToken` is falsy here, then the `else` block on L182 below happen (which also just logs to STDERR and should leave `subscriptions` as `undefined` which then gets passed into `userSubscribed()` and I guess causes the "Subscribe to Premium" link to get rendered.

src/app/(nextjs_migration)/(authenticated)/user/breaches/page.tsx
```diff
@@ -161,6 +161,45 @@ export default async function UserBreaches() {
...
+  let subscriptions;
+  if (result.ok) {
+    subscriptions = await result.json();
+  } else {
+    console.error("Could not refresh subscriptions:", result);
+  }
+
+  function userSubscribed(subscriptions: any) {
+    if (!subscriptions) {
+      console.error("No subscription data found");
```
pdehaan, line 187: Q: Do we need this `console.error()` or will it just fill our stderr log file on the prod servers with hundreds/millions of lines with the same message?

.env-dist
rhelmer: Right they aren't secret (they are passed query params in the URL the user sees when completing payment), however we do need to configure it for each of our environments.

Should we do the one set up for local development as the default? I think that's what we do for other defaults like SERVER_URL.

src/app/(nextjs_migration)/(authenticated)/user/breaches/page.tsx
```diff
@@ -161,6 +161,45 @@ export default async function UserBreaches() {
...
+    for (const subscription of subscriptions.subscriptions) {
+      if (
+        subscription.product_id === process.env.SUBSCRIBE_PREMIUM_PRODUCT &&
+        subscription.plan_id === process.env.SUBSCRIBE_PREMIUM_PLAN &&
+        subscription.status === "active"
+      ) {
+        return true;
+      }
+    }
+    console.debug("user not subscribed:", subscriptions);
```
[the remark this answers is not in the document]
rhelmer: Nope, thanks!

rhelmer:
> > I took a shot at this but if I getSession() and then modify it, I don't see my modification later when the page is rendered and I'm not sure why...
> 
> Just out of interest: How and from where did you modify the session?

I tried doing in `/api/v1/premium`, but after reading about it I think it'd need to be done in one of the callbacks for next-auth. 

I think my current approach isn't great (we probably don't need to check on every page load!) but I want to get something in place so we can start testing the end-to-end payment flow.

rhelmer:
> > > I took a shot at this but if I getSession() and then modify it, I don't see my modification later when the page is rendered and I'm not sure why...
> > 
> > 
> > Just out of interest: How and from where did you modify the session?
> 
> I tried doing in `/api/v1/premium`, but after reading about it I think it'd need to be done in one of the callbacks for next-auth.
> 
> I think my current approach isn't great (we probably don't need to check on every page load!) but I want to get something in place so we can start testing the end-to-end payment flow.

Oh and for the "how": I tried `getServerSession()` and then modifying that session object, which didn't work on the next page... I think it's because only a few fields are persisted on the server side, we could potentially store this state somewhere like the database or Redis.

rhelmer:
> It's not great to be calling this API on every page load, but I assume this at least works for now. Added a couple of nits, feel free to ignore.

Yes I agree - I want it more as a working example for testing. As I mentioned in another comment, I think we should probably persist the "subscribed" state somewhere (db or Redis), and check it less often.

Users have a month after unsubscribing before their features actually go away, and we need to figure out all the details on what happens when unsubscription is enforced...

rhelmer:
Oh, sorry one more top-level comment: I'm hopeful that we'll figure out why we only see the `subscription` claim in the FxA token on local dev (and not heroku). That would be simpler to reason about I think, provided we can solve the problem of automatically refreshing a user's token without requiring them to sign in again.

src/app/api/v1/premium/route.ts
```diff
@@ -10,20 +10,13 @@ export async function GET(req: NextRequest) {
     // NOTE: the email address passed here cannot
     // be trusted, we must also check that the
     // `subscriptions` claim on the FxA token contains
-    // "monitor".
+    // "monitor", or call the auth server API to confirm.
```
rhelmer, line 13: One thought, if we do need to keep using this API long-term: we might want to do the API call here, and then record in the database that this user is a subscriber. We'd still need to periodically check that the user is still subscribed, and take action, but we could record a timestamp so we don't have to call it more often than once a day or so.

We might want to have a cron job that checks for unsubscribed users and deletes their data. Maybe we could email them first to let them know, and give a link to come back and renew if they want?

.env-dist
```diff
@@ -120,4 +123,5 @@ E2E_TEST_ACCOUNT_PASSWORD=
 # Monitor Premium features
 PREMIUM_ENABLED=
 # Link to start user on the subscription process. PREMIUM_ENABLED must be set to `true`.
-SUBSCRIBE_PREMIUM_URL=
+PREMIUM_PRODUCT_ID=prod_NErZh679W62lai
+PREMIUM_PLAN_ID_US=price_1MUNq0Kb9q6OnNsL4BoJgepf
```
Vinnl, line 127: @rhelmer I just saw somewhere that we might even have monthly and annual plans, so we'd probably want to signify that too:

```suggestion
PREMIUM_PLAN_ID_US_MONTHLY=price_1MUNq0Kb9q6OnNsL4BoJgepf
```

PR #3149, opened 2023-06-20
MNTOR-1820: feature flags backend / db

mansaj:
# References: 
Jira: MNTOR-1820

# Description
Feature flag API and admin dashboard

# Screenshot (if applicable)

Not applicable.

# How to test


src/app/api/v1/admin/feature-flags/[flagId]/route.ts
```diff
@@ -0,0 +1,33 @@
...
+
+import { NextRequest, NextResponse } from "next/server";
+import { getToken } from "next-auth/jwt";
+import {
+  getFeatureFlagByName
+} from "../../../../../../db/tables/featureFlags";
+import { isAdmin } from "../../../../utils/auth";
+import appConstants from "../../../../../../appConstants";
+
+//  import appConstants from "../../../../../appConstants";
```
Comment (pdehaan) [inline src/app/api/v1/admin/feature-flags/[flagId]/route.ts:13]: ✂️  (we already do this above on L11 anyways)

src/app/api/v1/admin/feature-flags/[flagId]/route.ts
```diff
@@ -0,0 +1,33 @@
...
+
+export async function GET(req: NextRequest) {
+  const token = await getToken({ req });
+  if (typeof token?.email === "string" && isAdmin(token?.email)) {
+    // Signed in
+    const flagName = req.nextUrl.toString().split('/').pop() || ''
+    console.log({ flagName })
```
Comment (pdehaan) [inline src/app/api/v1/admin/feature-flags/[flagId]/route.ts:20]: ✂️

src/app/api/v1/admin/feature-flags/[flagId]/route.ts
```diff
@@ -0,0 +1,33 @@
...
+    try {
+      const flag = await getFeatureFlagByName(flagName)
+      return NextResponse.json(flag);
```
Comment (pdehaan) [inline src/app/api/v1/admin/feature-flags/[flagId]/route.ts:23]: It's possible that `getFeatureFlagByName()` returns `null`. Does that cause any issues w/ `NextResponse.json(null)`, or do we need to throw an error here if the flag name wasn't found?

src/app/api/v1/admin/feature-flags/[flagId]/route.ts
```diff
@@ -0,0 +1,33 @@
...
+  getFeatureFlagByName
+} from "../../../../../../db/tables/featureFlags";
+import { isAdmin } from "../../../../utils/auth";
+import appConstants from "../../../../../../appConstants";
+
+//  import appConstants from "../../../../../appConstants";
+
+export async function GET(req: NextRequest) {
+  const token = await getToken({ req });
+  if (typeof token?.email === "string" && isAdmin(token?.email)) {
```
Comment (pdehaan) [inline src/app/api/v1/admin/feature-flags/[flagId]/route.ts:17]: Looks like we're doing this pattern in at least 4&times;.
Should we move the `typeof token?.email === "string"` check into the `isAdmin(...)` method?

src/app/api/v1/admin/feature-flags/[flagId]/route.ts
```diff
@@ -0,0 +1,33 @@
...
+    } catch (e) {
+      console.error(e)
+      return NextResponse.json({ success: false }, { status: 500 });
+    }
+  }
+  else {
+    // Not Signed in, redirect to home
+    return NextResponse.redirect(appConstants.SERVER_URL, 301);
+  }
+}
```
Comment (pdehaan) [inline src/app/api/v1/admin/feature-flags/[flagId]/route.ts:33]: nit: add trailing linebreak

src/app/api/v1/admin/feature-flags/route.ts
```diff
@@ -0,0 +1,93 @@
...
+import {
+  getAllFeatureFlags,
+  addFeatureFlag,
+  enableFeatureFlagByName,
+  disableFeatureFlagByName,
+  updateAllowList,
+  updateWaitList
+} from "../../../../../db/tables/featureFlags";
+
+import { FeatureFlag } from "../../../../../db/tables/featureFlags";
```
pdehaan, line 16: nit: merge this w/ L7-14 above?

src/db/migrations/20230618104332_feature_flags.js
```diff
@@ -0,0 +1,25 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+ export function up (knex) {
+  return knex.schema
+    .createTable('feature_flags', table => {
+      table.string('name').primary().unique()
```
pdehaan, line 8: Does this knex magic auto-index this column?
I believe this was the only column that we're querying w/ `where=` clauses.

mansaj: Yes, a primary key index should be created

src/app/(nextjs_migration)/(authenticated)/admin/feature-flags/page.tsx
```diff
@@ -0,0 +1,65 @@
...
+
+export default async function FeatureFlagRootPage() {
+  const session = await getServerSession(authOptions);
+
+  if (!session?.user?.email) {
+    return redirect("/");
+  }
+
+  const admins = process.env.ADMINS?.split(",") ?? [];
+  const isAdmin = admins.includes(session.user.email);
```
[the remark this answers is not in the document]
mansaj: Good catch, fixed thanks

src/app/(nextjs_migration)/(authenticated)/admin/feature-flags/page.tsx
```diff
@@ -0,0 +1,60 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
```
rhelmer, line 1: @Vinnl was advising me to use the `(proper_react)` subdir for new code, should that apply here?

src/app/api/utils/auth.ts
```diff
@@ -222,3 +222,8 @@ export function bearerToken(req: NextRequest) {
   // The remaining portion, which should be the token
   return authHeader.substring(authHeader.indexOf(" ") + 1);
 }
+
+export function isAdmin(email: string) {
+  const admins = AppConstants.ADMINS?.split(",") ?? [];
```
rhelmer, line 227: Is `,` allowed in email addresses? I think it might technically be... since this is set in the env var and we control that I think it's OK in this case, just something to be aware of.

src/app/api/v1/admin/feature-flags/[flagId]/route.ts
```diff
@@ -0,0 +1,28 @@
...
+import { getToken } from "next-auth/jwt";
+import { getFeatureFlagByName } from "../../../../../../db/tables/featureFlags";
+import { isAdmin } from "../../../../utils/auth";
+import appConstants from "../../../../../../appConstants";
+
+export async function GET(req: NextRequest) {
+  const token = await getToken({ req });
+  if (typeof token?.email === "string" && isAdmin(token?.email)) {
+    // Signed in
+    const flagName = req.nextUrl.toString().split("/").pop() || "";
```
Comment (rhelmer) [inline src/app/api/v1/admin/feature-flags/[flagId]/route.ts:15]: `req.nextUrl.pathname` would be easier/safer I think (I think query params might break this?)

src/db/migrations/20230618104332_feature_flags.js
```diff
@@ -0,0 +1,25 @@
...
+      table.boolean('is_enabled').defaultTo(false)
+      table.text('description').default('')
+      table.specificType('dependencies', 'character varying(255)[]')
+      table.specificType('allow_list', 'character varying(255)[]')
+      table.specificType('wait_list', 'character varying(255)[]')
+      table.timestamp('created_at').defaultTo(knex.fn.now())
+      table.timestamp('modified_at').defaultTo(knex.fn.now())
```
rhelmer, line 15: I was going to ask for these if you hadn't already included them, nice :) Please see my other comment re: soft-delete too.

src/db/tables/featureFlags.ts
```diff
@@ -0,0 +1,129 @@
...
+  expiredAt?: number;
+  owner?: string;
+};
+
+async function getAllFeatureFlags() {
+  return await knex("feature_flags").returning("*");
+}
+
+async function getFeatureFlagByName(name: string) {
+  log.info("getFeatureFlagByName", name);
```
rhelmer, line 40: I was going to ask if we could do an audit log, I think this should suffice.

rhelmer (changes requested):
This looks great, I have a few small changes to suggest.

src/app/api/v1/admin/feature-flags/[flagId]/route.ts
Comment (rhelmer) [reply src/app/api/v1/admin/feature-flags/[flagId]/route.ts:15]: Actually looking into this, I think this is the right way to get `flagId` when using the `[]` syntax:
https://nextjs.org/docs/pages/building-your-application/routing/dynamic-routes

src/app/(nextjs_migration)/(authenticated)/admin/feature-flags/page.tsx
Vinnl: Yes please! I'd like to wholesale remove `(nextjs_migration)` once we release the new version, so anything that should live on beyond that would ideally not be in that directory.

src/app/(nextjs_migration)/(authenticated)/admin/feature-flags/page.tsx
```diff
@@ -0,0 +1,60 @@
...
+            <th>Name</th>
+            <th>Enabled</th>
+            <th>Dependencies</th>
+            <th>Allow List</th>
+            <th>Wait List</th>
+          </tr>
+        </thead>
+        <tbody>
+          {data.map((item) => (
+            <tr key="">
```
Vinnl, line 39: You'll want to use a key that uniquely identifies this particular row, so that React knows how to update the DOM if an element in `data` gets removed or added:

```suggestion
            <tr key={item.name}>
```

src/app/api/v1/admin/feature-flags/[flagId]/route.ts
```diff
@@ -0,0 +1,28 @@
...
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import { NextRequest, NextResponse } from "next/server";
+import { getToken } from "next-auth/jwt";
+import { getFeatureFlagByName } from "../../../../../../db/tables/featureFlags";
+import { isAdmin } from "../../../../utils/auth";
+import appConstants from "../../../../../../appConstants";
+
+export async function GET(req: NextRequest) {
+  const token = await getToken({ req });
```
Comment (Vinnl) [inline src/app/api/v1/admin/feature-flags/[flagId]/route.ts:12]: I think Next-Auth's idea is that we use `getSession` to read session data, rather than directly inspecting the data inside the token. I'm not sure yet what extra checks they wrap around it, but it's probably safer to go with that.

src/app/api/v1/admin/feature-flags/[flagId]/route.ts
Comment (Vinnl) [reply src/app/api/v1/admin/feature-flags/[flagId]/route.ts:15]: The `GET` function can take a second parameter, which is an object with the `params` property that contains the `flagId` property from the URL: https://nextjs.org/docs/app/building-your-application/routing/router-handlers#dynamic-route-segments

(Note that `/pages/` in your URL and the "Using Pages Router" in the top left-hand corner of the page you linked; Next.js is transitioning from one routing system to another, and we're already on the new one, the App Router. So make sure to look at the correct version of the docs.)

src/app/api/v1/admin/feature-flags/[flagId]/route.ts
```diff
@@ -0,0 +1,28 @@
...
+  if (typeof token?.email === "string" && isAdmin(token?.email)) {
+    // Signed in
+    const flagName = req.nextUrl.toString().split("/").pop() || "";
+    console.log({ flagName });
```
Comment (Vinnl) [inline src/app/api/v1/admin/feature-flags/[flagId]/route.ts:16]:
```suggestion
```

src/app/api/v1/admin/feature-flags/route.ts
```diff
@@ -0,0 +1,90 @@
...
+} from "../../../../../db/tables/featureFlags";
+
+import type { FeatureFlag } from "../../../../../db/tables/featureFlags";
+import { isAdmin } from "../../../utils/auth";
+import appConstants from "../../../../../appConstants";
+
+export async function GET(req: NextRequest) {
+  const token = await getToken({ req });
+  if (typeof token?.email === "string" && isAdmin(token?.email)) {
+    // Signed in
```
Vinnl, line 23: Same goes for the other request handlers, but I like the pattern of favouring early returns over `if...else` clauses, to make it easy to inspect required preconditions, as well as to reduce indentation:

```suggestion
  if (typeof token?.email !== "string" || !isAdmin(token?.email)) {
    return NextResponse.redirect(appConstants.SERVER_URL, 301);
  }
  // <contents of if{} block>
```

src/app/api/v1/admin/feature-flags/route.ts
```diff
@@ -0,0 +1,90 @@
...
+export async function POST(req: NextRequest) {
+  const token = await getToken({ req });
+  if (typeof token?.email === "string" && isAdmin(token?.email)) {
+    // Signed in
+    try {
+      const newFlag = (await req.json()) as FeatureFlag;
+      const resp = await addFeatureFlag(newFlag);
+      return NextResponse.json(resp);
+    } catch (e) {
+      console.error(e);
```
Vinnl, line 45: Should we use `mozlog` to help SRE more easily pinpoint errors?

src/app/api/v1/admin/feature-flags/route.ts
```diff
@@ -0,0 +1,90 @@
...
+      return NextResponse.json({ success: false }, { status: 500 });
+    }
+  } else {
+    // Not Signed in, redirect to home
+    return NextResponse.redirect(appConstants.SERVER_URL, 301);
+  }
+}
+
+type FeatureFlagPutRequest = {
```
Vinnl, line 54: :clap: Great work adding type definitions for request bodies.

If you export them, the frontend can also re-use them when composing the requests:

```suggestion
export type FeatureFlagPutRequest = {
```

src/app/api/v1/admin/feature-flags/route.ts
```diff
@@ -0,0 +1,90 @@
...
+  name: string;
+  isEnabled?: boolean;
+  allowList?: [string];
```
Vinnl, line 57: Just a note in case it's unintentional: `[string]` is an array with a single element (a `string`), whereas `string[]` (or `Array<string>`) is an array of strings.

src/db/migrations/20230618104332_feature_flags.js
Vinnl: I haven't used it myself, but I saw that Knex also [has a `timestamps` function](https://knexjs.org/guide/schema-builder.html#timestamps) that supposedly automatically adds both of these?

src/db/migrations/20230618104332_feature_flags.js
```diff
@@ -0,0 +1,25 @@
...
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+ export function up (knex) {
+  return knex.schema
+    .createTable('feature_flags', table => {
+      table.string('name').primary().unique()
+      table.boolean('is_enabled').defaultTo(false)
+      table.text('description').default('')
+      table.specificType('dependencies', 'character varying(255)[]')
```
Vinnl, line 11: I'm not sure I remember what the values of these were, but should these be [`json`](https://knexjs.org/guide/schema-builder.html#json) or [`jsonb`](https://knexjs.org/guide/schema-builder.html#jsonb) columns? Or is the `[]` a cool Postgres feature I haven't heard of yet that allows you to store arrays in columns?

src/db/migrations/20230618104332_feature_flags.js
```diff
@@ -0,0 +1,25 @@
...
+      table.specificType('allow_list', 'character varying(255)[]')
+      table.specificType('wait_list', 'character varying(255)[]')
+      table.timestamp('created_at').defaultTo(knex.fn.now())
+      table.timestamp('modified_at').defaultTo(knex.fn.now())
+      table.timestamp('expired_at')
+      table.string('owner')
```
Vinnl, line 17: Just wondering if it would be desirable to potentially have multiple owners?

Also, not sure how you both feel about this (it might just be noisy), but I sometimes feel like it can be useful to explicitly mark every column as `nullable` or `notNullable`, just so the former jumps out more.

src/db/tables/featureFlags.ts
```diff
@@ -0,0 +1,129 @@
...
+    owner: flag.owner,
+  };
+  const res = await knex("feature_flags").insert(featureFlagDb).returning("*");
+
+  return res[0];
+}
+
+async function deleteFeatureFlagByName(name: string) {
+  log.info("deleteFeatureFlagByName", name);
+  await knex("feature_flags").where("name", "=", name).delete();
```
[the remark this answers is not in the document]
Vinnl: Do we even need deletion? It's already possible to disable them, and deleting should primarily happen in code - so maybe the risks of this as a footgun are larger than the benefits?

Vinnl (commented):
The three main changes that I think are important to do:

- Move code out of `(nextjs_migration)`
- Use `getSession`
- Use the second parameter's `params` argument to read route data.

The current ways _might_ be harmless, but especially in this early phase where we're establishing code patterns, I'd like to stick as closely to the ways described in the docs as possible, just to avoid issues down the road when code gets copy-pasted.

The rest are just suggestions or straightforward.

src/app/api/v1/admin/feature-flags/[flagId]/route.ts
Comment (mansaj) [reply src/app/api/v1/admin/feature-flags/[flagId]/route.ts]: Thanks, I knew there'd be a way to do this! Fixed

src/app/api/v1/admin/feature-flags/[flagId]/route.ts
Comment (mansaj) [reply src/app/api/v1/admin/feature-flags/[flagId]/route.ts]: I think it's redundant, simplified

src/app/api/v1/admin/feature-flags/[flagId]/route.ts
Comment (mansaj) [reply src/app/api/v1/admin/feature-flags/[flagId]/route.ts]: Using session now

src/db/migrations/20230618104332_feature_flags.js
mansaj: Yea it's a postgres specific type that Knex doesn't support (fully).. Though this is a recommended way of using it with Knex

src/db/tables/featureFlags.ts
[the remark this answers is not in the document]
mansaj: @Vinnl I think delete can be nice (even if it's soft delete/ archive) to just have a way to hide it from the UI. One of the problems with feature flags (for a larger projects) is to have too many zombie flags..

I'll make the change for a soft delete

src/app/(nextjs_migration)/(authenticated)/admin/feature-flags/page.tsx
```diff
@@ -0,0 +1,65 @@
...
+  }
+
+  const admins = process.env.ADMINS?.split(",") ?? [];
+  const isAdmin = admins.includes(session.user.email);
+
+  if (!isAdmin) {
+    return notFound();
+  }
+
+  const AllFlagsTable = (featureFlags: {data: Array<any>}) => {
```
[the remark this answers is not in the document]
mansaj: Fixed thanks

src/app/(nextjs_migration)/(authenticated)/admin/feature-flags/page.tsx
```diff
@@ -0,0 +1,65 @@
...
+    const { data } = featureFlags
+    // Extract column names from the first object in the data array
+    const columns = Object.keys(data[0]);
+
+    return (
+      <>
```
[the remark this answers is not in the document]
mansaj: Fixed thanks

src/db/tables/featureFlags.ts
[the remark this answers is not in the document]
mansaj: Fixed

src/app/api/v1/admin/feature-flags/[flagId]/route.ts
Comment (mansaj) [reply src/app/api/v1/admin/feature-flags/[flagId]/route.ts]: Returning null shouldn't be an issue. These are admin endpoints, we will be the only ones consuming it

src/app/api/v1/admin/feature-flags/route.ts
mansaj: Import type only, fixed thanks

src/app/(nextjs_migration)/(authenticated)/admin/feature-flags/page.tsx
mansaj: Moved to (proper_react).. At this point, we can't see that what's in that folder from the frontend? @rhelmer @Vinnl

src/app/api/utils/auth.ts
mansaj: hmm interesting.. We might need another way to separate the addresses

src/db/migrations/20230618104332_feature_flags.js
mansaj: I think this way is more explicit (clearer)

src/db/migrations/20230618104332_feature_flags.js
mansaj: Hmm.. I think it's usually best practice to have one. In the future, we CAN add the functionality to transfer ownership.

src/app/(nextjs_migration)/(authenticated)/admin/feature-flags/page.tsx
Vinnl: According to the current file path, the URL to the page should be `/redesign/admin/feature-flags/` - is that what you mean?

src/db/tables/featureFlags.ts
```diff
@@ -8,7 +8,7 @@ import mozlog from "../../utils/log.js";
 const knex = Knex(knexConfig);
 const log = mozlog("DB.flags");
 
-type FeatureFlagDB = {
+export type FeatureFlagDB = {
```
Vinnl, line 11: I submitted a PR yesterday to provide type definitions for our database tables in the way that's recommended by its documentation, and that allows them to be recognised by the query builder. You might want to align these with that: https://github.com/mozilla/blurts-server/pull/3167

(And also here you might want to change `[string]` to `string[]`.)

Vinnl (approved):
Addresses the blockers that I've seen, nice job :+1:

src/app/(nextjs_migration)/(authenticated)/admin/feature-flags/page.tsx
mansaj: Yes thanks!

src/db/tables/featureFlags.ts
mansaj: @Vinnl Good call. I'll modify this after your PR lands

Vinnl: Landed! :)

PR #3150, opened 2023-06-21
Attempt to fix image remotePatterns for stage and prod

flozia:
# References: 
Jira: [MNTOR-1853](https://mozilla-hub.atlassian.net/browse/MNTOR-1853)

# Description

I see the fallback placeholder for the profile image working on dev, but on stage, the network request is failing because of CSP reasons. This is just a guess: Let’s try to add a trailing `/` since the referer hostname might need an exact match.

# Screenshot

Profile placeholder
| dev | stage |
| --- | --- |
| profile-dev | profile-stage |

# How to test

Only testable after the PR has been promoted to `stage`.

PR #3151, opened 2023-06-21
Allow all <Image> sources to be loaded in our CSP

Vinnl:
# References: 
Jira: MNTOR-1853
Figma: N/A

# Description

Profile images got blocked by our Content Security Policy:

> Content Security Policy: The page’s settings blocked the loading of a resource at https://profile.stage.mozaws.net/v1/avatar/v (“img-src”).

They were already allowlisted [via `<Image>`'s `remotePatterns`](https://nextjs.org/docs/app/api-reference/components/image#remotepatterns), but those weren't necessarily added to our Content Security Policy.

Supersedes #3150.

# How to test

Run `npm run build; npm start`, then verify that the profile image (or its placeholder) does get loaded.

### DIFF
--- a/next.config.js
+++ b/next.config.js
@@ -59,11 +59,16 @@ const nextConfig = {
               `connect-src 'self' ${
                 process.env.NODE_ENV === "development" ? "webpack://*" : ""
               } https://*.google-analytics.com https://*.analytics.google.com https://*.googletagmanager.com https://*.ingest.sentry.io`,
-              `img-src 'self' https://*.google-analytics.com https://*.googletagmanager.com https://firefoxusercontent.com https://mozillausercontent.com https://monitor.cdn.mozilla.net ${
-                process.env.FXA_ENABLED
-                  ? new URL(process.env.OAUTH_PROFILE_URI).origin
-                  : ""
-              }`,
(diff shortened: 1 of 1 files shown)


next.config.js
```diff
@@ -59,11 +59,16 @@ const nextConfig = {
               `connect-src 'self' ${
                 process.env.NODE_ENV === "development" ? "webpack://*" : ""
               } https://*.google-analytics.com https://*.analytics.google.com https://*.googletagmanager.com https://*.ingest.sentry.io`,
-              `img-src 'self' https://*.google-analytics.com https://*.googletagmanager.com https://firefoxusercontent.com https://mozillausercontent.com https://monitor.cdn.mozilla.net ${
-                process.env.FXA_ENABLED
-                  ? new URL(process.env.OAUTH_PROFILE_URI).origin
```
Vinnl, line 64: This should have already resulted in the profile image being allowlisted on stage, but I think these environment variables just aren't set to what I thought they would be there. Either way, since we're already hardcoding them in an allowlist above, I figured it made sense to re-use that, in case new URLs get added there later.

flozia (approved):

PR #3152, opened 2023-06-21
Set the base path for metadata

Vinnl:
This ensures that relative URLs (such as those in
`openGraph.images`) resolve to the correct URL, and thus fixes the problem called out by @pdehaan in https://github.com/mozilla/blurts-server/pull/3134#issuecomment-1593770158.

We should be able to properly test this on stage, but it does give confidence that a warning about this got removed from the build log, and that [the docs say](https://nextjs.org/docs/app/api-reference/functions/generate-metadata#metadatabase):

> The field's relative path will be composed with metadataBase to form a fully qualified URL.
If not configured, metadataBase is automatically populated with a [default value](https://nextjs.org/docs/app/api-reference/functions/generate-metadata#default-value).

It should also apply for all other pages:

> metadataBase is typically set in root app/layout.js to apply to URL-based metadata fields across all routes.

### DIFF
--- a/src/app/layout.tsx
+++ b/src/app/layout.tsx
@@ -3,6 +3,7 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
 import { ReactNode } from "react";
+import { Metadata } from "next";
 import { Inter } from "next/font/google";
 import "./globals.css";
 import { getServerSession } from "next-auth";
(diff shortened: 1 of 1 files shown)


src/app/layout.tsx
```diff
@@ -14,11 +15,15 @@ import { metropolis } from "./fonts/Metropolis/metropolis";
...
 
-export async function generateMetadata() {
+export async function generateMetadata(): Promise<Metadata> {
   const l10n = getL10n();
   return {
     title: l10n.getString("brand-fx-monitor"),
     description: l10n.getString("meta-desc-2"),
+    metadataBase:
+      typeof process.env.SERVER_URL === "string"
+        ? new URL(process.env.SERVER_URL)
```
pdehaan, line 25: Not sure when/if this could happen, but if your `SERVER_URL` is an _empty_ string, this will throw a TypeError on the `URL()` constructor call:

```js
let serverUrl = "";
new URL(serverUrl);

/*
Uncaught TypeError: Failed to construct 'URL': Invalid URL
    at <anonymous>:1:1
*/
```

rhelmer: `SERVER_URL` is defined in `.env-dist`, if it's not set I think we probably do want to throw here since lots of things won't work correctly. It's not something we can reasonable handle and keep the server functioning, I don't think.

rhelmer: If you want to explicitly catch I think that's fine, but I would re-throw. Up to you I think @Vinnl - I prefer to not handle exceptions unless we are sure we really can handle them (e.g. "swallowing" the exception by just logging it and moving on would just lead to later failures, so we're not really "handling" it)

rhelmer (approved):

flozia (approved):

PR #3154, opened 2023-06-21
Fix HIBP link in verification email

Vinnl:
Since this is the same fix as in https://github.com/mozilla/blurts-server/pull/3134/commits/add1e8ecf889fbc819b6c86ac34496fdc7cec876, I figured I'd quickly apply it here too.

# References: 
Jira: MNTOR-1858

### DIFF
--- a/src/views/emails/email2022.js
+++ b/src/views/emails/email2022.js
@@ -142,7 +142,12 @@ const emailFooter = (data, l10n) => {
       <p>
         ${getMessage('email-2022-hibp-attribution', {
         'hibp-link-attr': `href='${links(data).hibp}' rel='noopener'`
-      })}
+      })
+      // The following are special characters inserted by Fluent,
+      // which break the link when inserted into the tag.
+      ?.replaceAll("⁩", "")
+      .replaceAll("⁨", "")
+      }
       </p>
       <img
         alt='${getMessage('mozilla')}'


src/views/emails/email2022.js
```diff
@@ -142,7 +142,12 @@ const emailFooter = (data, l10n) => {
       <p>
         ${getMessage('email-2022-hibp-attribution', {
         'hibp-link-attr': `href='${links(data).hibp}' rel='noopener'`
-      })}
+      })
+      // The following are special characters inserted by Fluent,
+      // which break the link when inserted into the tag.
+      ?.replaceAll("⁩", "")
+      .replaceAll("⁨", "")
+      }
```
pdehaan, line 150: Still a bit curious about this.
Why is it needed and why isn't it needed _more/everywhere_?

Also, if we are using it in more than 1-2 places, should we create some helper function or react hook or wrapper around `getMessage()`/`getStringLookup()`?

flozia: A helper function would be nice. If I understood correctly [these Unicodes](https://www.w3.org/International/questions/qa-bidi-unicode-controls.en.html#basedirection) `U+2068` (FSI – FIRST-STRONG ISOLATE) and `U+2069` (PDI – POP DIRECTIONAL ISOLATE) are being used for marking the direction of text inline and I suspect that these are being added to strings that are being included as variables?

flozia (approved):

src/views/emails/email2022.js
Vinnl: Generally, it should _not_ be needed. `@fluent/react` (but apparently not the Fluent.js we were using) is inserting these characters around variables with a purpose (IIRC, to support RTL locales). However, this breaks when we do this weird thing where the localised strings contain plain HTML with a variable inside the tag (i.e. `<a { $attrs }>some text</a>`).

So the real solution is to not do that anymore - instead, we can use `elems` in `getFragment` or `<Localized>` to properly inject DOM elements, and the special characters can be preserved to do their job. So I definitely wouldn't want to encourage doing this in more places by creating a helper function.

PR #3155, opened 2023-06-21
MPP-1683, MPP-1681 - View all exposures on dashboard 

codemist:
# References: 
Jira: MNTOR-1683, MNTOR-1681
Figma: https://www.figma.com/file/CaEKIhvSJqf6KNIMzSkt40/Concepts-for-Monitor-MVP-Redesign?type=design&node-id=546-44852&t=H8fUXbL4NhS5pBQq-4

# Description

This utilizes the `Shell` page layout and inputs some `ExposureCard` elements into it.

# Screenshot (if applicable)
image


src/app/(proper_react)/Shell.module.scss
```diff
@@ -71,57 +79,58 @@
...
 
-      .mozillaLink {
-        padding: $spacing-md $spacing-2xl;
-      }
+    .mozillaLink {
+      padding: $spacing-md $spacing-2xl;
     }
   }
-}
+}
```
pdehaan, line 136: nit: add blank line at end of file.

src/app/(proper_react)/Dashboard.tsx
```diff
@@ -0,0 +1,19 @@
...
+export const Dashboard = () => {
+  const timeStamp = 1668144000; // Example Unix timestamp
+
+  return (
+    <ShellEl l10n={getL10n()} session={null}>
+      <h2>View all exposures that are fixed or in-progress</h2>
+      <ExposureCard exposureImg={FamilyTreeImage} exposureName={"Familytree.com"} exposureType={"dataBreach"} exposureDetailsLink={"linkeehere.com"} dateFound={timeStamp} statusPillType={"needAction"} />
+    </ShellEl>
+  );
+}
```
pdehaan, line 19: nit: Missing semi-colon?
Not sure if we need to rebase to get latest linting/prettier configs

src/app/(proper_react)/Shell.tsx
```diff
@@ -24,72 +24,74 @@ export const Shell = (props: Props) => {
...
-            <ul className={styles.externalLinks}>
+            </Link>
+            <ul>
+              {/* Note: If you add elements here, also add them to <MobileShell>'s navigation */}
               <li>
-                <a
-                  href="https://support.mozilla.org/kb/firefox-monitor-faq"
-                  title={l10n.getString("footer-external-link-faq-tooltip")}
+                <PageLink
+                  href="/redesign/user/dashboard"
```
pdehaan, line 41: Q: is "/redesign/" part of the final paths?

Only found 2 references, possibly both from this PR, but it feels like a very temporary path:
```sh
git grep -n "/redesign/" | cat

src/app/(proper_react)/MobileShell.tsx:71:                  href="/redesign/user/dashboard"
src/app/(proper_react)/Shell.tsx:41:                  href="/redesign/user/dashboard"
```

src/app/(proper_react)/Shell.tsx
```diff
@@ -24,72 +24,74 @@ export const Shell = (props: Props) => {
...
+                <li>
+                  <a href="https://www.mozilla.org/privacy/firefox-monitor">
+                    {l10n.getString("terms-and-privacy")}
+                  </a>
+                </li>
+                <li>
+                  <a href="https://github.com/mozilla/blurts-server">
+                    {l10n.getString("github")}
+                  </a>
+                </li>
```
pdehaan, line 91: Q: Should these `<a href="">`s use `target="_blank"` to spawn a new window/tab?

src/app/components/client/ExposureCard.module.scss
```diff
@@ -1,49 +1,54 @@
...
     li {
-      flex: 1;
       display: flex;
-      align-items: center;
-    }
-
-    .exposureImage {
-      max-width: 150px;
-      height: 100%;
+      align-items: center;
```
pdehaan, line 45: nit: looks like a few bonus trailing spaces here.

src/app/components/client/ExposureCard.module.scss
```diff
@@ -84,25 +95,46 @@
         list-style: none;
```
pdehaan, line 95: Above on L34 we changed this to `list-style-type: none;`
Not sure if that needs to change here as well or why it changed above.

src/app/components/client/ExposureCard.tsx
```diff
@@ -42,77 +42,81 @@ export const ExposureCard = (props: Props) => {
...
+    window.addEventListener('resize', handleResize);
+    handleResize();
+
+    return () => window.removeEventListener('resize', handleResize);
+  }, []);
+
+
+  const date = new Date(dateFound * 1000); // Multiply by 1000 to convert seconds to milliseconds
+
+  const formattedDate = date.toLocaleDateString("en-US", {
```
pdehaan, line 66: Q: should this do some user locale detection instead of forcing en-US locale formatting?

src/app/components/client/ExposureCard.tsx
```diff
@@ -42,77 +42,81 @@ export const ExposureCard = (props: Props) => {
...
     }
 
     return (
-      <>
+      <dl className={styles.detailsFoundItem}>
         <dt>
           <span className={styles.exposureTypeIcon}>{props.icon}</span>
           {headline}
         </dt>
         <dl>{description}</dl>
```
pdehaan, line 100: Should this be [`<dd>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dd) (Description _Details_ element) instead of a nested [`<dl>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dl) (Description _List_ element)?

src/app/components/server/StatusPill.tsx
```diff
@@ -5,15 +5,30 @@
...
+    className = "inProgress";
+  } else if (type === "fixed") {
+    stringContent = "Fixed";
+    className = "isFixed";
+  }
+
+  const classes = [styles.pill, styles[className]].filter(Boolean).join(" ");
+
+  return <div className={classes}>{stringContent}</div>;
+};
```
pdehaan, line 34: nit: add linebreak at end of file

src/app/(proper_react)/DashboardTopBanner.tsx
```diff
@@ -0,0 +1,100 @@
...
+      remaining?: number,
+    }
+  }
+}
+
+export const DashboardTopBanner = (props: DashboardTopBannerProps) => {
+  const contentData = {
+    LetsFixDataContent: {
+      headline: "Let's protect your data",
+      description: `We found your data in ${props.data?.dataBreaches} data breaches and ${props.data?.infoForSale} sites selling your personal info. We'll guide you on how to fix it.`,
```
pdehaan, line 27: re: `${props.data?.dataBreaches}` does this content get suppressed if there were no breaches or data passed?
Curious if this could potentially log:
> "We found your data in _undefined_ data breaches and _undefined_ sites selling your personal info..."

src/app/(proper_react)/Shell.module.scss
```diff
@@ -1,7 +1,15 @@
 @import "../tokens";
 
+main {
+  min-height: 100vh; // ensure container takes up full viewport height
+  position: relative;
+  display: flex;
+  flex-direction: column;
+  justify-content: space-between;
+}
+
```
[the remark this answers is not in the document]
Vinnl: I would add `main` to the `<Shell>`, as the children might include more than just the main content (for example, the dashboard has a tab bar in there). And in general, if all elements that wrap the footer have `height: 100%`, that should already result in it being pushed to the bottom, as far as I know - this is the case at `/redesign/user/dashboard`, for example.

src/app/(proper_react)/Shell.tsx
Vinnl: It is a temporary path, but it's to avoid conflicting with the existing paths, some of which we will replace. We'll remove that path segment when we go live with this.

src/app/components/client/ExposureCard.tsx
Vinnl, line 1: Could you move the strings in here to `pendingTranslations.ftl` as well?

src/app/components/client/ExposureCard.tsx
```diff
@@ -35,84 +35,88 @@ type DetailsFoundProps = {
...
     exposureName,
     exposureType,
     exposureDetailsLink,
     dateFound,
-    statusPillContent,
     statusPillType,
   } = props;
 
   const [detailsOpen, setDetailsOpen] = useState(false);
+  const [isMobile, setIsMobile] = useState(false);
```
Vinnl, line 49: Is it not possible to hide all the elements that listen to `isMobile` with media queries? That should be more performant AFAIK.

Otherwise, we have a `useMediaQuery` hook in Relay that might be useful to use in Monitor.

src/app/components/client/ExposureCard.tsx
```diff
@@ -18,15 +18,15 @@ import {
...
 import Image from "next/image";
 import { Button } from "../server/Button";
+import { ExposureType, ExposureTypeEl } from "../server/ExposureType";
 
-export type Props = {
+export type ExposureCardProps = {
   exposureImg: StaticImageData;
   exposureName: string;
-  exposureType: string;
+  exposureType: ExposureType;
```
Vinnl, line 26: Is this to distinguish between data breaches and data brokers? If so, we probably want to take an `exposureData` element of type `ScanResult | Breach` from [here](https://github.com/mozilla/blurts-server/blob/75e7d70988ddc122e2882feaa69aa807d0218819/src/external/onerep.js#L249-L264) and [here](https://github.com/mozilla/blurts-server/blob/75e7d70988ddc122e2882feaa69aa807d0218819/src/app/(nextjs_migration)/(authenticated)/user/breaches/breaches.d.ts#L65-L86), and write a [type guard function](https://www.typescriptlang.org/docs/handbook/advanced-types.html#user-defined-type-guards) that distinguishes between the two.

And possibly then pass the result on to breach- and data broker-specific components, so they can make assumptions about what properties are available?

src/app/components/client/ExposureCard.tsx
```diff
@@ -35,84 +35,88 @@ type DetailsFoundProps = {
...
-            <li>{exposureType}</li>
-            <li>{dateFound}</li>
+            {!isMobile ? <li><ExposureTypeEl type={exposureType} /></li> : ""}
+            {!isMobile ? <li>{formattedDate}</li> : ""}
             <li>
-              <StatusPill type={statusPillType} content={statusPillContent} />
+              <StatusPill type={statusPillType} />
             </li>
           </ul>
           <span
```
Vinnl, line 122: Hmm, I can't add comments to it, but `<ChevronDown>` doesn't have an alt text, even though the image does include information (i.e. what will happen when you click it).

src/app/components/client/ExposureCard.module.scss
```diff
@@ -84,25 +99,46 @@
...
+        @media screen and (min-width: $screen-md) {
           display: flex;
+          flex-direction: row;
+          align-items: center;
+        }
 
-          svg {
-            color: $color-purple-70;
+        li + li {
+         padding-left: 23px; // offset the exposure icon
```
Vinnl, line 113: With magic numbers I'd like to know how the number was calculated. For example, is the exposure icon 46px and does it need to be shifted halfway to the left?

That makes it easier for reviewers and future readers to know why it is there, and whether it still needs to be there. For example, is there any reason this could not be solved with a `gap`?

src/app/components/client/ExposureCard.tsx
```diff
@@ -133,42 +137,43 @@ export const ExposureCard = (props: Props) => {
...
+                <OpenInNew alt="" width="13" height="13" />
               </span>
-            </a>{" "}
+            </a>
+            {" "}
             Remove this profile to protect your privacy.
           </p>
           <div className={styles.exposureListOfExposureTypes}>
             <ul>
               <li>Your exposed info:</li>
```
Vinnl, line 152: This `<li>` feels like it's more a heading for the list, than that it is the first element of the list?

And I think for the rest, instead of a `<ul>`, you'll want a `<dl>`, and then the `<li>`s can be replaced by `<div>`s containing just the `<dt>`s and `<dd>`s.

src/app/components/client/ExposureCard.module.scss
```diff
@@ -84,25 +99,46 @@
...
+        .detailsFoundItem {
           dt {
             display: flex;
             gap: $spacing-sm;
             align-items: center;
+            position: relative;
+
+            span {
+              color: $color-purple-60;
+              position: absolute;
```
Vinnl, line 133: Why is this absolutely positioned? It looks like it should be part of the regular document flow? (And likewise for the `relative` above.)

src/app/components/client/ExposureCard.module.scss
```diff
@@ -30,20 +34,25 @@
...
-      height: 100%;
+      @media screen and (min-width: $screen-md) {
+        flex: 1; 
+      }  
+      
+      .exposureImage {
+        width: 120px;
+        height: 100%;
+        object-fit: cover;
+      }
```
Vinnl, line 55: Quick heads-up: AFAIK we haven't come up with a way to get proper rectangular company logos here, so this will probably need to become favicons at some point.

src/app/(proper_react)/Shell.module.scss
Vinnl, line 1: Are the changes in here still needed? (It's a bit hard to see what's new and not, with things moving around and formatting changes.)

src/app/(proper_react)/DashboardTopBanner.module.scss
```diff
@@ -0,0 +1,41 @@
...
+    flex: 1;
+  }
+
+  &:before {
+    content: "";
+    position: absolute;
+    top: 0;
+    left: 0;
+    width: 100%;
+    height: 100%;
```
Vinnl, line 36: I think it's because of this, but the entire top banner is unclickable for me. Any reason this couldn't be a background image on the container?

src/app/(proper_react)/DashboardTopBanner.module.scss
```diff
@@ -0,0 +1,41 @@
...
+    display: flex;
+    flex-direction: column;
+    gap: $spacing-sm;
+
+    h3 {
+      font: $text-title-2xs;
+    }
+
+    .cta {
+      margin-top: $spacing-sm;
```
Vinnl, line 21: Nit, but I'd prefer a `gap` on the parent if possible over a margin on the child, in the spirit of [composable CSS](https://non-traditional.dev/encapsulated-css-the-key-to-composable-layouts-94f11c177cc1): if we consistently keep parents responsible for layout their children, rather than children repositioning themselves sometimes, that makes it easier to debug issues, to move things around, and to understand the flow of our styles.

src/app/(proper_react)/DashboardTopBanner.module.scss
```diff
@@ -0,0 +1,41 @@
...
+      align-self: start;
+    }
+  }
+
+  & > * {
+    flex: 1;
```
Vinnl, line 27: I'd either set all properties with the short-hand, or only the property you're intending to set, to make it easier to understand for readers. That said, I'm not sure what effect this line has if it's applied to every child?

```suggestion
    flex-grow: 1;
```

src/app/(proper_react)/Dashboard.tsx
Vinnl, line 1: There's already a page for the dashboard in `(proper_react)/redesign/(authenticated)/user/dashboard/page.tsx` where you can add things under `<nav>`, could you build on that?

I expect that we'll probably want to use a pattern where we have a `page.tsx` that fetches the data and passes it to, say, a `<View>` component in the same directory that actually renders the page - that component could then be shown in a Storybook story, wrapped in a `<Shell>`. I can scaffold that out for you if you like.

Vinnl (commented):
I did a cursory look of things that jumped out to me, but there are a lot of changes in this PR, and there's code that went in with the setting-up-Storybook PR that hasn't really been reviewed as standalone code yet, so it was hard for me to get a feel for how it all fits together. It might be worth considering splitting things up into smaller PRs that are easier to land individually. Now I feel bad for springing a whole bunch of comments on you and still being afraid of missing things :/

There are also two accessibility violations listed on Storybook that you'll probably want to resolve:

image

(When I set up Jest, I'll integrate these checks into the test suite too, so then they'd break CI.)

pendingTranslations.ftl
```diff
@@ -11,3 +11,27 @@ main-nav-link-faq-tooltip = Frequently asked questions
...
+# Variables:
+# $exposure_type is the number of exposures found for a particular type, e.g. 3 found 
+exposure-card-num-found = { $exposure_num } found
+exposure-card-description-info-for-sale-part-one = This site is selling and publishing <data_broker_link>details about you</data_broker_link>
+exposure-card-description-info-for-sale-part-two = Remove this profile to protect your privacy.
+# Variables:
+# $data_breach_company is the company assosciated with the data breach.
+# $data_breach_date is the date of the data breach.
+exposure-card-description-data-breach-part-one = Your information was exposed in the <data_breach_link>{ $data_breach_company } data breach on { $data_breach_date }</data_breach_link>
+exposure-card-description-data-breach-part-two = We'll walk you through the steps to fix it.
```
pdehaan, line 36: nit: use smart quote for `We'll`

Not sure if we want to modify /docs/CODEOWNERS to include @flodolo as a codeowner for /pendingTranslations.ftl too. I'll let Flod decide in case it's too much noise to review pending vs waiting for them to land in locales/en/*.ftl.

 https://github.com/mozilla/blurts-server/blob/75e7d70988ddc122e2882feaa69aa807d0218819/docs/CODEOWNERS#L1-L2

pendingTranslations.ftl
```diff
@@ -11,3 +11,27 @@ main-nav-link-faq-tooltip = Frequently asked questions
...
+# Status Pill
+
+status-pill-action-needed = Action Needed
+status-pill-progress = In Progress
+status-pill-fixed = Fixed
+
+# Exposure Card
+
+exposure-card-family-members = Family members
+exposure-card-email = Type
```
pdehaan, line 24: Q: should this be "Type" or "Email"?

Storybook:
Screenshot 2023-06-26 at 4 27 04 PM

Figma:
https://www.figma.com/file/CaEKIhvSJqf6KNIMzSkt40/Concepts-for-Monitor-MVP-Redesign?type=design&node-id=546-44852&mode=design&t=bk52u2JrpN1imc6f-0

src/app/components/client/ExposureCard.tsx
```diff
@@ -132,43 +146,69 @@ export const ExposureCard = (props: Props) => {
...
+          // Data broker content
+          <div>
+            <p>
+              {l10n.getFragment("exposure-card-description-info-for-sale-part-one", 
+              {
+                elems: {"data_broker_link": <a href={props.exposureDetailsLink}/>}
+              })}
+              <a href={exposureDetailsLink}>
+                <span>
+                  <OpenInNew alt="" width="13" height="13" />
```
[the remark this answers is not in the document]
pdehaan: I think the empty `alt=` and `aria-label=` attributes are throwing some warnings in Storybook re: a11y

Screenshot 2023-06-26 at 5 05 57 PM

pendingTranslations.ftl
flodolo: The linter should be also extended to check for these files
https://github.com/mozilla/blurts-server/blob/main/.github/workflows/reference_linter.yaml

pendingTranslations.ftl
```diff
@@ -11,3 +11,27 @@ main-nav-link-faq-tooltip = Frequently asked questions
...
+exposure-card-phone-number = Phone number
+exposure-card-address = Address
+# Variables:
+# $exposure_type is the number of exposures found for a particular type, e.g. 3 found 
+exposure-card-num-found = { $exposure_num } found
```
flodolo, line 29:
```suggestion
exposure-card-num-found = 
  { $exposure_num ->
     *[other] { $exposure_num } found
  }
```

Technically identical for English, but this would trigger the plural UI in Pontoon (if we end up exposing to other locales).

pendingTranslations.ftl
```diff
@@ -11,3 +11,27 @@ main-nav-link-faq-tooltip = Frequently asked questions
...
+exposure-card-description-info-for-sale-part-one = This site is selling and publishing <data_broker_link>details about you</data_broker_link>
+exposure-card-description-info-for-sale-part-two = Remove this profile to protect your privacy.
+# Variables:
+# $data_breach_company is the company assosciated with the data breach.
+# $data_breach_date is the date of the data breach.
+exposure-card-description-data-breach-part-one = Your information was exposed in the <data_breach_link>{ $data_breach_company } data breach on { $data_breach_date }</data_breach_link>
+exposure-card-description-data-breach-part-two = We'll walk you through the steps to fix it.
+exposure-card-your-exposed-info = Your exposed info
```
flodolo, line 37: Always keep an empty line at the end of files.

src/app/(proper_react)/DashboardTopBanner.tsx
```diff
@@ -0,0 +1,98 @@
...
+      description: "We can now find exposures of your personal info on 190 data broker sites that publish and sell your personal info for a profit.",
+      cta: {
+        content: "Get first scan free",
+        onClick: () => { // do something 
+        }
+      }
+    },
+    NoExposuresFoundContent: {
+      headline: "No exposures found",
+      description: "Great news! We searched all known data breaches and 190 data broker sites that sell personal info and found no exposures. Upgrade to premium and we'll monitor for any new exposures.",
```
flodolo, line 46: premium or Premium?

src/app/(proper_react)/DashboardTopBanner.tsx
```diff
@@ -0,0 +1,98 @@
...
+      cta: {
+        content: "Get continuous protection",
+        onClick: () => { // do something 
+        }
+      }
+    },
+    ResumeBreachResolutionContent: {
+      headline: "Let's keep protecting your data",
+      description: `You still have ${props.data?.exposures?.remaining} exposures left to fix. Keepp going and protect yourself. We'll guide you step by step.`,
```
flodolo, line 55:
```suggestion
      description: `You still have ${props.data?.exposures?.remaining} exposures left to fix. Keep going and protect yourself. We'll guide you step by step.`,
```

src/app/components/client/ExposureCard.tsx
flodolo: If we ever expose this, by the time it happens we'll forget about this hard-coded locale, so I support Peter's suggestion.

src/app/components/client/ExposureCard.tsx
```diff
@@ -132,43 +146,69 @@ export const ExposureCard = (props: Props) => {
...
+                </span>
+              </a>
+              {". "}{l10n.getString("exposure-card-description-info-for-sale-part-two")}
```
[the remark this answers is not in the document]
flodolo: Absolutely. Never hard-code spaces or punctuation sign. This should be one string, not two strings glued together.

src/app/components/client/ExposureCard.tsx
```diff
@@ -132,43 +146,69 @@ export const ExposureCard = (props: Props) => {
...
+                </span>
+              </a>
+              {". "}{l10n.getString("exposure-card-description-data-breach-part-two")}
+            </p>
+          </div>
+          }
           <div className={styles.exposureListOfExposureTypes}>
             <ul>
-              <li>Your exposed info:</li>
+              <li>{l10n.getString("exposure-card-your-exposed-info")}:</li>
```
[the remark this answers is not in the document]
flodolo: Or just French…

flodolo (commented):
Please don't hardcode characters, even if this is just en-US for now (it might not be if we decide to add Spanish)
https://mozilla-l10n.github.io/documentation/localization/dev_best_practices.html

Also, why the mix of strings exposed in FTL and hard-coded strings (for banner). I only looked quickly at the latter, but punctuation needs fixing, and there might be more errors.

src/app/components/client/ExposureCard.tsx
```diff
@@ -132,48 +134,74 @@ export const ExposureCard = (props: Props) => {
...
+              </a>
+              {". "}{l10n.getString("exposure-card-description-data-breach-part-two")}
+            </p>
+          </div>
+          }
           <div className={styles.exposureListOfExposureTypes}>
-            <ul>
-              <li>Your exposed info:</li>
-              <li>
+            <dl>
```
pdehaan, line 174: I don't think `<dl>` is the right tag here since this isn't really a list (and we don't have nested child `<dt>` and `<dd>` elements).

src/app/components/client/ExposureCard.tsx
```diff
@@ -132,48 +134,74 @@ export const ExposureCard = (props: Props) => {
...
+                  icon={<LocationPin alt="" width="13" height="13" />}
                   whichExposed="address"
                   num={0}
                 />
-              </li>
-            </ul>
+              </div>
+            </dl>
             <span className={styles.fixItBtn}>
               <Button type={"primary"} content={"Lets fix it"} />
```
pdehaan, line 206: nit: this should be localized

src/app/components/server/ExposureType.tsx
```diff
@@ -0,0 +1,23 @@
...
+
+export const ExposureTypeEl = (props: Props) => {
+  const { type } = props;
+
+  let string = "";
+
+  if (type === "infoForSale") {
+    string = "Info for sale"
+  } else if (type === "dataBreach") {
+    string = "Data breach"
```
pdehaan, line 19: I assume these two `string`s should be localized too.

src/app/components/client/ExposureCard.module.scss
Vinnl: You can see them next to the company names at https://monitor.firefox.com/breaches.

src/app/(proper_react)/DashboardTopBanner.module.scss
Vinnl: Right, so then the values that are implicitly set by leaving them out (specifically: `flex-shrink: 1`) are relevant as well - so I'd still set those explicitly, so it's clear to the reader that you're doing that. So I'd suggest either

```suggestion
    flex: 1 1 auto;
```

 (assuming `flex-basis` doesn't actually have to be 0)

or

```suggestion
    flex-grow: 1;
    flex-shrink: 1;
```

src/app/components/client/ExposureCard.module.scss
```diff
@@ -56,56 +65,92 @@
...
+        @media screen and (min-width: $screen-md) {
           display: flex;
+          flex-direction: row;
+          align-items: center;
+        }
 
-          svg {
-            color: $color-purple-70;
+        div + div {
+         padding-left: 23px; // restore the offset of the exposure icon which is -23px
```
pdehaan, line 112: `23px` feels a bit magic-numbery. Should we convert it to a variable?

```sh
git grep -En "\-?23px" src | cat

src/app/components/client/ExposureCard.module.scss:112:         padding-left: 23px; // restore the offset of the exposure icon which is -23px
src/app/components/client/ExposureCard.module.scss:134:              margin-left: -23px; // offset the exposure icon
```

src/app/components/client/ExposureCard.tsx
```diff
@@ -132,48 +122,66 @@ export const ExposureCard = (props: Props) => {
...
+          {isScanResult(props.exposureType) ? 
+          // Data broker content
+          <div>
+            <p>
+              {l10n.getFragment("exposure-card-description-info-for-sale-part-one", 
+              {
+                elems: {"data_broker_link": <a href={exposureDetailsLink}/>}
+              })}
+              <a href={exposureDetailsLink}>
+                <span>
```
pdehaan, line 134: Q: does this need the wrapper `<span>`?

src/app/components/client/ExposureCard.tsx
```diff
@@ -132,48 +122,66 @@ export const ExposureCard = (props: Props) => {
...
+              {l10n.getFragment("exposure-card-description-data-breach-part-one", 
+              {
+                vars: {
+                  data_breach_company: exposureName,
+                  data_breach_date: props.exposureType.BreachDate
+                },
+                elems: {"data_breach_link": <a href={exposureDetailsLink}/>}
+              })}
+              <a href={exposureDetailsLink}>
+                <span>
```
pdehaan, line 153: Same Q here re: possibly unneeded `<span>`.

And not sure which PR it was in, but I believe @flodolo shared some concerns about these strings basically being 2-3 strings glued together, which doesn't often play well w/ localizering. (applies to this block/div and others above where we seem to be splitting into `*-part-one` and `*-part-two`).

src/app/components/client/ExposureCard.tsx
```diff
@@ -132,48 +122,66 @@ export const ExposureCard = (props: Props) => {
...
-                  icon={<LocationPin alt="" width="15" height="15" />}
-                  whichExposed="address"
-                  num={0}
-                />
-              </li>
-            </ul>
+            <dl>
+              <div>{l10n.getString("exposure-card-your-exposed-info")}:</div>
+              <DetailsFoundItem
+                icon={<MultipleUsers alt="" width="13" height="13" />}
```
pdehaan, line 165: Not sure if this is even worth of a nit or not, but not sure if we should rename anything for consistency:

- `<MultipleUsers />`
- `<PhoneIcon />`
- `<EmailIcon />`
- `<LocationPin />`

Not sure if first and last should have `Icon` suffixes or not.

src/app/components/client/ExposureCard.tsx
pdehaan: The "Lets fix it" looks hardcoded though, and Ctrl+Fing doesn't seem to show any results in any .ftl files (at least not on this page).

```sh
git grep -in "Lets fix it" | cat

src/app/components/client/ExposureCard.tsx:186:              <Button type={"primary"} content={"Lets fix it"} />
```

src/app/components/client/stories/ExposureCard.stories.ts
```diff
@@ -3,26 +3,83 @@
...
+  IsSpamList: false,
+  IsVerified: false,
+  LogoPath: "",
+  ModifiedDate: "",
+  Name: "",
+  PwnCount: 0,
+  recencyIndex: 0,
+  ResolutionsChecked: [],
+  Title: "Twitter"
+}
```
pdehaan, line 62: Same Q: as above in src/app/(proper_react)/Dashboard.stories.tsx on whether we need mocks for this stuff, or if that's a followup ticket.

src/app/components/server/ExposureType.tsx
```diff
@@ -0,0 +1,32 @@
...
+    string = l10n.getString("exposure-card-exposure-type-data-breach");
+  }
+
+  return <>{string}</>;
+};
+
+// Typeguard function
+export function isScanResult(obj: ScanResult | Breach): obj is ScanResult {
+  return (obj as ScanResult).data_broker !== undefined; // only ScanResult has an instance of data_broker
+}
```
pdehaan, line 32: nit, needs trailing linebreak.

src/app/components/server/Icons.tsx
```diff
@@ -197,7 +197,7 @@ export const EmailIcon = ({
       aria-label={alt}
       aria-hidden={alt === ""}
       xmlns="http://www.w3.org/2000/svg"
-      viewBox="0 0 14 14"
+      viewBox="0 0 15 15"
```
pdehaan, line 200: Why did this grow?
Might be moot since I think the icons in this file are several different widths and heights, per:
```sh
git grep -Pno 'viewBox=".*?"' src/app/components/server/Icons.tsx | cat

src/app/components/server/Icons.tsx:23:viewBox="0 0 16 14"
src/app/components/server/Icons.tsx:46:viewBox="0 0 16 16"
src/app/components/server/Icons.tsx:69:viewBox="0 0 18 18"
src/app/components/server/Icons.tsx:96:viewBox="0 0 12 12"
src/app/components/server/Icons.tsx:119:viewBox="0 0 13 7"
src/app/components/server/Icons.tsx:139:viewBox="0 0 15 14"
src/app/components/server/Icons.tsx:160:viewBox="0 0 16 16"
src/app/components/server/Icons.tsx:180:viewBox="0 0 14 14"
src/app/components/server/Icons.tsx:200:viewBox="0 0 15 15"
src/app/components/server/Icons.tsx:221:viewBox="0 0 11 14"
src/app/components/server/Icons.tsx:241:viewBox="0 0 13 13"
src/app/components/server/Icons.tsx:267:viewBox="0 0 16 16"
src/app/components/server/Icons.tsx:290:viewBox="0 0 24 24"
```

But since this says EmailIcon and I vaguely recall some *Icon questions above&hellip;

- `<MultipleUsers />` &mdash; `viewBox="0 0 16 16"`
- `<PhoneIcon />` &mdash; `viewBox="0 0 14 14"`
- `<EmailIcon />` &mdash; `viewBox="0 0 15 15"`
- `<LocationPin />` &mdash; `viewBox="0 0 11 14"`

pendingTranslations.ftl
```diff
@@ -22,3 +22,77 @@ toolbar-app-picker-by-mozilla = Made by { -brand-mozilla }
...
+exposure-card-exposure-type-data-broker = Info for sale
+exposure-card-exposure-type-data-breach = Data breach
+
+# Dashboard
+
+dashboard-top-banner-protect-your-data-title = Let's protect your data
+# Variables:
+# $data_breach_total_num is the total number of data breaches the user has.
+# $data_broker_total_num is the total number of data brokers selling the user's data.
+dashboard-top-banner-protect-your-data-description = We found your data in { $data_breach_total_num } data breaches and { $data_broker_total_num } sites selling your personal info. We'll guide you on how to fix it.
```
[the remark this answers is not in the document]
flodolo: This string works for English, but it's going to be impossible to localize in many languages (there is no support for multiple plurals). It's also broken in English if any of these is singular.

If possible, I would suggest content to find an alternative solution now, instead of trying to come up with something later on.

pendingTranslations.ftl
```diff
@@ -22,3 +22,77 @@ toolbar-app-picker-by-mozilla = Made by { -brand-mozilla }
...
+dashboard-top-banner-protect-your-data-cta = Let's fix it
+
+dashboard-top-banner-monitor-protects-your-even-more-title = Monitor now protects you even more
+# Variables:
+# $data_broker_sites_total_num is the total number of data broker sites available to scan.
+dashboard-top-banner-monitor-protects-your-even-more-description = We can now find exposures of your personal info on { $data_broker_sites_total_num } data broker sites that publish and sell your personal info for a profit.
```
flodolo, line 72: Same as the previous comment. If there is a plural, please use the structure with `*[other]` to trigger the plural UI (there are multiple strings in this patch).

flodolo (dismissed):

src/app/components/client/ExposureCard.tsx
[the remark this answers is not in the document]
flodolo: Yes, punctuation should always be exposed as part of the string, not hardcoded in the markup.

src/app/components/client/ExposureCard.tsx
flodolo: It really doesn't, quite the opposite. You're enforcing the English structure of the sentence on all locales, and hard-coding characters to do it (a colon here, a space after).

src/app/(proper_react)/Shell.module.scss
```diff
@@ -3,12 +3,17 @@
...
+  @media screen and (min-width: $screen-lg) {
+    gap: $layout-md;
+  } 
 }
 
 .mainMenu {
   display: none;
 
-  @media screen and (min-width: $screen-md) {
+  @media screen and (min-width: $screen-lg) {
```
Vinnl, line 16: Hmm, with just this change, the sidebar menu already disappears when the viewport width is smaller than `$screen-lg`, but the mobile menu only shows up when it's smaller than `$screen-md` - so in between, there's no way to navigate the site. I'm not sure why this change was made, but either we'll need an equivalent change in `MobileShell.module.scss`, or to revert this one.

(I'd lean towards the latter, as the mobile toolbar looks rather weird on screens as wide as `$screen-lg` - feels like there should be plenty of room for a sidebar. Instead, we probably want the Exposure Cards to switch to their thin layouts more quickly - either using [Container Queries](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Container_Queries) (not supported in older Safari) or using `flex-wrap`.)

src/app/(proper_react)/DashboardTopBanner.module.scss
```diff
@@ -0,0 +1,32 @@
+@import "../tokens";
+
+.container {
+  display: flex;
+  flex-direction: row;
+  position: relative;
```
Vinnl, line 6: In case you hadn't noticed, `position` is always a red flag for me :P But I think it's just a leftover in this case, right?

```suggestion
```

src/app/components/client/ExposureCard.module.scss
Vinnl: Ah OK, so you want the `dd` to be aligned with the heading, not the icon. In that case, I'd wrap the icon in a container with a fixed width (preferably defined in a Sass variable, to explain the magic number), and then add a `padding-inline-start` to the `<dd>` of the same size. That avoids much of the potential headaches of absolute positioning down the road, while still achieving the desired layout.

src/app/components/client/ExposureCard.module.scss
```diff
@@ -56,56 +65,92 @@
...
+
+.hideOnMobile {
+  display: none !important;
+  visibility: hidden;
+
+  @media screen and (min-width: $screen-md) {
+    display: block !important;
+    visibility: visible;
+  }
+}
```
Vinnl, line 156: It's also highly recommended to avoid `!important` - there's no coming back from that, as it can't be overridden by more `!important`s. Instead, since this is only applying to a single type of elements anyway, you could instead add this to the `li` of line 43:

```scss
      @media screen and (max-width: $screen-md) {
        &.hideOnMobile {
          display: none;
        }
      }
```

(`visibility:hidden` is also unnecessary when you already have `display: none`.)

That way, the selector is more specific than the plain `li` that sets the `display: flex`, and thus it can override it - and as an added bonus, future readers can infer that from the context.

src/app/components/client/ExposureCard.tsx
```diff
@@ -121,7 +111,7 @@ export const ExposureCard = (props: Props) => {
           >
             <ChevronDown
               className={detailsOpen ? styles.isOpen : ""}
-              alt=""
+              alt={l10n.getString("chevron-down-alt")}
```
Vinnl, line 114: This now says "Drop down button", which doesn't provide much info about what it does. More descriptive would be something like "Expand" or "Collapse", depending on the current state. In fact, might be useful to add that as a `title` too.

src/app/components/client/ExposureCard.tsx
```diff
@@ -18,101 +18,91 @@ import {
...
-      </>
+        <dd>{description}</dd>
+      </div>
     );
   };
   const elementCard = (
     <div>
       <div className={styles.exposureCard}>
         <div className={styles.exposureHeader}>
           <ul className={styles.exposureHeaderList}>
```
Vinnl, line 94: I missed this the first time (annoying, I know...), but imagining this list being read out to someone, I don't think it would make much sense to them. Could you convert it into a `<dl>` with the `<dt>`s containing descriptions of every heading (e.g. "Exposure name", "Date found", etc. - we should be able to use the same labels that we'll use as titles above the cards), and then visually hide the `<dt>`s? See [here](https://github.com/mozilla/blurts-server/blob/6422ebf1af7a8529289a30201313c3af41ef8e78/src/app/components/client/Chart.module.scss#L52-L63) for an example of how to visually hide something while keeping it available to screen readers.

src/app/(proper_react)/Dashboard.module.scss
```diff
@@ -0,0 +1,25 @@
...
+    margin-top: $layout-sm;
+
+    .exposureList {
+      list-style: none;
+      padding-top: $layout-xs;
+      padding-left: 0;
+      margin: 0;
+  
+      .exposureListItem {
+        margin-bottom: $spacing-md;
```
Vinnl, line 21: Referring to [composable CSS](https://non-traditional.dev/encapsulated-css-the-key-to-composable-layouts-94f11c177cc1) again, I think a good rule of thumb is that you almost always want to avoid `margin` if you can. So instead of this `margin-bottom`, you'd make the parent a flex vertical container with a `gap` and a `padding-block` if need be. Likewise for the margin on line 12.

Vinnl (commented):
Thanks for doing all this - it's a big chunk of work (and as you can tell, I can hardly keep up with reviews :grin: ). I know I can be a pain sometimes, so thanks for sticking with it. But we're getting pretty close! I've still got a couple of comments left, and the strings need some work, but we can't be far off now.

pendingTranslations.ftl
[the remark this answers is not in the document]
flodolo: Correct. As far as I can tell, there isn't any localization framework that supports that.

If you think about it, things get quickly out of hand with 6 possible plurals for each variable (I forgot everything about statistics, but hopefully the math is right):
* 1 plural -> 6 variants
* 2 plurals -> 36 variants

While English only has `one` and `other`, it gets much more complicated for Slavic languages.
https://www.unicode.org/cldr/charts/43/supplemental/language_plural_rules.html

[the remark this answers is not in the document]
flodolo: For the sake of completeness, technically Fluent supports multiple plurals in one string, but that's deprecated (Message Format 2 doesn't support them, and we want each plural variant to be a fully formed sentence).

This is pretty much impossible for a non-technical person to translate, and you'd have 6 variants instead of 2 for each plural.

```
dashboard-top-banner-protect-your-data-description = 
    We found your data in 
    { $data_breach_total_num ->
        [one] { $data_breach_total_num } data breach
       *[other] { $data_breach_total_num } data breaches
    } 
    and 
    { $data_broker_total_num ->
        [one] { $data_broker_total_num } site
       *[other] { $data_broker_total_num } sites
    } 
    selling your personal info. We’ll guide you on how to fix it.
```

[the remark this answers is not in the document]
flodolo: It's not ideal (you'd need to "glue" them together with spaces), but it would at least work.

src/app/components/client/ExposureCard.tsx
```diff
@@ -132,50 +149,68 @@ export const ExposureCard = (props: Props) => {
...
+            </p>
+          </div> : 
+          // Data breach content
+          <div>
+            <p>
+              {l10n.getFragment("exposure-card-description-data-breach-part-one", 
+              {
+                vars: {
+                  data_breach_company: exposureName,
+                  data_breach_date: exposureData.BreachDate
```
pdehaan, line 175: Q: Is this date localized somewhere, or is it the raw value from HIBP?

If they aren't localized, what date is 07/10/2020?

<details>
<summary>Spoiler alert</summary>

```js
const d = new Date("07/10/2020");
console.log(d.toString()); // 'Fri Jul 10 2020 00:00:00 GMT-0700 (Pacific Daylight Time)'
console.log("LOCALE | DATE\n:----:|----");
for (const locale of locales) {
  console.log(`${locale} | ${ new Intl.DateTimeFormat(locale).format(d)}`);
}
```

LOCALE | DATE
:----:|----
be | 10.7.2020
cs | 10. 7. 2020
cy | 10/7/2020
da | 10.7.2020
de | 10.7.2020
dsb | 10.7.2020
el | 10/7/2020
en-CA | 7/10/2020
en-GB | 10/07/2020
es-AR | 10/7/2020
es-CL | 10-07-2020
es-ES | 10/7/2020
es-MX | 10/7/2020
fi | 10.7.2020
fr | 10/07/2020
fy-NL | 10-7-2020
gn | 7/10/2020
he | 10.7.2020
hsb | 10.7.2020
hu | 2020. 07. 10.
ia | 10-07-2020
is | 10.7.2020
it | 10/7/2020
ja | 2020/7/10
ka | 10.7.2020
kab | 10/7/2020
kk | 10.07.2020
ko | 2020. 7. 10.
nl | 10-7-2020
nn-NO | 10.7.2020
pa-IN | 10/7/2020
pl | 10.07.2020
pt-BR | 10/07/2020
pt-PT | 10/07/2020
rm | 10-07-2020
ru | 10.07.2020
sk | 10. 7. 2020
sl | 10. 7. 2020
sq | 10.7.2020
sr | 10. 7. 2020.
sv-SE | 2020-07-10
tr | 10.07.2020
uk | 10.07.2020
vi | 10/7/2020
zh-CN | 2020/7/10
zh-TW | 2020/7/10

```

File: src/app/components/client/ExposureCard.tsx (hunk 46 above)
Comment (Vinnl) [reply src/app/components/client/ExposureCard.tsx]: I think the same answer as [here](https://github.com/mozilla/blurts-server/pull/3155#discussion_r1247678093) applies: yes, but it gets set by the `alt` prop for icons.

One thing I also noticed while inspecting the chevron in the Accessibility Inspector is this warning:

image

So I think the `<span>` that has the `onClick` handler should actually be a `<button>`.

File: src/app/components/client/ExposureCard.tsx (hunk 47 above)
Comment (Vinnl) [reply src/app/components/client/ExposureCard.tsx:120]: I meant

```html
<dl>
  <dt class="visually-hidden">Breached Company</dt>
  <dd>Familytree.com</dd>
  <dt class="visually-hidden">Date found</dt>
  <dd>11/09/23</dd>
</dl>
```

rather than the current structure:

```html
<ul>
  <li>Familytree.com</li>
  <li>11/09/23</li>
</ul>
```

src/app/(proper_react)/Dashboard.tsx
Vinnl: As discussed in the frontenders pairing just now, let's skip this for now until we're actually fetching data. (I thought this was intended to actually be visible on the site already.)

src/app/components/client/ExposureCard.tsx
```diff
@@ -57,7 +57,7 @@ export type ExposureCardProps = {
 };
 
 type DetailsFoundProps = {
-  whichExposed: string; // family | email | phone | address | credit card | password
+  whichExposed: string; // family | email | phone | address | creditcard | password
```
pdehaan, line 59: Instead of generic "string" should we inline the valid values?
```suggestion
  whichExposed: "family" | "email" | "phone" | "address" | "creditcard" | "password";
```

src/app/(proper_react)/Shell.module.scss
Vinnl:
```suggestion
  @media screen and (min-width: $screen-md) {
```

Otherwise there's no navigation on some screen sizes, and that breakage is easier to miss than the exposure cards not fitting:

image

Vinnl (approved):
If you apply [this change](https://github.com/mozilla/blurts-server/pull/3155#discussion_r1252825561), I think it's good to go for now.

With the move of `pendingTranslations.ftl`, you'll have to move the strings into `locales-pending/premium.ftl`, which I think might even happen automatically if you merge `main` into this branch. I can also do it for you if you want.

PR #3157, opened 2023-06-22
MNTOR-1403/data removal action and scan

rhelmer:
# References: 
Jira: MNTOR-1403

# Description

Basic UI for testing broker scanning and removal.

# How to test

Using the sandbox key (which always returns fabricated results):

1. Make sure to run DB migrations to add new tables and columns (`npm run db:migrate`)
2. Visit the "Welcome" flow at /redesign/user/welcome/
3. Proceed through free scan steps, and verify that screens defined in Figma appear
4. Verify that (fabricated) scan results are shown, and offer to subscribe to premium is offered
5. Subscribe to premium and verify that removal can be requested
6. Verify that removal succeeds
7. Unsuscribe from Premium and verify that data remains, but no new scans are performed and data removal cannot be requested for new scans
7a. NOTE: to set up this test, we need to figure out how to avoid waiting for a month for the cancellation to take effect, and we also will want to coordinate loading new scan results to verify that they do not appear for lapsed subscription customers)

If you'd like to re-test, you'll need to either log in as a new user, or remove the scan history from the `onerep_scans` table - the system only allows one scan currently. 

- [x] Prerequisite PRs have landed (https://github.com/mozilla/blurts-server/pull/3166)

### DIFF
--- a/public/nextjs_migration/client/js/welcome.js
+++ b/public/nextjs_migration/client/js/welcome.js
@@ -1,0 +1,35 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+function init() {
+  let intervalCount = 1;
+
+  const progress = document.querySelector("#progress");
+  let results;
+
+  const interval = setInterval(async () => {
+    if (!results) {
+      const response = await fetch("/api/v1/user/welcome");
+      const body = await response.json();
+
+      if (response.ok && body.scan_results[0].onerep_scan_results !== null) {
+        results = body.scan_results[0].onerep_scan_results;
+      }
+    }
+
+    const percentComplete = intervalCount * 6;
+
+    if (percentComplete >= 100) {
+      progress.textContent = "Scan complete";
+      clearInterval(interval);
+
+      window.location = "results";
+    } else {
+      progress.textContent = `${percentComplete}%`;
+      intervalCount++;
+    }
+  }, 1000);
+}
+
+init();
--- a/src/app/(nextjs_migration)/(authenticated)/user/breaches/page.tsx
+++ b/src/app/(nextjs_migration)/(authenticated)/user/breaches/page.tsx
@@ -22,6 +22,7 @@ import ImageIconEmail from "../../../../../client/images/icon-email.svg";
 import { BreachesTable } from "../../../components/server/BreachesTable";
 import { getComponentAsString } from "../../../functions/server/getComponentAsString";
 import { getCountryCode } from "../../../../functions/server/getCountryCode";
+import { isUserSubscribed } from "../../../../functions/server/isUserSubscribed";
 
 export function generateMetadata() {
   const l10n = getL10n();
@@ -88,43 +89,6 @@ export default async function UserBreaches() {
     }>;
   };
 
-  // Fetch list of subscriptions.
-  let subscriptions;
-  const bearerToken = session?.user.subscriber?.fxa_access_token;
-  if (bearerToken) {
-    const result = await fetch(
-      `${process.env
-        .OAUTH_API_URI!}/oauth/mozilla-subscriptions/customer/billing-and-subscriptions`,
-      {
-        headers: {
-          "Content-Type": "application/json",
-          authorization: `Bearer ${bearerToken}`,
-        },
-      }
-    );
-    if (result.ok) {
-      subscriptions = await result.json();
-    }
-  } else {
-    console.error("User has no bearer token");
-  }
-
-  const isUserSubscribed = (subscriptions: FxaSubscriptionResponse) => {
-    if (!subscriptions) {
-      return false;
-    }
-    for (const subscription of subscriptions.subscriptions) {
-      if (
-        subscription.product_id === process.env.PREMIUM_PRODUCT_ID &&
-        subscription.plan_id === process.env.PREMIUM_PLAN_ID_US &&
-        subscription.status === "active"
-      ) {
-        return true;
-      }
-    }
-    return false;
-  };
-
   return (
     <>
       {/* These scripts predate the use of React and thus shouldn’t wait for
@@ -153,7 +117,7 @@ export default async function UserBreaches() {
 
       <main data-partial="breaches">
         {process.env.PREMIUM_ENABLED === "true" &&
-        !isUserSubscribed(subscriptions) ? (
+        !(await isUserSubscribed()) ? (
           <section>
             <a
               className="button primary"
--- a/src/app/(proper_react)/redesign/(authenticated)/user/welcome/details/page.tsx
+++ b/src/app/(proper_react)/redesign/(authenticated)/user/welcome/details/page.tsx
@@ -1,0 +1,95 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import { isEligible } from "../../../../../../functions/server/onerep";
+import Script from "next/script";
+
+export function generateMetadata() {
+  return {
+    title: "Welcome - Details",
+  };
+}
+
+export default async function UserWelcomeDetails() {
+  const eligible = await isEligible();
+
+  if (!eligible) {
+    return (
+      <main>
+        <h2>You have already used your free scan.</h2>
+      </main>
+    );
+  } else {
+    return (
+      <>
+        <Script
+          type="module"
+          src="/nextjs_migration/client/js/welcome-details.js"
+        ></Script>
+
+        <main>
+          <h1>Enter the details you want to protect</h1>
+          <p>
+            We’ll use this to find exposures of your personal information, and
+            then guide you step-by-step on how to fix it.
+          </p>
+
+          <section>
+            <form name="details">
+              <table>
+                <tbody>
+                  <tr>
+                    <td>
+                      <label htmlFor="firstname">First name: </label>
+                    </td>
+                    <td>
+                      <input type="text" name="firstname" id="firstname" />
+                    </td>
+                  </tr>
+                  <tr>
+                    <td>
+                      <label htmlFor="lastname">Last name: </label>
+                    </td>
+                    <td>
+                      <input type="text" name="lastname" id="lastname" />
+                    </td>
+                  </tr>
+                  <tr>
+                    <td>
+                      <label htmlFor="city">City: </label>
+                    </td>
+                    <td>
+                      <input type="text" name="city" id="city" />
+                    </td>
+                    <td>
+                      <label htmlFor="state">State: </label>
+                    </td>
+                    <td>
+                      <input type="text" name="state" id="state" />
+                    </td>
+                  </tr>
+                  <tr>
+                    <td>
+                      <label htmlFor="dob">Date of birth: </label>
+                    </td>
+                    <td>
+                      <input type="date" name="dob" id="dob" />
+                    </td>
+                  </tr>
+                  <tr>
+                    <td>
+                      <button type="submit" className="button primary">
+                        Start my free scan
+                      </button>
+                    </td>
+                  </tr>
+                </tbody>
+              </table>
+            </form>
+          </section>
+        </main>
+      </>
+    );
+  }
+}
--- a/src/app/(proper_react)/redesign/(authenticated)/user/welcome/page.tsx
+++ b/src/app/(proper_react)/redesign/(authenticated)/user/welcome/page.tsx
@@ -1,0 +1,42 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import { isEligible } from "../../../../../functions/server/onerep";
+
+export function generateMetadata() {
+  return {
+    title: "Welcome",
+  };
+}
+
+export default async function UserWelcome() {
+  const eligible = await isEligible();
+
+  if (!eligible) {
+    return (
+      <main>
+        <h2>You have already used your free scan.</h2>
+      </main>
+    );
+  }
+
+  return (
+    <main>
+      <h1>Welcome to Monitor. Let’s find your exposed information.</h1>
+      <p>
+        To find and fix your exposed data, we’ll need to ask for your personal
+        info. Rest assured we will safeguard your information like it’s our own.
+      </p>
+
+      <p>
+        We believe in your right to privacy, so basic Monitor is always free.
+        Upgrade to Premium for continuous protection and automatic removal of
+        your personal info.
+      </p>
+      <a href="welcome/details">
+        <button className="button primary">Start my free scan</button>
+      </a>
+    </main>
+  );
+}
--- a/src/app/(proper_react)/redesign/(authenticated)/user/welcome/remove/page.tsx
+++ b/src/app/(proper_react)/redesign/(authenticated)/user/welcome/remove/page.tsx
@@ -1,0 +1,43 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import { getOnerepProfileId } from "../../../../../../../db/tables/subscribers";
+import { getServerSession } from "next-auth";
+import { authOptions } from "../../../../../../api/utils/auth";
+import {
+  activateProfile,
+  optoutProfile,
+} from "../../../../../../functions/server/onerep";
+import Script from "next/script";
+import { isUserSubscribed } from "../../../../../../functions/server/isUserSubscribed";
+
+export function generateMetadata() {
+  return {
+    title: "Welcome - Results",
+  };
+}
+
+export default async function UserWelcomeRemove() {
+  const session = await getServerSession(authOptions);
+  if (!session?.user?.subscriber?.id) {
+    throw new Error("No session");
+  }
+
+  if (process.env.PREMIUM_ENABLED !== "true" || !(await isUserSubscribed())) {
+    return (
+      <Script id="redirect">window.location = &quot;results&quot;;</Script>
+    );
+  }
+
+  const result = await getOnerepProfileId(session.user.subscriber.id);
+  const profileId = result[0]["onerep_profile_id"] as number;
+  await activateProfile(profileId);
+  await optoutProfile(profileId);
+
+  return (
+    <main>
+      <h2>Profile activated and removal has started</h2>
+    </main>
+  );
+}
--- a/src/app/(proper_react)/redesign/(authenticated)/user/welcome/results/page.tsx
+++ b/src/app/(proper_react)/redesign/(authenticated)/user/welcome/results/page.tsx
@@ -1,0 +1,61 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import { getOnerepProfileId } from "../../../../../../../db/tables/subscribers";
+import { getServerSession } from "next-auth";
+import { authOptions } from "../../../../../../api/utils/auth";
+import { getLatestOnerepScan } from "../../../../../../../db/tables/onerep_scans";
+import { isUserSubscribed } from "../../../../../../functions/server/isUserSubscribed";
+
+export function generateMetadata() {
+  return {
+    title: "Welcome - Results",
+  };
+}
+
+export default async function UserWelcomeResults() {
+  const session = await getServerSession(authOptions);
+  if (!session?.user?.subscriber?.id) {
+    throw new Error("No session");
+  }
+
+  const result = await getOnerepProfileId(session.user.subscriber.id);
+  const profileId = result[0]["onerep_profile_id"] as number;
+  const scanResult = await getLatestOnerepScan(profileId);
+
+  const results = JSON.stringify(scanResult, null, 2);
+
+  const ctaButton = async () => {
+    if (process.env.PREMIUM_ENABLED === "true") {
+      if (await isUserSubscribed()) {
+        return (
+          <a className="button primary" href="remove">
+            <button className="button primary">Remove my Data</button>
+          </a>
+        );
+      } else {
+        return (
+          <a
+            className="button primary"
+            href={`${process.env.FXA_SUBSCRIPTIONS_URL ?? ""}/products/${
+              process.env.PREMIUM_PRODUCT_ID ?? ""
+            }?plan=${process.env.PREMIUM_PLAN_ID_US ?? ""}`}
+          >
+            <button className="button primary">Subscribe to Premium</button>
+          </a>
+        );
+      }
+    } else {
+      return <></>;
+    }
+  };
+
+  return (
+    <main>
+      {ctaButton()}
+      <h2>Results</h2>
+      <pre>{results}</pre>
+    </main>
+  );
+}
--- a/src/app/(proper_react)/redesign/(authenticated)/user/welcome/scanning/page.tsx
+++ b/src/app/(proper_react)/redesign/(authenticated)/user/welcome/scanning/page.tsx
@@ -1,0 +1,100 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import { getOnerepProfileId } from "../../../../../../../db/tables/subscribers";
+import { getServerSession } from "next-auth";
+import { authOptions } from "../../../../../../api/utils/auth";
+import {
+  createScan,
+  listScanResults,
+  getScanDetails,
+  isEligible,
+} from "../../../../../../functions/server/onerep";
+import Script from "next/script";
+import {
+  setOnerepScan,
+  setOnerepScanResults,
+} from "../../../../../../../db/tables/onerep_scans";
+
+export function generateMetadata() {
+  return {
+    title: "Welcome - Scanning",
+  };
+}
+
+export default async function UserWelcomeScanning() {
+  const eligible = await isEligible();
+  if (!eligible) {
+    return (
+      <main>
+        <h2>You have already used your free scan.</h2>
+      </main>
+    );
+  }
+
+  const session = await getServerSession(authOptions);
+  if (!session?.user?.subscriber?.id) {
+    throw new Error("No session");
+  }
+
+  const result = await getOnerepProfileId(session.user.subscriber.id);
+  const profileId = result[0]["onerep_profile_id"] as number;
+  const scan = await createScan(profileId);
+  await setOnerepScan(profileId, scan.id);
+
+  let iterations = 0;
+  const totalIterations = 15;
+
+  const interval = setInterval(
+    () =>
+      void (async () => {
+        const scanDetails = await getScanDetails(profileId, scan.id);
+
+        // Give up after set number of iterations.
+        if (iterations >= totalIterations) {
+          clearInterval(interval);
+        } else if (scanDetails.status === "finished") {
+          clearInterval(interval);
+
+          const scanListFull = [];
+          const firstPage = await listScanResults(profileId, { per_page: 100 });
+          // Results are paginated, use per_page maximum and collect all pages into one result.
+          if (firstPage.meta.last_page > 1) {
+            let currentPage = 2;
+            while (currentPage <= firstPage.meta.last_page) {
+              const nextPage = await listScanResults(profileId, {
+                per_page: 100,
+              });
+              currentPage++;
+              nextPage.data.forEach((element: object) =>
+                scanListFull.push(element)
+              );
+            }
+          } else {
+            scanListFull.push(firstPage.data);
+          }
+          // Store full list of results in the DB.
+          await setOnerepScanResults(profileId, scan.id, {
+            data: scanListFull[0],
+          });
+        } else {
+          iterations++;
+        }
+      })(),
+    1000
+  );
+
+  return (
+    <>
+      <Script
+        type="module"
+        src="/nextjs_migration/client/js/welcome.js"
+      ></Script>
+      <main>
+        <h2 id="status">Scanning for exposures...</h2>
+        <div id="progress">0%</div>
+      </main>
+    </>
+  );
+}
--- a/src/app/api/v1/user/welcome/route.ts
+++ b/src/app/api/v1/user/welcome/route.ts
@@ -1,0 +1,94 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import { getServerSession } from "next-auth";
+import { authOptions } from "../../../../api/utils/auth";
+import { NextRequest, NextResponse } from "next/server";
+
+import { createProfile, isEligible } from "../../../../functions/server/onerep";
+import type { ProfileData } from "../../../../functions/server/onerep";
+import AppConstants from "../../../../../appConstants";
+import {
+  getOnerepProfileId,
+  getSubscriberByEmail,
+} from "../../../../../db/tables/subscribers";
+
+import {
+  getLatestOnerepScan,
+  setOnerepProfileId,
+} from "../../../../../db/tables/onerep_scans";
+import { setProfileDetails } from "../../../../../db/tables/onerep_profiles";
+import { StateAbbr } from "../../../../../utils/states";
+
+export async function POST(req: NextRequest) {
+  const session = await getServerSession(authOptions);
+
+  const eligible = await isEligible();
+  if (!eligible) {
+    throw new Error("User has no more manual scans left");
+  }
+
+  const params = await req.json();
+
+  const requiredParams = ["firstname", "lastname", "city", "state", "dob"];
+  requiredParams.forEach((param) => {
+    if (!params[param]) {
+      throw new Error(`${param} is required`);
+    }
+  });
+
+  const profileData: ProfileData = {
+    first_name: params.firstname,
+    last_name: params.lastname,
+    addresses: [
+      { city: params.city as string, state: params.state as StateAbbr },
+    ],
+    birth_date: params.dob,
+  };
+
+  if (typeof session?.user?.email === "string") {
+    try {
+      const subscriber = await getSubscriberByEmail(session.user.email);
+      // FIXME check if scan has already been performed and return early if so.
+
+      if (!subscriber.onerep_profile_id) {
+        const profileId = await createProfile(profileData);
+        await setOnerepProfileId(subscriber, profileId);
+        await setProfileDetails(profileId, profileData);
+      }
+
+      return NextResponse.json({ success: true }, { status: 200 });
+    } catch (e) {
+      console.error(e);
+      return NextResponse.json({ success: false }, { status: 500 });
+    }
+  } else {
+    // Not Signed in, redirect to home
+    return NextResponse.redirect(AppConstants.SERVER_URL, 302);
+  }
+}
+
+export async function GET(_req: NextRequest) {
+  const session = await getServerSession(authOptions);
+  if (typeof session?.user?.email === "string") {
+    try {
+      const subscriber = await getSubscriberByEmail(session.user.email);
+      const profileId = (await getOnerepProfileId(subscriber.id))[0][
+        "onerep_profile_id"
+      ] as number;
+
+      const scanResults = await getLatestOnerepScan(profileId);
+      return NextResponse.json(
+        { success: true, scan_results: scanResults },
+        { status: 200 }
+      );
+    } catch (e) {
+      console.error(e);
+      return NextResponse.json({ success: false }, { status: 500 });
+    }
+  } else {
+    // Not Signed in, redirect to home
+    return NextResponse.redirect(AppConstants.SERVER_URL, 302);
+  }
+}
--- a/src/app/functions/server/isUserSubscribed.ts
+++ b/src/app/functions/server/isUserSubscribed.ts
@@ -1,0 +1,52 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import { getServerSession } from "next-auth";
+import { authOptions } from "../../api/utils/auth";
+
+type FxaSubscriptionResponse = {
+  subscriptions: Array<{
+    product_id: string;
+    plan_id: string;
+    status: "active";
+  }>;
+};
+
+export const isUserSubscribed = async () => {
+  // Fetch list of subscriptions.
+  let subscriptions: FxaSubscriptionResponse;
+
+  const session = await getServerSession(authOptions);
+  if (!session?.user?.subscriber?.id) {
+    throw new Error("No session");
+  }
+
+  const bearerToken = session?.user.subscriber?.fxa_access_token;
+  if (bearerToken) {
+    const url = `${
+      process.env.OAUTH_API_URI ?? ""
+    }/oauth/mozilla-subscriptions/customer/billing-and-subscriptions`;
+    const result = await fetch(url, {
+      headers: {
+        "Content-Type": "application/json",
+        authorization: `Bearer ${bearerToken}`,
+      },
+    });
+    if (result.ok) {
+      subscriptions = await result.json();
+      for (const subscription of subscriptions.subscriptions) {
+        if (
+          subscription.product_id === process.env.PREMIUM_PRODUCT_ID &&
+          subscription.plan_id === process.env.PREMIUM_PLAN_ID_US &&
+          subscription.status === "active"
+        ) {
+          return true;
+        }
+      }
+      return false;
+    }
+  } else {
+    console.error("User has no bearer token");
+  }
+};
--- a/src/db/migrations/20230623152940_add_onerep_scans_table.js
+++ b/src/db/migrations/20230623152940_add_onerep_scans_table.js
@@ -1,0 +1,29 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+/**
+ * @param { import("knex").Knex } knex
+ * @returns { Promise<void> }
+ */
+export function up (knex) {
+  return knex.schema
+    .createTable('onerep_scans', table => {
+      table.increments('id').primary()
+      table.integer('onerep_profile_id').references('subscribers.onerep_profile_id').nullable()
+      table.integer('onerep_scan_id')
+      table.jsonb('onerep_scan_results')
+      table.timestamp('created_at').defaultTo(knex.fn.now())
+      table.timestamp('updated_at').defaultTo(knex.fn.now())
+      table.index('onerep_profile_id')
+      table.index('onerep_scan_id')
+    })
+}
+
+/**
+ * @param { import("knex").Knex } knex
+ * @returns { Promise<void> }
+ */
+export function down (knex) {
+  return knex.schema.dropTable('onerep_scans')
+}
--- a/src/db/tables/subscribers.js
+++ b/src/db/tables/subscribers.js
@@ -202,21 +202,6 @@ async function removeFxAData (subscriber) {
 
 /**
  * @param {import('../../app/(nextjs_migration)/(authenticated)/user/breaches/breaches.js').Subscriber} subscriber
- * @param {number} onerepProfileId
- */
-async function setOnerepProfileId (subscriber, onerepProfileId) {
-  await knex('subscribers')
-    .where('id', subscriber.id)
-    .update({
-      onerep_profile_id: onerepProfileId,
-      // @ts-ignore knex.fn.now() results in it being set to a date,
-      // even if it's not typed as a JS date object:
-      updated_at: knex.fn.now(),
-    })
-}
-
-/**
- * @param {import('../../app/(nextjs_migration)/(authenticated)/user/breaches/breaches.js').Subscriber} subscriber
  */
 async function setBreachesLastShownNow (subscriber) {
   // TODO: turn 2 db queries into a single query (also see #942)
@@ -430,6 +415,15 @@ async function updateMonthlyEmailOptout (token) {
     .where('primary_verification_token', token)
 }
 
+/**
+ * @param {number} subscriberId
+ */
+async function getOnerepProfileId (subscriberId) {
+  return await knex('subscribers')
+    .select('onerep_profile_id')
+    .where('id', subscriberId)
+}
+
 function getSubscribersWithUnresolvedBreachesQuery () {
   return knex('subscribers')
     .whereRaw('monthly_email_optout IS NOT TRUE')
@@ -471,6 +465,7 @@ async function joinEmailAddressesToSubscriber (subscriber) {
   return subscriber
 }
 export {
+  getOnerepProfileId,
   getSubscriberByToken,
   getSubscribersByHashes,
   getSubscriberByTokenAndHash,
@@ -484,7 +479,6 @@ export {
   updateFxAData,
   removeFxAData,
   updateFxAProfileData,
-  setOnerepProfileId,
   setBreachesLastShownNow,
   setAllEmailsToPrimary,
   setBreachesResolved,
--- a/public/nextjs_migration/client/js/welcome-details.js
+++ b/public/nextjs_migration/client/js/welcome-details.js
@@ -1,0 +1,24 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+const form = document.forms.details;
+form.addEventListener("submit", (e) => {
+  e.preventDefault();
+
+  const body = JSON.stringify(Object.fromEntries(new FormData(e.target)));
+  fetch("/api/v1/user/welcome", {
+    method: "POST",
+    headers: {
+      "Content-Type": "application/json",
+    },
+    body,
+  }).then(async (result) => {
+    const message = await result.json();
+    if (message && message.success === true) {
+      window.location = "scanning";
+    } else {
+      throw new Error("Could not submit profile:", message);
+    }
+  });
+});
--- a/src/app/functions/server/onerep.ts
+++ b/src/app/functions/server/onerep.ts
@@ -2,22 +2,23 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
+import { getServerSession } from "next-auth";
 import AppConstants from "../../../appConstants.js";
+import { getOnerepProfileId } from "../../../db/tables/subscribers.js";
 import mozlog from "../../../utils/log.js";
 import {
   E164PhoneNumberString,
   ISO8601DateString,
-  parseE164PhoneNumber,
-  parseIso8601Datetime,
 } from "../../../utils/parse.js";
-import { StateAbbr, usStates } from "../../../utils/states.js";
+import { StateAbbr } from "../../../utils/states.js";
+import { getLatestOnerepScan } from "../../../db/tables/onerep_scans";
+import { authOptions } from "../../api/utils/auth";
 const log = mozlog("external.onerep");
 
 export type ProfileData = {
   first_name: string;
   last_name: string;
-  city: string;
-  state: StateAbbr;
+  addresses: [{ city: string; state: StateAbbr }];
   birth_date?: ISO8601DateString;
   age?: ISO8601DateString;
   phone_number?: E164PhoneNumberString;
@@ -104,8 +105,8 @@ export async function createProfile(profileData: ProfileData): Promise<number> {
     birth_date: profileData.birth_date,
     addresses: [
       {
-        state: profileData.state,
-        city: profileData.city,
+        state: profileData.addresses[0].state,
+        city: profileData.addresses[0].city,
       },
     ],
   };
@@ -115,7 +116,9 @@ export async function createProfile(profileData: ProfileData): Promise<number> {
   });
   if (!response.ok) {
     log.info(
-      `Failed to create OneRep profile: [${response.status}] [${response.statusText}]`
+      `Failed to create OneRep profile: [${response.status}] [${
+        response.statusText
+      }] [${JSON.stringify(await response.json())}]`
     );
     throw new Error(
       `Failed to create OneRep profile: [${response.status}] [${response.statusText}]`
@@ -130,36 +133,6 @@ export async function createProfile(profileData: ProfileData): Promise<number> {
     url: string;
   } = await response.json();
   return savedProfile.id;
-}
-
-export function parseExposureScanData(body: ScanResult): ProfileData | null {
-  const state = usStates.find(
-    (state) => typeof body === "object" && state === body.addresses[0].state
-  );
-  if (
-    typeof body !== "object" ||
-    typeof body.first_name !== "string" ||
-    body.first_name.length === 0 ||
-    typeof body.last_name !== "string" ||
-    body.last_name.length === 0 ||
-    typeof body.addresses[0].city !== "string" ||
-    body.addresses[0].city.length === 0 ||
-    typeof body.addresses[0].state !== "string" ||
-    typeof state !== "string" ||
-    (typeof body.age !== "string" && typeof body.age !== "undefined") ||
-    (typeof body.phones !== "string" && typeof body.phones !== "undefined")
-  ) {
-    return null;
-  }
-
-  return {
-    first_name: body.first_name,
-    last_name: body.last_name,
-    city: body.addresses[0].city,
-    state,
-    age: parseIso8601Datetime(body.age)?.toString() ?? undefined,
-    phone_number: parseE164PhoneNumber(body.phones) ?? undefined,
-  };
 }
 
 export async function activateProfile(profileId: number): Promise<void> {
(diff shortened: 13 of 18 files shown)


Vinnl:
Hey Rob, small request: could you do new (UI) work inside `/app/(proper_react)/`. I'm hoping to be able to delete the `(nextjs_migration)` folder altogether after the new version goes live.

src/app/(nextjs_migration)/(authenticated)/user/welcome/details/page.tsx
```diff
@@ -0,0 +1,75 @@
...
+        <p>
+          We’ll use this to find exposures of your personal information, and
+          then guide you step-by-step on how to fix it.
+          <a>Why do we need this info?</a>
+        </p>
+
+        <section>
+          <form method="post" action="scanning">
+            <div>
+              <label for="name">First name: </label>
```
pdehaan, line 49: I think all these `for=` attributes are wrong.

```
Error: The value of the for attribute of the label element must be the ID of a non-hidden form control.

[From line 14, column 15; to line 14, column 32](https://validator.w3.org/nu/#l14c32)

          <label for="name">First
```

```suggestion
              <label for="firstname">First name: </label>
```

src/app/(nextjs_migration)/(authenticated)/user/welcome/details/page.tsx
```diff
@@ -0,0 +1,75 @@
...
+              <input type="text" name="firstname" id="firstname"></input>
```
pdehaan, line 50: W3c validator didn't like these `</input>` tags.

```
Error: Stray end tag input.

[From line 23, column 66; to line 23, column 73](https://validator.w3.org/nu/#l23c73)

itystate"></input>↩
```

Maybe this instead?

```suggestion
              <input type="text" name="firstname" id="firstname" />
```

src/app/(nextjs_migration)/(authenticated)/user/welcome/details/page.tsx
```diff
@@ -0,0 +1,75 @@
...
+            </div>
+            <div>
+              <label for="name">Last name: </label>
+              <input type="text" name="lastname" id="lastname"></input>
```
pdehaan, line 54:
```suggestion
              <label for="lastname">Last name: </label>
              <input type="text" name="lastname" id="lastname" />
```

src/app/(nextjs_migration)/(authenticated)/user/welcome/details/page.tsx
```diff
@@ -0,0 +1,75 @@
...
+            </div>
+            <div>
+              <label for="name">City and state: </label>
+              <input type="text" name="citystate" id="citystate"></input>
```
pdehaan, line 58:
```suggestion
              <label for="citystate">City and state: </label>
              <input type="text" name="citystate" id="citystate" />
```

src/app/(nextjs_migration)/(authenticated)/user/welcome/details/page.tsx
```diff
@@ -0,0 +1,75 @@
...
+            </div>
+            <div>
+              <label for="name">Date of birth: </label>
+              <input type="date" name="dob" id="dob"></input>
```
pdehaan, line 62:
```suggestion
              <label for="dob">Date of birth: </label>
              <input type="date" name="dob" id="dob" />
```

src/app/(nextjs_migration)/(authenticated)/user/welcome/details/page.tsx
```diff
@@ -0,0 +1,75 @@
...
+      images: ["/images/og-image.webp"],
+    },
+  };
+}
+
+export default async function UserBreaches() {
+  const l10n = getL10n();
+
+  return (
+    <>
```
pdehaan, line 37: nit: do we need the wrapper `<>` tag if there is only one child tag, `<main>`?

src/app/(nextjs_migration)/(authenticated)/user/welcome/page.tsx
```diff
@@ -0,0 +1,56 @@
...
+      images: ["/images/og-image.webp"],
+    },
+  };
+}
+
+export default async function UserBreaches() {
+  const l10n = getL10n();
+
+  return (
+    <>
```
pdehaan, line 36: Same as above, do we need this wrapper `<>` tag?

src/app/(nextjs_migration)/(authenticated)/user/welcome/page.tsx
```diff
@@ -0,0 +1,56 @@
...
+          own.
+        </p>
+
+        <p>
+          We believe in your right to privacy, so basic Monitor is always free.
+          Upgrade to Premium for continuous protection and automatic removal of
+          your personal info.
+        </p>
+
+        <p>See how we protect your data</p>
```
pdehaan, line 51: Should this have a period, or be a link? It feels a bit funny here.

src/app/(nextjs_migration)/(authenticated)/user/welcome/scanning/page.tsx
```diff
@@ -0,0 +1,107 @@
...
+    }
+    const result = await fetch(`https://api.onerep.com/${path}`, options);
+    if (!result.ok) {
+      throw new Error("Error connecting to provider");
+    }
+    return result.json();
+  }
+
+  const profile = await callOneRep("POST", "profiles");
+  const scan = await callOneRep("POST", `profiles/${profile.id}/scans`);
```
pdehaan, line 73: Q: is it possible that `profile` here can be `null`|`undefined` (causing the `profile.id` to error)?

src/app/(nextjs_migration)/(authenticated)/user/welcome/scanning/page.tsx
```diff
@@ -0,0 +1,107 @@
...
+
+  let iterations = 0;
+  const interval = setInterval(async () => {
+    const scanDetails = await callOneRep(
+      "GET",
+      `profiles/${profile.id}/scans/${scan.id}`
```
pdehaan, line 79: I guess same Q as above re: can `scan` results be `null`|`undefined` from OneRep causing this to error?

rhelmer:
Thanks @pdehaan - I'm going to move this to a proper React component, this was just an example so I could demo it earlier (which is why the PR is in draft mode). Per @Vinnl I'm going to start using the `(proper_react)` vs. the `(nextjs_migration)` (re-uses our old Express templates) for new code, but I'll keep your comments in mind as I port things over.

src/app/(proper_react)/redesign/(authenticated)/user/welcome/scanning/page.tsx
```diff
@@ -0,0 +1,123 @@
...
+    },
+  };
+}
+
+export default async function UserWelcomeScanning() {
+  const session = await getServerSession(authOptions);
+
+  const current = 0;
+  const total = 672;
+  const percentage = ((current / total) * 100).toFixed(1);
```
pdehaan, line 42: Q: Is it possible that `total` is `0` here and the universe (and/or multiverse) folds in on itself? Or is there a guard somewhere else for a divide by zero error?

Oh, interesting, I expected a RangeError, but looks like this might just return a `NaN` (in about:blank console and Node REPL):

```js
const percentage = ((0 / 0) * 100).toFixed(1);
console.log(percentage); // 'NaN'
```

Which would possibly give us `<h1>NaN%</h1>` below on L120.

src/app/(proper_react)/redesign/(authenticated)/user/welcome/scanning/page.tsx
```diff
@@ -0,0 +1,123 @@
...
+      options
+    );
+    if (!result.ok) {
+      throw new Error("Error connecting to provider");
+    }
+    return result.json();
+  }
+
+  // @ts-ignore FIXME
+  const result = await getOnerepProfileId(session?.user?.subscriber?.id);
```
pdehaan, line 68: And while my special 👁️'s are here, re: abundance of `?.` operators does this mean we could be calling `getOnerepProfileId(undefined)` and we should avoid that if the user is not a subscriber (or logged in, or whatever)?

Also, a bit of a nit but this should probably be "OneRep", per https://onerep.com/ (ie: "© 2023 OneRep LLC").
But I think that code is probably outside the scope of this PR, but I'm a stickler for style guides.

We call it `callOneRep()` above (L44), but also `*Onerep*` in these imports (from the /db/tables/subscribers table):

https://github.com/mozilla/blurts-server/blob/911c5ec128440278cd90e66f9d37e719374080f8/src/app/(proper_react)/redesign/(authenticated)/user/welcome/scanning/page.tsx#L7-L12

```sh
git grep -n "Onerep" src # 22 results
```

Although I think almost all instances of "Onerep" in the codebase referred to the functions/exports from [src/db/tables/subscribers.js exports](https://github.com/mozilla/blurts-server/blob/911c5ec128440278cd90e66f9d37e719374080f8/src/db/tables/subscribers.js#L475-L493):

```js
476:  getLatestOnerepScan,
477:  getOnerepProfileId,
...
491:  setOnerepProfileId,
492:  setOnerepScan,
493:  setOnerepScanResults,
```

src/app/api/v1/user/welcome/route.ts
```diff
@@ -0,0 +1,79 @@
...
+  if (method !== "GET" && method !== "HEAD") {
+    //@ts-ignore FIXME
+    options.body = body;
+  }
+  const result = await fetch(`${process.env.ONEREP_API_BASE}/${path}`, options);
+  if (!result.ok) {
+    throw new Error("Error connecting to provider");
+  }
+  return result.json();
+}
```
pdehaan, line 32: This looks very similar to (but not exactly the same as) `callOneRep()` method https://github.com/mozilla/blurts-server/blob/911c5ec128440278cd90e66f9d37e719374080f8/src/app/(proper_react)/redesign/(authenticated)/user/welcome/scanning/page.tsx#L44-L65
I think the only difference is one function handles an optional "POST" `body` payload and the other doesn't.
Q: Should we make this some shared utility lib function somewhere and share between the two files?

Also, should `method` be a generic `:string` type, or should we hardcode constant `"GET"|"POST"`? Those were the only two methods I found in use (via <kbd>git grep -n -A1 "callOneRep(" src</kbd>) although the code above does a check for `!== "HEAD"`).

src/db/tables/subscribers.js
```diff
@@ -196,6 +196,34 @@ async function setOnerepProfileId (subscriber, onerepProfileId) {
...
+ */
+async function setOnerepScanResults (onerepProfileId, oneRepScanId, oneRepScanResults) {
+  await knex('onerep_scans')
+    .where('onerep_profile_id', onerepProfileId)
+    .andWhere('onerep_scan_id', oneRepScanId)
+    .update({
+      onerep_scan_results: oneRepScanResults,
+      updated_at: knex.fn.now()
+    })
+}
```
pdehaan, line 225:
```suggestion
/**
 * @param {number} oneRepProfileId
 * @param {number} oneRepScanId
 * @param {object} oneRepScanResults
 */
async function setOneRepScanResults (oneRepProfileId, oneRepScanId, oneRepScanResults) {
  await knex('onerep_scans')
    .where('onerep_profile_id', oneRepProfileId)
    .andWhere('onerep_scan_id', oneRepScanId)
    .update({
      onerep_scan_results: oneRepScanResults,
      updated_at: knex.fn.now()
    })
}
```

PROTIP: I also think you can pass an object into the Knex.js `where` clause instead of chained `.where().andWhere()`; https://knexjs.org/guide/query-builder.html#where:

```js
knex('users')
  .where({
    first_name: 'Test',
    last_name:  'User'
  })
  .select('id')
```

src/customTypes.d.ts
```diff
@@ -80,3 +80,29 @@ interface Window {
...
+  }[];
+  last_name: string;
+  relatives: string[];
+  created_at: string | Date;
+  first_name: string;
+  profile_id: number;
+  updated_at: string | Date;
+  data_broker: string;
+  middle_name: string | null;
+  data_broker_id: number;
```
pdehaan, line 107: nit: not sure this is any better/clearer, but roughly grouped by `*_name`, plus grouped `data_broker*` and date fields together.
```suggestion
  first_name: string;
  middle_name: string | null;
  last_name: string;
  relatives: string[];
  profile_id: number;
  data_broker: string;
  data_broker_id: number;
  created_at: string | Date;
  updated_at: string | Date;
```

src/app/functions/server/getOneRepResult.ts
```diff
@@ -0,0 +1,32 @@
...
+  const bearerToken = process.env.ONEREP_API_KEY;
+  const options = {
+    method,
+    headers: {
+      Authorization: `Bearer ${bearerToken}`,
+      "Content-Type": "application/json",
+    },
+  };
+
+  if (body) {
```
pdehaan, line 19: One subtle thing we changed here was the `method` check above from src/app/api/v1/user/welcome/route.ts#L23

```diff
-if (method !== "GET" && method !== "HEAD") {
+if (body) {
```

is `POST` the only valid method that supports `body` payloads, and do we want to add some check back in to make sure somebody isn't accidentally trying to pass a body along w/ a `GET` request?

src/app/functions/server/getOneRepResult.ts
```diff
@@ -0,0 +1,32 @@
...
+    // @ts-ignore FIXME
+    options.body = body;
+  }
+
+  const result = await fetch(`${process.env.ONEREP_API_BASE}/${path}`, options);
+  if (!result.ok) {
+    throw new Error(
+      `Error connecting to provider: ${process.env.ONEREP_API_BASE},
+       ${JSON.stringify(await result.json())}`
```
pdehaan, line 28: Hhmm. is it possible that the `result` may not be JSON object on a `!ok` response and `await result.json()` can throw an error here?
Not entirely sure what that means or how it could happen (since I've never used OneRep API before), but maybe on a 404 path or invalid bearer token/config, or 5xx API server error?

Possibly not a concern since we don't seem to be in a `try..catch` block and presumably the mythical `await result.json()` error might just rethrow upwards to the callee anyways. <i>🤷</i>

src/app/(nextjs_migration)/(authenticated)/user/welcome/details/page.tsx
```diff
@@ -0,0 +1,75 @@
...
+      <main>
+        <h1>Enter the details you want to protect</h1>
+        <p>
+          We’ll use this to find exposures of your personal information, and
+          then guide you step-by-step on how to fix it.
+          <a>Why do we need this info?</a>
```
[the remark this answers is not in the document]
rhelmer: This isn't the final copy (also why I'm hardcoding English :)), I'm working on the backend parts while @codemist is building out the components. I just want it to be in a demo-able state.

It does looks weird to have this even in a demo state though, removed it.

src/app/(nextjs_migration)/(authenticated)/user/welcome/details/page.tsx
rhelmer: Yeah, sorry I wasn't expecting anyone to jump on it when it was in draft state. No worries though, it's super helpful to have eyes on things as early as possible, but I don't want you to waste your time on stuff that's not quite ready either.

src/app/(nextjs_migration)/(authenticated)/user/welcome/details/page.tsx
rhelmer: Huh, I'm surprised to hear that actually... will take a look.

src/app/(nextjs_migration)/(authenticated)/user/welcome/page.tsx
rhelmer: Yeah I don't have anything to link it to, will remove for now thanks.

src/app/(nextjs_migration)/(authenticated)/user/welcome/scanning/page.tsx
rhelmer: In general I'd prefer to not catch errors (preferring to let a global error handler catch them) unless we're sure we can actually handle them. I don't think there's anything useful we can do if this fails, the user shouldn't be at this page unless they successfully created a profile and a scan and they were both saved in the database. If this sort of thing fails, it likely means the database is unavailable or the network is down, so we can't really provide the user much of an experience at this point.

src/app/(nextjs_migration)/(authenticated)/user/welcome/scanning/page.tsx
rhelmer: Actually I do think defining this sort of thing in the type definition would be the most useful thing to do for now, I'll make sure that we're doing the right thing there. As above though, if the scan results aren't present and the user got this far, then something is wrong.

The intended flow is like this:

1. welcome screen
2. user accepts, redirects to details page: user enters information
3. form post to API endpoint: information is validated, onerep profile created and stored in db
4. redirect to scanning page: progress is shown, and scan is initiated and recorded in db
5. redirect to results page: scan result is displayed

At each step, the results are stored in our database so if it's inaccessible that's likely down. We have plans to improve this by introducing queuing (which could fail too of course but e.g. CloudTasks are more distributed and more reliable in general than our primary db, so it will be a step in the right direction).

src/app/(proper_react)/redesign/(authenticated)/user/welcome/scanning/page.tsx
rhelmer: This is pretty different now, shouldn't have the same problem.

src/app/(proper_react)/redesign/(authenticated)/user/welcome/scanning/page.tsx
rhelmer: This is bugging me too, might as well make it consistent now. Thanks!

src/app/api/v1/user/welcome/route.ts
rhelmer: I forgot that @Vinnl already landed a very similar set of functions so I switched to that, and his is much nicer :)

src/app/(nextjs_migration)/(authenticated)/user/welcome/details/page.tsx
cadeyrn: A trailing slash is not needed either. Just write:

`<input type="text" name="firstname" id="firstname">`

It's not an error to use a trailing slash but the W3C validator gives you the following information if you use it:

> Trailing slash on void elements has no effect and interacts badly with unquoted attribute values.

src/app/(proper_react)/redesign/(authenticated)/user/welcome/remove/page.tsx
```diff
@@ -0,0 +1,102 @@
...
+      }
+    );
+    if (result.ok) {
+      subscriptions = await result.json();
+    }
+  } else {
+    console.error("User has no bearer token");
+  }
+
+  type FxaSubscriptionResponse = {
```
pdehaan, line 59: tiniest of nits: In other places we call it "FxA"

https://github.com/mozilla/blurts-server/blob/d67b9acd84c6923114327c76408687fdf8ab1730/src/db/tables/subscribers.js#L135

Although looks like we use `FxA` about 5:3 over `Fxa`, so&hellip;

<details>
<summary><kbd>git grep -Pn "(const|let|var|function) [a-zA-Z]+FxA[A-Z][a-zA-Z]+" | wc -l # 5</kbd></summary>

```sh
src/app/api/v1/fxa-rp-events/route.ts:186:        const currentFxAProfile = subscriber?.fxa_profile_json || {};
src/controllers/fxaRpEvents.js:151:        const currentFxAProfile = subscriber?.fxa_profile_json || {}
src/db/tables/subscribers.js:135:async function updateFxAData (subscriber, fxaAccessToken, fxaRefreshToken, fxaProfileData) {
src/db/tables/subscribers.js:163:async function updateFxAProfileData (subscriber, fxaProfileData) {
src/db/tables/subscribers.js:180:async function removeFxAData (subscriber) {
```
</details>


<details>
<summary><kbd>git grep -Pn "(const|let|var|function) [a-zA-Z]+Fxa[A-Z][a-zA-Z]+" | wc -l # 3</kbd></summary>

```sh
src/app/api/v1/fxa-rp-events/route.ts:60:const authenticateFxaJWT = async (req: NextRequest) => {
src/controllers/fxaRpEvents.js:47:const authenticateFxaJWT = async (req) => {
src/db/tables/subscribers.js:59:async function getSubscriberByFxaUid (uid) {
```
</details>

src/app/(proper_react)/redesign/(authenticated)/user/welcome/remove/page.tsx
```diff
@@ -0,0 +1,102 @@
...
+      }
+    }
+    return false;
+  };
+
+  if (
+    process.env.PREMIUM_ENABLED !== "true" ||
+    !isUserSubscribed(subscriptions)
+  ) {
+    return NextResponse.redirect(process.env.SERVER_URL ?? "results", 301);
```
pdehaan, line 87: Is there a case where `process.env.SERVER_URL` would be falsy?
We seem to use the config a lot so was curious what'd break if this was to be undefined.

src/app/(proper_react)/redesign/(authenticated)/user/welcome/results/page.tsx
```diff
@@ -0,0 +1,111 @@
...
+  return (
+    <main>
+      {process.env.PREMIUM_ENABLED === "true" &&
+      !isUserSubscribed(subscriptions) ? (
+        <div>
+          <a
+            className="button primary"
+            href={`${process.env.FXA_SUBSCRIPTIONS_URL}/products/${process.env.PREMIUM_PRODUCT_ID}?plan=${process.env.PREMIUM_PLAN_ID_US}`}
+          >
+            <button className="button primary">Subscribe to Premium</button>
```
pdehaan, line 97: Same comment here... Do we need the `<button>` wrapper? We're already setting `className="button primary"` above on the `<a>` tag on L94.

src/app/(proper_react)/redesign/(authenticated)/user/welcome/results/page.tsx
```diff
@@ -0,0 +1,111 @@
...
+          </a>
+        </div>
+      ) : (
+        <div>
+          <a className="button primary" href="remove">
+            <button className="button primary">Remove my Data</button>
```
pdehaan, line 103: Same Q here re: `<button>`.

src/app/(proper_react)/redesign/(authenticated)/user/welcome/scanning/page.tsx
```diff
@@ -0,0 +1,108 @@
...
+    } else if (scanDetails.status === "finished") {
+      clearInterval(interval);
+
+      const scanListFull = [];
+      const firstPage = await listScanResults(profileId, { per_page: 100 });
+      // Results are paginated, use per_page maximum and collect all pages into one result.
+      if (firstPage.meta.last_page > 1) {
+        let currentPage = 2;
+        while (currentPage <= firstPage.meta.last_page) {
+          const nextPage = await listScanResults(profileId, { per_page: 100 });
```
pdehaan, line 80: How does `listScanResults()` know/track what the current page we're requesting is?
I see a reference to `currentPage` which we set/increment, but not sure how we're passing that to `listScanResults()`.

I think this code is implying we should be passing an optional `page` param  (which will then get passed to the request's query params.
TLDR: Should we be passing `{ per_page: 100, page: currentPage }`?
https://github.com/mozilla/blurts-server/blob/d67b9acd84c6923114327c76408687fdf8ab1730/src/external/onerep.js#L277-L286

src/db/migrations/20230623152940_add_onerep_scans_table.js
```diff
@@ -0,0 +1,28 @@
...
+/**
+ * @param { import("knex").Knex } knex
+ * @returns { Promise<void> }
+ */
+export function up (knex) {
+  return knex.schema
+    .createTable('onerep_scans', table => {
+      table.increments('id').primary()
+      table.integer('onerep_profile_id').references('subscribers.onerep_profile_id').nullable()
+      table.integer('onerep_scan_id')
```
pdehaan, line 14: Does `onerep_scan_id` need an index too?
We seem to use `where` on both `onerep_scan_id` and `onerep_profile_id` but I only see an explicit index on `onerep_profile_id` below on L18.

Sample query:
https://github.com/mozilla/blurts-server/blob/d67b9acd84c6923114327c76408687fdf8ab1730/src/db/tables/onerep_scans.js#L14-L18

src/db/tables/onerep_scans.js
```diff
@@ -0,0 +1,82 @@
...
+    })
+}
+
+export {
+  getLatestOnerepScan,
+  getOnerepScanResults,
+  setOnerepProfileId,
+  setOnerepScan,
+  setOnerepScanResults
+}
```
pdehaan, line 82: nit: add newline.

src/db/tables/profiles.js
```diff
@@ -0,0 +1,46 @@
...
+      state_code: profileData.addresses[0].state_name,
+      date_of_birth: profileData.birth_date,
+      created_at: knex.fn.now()
+    })
+}
+
+export {
+  getProfileDetails,
+  setProfileDetails
+}
```
pdehaan, line 46: nit: add newline

src/external/onerep.js
```diff
@@ -309,3 +313,48 @@ export async function listScanResults(profileId, options = {}) {
...
+  return true;
+}
+
+/**
+ *
+ * @param {number} profileId
+ * @param {number} scanId
+ * @returns {Scan}
+ */
+export async function showScanDetails(profileId, scanId) {
```
pdehaan, line 347: nit: should this be _get_ScanDetails() or _fetch_ScanDetails()?
"show" sounds possibly confusing since it's merely fetching JSON and not displaying anything.

src/app/(proper_react)/redesign/(authenticated)/user/welcome/scanning/page.tsx
rhelmer: FYI I put some guards in to ensure that users have a session and a bearer token and exit early if not, but this is likely unnecessary since this is all under an authenticated route. I don't really understand Next deeply enough to know what's safe, so I am leaning on the "duplicate checks" side.

src/app/(proper_react)/redesign/(authenticated)/user/welcome/details/page.tsx
```diff
@@ -0,0 +1,66 @@
...
+    <main>
+      <h1>Enter the details you want to protect</h1>
+      <p>
+        We’ll use this to find exposures of your personal information, and then
+        guide you step-by-step on how to fix it.
+        <a>Why do we need this info?</a>
+      </p>
+
+      <section>
+        <form method="post" action="http://localhost:6060/api/v1/user/welcome">
```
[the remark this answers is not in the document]
rhelmer: Thanks! I think `welcome` isn't a great name for this API endpoint, I may change this, but hardcoding server name is bad. I *think* I remember there being a reason we needed an absolute not relative URL here, but if that's the case this should be using the `SERVER_URL` variable anyway, so thanks for catching that.

src/db/tables/subscribers.js
```diff
@@ -196,6 +196,34 @@ async function setOnerepProfileId (subscriber, onerepProfileId) {
...
+ * @param {number} oneRepScanId
+ */
+async function setOnerepScan (onerepProfileId, oneRepScanId) {
+  await knex('onerep_scans')
+    .insert({
+      onerep_profile_id: onerepProfileId,
+      onerep_scan_id: oneRepScanId,
+      created_at: knex.fn.now()
+    })
+}
```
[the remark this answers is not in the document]
rhelmer: We're already using `Onerep`/`onerep_` in various places, I think it might be the winning convention, but I'll check. @Vinnl had an opinion on this IIRC

src/db/tables/subscribers.js
rhelmer: Yeah, I know I'm doing something wrong when the caller needs to do stuff like `result[0].blah` :)

public/nextjs_migration/client/js/welcome.js
```diff
@@ -0,0 +1,35 @@
...
+    if (!results) {
+      const response = await fetch("/api/v1/user/welcome");
+      const body = await response.json();
+
+      if (response.ok && body.scan_results[0].onerep_scan_results !== null) {
+        results = body.scan_results[0].onerep_scan_results;
+      }
+    }
+
+    const percentComplete = intervalCount * 6;
```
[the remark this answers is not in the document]
rhelmer: Yes it's completely arbitrary. Will be replaced shortly after landing by a React component (that will do essentially the same thing, just make sure the user knows things haven't crashed and aren't hanging, but we have no visibility into incremental progress AFAIK). This is something we're going to confirm soon though, I can't find anything in the documented API that gives us this kind of incremental progress for a manual scan.

public/nextjs_migration/client/js/welcome.js
```diff
@@ -0,0 +1,35 @@
...
+
+    if (percentComplete >= 100) {
+      progress.textContent = "Scan complete";
```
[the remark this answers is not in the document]
rhelmer: Yes, the reason this and other places are not is because I want to use Fluent *but* this is an English-only feature, at least for now, so I want to do it in a way that won't ping localizers and waste their time. Leaving this open though since I agree we need to be using Fluent even for english-only strings.

public/nextjs_migration/client/js/welcome.js
```diff
@@ -0,0 +1,35 @@
...
+      clearInterval(interval);
+
+      window.location = "results";
+    } else {
+      progress.textContent = `${percentComplete}%`;
```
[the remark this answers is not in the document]
rhelmer: This isn't the final UI, but TIL :) I don't mind doing it properly even for this first phase.

src/app/(proper_react)/redesign/(authenticated)/user/welcome/details/page.tsx
```diff
@@ -0,0 +1,106 @@
...
+      title: l10n.getString("brand-fx-monitor"),
+      description: l10n.getString("meta-desc-2"),
+      images: ["/images/og-image.webp"],
+    },
+    openGraph: {
+      title: l10n.getString("brand-fx-monitor"),
+      description: l10n.getString("meta-desc-2"),
+      siteName: l10n.getString("brand-fx-monitor"),
+      type: "website",
+      url: process.env.SERVER_URL,
```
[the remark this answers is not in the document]
rhelmer: I'm not even sure we should bother with it here since it's an authenticated route...

src/app/(proper_react)/redesign/(authenticated)/user/welcome/scanning/page.tsx
rhelmer: I'll add a comment, I realize now this is very non-obvious.

src/db/migrations/20230623152940_add_onerep_scans_table.js
rhelmer: Yes I think it's appropriate here... I don't expect this table to grow at the same rate as our `subscribers` table, but it's still good practice (and I imagine that joining based on the onerep ID will be the normal use case).

We'll also use this table to keep an eye on how many scans have been performed, and how long they take, I need to think about this and test it a bit. It would be pretty easy to partition this table if it becomes performance critical (partitioning by month would be a natural fit with the constraints we have on number of scans).

src/external/onerep.js
rhelmer: It's actually called both in the docs ("show" and "get"): https://docs.onerep.com/#operation/getScan

If `get` seems clearer I am fine with that.

src/db/migrations/20230623152940_add_onerep_scans_table.js
Vinnl, line 1: Could you also add a `OnerepScanRow` to https://github.com/mozilla/blurts-server/blob/b601bb75449e51a545dc393b25a5c5c38c6b7ccc/src/knex-tables.d.ts and then list it in the `Tables` interface at the bottom of that file? Ideally in the same commit in which you add this migration, so that we can `git blame` the type to see which migration introduced each column.

src/db/migrations/20230703010356_add-profile-table.js
Vinnl, line 1: And likewise:

> Could you also add a `OnerepProfileRow` to https://github.com/mozilla/blurts-server/blob/b601bb75449e51a545dc393b25a5c5c38c6b7ccc/src/knex-tables.d.ts and then list it in the `Tables` interface at the bottom of that file? Ideally in the same commit in which you add this migration, so that we can `git blame` the type to see which migration introduced each column.

src/db/migrations/20230703010356_add-profile-table.js
```diff
@@ -0,0 +1,27 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+export function up(knex) {
+  return knex.schema
+    .createTable('profiles', table => {
```
Vinnl, line 7: Let me make a prediction that someone's going to get confused in the future if we don't rename this :sweat_smile: 

```suggestion
    .createTable('onerep_profiles', table => {
```

src/db/migrations/20230703010356_add-profile-table.js
```diff
@@ -0,0 +1,27 @@
...
+      table.increments('id').primary()
+      table.integer('onerep_profile_id').references('subscribers.onerep_profile_id').nullable()
```
Vinnl, line 9: Hmm... Do we want to store OneRep profiles if we don't know which user they belong to? (Referring to the `.nullable()` here.)

src/db/tables/onerep_scans.js
Vinnl, line 1: Any reason to not use TypeScript here?

src/app/(proper_react)/redesign/(authenticated)/user/welcome/scanning/page.tsx
```diff
@@ -0,0 +1,108 @@
...
+          currentPage++;
+          nextPage.data.forEach((element: object) =>
```
Vinnl, line 82: Wouldn't we be able to use this specific type from `onerep.ts` here? (In fact, I think that might actually even get inferred if you remove `:object` altogether?)

```suggestion
          nextPage.data.forEach((element: ScanResult) =>
```

src/db/tables/onerep_scans.js
```diff
@@ -0,0 +1,82 @@
...
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+ import initKnex from 'knex'
+ import knexConfig from '../knexfile.js'
+ const knex = initKnex(knexConfig)
+ 
+/**
+ * @param {number} onerepProfileId
+ * @param {number} oneRepScanId
+ * @returns {Promise<object>} oneRepScanResults
```
Vinnl, line 12: I think this and other references to the unspecific `object` type in this file can also instead refer to `ScanResult` from `onerep.ts` directly?

src/db/tables/profiles.js
```diff
@@ -0,0 +1,48 @@
...
+
+import initKnex from 'knex'
+import knexConfig from '../knexfile.js'
+
+import { ProfileData } from "../../external/onerep.js";
+
+const knex = initKnex(knexConfig)
+
+/**
+ * @param {number} onerepProfileId
```
Vinnl, line 13: Doesn't this result in an object with a whole bunch of properties, rather than just the profile ID? (In fact, I think TypeScript should be able to infer the correct types if you add a type definition to `knex-tables.d.ts`.)

src/db/tables/profiles.js
Vinnl, line 1: Here too: any reason not to use TypeScript? (And also maybe rename this to `onerep_profiles`.)

src/external/onerep.js
```diff
@@ -137,6 +141,11 @@ export async function activateProfile(profileId) {
     method: "PUT",
   });
   if (!response.ok) {
+    const result = await response.json();
+    if (result.message === "Profile is already active") {
+      // TODO is there an advantage to checking first or is this that a wasted API call?
```
Vinnl, line 146: Note that this will throw an unrelated error if the response body does not include a `.message` property. And given that I think we generally would not expect `activateProfile` to be called multiple times for the same profile (i.e. it's a genuine error condition), I think we _do_ want to log it?

public/nextjs_migration/client/js/welcome.js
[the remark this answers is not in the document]
Vinnl: Just a note that we do have a `locales-pending` directory now that won't get submitted to localisers yet. That said, it would be good to only have strings in there that might eventually go to localisers (even if just to `es-MX`), which I don't think is the case for these proof-of-concept pages.

src/app/(nextjs_migration)/(authenticated)/user/breaches/page.tsx
```diff
@@ -148,8 +113,7 @@ export default async function UserBreaches() {
       <Script type="module" src="/nextjs_migration/client/js/dialog.js" />
 
       <main data-partial="breaches">
-        {process.env.PREMIUM_ENABLED === "true" &&
-        !isUserSubscribed(subscriptions) ? (
+        {process.env.PREMIUM_ENABLED === "true" && !isUserSubscribed() ? (
```
Vinnl, line 116:
```suggestion
        {process.env.PREMIUM_ENABLED === "true" && !(await isUserSubscribed()) ? (
```

This would've been caught by [this linting rule](https://typescript-eslint.io/rules/no-floating-promises/), so I'll submit a PR to enable it.

Vinnl (commented):
Sorry to have taken so long, but I've finally reviewed it! Left a bunch of comments, hope they're helpful.

src/db/tables/subscribers.js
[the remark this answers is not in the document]
Vinnl: Heh yeah, a soft opinion: I'd use a single capital letter for all external names, just for predictability in edge cases, e.g. `isRSSXML` vs `isRssXml`, `isJsonld` vs `isJsonLd` vs `isJSONLD`, or `getFxaProfile` vs `getFxAProfile` vs `getFXAProfile`. I won't die on that hill though :)

src/app/api/v1/user/welcome/route.ts
```diff
@@ -0,0 +1,102 @@
...
+    throw new Error("First name is required");
+  } else if (!params.has("lastname")) {
+    throw new Error("Last name is required");
+  } else if (!params.has("city")) {
+    throw new Error("City is required");
+  } else if (!params.has("state")) {
+    throw new Error("State is required");
+  } else if (!params.has("dob")) {
+    throw new Error("Date of Birth is required");
+  }
```
flozia, line 43: Suggestion: Should we check not only the presence of the param but also the actual value so we are catching empty strings?

```suggestion
  const requiredParams = ['firstname', 'lastname', 'city', 'state', 'dob'];
  requiredParams.forEach(param => {
    if (!params.get(param)) {
      throw new Error(`${param} is required`);
    }
  });
```

src/app/functions/server/isUserSubscribed.ts
```diff
@@ -0,0 +1,52 @@
...
+
+  const session = await getServerSession(authOptions);
+  if (!session?.user?.subscriber?.id) {
+    throw new Error("No session");
+  }
+
+  const bearerToken = session?.user.subscriber?.fxa_access_token;
+  if (bearerToken) {
+    const result = await fetch(
+      `${process.env.OAUTH_API_URI}/oauth/mozilla-subscriptions/customer/billing-and-subscriptions`,
```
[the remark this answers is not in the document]
rhelmer: I did talk to subplat folks last week - it turns out there are two slightly different scopes (`profile:subscription` vs. the subscriptions URI that allows calling this API). The former is what should cause the product to show up in the user's profile (JWT).

The only reason we'd need to call these APIs is if we need to be able to differentiate plans (to unlock different features for different plans, which we don't currently have plans for), or if we need more information about subscribers (again no plans for this).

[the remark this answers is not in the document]
rhelmer: > I'm still a bit nervous that we haven't found a good way to do this. Do you know if we can setup SubPlat to pass a query param to Monitor when it redirects the user back to us? That's how we refresh the user's session on Relay, and I think we might possibly be able to solve it by calling [`session.update`](https://next-auth.js.org/getting-started/client#updating-the-session) at that point, combined with a server-side call to update the profile data?

Also yes we get a query param and I will try calling `session.update` and see if that works to refresh the session. Thanks!

src/db/tables/profiles.js
rhelmer: No, I just started this locally before we started talking about TS for knex, I can do this.

src/app/(nextjs_migration)/(authenticated)/user/welcome/details/page.tsx
rhelmer: Hm, our JSX linter wants a closing tag here, I'll do that for now. Note that the final HTML output might look slightly different! JSX is XML-ish.

public/nextjs_migration/client/js/welcome.js
```diff
@@ -0,0 +1,35 @@
...
+
+  const progress = document.querySelector("#progress");
+  let results;
+
+  const interval = setInterval(async () => {
+    if (!results) {
+      const response = await fetch("/api/v1/user/welcome");
+      const body = await response.json();
+
+      if (response.ok && body.scan_results[0].onerep_scan_results !== null) {
```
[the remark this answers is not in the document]
rhelmer: I mean, it shouldn't happen, but browser extensions or network problems etc. can happen...

src/app/functions/server/isUserSubscribed.ts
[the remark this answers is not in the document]
rhelmer: So I think we should do this separately, because the real underlying problem here is that we have too many test client IDs floating around, and they do not have the proper scopes. This actually works on localhost/dev/stage right now, so I'd like to land as-is and then clean up the client IDs (which might take a few days for SRE), then land a targeted fix that just changes this to use the `subscriptions` claim instead of this API.

src/app/api/v1/user/welcome/route.ts
```diff
@@ -0,0 +1,102 @@
...
+
+export async function POST(req: NextRequest) {
+  const token = await getToken({ req });
+
+  const eligible = await isEligible();
+  if (!eligible) {
+    throw new Error("User has no more manual scans left");
+  }
+
+  const params = new URLSearchParams(await req.text());
```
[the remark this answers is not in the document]
rhelmer: Yeah, HTML forms are just so simple :P but sure I'll make it JSON.

src/db/tables/profiles.js
rhelmer: Hm well this is `getProfileDetails`, that just takes the ID... the function below takes a `ProfileData` object so that seems fine?

src/app/(nextjs_migration)/(authenticated)/user/breaches/page.tsx
rhelmer: Yeah thanks I rebased and the linter caught it, nice work :) Already fixed

src/app/(proper_react)/redesign/(authenticated)/user/welcome/details/page.tsx
```diff
@@ -0,0 +1,106 @@
...
+                  <td>
+                    <label htmlFor="dob">Date of birth: </label>
+                  </td>
+                  <td>
+                    <input type="date" name="dob" id="dob"></input>
+                  </td>
+                </tr>
+                <tr>
+                  <td>
+                    <button className="button secondary">Go back</button>
```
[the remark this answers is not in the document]
rhelmer: Oh hm yeah that's interesting... I guess it should be outside of the form (there's probably a way to prevent it from JS too)

src/external/onerep.js
rhelmer: Hm yes I wasn't sure whether to throw an error or continue... I also considered making a call to check if the user was already activated, but this seems simpler and doesn't use an extra API call.

I think the only time this would matter is if the profile was activated but calling out-out failed for some reason, it's not harmful to try again. However it shouldn't be necessary.

src/db/migrations/20230703010356_add-profile-table.js
```diff
@@ -0,0 +1,27 @@
...
+      table.index('onerep_profile_id')
+    })
+}
+
+/**
+ * @param { import("knex").Knex } knex
+ * @returns { Promise<void> }
+ */
+export function down(knex) {
+  return knex.schema.dropTable('profiles')
```
pdehaan, line 26: See  https://github.com/mozilla/blurts-server/pull/3157/files#r1255721542 suggestion above about potentially renaming this table from "profiles" to "onerep_profiles".

src/db/tables/onerep_scans.js
pdehaan: :+1: to this. I was going to file a ticket or ask in Jira, but do we have an exiting issue about converting all existing .js files .ts [where possible] for consistency? It'd be nice to have a repo that is mostly .ts and .tsx so we can easily share type definitions (versus JSDoc style).

src/db/migrations/20230703010356_add-profile-table.js
```diff
@@ -0,0 +1,27 @@
...
+
+export function up(knex) {
+  return knex.schema
+    .createTable('profiles', table => {
+      table.increments('id').primary()
+      table.integer('onerep_profile_id').references('subscribers.onerep_profile_id').nullable()
+      table.integer('first_name')
+      table.integer('last_name')
+      table.integer('city_name')
+      table.integer('state_code')
```
pdehaan, line 13: Should these be `.integer()` or `.string()`?

```sh
git grep -Pho "table\.\w+\(" | sort | uniq

table.boolean(
table.date(
table.dropColumn(
table.dropColumns(
table.dropIndex(
table.dropTimestamps(
table.increments(
table.index(
table.integer(
table.jsonb(
table.renameColumn(
table.specificType(
table.string(
table.text(
table.timestamp(
table.timestamps(
```

src/app/functions/server/isUserSubscribed.ts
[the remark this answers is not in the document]
Vinnl: Oh yeah, I think fixing that is outside of the scope of this PR, but just wanted to see if we could come up with some more plans on how to fix this. Note that I don't expect that `session.update` will do anything unless we also do another [`userinfo` call](https://github.com/mozilla/blurts-server/blob/7107f728e0d2d90af78c45b98e1edf206af86581/src/app/api/utils/auth.ts#L65-L75) and then update the profile.

src/app/(proper_react)/redesign/(authenticated)/user/welcome/details/page.tsx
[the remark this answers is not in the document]
Vinnl: HTML `<button>` elements by default have a `type="submit"`. If you set it to `type="button"` it shouldn't submit anymore. (But doesn't matter a whole lot, since it's just a demo page.)

src/app/(nextjs_migration)/(authenticated)/user/welcome/details/page.tsx
Vinnl: Yeah, it's not actually HTML, it's roughly syntactic sugar for programmatically creating DOM elements. Running the W3C validator on the input source won't help much - I imagine it doesn't like attributes like `className` or `htmlFor` either :sweat_smile:

src/db/tables/profiles.js
Vinnl: Oh whoops, I read `@returns` instead of `@param`. Never mind then! :sweat_smile:

src/external/onerep.js
```diff
@@ -309,3 +318,48 @@ export async function listScanResults(profileId, options = {}) {
   }
   return response.json();
 }
+
+export async function isEligible() {
```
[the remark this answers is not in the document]
rhelmer: The first step must be to create a OneRep profile, which is where we get this ID from.

I think we should use the `onrep_scans` table, it's indexed by `onerep_profile_id` so it's very easy to just look up how many scans a user has performed and what the status of them is.

src/app/(proper_react)/redesign/(authenticated)/user/welcome/remove/page.tsx
```diff
@@ -0,0 +1,59 @@
...
+  if (process.env.PREMIUM_ENABLED !== "true" || !isUserSubscribed()) {
+    return (
+      <Script id="redirect">window.location = &quot;results&quot;;</Script>
+    );
+  }
+
+  const result = await getOnerepProfileId(session.user.subscriber.id);
+  const profileId = result[0]["onerep_profile_id"];
+  await activateProfile(profileId);
+  await optoutProfile(profileId);
```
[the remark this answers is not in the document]
mansaj: I think the removal feature is only for premium. Both free/premium users will hit that URI?

src/app/(proper_react)/redesign/(authenticated)/user/welcome/details/page.tsx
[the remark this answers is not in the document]
rhelmer: I just removed it for now :P

src/app/(proper_react)/redesign/(authenticated)/user/welcome/results/page.tsx
```diff
@@ -0,0 +1,61 @@
...
+  }
+
+  const result = await getOnerepProfileId(session.user.subscriber.id);
+  const profileId = result[0]["onerep_profile_id"];
+  const scanResult = await getLatestOnerepScan(profileId);
+
+  const results = JSON.stringify(scanResult, null, 2);
+
+  const ctaButton = async () => {
+    if (process.env.PREMIUM_ENABLED) {
```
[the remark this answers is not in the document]
rhelmer: Yes I think we should be explicit here, truthiness might do the wrong thing

src/db/migrations/20230703010356_add-profile-table.js
rhelmer: Yeah I noticed that yesterday when rebasing and re-testing from scratch, thanks

src/app/api/v1/user/welcome/route.ts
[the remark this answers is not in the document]
rhelmer: done

src/db/tables/profiles.js
```diff
@@ -0,0 +1,46 @@
...
+      'state_code',
+      'date_of_birth',
+      'created_at'
+    )
+    .where('onerep_profile_id', onerepProfileId)
+}
+
+/**
+ * @param {number} onerepProfileId
+ * @param {import("../../app/functions/server/onerep.js")} profileData
```
mansaj, line 28:
```suggestion
 * @param {import("../../app/functions/server/onerep.js".ProfileData)} profileData
```

src/db/tables/profiles.js
```diff
@@ -0,0 +1,46 @@
...
+ */
+async function setProfileDetails (onerepProfileId, profileData) {
+  await knex('profiles')
+    .insert({
+      onerep_profile_id: onerepProfileId,
+      first_name: profileData.first_name,
+      last_name: profileData.last_name,
+      city_name: profileData.addresses[0].city_name,
```
mansaj, line 36: city

src/db/tables/profiles.js
```diff
@@ -0,0 +1,46 @@
...
+      state_code: profileData.addresses[0].state_name,
```
mansaj, line 37: state?

mansaj: In my `profile` table, these values aren't set
Screenshot 2023-07-11 at 4 20 50 PM

src/db/tables/onerep_scans.js
```diff
@@ -0,0 +1,82 @@
...
+    .where('onerep_profile_id', onerepProfileId)
+    .andWhere('onerep_scan_id', oneRepScanId)
+    .update({
+      onerep_scan_results: oneRepScanResults,
+      updated_at: knex.fn.now()
+    })
+}
+
+export {
+  getLatestOnerepScan,
```
mansaj, line 77: We should sync up on the verbiage for actions. 
So far we have inherited:
`get..`
`update...`
`add..`
`delete..`
`remove..`

and we are adding here:
`set...`

src/db/tables/profiles.js
rhelmer: Thanks!

src/db/migrations/20230703010356_add-profile-table.js
rhelmer: I'm not against it, I chose to not do it because it's data that's submitted directly to us, and it may not always be strictly limited to this... we can change it later, but there's a cost to migrating the data.

src/app/(proper_react)/redesign/(authenticated)/user/welcome/page.tsx
```diff
@@ -0,0 +1,58 @@
...
+        info. Rest assured we will safeguard your information like it’s our own.
+      </p>
+
+      <p>
+        We believe in your right to privacy, so basic Monitor is always free.
+        Upgrade to Premium for continuous protection and automatic removal of
+        your personal info.
+      </p>
+      <a href="welcome/details">
+        <button className="button primary">Start my free scan</button>
```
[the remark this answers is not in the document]
rhelmer: Last I checked that CSS hadn't made its way into the `(proper_react)` dir yet, but I have been keeping this branch up-to-date and it may work now, will check thanks.

src/app/(proper_react)/redesign/(authenticated)/user/welcome/remove/page.tsx
```diff
@@ -0,0 +1,102 @@
...
+      url: process.env.SERVER_URL,
+      images: ["/images/og-image.webp"],
+    },
+  };
+}
+
+export default async function UserWelcomeRemove() {
+  const session = await getServerSession(authOptions);
+  if (!session?.user?.subscriber?.id) {
+    throw new Error("No session");
```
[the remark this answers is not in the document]
rhelmer: We're going to replace this with a proper component that's based more closely on Figma, I don't want to build out a way to show errors to the users since that's happening in parallel elsewhere :)

src/app/(proper_react)/redesign/(authenticated)/user/welcome/scanning/page.tsx
```diff
@@ -0,0 +1,108 @@
...
+      iterations++;
+    }
+  }, 1000);
+
+  return (
+    <>
+      <Script
+        type="module"
+        src="/nextjs_migration/client/js/welcome.js"
+      ></Script>
```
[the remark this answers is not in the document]
rhelmer: Right, that's why I left this in the `nextjs_migration` dir :) I had some trouble mixing client and server side React components, I am eager to learn how this works.

src/app/api/v1/user/welcome/route.ts
```diff
@@ -0,0 +1,102 @@
...
+} from "../../../../../db/tables/subscribers";
+
+import {
+  getLatestOnerepScan,
+  setOnerepProfileId,
+} from "../../../../../db/tables/onerep_scans";
+import { setProfileDetails } from "../../../../../db/tables/profiles";
+
+export async function POST(req: NextRequest) {
+  const token = await getToken({ req });
```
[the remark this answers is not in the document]
rhelmer: Thanks, thought I had replaced this already.

src/app/api/v1/user/welcome/route.ts
```diff
@@ -0,0 +1,102 @@
...
+    ],
+    birth_date: params.get("dob") || "",
+  };
+
+  if (typeof token?.email === "string") {
+    try {
+      const subscriber = await getSubscriberByEmail(token.email);
+      // FIXME check if scan has already been performed and return early if so.
+
+      if (!subscriber.onerep_profile_id) {
```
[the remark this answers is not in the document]
rhelmer: I think we do want to use webhooks, but they are very difficult to test locally.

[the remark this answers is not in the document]
rhelmer: I think the problem is the way the ordering works here:

1. user must create a profile, and we must know the profile ID
2. a free scan is created, using the profile ID (which returns a scan ID)
3. a premium users profile may be activated, which causes automatically monthly scan
4. a premium users may trigger opt-out

Let's discuss further though, maybe I am not thinking through this clearly, but I the idea is that we can use the scan table to tell how many scans each user has performed, as well as other statistics such as total scans per time period, and how long each scan took.

src/db/tables/onerep_scans.js
rhelmer: > 👍 to this. I was going to file a ticket or ask in Jira, but do we have an exiting issue about converting all existing .js files .ts [where possible] for consistency? It'd be nice to have a repo that is mostly .ts and .tsx so we can easily share type definitions (versus JSDoc style).

No I don't think we do, we've been migrating as we go. At some point we need to excise a bunch of the old code (rather than convert it).

src/db/tables/onerep_scans.js
rhelmer: Hm, `add` feels like it's be additive... do you have a specific proposal @mansaj, do you think this should be e.g. `addOnerepProfileId`?

src/app/(proper_react)/redesign/(authenticated)/user/welcome/remove/page.tsx
```diff
@@ -0,0 +1,43 @@
...
+  }
+
+  if (process.env.PREMIUM_ENABLED !== "true" || !(await isUserSubscribed())) {
+    return (
+      <Script id="redirect">window.location = &quot;results&quot;;</Script>
```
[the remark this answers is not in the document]
rhelmer: Yes, but this is going to be replaced with proper client-side components (that's being worked on in parallel) so I think this is clear enough for demonstration.

src/app/(proper_react)/redesign/(authenticated)/user/welcome/details/page.tsx
```diff
@@ -0,0 +1,92 @@
...
+                  <tr>
+                    <td>
+                      <label htmlFor="citystate">City and state: </label>
+                    </td>
+                    <td>
+                      <input type="text" name="city" id="city" />
+                    </td>
+                    <td>
+                      <input type="text" name="state" id="state" />
+                    </td>
```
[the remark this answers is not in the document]
rhelmer: Yes, I think the JSON route that parses this is the more important bit (I didn't want to deal with parsing the city/state here since that's being worked on in parallel as part of a lookup component

src/app/(proper_react)/redesign/(authenticated)/user/welcome/details/page.tsx
```diff
@@ -0,0 +1,92 @@
...
+                    <td>
+                      <label htmlFor="lastname">Last name: </label>
+                    </td>
+                    <td>
+                      <input type="text" name="lastname" id="lastname" />
+                    </td>
+                  </tr>
+                  <tr>
+                    <td>
+                      <label htmlFor="citystate">City and state: </label>
```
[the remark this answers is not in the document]
rhelmer: Good catch thanks

pdehaan:
@rhelmer I think this looks good to me, but I know you said a lot of the nits were just local-only or temporary code while we work on other components.
Let me know if you want me to officially "Approve" this if it's somehow blocking you and you want to merge this ASAP to work on other PRs that might be blocked.

mansaj (approved):

src/external/onerep.js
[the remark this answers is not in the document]
Vinnl: Yeah, I know it has to be the first step on the OneRep side, but can't we just not insert that into our database until after the scan has completed?

src/app/(proper_react)/redesign/(authenticated)/user/welcome/remove/page.tsx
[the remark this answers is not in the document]
Vinnl: But can't the API just check whether the user has Premium and decide whether to call it in the API call?

src/app/api/v1/user/welcome/route.ts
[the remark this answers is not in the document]
Vinnl: Poorly-timed PTO on my side, but your ordering sounds correct, and the idea of keeping a `onerep_scans` table as well. My suggestion was only to remember the profile ID in memory after step 1, and then once step 2 is complete, to only then store it in the database.

PR #3158, opened 2023-06-22
Proxy company icons

Vinnl:
Downloading them to public/ after the app is built doesn't result in the server serving them, so this adds an API that reads the icons from disk and serves them up. This is really a stopgap solution; we really want to get around to downloading them in a cron job and storing them on our CDN, but I don't know enough of our infrastructure at the moment to know how to do that.

### DIFF
--- a/src/app/api/breach-icon/[breach-domain]/route.ts
+++ b/src/app/api/breach-icon/[breach-domain]/route.ts
@@ -1,0 +1,26 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import { readFile } from "node:fs/promises";
+import { dirname, resolve as pathResolve } from "node:path";
+import { fileURLToPath } from "node:url";
+import { NextResponse } from "next/server";
+
+export async function GET(
+  request: Request,
+  context: { params: { "breach-domain": string } }
+) {
+  const logoPath = pathResolve(
+    dirname(fileURLToPath(import.meta.url)),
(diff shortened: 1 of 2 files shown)


rhelmer:
> Downloading them to public/ after the app is built doesn't result in the server serving them, so this adds an API that reads the icons from disk and serves them up. This is really a stopgap solution; we really want to get around to downloading them in a cron job and storing them on our CDN, but I don't know enough of our infrastructure at the moment to know how to do that.

Thanks, yes I think adding it to cron is easy enough but we need to ensure the CDN is accessible to us and set up the way we need (there's a ticket for that filed already). We could store the icons in something else like Redis/Postgres, but I agree the CDN is probably the most appropriate thing to use here ultimately.

rhelmer (approved):

src/app/api/breach-icon/[breach-domain]/route.ts
```diff
@@ -0,0 +1,26 @@
...
+export async function GET(
+  request: Request,
+  context: { params: { "breach-domain": string } }
+) {
+  const logoPath = pathResolve(
+    dirname(fileURLToPath(import.meta.url)),
+    `../../../../../public/logo_cache/${context.params["breach-domain"]}.ico`
+  );
+  try {
+    const file = await readFile(logoPath);
```
Comment (pdehaan) [inline src/app/api/breach-icon/[breach-domain]/route.ts:19]: I know this is just a temporary solution, but any concerns here with a disk i/o operation on every logo request?

Comment (Vinnl) [reply src/app/api/breach-icon/[breach-domain]/route.ts:19]: I'd say yes, it feels pretty concerning to me. @mansaj is now looking at setting up the cronjob+CDN, so hopefully we won't need this PR.

Vinnl:
Superseded by https://github.com/mozilla/blurts-server/pull/3165 and https://github.com/mozilla/blurts-server/pull/3172.

PR #3160, opened 2023-06-22
Use client-only component for email input

flozia:
# References: 
Jira: [MNTOR-1863](https://mozilla-hub.atlassian.net/browse/MNTOR-1863)

# Description

Browser extensions — in our test case the Relay browser extension — are potentially causing a React hydration error when manipulating the input. Strangely enough, adding `suppressHydrationWarning` to the input and/or the parent `<form>` does not prevent this from happening. One possible solution would be to only render the input client-side until we have a better solution.

### DIFF
--- a/src/app/(nextjs_migration)/components/client/ExposureScanInput.tsx
+++ b/src/app/(nextjs_migration)/components/client/ExposureScanInput.tsx
@@ -1,0 +1,26 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+"use client";
+
+import dynamic from "next/dynamic";
+import { useL10n } from "../../../hooks/l10n";
+
+const ExposureScanInput = () => {
+  const l10n = useL10n();
+
+  return (
+    <input
+      id="scan-email-address"
+      name="email"
+      type="email"
+      placeholder={l10n.getString("exposure-landing-hero-email-placeholder")}
+      required
+    />
+  );
+};
+
+export default dynamic(() => Promise.resolve(ExposureScanInput), {
+  ssr: false,
+});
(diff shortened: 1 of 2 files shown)


rhelmer (approved):
lgtm - as discussed in Slack:

1. I think the Relay extension could be a better citizen here and not modify the DOM (or at least not modify it as soon), *but*

src/app/(nextjs_migration)/components/client/ExposureScanInput.tsx
```diff
@@ -0,0 +1,26 @@
...
+      name="email"
+      type="email"
+      placeholder={l10n.getString("exposure-landing-hero-email-placeholder")}
+      required
+    />
+  );
+};
+
+export default dynamic(() => Promise.resolve(ExposureScanInput), {
+  ssr: false,
```
rhelmer, line 25: Hm, is it required to do both this and `use client`? Or are the subtly different in some way? :)

rhelmer (approved):

Vinnl:
Yeah, I think this would be an endless game of whack-a-mole where we couldn't do server-side rendering at all, just because any extension could potentially mess with DOM updates. Extensions shouldn't mess with the DOM that extensively anyway, and if [this PR](https://github.com/mozilla/fx-private-relay-add-on/pull/462) finally gets merged some day, the Relay extension should be pretty harmless too. It'll probably still cause a warning, but I don't think it'll cause issues. Other extensions of course might.

Btw @rhelmer, I've also got Sentry logging of frontend errors enabled on one of my own projects, and we'll see _a lot_ of extension-generated errors in there. Just a heads-up :)

src/app/(nextjs_migration)/components/client/ExposureScanInput.tsx
flozia: `use client` makes this component a client component, but that does not mean they are client-only. The initial DOM elements are still being rendered server-side. Adding a dynamic export with `ssr: false` ensures that there is no part of the component that’s executed on the server.

flozia:
> So I'd close this PR, I think.

I’m fine with that. Just wanted to get this PR up so we have a possible fix in case the issue rises in priority — for now I’ll go ahead and close it.

PR #3161, opened 2023-06-23
Don’t ignore script preloading

flozia:
# References: 
Jira: [MNTOR-1863](https://mozilla-hub.atlassian.net/browse/MNTOR-1863)

# Description

PR https://github.com/mozilla/blurts-server/pull/3137 made scripts load before hydration.  Even though they are not waiting for hydration to load they are still not being preloaded on `stage`. The console complains that the request credentials are not matching and adding `crossOrigin=anonymous` should resolve the issue:
```
A preload for 'https://stage.firefoxmonitor.nonprod.cloudops.mozgcp.net/nextjs_migration/client/js/resizeObserver.js' is found, but is not used because the request credentials mode does not match. Consider taking a look at crossorigin attribute.
```


Vinnl (approved):
TIL!

PR #3165, opened 2023-06-23
MNTOR-1838: cronjob for breach logos

mansaj:
# References: 
Jira: MNTOR-1838

# Description
Adding logo fetching to an existing cronjob (sync breaches). On top of updating the latest breaches from HIBP, the script will also fetch logos based on the breaches db and upload them to a public bucket in AWS S3.

The frontend will use the logo assets hosted on s3 instead of local assets

# Screenshot (if applicable)
`http://s3.amazonaws.com/firefoxmonitor-dev-monitor-cdn-dev-static-website/`
Screenshot 2023-06-26 at 12 55 32 PM

# How to test
`node src/scripts/syncBreaches.js`

### DIFF
--- a/package.json
+++ b/package.json
@@ -12,7 +12,7 @@
     "build": "next build",
     "start": "next start",
     "lint": "stylelint '**/*.scss' && prettier --check './src' && next lint",
-    "fix": "prettier --write './src'",
+    "fix": "prettier --write './src' && next lint --fix && stylelint --fix '**/*.scss'",
     "test": "NODE_OPTIONS=--loader=testdouble c8 ava",
     "e2e": "playwright test src/e2e/",
     "functions": "nodemon --exec 'functions-framework --target=app --signature-type=http --source=src/cloud-functions'",
--- a/package.json
+++ b/package.json
@@ -49,7 +49,6 @@
     "npm": "8.19.3"
   },
   "dependencies": {
-    "@aws-sdk/client-s3": "^3.358.0",
     "@fluent/bundle": "^0.17.1",
     "@fluent/langneg": "^0.6.2",
     "@fluent/react": "^0.15.0",
@@ -60,6 +59,7 @@
     "@types/node": "^20.1.1",
     "@types/react": "^18.2.6",
     "@types/react-dom": "^18.2.4",
+    "aws-sdk": "^2.1404.0",
     "client-oauth2": "^4.3.3",
     "csrf-csrf": "^2.2.2",
     "dotenv": "^16.0.3",
--- a/package.json
+++ b/package.json
@@ -12,6 +12,7 @@
     "build": "next build",
     "start": "next start",
     "lint": "stylelint '**/*.scss' && prettier --check './src' && next lint",
+    "fix": "prettier --write './src'",
     "test": "NODE_OPTIONS=--loader=testdouble c8 ava",
     "e2e": "playwright test src/e2e/",
     "functions": "nodemon --exec 'functions-framework --target=app --signature-type=http --source=src/cloud-functions'",
@@ -48,6 +49,7 @@
     "npm": "8.19.3"
   },
   "dependencies": {
+    "@aws-sdk/client-s3": "^3.358.0",
     "@fluent/bundle": "^0.17.1",
     "@fluent/langneg": "^0.6.2",
     "@fluent/react": "^0.15.0",
--- a/src/scripts/syncBreaches.js
+++ b/src/scripts/syncBreaches.js
@@ -11,11 +11,12 @@
 
 import { req, formatDataClassesArray } from '../utils/hibp.js'
 import { getAllBreaches, upsertBreaches } from '../db/tables/breaches.js'
-import { mkdir, readdir } from "node:fs/promises";
+import { readdir } from "node:fs/promises";
 import { resolve as pathResolve } from "node:path";
 import { finished } from 'node:stream/promises';
-import fs from "node:fs";
+import { createWriteStream } from "node:fs";
 import { Readable } from 'node:stream';
+import os from 'node:os';
 import aws from 'aws-sdk';
 
 // Get breaches logos and uploads to s3
@@ -48,21 +49,19 @@ async function uploadToS3(fileName, fileStream) {
 }
 
 export async function getBreachIcons(breaches) {
-  let logoMap;
   const breachDomains = breaches
     .map((breach) => breach.Domain)
     .filter((breachDomain) => breachDomain.length > 0);
 
   // make logofolder if it doesn't exist
-  const logoFolder = "public/logo_cache/";
-  if (!fs.existsSync(logoFolder)) await mkdir(logoFolder);
+  const logoFolder = os.tmpdir();
+  console.log(`Logo folder: ${logoFolder}`)
 
   // read existing logos
   const existingLogos = await readdir(logoFolder);
   console.log(`existing logos: ${existingLogos.length}`)
 
-  // TODO: Batch to limit memory use?
-  const logoMapElems = (await Promise.all(
+  (await Promise.all(
     breachDomains.map(async (breachDomain) => {
       const logoFilename = breachDomain.toLowerCase() + ".ico";
       const logoPath = pathResolve(logoFolder, logoFilename);
@@ -74,14 +73,11 @@ export async function getBreachIcons(breaches) {
       const res = await fetch(
         `https://icons.duckduckgo.com/ip3/${breachDomain}.ico`);
       await uploadToS3(logoFilename, Buffer.from(await res.arrayBuffer()))
-      const fileStream = fs.createWriteStream(logoPath, { flags: 'wx' });
+      const fileStream = createWriteStream(logoPath, { flags: 'wx' });
       const bodyReadable = Readable.fromWeb(res.body)
       await finished(bodyReadable.pipe(fileStream));
     })
   ));
-
-  logoMap = new Map(logoMapElems.filter(e => e != null));
-
 }
 
 
--- a/src/scripts/syncBreaches.js
+++ b/src/scripts/syncBreaches.js
@@ -13,9 +13,9 @@ import { req, formatDataClassesArray } from '../utils/hibp.js'
 import { getAllBreaches, upsertBreaches } from '../db/tables/breaches.js'
 import { mkdir, readdir } from "node:fs/promises";
 import { resolve as pathResolve } from "node:path";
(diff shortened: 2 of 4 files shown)


src/scripts/syncBreaches.js
```diff
@@ -11,7 +11,83 @@
 
 import { req, formatDataClassesArray } from '../utils/hibp.js'
 import { getAllBreaches, upsertBreaches } from '../db/tables/breaches.js'
+import { mkdir, readdir } from "node:fs/promises";
+import { resolve as pathResolve } from "node:path";
+import { finished } from 'stream/promises';
+import fs from "fs";
```
pdehaan, line 17: Self note since this was semi-confusing to me initially (re: why we need "node:fs/promises" (L14) and "fs"):

- `fs.existsSync()` L59
- `fs.createWriteStream()` L77
- `fs.promises.mkdir()` L59
- `fs.promises.readdir()` L62

Not sure if we could possibly use [`fs.mkdirSync()`](https://nodejs.org/dist/latest-v18.x/docs/api/fs.html#fsmkdirsyncpath-options) and [`fs.readdirSync()`](https://nodejs.org/dist/latest-v18.x/docs/api/fs.html#fsreaddirsyncpath-options) and then not need the `fs:promises` async versions.

nit: can we prefix this with "node:fs" to be consistent with L14-15 above?

src/scripts/syncBreaches.js
```diff
@@ -11,7 +11,83 @@
 
 import { req, formatDataClassesArray } from '../utils/hibp.js'
 import { getAllBreaches, upsertBreaches } from '../db/tables/breaches.js'
+import { mkdir, readdir } from "node:fs/promises";
+import { resolve as pathResolve } from "node:path";
+import { finished } from 'stream/promises';
```
pdehaan, line 16: nit: can we prefix this w/ [`node:stream/promises`](https://nodejs.org/dist/latest-v18.x/docs/api/stream.html#streams-promises-api) for consistency?
```suggestion
import { finished } from 'node:stream/promises';
```

src/scripts/syncBreaches.js
```diff
@@ -11,7 +11,83 @@
...
+import fs from "fs";
+import { Readable } from 'stream';
```
pdehaan, line 18: Same here re: `node:stream`
```suggestion
import { Readable } from 'node:stream';
```

Possibly also sort so the fs* and stream* dependencies are closer.
I guess we could also possibly group these into a single, albeit confusing, import like:

```js
import { Readable, promises: stream  } from "node:stream";
```
Where now Readable is, well, whatever that is, and `stream` alias is the async promise-API methods. Although not sure that's any better/clearer, and now I'm just over-nit-ing.

src/scripts/syncBreaches.js
```diff
@@ -11,7 +11,83 @@
...
+  await s3.upload(uploadParams).promise()
+  console.log('Successfully uploaded data to ' + Bucket + '/' + fileName)
+} catch (err) {
+  console.error(err, err.stack)
+}
+}
+
+export async function getBreachIcons(breaches) {
+  let logoMap;
+  async function fetchBreachIcons() {
```
pdehaan, line 52: nit: not sure if we have a style guide or not, but sometimes nested function definitions can get confusing.
Can we move this up to the root level (since I think it's hoisted in scope anyways) or are we depending on some leaky variables or something?

src/scripts/syncBreaches.js
```diff
@@ -11,7 +11,83 @@
...
+          const logoFilename = breachDomain.toLowerCase() + ".ico";
+          const logoPath = pathResolve(logoFolder, logoFilename);
+          if (existingLogos.includes(logoFilename)) {
+            console.log('skipping ', logoFilename)
+            return;
+          }
+          console.log(`fetching: , ${logoFilename}`)
+          const res = await fetch(
+            `https://icons.duckduckgo.com/ip3/${breachDomain}.ico`);
+          const fileStream = fs.createWriteStream(logoPath, { flags: 'wx' });
```
pdehaan, line 77: Interesting, does this need an "x" bit? (Looks like the [default `flags` is already `w`](https://nodejs.org/api/fs.html#fscreatewritestreampath-options), but maybe just curious why the "x" is there). Or maybe I'm completely missing the point. I assumed it was "x" for executable, but looks like it might be something else. Per https://nodejs.org/api/fs.html#file-system-flags:
> - `'wx'`: Like `'w'` but fails if the path exists.

Unclear on what the word "fails" means here. For example, if this (or any other async `fetch()`, `finished()`, or `uploadToS3()` `throw`s an error, I'm not sure who/where is listening/catching. Would one bad logo fetch break the entire `.map()` loop and abort loading any other logos?

src/scripts/syncBreaches.js
```diff
@@ -11,7 +11,83 @@
...
+      .map((breach) => breach.Domain)
+      .filter((breachDomain) => breachDomain.length > 0);
+
+    // make logofolder if it doesn't exist
+    const logoFolder = "public/logo_cache/";
+    if (!fs.existsSync(logoFolder)) await mkdir(logoFolder);
+
+    // read existing logos
+    const existingLogos = await readdir(logoFolder);
+    console.log({existingLogos})
```
pdehaan, line 63: ✂️❓ 
```suggestion
```

src/scripts/syncBreaches.js
```diff
@@ -11,7 +11,83 @@
...
+            return;
+          }
+          console.log(`fetching: , ${logoFilename}`)
+          const res = await fetch(
+            `https://icons.duckduckgo.com/ip3/${breachDomain}.ico`);
+          const fileStream = fs.createWriteStream(logoPath, { flags: 'wx' });
+          const bodyReadable = Readable.fromWeb(res.body)
+          await finished(bodyReadable.pipe(fileStream));
+          await uploadToS3(logoFilename, res.fileStream)
+      })
```
pdehaan, line 81: Does this need to return a value to the `map()` iterator? I see we have an early exit above on L72, which I assume will return `null` or `undefined` which then gets filtered by the `.filter()` call below on L84. But I'm not clear on what these values would be and how items in `logoMapElems` gets populated.

pdehaan:
Curious when/where this future script will ultimately be called.
Or maybe I'm more confused as to how HIBP notifies us of new breaches. Should that new breach callback (if something like that exists) just fetch new logos and that way we only ever need to check for one single logo on the CDN? (not to imply that a new breach needs a logo, for example, if its a breach with `Domain: ""` or a breach of an existing domain breach where we already have a logo).

But I should also probably not comment on WIP/draft PRs since I'm just adding noise.

src/scripts/syncBreaches.js
mansaj: Taken out. thanks!

src/scripts/syncBreaches.js
```diff
@@ -11,7 +11,81 @@
...
+      const logoFilename = breachDomain.toLowerCase() + ".ico";
+      const logoPath = pathResolve(logoFolder, logoFilename);
+      if (existingLogos.includes(logoFilename)) {
+        console.log('skipping ', logoFilename)
+        return;
+      }
+      console.log(`fetching: , ${logoFilename}`)
+      const res = await fetch(
+        `https://icons.duckduckgo.com/ip3/${breachDomain}.ico`);
+      await uploadToS3(logoFilename, Buffer.from(await res.arrayBuffer()))
```
pdehaan, line 76: Does S3 care if the file is already there?
If I'm following this correctly the script will check if files exist LOCALLY and if not, it will fetch them from DDG and then upload to S3. So I guess I'm mildly curious on whether a file is missing locally but possibly already exists in S3, will S3 throw errors.

I guess the alternative is to not worry about local assets and query S3 bucket if a logo already exists and if so, skip and if not, upload. Although that comes at the expense of an extra 670ish GET requests to S3 to see if images exist (and 99.999999% they will).

Or conversely, if we're running the cron every... 6-24 hours, we could just filter the breaches by records that have a `ModifiedDate` that is around the same cron period. Possibly not perfect, but means 99.9999% of the time we don't need to fetch or check-for _any_ images in S3 if there haven't been any changes to the hibp breaches feed.

src/scripts/syncBreaches.js
```diff
@@ -11,7 +11,81 @@
...
+      const fileStream = fs.createWriteStream(logoPath, { flags: 'wx' });
+      const bodyReadable = Readable.fromWeb(res.body)
+      await finished(bodyReadable.pipe(fileStream));
+    })
+  ));
+
+  logoMap = new Map(logoMapElems.filter(e => e != null));
```
pdehaan, line 83: I don't think we're using `logoMap` anywhere. It seems to be local to this fn (see L51 above) but we aren't returning it from this fn; not that it mattes since the `getBreachIcons()` call on L104 below doesn't expect/use/want a return value.

src/scripts/syncBreaches.js
pdehaan: Or here's a rough version where I scraped the S3 bucket XML's listing and scanned for missing logos: https://github.com/pdehaan/blurts-s3-logos
Hopefully there is a better method to check S3 using the S3 API versus me scraping XML, but hey... I only actually needed 2 network requests to see which breaches from HIBP were missing from the S3 bucket (one for HIBP /breaches and the other for the [S3 bucket](http://s3.amazonaws.com/firefoxmonitor-dev-monitor-cdn-dev-static-website/)).

Spoiler alert: No logos were missing in S3, so I had to fake an item into the `breachDomains` Set to see if my logo checking was working.

Vinnl: I think it makes sense to overwrite existing logos, since companies might update their logos anyway, so we probably want the newest versions.

mansaj: S3 upload will always overwrite / replace. S3 has a concept of versioning that's built in so the old version is preserved by default.. though we can configure it (SRE) with different retention policies etc. Yea I think we are ok doing an overwrite here

package.json
```diff
@@ -58,6 +59,7 @@
     "@types/node": "^20.1.1",
     "@types/react": "^18.2.6",
     "@types/react-dom": "^18.2.4",
+    "aws-sdk": "^2.1404.0",
```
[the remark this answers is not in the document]
mansaj: Yup.. hmm

src/scripts/syncBreaches.js
mansaj: Took it out. thanks for catching this!

PR #3166, opened 2023-06-24
MNTOR-1403 - land mock data, type definition, and faker code for OneRep types

rhelmer:
# References: 
Jira: MNTOR-1403

# Description

I split this from https://github.com/mozilla/blurts-server/pull/3157 to make it easier to build out the front-end and write tests while the core scan/removal functionality is being worked on:

- type definition for OneRep `ScanResult` (and others)
- function to generate mock `ScanResult` data using faker
- static mock data (generated by above)


src/apiMocks/mockData.ts
```diff
@@ -0,0 +1,202 @@
...
+import { OneRepScanResult } from "../customTypes";
+
+// This is a full list of scan results, all pages, and would normally be
+// stored in the `onerep_scan_results.onerep_scan_results` column
+// as `jsonb`.
+export const mockedOneRepScanResult = {
+  data: [
+    {
+      id: 3038059606048768,
+      age: 3392703368790016,
```
pdehaan, line 15: Unclear on this is a "real" mock `age` or a random number, but&hellip;

Does OneRep not have public API docs?
I found https://onerep.com/api but it seems to want me to submit a request to OneRep to see if I "qualify" to read their API docs.

I had assumed from the schema below that `age` would have been my age in years. But even if it's milliseconds that isn't making sense.

If that is a JavaScript date in ms, it looks like it's from the year 109480, which seems implausible. Even if i divide by 1000 and reparse, it thinks the date is 2077. So I must be totally misunderstanding what this `age` is if it isn't my age or possibly milliseconds.
```js
new Date(3392703368790016).toLocaleDateString() // '6/27/109480'
new Date(3392703368790016/1000).toLocaleDateString() // '7/5/2077'
```

Oh, unless it's nanoseconds:

```js
new Date(3392703368790016 / 1000 / 1000).toLocaleDateString() // '2/8/1970'
new Date(3392703368790016 / 1000 / 1000).getTime() // 3_392_703_368

const MS_IN_DAY = 1000 * 60 * 60 * 24; // 86_400_000
console.log(3_392_703_368 / 86_400_000); // 39.27 days
```

src/apiMocks/mockData.ts
```diff
@@ -0,0 +1,202 @@
...
+
+/**
+ * Generates scan result with randomly-generated mock data.
+ *
+ * @returns {OneRepScanResult} - A single scan result.
+ */
+export function createRandomScan(): OneRepScanResult {
+  return {
+    id: faker.number.int() || null,
+    age: faker.number.int(),
```
pdehaan, line 157: I left a confusing, rambling message above, but if this is meant to be a human age, we might want to cap this at something more reasonable (like 100 or 105).

https://fakerjs.dev/api/number.html#int
```js
age: faker.number.int(100),
age: faker.number.int({ max: 105 }),
age: faker.number.int({ min: 14, max: 97 });
```

src/apiMocks/mockData.ts
```diff
@@ -0,0 +1,202 @@
...
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import { faker } from "@faker-js/faker";
+import { OneRepScanResult } from "../customTypes";
+
+// This is a full list of scan results, all pages, and would normally be
+// stored in the `onerep_scan_results.onerep_scan_results` column
+// as `jsonb`.
+export const mockedOneRepScanResult = {
```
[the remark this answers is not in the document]
pdehaan: Personal curiosity, but why hardcode the object versus just creating an array w/ 3 `createRandomScan()` results? Or do we need stable mock values for test asserts?

src/customTypes.d.ts
```diff
@@ -80,3 +80,29 @@ interface Window {
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
   gtag: any;
 }
+
+export type OneRepScanResult = {
```
Vinnl, line 84: I think it's probably good to limit this file to limit this file to providing [ambient type definitions](https://www.typescriptlang.org/docs/handbook/modules.html#working-with-other-javascript-libraries), and (especially now that we can write TS ourselves) define types returned by our own functions next to the code where data matching that shape originates.

I'll also note that there is already a type for the API response that OneRep provides for Scan Results: https://github.com/mozilla/blurts-server/blob/75e7d70988ddc122e2882feaa69aa807d0218819/src/external/onerep.js#L249-L264

If this is defines the structure of database columns that we define ourselves, it's probably a good idea not to name it `OneRep`. In fact, Knex has a more standard way to define types for it table rows - let me create a separate PR to demonstrate that.

Vinnl: https://github.com/mozilla/blurts-server/pull/3167

src/apiMocks/mockData.ts
```diff
@@ -0,0 +1,202 @@
...
+    link: faker.internet.url(),
+    emails: [
+      faker.internet.exampleEmail(),
+      faker.internet.exampleEmail(),
+      faker.internet.exampleEmail(),
+    ],
+    phones: [faker.phone.number(), faker.phone.number(), faker.phone.number()],
+    status: "new", // state is always "new" by default, see the `OneRepScanResult` type definition.
+    scan_id: faker.number.int(),
+    addresses: [
```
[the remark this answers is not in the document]
rhelmer: Nice, thanks!

src/customTypes.d.ts
rhelmer: OK I ended up extracting the types from the code you landed a while back (thanks for pointing that out!) and added its own `.d.ts` file, I might not totally be understanding where you want to go wrt types in Knex but we can discuss further. I stopped using `OneRep` in the type defs.

rhelmer:
I went ahead and made faker generate this data on-the-fly, and set faker to always use the same random seed, so we will get stable results for testing if things change. 

@codemist this means you can just call `mockedOneRepScanResults()` and it will return the same results every time, but if we want to change things it's easy to do so. Does this work for your purposes? Thanks!

src/apiMocks/mockData.ts
```diff
@@ -0,0 +1,58 @@
...
+const fakerSeed = 123;
+
+// This is a full list of scan results, all pages, and would normally be
+// stored in the `onerep_scan_results.onerep_scan_results` column
+// as `jsonb`.
+export function mockedOneRepScanResults() {
+  faker.seed(fakerSeed);
+  return {
+    data: Array.from({ length: 3 }, () => createRandomScan()),
+  };
```
pdehaan, line 20: 😍

src/app/functions/server/onerep.d.ts
```diff
@@ -0,0 +1,72 @@
...
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+export type ProfileData = {
+  first_name: string;
+  last_name: string;
+  city: string;
+  state: import("../../../utils/states").StateAbbr;
+  birth_date?: import("../../../utils/parse.js").ISO8601DateString;
+  phone_number?: import("../../../utils/parse.js").E164PhoneNumberString;
```
pdehaan, line 11: nit: can we `import` these types above on L4-ish?
Seems like we're using:
- `import("../../../utils/states")` 1&times;
- `import("../../../utils/parse.js")` 6&times;
    - `ISO8601DateString` 5&times;
    - `E164PhoneNumberString` 1&times;

src/app/functions/server/onerep.ts
```diff
@@ -72,26 +55,18 @@ export async function createProfile(profileData) {
...
-   *   url: string,
-   * }}
-   */
-  const savedProfile = await response.json();
+
+  const savedProfile: {
+    id: number;
+    status: "active" | "inactive";
+    created_at: import("../../../utils/parse.js").ISO8601DateString;
+    updated_at: import("../../../utils/parse.js").ISO8601DateString;
```
pdehaan, line 63: Re: Since we're already importing from `../../../utils/parse.js` above in https://github.com/mozilla/blurts-server/blob/4eee0ae6bfbfcf0c5474f8f144127c1fbb3c52be/src/app/functions/server/onerep.ts#L7-L10

Can we import the `ISO8601DateString` type above on L9.5 or something and then just use the type here instead of dynamic type imports?:

```suggestion
    created_at: ISO8601DateString;
    updated_at: ISO8601DateString;
```

src/app/functions/server/onerep.ts
```diff
@@ -2,17 +2,19 @@
...
+} from "../../../utils/parse.js";
+import { usStates } from "../../../utils/states.js";
 const log = mozlog("external.onerep");
 
-/**
- * @param {string} path
- * @param {Parameters<typeof fetch>[1]} [options]
- */
-async function onerepFetch(path, options = {}) {
+async function onerepFetch(
```
pdehaan, line 14: nit: should this be `onerepFetch()` or `oneRepFetch()`?

src/utils/states.d.ts
```diff
@@ -0,0 +1,67 @@
...
+  "WY",
+  "AS",
+  "GU",
+  "MP",
+  "PR",
+  "VI",
+  "UM",
+  "MH",
+  "FM",
+  "PW"
```
pdehaan, line 65: :+1:

src/app/functions/server/onerep.ts
Vinnl: If it should be `oneRepFetch` then it should also be `uSStates`, so... I'd stay with `onerepFetch`.

src/app/functions/server/onerep.d.ts
```diff
@@ -0,0 +1,72 @@
...
+  from: number;
+  last_page: number;
+  per_page: number;
+  to: number;
+  total: number;
+};
+export type Scan = {
+  id: number;
+  profile_id: number;
+  status: "in_progress" | "finished";
```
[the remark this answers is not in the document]
Vinnl: These type represent what OneRep returns to us, which does not include `action_needed`. I think we'll track separately in our database whether a free user has manually resolved an exposure - do you know what the plan for that is Rob? If so, the appropriate place for that would then be in our [database types](https://github.com/mozilla/blurts-server/blob/main/src/knex-tables.d.ts).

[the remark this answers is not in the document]
rhelmer: In PR #3157 I've added a new table `onerep_scans`, yes I think this kind of record-keeping should be there.

One thing to keep in mind is the result of a single scan might return many individual records across different brokers, so if we need users to be able to manually resolve individual breaches then we either need to modify the `onerep_scan_results` which is JSON, or we need to have some other way (e.g. a table to map them). I think we should resolve this in #3157

src/app/functions/server/onerep.d.ts
```diff
@@ -0,0 +1,72 @@
...
+  addresses: Array<{
+    city: string;
+    state: string;
+    street: string;
+    zip: string;
+  }>;
+  relatives: string[];
+  phones: string[];
+  emails: string[];
+  data_broker: string;
```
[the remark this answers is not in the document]
rhelmer: Hm, looking at the sample data provided:

```
"link": "https://example.com/john-smith",
"data_broker": "example.com",
```

I believe we're using the S3 CDN again for icons? So I think we can use the same endpoint that we do for breaches.

src/utils/parse.d.ts
Vinnl, line 1: Since this types our own code, I don't think we need this `.d.ts` file. (The `.js` file already contains type definitions - at most, we could convert that entire file to a `.ts` file.)

src/utils/states.d.ts
Vinnl, line 1: Since this types our own code, I don't think we need this `.d.ts` file. (The `.js` file already contains type definitions - at most, we could convert that entire file to a `.ts` file.)

src/appConstants.d.ts
Vinnl, line 1: This appears to be some unused remnant from a typings generator?

Vinnl (commented):
I think the other three `.d.ts` files are unneeded as well :)

src/appConstants.d.ts
rhelmer: Yes all of these are from when I ran `tsc` and I left them because those are still JS files, but since we're using jsdoc annotations in the JS files (and will presumably convert these to .ts files with the definitions inline at some point). I think your reasoning makes sense, thanks!

rhelmer:
I just tested to make sure it actually still works after the latest round of changes and looks reasonable:
```
{
  id: 5135502900461568,
  age: 72,
  url: 'https://single-dispute.info',
  link: 'https://composed-value.info/',
  emails: [
    'Hilma22@example.com',
    'Nelson52@example.net',
    'Mackenzie.Hoppe-Thiel@example.net'
  ],
  phone_numbers: [
    '929-931-4860 x512',
    '(542) 885-8452 x0283',
    '(284) 684-2928 x184'
  ],
  status: 'new',
  scan_id: 892470038626304,
  addresses: [
    {
      zip: '59738',
      city: 'Lake Tatyanafurt',
      state: 'SD',
      street: 'Antwan Hill'
    },
    {
      zip: '85168',
      city: 'Lake Hermanntown',
      state: 'WI',
      street: 'Luettgen Pike'
    },
    {
      zip: '55195',
      city: 'Uptonfield',
      state: 'KY',
      street: 'Dandre Expressway'
    }
  ],
  last_name: 'Stiedemann',
  relatives: [ 'Delia Hayes', 'Evan Mraz II', 'Kelly Williamson' ],
  created_at: '2023-07-05T09:58:03.531Z',
  first_name: 'Adrain',
  profile_id: 6544684195250176,
  updated_at: '2023-07-04T19:09:29.123Z',
  data_broker: 'remorseful-farmer.org',
  middle_name: 'Micah',
  data_broker_id: 8166515205472256,
  birth_date: '1993-12-26T19:35:54.641Z'
}
```

I'm having a little trouble convincing typescript that Fake will generate probably state code abbreviations, I used `ts-ignore` for now but if there's an obvious fix it'd be nice to get that too.

pdehaan:
Looking at @rhelmer's sample data in https://github.com/mozilla/blurts-server/pull/3166#issuecomment-1622075259 and https://docs.onerep.com/#operation/getScanResult i found a few possible nits:
1. OneRep seemingly returning `age`s as "string" and we're using "number/int" in our faker results. Not a big deal, but 🤷.
2. Not sure if I'm looking at the correct response/payload, but I didn't see `birth_date` or `scan_id` in the https://docs.onerep.com/#operation/getScanResult "200" response payload. But I do see `birth_date` on https://docs.onerep.com/#operation/getProfiles so maybe I'm looking at the wrong API endpoints or the docs are misleading.
3. OneRep's `middle_name` _might_ be a single character where we're returning a full name. I did a Ctrl+F in the docs page and all 11/11 results seemed to just be an initial.
4. Our Faker addresses have no house number, only street name (and city/state/zip).
5. Our Faker data might have `phone_numbers[]` key, but https://docs.onerep.com/#operation/getScanResult seems to use `phones[]`. If I look at https://docs.onerep.com/#operation/getProfiles endpoint I see an inconsistently named `phone_numbers[]` result, but it's also an array of objects not array of strings.
    - https://docs.onerep.com/#operation/getScanResult `"phones": ["..."]`
    - https://docs.onerep.com/#operation/getScanResults `"phones": ["..."]`
    - https://docs.onerep.com/#operation/activateProfile `"phone_numbers": [{...}]`
    - https://docs.onerep.com/#operation/createProfile `"phone_numbers": [{...}]`
    - https://docs.onerep.com/#operation/deactivateProfile `"phone_numbers": [{...}]`
    - https://docs.onerep.com/#operation/getProfile `"phone_numbers": [{...}]`
    - https://docs.onerep.com/#operation/getProfiles `"phone_numbers": [{...}]`
    - https://docs.onerep.com/#operation/updateProfile `"phone_numbers": [{...}]`

src/apiMocks/mockData.ts
```diff
@@ -0,0 +1,64 @@
...
+      faker.phone.number(),
+      faker.phone.number(),
+    ],
+    status: "new", // state is always "new" by default, see the `ScanResult` type definition.
+    scan_id: faker.number.int(),
+    //@ts-ignore FIXME figure out how to convince TS that Fake will return a valid abbreviated state
+    addresses: Array.from({ length: 3 }, () => ({
+      zip: faker.location.zipCode(),
+      city: faker.location.city(),
+      state: faker.location.state({ abbreviated: true }),
```
Vinnl, line 50: You can use [type assertions](https://www.typescriptlang.org/docs/handbook/2/everyday-types.html#type-assertions) to tell TS that you know better than it what the type is going to be:

```suggestion
    addresses: Array.from({ length: 3 }, () => ({
      zip: faker.location.zipCode(),
      city: faker.location.city(),
      state: faker.location.state({ abbreviated: true }) as StateAbbr,
```

tsconfig.json
```diff
@@ -33,7 +33,7 @@
     // "types": [],                                      /* Specify type package names to be included without being referenced in a source file. */
     // "allowUmdGlobalAccess": true,                     /* Allow accessing UMD globals from modules. */
     // "moduleSuffixes": [],                             /* List of file name suffixes to search when resolving a module. */
-    // "allowImportingTsExtensions": true,               /* Allow imports to include TypeScript file extensions. Requires '--moduleResolution bundler' and either '--noEmit' or '--emitDeclarationOnly' to be set. */
+    "allowImportingTsExtensions": true,               /* Allow imports to include TypeScript file extensions. Requires '--moduleResolution bundler' and either '--noEmit' or '--emitDeclarationOnly' to be set. */
```
Vinnl, line 36: Huh - just wondering, why was this necessary?

Vinnl (approved):
Looking good!

tsconfig.json
rhelmer: Oops, sorry I wrote a little test script and enabled it for convenience, will remove thanks!

src/apiMocks/mockData.ts
rhelmer: Oh, that's what I've been missing. Thanks!

rhelmer:
> Looking at @rhelmer's sample data in [#3166 (comment)](https://github.com/mozilla/blurts-server/pull/3166#issuecomment-1622075259) and https://docs.onerep.com/#operation/getScanResult i found a few possible nits:
> 
>     1. OneRep seemingly returning `age`s as "string" and we're using "number/int" in our faker results. Not a big deal, but 🤷.
> 
>     2. Not sure if I'm looking at the correct response/payload, but I didn't see `birth_date` or `scan_id` in the https://docs.onerep.com/#operation/getScanResult "200" response payload. But I do see `birth_date` on https://docs.onerep.com/#operation/getProfiles so maybe I'm looking at the wrong API endpoints or the docs are misleading.
> 
>     3. OneRep's `middle_name` _might_ be a single character where we're returning a full name. I did a Ctrl+F in the docs page and all 11/11 results seemed to just be an initial.
> 
>     4. Our Faker addresses have no house number, only street name (and city/state/zip).
> 
>     5. Our Faker data might have `phone_numbers[]` key, but https://docs.onerep.com/#operation/getScanResult seems to use `phones[]`. If I look at https://docs.onerep.com/#operation/getProfiles endpoint I see an inconsistently named `phone_numbers[]` result, but it's also an array of objects not array of strings.
>        
>        * https://docs.onerep.com/#operation/getScanResult `"phones": ["..."]`
>        * https://docs.onerep.com/#operation/getScanResults `"phones": ["..."]`
>        * https://docs.onerep.com/#operation/activateProfile `"phone_numbers": [{...}]`
>        * https://docs.onerep.com/#operation/createProfile `"phone_numbers": [{...}]`
>        * https://docs.onerep.com/#operation/deactivateProfile `"phone_numbers": [{...}]`
>        * https://docs.onerep.com/#operation/getProfile `"phone_numbers": [{...}]`
>        * https://docs.onerep.com/#operation/getProfiles `"phone_numbers": [{...}]`
>        * https://docs.onerep.com/#operation/updateProfile `"phone_numbers": [{...}]`

Thanks, it's very possible I got out of sync (and it's easy to mix up the input e.g. the form POST sent to OneRep vs. the output that we get back from their API and/or webhooks). I'll go over everything and compare to the docs. 

The publish using OpenAPI, I wonder if there's a way to automatically consume those and convert to TS types and so on?

rhelmer:
OK I've aligned the types and faker data with the OneRep API - as we build out tests we'll be able to tell if there any small differences.

src/app/(proper_react)/Dashboard.stories.tsx
```diff
@@ -27,7 +28,7 @@ const ScanResultMockItem: ScanResult = {
   addresses: [
     {
       city: "123",
-      state: "State",
+      state: "State" as StateAbbr,
```
rhelmer, line 31: I made a few small corrections to existing mocks - I wasn't sure if you wanted to use the new API mock class here or not @codemist? I can add that if so

src/app/functions/server/onerep.ts
```diff
@@ -0,0 +1,283 @@
...
+    id: number;
+    status: "active" | "inactive";
+    created_at: ISO8601DateString;
+    updated_at: ISO8601DateString;
+    url: string;
+  } = await response.json();
+  return savedProfile.id;
+}
+
+export function parseExposureScanData(body: ScanResult): ProfileData | null {
```
rhelmer, line 135: It took me a little bit to understand what this was intended for (I had to review the older controller code)... I think in the latest code there isn't really a call for parsing the scan result and outputting `ProfileData`, we're treating them separately like this:

1. `ProfileData` is the data that users enter in the form (first/last name, city, state, dob)
2. `ScanResults` from the remote API call are stored as-is in our database

Profiles and scan results are stored in separate tables and can be joined by profile ID. There may be more than one scan per user (they are differentiated by ID), but I think we want to keep these as separate classes of data.

rhelmer: That said, for the moment we are allowing `ProfileData` to contain an optional `age` or `birth_date` field to keep the existing code from breaking, but it's something we can likely clean up. I'm not sure if we will need this parsing code at that point, but it could be re-purposed as a runtime guard perhaps.

src/apiMocks/mockData.ts
```diff
@@ -0,0 +1,54 @@
...
+    last_name: faker.person.lastName(),
+    middle_name: faker.person.middleName(),
+    age: faker.number.int({ min: 14, max: 120 }).toString(),
+    status: "new", // state is always "new" by default, see the `ScanResult` type definition.
+    addresses: Array.from({ length: 3 }, () => ({
+      zip: faker.location.zipCode(),
+      city: faker.location.city(),
+      state: faker.location.state({ abbreviated: true }) as StateAbbr,
+      street: faker.location.street(),
+    })),
```
pdehaan, line 43: Interesting, I just noticed that addresses don't have countries. Do we think this feature will ever extend outside of the USA?

src/apiMocks/mockData.ts
```diff
@@ -0,0 +1,54 @@
...
+    phones: [faker.phone.number()],
+    emails: [faker.internet.exampleEmail()],
```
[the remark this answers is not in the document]
rhelmer: Yes, I don't really like using actually-routable emails in tests, just in case we accidentally send a test email etc.

src/apiMocks/mockData.ts
rhelmer: No plans at this time, they would need to extend their API (likely by adding a new field, but no reason to worry yet :))

rhelmer:
> Looking at @rhelmer's sample data in [#3166 (comment)](https://github.com/mozilla/blurts-server/pull/3166#issuecomment-1622075259) and https://docs.onerep.com/#operation/getScanResult i found a few possible nits:
> 
>     1. OneRep seemingly returning `age`s as "string" and we're using "number/int" in our faker results. Not a big deal, but 🤷.
> 
>     2. Not sure if I'm looking at the correct response/payload, but I didn't see `birth_date` or `scan_id` in the https://docs.onerep.com/#operation/getScanResult "200" response payload. But I do see `birth_date` on https://docs.onerep.com/#operation/getProfiles so maybe I'm looking at the wrong API endpoints or the docs are misleading.
> 
>     3. OneRep's `middle_name` _might_ be a single character where we're returning a full name. I did a Ctrl+F in the docs page and all 11/11 results seemed to just be an initial.
> 
>     4. Our Faker addresses have no house number, only street name (and city/state/zip).
> 
>     5. Our Faker data might have `phone_numbers[]` key, but https://docs.onerep.com/#operation/getScanResult seems to use `phones[]`. If I look at https://docs.onerep.com/#operation/getProfiles endpoint I see an inconsistently named `phone_numbers[]` result, but it's also an array of objects not array of strings.
>        
>        * https://docs.onerep.com/#operation/getScanResult `"phones": ["..."]`
>        * https://docs.onerep.com/#operation/getScanResults `"phones": ["..."]`
>        * https://docs.onerep.com/#operation/activateProfile `"phone_numbers": [{...}]`
>        * https://docs.onerep.com/#operation/createProfile `"phone_numbers": [{...}]`
>        * https://docs.onerep.com/#operation/deactivateProfile `"phone_numbers": [{...}]`
>        * https://docs.onerep.com/#operation/getProfile `"phone_numbers": [{...}]`
>        * https://docs.onerep.com/#operation/getProfiles `"phone_numbers": [{...}]`
>        * https://docs.onerep.com/#operation/updateProfile `"phone_numbers": [{...}]`

Thanks, I think I got all of these except for middle name - good point about their examples but the docs just say "string" not "character", so I think we shouldn't sweat this one. We'll see what the real production data looks like soon enough.

src/app/functions/server/onerep.ts
```diff
@@ -0,0 +1,283 @@
...
+  | "optout_in_progress"
+  | "waiting_for_verification"
+  | "removed";
+
+async function onerepFetch(
+  path: string,
+  options: Parameters<typeof fetch>[1] = {}
+) {
+  const url = "https://api.onerep.com" + path;
+  const headers = new Headers(options.headers);
```
mansaj, line 90: perhaps this part (setting header for bearer token) can be stripped out into an util thing? 
We have a corresponding getter in `/app/api/utils/auth.ts`

PR #3167, opened 2023-06-26
Add type definitions for database queries

Vinnl:
This allows many Knex operations to be typed, and allows us to track where the data we're dealing with came directly from the database.

Note that I marked the `updated_at` columns as required on updates, so this also updates a couple of updates to actually update that column.


src/customTypes.d.ts
```diff
@@ -2,37 +2,6 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
-import { Subscriber } from "./app/(nextjs_migration)/(authenticated)/user/breaches/breaches";
-
-interface HTMLElement {
```
Vinnl, line 7: These were to tack on types for our partials when they were generated in controllers, and thus are no longer necessary.

src/knex-tables.d.ts
```diff
@@ -0,0 +1,148 @@
...
+    is_spam_list: boolean;
+    is_malware: boolean;
+  }
+  type BreachOptionalColumns = Extract<
+    keyof BreachRow,
+    "domain" | "description"
+  >;
+  type BreachAutoInsertedColumns = Extract<keyof BreachRow, "id">;
+
+  interface Tables {
```
Vinnl, line 115: They keys in this interface correspond to database tables, and they're used when calling e.g. `knex("subscribers")` to use the correct types for subsequent operations, if possible. See https://knexjs.org/guide/#typescript.

mansaj (approved):
Thanks for this... This reminded me.. I'll open a PR to clean up the `subscriber` table a little bit

PR #3168, opened 2023-06-27
Port tests to Jest

Vinnl:
Most notable to look out for:

- I re-added the `getCountryCode` to only show certain recommendations in certain countries. (It got lost in the port over, so we showed the same recommendations as in the US everywhere.)
- To keep the cloud function tests compatible with the Next.js setup, I had to remove the top-level await and wrap it in an instantly-invoked `init` function. Ideally, we fix the discrepancy by creating separate `package.json`s for the website, cloud functions, and cron jobs, and have each have its own independent pipeline and test environment.
- I did not port `settings.test.js`. The code under tests has now mostly moved to `api/v1/user/email/route.ts` and `api/v1/user/verify-email/route.ts`, but I couldn't quite figure out yet how to mock out `NextResponse` and Next-Auth. There are probably ways around that (e.g. extracting out the relevant code to separate modules), but since the original tests have a lot commented-out, and mock out almost all the actual code to only verify responses and status codes, which is fairly trivial code, I opted to just not port that right now.

The Jest configuration was generated using `jest --init`.


jest.config.cjs
```diff
@@ -0,0 +1,214 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
```
pdehaan, line 1: A *.cjs file, haven't seen one of those in a while!
If we can't convert this to .ts or .js we should make sure this is getting picked up by ESLint+Prettier workflows.

rhelmer (commented):
It looks like .ts is supported: https://jestjs.io/docs/configuration

rhelmer (approved):

jest.config.cjs
Vinnl: Good catch, f1e3cbf92.

Unfortunately Jest can't run in an MJS context, which is the default for `.js` and `.ts` files in our project (because of the `"type": "module"` in our `package.json`), so using a different extension results in errors like this one:

```
SyntaxError: Cannot use import statement outside a module
```

PR #3171, opened 2023-06-28
Update to AWS SDK v3, and don't store 404'd logos

Vinnl:
This migration was done using

    npx aws-sdk-js-codemod -t v2-to-v3

As listed on https://www.npmjs.com/package/aws-sdk:

> AWS SDK for JavaScript v3 is the latest and recommended version,
> which has been GA since December 2020.

The new version also modularises the different clients, hopefully
resulting in less frequent dependency updates that mostly do not
affect us.

I was not able to test this yet, since I don't have the credentials. This includes the `deleteFromS3` function - which I'm not sure we need long term, but it would be good to remove the 404 icons we've already stored on S3.

### DIFF
--- a/src/scripts/syncBreaches.js
+++ b/src/scripts/syncBreaches.js
@@ -18,7 +18,7 @@ import { createWriteStream } from "node:fs";
 import { Readable } from 'node:stream';
 import os from 'node:os';
 import { Upload } from "@aws-sdk/lib-storage";
-import { S3, DeleteObjectCommand } from "@aws-sdk/client-s3";
+import { S3 } from "@aws-sdk/client-s3";
 
 // Get breaches logos and uploads to s3
 const accessKeyId = process.env.AWS_ACCESS_KEY_ID;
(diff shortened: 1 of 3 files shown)


src/scripts/syncBreaches.js
```diff
@@ -72,10 +92,14 @@ export async function getBreachIcons(breaches) {
...
-      const fileStream = createWriteStream(logoPath, { flags: 'wx' });
-      const bodyReadable = Readable.fromWeb(res.body)
-      await finished(bodyReadable.pipe(fileStream));
+      if (res.status === 200) {
+        await uploadToS3(logoFilename, Buffer.from(await res.arrayBuffer()))
+        const fileStream = createWriteStream(logoPath, { flags: 'wx' });
+        const bodyReadable = Readable.fromWeb(res.body)
+        await finished(bodyReadable.pipe(fileStream));
+      } else {
+        await deleteFromS3(logoFilename);
```
mansaj, line 101: Is there a reason we'd want to explicitly delete an old `Ico` if it exists? I think it would probably be safer keeping it in case `https://icons.duckduckgo.com/ip3` is temporarily offline (then our bucket will be emptied)

Alternatively, I'd suggest fallback on a google URI as a mechanism to fetch in case `icons.duckduckgo.com` is offline

Vinnl: Yeah I'm not sure, but I at least wanted to delete the 404 icons that we've already uploaded to the dev bucket. But removed the deletion in de63bc95e.

mansaj: Thanks. Let me test this locally real quick

mansaj (approved):
Tested locally

src/scripts/syncBreaches.js
```diff
@@ -17,15 +17,16 @@ import { finished } from 'node:stream/promises';
 import { createWriteStream } from "node:fs";
 import { Readable } from 'node:stream';
 import os from 'node:os';
-import aws from 'aws-sdk';
+import { Upload } from "@aws-sdk/lib-storage";
+import { S3, DeleteObjectCommand } from "@aws-sdk/client-s3";
```
mansaj, line 21: `DeleteObjectCommand` can be deleted now

Vinnl: Thanks for testing, deleted in e6c07ec67.

PR #3172, opened 2023-06-28
Fetch company icons from S3

Vinnl:
# References: 
Jira: MNTOR-1838

This does a HEAD request to every `.ico` file on S3 and, if it gets a successful response, adds them to the logo map. Unfortunately, every breach currently has an icon available because we currently also upload DDG's 404 icon to S3 if no icon was found, so we'll only be able to test the fallback once https://github.com/mozilla/blurts-server/pull/3171 has been run.


next.config.js
```diff
@@ -34,6 +34,11 @@ const nextConfig = {
         protocol: "https",
         hostname: "profile.stage.mozaws.net",
       },
+      // Stores breached companies' logos
```
flozia, line 37:
```suggestion
      // Stores breached companies logos
```

src/app/functions/server/getBreaches.ts
```diff
@@ -65,49 +65,21 @@ export async function getBreachIcons(breaches: Breach[]): Promise<LogoMap> {
...
     const logoMapElems = (await Promise.all(
-      breachDomains.map((breachDomain) => {
-        return new Promise((resolve, reject) => {
-          const logoFilename = breachDomain.toLowerCase() + ".ico";
-          const logoPath = pathResolve(logoFolder, logoFilename);
-          if (existingLogos.includes(logoFilename)) {
-            resolve([breachDomain, `/logo_cache/${logoFilename}`]);
-            return;
+      breachDomains.map(async (breachDomain) => {
+        const logoUrl = `https://s3.amazonaws.com/${
```
flozia, line 71: Question: Should we add `https://s3.amazonaws.com/` to an `.env` variable as well or do we not expect this to be changed?

flozia (approved):
Looks good!

next.config.js
Vinnl: I think this is correct? The logos are from the breached companies, so it's a possessive (alternative: "breached companies's logos").

src/app/functions/server/getBreaches.ts
Vinnl: In theory no, but I did expect us to use our own domain - I asked about that [here](https://mozilla.slack.com/archives/C04BLQVN151/p1687945004183639?thread_ts=1687809528.834059&cid=C04BLQVN151).

next.config.js
flozia: Oh, might be — sorry I though it just sneaked its way in.

mansaj (approved):
I'll look into the cloudfront cache

src/app/functions/server/getBreaches.ts
pdehaan: Don't think I have access to that Slack channel/thread, but looking at the Wayback Machine, it looks like our previous Monitor CDN was something like https://monitor.cdn.mozilla.net/img/logos/Abandonia.png

pdehaan:
Sorry, quick Q, but do we still need these requests to ddg?
```sh
git grep -A3 -B1 -n "duckduckgo" | cat

src/scripts/syncBreaches.js-77-      const res = await fetch(
src/scripts/syncBreaches.js:78:        `https://icons.duckduckgo.com/ip3/${breachDomain}.ico`);
src/scripts/syncBreaches.js-79-      if (res.status !== 200) {
src/scripts/syncBreaches.js-80-        return;
src/scripts/syncBreaches.js-81-      }
--
src/utils/hibp.js-206-        }
src/utils/hibp.js:207:        get(`https://icons.duckduckgo.com/ip3/${breachDomain}.ico`, (response) => {
src/utils/hibp.js-208-          if (response.statusCode !== 200) {
src/utils/hibp.js-209-            resolve(null)
src/utils/hibp.js-210-            return
```

Vinnl:
@pdehaan The ones in `syncBreaches.js`, yes, as that is the cron job that fetches them. The one in `hibp.js` no, but that's old Express code that's no longer used. (See MNTOR-1645 for the cleanup ticket, where I added them.)

PR #3191, opened 2023-07-05
MNTOR-1052 - wait for all promises to settle, not to resolve

rhelmer:
# References: 
Jira: MNTOR-1052

# Description

This cron job is occasionally hanging indefinitely on staging, preventing icons from being updated. If I stub out the S3 upload function locally, I can see that when icon fetching fails (which happens fairly often), the script hangs.

I believe that the problem is the use of `Promise.all()` vs. `Promise.allSettled()`. `all()` will stop at the first rejection, we *could* handle it, but what we really want is to keep iterating and trying the rest of the icons, so `allSettled()` is a better fit here anyway.

# Screenshot (if applicable)

Not applicable.

# How to test

I tested by running locally in a loop, putting it into service in stage and observing should help. We should set up monitoring for this job, maybe using [Sentry's new cron feature](https://docs.sentry.io/platforms/node/crons/), as a next step - I'll file a separate issue for that.


Vinnl (approved):
Good find, let's try this on stage ASAP :)

PR #3197, opened 2023-07-06
Do not break if window.gtag is not available

Vinnl:
This may happen if e.g. AdBlocker Ultimate is installed. Unfortunately, I couldn't reproduce, so I don't know how to test, but QA (Robert) confirmed that this was fixed when I deployed it to dev (and that it broke there before I deployed).


flozia (approved):
Would we need to add optional chaining to other gtag events as well? Like:
- `added_email`
- `resolved_breach_item`
- `opened_closed_dialog`
- `changed_email_preference`
- `removed_email`
- `opened_closed_user_menu`

flozia (approved):

PR #3199, opened 2023-07-07
Enable type-checked linting rules

Vinnl:
This was triggered by my wondering if [this](https://github.com/mozilla/blurts-server/pull/3157#discussion_r1255746579) could've been caught by ESLint. It can, if we allow it to make use of the type information that TypeScript can provide it :)

The main disadvantage is that our linting will get slower, but it will catch more potential bugs. I did turn off four of the recommended rules, since they were found all over the codebase, but the other recommended rules were relatively easy to fix, and mostly did seem to catch potential bugs or unnecessary Promise-wrapping. Let me know what you think.


.eslintrc.json
```diff
@@ -93,6 +93,31 @@
...
+        // > You must therefore provide a value for the "parserOptions.project"
+        // > property for @typescript-eslint/parser.
+        "project": "tsconfig.json"
+      },
+      "rules": {
+        // These rules trigger lots of times; possibly we can enable them later:
+        "@typescript-eslint/no-unsafe-argument": "off",
+        "@typescript-eslint/no-unsafe-assignment": "off",
+        "@typescript-eslint/no-unsafe-member-access": "off",
+        "@typescript-eslint/no-unsafe-call": "off"
```
maxxcrawford, line 119: praise: I'm all for easing into strict linting.

maxxcrawford (approved):

.eslintrc.json
```diff
@@ -93,6 +93,31 @@
         "import/no-named-as-default-member": "off",
         "import/no-unresolved": "off"
       }
+    },
+    {
+      // Only enable rules that depend on type checking on TS files,
+      // to avoid a barrage of warnings for older code:
+      "files": ["**/*.{ts,tsx}"],
+      "excludedFiles": ["src/app/(nextjs_migration)/**/*"],
```
pdehaan, line 101: Q: Why excluding this directory specifically?

src/app/api/v1/fxa-rp-events/route.ts
```diff
@@ -37,16 +37,16 @@ const getJwtPubKey = async () => {
...
       "getJwtPubKey",
       `fetched jwt public keys from: ${jwtKeyUri} - ${keys.length}`
     );
     return keys;
-  } catch (e) {
+  } catch (e: unknown) {
     console.error("getJwtPubKey", `Could not get JWT public key: ${jwtKeyUri}`);
     captureException(
-      new Error(`Could not get JWT public key: ${jwtKeyUri} - ${e}`)
+      new Error(`Could not get JWT public key: ${jwtKeyUri} - ${e as string}`)
```
pdehaan, line 49: What is `e as string` casting doing?
Is `e` an Error object here? Should this be `${e.message}` or does casting `as string` somehow capture the stack as well?

src/app/api/v1/premium/route.ts
```diff
@@ -16,7 +16,7 @@ export async function GET(req: NextRequest) {
     const _email = searchParams.get("email");
 
     return NextResponse.redirect(
-      `${process.env.SERVER_URL}/user/breaches`,
+      `${process.env.SERVER_URL ?? ""}/user/breaches`,
```
pdehaan, line 19: In what cases would `process.env.SERVER_URL` be missing/falsy?
I think we use it elsewhere without checking or using nullish coalescing for a fallback.

<details>
<summary><kbd>git grep -En "process\.env\.SERVER_URL"</kbd></summary>

```sh
src/app/(nextjs_migration)/(authenticated)/user/breaches/page.tsx:38:      url: process.env.SERVER_URL,
src/app/(nextjs_migration)/(guest)/breach-details/[breachName]/page.tsx:72:      url: process.env.SERVER_URL,
src/app/(nextjs_migration)/(guest)/breaches/page.tsx:26:      url: process.env.SERVER_URL,
src/app/api/v1/premium/route.ts:19:      `${process.env.SERVER_URL ?? ""}/user/breaches`,
src/app/api/v1/user/verify-email/route.ts:16:      (process.env.SERVER_URL ?? "") + "/user/settings",
src/app/layout.tsx:23:      typeof process.env.SERVER_URL === "string"
src/app/layout.tsx:24:        ? new URL(process.env.SERVER_URL)
src/app/layout.tsx:37:      url: process.env.SERVER_URL,
src/appConstants.js:69:if (!process.env.SERVER_URL && (process.env.NODE_ENV) === 'heroku') {
src/appConstants.js:70:  process.env.SERVER_URL = `https://${process.env.HEROKU_APP_NAME}.herokuapp.com`
```
</details>

.eslintrc.json
Vinnl: Heh, actually, I was going to say that there's lots of old copy-pasted code there from the pre-TS era and thus making it pass was too painful, but it was actually doable, so this exclude is now excluded: 68e51ef1e

src/app/api/v1/fxa-rp-events/route.ts
Vinnl: It's a TypeScript type assertion, which only works at build time to appease the TypeScript compiler - in other words, it doesn't change runtime behaviour. The linter is complaining about `e` being of type `any`, i.e. TS is unable to verify that it's actually loggable. Now, I don't know for sure whether it actually casts to a string at runtime, but given that the code already existed, I just assumed that it did.

src/app/api/v1/premium/route.ts
Vinnl: Heh, never, that env var is set in all our environments, but the lint rule doesn't want us to use potentially-undefined values in strings (which generally is a good rule-of-thumb).

PR #3209, opened 2023-07-10
Load the dashboard into an actual page

Vinnl:
There is still a bunch of hardcoded data, and it doesn't show data broker exposures yet, but at least it's a more realistic setting - it already exposed a couple of places where things didn't work perfectly (e.g. components with a transparent background, or data in a different shape than expected).

Note that unfortunately, I wasn't able to re-use the Storybook story in the unit test, because the import of `getL10n` would trigger an error because that imports `headers` from `next/headers`, which in turn triggers an error about not running on the server side. We could investigate that later, or possibly Next.js updates its Jest support to work better, or we could just not re-use the Storybook and its mock data.

## How to test

Visit `/redesign/user/dashboard` while logged in - you should see `<ExposureCard>`s listing your breaches, albeit all with the Twitter logo for now, and with every breach listed as "Fixed". Also, I've updated the Storybook to remove the mock "Shell" story (you can see what the Shell looks like when looking at the Dashboard or, later, the Settings), and the old non-[page](https://nextjs.org/docs/app/building-your-application/routing/pages-and-layouts) dashboard, and instead added [a story](https://deploy-preview-3209--fx-monitor-storybook.netlify.app/?path=/story/pages-dashboard--dashboard) that renders the actual page (with mock data).


Vinnl:
@codemist Could it be that the account you're logged in with does not have any verified emails with breaches? (Which brings another good observation to light: we need an empty state :) )

src/app/(proper_react)/redesign/(authenticated)/admin/feature-flags/page.tsx
```diff
@@ -60,7 +60,7 @@ export default async function FeatureFlagPage() {
     <div className={styles.wrapper}>
       <nav className={styles.tabBar}>
         <div className={styles.end}>
-          <Toolbar session={session} />
+          <Toolbar user={session} />
```
maxxcrawford, line 63: praise: code legibility

src/app/(proper_react)/redesign/(authenticated)/user/dashboard/View.tsx
```diff
@@ -0,0 +1,82 @@
...
+  locale: string;
+};
+
+export const View = (props: Props) => {
+  const l10n = useL10n();
+
+  return (
+    <div className={styles.wrapper}>
+      <Toolbar user={props.user}>
+        TODO:{" "}
```
maxxcrawford, line 28: question (non-blocking): What is this referencing? Should this be a comment?

src/app/(proper_react)/redesign/(authenticated)/user/dashboard/page.tsx
```diff
@@ -25,14 +26,12 @@ export default async function DashboardPage() {
...
-        TODO:{" "}
-        <a href="https://react-spectrum.adobe.com/react-aria/useTabList.html">
-          add a tab list
-        </a>
-      </Toolbar>
-    </div>
-  );
+  if (!session?.user) {
+    return redirect("/");
+  }
```
maxxcrawford, line 31: question: Did we have this logic somewhere else?! It seems super important. 😅

src/app/components/client/ExposureCard.tsx
```diff
@@ -67,10 +70,14 @@ export const ExposureCard = (props: ExposureCardProps) => {
...
     exposureDetailsLink,
     statusPillType,
+    locale,
   } = props;
 
   const l10n = useL10n();
   const [exposureCardExpanded, setExposureCardExpanded] = useState(false);
+  const dateFormatter = new Intl.DateTimeFormat(locale, {
+    dateStyle: "medium",
+  });
```
maxxcrawford, line 80: suggestion: Add a comment about what `dateStyle: "medium` means? (Example format, maybe?)

src/app/components/client/ExposureCard.module.scss
```diff
@@ -5,9 +5,9 @@
   --exposureDetailsIconWidth: 25px;
 
   font: $text-body-sm;
+  background-color: $color-white;
   border: 2px solid rgba($color-purple-70, 0.2);
   border-radius: $border-radius-md;
-  max-width: $content-xl;
```
maxxcrawford, line 10: question (non-blocking): Why this change?

maxxcrawford (commented):
Ahh...The rubber meeting the road. Thanks for knocking this out, @Vinnl 

My review is mostly questions! 

I had one larger follow-up: On Relay we had different mock data profiles. Is that something we have road-maps, etc? It'll be useful to be able to jump into specific user-types going forward.

src/app/(proper_react)/redesign/(authenticated)/user/dashboard/View.tsx
```diff
@@ -0,0 +1,82 @@
...
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+"use client";
+
+import { Session } from "next-auth";
+import styles from "./View.module.scss";
+import TwitterImage from "../../../../../components/client/assets/twitter-icon.png";
+import { Toolbar } from "../../../../../components/client/toolbar/Toolbar";
+import { DashboardTopBanner } from "../../../../DashboardTopBanner";
```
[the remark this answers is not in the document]
Vinnl: Yeah, makes sense to me! 2fdd29891

src/app/(proper_react)/redesign/(authenticated)/user/dashboard/View.tsx
```diff
@@ -0,0 +1,82 @@
...
+                    return (
+                      <li
+                        key={`${verifiedEmail.email}_${breach.Id.toString()}`}
+                        className={styles.exposureListItem}
+                      >
+                        <ExposureCard
+                          exposureImg={TwitterImage}
+                          exposureData={breach}
+                          exposureName={breach.Name}
+                          exposureDetailsLink={""}
```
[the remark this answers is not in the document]
Vinnl: That is a good question. On the current website, I'd expect that to link to e.g. https://monitor.firefox.com/breach-details/Twitter, but I'm not sure where it's going to go with Bedrock taking over those pages.

src/app/(proper_react)/redesign/(authenticated)/user/dashboard/View.module.scss
```diff
@@ -0,0 +1,38 @@
...
+    font: $text-body-md;
+  }
+}
+
+.exposureList {
+  list-style-type: none;
+  padding-inline: 0;
+  display: flex;
+  flex-direction: column;
+  gap: $spacing-lg;
```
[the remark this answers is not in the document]
Vinnl: Hmm, I based this on the designs having `24px` in between them, which is the exact value of `$spacing-lg`. Should I be looking elsewhere to determine this size?

image

src/app/(proper_react)/redesign/(authenticated)/user/dashboard/View.tsx
Vinnl: It's here at the top:

image

When I originally added that I put it there to make sure it didn't get missed when someone went to actually implement those tabs. It should be gone before we go live anyway.

src/app/(proper_react)/redesign/(authenticated)/user/dashboard/page.tsx
Vinnl: Oh yeah, I guess technically it's redundant with this: https://github.com/mozilla/blurts-server/blob/8c7dc84860ac780fca25cfe4aa2bfefc50f16f4a/src/app/(proper_react)/redesign/(authenticated)/layout.tsx#L13-L15

Although better safe than sorry I suppose, and now TS also knows for sure that `session.user` is defined.

src/app/components/client/ExposureCard.tsx
Vinnl: Added a link to the docs in 51c9261f0, though technically how it gets formatted depends on `locale` _and_ the user's browser. But for `en`, it's basically short-but-not-ambiguous, in the format "Jul 10, 2023".

src/app/components/client/ExposureCard.module.scss
Vinnl: Otherwise there was empty space to the right of the cards, which I didn't see in the designs (nor anything to indicate that they should have a max width). So

image

vs

image

Vinnl:
@maxxcrawford 

> I had one larger follow-up: On Relay we had different mock data profiles. Is that something we have road-maps, etc? It'll be useful to be able to jump into specific user-types going forward.

Not explicitly on the road map anywhere, but I think we'll probably set those up as we go and as-needed for testing. @rhelmer has already set up some infrastructure for dynamically-generated mock data in https://github.com/mozilla/blurts-server/pull/3166.

We won't be able to do it the same way as we do in Relay, since we fetch data on the server side, but in my mind this is a perfect use case for Storybook.

PR #3210, opened 2023-07-10
Unblock Storybook-in-tests with l10n

Vinnl:
Accessing next/headers in tests results in an error because it's not in a Server Component context, so calling our regular `getL10n()` functions from inside Storybook would result in test failures. Instead, I replaced that call with a function that just loads the English strings directly, rather than looking at request headers.

Note that there are also a couple of test changes - now that proper English strings are loaded, some of the tests would fail without accommodating that :)

I've set https://github.com/mozilla/blurts-server/pull/3209's branch as the target, because this also resolves what I mentioned there:

> Note that unfortunately, I wasn't able to re-use the Storybook story in the unit test

### DIFF
--- a/.storybook/preview.tsx
+++ b/.storybook/preview.tsx
@@ -9,28 +9,14 @@ import "../src/app/globals.css";
 import { L10nProvider } from "../src/contextProviders/localization";
 import { metropolis } from "../src/app/fonts/Metropolis/metropolis";
 import { ReactAriaI18nProvider } from "../src/contextProviders/react-aria";
+import { getEnL10nBundlesSync } from "../src/app/functions/server/mockL10n";
 
 const inter = Inter({ subsets: ["latin"], variable: "--font-inter" });
-
-function loadL10nModule() {
-  if (process.env.NODE_ENV === "test") {
-    // In Jest, loading this module throws errors because it uses
-    // require.context(). Since we don't need to look at actual strings in our
-    // unit tests, we can just skip loading the bundles.
-    return {
-      getL10nBundles: () => [],
-      getLocale: () => "en",
-    };
-  }
-
-  return require("../src/app/functions/server/l10n");
-}
 
 const AppDecorator: Exclude<Preview["decorators"], undefined>[0] = (
   storyFn
 ) => {
-  const { getL10nBundles, getLocale } = loadL10nModule();
-  const l10nBundles = getL10nBundles();
+  const l10nBundles = getEnL10nBundlesSync();
 
   useEffect(() => {
     // We have to add these classes to the body, rather than simply wrapping the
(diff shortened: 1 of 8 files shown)


locales-pending/premium.ftl
```diff
@@ -13,6 +13,8 @@ main-nav-link-dashboard-label = Dashboard
 main-nav-link-faq-label = FAQs
 main-nav-link-faq-tooltip = Frequently asked questions
 
+mobile-menu-label = Main menu
```
Vinnl, line 16: I added this because the a11y test now caught that the mobile menu did not have a unique label.

src/app/(proper_react)/(onboarding)/user/welcome/Onboarding.stories.tsx
```diff
@@ -14,5 +14,5 @@ export default meta;
 type Story = StoryObj<typeof OnboardingEl>;
 
 export const Onboarding: Story = {
-  render: () => <OnboardingEl user={{}} />,
+  render: () => <OnboardingEl user={{ email: "example@example.com" }} />,
```
Vinnl, line 17: Sorry, this is an unrelated change I snuck in - I saw a red squiggly that this mock data did not include an email address.

src/app/functions/server/mockL10n.ts
Vinnl, line 1: This is a bit of a complex file, but basically it's a copy of the `l10n.ts` file in the same dir, but modified to always load the `en` locales without checking request headers (which are not available in tests), and to use `require.context` for Storybook (which uses Webpack, which provides that function) and `node:fs` in Jest (which runs in a Node context).

.storybook/preview.tsx
```diff
@@ -44,9 +30,7 @@ const AppDecorator: Exclude<Preview["decorators"], undefined>[0] = (
 
   return (
     <L10nProvider bundleSources={l10nBundles}>
-      <ReactAriaI18nProvider locale={getLocale(l10nBundles)}>
-        {storyFn()}
-      </ReactAriaI18nProvider>
+      <ReactAriaI18nProvider locale={"en"}>{storyFn()}</ReactAriaI18nProvider>
```
pdehaan, line 33: Not sure it needs the curly braces in `locale={"en"}`:
```suggestion
      <ReactAriaI18nProvider locale="en">{storyFn()}</ReactAriaI18nProvider>
```

src/app/(proper_react)/(onboarding)/user/welcome/Onboarding.test.tsx
```diff
@@ -33,7 +33,7 @@ it("can open a dialog with more information", async () => {
   expect(screen.queryByRole("dialog")).not.toBeInTheDocument();
 
   const explainerTrigger = screen.getByRole("button", {
-    name: "onboarding-get-started-content-explainer",
+    name: "See how we protect your data",
```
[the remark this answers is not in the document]
Vinnl: Yeah, it would. (That was already the case, though, as string changes would also require a new l10n ID.) [The idea](https://testing-library.com/docs/queries/about#priority) is that we simulate what the user would do, so that e.g. if a "Sign up" button were to suddenly say "Sign in", the tests break because the user's expected flow would break.

It's not that much work to fix since the test runner should clearly indicate which tests fail, and strings don't change that often, so IMHO that's a reasonable trade-off. (But: definitely something we could discuss in the tech meeting if you're interested.)

.storybook/preview.tsx
Vinnl: Fairly sure it does not: 95f9e8acf.

PR #3214, opened 2023-07-12
Fix: FxA profile image on prod

flozia:
# Description

Fixes the users FxA profile image.

# Screenshot

| Stage | Prod |
| --- | --- |
| fxa-profile-image-stage | fxa-profile-image-prod |


rhelmer (approved):
We should figure out how to substitute a variable here, but fixing prod now sounds good thanks!

mansaj (approved):

PR #3219, opened 2023-07-14
feature: add onerep webhook handler for completed scan

rhelmer:
# References: 
Jira: MNTOR-1403

# Description

Add webhook handler for OneRep scan results.

# How to test

1. use curl, or
2. use a service like ngrok and register a webhook from sandbox env https://docs.onerep.com/#tag/Webhook-Endpoints, or
3. wait for this to be live on dev server :) 


.env-dist
```diff
@@ -72,7 +72,9 @@ HIBP_NOTIFY_TOKEN=unsafe-default-token-for-dev
 HIBP_BREACH_DOMAIN_BLOCKLIST=a-blocked-domain.com,another-blocked-domain.org
 
 # OneRep API for exposure scanning
+ONEREP_API_BASE=https://api.onerep.com
```
pdehaan, line 75: Not sure this is being used.

```sh
git grep -n "ONEREP_API_BASE" | cat
.env-dist:75:ONEREP_API_BASE=https://api.onerep.com

git grep -n "https://api.onerep.com" | cat
.env-dist:75:ONEREP_API_BASE=https://api.onerep.com
src/app/functions/server/onerep.ts:90:  const url = "https://api.onerep.com" + path;
```

src/app/functions/server/onerep.ts
```diff
@@ -296,3 +296,27 @@ export async function getScanDetails(
...
+        per_page: 100,
+      });
+      currentPage++;
+      nextPage.data.forEach((element: object) => scanListFull.push(element));
+    }
+  } else {
+    scanListFull.push(firstPage.data);
+  }
+
+  return scanListFull[0];
```
pdehaan, line 321: what does this `[0]` do?
Looks like we're looping while `currentPage <= firstPage.meta.last_page` and pushing results into `scanListFill[]` Array. But then we might only be returning the first 100 paged results anyways. Should this be `return scanListFull;` or `return scanListFull.flat();`?

.env-dist
rhelmer: Thanks... I think we should, I'll modify `onerep.ts`. I was wondering why this wasn't set

src/app/functions/server/onerep.ts
rhelmer: I just copy/pasted this from the code I landed... I was hoping the webhook had the payload, but looks like it just lets us know the scan is ready and we still need to do the multi-page thing.

I'll go over this code again since it looks like we need to keep it, was hoping it'd only be for local testing.

rhelmer: Hm yeah I need to test this (I guess I can reduce the `per_page` setting to do this), but I think you're right - we want to flatten out the array, not just pick the first one.

src/app/functions/server/onerep.ts
```diff
@@ -296,3 +300,27 @@ export async function getScanDetails(
...
+  const scanPagesAll = [];
+  const firstPage = await listScanResults(profileId, {
+    per_page: 100,
+  });
+  // Results are paginated, use per_page maximum and collect all pages into one result.
+  if (firstPage.meta.last_page > 1) {
+    let currentPage = 2;
+    while (currentPage <= firstPage.meta.last_page) {
+      const nextPage = await listScanResults(profileId, {
+        per_page: 100,
```
pdehaan, line 316: Q: Does _this_ version need to pass `page: currentPage,` to track the current page number?

rhelmer: Ugh yes thanks

rhelmer:
I also tested this on Heroku and it seems to work well, I've registered a new webhook handler (which is now temporarily disabled until we merge this) and I've deleted the handler I was using for local development.

I'm not sure how we're going to use webhooks for multiple environments (dev and stage) though, will investigate. Until we have a resolution, I suggest we register both and only have one enabled at a time - once this feature graduates to stage we can disable it on dev. We may want to fall back to the polling approach that we've preserved for local dev on Heroku as well, if there isn't a good way to differentiate the environments.

.env-dist
```diff
@@ -72,7 +72,9 @@ HIBP_NOTIFY_TOKEN=unsafe-default-token-for-dev
...
 ONEREP_API_KEY=
+ONEREP_WEBHOOK_SECRET=
```
mansaj, line 77: Is this different from ONEREP_API_KEY?

src/app/api/v1/onerep-events/route.ts
```diff
@@ -0,0 +1,90 @@
...
+    captureException(ex);
+
+    return NextResponse.json({ success: false }, { status: 401 });
+  }
+
+  try {
+    const result: OnerepWebhookRequest = JSON.parse(finalBuffer.toString());
+
+    if (result.type !== "scan.completed") {
+      throw new Error("Unexpected OneRep webhook type received");
```
mansaj, line 67: We should be handling all kinds of requests here? ie. `scan_result.*` are also events that can be expected?

src/app/api/v1/onerep-events/route.ts
```diff
@@ -0,0 +1,90 @@
...
+  } catch (ex) {
+    console.error(ex);
+    captureException(ex);
+
+    return NextResponse.json({ success: false }, { status: 401 });
+  }
+
+  try {
+    const result: OnerepWebhookRequest = JSON.parse(finalBuffer.toString());
+
```
mansaj, line 65: Can we add a debug log here for the result? It would be useful for heroku/stage debugging

.env-dist
mansaj: Ah yes, that makes sense. We need to register all 3 endpoints, ask SRE to add the env vars for stage/prod, and record the dev secrets in our shared space (also add the env var to heroku)

src/app/api/v1/onerep-events/route.ts
mansaj: I wonder if we should fail the other result types silently (log but not throw an error)

rhelmer: Yeah good idea

mansaj (approved):

PR #3222, opened 2023-07-17
MNTOR-1889: Add “Find exposures” step to onboarding and wire up backend

flozia:
# References: 
Jira: [MNTOR-1889](https://mozilla-hub.atlassian.net/browse/MNTOR-1889)
Figma: https://github.com/mozilla/blurts-server/compare/MNTOR-1889-Wireup-enter-info-backend?expand=1

# Description

This PR adds the “Find exposure” step to the onboarding flow.

# How to test

1. If you already [created `locationAutocompleteData.json` locally](https://github.com/mozilla/blurts-server#prerequisites) you can skip this step:
```
npm run create-location-data
```

2. Add the new OneRep variables in your `.env`:
```
ONEREP_API_KEY=[add-sandbox-key-here]
ONEREP_API_BASE=https://api.onerep.com
# Can be empty for testing this PR
ONEREP_WEBHOOK_SECRET=
ONEREP_DATA_BROKER_COUNT=190
```

4. If you already have run a scan you can delete the entries from your local database:
```
delete from onerep_scans; delete from onerep_profiles; update subscribers set onerep_profile_id = null;
```

5. In case you have not run the migration script locally for a while you can do so with:
```
npm run db:migrate
```

6. Visit `/redesign/user/welcome` and go through the welcome onboarding.

PR #3233, opened 2023-07-19
MNTOR-1675: Welcome scan progress label

flozia:
# References: 
Jira: [MNTOR-1675](https://mozilla-hub.atlassian.net/browse/MNTOR-1675)
Figma: https://www.figma.com/file/CaEKIhvSJqf6KNIMzSkt40/Concepts-for-Monitor-MVP-Redesign?type=design&node-id=1398-73778&t=LMpwec8NLhVLuzS7-4

# Description

This add the correct labels to the progress bar in the “find exposures” welcome step.

# Screenshot

https://github.com/mozilla/blurts-server/assets/13835474/ae118c7b-db6f-491f-bab1-9fc698708943

# How to test

Same steps as for https://github.com/mozilla/blurts-server/pull/3222


src/db/tables/breaches.js
```diff
@@ -18,6 +18,20 @@ async function getAllBreaches() {
...
 }
 
+/**
+ * Get all count from "breaches" table
+ *
+ * @returns Count of all records from "breaches" table
+ */
+async function getAllBreachesCount() {
+  const breachesCount = await knex('breaches')
+    .count({ count: "*" })
```
rhelmer, line 28: This table is not very large, but if it became a problem we could consider doing an index-only scan (e.g. if the index is likely to be significantly smaller than the table).

src/db/tables/breaches.js
```diff
@@ -18,6 +18,20 @@ async function getAllBreaches() {
...
+    .first()
+    .then(result => result?.count)
+  // @ts-ignore The above query is returning a string
+  return parseInt(breachesCount, 10)
```
rhelmer, line 32:
```suggestion
  return parseInt(breachesCount?.toString() || "", 10)
```

rhelmer (approved):
I think we could avoid the `@ts-ignore`, this should return `NaN` if the empty string is passed to `parseInt()`

src/db/tables/breaches.js
flozia: Good suggestion! Made the change in [bd3a6df](https://github.com/mozilla/blurts-server/pull/3233/commits/bd3a6df8c2ed1e6c2e59780192ad8b0f2fc822d6).

src/db/tables/breaches.js
flozia: Thanks — changed in [bd3a6df](https://github.com/mozilla/blurts-server/pull/3233/commits/bd3a6df8c2ed1e6c2e59780192ad8b0f2fc822d6).

PR #3235, opened 2023-07-20
fix: Dashboard route protection

flozia:
# Description

Fixes the dashboard route protection and redirect url.

### DIFF
--- a/src/app/(proper_react)/redesign/(authenticated)/user/dashboard/page.tsx
+++ b/src/app/(proper_react)/redesign/(authenticated)/user/dashboard/page.tsx
@@ -3,10 +3,7 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
 import { getServerSession } from "next-auth";
-import {
-  canSubscribeToPremium,
-  hasSetupOnerep,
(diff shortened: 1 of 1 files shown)


src/app/(proper_react)/redesign/(authenticated)/user/dashboard/page.tsx
```diff
@@ -20,27 +17,23 @@ import { getLatestOnerepScan } from "../../../../../../db/tables/onerep_scans";
...
   const countryCode = getCountryCode(headersList);
 
   const result = await getOnerepProfileId(session.user.subscriber.id);
   const profileId = result[0]["onerep_profile_id"] as number;
-  const scanResult = await getLatestOnerepScan(profileId);
-  const scanResultItems = scanResult?.onerep_scan_results?.data ?? [];
-
   if (
-    !hasSetupOnerep(session?.user) &&
+    !profileId &&
```
Vinnl, line 29: @flozia Isn't this the exact same logic as in `hasSetupOneReq`, but now no longer in a function name that describes why it does this check? (And also theoretically failing if the profile ID is 0, though I don't think that's possible.)

PR #3237, opened 2023-07-21
Implement feature flag UI

rhelmer:
# Description

Basic admin page for feature flags.

I'm not super happy with the user experience, but it does work - currently it adds a new flag when you click the "Add flag" button, and saves form fields and the enabled checkbox for existing flags when focus moves away from them (blur).

It doesn't allow deleting flags right now, just adding them and disabling/enabling them, and modifying all fields.

# Screenshot (if applicable)

Not applicable.

# How to test

- `npm run dev` and load http://localhost:6060/redesign/admin/feature-flags
- add new flag(s)
- modify fields, they are saved as soon as focus changes from the element


rhelmer:
@pdehaan BTW this is in draft mode, so I can get feedback about what I'm doing wrong from front-end folks :) I know it needs tests and better input validation too.

src/app/functions/server/onerep.ts
```diff
@@ -265,6 +266,16 @@ export async function isEligible() {
...
+  const flag = await getFlag("FreeBrokerScan");
+
+  if (!flag.isEnabled) {
+    throw new Error("Flag is not enabled");
+  }
+
+  if (!flag.allowList?.includes(session.user.email)) {
+    throw new Error("User is not on allow list for feature");
+  }
+
```
rhelmer, line 278: This is an example of how to gate a feature on a flag, `getFlag()` server function returns a `FeatureFlag` type.

If we need to do feature flags on the client side, we could expose a client function too (feature flags are exposed via an API it could fetch).

rhelmer: If we expected _n_ (such as `allowList` here) to be very large (thousands or millions of entries), using `Set` instead of an array here would be advantageous. It might be faster still to build a SQL query that lets the database look it up and just return the answer vs. pulling all the data and doing it in TS (but we'd probably want to have a few tables with indexes instead of a column with array types as the schema is now).

I don't think it's likely we'll see those kind of numbers to warrant the more complex schema at this time. Using `Set` might be both more efficient for lookups and just as convenient though so we should consider using it here.

pdehaan:
> _@pdehaan BTW this is in draft mode, so I can get feedback about what I'm doing wrong from front-end folks_

I feel both simultaneously "seen" and "attacked". 😍

src/app/(proper_react)/redesign/(authenticated)/admin/feature-flags/AddFeatureFlag.tsx
```diff
@@ -0,0 +1,78 @@
...
+
+    const options = {
+      method: "POST",
+      headers: {
+        "Content-Type": "application/json",
+      },
+      body: JSON.stringify(data),
+    };
+
+    // Send the form data to our forms API on Vercel and get a response.
```
pdehaan, line 31: Wait, are we using Vercel? I thought we were using GCP and Netlify.

src/app/(proper_react)/redesign/(authenticated)/admin/feature-flags/ModifyInputField.tsx
```diff
@@ -0,0 +1,40 @@
...
+export const ModifyInputField = (props: {
+  id: string | undefined;
+  name: string | undefined;
+  defaultValue: string | undefined;
+}) => {
+  // @ts-ignore FIXME determine correct event type
+  const handleChange = void (async (event) => {
+    const id = event.target.id;
+    const name = event.target.name;
+    const value = event.target.value;
```
pdehaan, line 16: This code looks similar to the previous file, except the previous file had a bunch of optional chaining `?.` stuffs for `event` and `target`:

```js
    const data = {
      name: event?.target?.name.value,
      …
```

src/app/(proper_react)/redesign/(authenticated)/admin/feature-flags/ModifyInputField.tsx
```diff
@@ -0,0 +1,40 @@
...
+
+    const endpoint = `/api/v1/admin/feature-flags/${name as string}`;
```
pdehaan, line 18: Possibly just a TypeScript-ism that I'm not familiar with yet, but should this be:
```diff
-const endpoint = `/api/v1/admin/feature-flags/${name as string}`;
+const endpoint = `/api/v1/admin/feature-flags/${name}` as string;
```

Considering earlier we were casting the entire `const` `as string`:
https://github.com/mozilla/blurts-server/blob/ad093cf82259520737ea19e3598c83a1a777c933/src/app/(proper_react)/redesign/(authenticated)/admin/feature-flags/AddFeatureFlag.tsx#L21

src/app/(proper_react)/redesign/(authenticated)/admin/feature-flags/page.tsx
```diff
@@ -37,16 +40,50 @@ export default async function FeatureFlagPage() {
...
+              <td>
+                <ToggleFlagEnabled
+                  id="isEnabled"
+                  name={item.name}
+                  isEnabled={item.is_enabled}
+                />
+                {item.is_enabled}
+              </td>
+              <td>
+                {" "}
```
pdehaan, line 59: Q: Why does this need a leading space?

src/app/(proper_react)/redesign/(authenticated)/admin/feature-flags/page.tsx
```diff
@@ -37,16 +40,50 @@ export default async function FeatureFlagPage() {
...
             <tr key={item.name}>
               <td>{item.name}</td>
-              <td>{String(item.is_enabled)}</td>
-              <td>{item.dependencies}</td>
-              <td>{item.allow_list?.join(", ")}</td>
-              <td>{item.wait_list?.join(", ")}</td>
+              <td>
+                <ToggleFlagEnabled
+                  id="isEnabled"
+                  name={item.name}
```
pdehaan, line 53: Q: Does this work? It seems from the code below that _everything_ will have a `name` of `item.name`:
- `id="isEnabled` (L52)
- `id="dependencies"` (L62)
- `id="allowList"` (L69)
- `id="waitList"` (L76)
- `id="owner"` (L83)

src/app/api/v1/admin/feature-flags/[flagId]/route.ts
```diff
@@ -28,3 +35,41 @@ export async function GET(
...
+
+      return NextResponse.json({ success: true }, { status: 200 });
+    } catch (e) {
+      console.error(e);
+      return NextResponse.json({ success: false }, { status: 500 });
+    }
+  } else {
+    // Not Signed in, redirect to home
+    return NextResponse.redirect(appConstants.SERVER_URL, 301);
+  }
```
Comment (pdehaan) [inline src/app/api/v1/admin/feature-flags/[flagId]/route.ts:74]: tiniest of nits, and not sure if we have a style guide, but would it be easier to invert the `if..else` and just abort early if the user is not auth'ed and reduce a level of nesting?

```js
  if (!isAdmin(session?.user?.email || "")) {
    return NextResponse.redirect(appConstants.SERVER_URL, 301);
  }

  // Signed in
  try {
    const flagName = req.nextUrl.pathname.split("/").at(-1);
    ...
```

src/app/functions/server/onerep.ts
```diff
@@ -265,6 +266,16 @@ export async function isEligible() {
     throw new Error("No session");
   }
 
+  const flag = await getFlag("FreeBrokerScan");
+
+  if (!flag.isEnabled) {
+    throw new Error("Flag is not enabled");
+  }
+
+  if (!flag.allowList?.includes(session.user.email)) {
```
pdehaan, line 275: Not sure if it matters, but I have cursed eyes. Elsewhere in these files we're doing:
```js
  if (!session?.user?.email) {
    return redirect("/");
  }
```

Does this need the additional optional chaining on `session?.user?.email`? Or maybe not if we're aborting earlier on L265-267 (although that's checking `session.user.subscriber.id`, but one might infer the other):
```js
  if (!session?.user?.subscriber?.id) {
    throw new Error("No session");
  }
```

pdehaan:
Q: Is https://github.com/mozilla/blurts-server/blob/main/docs/adr/0005-feature-flags.md ADR still in "proposed" state, or do we need to update that to "accepted" (or whatever you use in this project) and document up the accepted solution?
I did see this somewhere near the middle of the doc:
> _"A team decision has been made on Thursday, May 25. We will go with the `environment variable` approach similar to Relay."_

src/app/(proper_react)/redesign/(authenticated)/admin/feature-flags/AddFeatureFlag.tsx
```diff
@@ -0,0 +1,78 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+"use client";
+
+export const AddFeatureFlag = () => {
+  // @ts-ignore FIXME determine correct event type
+  const handleSubmit = void (async (event) => {
```
rhelmer, line 9: Hm, `next lint` accepted this but it's actually wrong and the handler isn't being called so the UI isn't working... I'll figure this out.

src/app/(proper_react)/redesign/(authenticated)/admin/feature-flags/page.tsx
rhelmer: This is something VSCode seems to autocomplete, it's really annoying. I'll remove.

src/app/(proper_react)/redesign/(authenticated)/admin/feature-flags/page.tsx
rhelmer: `item.name` should be the name of the flag, I believe.

src/app/api/v1/admin/feature-flags/[flagId]/route.ts
Comment (rhelmer) [reply src/app/api/v1/admin/feature-flags/[flagId]/route.ts:74]: I think "whatever `prettier` does, unless we override it" is our de factor style guide :P But yes I think you're right that earlier return is a better pattern, thanks.

rhelmer:
> Q: Is https://github.com/mozilla/blurts-server/blob/main/docs/adr/0005-feature-flags.md ADR still in "proposed" state, or do we need to update that to "accepted" (or whatever you use in this project) and document up the accepted solution? I did see this somewhere near the middle of the doc:
> 
> > _"A team decision has been made on Thursday, May 25. We will go with the `environment variable` approach similar to Relay."_

Yes this was the plan at the time, but for this and upcoming features we realized we were going to need to have more complex flags than just a simple boolean on/off switch (such as allow lists, and flags which depend on other flags).

The backend functionality for this more complex type of flag has been implemented, but there is no UI and I think asking SRE to modify the database every time we want to modify an allowlist or flip a feature on/off is pretty onerous for them. I'll see if we can amend the ADR, good call-out thanks.

src/app/(proper_react)/redesign/(authenticated)/admin/feature-flags/page.tsx
pdehaan: Yes, but isn't it giving every single `<input>` the same `name=` value then (despite having different `id=` values)?
Maybe I'm wrong, it just looks a bit suspicious.

src/app/api/v1/admin/feature-flags/[flagId]/route.ts
Comment (pdehaan) [reply src/app/api/v1/admin/feature-flags/[flagId]/route.ts:74]: Only because I mentioned Next.js middleware and started googling: https://nextjs.org/docs/app/building-your-application/routing/middleware

Not sure if that's something we'd want to poke at (although not in this PR, just... in general). Basically create a middleware that looks for the `/(authenticated)/` route group [if that's possible` and auto-redirect to the homepage if they're not logged in so we don't need to do that in _n_ places.

src/app/(proper_react)/redesign/(authenticated)/admin/feature-flags/AddFeatureFlag.tsx
rhelmer: No, sorry this is just from a tutorial, I'll remove it. We're using GCP and k8s :)

src/app/(proper_react)/redesign/(authenticated)/admin/feature-flags/ModifyInputField.tsx
rhelmer: Yeah I think the better approach is going to be defining the type here and using a type guard to catch problems at run time, I'm trying to figure out if I can re-use the existing `FeatureFlag` type since it's really close.

rhelmer: (`event.target` is basically a string-ified version of `FeatureFlag` with `.value` on each leaf node, I'm not sure if there's an elegant way to declare that for the typescript compiler)

src/app/api/v1/admin/feature-flags/[flagId]/route.ts
Comment (rhelmer) [reply src/app/api/v1/admin/feature-flags/[flagId]/route.ts:74]: Yes I had the same question... I'd like to be able to be able to just trust the authenticated route to do the right thing and not have redundant checks everywhere.

Comment (rhelmer) [reply src/app/api/v1/admin/feature-flags/[flagId]/route.ts:74]: One problem is that we keep the APIs in `./src/app/api` and each endpoint implementation does it's own checking, I wonder if we could simplify there as well.

Comment (rhelmer) [reply src/app/api/v1/admin/feature-flags/[flagId]/route.ts:74]: Oh, and another thing - we need to check not only that the user is authenticated but *also* that they are using an admin account, I wonder if this is something we could do using middleware as well (and how we could test to make sure things aren't misconfigured and leaving the route open to normal users etc)

rhelmer:
OK I figured out how to refresh server-side components without a full page refresh, so it's much nicer to use now - however I think the way I am saving changes when the focus leaves each element feels clumsy, I think it might be better to use a button to save each row, or a single button to submit the whole form (client-side without a page reload).

I think a toast element to confirm that changes were saved (or that an error occurred) would be really good as well.

src/app/api/v1/admin/feature-flags/[flagId]/route.ts
Comment (pdehaan) [reply src/app/api/v1/admin/feature-flags/[flagId]/route.ts:74]: Yeah, if we can do middleware for everything in `(authenticated)` maybe we create another route group for `(admin)` which adds bonus middleware for being both logged in AND part of some blessed admin group.

I believe FxA recently added a bunch of extended role based permissions for admin actions last year: https://mozilla.github.io/ecosystem-platform/reference/admin-panel#features-guarded-by-permissions 

I also think their https://github.com/mozilla/fxa/blame/main/packages/fxa-admin-panel/src/components/Guard/index.tsx system is protected by being both (a) behind VPN, and (b) accessible to only certain LDAP groups. But maybe that's right overkill for experiments if we only want to use some binary "isAdmin" flag in the database and not more granular roles like "can create feature flags" and "can delete feature flags" roles.

src/app/(proper_react)/redesign/(authenticated)/admin/feature-flags/components/AddFeatureFlag.tsx
```diff
@@ -33,23 +33,26 @@ export const AddFeatureFlag = () => {
...
+
     const data = {
-      name: event.target.name.value,
-      isEnabled: event.target.isEnabled ? true : false,
+      name: eventTarget.name.value,
+      isEnabled: eventTarget.isEnabled.checked ? true : false,
       dependencies: event.target.dependencies.value.split(","),
       allowList: event.target.allowList.value.split(","),
       waitList: event.target.waitList.value.split(","),
       owner: event.target.owner.value,
```
pdehaan: Q: Should these be updated to use `eventTarget.` instead of `event.target.` too?
```suggestion
      dependencies: eventTarget.dependencies.value.split(","),
      allowList: eventTarget.allowList.value.split(","),
      waitList: eventTarget.waitList.value.split(","),
      owner: eventTarget.owner.value,
```

src/app/(proper_react)/redesign/(authenticated)/admin/feature-flags/components/ModifyInputField.tsx
```diff
@@ -12,13 +25,16 @@ export const ModifyInputField = (props: {
...
+    event: FocusEventHandler<HTMLFormElement> & {
+      target: FeatureFieldElement;
+    }
+  ) => {
+    const id = event.target.id.value;
+    const name = event.target.name.value;
+    const value = event.target.value.value;
 
-    const endpoint = `/api/v1/admin/feature-flags/${name as string}`;
+    const endpoint = `/api/v1/admin/feature-flags/${name }`;
```
pdehaan, line 37:
```suggestion
    const endpoint = `/api/v1/admin/feature-flags/${name}`;
```

src/app/(proper_react)/redesign/(authenticated)/admin/feature-flags/components/ToggleFlagEnabled.tsx
```diff
@@ -12,13 +19,18 @@ export const ToggleFlagEnabled = (props: {
...
+    }
+  ) => {
+    const eventTarget: FeatureToggleElement = event.target;
+
+    const id = eventTarget.id;
+    const name = eventTarget.name;
+    const isEnabled = eventTarget.checked ? true : false;
 
-    const endpoint = `/api/v1/admin/feature-flags/${name as string}`;
+    const endpoint = `/api/v1/admin/feature-flags/${name }`;
```
pdehaan, line 33:
```suggestion
    const endpoint = `/api/v1/admin/feature-flags/${name}`;
```

src/app/(proper_react)/redesign/(authenticated)/admin/feature-flags/page.tsx
rhelmer: The way it's intended to work is that each `<input>` element has an `id` to identify what it is, and each row of elements has a `name` indicating the name of the flag that it modifies.

src/app/functions/server/onerep.ts
mansaj: I agree, we really shouldn't have more than a few tens or even hundreds of people on the allow list. Anything more than would probably require more sophisticated admin controls (groups, different policies etc)

src/app/api/v1/admin/feature-flags/[flagId]/route.ts
Comment (mansaj) [reply src/app/api/v1/admin/feature-flags/[flagId]/route.ts:74]: I looked at a couple of tutorials for auth middleware, couldn't get any of them to work in our context. I agree it's definitely something we should look into, we definitely shouldn't have these if/else everywhere for session user checking

src/app/api/v1/admin/feature-flags/[flagId]/route.ts
```diff
@@ -28,3 +35,41 @@ export async function GET(
     return NextResponse.redirect(appConstants.SERVER_URL, 301);
   }
 }
+
+export async function POST(req: NextRequest) {
```
Comment (mansaj) [inline src/app/api/v1/admin/feature-flags/[flagId]/route.ts:39]: I think this should probably be a PUT

src/app/(proper_react)/redesign/(authenticated)/admin/feature-flags/components/AddFeatureFlag.tsx
```diff
@@ -0,0 +1,114 @@
...
+
+    const data = {
+      name: eventTarget.name.value,
+      isEnabled: eventTarget.isEnabled.checked ? true : false,
+      dependencies: eventTarget.dependencies.value.split(","),
+      allowList: eventTarget.allowList.value.split(","),
+      waitList: eventTarget.waitList.value.split(","),
+      owner: eventTarget.owner.value,
+    };
+
```
mansaj, line 52: Might be nice to add a empty check for `data.name`. I accidentally clicked the add button too many times and ended up with a few empty rows

mansaj (approved):
Works fine, LGTM pending fixes and comments

src/app/(proper_react)/redesign/(authenticated)/admin/feature-flags/components/AddFeatureFlag.tsx
rhelmer: Yeah same :(

src/app/(proper_react)/redesign/(authenticated)/admin/feature-flags/ModifyInputField.tsx
rhelmer: OK I declared in the interface that these fields can be null more consistently, and put really basic checking in place (I think we could do a more abstract type guard but I'd like to coordinate with front-end folks who know this stuff better)

PR #3238, opened 2023-07-21
Add locales-pending folder to CODEOWNERS for l10n

flodolo:
Given the amount of content landing in `locales-pending`, it's better to have review early than to slow things down later, if we decide to expose it for localization.


flodolo:
Replaced by #3239

PR #3246, opened 2023-07-24
L10 hook hydration error

flozia:
# Description

This PR fixes the l10 hook hydration errors, adds some missing strings, and addresses some warnings.

# How to test

Visit the new dashboard `/redesign/user/dashboard` and page where we utilize the `l10n` hook and make sure you are not getting any hydration errors.


locales-pending/premium.ftl
```diff
@@ -41,6 +41,11 @@ exposure-chart-legend-value-nr = { $nr }x
 
 # Here's What We Fixed Progress Card
 
+exposure-card-company-logo = Company logo
+exposure-card-company = Company domain
+exposure-card-exposure-type = Exposure types
+exposure-card-date-found = Date of the exposure
```
flozia, line 47: @codemist These strings were missing in the `.ftl` files and I’ve seen a warning about it in the console. Are these strings correct or could you point me to the content?

flodolo (commented):
Wondering what's _hydration_ means in this context.

flodolo (approved):

src/app/hooks/l10n.ts
```diff
@@ -21,17 +21,25 @@ export type GetFragment = (
...
-    l10n.getElement(createElement(Fragment, null, fallback ?? id), id, args);
+  useEffect(() => {
+    setIsMounted(true);
+  }, [isMounted]);
+
+  const getFragment: GetFragment = (id, args, fallback) => {
+    return isMounted
+      ? l10n.getElement(createElement(Fragment, null, fallback ?? id), id, args)
+      : null;
+  };
```
flozia, line 42: > Wondering what's _hydration_ means in this context.

@flodolo The [hydration errors](https://nextjs.org/docs/messages/react-hydration-error) are specific to React. They are thrown when the HTML content of an element differs between the server and client after React is hydrated (loaded client-side resources).

We have the l10n hook in this file for client-side excecution and [src/app/functions/server/l10n.ts](https://github.com/mozilla/blurts-server/blob/main/src/app/functions/server/l10n.ts) for server-side rendering. Due to how server and client components work since React18 and Next.js 13 client components are being pre-rendered on the server if possible. The client l10n hook will not inject elements on the server though which is resulting in different content between server and client, thus throwing a hydration error which we can avoid with returning `null` at first.

src/app/(proper_react)/redesign/(authenticated)/user/dashboard/View.tsx
```diff
@@ -172,37 +172,39 @@ export const View = (props: Props) => {
...
+        </li>
+      ) : (
+        // Breaches result
+        <li
+          key={`breach-${exposure.Id}-${index}`}
+          className={styles.exposureListItem}
+        >
+          <ExposureCard
+            exposureData={exposure}
+            exposureName={exposure.Name}
```
pdehaan, line 199: What's this? The capitalization is "Weird" which makes me think it comes from HIBP; but wanted to confirm we aren't displaying the `Name` (since we typically only display the `breach.Title` to end users).

locales-pending/premium.ftl
```diff
@@ -41,6 +41,11 @@ exposure-chart-legend-value-nr = { $nr }x
 
 # Here's What We Fixed Progress Card
 
+exposure-card-company-logo = Company logo
+exposure-card-company = Company domain
+exposure-card-exposure-type = Exposure types
```
pdehaan, line 46: Slight discrepancy here between "type" and "types".
Not sure if we should pluralize the ID or singular-ize the string.

src/app/(proper_react)/redesign/(authenticated)/user/dashboard/View.tsx
```diff
@@ -172,37 +172,39 @@ export const View = (props: Props) => {
...
+            fromEmail={email}
+            exposureDetailsLink={""}
```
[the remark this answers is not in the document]
flozia: Updated in [6ea0351](https://github.com/mozilla/blurts-server/pull/3246/commits/6ea0351f6dca99ec79fed1791ee4877c958ab0c5).

locales-pending/premium.ftl
flozia: Updated in [6ea0351](https://github.com/mozilla/blurts-server/pull/3246/commits/6ea0351f6dca99ec79fed1791ee4877c958ab0c5).

src/app/(proper_react)/redesign/(authenticated)/user/dashboard/View.tsx
flozia: Correct, that’s data coming from HIBP. Updated to use `.Title` instead of `.Name` in [56f344b](https://github.com/mozilla/blurts-server/pull/3246/commits/56f344b20716a565a39a0e5a5a6214415ccf9a77).

rhelmer (approved):

PR #3251, opened 2023-07-25
Add Fix Flow Components to Storybook

maxxcrawford:
## Summary

- Add `<FixNavigation />` and `<DataBrokerProfiles />` components to Storybook

## Testing

- Run storybook
- Review components


src/app/components/client/stories/DataBrokerProfiles.stories.ts
```diff
@@ -0,0 +1,455 @@
...
+
+// More on how to set up stories at: https://storybook.js.org/docs/react/writing-stories/introduction
+const meta: Meta<typeof DataBrokerProfiles> = {
+  title: "DataBrokerProfiles",
+  component: DataBrokerProfiles,
+};
+export default meta;
+type Story = StoryObj<typeof DataBrokerProfiles>;
+
+const MockScanResults = [
```
pdehaan, line 18: Instead of hardcoding 10(?) results here, do we want to use mocks from https://github.com/mozilla/blurts-server/blob/af9ee12e94844cd83408b38cd2c192805f44d746/src/apiMocks/mockData.ts#L29-L54

We even have this utility fn that returns an array with `3` scan results, but should be easy enough to configure to return a optional variable number https://github.com/mozilla/blurts-server/blob/af9ee12e94844cd83408b38cd2c192805f44d746/src/apiMocks/mockData.ts#L17-L22

Using the mock factory would reduce this file by like 431 lines of code.

maxxcrawford: Love this.

PR #3261, opened 2023-07-27
Redirect to auth trigger URL

flozia:
# Description

Redirect back to the URL that initiated authentication if not explicitly set otherwise.

### DIFF
--- a/src/app/(nextjs_migration)/components/client/SignInButton.tsx
+++ b/src/app/(nextjs_migration)/components/client/SignInButton.tsx
@@ -5,27 +5,30 @@
 "use client";
 
 import { signIn } from "next-auth/react";
+import { usePathname } from "next/navigation";
 import { useL10n } from "../../../hooks/l10n";
 
 export type Props = {
   autoSignIn?: boolean;
 };
 
-function initSignIn() {
-  void signIn("fxa", { callbackUrl: "/user/breaches" });
+function initSignIn(callbackUrl: string) {
+  void signIn("fxa", { callbackUrl });
 }
 
 export const SignInButton = ({ autoSignIn }: Props) => {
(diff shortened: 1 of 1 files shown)


src/app/(nextjs_migration)/components/client/SignInButton.tsx
```diff
@@ -5,27 +5,30 @@
...
+import { usePathname } from "next/navigation";
 import { useL10n } from "../../../hooks/l10n";
 
 export type Props = {
   autoSignIn?: boolean;
 };
 
-function initSignIn() {
-  void signIn("fxa", { callbackUrl: "/user/breaches" });
+function initSignIn(callbackUrl: string) {
```
rhelmer, line 15: hm, how do we use this from e.g. src/app/(proper_react)/redesign/(authenticated)/user/welcome/page.tsx - does it need to be a prop like `autoSignIn`, so the caller can specify the right callback URL?

rhelmer (changes requested):

src/app/(nextjs_migration)/components/client/SignInButton.tsx
flozia: It should be already working from `src/app/(proper_react)/redesign/(authenticated)/user/welcome/page.tsx`. We have the `SignInButton` set there: https://github.com/mozilla/blurts-server/blob/main/src/app/(proper_react)/redesign/(authenticated)/user/welcome/page.tsx#L18.

rhelmer: Oh, I misunderstood where `pathname` was coming from, no this makes sense now thanks!

rhelmer (approved):

PR #3262, opened 2023-07-28
Add logging and simplify isEligible check

rhelmer:
# Description

Tiny followup from eng meeting to logging to explain why user is ineligible for free scan.

While in there, I noticed the date check for the scan is redundant - free scan is only ever needed once.

# How to test

Ensure logging in the following scenarios:

1. `FreeBrokerScan` flag does not exist
2. flag exists, but is disabled
3. flag exists and is enabled, but user is not on allow list
4. above, but free scan has been used


mansaj:
@rhelmer do we want to throw these in sentry as well?

rhelmer:
> @rhelmer do we want to throw these in sentry as well?

I think these are more warnings that we might use for debugging than errors... someone trying to do another scan, or access a feature that isn't enabled, etc. feel more on par with e.g. 404 errors (which I'd expect to just be logged) and not e.g. 500 errors (which are important and should be logged and tracked somewhere all engineers can see like Sentry).

mansaj (approved):

PR #3270, opened 2023-07-31
Focus-related accessibility refinements

Vinnl:
I ran into a couple of issues tabbing through the dashboard with my keyboard - I couldn't see what I had focused. This resolves that for a couple of elements.

I also ran into a couple of issues trying to use the filter through my keyboard, but since I couldn't actually filter my list either, I assume that that just isn't done yet?


flozia (approved):
Thanks for making these adjustments.

Regarding the filter: QA already [filed an issue](https://mozilla-hub.atlassian.net/browse/MNTOR-1977) for the filter. I also noticed it while working on the “fixed tab”. I [tweaked the filter logic here](https://github.com/mozilla/blurts-server/pull/3269/files#diff-d701455e388547a86562525e090b5bd1be6d2f06cee5aa5187d602f12253640dR154), which should resolve the issue.

PR #3274, opened 2023-08-01
Up test coverage threshold for future code to 100%

Vinnl:
All future code that is inside a file that gets executed as part of a test, test coverage should be 100%. I have added exceptions for existing files under test that do not reach this threshold; ideally those exceptions get removed over time.

We will probably want to have a conversation about which code does not need unit tests; those can either be added to the `coveragePathIgnorePatterns` property in `jest.config.cjs` if those reasons apply to an entire file; otherwise, specific lines of code can be ignored using specific comments as described in https://github.com/istanbuljs/v8-to-istanbul/blob/fca5e6a9e6ef38a9cdc3a178d5a6cf9ef82e6cab/README.md#ignoring-uncovered-lines

I have already applied a couple of such comments to existing code. For example, `useElementWidth` heavily interacts with `window` to keep a state variable up-to-date with an element's size; testing that would involve extensively mocking resize events. Test failures would be more likely to be caused by problems in the test than by actual bugs in the code, so writing a test for that is probably not worth it.

This is a proposal, and even if we merge this, we can adjust course later if it turns out not to work well for us.

### DIFF
--- a/docs/test-coverage.md
+++ b/docs/test-coverage.md
@@ -1,0 +1,29 @@
+# How to get to 100% test coverage
+
+We have set up a very strict test coverage threshold: 100% branch coverage. How can you achieve that?
+
+The first thing to realise is: we don't actually have 100% coverage. Not all code needs unit tests, however, we want it to be a conscious choice not to test a particular code branch. Thus, we make liberal use of [ignore comments](https://github.com/istanbuljs/v8-to-istanbul/blob/fca5e6a9e6ef38a9cdc3a178d5a6cf9ef82e6cab/README.md#ignoring-uncovered-lines), but those should be accompanied by a comment explaining why that code is OK to ignore. And maybe this is all too much of a hassle, and we disable these thresholds again.
+
+## The coverage report
+
+One very important tool in your toolchain is the coverage report. After running `npm test` locally, you'll be able to find a graphical coverage report inside `/coverage/lcov-report/index.html`. Open that in a web browser, and you'll be able to see which code paths you forgot to cover in a test. Or alternatively, which code paths should be marked with an ignore comment.
+
+## How to write tests for UI
+
+For tests concerning the UI in particular, you can get the most bang for your buck by simulating user actions, resisting the urge to drop down to the component level and mocking out individual props. [Think integration tests, not unit tests](https://kentcdodds.com/blog/write-tests), in so far as there's a clear distinction in the first place. Make your tests [reason about the UI like users do](https://testing-library.com/docs/queries/about#priority) (for example, by finding a button with a particular label to click on, rather than the fifth child of an element with class name `navigation`).
+
+Both of those measures ensure we won't have to update a whole bunch of unit tests unless we actually meaningfully change the behaviour for users, while still verifying the behaviour of a whole slew of components in one go.
+
+## When is it OK not to add a test?
+
+What follows is a list of some reasons not to write a unit test - your ignore comments can refer to these reasons. However, this list is not exhaustive! If you come across another valid reason not to test a bit of code, feel free to add that to this list and then reference it in your code.
+
+### Mock-heavy
+
+If most of the code under test would be interacting with mocks, writing the test would be a lot of work, would require lots of maintenance to keep the mocks up-to-date, while not being very likely to actually catch bugs. This is particularly likely for code that has lots of side effects, e.g. because it executes database queries or sends HTTP requests.
+
+Feel free to mark such code as skipped for coverage. However, do consider extracting code that is side-effect-free into a separate, testable, function. For example, if you have code that sends an HTTP request, then parses the response, that parsing code can be extracted into a separate function that takes the response body and converts it into parsed data.
+
+### Trivial
+
+Some code is exceptionally trivial, such that it is unlikely to contain bugs. When you're in the habit of writing unit tests, it might still be easy enough to set up a quick unit test because why not, but sometimes the effort probably won't pay off.
--- a/jest.config.cjs
+++ b/jest.config.cjs
@@ -37,9 +37,14 @@ const customJestConfig = {
   coverageDirectory: "coverage",
 
   // An array of regexp pattern strings used to skip coverage collection
-  // coveragePathIgnorePatterns: [
-  //   "/node_modules/"
-  // ],
+  coveragePathIgnorePatterns: [
+    "/node_modules/",
+    "<rootDir>/src/db/knexfile.js",
+    // Old, pre-Next.js code assumed to be working:
+    "<rootDir>/src/appConstants.js",
+    "<rootDir>/src/views/",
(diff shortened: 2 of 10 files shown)


rhelmer (approved):
lgtm overall, thanks for putting this together! Added a few thoughts and a question.

I'd really like to fix (or remove if we can't / don't want to) the code coverage badge on the README in our repo - however I'd like to bring up the topic of updating our docs and README in our next tech meeting anyway so we can tackle the code coverage badge issue then if you don't want to do it here.

docs/test-coverage.md
```diff
@@ -0,0 +1,29 @@
...
+
+## When is it OK not to add a test?
+
+What follows is a list of some reasons not to write a unit test - your ignore comments can refer to these reasons. However, this list is not exhaustive! If you come across another valid reason not to test a bit of code, feel free to add that to this list and then reference it in your code.
+
+### Mock-heavy
+
+If most of the code under test would be interacting with mocks, writing the test would be a lot of work, would require lots of maintenance to keep the mocks up-to-date, while not being very likely to actually catch bugs. This is particularly likely for code that has lots of side effects, e.g. because it executes database queries or sends HTTP requests.
+
+Feel free to mark such code as skipped for coverage. However, do consider extracting code that is side-effect-free into a separate, testable, function. For example, if you have code that sends an HTTP request, then parses the response, that parsing code can be extracted into a separate function that takes the response body and converts it into parsed data.
```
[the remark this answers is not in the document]
Vinnl: Yeah agreed. I'm being somewhat deliberately fuzzy - "most", "a lot of", "lots of", "not ... very likely". I didn't want to rule out mocks altogether, and the default should still be to write unit tests, but at some point, the effort invested probably doesn't outweigh the payoff, and at that point you can add an exception for a bit of code and point to this section.

jest.config.cjs
```diff
@@ -53,7 +58,185 @@ const customJestConfig = {
...
+      lines: 0,
+      statements: 14,
+    },
+    "src/db/tables/*.js": {
+      branches: 0,
+      functions: 0,
+      lines: 0,
+      statements: 0,
+    },
+  },
```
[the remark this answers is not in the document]
Vinnl: Ideally we never would - what would happen is that CI would start complaining if those new files/directories are not covered by unit tests :)

(One bit of nuance: the coverage report only measures files that actually get loaded during tests. So if not a single line of code in a file is imported when running the tests, it will not be counted as uncovered.)

jest.config.cjs
```diff
@@ -53,7 +58,185 @@ const customJestConfig = {
...
 
   // An object that configures minimum threshold enforcement for coverage results
-  // coverageThreshold: undefined,
+  coverageThreshold: {
+    global: {
+      branches: 100,
+      functions: 100,
+      lines: 100,
+      statements: 100,
+    },
```
pdehaan, line 67: 😍

src/app/functions/server/getCountryCode.ts
```diff
@@ -19,9 +19,9 @@ export function getCountryCode(
   const acceptLanguage = headers.get("Accept-Language");
   if (acceptLanguage) {
     const acceptedLocales = acceptLanguage.split(",");
-    const primaryLocale = acceptedLocales[0] ?? "";
+    const primaryLocale = acceptedLocales[0];
     const [locale, _weight] = primaryLocale.split(";");
-    const [_language, country] = (locale ?? "").split("-");
+    const [_language, country] = locale.split("-");
```
pdehaan, line 24: Q: why these changes?
Seems possibly unrelated to code coverage thresholds.

src/app/functions/server/mockL10n.ts
```diff
@@ -15,11 +15,15 @@ import type { LocaleData } from "./l10n";
 /* eslint-disable @typescript-eslint/no-var-requires */
 
 export function getEnL10nBundlesSync(): LocaleData[] {
+  // Ignored for test coverage, since tests don't run in a Webpack context:
+  /* c8 ignore next 2 */
```
pdehaan, line 19: Should this be `next 2` or `next 3`, or maybe i'm confused by line counting on ternary operators.

src/app/(proper_react)/redesign/(authenticated)/user/dashboard/DashboardTopBanner.tsx
```diff
@@ -43,6 +43,8 @@ export const DashboardTopBanner = (props: DashboardTopBannerProps) => {
       ),
       cta: {
         content: l10n.getString("dashboard-top-banner-protect-your-data-cta"),
+        // Ignored for test coverage; to be replaced by a link:
+        /* c8 ignore next 4 */
```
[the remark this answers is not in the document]
Vinnl: It's a shame we're no longer using Istanbul for measuring coverage - that [had](https://github.com/gotwarlost/istanbul/blob/master/ignoring-code-for-coverage.md) `ignore if` and `ignore else` comments, and `ignore next` appeared to skip a whole node in the tree (e.g. a function block) rather than a single line.

But yes, it can get annoying with Prettier and having to go back because of CI rejecting your code. My rule of thumb was kinda "big code blocks use `start ... stop`, a couple of lines use `next`. But there's something to be said for not using `n` and always using `start ... stop` if it's more than a single line.

...in fact, I was on my way replacing them, but then I realised that that would increase the risk of future edits to the code staying within the `ignore` block, and then not get covered with unit tests. The advantage of using specific line numbers, is that future edits to the code will very likely trigger a reconsidering of whether that code shouldn't in fact need unit tests as well. 

So maybe a good rule of thumb is: `next n` if future edits are likely to need unit tests, and `start ... stop` if they don't?

src/app/functions/server/getCountryCode.ts
Vinnl: The `?? ""` was marked as uncovered, and didn't actually do anything (`.split` always returns an array with at least one element). So rather than adding an exception, I figured I'd apply this small change to get it to 100% coverage.

src/contextProviders/localization.tsx
```diff
@@ -31,6 +31,9 @@ export const L10nProvider = (props: {
   // To enable server-side rendering, all tags are converted to plain text nodes.
   // They will be upgraded to regular HTML elements in the browser:
   const parseMarkup: MarkupParser | undefined =
+    // Ignored for test coverage, since `document` is mocked (and thus defined)
+    // in unit tests:
+    /* c8 ignore next 7 */
```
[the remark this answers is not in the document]
Vinnl: `7`; the other branch of the ternary does get hit in a unit test.

src/app/functions/server/mockL10n.ts
Vinnl: Same here; it's `2`, because the Node context branch does get hit in unit tests.

mansaj (approved):

PR #3281, opened 2023-08-01
MNTOR-1994/use flags to disable UI in response to removal/free scan flags

rhelmer:
# References: 
Jira: MNTOR-1994

# Description

This includes a few commits:

1. refactor `getFlag` server-side React function to `isFlagEnabled`, moving more logic inside this function to make things easier for callers
2. use the new `isFlagEnabled` function from the page server component to pass props to the client-side components on the dashboard, and use these props to enable or disable features.

I tried to keep things simple for starters, but we can make these more advanced as we identify more use cases.

# Screenshot (if applicable)

Screenshot 2023-08-01 at 3 43 56 PM
Screenshot 2023-08-01 at 3 43 41 PM
Screenshot 2023-08-01 at 3 43 30 PM
Screenshot 2023-08-01 at 3 43 15 PM

# How to test

Try enabling/disabling feature flags and reloading dashboard, see screenshots.


src/app/(proper_react)/redesign/(authenticated)/user/dashboard/View.tsx
```diff
@@ -216,17 +220,23 @@ export const View = (props: Props) => {
...
+  let type = "NoContent";
+  if (
+    props.featureFlagsEnabled?.FreeBrokerScan &&
+    props.featureFlagsEnabled?.PremiumBrokerRemoval
+  ) {
+    type = isScanResultItemsEmpty
+      ? "DataBrokerScanUpsellContent"
+      : "LetsFixDataContent";
+  }
+
```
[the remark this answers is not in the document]
rhelmer: lgtm but I need to apply it locally since I made some conflicting unrelated changes in the meantime, thanks!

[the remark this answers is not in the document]
rhelmer: Oh hm actually wait I don't think I do understand... my intent was to set the `type` to `NoContent`, unless both flags are enabled. I think what we really want is more complicated (different content should be eligible to be shown depending on which combination of flags is enabled), but I don't want to overcomplicate things either.

I think the states we care about are something like this:

src/app/(proper_react)/redesign/(authenticated)/user/dashboard/View.tsx
```diff
@@ -216,17 +220,23 @@ export const View = (props: Props) => {
...
       <Toolbar user={props.user} />
       <div className={styles.dashboardContent}>
         <DashboardTopBanner
           bannerData={props.bannerData}
-          type={
-            isScanResultItemsEmpty
-              ? "DataBrokerScanUpsellContent"
-              : "LetsFixDataContent"
-          }
+          type={type}
```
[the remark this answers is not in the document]
rhelmer: OK I get it now sorry, things got confusing from merging changes in the meantime. Thanks!

src/app/functions/server/featureFlags.ts
```diff
@@ -34,5 +46,19 @@ export async function getFlag(name: string): Promise<FeatureFlag | undefined> {
     owner: data.owner,
   };
 
-  return flag;
+  if (!flag.isEnabled) {
+    console.warn("Flag is not enabled:", flag.name);
+    return false;
+  }
```
pdehaan, line 52: Not sure if it belongs here or elsewhere, but should this check if the flag hasn't been soft deleted?
Alternately, we could auto-disable experiments when soft deleting them in the database and then the `.isEnabled` check would ignore deleted experiments.

src/app/functions/server/featureFlags.ts
```diff
@@ -7,10 +7,22 @@ import {
...
+export type FeatureFlagsEnabled = {
+  FreeBrokerScan: boolean;
+  PremiumBrokerRemoval: boolean;
+};
+
+export async function isFlagEnabled(name: string): Promise<boolean> {
+  const session = await getServerSession(authOptions);
+  if (!session?.user?.email) {
+    throw new Error("No session");
+  }
```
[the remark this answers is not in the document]
rhelmer: Yes we do want this, I think we should file a ticket and make it an explicit field whether it's a public flag or not.

src/app/functions/server/featureFlags.ts
```diff
@@ -7,10 +7,22 @@ import {
...
 } from "../../../db/tables/featureFlags";
+import { getServerSession } from "next-auth";
+import { authOptions } from "../../api/utils/auth";
+
+export type FeatureFlagsEnabled = {
+  FreeBrokerScan: boolean;
+  PremiumBrokerRemoval: boolean;
+};
+
+export async function isFlagEnabled(name: string): Promise<boolean> {
```
Vinnl, line 18: You can make this same change for every function inside `tables/featureFlags.ts`.

```suggestion
export async function isFlagEnabled(name: keyof FeatureFlagsEnabled): Promise<boolean> {
```

src/app/functions/server/featureFlags.ts
```diff
@@ -34,5 +46,19 @@ export async function getFlag(name: string): Promise<FeatureFlag | undefined> {
     owner: data.owner,
   };
 
-  return flag;
+  if (!flag.isEnabled) {
+    console.warn("Flag is not enabled:", flag.name);
```
Vinnl, line 50: Aren't the `console` messages (both this and the others in this file) going to be very noisy in our logs?

src/app/functions/server/featureFlags.ts
```diff
@@ -7,10 +7,22 @@ import {
...
+import { getServerSession } from "next-auth";
+import { authOptions } from "../../api/utils/auth";
+
+export type FeatureFlagsEnabled = {
+  FreeBrokerScan: boolean;
+  PremiumBrokerRemoval: boolean;
+};
+
+export async function isFlagEnabled(name: string): Promise<boolean> {
+  const session = await getServerSession(authOptions);
```
Vinnl, line 19: This might be more of a personal preference than a valid remark, so feel free to ignore, but I tend to try minimising (ideally to 0) the number of side effects in my function calls. So it might be preferable to just have the caller pass in a `user` object?

- In most use cases, I expect that the caller will have access to that already, so it would save an additional unwrapping of the JWT.
- It would remove the need to mock the module in the test.
- It would allow checking if some _other_ user is able to see something (possibly useful in an admin page, or if the user has just logged out).
- We can do away with the runtime error by making it a required parameter. (Or actually make it optional, for flags without an allowlist.)

src/app/functions/server/featureFlags.ts
```diff
@@ -7,10 +7,22 @@ import {
...
+    throw new Error("No session");
+  }
 
-export async function getFlag(name: string): Promise<FeatureFlag | undefined> {
   if (!name) {
-    const err = new Error("No name provided to getFlag");
+    const err = new Error("No name provided to isFlagEnabled");
     captureException(err);
     throw err;
   }
```
Vinnl, line 28: Do we expect this function to be called by non-TypeScript code? Otherwise this check is redundant with the type.

src/app/functions/server/featureFlags.ts
```diff
@@ -34,5 +46,19 @@ export async function getFlag(name: string): Promise<FeatureFlag | undefined> {
...
+  if (!flag.allowList?.length) {
+    console.info("Flag does not have an allow list, enabling:", flag.name);
+  }
+
+  if (flag.allowList?.length && !flag.allowList?.includes(session.user.email)) {
+    console.warn("User is not on allow list for flag:", flag.name);
+    return false;
+  }
+
+  return true;
```
Vinnl, line 63: Minor nit, and would also require changes higher up, but I feel like accidentally enabling a flag might be worse than accidentally disabling it, and so the default might be better as `false`?

```suggestion
  if (!flag.allowList?.length || flag.allowList.includes(session.user.email)) {
    return true;
  }

  return false;
```

src/app/functions/server/featureFlags.test.ts
```diff
@@ -0,0 +1,58 @@
...
+jest.mock("../../../db/tables/featureFlags", () => ({
+  getFeatureFlagByName: jest.fn(),
+}));
+
+import { isFlagEnabled } from "./featureFlags";
+
+beforeEach(() => {
+  jest.resetModules();
+});
+
```
Vinnl, line 29: Hmm, I'd swear that I'd enabled `resetMocks` and `resetModules` by default when I set up Jest, but [apparently not](https://github.com/mozilla/blurts-server/blob/698e67a462a3a30ea46c0d243b166d0f0aadce73/jest.config.cjs#L126-L130). Either way, I think it's preferable to enable that globally rather than setting that for each (or worse, some) individual test, for predictability. Tests should never rely on the results from previous tests anyway.

src/app/functions/server/featureFlags.test.ts
```diff
@@ -0,0 +1,58 @@
...
+it("throws if there is no valid session", async () => {
+  await expect(isFlagEnabled("testFlag")).resolves.toBe(false);
```
Vinnl, line 31: I suggested elsewhere to take the user object as an argument, but as an FYI, this test description doesn't seem to match what it does. The current implementation is actually the " or nonexistent" part of the next test.

Checking correct behaviour for no session will be easier if it's passed in as an argument, but as an FYI, this is how you could've done it:

```suggestion
it("throws if there is no valid session", async () => {
  const nextAuthMock = jest.requireMock("next-auth");
  nextAuthMock.getServerSession.mockImplementationOnce(() => undefined);
  const { isFlagEnabled } = await import("./featureFlags");
  await expect(() => isFlagEnabled("testFlag")).rejects.toThrow("No session");
```

Unfortunately after mocking a module you have to import the module that depends on it again.

src/app/functions/server/featureFlags.test.ts
```diff
@@ -0,0 +1,58 @@
...
+});
+
+it("returns false if flag not provided or nonexistent", async () => {
+  //@ts-ignore TS should prevent, but let's test at runtime anyway.
```
Vinnl, line 35: You can even turn this into a test to verify that the type definition also disallows it at build time:

```suggestion
  // @ts-expect-error TS should prevent, but let's test at runtime anyway.
```

See https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-9.html#-ts-expect-error-comments

src/app/functions/server/featureFlags.test.ts
```diff
@@ -0,0 +1,58 @@
...
+  await expect(isFlagEnabled()).rejects.toThrowError(
+    "No name provided to isFlagEnabled"
+  );
+
+  await expect(isFlagEnabled("invalidFlag")).resolves.toBe(false);
```
Vinnl, line 40: And with my `keyof` suggestion elsewhere, this should start erroring at build time too:

```suggestion
  // @ts-expect-error
  await expect(isFlagEnabled("invalidFlag")).resolves.toBe(false);
```

src/app/functions/server/featureFlags.test.ts
```diff
@@ -0,0 +1,58 @@
...
+});
+
+it("returns flag state if exists", async () => {
+  //@ts-ignore jest mock
+  getFeatureFlagByName.mockImplementation((name) => {
```
Vinnl, line 45: `ts-ignore` ignores _all_ potential errors; I'd prefer telling TypeScript that we know it to be of a particular type:

```suggestion
  (getFeatureFlagByName as jest.Mock).mockImplementation((name) => {
```

src/app/components/client/ExposureCard.tsx
```diff
@@ -59,6 +60,7 @@ export type ExposureCardProps = {
   locale: string;
   fromEmail?: string;
   color: string;
+  featureFlagsEnabled: FeatureFlagsEnabled;
```
Vinnl, line 63: I think I'd split this up into _just_ the feature flags we're interested in, or just pick those from that object, because you're constructing this object manually on the dashboard page. Otherwise, every time we add a new flag to `FeatureFlagsEnabled`, the dashboard page and all tests will have to be updated to add it as well, even if it's not at all relevant.

For example:

```suggestion
  premiumBrokerRemovalEnabled: boolean;
```

or

```suggestion
  featureFlagsEnabled: Pick<FeatureFlagsEnabled, "FreeBrokerScan" | "PremiumBrokerRemoval">;
```

src/app/(proper_react)/redesign/(authenticated)/user/dashboard/View.tsx
```diff
@@ -27,13 +27,15 @@ import { DashboardSummary } from "../../../../../functions/server/dashboard";
...
 import AllFixedLogo from "./images/dashboard-all-fixed.svg";
+import { FeatureFlagsEnabled } from "../../../../../functions/server/featureFlags";
 
 export type Props = {
   user: Session["user"];
   userBreaches: UserBreaches;
   userScannedResults: ScanResult[];
   bannerData: DashboardSummary;
   locale: string;
+  featureFlagsEnabled: FeatureFlagsEnabled;
```
Vinnl, line 38: Likewise, I'd limit this to just the relevant feature flags.

Vinnl (approved):
Most important changes I'd like to see:

- only allow known feature flags rather than generic strings.
- don't require passing in the full list of feature flags to the `<ExposureCard>`.

src/app/functions/server/featureFlags.ts
rhelmer: Possibly, I'm a bit worried about how to debug what's wrong if things aren't working for someone... I'll think about it some more, maybe there's a better way to do this.

src/app/functions/server/featureFlags.ts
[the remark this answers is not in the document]
pdehaan: As you wish: MNTOR-2008

src/app/functions/server/featureFlags.ts
Vinnl: Isn't `mozlog` intended to allow us to log messages with different levels that only show up on particular environments?

src/app/functions/server/featureFlags.ts
rhelmer: We haven't fully implemented this yet (e.g. there's no way to do this in the UI) but I'll add the check for it here anyway, thanks!

src/app/functions/server/featureFlags.ts
rhelmer: Yes it is, `mozlog` is in a bit of an odd state right now but I think there's been some work recently on bringing it up-to-date. I am not sure it does very good structured logging for GCP or anything else that improves much over the `console.*` levels (`info/warning/error` etc), which are handled OK by the GCP logging tools.

src/app/functions/server/featureFlags.ts
rhelmer: That's a good idea, thanks! I'd prefer to have more side-effect-free functions too.

src/app/functions/server/featureFlags.test.ts
rhelmer: Ah I wasn't sure how to do this - thanks!

src/app/functions/server/featureFlags.ts
rhelmer: OK done - left the mock in place for the DB call, but that seems reasonable in this case IMHO and is fairly painless to mock, but we could consider if there's a better way.

rhelmer:
@Vinnl FYI we backslid a bit on branch coverage - I think that's expected in some cases though since we don't necessarily want to thoroughly test every combination when we know it's temporary (in this case I think we may not even need these two flags to be separate like we thought we did when I originally filed this issue).

That said, I think it's useful enough as an example to land as-is. We should be sure to up the test coverage when we remove flags though, I'm not sure how best to systematize this.

Vinnl:
@rhelmer Yeah that's fair enough, though I think it's probably good to mark that temporary code as ignored for coverage and a comment explaining that it's temporary, rather than adjusting the thresholds. (Actually, "This code has a known expiration date" is a good common case to add to the list of valid exceptions.

pdehaan:
> @Vinnl: _Yeah that's fair enough, though I think it's probably good to mark that temporary code as ignored for coverage and a comment explaining that it's temporary, rather than adjusting the thresholds. (Actually, "This code has a known expiration date" is a good common case to add to the list of valid exceptions.)_

Q: How do we audit that though? 
There's a [famous Russian proverb](https://allauthor.com/quotes/60176/) _"Nothing is more permanent than a temporary solution."_

So I'd be afraid of adding so many temporary patches and ignore statements to keep coverage and linting happy, and then we never get around to paying off the large amounts of tech debt that we've accumulated before moving onto other projects and just kicking the can further and further down the road. Unless we aggressively say that we can't release until all the TODOs and temporary solutions have been addressed in the codebase.

Vinnl:
@pdehaan Oh sorry, I'm not saying the annotations should be temporary - I meant that if the _code under test_ is temporary, that's a valid reason for it to be marked as untested.

I do agree that we should be keeping a watchful eye on whether code is _actually_ temporary, but that goes beyond just unit tests. If we suspect there's a risk that it'll actually stick around for a long time, then we should maintain the same standards as we do elsewhere, including writing unit tests. But it appears that in this case, it was already decided that this did meet the "temporary" bar.

PR #3291, opened 2023-08-04
Group related Dependabot updates

Vinnl:
This should hopefully cut down on some of the noise of ESLint updates, and ensure that packages that break unless they are updated together are actually updated together.

See https://github.blog/changelog/2023-06-30-grouped-version-updates-for-dependabot-public-beta/


rhelmer (approved):
Sounds great to me, thanks!

PR #3297, opened 2023-08-09
MNTOR-1834 - False door test

codemist:
# References: 
Jira: MNTOR-1834
Figma: https://www.figma.com/file/8mehWAxiNiPIe3WN4W2gx0/Firefox-Monitor-Premium-%7C-Marketing?type=design&node-id=81-65&mode=design&t=Aipyoiird2ZfxRNK-4

# Description
This adds the false door test banner on all pages of Monitor, which is dismissed through cookies. 

Note: Add the follow to your `.env` file:
`FALSE_DOOR_TEST_LINK_PHASE_ONE=https://www.mozilla.org/en-US/newsletter/monitor-waitlist/`

Note: I installed [react-cookie](https://www.npmjs.com/package/react-cookie) here, we'll probably need to deprecate it in favour of our own cookie hook - opened up a ticket [here](https://mozilla-hub.atlassian.net/browse/MNTOR-2043).

# Screenshot (if applicable)
image
 dashboard variant:
image

# Checklist (Definition of Done)
- [ ] Localization strings (if needed) have been added.
- [ ] Commits in this PR are minimal and [have descriptive commit messages](https://chris.beams.io/posts/git-commit/).
- [ ] I've added or updated the relevant sections in readme and/or code comments
- [ ] I've added a unit test to test for potential regressions of this bug.
- [ ] Product Owner accepted the User Story (demo of functionality completed) or waived the privilege.
- [ ] All acceptance criteria are met.
- [ ] Jira ticket has been updated (if needed) to match changes made during the development process.
- [ ] Jira ticket has been updated (if needed) with suggestions for QA when this PR is deployed to stage.

PR #3303, opened 2023-08-11
MNTOR-2022: Integrate Glean.js & hook up generic events

rosahbruno:
# References: 
Jira: MNTOR-2022
Figma: 

# Description
Adds and configures Glean.js for Monitor. Since we are using Glean.js in the front-end, there are some specific ways in which we need to use Glean.

There are a few sharp edges I tried to address
- Next.js has no entry file that runs for every app load, so Glean needs to be initialized for every route. To manage this, I created a custom `useGlean` hook that handles initialization and returns the objects you need to record data.
- You cannot record Glean data directly on a non-client component, making things like recording a page view difficult. A pattern that we can use to handle this would be something like the `PageLoad` component which you can drop onto any page with whatever props you would like to pass as event extras. Then on first mount, the `PageLoad` component will record an event and then not run again as long as it doesn't re-render. So putting this at the top-level of a component and making sure it isn't a child component of a re-render heavy component is important.
- Rather than set up individual pings, we can send events as they happen using Glean's `maxEvent: 1` capability. This will record events and send a ping with those events immediately, similar to recording an event in GA.

# Screenshot (if applicable)

Not applicable.

# How to test
1. `npm install`
2. `npm run build-glean`
3. `npm run dev`
4. Open the browser console
5. Observe a block that shows data about the ping that was just submitted (when loading the home page).
6. Open https://debug-ping-preview.firebaseapp.com
7. Find the debug tag, currently checked in tag is "monitor-testing"
8. Confirm that the ping UUID in the browser console can be found in the debug ping viewer (sometimes a ping can take a few minutes to show up).


rosahbruno:
The error in the failing Playwright tests doesn't appear to mention anything that I added, though I could be missing something.

rhelmer:
> The error in the failing Playwright tests doesn't appear to mention anything that I added, though I could be missing something.

Looks like it's a problem in the build step, I think this is it @rosahbruno:
```
src/app/(proper_react)/redesign/(authenticated)/user/dashboard/fix/layout.tsx
Type error: Layout "src/app/(proper_react)/redesign/(authenticated)/user/dashboard/fix/layout.tsx" has an invalid "default" export:
  Type "Props" is not valid.
```

next.config.js
```diff
@@ -67,7 +67,7 @@ const nextConfig = {
               "script-src-attr 'none'",
               `connect-src 'self' ${
                 process.env.NODE_ENV === "development" ? "webpack://*" : ""
-              } https://*.google-analytics.com https://*.analytics.google.com https://*.googletagmanager.com https://*.ingest.sentry.io`,
+              } https://*.google-analytics.com https://*.analytics.google.com https://*.googletagmanager.com https://*.ingest.sentry.io https://incoming.telemetry.mozilla.org`,
```
rhelmer, line 70: This is fine for the a/a test, but we should consider routing through our own proxy eventually (I know this host is on some block lists).

src/app/hooks/useGlean.ts
```diff
@@ -0,0 +1,33 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+"use client";
+
+import { useEffect } from "react";
+import Glean from "@mozilla/glean/web";
```
rhelmer, line 8: Do we need something like https://www.mozilla.org/en-US/privacy/websites/data-preferences/ ?

rhelmer (commented):
This is looking great, just a few thoughts. Thanks for working on this @rosahbruno!

src/app/hooks/useGlean.ts
Dexterp37: @rhelmer if GA doesn't need it, Glean doesn't need it. If you're product needs it, then you can/should add it. You might want to double check with legal, but there's no Glean-specific requirement around this.

src/app/(nextjs_migration)/components/client/PageLoad.tsx
```diff
@@ -0,0 +1,27 @@
...
+}
+
+// Empty component that records a page view on first load.
+export const PageLoad = ({ pageTitle }: PageLoadProps) => {
+  const { appEvents } = useGlean();
+
+  // On first load of the page, record a page view.
+  useEffect(() => {
+    if (pageTitle) {
+      appEvents.pageView.record({ label: pageTitle });
```
Dexterp37, line 21: You likely want to record the page path here as well (or that's pretty much the one thing you want to record)

src/telemetry/generated/appEvents.js
```diff
@@ -0,0 +1,39 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+// AUTOGENERATED BY glean_parser v8.1.1. DO NOT EDIT. DO NOT COMMIT.
+
+import EventMetricType from "@mozilla/glean/private/metrics/event";
```
Dexterp37, line 7: Please let's not commit generated code and, in fact, let's add the generated dir to the .gitignore to be sure :)

src/telemetry/metrics.yaml
```diff
@@ -0,0 +1,52 @@
...
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+---
+$schema: moz://mozilla.org/schemas/glean/metrics/2-0-0
+
+app_events:
+  page_view:
+    type: event
+    lifetime: ping
```
Dexterp37, line 11: @rosahbruno can this be left out? If so, let's leave this out, it's the only allowed value :)

src/telemetry/metrics.yaml
```diff
@@ -0,0 +1,52 @@
...
+    send_in_pings:
```
Dexterp37, line 12: Let's leave this out, evens already go there.

src/app/(nextjs_migration)/components/client/PageLoad.tsx
Dexterp37: After seeing @rhelmer slack, we likely need to add everything that's in https://support.google.com/analytics/answer/9234069?hl=en

flozia:
> > The error in the failing Playwright tests doesn't appear to mention anything that I added, though I could be missing something.
> 
> Looks like it's a problem in the build step, I think this is it @rosahbruno:
> 
> ```
> src/app/(proper_react)/redesign/(authenticated)/user/dashboard/fix/layout.tsx
> Type error: Layout "src/app/(proper_react)/redesign/(authenticated)/user/dashboard/fix/layout.tsx" has an invalid "default" export:
>   Type "Props" is not valid.
> ```

This test failure has been fixed with a [recent PR](https://github.com/mozilla/blurts-server/pull/3297/commits/2e52c3befd4400856a9d90737d11e1a44fdd2daa) — after rebasing with `main` issue should be resolved.

package.json
```diff
@@ -23,7 +23,8 @@
     "storybook": "storybook dev -p 6006",
     "build-storybook": "storybook build",
     "create-location-data": "node src/scripts/uploadAutoCompleteLocations.js",
-    "get-location-data": "node src/scripts/getAutoCompleteLocations.js"
+    "get-location-data": "node src/scripts/getAutoCompleteLocations.js",
+    "build-glean": "glean translate src/telemetry/metrics.yaml -f javascript -o src/telemetry/generated"
```
Vinnl, line 27: Are there long-form versions of `-f` and `-o`? If so, I'd prefer we use those so future readers have a better idea of what they do :)

src/app/(nextjs_migration)/components/client/PageLoad.tsx
```diff
@@ -0,0 +1,27 @@
...
+    }
+  }, [appEvents, pageTitle]);
+
+  // This component doesn't render anything.
+  return <></>;
```
Vinnl, line 26: This makes me think it probably makes more sense to make this a hook, rather than a component?

src/telemetry/metrics.yaml
```diff
@@ -0,0 +1,52 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
```
Vinnl, line 4: As someone unfamiliar with this configuration file: could we also add a link to the docs on available options here?

src/telemetry/generated/appEvents.js
```diff
@@ -0,0 +1,39 @@
...
+ */
+export const pageView = new EventMetricType(
+  {
+    category: "app_events",
+    name: "page_view",
+    sendInPings: ["events"],
+    lifetime: "ping",
+    disabled: false,
+  },
+  ["label"]
```
Vinnl, line 38: Not sure if this is the right place for a feature request, but it would be great if the type definition of `ExtraMap` / `ExtraValues` could be derived from the values passed here, so that we can get autocompletion when we pass an argument to `appEvents.pageView.record`.

src/app/hooks/useGlean.ts
```diff
@@ -0,0 +1,33 @@
...
+import * as appEvents from "../../telemetry/generated/appEvents";
+
+// Custom hook that initializes Glean and returns the Glean objects required
+// to record data.
+export const useGlean = () => {
+  // Initialize Glean only on the first render
+  // of our custom hook.
+  useEffect(() => {
+    Glean.initialize("monitor-testing", true, {
```
Vinnl, line 17: I don't know how bad it is if this function gets called multiple times, but it currently gets called every time a component gets loaded that has `useGlean` (so currently twice), and also gets called again whenever new components get loaded that include it (e.g. on route changes).

src/app/hooks/useGlean.ts
```diff
@@ -0,0 +1,33 @@
...
+      // This will submit an events ping every time an event is recorded.
+      maxEvents: 1,
+    });
+
+    // Enable logging pings to the browser console.
+    Glean.setLogPings(true);
+
+    // Set the tag and sends pings to the debug ping viewer.
+    Glean.setDebugViewTag("monitor-testing");
```
Vinnl, line 26: We might only want to enable this if `APP_ENV` has a particular value (e.g. `local`, `development` or `heroku` - @rhelmer probably knows the relevant values).

src/app/hooks/useGlean.ts
```diff
@@ -0,0 +1,33 @@
...
+  }, []);
+
+  // Return all generated Glean objects required for recording data.
+  return {
+    appEvents,
+  };
```
Vinnl, line 32: What's the benefit of wrapping it in another object - is this going to return additional data later?

```suggestion
  return appEvents;
```

Vinnl (commented):
I know this is still draft, and I don't have all the context, but when I asked, @rhelmer said it could be helpful if I already took a look, so here we are :)

src/app/(nextjs_migration)/components/client/PageLoad.tsx
rosahbruno: I made this a component rather than a hook since I am using this on pages like `src/app/(nextjs_migration)/(guest)/page.tsx`. If this was a react component, then yes we could use it as a hook.

I am not an expert with SSR projects, is there a way that we could use a hook on that page?

src/app/hooks/useGlean.ts
rosahbruno: There could be depending on how this is set up. Since I am only exporting events from the same category it only creates one object. If you would add a different category header in the `metrics.yaml`, then it would generate two separate objects.

So yes for now we can return just the one object but over time this will likely become multiple objects with different categories.

src/telemetry/generated/appEvents.js
Dexterp37: @Vinnl would you kindly file a bug in bugzilla's Glean SDK component? It's a totally reasonable request

src/app/(nextjs_migration)/components/client/PageLoad.tsx
Vinnl: Ah gotcha. Well, I'd caveat this with that everyone's new to the whole server/client component shift, but... Yeah, it'd need a client component I think. There aren't yet estasblished patterns and best practices, but I think that I'd still prefer the functionality in a hook, and then having this component only load that hook - to clarify that the only reason the component is there, is to make the hook run client-side. But not a blocker :)

src/telemetry/generated/appEvents.js
Vinnl: Sure! https://bugzilla.mozilla.org/show_bug.cgi?id=1850542

Note that I did not know which version to specify - the lockfile says that it's `2.0.1`, which was not an option in Bugzilla, so I set it to `unspecified`.

Dexterp37: @Vinnl that's perfect, thank you for filing!

src/app/hooks/useGlean.ts
```diff
@@ -0,0 +1,35 @@
...
+      // This will submit an events ping every time an event is recorded.
+      maxEvents: 1,
+    });
+
+    if (process.env.NODE_ENV === "development") {
+      // Enable logging pings to the browser console.
+      Glean.setLogPings(true);
+
+      // Set the tag and sends pings to the debug ping viewer.
+      Glean.setDebugViewTag("monitor-testing");
```
Dexterp37, line 27: I would recommend leaving this line out completely and enable things from the console as needed when manual QA is needed.

src/app/hooks/useGlean.ts
rhelmer: `APP_ENV` can be accessed by front-end code with `NEXT_PUBLIC_APP_ENV`, the expected values are listed here: https://github.com/mozilla/blurts-server/blob/d1bf222bb285cfdb7cc6489db876997b24b89f20/.env-dist#L4-L6

src/app/(nextjs_migration)/components/client/PageLoad.tsx
rhelmer: Yes, I think for this PR we don't need to, Monitor team can do that as a followup - we'd like to just land something simple enough to do an A/A dry-run test of Nimbus integration.

src/app/(nextjs_migration)/components/client/SignInButton.tsx
```diff
@@ -19,16 +20,22 @@ function initSignIn(callbackUrl: string) {
...
 
+  const signInClick = () => {
+    appEvents.buttonClick.record({ label: "sign-in" });
+    initSignIn(callbackUrl);
+  };
+
   return (
     <button
-      onClick={() => initSignIn(callbackUrl)}
+      onClick={() => signInClick()}
```
rhelmer, line 38: Do we need to pass `callbackUrl` here?

src/app/hooks/useGlean.ts
rhelmer: I'm fine either way, we just want to make it easy for folks to see the data locally.

src/telemetry/metrics.yaml
```diff
@@ -0,0 +1,49 @@
...
+---
+$schema: moz://mozilla.org/schemas/glean/metrics/2-0-0
+
+app_events:
+  page_view:
+    type: event
+    description: |
+      A generic page view event with extras.
+    bugs:
+      # TODO: Add bug link
```
rhelmer, line 17: Can the TODO comment be removed?

src/telemetry/metrics.yaml
```diff
@@ -0,0 +1,49 @@
...
+      - https://bugzilla.mozilla.org/1691365
+    data_reviews:
+      # TODO: Add data review link
```
rhelmer, line 20: Can the TODO comment be removed?

rhelmer: There are a few more below as well.

rhelmer (changes requested):

src/app/(nextjs_migration)/(guest)/page.tsx
```diff
@@ -28,6 +29,7 @@ export default function Home() {
         crossOrigin="anonymous"
       />
       <Script type="module" src="/nextjs_migration/client/js/landing.js" />
+      <PageLoad pageTitle="Home" />
```
pdehaan, line 32: This is just for stats logging and isn't user facing, correct? (if it is user facing, we'd need to localize)

pdehaan: Also, below we have some code like a this:
```js
appEvents.buttonClick.record({ label: "sign-in" });
```

Should this be "home" to be consistent lowercasing w/ "sign-in"?

rosahbruno: I am going to change this `PageLoad` component to send the current URL path instead, so yes I agree with your comments but this is going to change.

src/app/(nextjs_migration)/components/client/SignInButton.tsx
rosahbruno: I am just removing that sign in button click event. I replaced the two arbitrary events I created with something more practical, which is an event that runs on a page load that records the current URL path.

rosahbruno:
@rhelmer 
I had to add an additional NPM script for creating the Glean generated files (`npm run build-glean`). How should this be called from CI? Since we aren't checking in the Glean generated files, the `lint` job fails every time since currently when CI runs nothing is generating those files.

When testing I just updated the NPM `postinstall` script to look like `patch-package && npm run build-glean`. <- This feels a bit hacky, what would be the preferred way to make sure those generated files get created every time we need them?

Dexterp37:
> @rhelmer I had to add an additional NPM script for creating the Glean generated files (`npm run build-glean`). How should this be called from CI? Since we aren't checking in the Glean generated files, the `lint` job fails every time since currently when CI runs nothing is generating those files.

@rosahbruno @rhelmer one way to solve it would be to add a new step [here](https://github.com/mozilla/blurts-server/blob/d1bf222bb285cfdb7cc6489db876997b24b89f20/.github/workflows/reference_linter.yaml#L20) to generate the files before the linting runs, right?

This is needed since the product is not being built in the linter task, otherwise it would be covered by the other suggested change.

package.json
```diff
@@ -23,7 +23,8 @@
     "storybook": "storybook dev -p 6006",
     "build-storybook": "storybook build",
     "create-location-data": "node src/scripts/uploadAutoCompleteLocations.js",
-    "get-location-data": "node src/scripts/getAutoCompleteLocations.js"
+    "get-location-data": "node src/scripts/getAutoCompleteLocations.js",
+    "build-glean": "glean translate src/telemetry/metrics.yaml --format javascript --output src/telemetry/generated"
```
Dexterp37, line 27: @rosahbruno I believe this must be added to the `build` script/action on line 12 just before `next build`.

next.config.js
Dexterp37: @rhelmer yes. I believe we (Data Org) should prioritize providing guidelines for this. Let's work together after the A/A test.

src/app/hooks/useGlean.ts
```diff
@@ -0,0 +1,37 @@
...
+import Glean from "@mozilla/glean/web";
+import * as appEvents from "../../telemetry/generated/appEvents";
+
+// Custom hook that initializes Glean and returns the Glean objects required
+// to record data.
+export const useGlean = () => {
+  // Initialize Glean only on the first render
+  // of our custom hook.
+  useEffect(() => {
+    Glean.initialize("monitor-testing", true, {
```
Dexterp37, line 17:
```suggestion
    Glean.initialize("monitor.frontend", true, {
```

I believe we should change this before merging the PR. I think `monitor.frontend` might be an appropriate choice (it follows the same naming pattern as [FxA](https://dictionary.telemetry.mozilla.org/apps/accounts_frontend?itemType=app_ids&page=1)). @rhelmer thoughts?

src/telemetry/metrics.yaml
```diff
@@ -0,0 +1,30 @@
...
+      - https://bugzilla.mozilla.org/show_bug.cgi?id=1691365#c0
+    data_sensitivity:
+      - technical
```
Dexterp37, line 23: Doesn't matter too much, but this is not "technical", it's "interaction" (see https://wiki.mozilla.org/Data_Collection)

src/telemetry/metrics.yaml
```diff
@@ -0,0 +1,30 @@
...
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+# For more information on configuring this file:
+# https://mozilla.github.io/glean/book/reference/yaml/metrics.html
+
+---
+$schema: moz://mozilla.org/schemas/glean/metrics/2-0-0
+
+app_events:
+  page_view:
```
Dexterp37, line 12:
```suggestion
page:
  view:
```

src/telemetry/metrics.yaml
```diff
@@ -0,0 +1,30 @@
...
+    type: event
+    description: |
+      A generic page view event with extras.
+    bugs:
+      # TODO: Add bug link
```
Dexterp37, line 17:
```suggestion
      - https://mozilla-hub.atlassian.net/browse/MNTOR-2022
```

src/telemetry/metrics.yaml
```diff
@@ -0,0 +1,30 @@
...
+# https://mozilla.github.io/glean/book/reference/yaml/metrics.html
+
+---
+$schema: moz://mozilla.org/schemas/glean/metrics/2-0-0
+
+app_events:
+  page_view:
+    type: event
+    description: |
+      A generic page view event with extras.
```
Dexterp37, line 15:
```suggestion
      An event triggered when the page is loaded.
```

Dexterp37 (changes requested):

src/app/hooks/useGlean.ts
Dexterp37, line 17: **Important**: depending on the outcome of [this slack conversation](https://mozilla.slack.com/archives/C05ET81EPPW/p1693564326713539?thread_ts=1692727504.942899&cid=C05ET81EPPW), we may need to change the upload enabled parameter, currently fixed to `true`, to something else.

rosahbruno:
> @rhelmer I had to add an additional NPM script for creating the Glean generated files (`npm run build-glean`). How should this be called from CI? Since we aren't checking in the Glean generated files, the `lint` job fails every time since currently when CI runs nothing is generating those files.
> 
> When testing I just updated the NPM `postinstall` script to look like `patch-package && npm run build-glean`. <- This feels a bit hacky, what would be the preferred way to make sure those generated files get created every time we need them?

@rhelmer 
Is there an obvious place to put this I am missing? Looks like for lint I could just [add the command here](https://github.com/rosahbruno/blurts-server/blob/708fd24ab8fb8b1c78388642c0e871bf0098cbf7/.circleci/config.yml).

Would everything else be covered by adding this to the `build` command?

rhelmer:
> > @rhelmer I had to add an additional NPM script for creating the Glean generated files (`npm run build-glean`). How should this be called from CI? Since we aren't checking in the Glean generated files, the `lint` job fails every time since currently when CI runs nothing is generating those files.
> > When testing I just updated the NPM `postinstall` script to look like `patch-package && npm run build-glean`. <- This feels a bit hacky, what would be the preferred way to make sure those generated files get created every time we need them?
> 
> @rhelmer Is there an obvious place to put this I am missing? Looks like for lint I could just [add the command here](https://github.com/rosahbruno/blurts-server/blob/708fd24ab8fb8b1c78388642c0e871bf0098cbf7/.circleci/config.yml).
> 
> Would everything else be covered by adding this to the `build` command?

Hey sorry for not responding, I saw @Dexterp37's response here https://github.com/mozilla/blurts-server/pull/3303#issuecomment-1702510990 does that cover what you are looking for?

rosahbruno:
@rhelmer 

No, it doesn't cover all cases. We are still unable to run this command in the `lint` stage because that executor doesn't have python installed. Putting that command in just the `build` script only covers when we run that command. That will work for CI and making builds, but not necessarily when developing locally since you aren't likely to run that command on your machine.

I am just looking for suggestions from people more familiar with the repo on how they think this should be added before I make new CI steps or update the project set up steps.

Some projects just include as part of their project set up that you run the `npm run build-glean` step. I can add this to the `README` so that locally you run it before anything else. Then we would also just need a new CI step to generate those files.

rosahbruno:
I build and lint steps are now working. The PR itself was working until I pulled in the changes that moved the `Content-Security-Policy` stuff to `middleware.ts` instead of the `next.config.js`. I am unable to run the app and get browser errors after adding those commits.

Vinnl:
@rosahbruno Those changes are mine; could you push a branch with the errors up somewhere, and/or share the errors you're getting here?

rosahbruno:
@Vinnl 
I am seeing a blank screen on load with this in the console.

Screenshot 2023-09-07 at 9 57 36 AM

rhelmer (approved):

PR #3293, opened 2023-08-04
MNTOR-1923: Monitor premium sub fxa event handler

mansaj:
# References: 
Jira: MNTOR-1923

# Description
* On `subscribe`
  * activate OneRep profile
  * start opt-out process 
* On `unsubscribe`
  * deactivate OneRep profile (auto stops opt-out)

# Screenshot (if applicable)

Not applicable.

# How to test

# Checklist (Definition of Done)
- [ ] Localization strings (if needed) have been added.
- [ ] Commits in this PR are minimal and [have descriptive commit messages](https://chris.beams.io/posts/git-commit/).
- [ ] I've added or updated the relevant sections in readme and/or code comments
- [ ] I've added a unit test to test for potential regressions of this bug.
- [ ] Product Owner accepted the User Story (demo of functionality completed) or waived the privilege.
- [ ] All acceptance criteria are met.
- [ ] Jira ticket has been updated (if needed) to match changes made during the development process.
- [ ] Jira ticket has been updated (if needed) with suggestions for QA when this PR is deployed to stage.

PR #3307, opened 2023-08-14
MNTOR-1948/record data broker resolution time

rhelmer:
# References: 
Jira: MNTOR-1948

# Description

This uses a record-keeping approach in Postgres to store broker scans, so we can tell them apart. It works like this:

1. Free `manual` scan is created (one per user, creating 1 row)
2. The `manual` scan is updated when the webhook is received (updating the 1 existing row)
3. `initial` (happens on subscribe) and `monitoring` automated monthly scan creates 1 row each time the webhook is received

From my testing so far, the scan ID returned by (1) is used for the scans in (2), so we can't make `onerep_scan_id` unique. However that table has its own internal primary key, and each row stores so I think that's OK.

# How to test

In order to allow testing on dev/stage/prod and also accommodate local testing, I've implemented the following:

1. Basic `subscribe` and `unsubscribe` pages which activate profiles and initiate the opt-out/removal process
2. When in dev/heroku mode, the `subscribe` additionally page polls the server for the latest scan, and displays + stores the results in Postgres
3. When in non-dev/non-heroku mode, the `subscribe` does not poll the server (webhook should be used)

I'm working on adding better tests for the OneRep function, which is complicated by the fact that it uses the `next/headers` library but our Jest project uses `testEnvironment: "jsdom"`... I may just move the logic to a utility library - I think it will be too annoying to use if we require header to be passed into the (very simple) `onerepFetch()` function, which is simply a wrapper around `fetch`.

### DIFF
--- a/src/app/(proper_react)/redesign/(authenticated)/user/dashboard/fix/data-broker-profiles/automatic-remove/page.tsx
+++ b/src/app/(proper_react)/redesign/(authenticated)/user/dashboard/fix/data-broker-profiles/automatic-remove/page.tsx
@@ -129,7 +129,10 @@ export default function AutomaticRemove() {
                     { price: "X.XX" }
                   )}
             </span>
-            <Button variant="primary">
+            <Button
+              variant="primary"
+              onClick={() => (window.location.href = "../../subscribed")} // TODO replace with final UI
+            >
               {selectedPlanIsYearly
                 ? l10n.getString(
                     "fix-flow-data-broker-profiles-automatic-remove-features-select-plan-yearly-button"
--- a/src/app/(proper_react)/redesign/(authenticated)/user/dashboard/subscribed/page.tsx
+++ b/src/app/(proper_react)/redesign/(authenticated)/user/dashboard/subscribed/page.tsx
@@ -1,0 +1,75 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import { getServerSession } from "next-auth";
+import { getOnerepProfileId } from "../../../../../../../db/tables/subscribers";
+import { authOptions } from "../../../../../../api/utils/auth";
+import {
+  activateProfile,
+  getAllScanResults,
+  isEligibleForPremium,
+  optoutProfile,
+} from "../../../../../../functions/server/onerep";
+import {
+  getLatestOnerepScan,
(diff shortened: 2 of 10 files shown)


rhelmer:
> I'm working on adding better tests for the OneRep React server functions and the webhook, which is complicated by the fact that it uses the `next/headers` library but our Jest project uses `testEnvironment: "jsdom"`... I may just move the logic to a utility library - I think it will be too annoying to use if we require header to be passed into the (very simple) `onerepFetch()` function, which is simply a wrapper around `fetch`.

I took a stab at this, I'm not super happy with it but now the caller must pass the `Session["user"]` object and the country code (from `getCountryCode()` which takes headers as an argument), so the caller feels like a lot:

```
await isEligibleForFreeScan(session.user, getCountryCode(header()));
```

For the other functions in the `onerep` module, I managed to mock the global `fetch` object so we could test the logic in these functions and ensure the `onerepFetch` wrapper works correctly. I considered passing a `fetch`-like object as an argument to the functions but I don't want to make things too clumsy for callers, I think overriding the global `fetch` is not too terrible of a compromise in this case.

mansaj:
Does this cover everything in https://github.com/mozilla/blurts-server/pull/3293?

rhelmer:
> Does this cover everything in #3293?

@mansaj Well it does technically handle this functionality:
```
    On subscribe
        activate OneRep profile
        start opt-out process
    On unsubscribe
        deactivate OneRep profile (auto stops opt-out)
```

But it doesn't use the FxA endpoint, it provides a few SSR pages that implement it directly. This is because for Foxfooding we want to skip subplat. We will eventually need the FxA webhook handler to handle this though when we use subplat for real.

So I think we probably do still want #3293, we don't need it for Foxfooding but we need it for the final end-to-end test and for when the feature is generally available.

mansaj:
> > Does this cover everything in #3293?
> 
> @mansaj Well it does technically handle this functionality:
> 
> ```
>     On subscribe
>         activate OneRep profile
>         start opt-out process
>     On unsubscribe
>         deactivate OneRep profile (auto stops opt-out)
> ```
> 
> But it doesn't use the FxA endpoint, it provides a few SSR pages that implement it directly. This is because for Foxfooding we want to skip subplat. We will eventually need the FxA webhook handler to handle this though when we use subplat for real.
> 
> So I think we probably do still want #3293, we don't need it for Foxfooding but we need it for the final end-to-end test and for when the feature is generally available.

Ah yes, I wasn't sure if you wanted to merge that branch into this one

mansaj (approved):

PR #3310, opened 2023-08-16
MNTOR-1985: Reset filter values and other fixes

flozia:
# References: 
Jira: [MNTOR-1985](https://mozilla-hub.atlassian.net/browse/MNTOR-1985)

# Description

This addresses and fixes the following behaviors:
- apply filter changes when a user resets the filter
- resets the current filter selection if a user did not confirm their changes
- clicking inside the filter does not close the filter


src/app/components/client/ExposuresFilter.tsx
```diff
@@ -186,6 +191,14 @@ export const ExposuresFilter = ({
...
   );
 
+  // Reset current filter state if selection has not been confirmed.
+  useEffect(() => {
+    const shouldResetFilterState = filterValues !== filterState;
+    if (!filterDialogState.isOpen && shouldResetFilterState) {
+      setFilterState(filterValues);
+    }
+  }, [filterDialogState, filterValues, filterState]);
+
```
Vinnl, line 201: This feels wrong to me, since no [side effects](https://en.wikipedia.org/wiki/Side_effect_(computer_science)) happen here. Wouldn't it also work to pass an `onOpenChange` callback to `useOverlayTriggerState`, which calls `setFilterState(filterValues)` whenever the popover closes?

src/app/components/client/ExposuresFilter.tsx
```diff
@@ -86,12 +87,17 @@ export const ExposuresFilter = ({
 
   const dismissButtonRef = useRef<HTMLButtonElement>(null);
   const dismissButtonProps = useButton(
-    { onPress: () => filterDialogState.close() },
+    {
+      onPress: () => {
+        setFilterState(filterValues);
+        setFilterValues(filterValues);
```
Vinnl, line 93: Does this line do anything? (And if so, why?)

```suggestion
```

Vinnl (commented):
The `useEffect` doesn't feel great to me, so I'd like to either see that replaced, or to understand why it's necessary :)

src/app/components/client/ExposuresFilter.tsx
flozia: Thanks for the suggestion — moved to `onOpenChange` in [0689e3b](https://github.com/mozilla/blurts-server/pull/3310/commits/0689e3b9f0d2b6add695327971a197df96b0bfc6).

Vinnl (approved):

src/app/components/client/ExposuresFilter.tsx
```diff
@@ -241,9 +254,16 @@ export const ExposuresFilter = ({
       </div>
       {explainerDialogState.isOpen && explainerDialog}
       {filterDialogState.isOpen && (
-        <Popover state={filterDialogState} triggerRef={filterBtnRef}>
-          <FilterDialog>
-            <div className={styles.exposuresFilterWrapper} {...overlayProps}>
+        <Popover
+          crossOffset={100}
+          isOpen={filterDialogState.isOpen}
```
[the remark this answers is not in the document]
flozia: I remember working around an issue I was not able to reproduce anymore and agreed that we should only pass the state.

PR #3312, opened 2023-08-17
MNTOR-2059: Feature flags add/update edge cases

mansaj:
# References: 
Jira: MNTOR-2059

# Description
Feature flags add/update should guard against some edge cases:
* allow/wait lists should ignore whitespaces
* allow/wait lists should not accept empty strings
* allow/wait lists should not accept extra commas between/before/after array elements

# Screenshot (if applicable)

Not applicable.

# How to test
`/redesign/admin/feature-flags`

### DIFF
--- a/src/db/tables/featureFlags.ts
+++ b/src/db/tables/featureFlags.ts
@@ -42,8 +42,8 @@ async function addFeatureFlag(flag: FeatureFlag) {
     is_enabled: flag.isEnabled,
     description: flag.description,
     dependencies: flag.dependencies,
-    allow_list: flag.allowList,
-    wait_list: flag.waitList,
+    allow_list: flag.allowList?.map((email) => email.trim()).filter((e) => !!e),
+    wait_list: flag.waitList?.map((email) => email.trim()).filter((e) => !!e),
     expired_at: flag.expiredAt,
     owner: flag.owner,
   };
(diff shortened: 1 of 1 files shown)


src/db/tables/featureFlags.ts
```diff
@@ -42,8 +42,8 @@ async function addFeatureFlag(flag: FeatureFlag) {
     is_enabled: flag.isEnabled,
     description: flag.description,
     dependencies: flag.dependencies,
-    allow_list: flag.allowList,
-    wait_list: flag.waitList,
+    allow_list: flag.allowList?.map((e) => e.trim()).filter((e) => !!e),
+    wait_list: flag.waitList?.map((e) => e.trim()).filter((e) => !!e),
```
pdehaan, line 46: Q: is there a way/need to merge `.map().filter()` to reduce looping over arrays? Or is the data set so small we don't really need to worry?

But wasn't sure if something like this would work:
```js
flag.allowList?.filter((e) => !!e.trim()),
```

or a slightly uglier [and untested] `.reduce()`:
```js
flag.allowList.reduce((acc, e) => {
  e = e.trim();
  return !!e ? [...acc, e] : acc;
}, []);

mansaj: Thanks @pdehaan,  great suggestion

mansaj: actually, sorry, the filter one won't work because we need the trim() to apply to the emails

mansaj: The reduce function actually returns a string, we need string[]

mansaj: K I've modified the reduce, I think that's the right approach

rhelmer (approved):

PR #3304, opened 2023-08-11
MNTOR-2018: More tests

mansaj:
# References: 
Jira: MNTOR-2018

# Description
Better tests

# Screenshot (if applicable)

Not applicable.

# How to test
`npm test`

# Checklist (Definition of Done)
- [ ] Localization strings (if needed) have been added.
- [ ] Commits in this PR are minimal and [have descriptive commit messages](https://chris.beams.io/posts/git-commit/).
- [ ] I've added or updated the relevant sections in readme and/or code comments
- [ ] I've added a unit test to test for potential regressions of this bug.
- [ ] Product Owner accepted the User Story (demo of functionality completed) or waived the privilege.
- [ ] All acceptance criteria are met.
- [ ] Jira ticket has been updated (if needed) to match changes made during the development process.
- [ ] Jira ticket has been updated (if needed) with suggestions for QA when this PR is deployed to stage.

PR #3316, opened 2023-08-17
Ensure tests can run in isolation

Vinnl:
The primary change here is to change `resetMocks` and `resetModules` to `true`, to ensure each unit test can run independently. To do so, I did have to update the tests from https://github.com/mozilla/blurts-server/pull/3304 to define the mocks each specific test depended on inside that test, rather than in a `beforeAll` call that set them up in the precise order that the tests had to run in as well.

I wasn't entirely sure from the test descriptions what behaviour they were verifying, so I updated the descriptions to specify what I _though_ they did. Let me know if they're wrong.

### DIFF
--- a/jest.config.cjs
+++ b/jest.config.cjs
@@ -225,12 +225,12 @@ const customJestConfig = {
       statements: 68,
     },
     "src/cloud-functions/index.js": {
-      branches: 63,
+      branches: 60,
       functions: 66,
       lines: 60,
       statements: 60,
     },
-    "src/utils/*.{js,ts}": {
+    "src/utils/*.js": {
       branches: 19,
       functions: 0,
       lines: 0,
@@ -319,10 +319,10 @@ const customJestConfig = {
   // reporters: undefined,
 
   // Automatically reset mock state before every test
-  resetMocks: false,
+  resetMocks: true,
 
   // Reset the module registry before running each individual test
-  resetModules: false,
+  resetModules: true,
 
   // A path to a custom resolver
   // resolver: undefined,
--- a/src/utils/subscriberBreaches.ts
+++ b/src/utils/subscriberBreaches.ts
@@ -71,17 +71,7 @@ export async function getSubBreaches(
     );
 
     // breach resolution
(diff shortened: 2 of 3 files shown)


jest.config.cjs
```diff
@@ -225,12 +225,12 @@ const customJestConfig = {
       statements: 68,
     },
     "src/cloud-functions/index.js": {
-      branches: 63,
+      branches: 60,
```
Vinnl, line 228: The annoying thing here is that I don't know why this coverage went down, given that I did not actually touch that file :/ If anyone has any thoughts on that, let me know.

src/utils/subscriberBreaches.ts
```diff
@@ -74,7 +74,13 @@ export async function getSubBreaches(
     const breachResolution = subscriber.breach_resolution
```
mansaj, line 74: Yea I think this is off.. it should just be something like:
`subscriber.breach_resolution?.[email.email] ?? {}`

mansaj: Good catch!

rhelmer (approved):

jest.config.cjs
mansaj: We can always open up the `index.html` in coverage folder to see exactly which branches are no longer being tested. But it's probably due to changes to underlying functions (which I don't see any connections between this file and `index.js`.. weird) that file depends on

mansaj (approved):

jest.config.cjs
Vinnl: Unfortunately I see no changes when eyeballing it, other than the actual number:

Screenshot 2023-08-18 at 16-23-12 Code coverage report for src_cloud-functions_index js
Screenshot 2023-08-18 at 16-23-23 Code coverage report for src_cloud-functions_index js

src/utils/subscriberBreaches.ts
Vinnl: Cool, added in 143426815. Catching stuff like this is what the coverage report is useful for!

PR #3329, opened 2023-08-18
fix: filter on en-US explicitly

rhelmer:
# References: 
Jira: MNTOR-1834

# Description

Our locale filtering is a little too strict now - it only fires if `en` is higher in the browser language settings, the default for Firefox and Chrome are going to be `en-US` followed by `en`.

# Screenshot (if applicable)

image

# How to test

Use new browser profile with en-US build


mansaj (approved):

pdehaan:
Not sure I understand this&hellip; (spoiler alert: I'm not very bright)

So if I live in the US, but my primary/preferred browsing language is Portuguese or Spanish, because I was born in _(insert country that speaks one of those languages, or I'm just trying to learn Portuguese)_, I wouldn't be able to sign up for Premium?
Do we have a precedent for that elsewhere? I thought we generally just went with their physical location via MaxMind geoip and didn't specifically filter based on language. Or if I'm understanding correctly, even if I'm in the US and have my primary language set to en-CA, I still wouldn't get the Premium Experience<sup>(tm)</sup> unless I changed my browser settings to en-US in the first spot.

src/app/(nextjs_migration)/layout.tsx
```diff
@@ -47,7 +47,7 @@ export default async function MigrationLayout({
       {falseDoorFlag &&
         waitlistLink &&
         countryCode.toLowerCase() === "us" &&
-        locale === "en" && <HandleFalseDoorTest link={waitlistLink} />}
+        locale === "en-US" && <HandleFalseDoorTest link={waitlistLink} />}
```
Vinnl, line 50: This covers the locales we currently support (`en` and `en-CA`), and the ones we might support in the future:

```suggestion
        ["en", "en-GB", "en-US", "en-CA"].includes(locale) && <HandleFalseDoorTest link={waitlistLink} />}
```

Vinnl (commented):
I think the better solution here would be to use [the `getLocale` function](https://github.com/mozilla/blurts-server/blob/d4ff67a2e8ea9ae63815f95676b566b24b8b6209/src/app/functions/server/l10n.ts#L23-L25) with `l10nBundles` as an argument - that will return the locale of the `.ftl` files we loaded based on the Accept-Language header - so just checking for `en` and `en-CA` then covers _all_ possible accept languages that resolve to English.

And @pdehaan: AFAIK this PR doesn't actually block people from signing up for Premium (well, other than that _no one_ is able to), it just blocks them from signing up for an English-language notification once Premium is available.

src/app/(nextjs_migration)/layout.tsx
Vinnl: Oh heh, [that was actually the old situation](https://github.com/mozilla/blurts-server/pull/3327/files). So to clarify, the old situation did [already look at the Accept-Language headers](https://github.com/mozilla/blurts-server/blob/d4ff67a2e8ea9ae63815f95676b566b24b8b6209/src/app/functions/server/l10n.ts#L51-L52), so I'm not quite sure why that was changed?

rhelmer:
> Not sure I understand this… (spoiler alert: I'm not very bright)
> 
> So if I live in the US, but my primary/preferred browsing language is Portuguese or Spanish, because I was born in _(insert country that speaks one of those languages, or I'm just trying to learn Portuguese)_, I wouldn't be able to sign up for Premium? Do we have a precedent for that elsewhere? I thought we generally just went with their physical location via MaxMind geoip and didn't specifically filter based on language. Or if I'm understanding correctly, even if I'm in the US and have my primary language set to en-CA, I still wouldn't get the Premium Experience(tm) unless I changed my browser settings to en-US in the first spot.

No, I think this is a good point... and in fact I do thing we should reconsider generally if doing geo and language restrictions are a good enough proxy for citizenship (it's common for American citizens to live abroad and for them to prefer a language other than English, for example).

rhelmer:
> > Not sure I understand this… (spoiler alert: I'm not very bright)
> > So if I live in the US, but my primary/preferred browsing language is Portuguese or Spanish, because I was born in _(insert country that speaks one of those languages, or I'm just trying to learn Portuguese)_, I wouldn't be able to sign up for Premium? Do we have a precedent for that elsewhere? I thought we generally just went with their physical location via MaxMind geoip and didn't specifically filter based on language. Or if I'm understanding correctly, even if I'm in the US and have my primary language set to en-CA, I still wouldn't get the Premium Experience(tm) unless I changed my browser settings to en-US in the first spot.
> 
> No, I think this is a good point... and in fact I do thing we should reconsider generally if doing geo and language restrictions are a good enough proxy for citizenship (it's common for American citizens to live abroad and for them to prefer a language other than English, for example).

In this case though the false door acceptance criteria was already figured out a while back, so I'd prefer to ship as-is and have this conversation, and then possibly open it up (vs. holding off, there's no reason not to start this test even if it's over-filtered IMHO)

PR #3331, opened 2023-08-24
add automation to verify landing page footer redirect links

mozrokafor:
# References: 
Jira: [MNTOR-2066](https://mozilla-hub.atlassian.net/browse/MNTOR-2066)

# Description
Add automation to verify landing page footer redirect links [ref URL](https://testrail.stage.mozaws.net/index.php?/cases/view/2095095)

### DIFF
--- a/src/e2e/pages/landingPage.js
+++ b/src/e2e/pages/landingPage.js
@@ -49,7 +49,7 @@ export class LandingPage {
     return {
       // identify expected URLs
       mozillaLogoUrl: "https://www.mozilla.org",
-      allBreachesUrl: "https://stage.firefoxmonitor.nonprod.cloudops.mozgcp.net/breaches",
+      allBreachesUrl: "/breaches",
       FAQUrl: "/firefox-monitor-faq",
       TermsUrl: "/privacy/firefox-monitor",
       GithubUrl: "https://github.com/mozilla/blurts-server",
--- a/src/e2e/pages/landingPage.js
+++ b/src/e2e/pages/landingPage.js
@@ -12,6 +12,17 @@ export class LandingPage {
     this.howItWorksSec = page.locator('.how-it-works')
     this.questionsAboutSec = page.locator('.top-questions-about-monitor')
     this.seeIfDataBreachSec = page.locator('.see-if-data-breach')
+    this.MozillaFooterLogoImage = page.getByAltText("Mozilla")
+    this.MozillaFooterLogoLink = page.locator('footer a').first()
+    this.AllBreachesLink = page.getByRole('link', { name: 'All Breaches' })
+    this.FAQLink = page.locator('footer a:has-text("faq")')
+    this.TermsLink = page.getByRole('link', { name: 'Terms & Privacy' })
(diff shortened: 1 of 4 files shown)


src/e2e/specs/landing.spec.js
```diff
@@ -8,6 +8,26 @@ test.describe('Landing Page element verification', () => {
   test.beforeEach(async ({ landingPage }) => {
     await landingPage.open()
   })
+  test(' Verify that the site footer options work correctly ', async ({ landingPage }) => {
+    // link to testrail case
+    test.info().annotations.push({
```
mansaj, line 13: Are we supposed to add a testrail case for each of our tests? @mozrokafor

mansaj (approved):
LGTM

src/e2e/utils/helpers.js
```diff
@@ -22,6 +22,10 @@ export const setEnvVariables = async (email) => {
   process.env.E2E_TEST_BASE_URL = ENV_URLS[process.env.E2E_TEST_ENV] || 'https://stage.firefoxmonitor.nonprod.cloudops.mozgcp.net'
 }
 
+export const getBaseUrl = async () => {
+  return ENV_URLS[process.env.E2E_TEST_ENV] || ENV_URLS.local
```
Vinnl, line 26: Probably won't be an issue in this case, but I think it's a good habit to get into to use `??` instead of `||` for non-boolean values, to avoid accidentally matching on falsy non-boolean values.

```suggestion
  return ENV_URLS[process.env.E2E_TEST_ENV] ?? ENV_URLS.local
```

src/e2e/pages/landingPage.js
```diff
@@ -12,6 +12,17 @@ export class LandingPage {
     this.howItWorksSec = page.locator('.how-it-works')
     this.questionsAboutSec = page.locator('.top-questions-about-monitor')
     this.seeIfDataBreachSec = page.locator('.see-if-data-breach')
+    this.MozillaFooterLogoImage = page.getByAltText("Mozilla")
+    this.MozillaFooterLogoLink = page.locator('footer a').first()
```
Vinnl, line 16: This seems to match any link in the footer, not necessarily that Mozilla logo. I think you should be able to use the alt text as the name here as well, assuming that `name` is the accessible name/

Vinnl (approved):
I don't see major issues with this code, but I'm wondering if this is not getting too specific in our end-to-end tests? As in: if this test fails ten times, how often will that be because something actually broke, and how often will that be because some text broke, because CI ran out of memory, or some other error that does not indicate broken functionality? I feel like it might lean too much towards the latter, and thus might not be worth the extra maintenance work it introduces?

src/e2e/pages/landingPage.js
mozrokafor: To answer your concern, Maintenance and Reliability were all considered before these scripts are created. as part of an upcoming ticket, will be add labels to these tests as we'll need to run only the core flows during PRs however, we'll need to automate the automate-ables from the QA regression

PR #3340, opened 2023-08-30
Make coverage report help find untested code

Vinnl:
By lowering coverage thresholds, uncovered code would still stand out in coverage reports, making it hard to find what new code was not covered yet. With this change, existing untested code is marked as skipped for the coverage report, so only new code will be marked as uncovered.

This makes it harder to find which old code is still uncovered, but since it's not particularly likely (nor useful, since the code has been working so far) we'll ever fully cover all that with unit tests, that's a fair trade-off if it means future code will be properly tested.

And they said programming was creative, not menial, work...


Vinnl:
Yeah that's a good point. Ideally we could distinguish between pre-unit test code and new code, and this PR removes that distinction. I think Rob was also going to look at adding an ESLint rule that enforced that, and since I used pretty much the same comment everywhere in this PR, maybe he could add an exception for that comment - or just not run the rule in the previously allowlisted files.

PR #3342, opened 2023-08-30
Remove irrelevant warnings in tests

Vinnl:
I'm not sure about this approach, but together with https://github.com/mozilla/blurts-server/pull/3340 and https://github.com/mozilla/blurts-server/pull/3341, this should ensure our console output when running tests is only related to code we're currently working on - happy to hear suggestions on other methods of achieving that.


jest.setup.ts
```diff
@@ -16,6 +17,11 @@ setProjectAnnotations(
 
 expect.extend(toHaveNoViolations);
 
+// Prevent logs from cluttering up actual problems in our tests:
+failOnConsole({
+  shouldFailOnWarn: true,
```
rhelmer, line 22: 💯

rhelmer (approved):

PR #3346, opened 2023-08-31
Remove debug logs

Vinnl:
These logs contained no further information w.r.t. where they came from or what they were needed for, making it hard to find the logs you *do* need. Presumably, they were used while implementing, rather than added to aid observability.

(The specific trigger for this was the `guidedBreaches` log taking up all my console space, making it very hard to find the log I was looking for - but it was also hard to find where it got logged.)


pdehaan:
/[`no-console`](https://eslint.org/docs/latest/rules/no-console) enters the chat

We could add that rule to root ESLint config and make it an error. That way you fail the build by leaving in `console.*` unless you explicitly have an `eslint-disable-next-line no-console` comment.

Vinnl:
Yes, I'd be in favour of that, but I also left quite a few log statements in for now for which I didn't know their importance, so I probably wouldn't be the right person to add those `disable` comments.

src/scripts/syncBreaches.js
```diff
@@ -99,7 +99,6 @@ if (seen.size !== breaches.length) {
 
   // get
   const result = await getAllBreaches()
-  console.log(result.length)
```
mansaj, line 102: We'd want to keep this console log. It's good to know in the logs how many total breaches we currently have

src/scripts/recencyToBreachId.js
```diff
@@ -72,14 +71,12 @@ do {
 
         // find breach id for the breach
         const breachId = allBreaches.find(b => b.Name === breachName)?.Id
-        console.log({ breachId })
```
mansaj, line 75: This file can be deleted

src/scripts/kube-jobs/recencyToBreachId.js
```diff
@@ -144,14 +143,12 @@ for (const subscriber of subscribersArr) {
 
       // find breach id for the breach
       const breachId = allBreaches.find(b => b.Name === breachName)?.Id
-      console.log({ breachId })
```
mansaj, line 147: This file can be deleted

src/app/(proper_react)/redesign/(authenticated)/user/dashboard/page.tsx
```diff
@@ -9,10 +9,7 @@ import { View } from "./View";
 import { authOptions } from "../../../../../api/utils/auth";
 import { dashboardSummary } from "../../../../../functions/server/dashboard";
 import { getCountryCode } from "../../../../../functions/server/getCountryCode";
-import {
-  getSubscriberBreaches,
-  guidedExperienceBreaches,
-} from "../../../../../functions/server/getUserBreaches";
+import { getSubscriberBreaches } from "../../../../../functions/server/getUserBreaches";
```
mansaj, line 12: It's fine to remove this now.. Though we will need to fetch it somewhere for the guided experience data

src/scripts/recencyToBreachId.js
Vinnl: Wait, you mean the whole `recencyToBreachId.js`?

src/app/(proper_react)/redesign/(authenticated)/user/dashboard/page.tsx
Vinnl: Yes, but right now, this was really crowding my logs, and after removing it complaining about an unused import. We can reimport it when we implement the guided experience :)

src/scripts/syncBreaches.js
Vinnl: 9ce1d2fd9

src/scripts/recencyToBreachId.js
mansaj: Sure, the migration is over

Vinnl: Cool! I wasn't aware what the script did, hence the hesitation :) Removed in 9963b765fc1c5eff0ab582dae07f2f69e6056042.

PR #3349, opened 2023-08-31
Do not prompt ineligible users to start a scan

Vinnl:
# References: 
Jira: MNTOR-2119
Figma: N/A

# Description

The chart would prompt users who hadn't run a scan yet to start a free scan - even those who unfortunately cannot. This fixes that.

# Screenshot (if applicable)

image

# How to test

Set your language to something other than `en-US`, then visit the dashboard with a free user who has not run a scan yet. On `main`, you see the above prompt; in this branch, you shouldn't.

### DIFF
--- a/src/app/(proper_react)/redesign/(authenticated)/user/dashboard/View.tsx
+++ b/src/app/(proper_react)/redesign/(authenticated)/user/dashboard/View.tsx
@@ -40,6 +40,7 @@ export type Props = {
   user: Session["user"];
   userBreaches: SubscriberBreach[];
   userScannedResults: ScanResult[];
+  isEligibleForFreeScan: boolean;
   countryCode?: string;
 };
 
@@ -110,6 +111,8 @@ export const View = (props: Props) => {
       }
     }
 
+    // TODO: Add unit test when changing this code:
+    /* c8 ignore next */
     return exposure.isResolved ? "fixed" : "needAction";
   };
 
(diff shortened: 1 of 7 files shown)


src/app/(proper_react)/redesign/(authenticated)/user/dashboard/View.tsx
```diff
@@ -110,6 +111,8 @@ export const View = (props: Props) => {
       }
     }
 
+    // TODO: Add unit test when changing this code:
+    /* c8 ignore next */
```
Vinnl, line 115: I snuck this in - because we use faker data, sometimes our unit test cases hit this code path, and sometimes they don't, depending on the shape of generated exposures, thereby sometimes not resulting in 100% "coverage". Down the road we should make our unit tests deterministic and consciously test this code path.

src/app/components/client/Chart.tsx
```diff
@@ -190,7 +191,7 @@ export const DoughnutChart = (props: Props) => {
                 ))}
               </tbody>
             </table>
-            {!props.hasRunScan ? prompt : null}
+            {!props.hasRunScan && props.isEligibleForFreeScan ? prompt : null}
```
flozia, line 194: Nit: This could be just:
```suggestion
            {(!props.hasRunScan && props.isEligibleForFreeScan) && prompt}
```

flozia (approved):
Looks good!

PR #3332, opened 2023-08-25
MNTOR-2069-README

mozrokafor:
# References: 
Jira: [MNTOR-2069](https://mozilla-hub.atlassian.net/browse/MNTOR-2069)

# Description
added readme and debug and update snapshots commands to package.json

PR #3351, opened 2023-08-31
Landing page content display verification

mozrokafor:
# References: 
Jira: [MNTOR-2121](https://testrail.stage.mozaws.net/index.php?/cases/view/2095093)

# Description
Test Automation to verify that the Landing Page content is displayed correctly 

[MNTOR-2121]: https://mozilla-hub.atlassian.net/browse/MNTOR-2121?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/e2e/pages/landingPage.js
```diff
@@ -70,9 +86,21 @@ export class LandingPage {
...
+      heroHeaderText: "Find out if your personal information has been compromised",
+      heroDescriptionText: "Stay safe with privacy tools from the makers of ⁨Firefox⁩ that protect you from hackers and companies that publish and sell your personal information. We’ll alert you of any known data breaches, find and remove your exposed info and continually watch for new exposures.",
+      heroEmailInputPlaceholder: "Enter email address",
+    }
+  }
+
   async checkBanner(){
-    if(await this.falseDoorBanner.isVisible()){
-      await this.falseBannerCloseButton.click()
+    await this.page.waitForSelector('//div[starts-with(@class, "FalseDoorBanner_falseDoorTest")]',{
```
pdehaan, line 98: Q: Is there a way to do this without xpath (since I dont think we use them elsewhere in selectors/locators)?

src/e2e/pages/landingPage.js
```diff
@@ -23,6 +26,19 @@ export class LandingPage {
...
+    this.heroHeader = page.locator('.hero div h1').innerText()
+    this.heroDescription = page.locator('.hero p').innerText()
+    this.heroEmailInput = page.locator('#scan-email-address')
+    this.heroCTAButton = page.locator('form button:has-text("Check for breaches")')
+    this.heroImage = page.locator('.hero figure img')
+    this.whyUseMonitorSection = page.locator('.why-use-monitor')
+    this.howItWorksSection = page.locator('.how-it-works')
+    this.yourPrivacyIsSafeWithUsSection = page.locator('.safe-with-us')
+    this.landingFAQSection = page.locator('.top-questions-about-monitor')
+    this.landingFAQSectionButtons = page.locator('.top-questions-about-monitor details summary')
```
pdehaan, line 40: not sure this is used anywhere below in any tests. 🤷

src/e2e/specs/landing.spec.js
```diff
@@ -40,4 +42,83 @@ test.describe('Landing Page element verification', () => {
...
+    // 2. Verify the "Why use Firefox Monitor?" section.
+    await expect(async () => {
+      await expect(await landingPage.whyUseMonitorSection).toHaveScreenshot(
+        `${process.env.E2E_TEST_ENV}-whyUseMonitorSection.png`
+      )
+    }).toPass({
+      timeout: 2000
+    })
+
+    // 3. Verify the "Hereâ€™s how it works" section.
```
pdehaan, line 88:
```suggestion
    // 3. Verify the "Here's how it works" section.
```

src/e2e/specs/landing.spec.js
```diff
@@ -40,4 +42,83 @@ test.describe('Landing Page element verification', () => {
...
+    // 5. Verify the "Top questions about Firefox Monitor" section.
+    await expect(async () => {
+      await expect(await landingPage.landingFAQSection).toHaveScreenshot(
+        `${process.env.E2E_TEST_ENV}-landingFAQSection.png`
+      )
+    }).toPass({
+      timeout: 2000
+    })
+
+    // 6. Verify the "See if youâ€™ve been in a data breach" section.
```
pdehaan, line 115:
```suggestion
    // 6. Verify the "See if you've been in a data breach" section.
```

src/e2e/specs/landing.spec.js-snapshots/local-heroCTAButton-chromium-darwin.png
pdehaan, line 1: Interesting! Were these generated locally on your macOS (assuming based on the `-darwin` suffix)?
Would they pass in CI or do we need to have CI generate it's own CI assets assuming it's running some variant of Linux image?

src/e2e/pages/landingPage.js
mozrokafor: this is a temp banner to my knowledge, this will be removed as soon as that banner is removed.

src/e2e/specs/landing.spec.js-snapshots/local-heroCTAButton-chromium-darwin.png
mozrokafor: its my machine unfortunately

src/e2e/specs/landing.spec.js
```diff
@@ -40,4 +42,83 @@ test.describe('Landing Page element verification', () => {
...
+
+      // verify hero description;
+      expect(await landingPage.heroDescription).toBe(links.heroDescriptionText)
+
+      // verify email field placeholder
+      expect(await landingPage.heroEmailInput.getAttribute('placeholder')).toBe(links.heroEmailInputPlaceholder)
+
+      // verify "Check for breaches" blue button;
+      await expect(await landingPage.heroCTAButton).toHaveScreenshot(
+        `${process.env.E2E_TEST_ENV}-heroCTAButton.png`
```
mansaj, line 68: How are these pngs generated?

src/e2e/specs/landing.spec.js
```diff
@@ -40,4 +42,83 @@ test.describe('Landing Page element verification', () => {
...
+      )
+
+      // verify An appropriate image (fishing pole fishing an email icon from a laptop cluttered by notifications);
+      await expect(await landingPage.heroImage).toHaveScreenshot(
+        `${process.env.E2E_TEST_ENV}-heroImage.png`
+      )
+    }).toPass({
+      timeout: 2000
```
mansaj, line 76: Is there a reason for the timeout to be a const?

src/e2e/specs/landing.spec.js
Vinnl: Yeah, as I mentioned in https://github.com/mozilla/blurts-server/pull/3332#discussion_r1308847119, if we generate these locally, they're bound to break (in fact, they wouldn't even match in the current state, as the CI tests don't run on `darwin`), so I'd expect these to be generated in CI, and there to be docs somewhere describing how to regenerate them (i.e. push a failing test run, and then download the artefacts with the mismatched screenshots).

src/e2e/specs/landing.spec.js
```diff
@@ -40,4 +42,83 @@ test.describe('Landing Page element verification', () => {
       timeout: 2000
     })
   })
+
+  test('Verify that the Landing Page content is displayed correctly', async ({ landingPage }) => {
+    // link to testrail case
+    test.info().annotations.push({
+      type: "testrail",
+      description: "https://testrail.stage.mozaws.net/index.php?/cases/view/2095093",
```
Vinnl, line 50: Oooh this is neat :clap:

src/e2e/specs/landing.spec.js
mozrokafor: the 2 seconds is good to have, centralizing this will be in a follow up pr

src/e2e/specs/landing.spec.js
mozrokafor: +1 , its already on the TODO list figure there's no rush especially when a big visual change is coming

src/e2e/pages/landingPage.js
```diff
@@ -72,9 +87,21 @@ export class LandingPage {
     await this.firefoxLogo.click()
   }
 
+  async heroSectionTexts(){
+    return {
+      heroHeaderText: "Find out if your personal information has been compromised",
+      heroDescriptionText: "Stay safe with privacy tools from the makers of Firefox that protect you from hackers and companies that publish and sell your personal information. We'll alert you of any known data breaches, find and remove your exposed info and continually watch for new exposures.",
```
pdehaan, line 93: Unrelated your PR, but now I'm wondering if the .ftls are using the wrong straight quotes for "We'll". 👀

pdehaan (commented):
Re: snapshots,
I see 7 snapshots, but for "local-" only and only on Chromium+Darwin. I think I was just looking at your #3539 PR and it had baselines for `stage-` and `prod-` (but no `local-`) but also for the union of `firefox-` and `linux`.
So not sure if these 7 snapshots need to be 14 (or 21) if we include stage+prod envs, and then multiply by 2 for "-firefox-" as well as "-chromium-" (then add another 14 in the future when we add Safari).

src/e2e/pages/landingPage.js
Vinnl, line 1: Shouldn't this be a TS file now? (and the spec as well?)

src/e2e/pages/landingPage.js
```diff
@@ -25,6 +28,18 @@ export class LandingPage {
     this.falseBannerCloseButton = page.locator('#close-button')
     this.scanEmailAddressInput = page.locator('#scan-email-address')
     this.checkForBreachesButton =  page.getByRole('button', {name: 'Check for breaches'})
+
+    // hero section
```
Vinnl, line 32: Just food for thought: with all these selectors, is this page object even useful? Doesn't seem like they get re-used often, in which case it's just another layer of indirection.

src/e2e/pages/landingPage.js
```diff
@@ -25,6 +28,18 @@ export class LandingPage {
...
+    this.heroHeader = page.locator('.hero div h1').innerText()
+    this.heroDescription = page.locator('.hero p').innerText()
```
Vinnl, line 34: Don't page objects generally just contain the selectors and possibly methods to interact with the page - but if tests want to view the contents of an element, they call `.innerText()` themselves?

src/e2e/pages/landingPage.js
```diff
@@ -72,9 +87,21 @@ export class LandingPage {
     await this.firefoxLogo.click()
   }
 
+  async heroSectionTexts(){
```
Vinnl, line 90: This looks like something that's more appropriate in a test, rather than in the page object? In any case, there's no reason for the method to be `async`.

src/e2e/specs/landing.spec.js-snapshots/local-heroCTAButton-chromium-darwin.png
Vinnl, line 1: Re this and other `-darwin` files: I think it's best to only include files in the repo that are actually used in CI and thus can be verified by anyone, and for which regressions can actually be checked. Otherwise they're bound to get out-of-date and just take up space/be confusing when opened.

Vinnl (commented):
I still don't really think committing locally-generated screenshots is very useful at best and is misleading at worst, so I'm not sure if merging this PR at this point in time is an improvement, unfortunately :/

mozrokafor:
updating with new landing page

PR #3357, opened 2023-09-04
Remove `unsafe-inline` from the script-src CSP

Vinnl:
As per
https://github.com/vercel/next.js/discussions/51039#discussioncomment-6889996, this follows the new documentation at
https://nextjs.org/docs/app/building-your-application/configuring/content-security-policy#adding-a-nonce-with-middleware.

Note that this does not remove the `unsafe-inline` from the `style-src` directive yet; that was already present before the Next.js migration, and is less risky than script-src, I think. We can remove that once we've removed all old code with the associated inline <style> elements.

I pushed this to dev for a bit, and it appeared to work well enough there - though worth keeping an eye on after merging. Also, the page breaks harder if an extension, such as Relay, tries to inject a script into the page (or at least, I think that's what the problem is).


rhelmer (approved):

Vinnl:
Whoops, accidentally left the "Do not merge" label on there from when I made a mistake when testing.

PR #3359, opened 2023-09-05
Split ExposureCard into Breach and Broker cards

Vinnl:
# References: 
Jira: MNTOR-1694
Figma: N/A

# Description

This is the first step in adding exposure cards to the guided resolution flow. It's mostly a refactoring that simplifies including `<ExposureCard>` in other components (currently the dashboard, but in a followup PR also the guided resolution flow), which should now no longer have to pass in different properties depending on whether it's a ScanResult or a SubscriberBreach - all that is now handled by the ExposureCard.

To do so, the ExposureCard now just first determines whether it's been given a ScanResult or a SubscriberBreach, and then delegates to type-specific components that deal just with that type. In essence, those two components are copies of the former ExposureCard, but with every `if` statement that dealt with the different types removed.

I did also make a couple of changes that made it re-use more of our existing code, specifically for date parsing, and for rendering a fallback icon (a circle with a single letter). For the latter I did copy over the list of colours @codemist had added before, since it was more extensive than the list the `<BreachLogo>` component used - thus, the existing website should also start showing those colours.

And finally, I made the mock data more deterministic, to ensure that the Storybook and tests see at least one example of every state an exposure (whether breach or broker) can be in.

# How to test

There should be no visible behaviour changes, other than that the production website should have a wider palette of colours for fallback icons.

- [ ] Product Owner accepted the User Story (demo of functionality completed) or waived the privilege. N/A


src/app/(proper_react)/redesign/(authenticated)/user/dashboard/View.tsx
```diff
@@ -128,34 +103,17 @@ export const View = (props: Props) => {
   const filteredExposures = filterExposures(tabSpecificExposures, filters);
 
   const exposureCardElems = filteredExposures.map((exposure: Exposure) => {
-    const status = getExposureStatus(exposure);
-
-    return isScanResult(exposure) ? (
-      // Scanned result
-      <li key={`scan-${exposure.id}`} className={styles.exposureListItem}>
-        <ExposureCard
```
Vinnl, line 136: Looks, there's just one `<ExposureCard>` left, and it takes fewer props :)

src/app/components/client/ExposureCard.tsx
```diff
@@ -70,23 +50,254 @@ type BreachExposureCategoryProps = {
...
+  );
+};
+
+export type ScanResultCardProps = {
+  exposureImg?: StaticImageData;
+  scanResult: ScanResult;
+  locale: string;
+  isPremiumBrokerRemovalEnabled: boolean;
+};
+const ScanResultCard = (props: ScanResultCardProps) => {
```
Vinnl, line 67: Most of this "new" code is a verbatim copy from the old `<ExposureCard>` component, but only the code branches related to scan results.

src/app/components/client/ExposureCard.tsx
```diff
@@ -70,23 +50,254 @@ type BreachExposureCategoryProps = {
...
-    locale,
-    color,
-    featureFlagsEnabled,
-  } = props;
+export type SubscriberBreachCardProps = {
+  exposureImg?: StaticImageData;
+  subscriberBreach: SubscriberBreach;
+  locale: string;
+};
+const SubscriberBreachCard = (props: SubscriberBreachCardProps) => {
```
Vinnl, line 299: Most of this component is also a verbatim copy of the old `<ExposureCard>` component, but only the branches relating to subscriber breaches.

src/app/components/server/BreachLogo.module.scss
```diff
@@ -0,0 +1,112 @@
...
+span.breachLogo {
+  line-height: 1.75rem;
+  width: 1.75rem;
+  border-radius: 1.75rem;
+  font-size: 1rem;
+  font-weight: bold;
+  text-transform: uppercase;
+  text-align: center;
+
+  &.purple40 {
```
Vinnl, line 21: This file is "new" because I changed it from `.css` to `.scss`, but the `&.<colour>` classes are what's actually new, incorporating the colours that were first confined to `getRandomLightNebulaColor` in the dashboard page.

src/utils/parse.js
```diff
@@ -40,9 +40,11 @@ export function parseIso8601Datetime (datetime) {
   // > Support for ISO 8601 formats differs in that date-only strings
   // > (e.g. "1970-01-01") are treated as UTC, not local.
   // https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/Date#parameters
-  try {
-    return new Date(datetime)
-  } catch (_e) {
-    return null
```
Vinnl, line 46: Well would you look at that, my unit test actually found a bug :) Turns out the Date constructor doesn't throw an error if it receives an unparseable date.

src/app/components/client/ExposureCard.tsx
```diff
@@ -316,26 +429,30 @@ export const ExposureCard = (props: ExposureCardProps) => {
...
             </dd>
             <dt className={styles.visuallyHidden}>
               {l10n.getString("exposure-card-date-found")}
             </dt>
             <dd className={styles.hideOnMobile}>
-              {dateFormatter.format(props.dateFound)}
+              {dateFormatter.format(
+                // We should be able to result that HIBP's `addedDate` property is a properly-formatted data string
+                // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+                parseIso8601Datetime(subscriberBreach.addedDate)!
```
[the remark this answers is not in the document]
Vinnl: Heh, that was a fun one, well spotted. So apparently, `SubscriberBreach.addedDate`, at least in this case, is actually a `Date` object rather than an ISO 8601 date string (and likewise for `modifiedDate` and `breachDate`). The previous date formatted just passed it into the `Date` constructor, which then just returned that same date, but `parseIso8601DatetimeString` actually checks if it's a string, and returns `null` if not.

So I just submitted https://github.com/mozilla/blurts-server/pull/3369 that should ensure it is consistently a `Date` object. Once that is merged, I can update this PR to not call `parseIso8601DatetimeString` for `SubscriberBreach`es. (And TypeScript will prevent me from merging this until I do, so I won't forget :) )

src/app/(proper_react)/redesign/(authenticated)/user/dashboard/View.tsx
```diff
@@ -128,34 +103,17 @@ export const View = (props: Props) => {
...
-          color={getRandomLightNebulaColor(exposure.data_broker)}
-          featureFlagsEnabled={props.featureFlagsEnabled}
-        />
-      </li>
-    ) : (
-      // Breaches result
-      <li key={`breach-${exposure.id}`} className={styles.exposureListItem}>
+    return (
+      <li
+        key={`${isScanResult(exposure) ? "scan" : "breach"}-${exposure.id}`}
```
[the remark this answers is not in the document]
Vinnl: The ternary condition just ensures that the prefix is either `scan` or `breach`, which is then followed by the `.id` property of the exposure, regardless of whether it's a scan or a breach. So yes, both scans and breaches have IDs.

src/utils/parse.test.ts
```diff
@@ -0,0 +1,53 @@
...
+        "This is more than 16 characters and therefore unlikely to be a phone number."
+      )
+    ).toBeNull();
+  });
+
+  it("returns null if given something that does not start with a +", () => {
+    expect(parseE164PhoneNumber("0031612345678")).toBeNull();
+  });
+
+  it("returns null if given something that does not consit of a + followed by numbers", () => {
```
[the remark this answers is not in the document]
Vinnl: Thanks! c9ebaf8f6

PR #3367, opened 2023-09-07
Make Knex find the type definitions for OnerepScan

Vinnl:
Although the type definition existed, it wasn't linked to the "onerep_scans" table name. With this change, that table name is linked to the table name, and some columns are marked as required or prohibited for INSERT and UPDATE queries. This means that you get autocompletion for arguments to method calls after `connection("onerep_scans")`.

(The same already exists for the other tables.)


rhelmer (approved):
Nice catch, thanks!

PR #3368, opened 2023-09-07
Remove `buttonType` prop from <Button>

Vinnl:
If a button has a link target, it should always be a link, so no need for the caller to think about that explicitly.

One potential future improvement is to use an `<a>` if `href` doesn't start with `/`, but I'm not sure if we'll ever use `<Button>` for external links (maybe to FxA?).


flozia (approved):
Makes a lot of sense — thanks for the change!

PR #3371, opened 2023-09-07
var setting

mozrokafor:
Should fix issue with playwright tests


Vinnl (changes requested):
See https://mozilla.slack.com/archives/CEV7MGSQH/p1694091554045079?thread_ts=1694085902.899519&cid=CEV7MGSQH - I don't think that this is the issue. (In reality, it's three separate issues that all manifest in the same way.)

mozrokafor:
not needed anymore

PR #3372, opened 2023-09-07
Explicitly verify in CI that the build succeeds

Vinnl:
Before this, the only place the build happened was before running the Playwright tests (and in the Docker image), which obscured the root cause if build failures happen.

Since the primary failure is type checking failing, which in essence is a form of linting, I added it there, but happy to move it elsewhere.


mozrokafor (approved):
LGTM

mansaj (approved):

PR #3380, opened 2023-09-08
MNTOR-2120 - re enable breach alerts with pubsub

rhelmer:
# References: 
Jira: MNTOR-2120

# Description

Re-enable email breach alerts using GCP PubSub for queue, and cron job for processing results (to be run hourly).

See https://github.com/mozilla/blurts-server/pull/2976 for rationale behind this approach.

# How to test

See README changes for running local emulator, and set env vars:
```
GCP_PUBSUB_PROJECT_ID
GCP_PUBSUB_TOPIC_NAME
GCP_PUBSUB_SUBSCRIPTION_NAME
```


src/app/api/v1/hibp/notify/route.ts
```diff
@@ -0,0 +1,63 @@
...
+    return NextResponse.json({ success: "false" }, { status: 401 });
+  }
+
+  const json = await req.json();
+
+  if (!(json.breachName && json.hashPrefix && json.hashSuffixes)) {
+    console.error(
+      "HIBP breach notification: requires breachName, hashPrefix, and hashSuffixes."
+    );
+    return NextResponse.json({ success: "false" }, { status: 500 });
```
mansaj, line 32: Nit: prob a 400 instead of 500

mansaj, line 32: Nit: user request error should be 400?

mansaj (approved):
Didn’t test it but the code lgtm 


Also pending unit test fixes

src/scripts/emailBreachAlerts.js
```diff
@@ -0,0 +1,260 @@
...
+});
+
+const checkInId = Sentry.captureCheckIn({
+  monitorSlug: SENTRY_SLUG,
+  status: "in_progress",
+});
+
+// Only process this many messages before exiting.
+const maxMessages = 1000;
+const projectId = process.env.GCP_PUBSUB_PROJECT_ID;
```
mansaj, line 46: These env vars can be added to the env file?

src/scripts/emailBreachAlerts.js
```diff
@@ -0,0 +1,260 @@
...
+      });
+    })
+    .catch((err) => console.error(err))
+    .finally(async() => {
+      // Tear down knex connection pools
+      await knexSubscribers.destroy();
+      await knexEmailAddresses.destroy();
+      await knexHibp.destroy();
+    });
+}
```
pdehaan: nit: add newline

package.json
```diff
@@ -80,6 +83,7 @@
     "react-cookie": "^4.1.1",
     "react-dom": "^18.2.0",
     "react-stately": "^3.23.0",
+    "request": "^2.88.2",
```
pdehaan, line 86: Where is this used? I'm not seeing an import for it ([plus it's a really old, deprecated package](https://www.npmjs.com/package/request))

package.json
```diff
@@ -63,6 +65,7 @@
     "client-oauth2": "^4.3.3",
     "csrf-csrf": "^2.2.2",
     "dotenv": "^16.0.3",
+    "encoding": "^0.1.13",
```
pdehaan, line 68: Not sure where this is used, unless it's some side effect of the previous GCP dependencies above and this is bumping their version numbers.

src/app/api/v1/hibp/notify/route.ts
```diff
@@ -0,0 +1,63 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import { NextRequest, NextResponse } from "next/server";
+import { bearerToken } from "../../../utils/auth";
+
+import { PubSub } from "@google-cloud/pubsub";
+
+const projectId = "rhelmer-monitor-local-dev";
```
pdehaan, line 10: Is this for everybody, or just your local env?

src/app/api/v1/hibp/notify/route.ts
pdehaan: Same Q as above, re: `"false"` vs `false`.

src/app/api/v1/hibp/notify/route.ts
```diff
@@ -0,0 +1,63 @@
...
+    if (process.env.NODE_ENV === "development") {
+      try {
+        await pubsub.createTopic(topicName);
+        await pubsub.topic(topicName).createSubscription(subscriptionName);
+      } catch (ex) {
+        console.debug(ex);
+      }
+    } else {
+      console.error("Topic not found:", topicName);
+      return NextResponse.json({ success: "false" }, { status: 500 });
```
pdehaan, line 52: Use Boolean?
```suggestion
      return NextResponse.json({ success: false }, { status: 500 });
```

src/app/api/v1/hibp/notify/route.ts
```diff
@@ -0,0 +1,63 @@
...
+    }
+  }
+
+  try {
+    await topic.publishMessage({ json });
+    return NextResponse.json({ success: true }, { status: 200 });
+  } catch (ex) {
+    console.error("Error queueing HIBP breach:", ex);
+    return NextResponse.json({ success: "false" }, { status: 500 });
```
pdehaan, line 61: Use Boolean?
```suggestion
    return NextResponse.json({ success: false }, { status: 500 });
```

src/scripts/emailBreachAlerts.js
pdehaan, line 1: Not sure if we have a policy on files being TypeScript vs JavaScript.

src/scripts/emailBreachAlerts.js
```diff
@@ -0,0 +1,260 @@
...
+  getAllBreachesFromDb,
+  knexHibp
+} from "../utils/hibp.js";
+
+const SENTRY_SLUG = "cron-breach-alerts";
+
+Sentry.init({
+  environment: process.env.APP_ENV,
+  dsn: process.env.SENTRY_DSN,
+  tracesSampleRate: 1.0,
```
pdehaan, line 36: Q: should this be an ENV var so we can tweak it in Heroku/dev envs?

src/app/api/v1/hibp/notify/route.ts
```diff
@@ -0,0 +1,63 @@
...
+    return NextResponse.json({ success: "false" }, { status: 401 });
+  }
+
+  const json = await req.json();
+
+  if (!(json.breachName && json.hashPrefix && json.hashSuffixes)) {
+    console.error(
+      "HIBP breach notification: requires breachName, hashPrefix, and hashSuffixes."
+    );
+    return NextResponse.json({ success: "false" }, { status: 400 });
```
pdehaan, line 32: Use Boolean?
```suggestion
    return NextResponse.json({ success: false }, { status: 400 });
```

src/app/api/v1/hibp/notify/route.ts
```diff
@@ -0,0 +1,63 @@
...
+/**
+ * Whenever a breach is detected on the HIBP side, HIBP sends a request to this endpoint.
+ * The payload is checked for validity, and immediately queued if it is valid.
+ *
+ * @param req
+ */
+export async function POST(req: NextRequest) {
+  const headerToken = bearerToken(req);
+  if (headerToken !== process.env.HIBP_NOTIFY_TOKEN) {
+    return NextResponse.json({ success: "false" }, { status: 401 });
```
pdehaan, line 23: Use Boolean?
```suggestion
    return NextResponse.json({ success: false }, { status: 401 });
```

<details>
<summary><kbd>git grep -Pho "success: (.*?)\s" | sort | uniq -c</kbd></summary>

```
   5 success: "false"
  23 success: false
   6 success: true
   8 success: true,
```
</details>

src/app/api/v1/hibp/notify/route.ts
rhelmer: Oops, I meant to change this to use the env vars, not just the subscriber. Thanks!

package.json
rhelmer: I'm pretty sure this and the `encoding` one below came up while testing the local GCP pubsub emulator, but I will double-check. If that's true (and it's not needed for production) I will move these to dev dependencies, but I'm not totally sure yet.

src/scripts/emailBreachAlerts.js
rhelmer: I think in this case we want 100%, because it runs hourly and it's under our control (there's only going to be a single script running at a time).

I'd like to eventually move the subscriber out of cron and into e.g. a pool of HTTP workers, we will probably need to reconsider.

src/scripts/emailBreachAlerts.js
rhelmer: Typescript in general, but since this is a cron job I want to make it easy for SRE to support (all of the existing cron jobs are JS). I would like to do this though, just as a separate step, filed https://mozilla-hub.atlassian.net/browse/MNTOR-2142 Thanks!

package.json
rhelmer: OK re-tested and they don't seem to be needed, was probably from some local debugging rabbit hole, good catch! Removed

.env-dist
```diff
@@ -145,3 +145,7 @@ STATS_TOKEN=
 
 # Support article link
 NEXT_PUBLIC_MONITOR_SUPPORT_URL=https://support.mozilla.org/kb/firefox-monitor
+
+# GCP PubSub Project ID and subscription name
+GCP_PUBSUB_PROJECT_ID=
+GCP_PUBSUB_SUBSCRIPTION_NAME=hibp-breaches-sub
```
pdehaan, line 151: From some random code below, but do we need to add `GCP_PUBSUB_TOPIC_NAME` to the .env-dist file too?
```js
const projectId = process.env.GCP_PUBSUB_PROJECT_ID;
const topicName = process.env.GCP_PUBSUB_TOPIC_NAME;
const subscriptionName = process.env.GCP_PUBSUB_SUBSCRIPTION_NAME;
```

rhelmer: Yes, I noticed that after landing, will file a small followup. The code detects if any of these aren't set and error out with a specific message, but I think having it in the `.env-dist` is still the right thing to do.

rhelmer: https://github.com/mozilla/blurts-server/pull/3387/

PR #3383, opened 2023-09-08
Store scan results in a separate table

Vinnl:
As discussed on Slack, this removes the scan results column that contains a JSON blob from the `onerep_scans` table, and instead creates a new `onerep_scan_results` table that holds the results - this will enable us to allow the user to resolve individual scan results from a scan. In a followup PR, I'll extend that table to add a boolean `manually_resolved` column, and an API to toggle that.

Note: this builds on https://github.com/mozilla/blurts-server/pull/3367 and https://github.com/mozilla/blurts-server/pull/3373, so merging those first should allow shrinking this PR a bit.

### DIFF
--- a/src/apiMocks/mockData.ts
+++ b/src/apiMocks/mockData.ts
@@ -3,7 +3,7 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
 import { faker } from "@faker-js/faker";
-import { ScanResult } from "../app/functions/server/onerep";
+import { OnerepScanResultRow } from "knex/types/tables";
 import {
   RemovalStatus,
   RemovalStatusMap,
@@ -24,8 +24,7 @@ import {
 const fakerSeed = 123;
 
 // This is a full list of scan results, all pages, and would normally be
-// stored in the `onerep_scan_results.onerep_scan_results` column
-// as `jsonb`.
+// stored in the `onerep_scan_results` table.
 export function mockedOneRepScanResults() {
   faker.seed(fakerSeed);
   return {
@@ -43,17 +42,20 @@ export type RandomScanOptions = Partial<{
  * Generates scan result with randomly-generated mock data.
  *
  * @param options
- * @returns {ScanResult} - A single scan result.
+ * @returns A single scan result.
  */
-export function createRandomScan(options: RandomScanOptions = {}): ScanResult {
+export function createRandomScan(
+  options: RandomScanOptions = {}
+): OnerepScanResultRow {
   faker.seed(options.fakerSeed);
   return {
     id: faker.number.int(),
-    profile_id: faker.number.int(),
+    onerep_scan_result_id: faker.number.int(),
+    onerep_scan_id: faker.number.int(),
     first_name: faker.person.firstName(),
     last_name: faker.person.lastName(),
     middle_name: faker.person.middleName(),
-    age: faker.number.int({ min: 14, max: 120 }).toString(),
+    age: faker.number.int({ min: 14, max: 120 }),
     status:
       options.status ??
       (faker.helpers.arrayElement(
@@ -71,11 +73,8 @@ export function createRandomScan(options: RandomScanOptions = {}): ScanResult {
     link: faker.internet.url(),
     data_broker: faker.internet.domainName(),
     data_broker_id: faker.number.int(),
-    created_at:
-      options.createdDate?.toISOString() ??
-      faker.date.recent({ days: 2 }).toISOString(),
-    updated_at: faker.date.recent({ days: 1 }).toISOString(),
-    url: faker.internet.url(),
+    created_at: options.createdDate ?? faker.date.recent({ days: 2 }),
+    updated_at: faker.date.recent({ days: 1 }),
   };
 }
 
--- a/src/db/migrations/20230907143204_add_onerep_scanresults_table.js
+++ b/src/db/migrations/20230907143204_add_onerep_scanresults_table.js
@@ -61,7 +61,10 @@ export async function up(knex) {
 
   if (scanResultRows.length > 0) {
     await knex("onerep_scan_results")
-      .insert(scanResultRows);
+      .insert(scanResultRows)
+      // On Heroku, we have multiple rows with the same `onerep_scan_id`:
+      .onConflict("onerep_scan_result_id")
+      .ignore();
   }
 
 
--- a/src/db/migrations/20230907143204_add_onerep_scanresults_table.js
+++ b/src/db/migrations/20230907143204_add_onerep_scanresults_table.js
@@ -1,0 +1,85 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+/**
+ * @param { import("knex").Knex } knex
+ * @returns { Promise<void> }
+ */
+export async function up(knex) {
+  await knex.schema.alterTable("onerep_scans", table => {
+    table.unique("onerep_scan_id");
+  });
+
+  await knex.schema
+    .createTable("onerep_scan_results", table => {
+      table.increments("id").primary();
+      table.integer("onerep_scan_result_id").notNullable();
+      table.integer("onerep_scan_id").references("onerep_scans.onerep_scan_id").notNullable();
+      table.string("link").notNullable();
+      table.integer("age").nullable();
+      table.string("data_broker").notNullable();
+      table.integer("data_broker_id").notNullable();
+      table.jsonb("emails").notNullable();
+      table.jsonb("phones").notNullable();
+      table.jsonb("addresses").notNullable();
+      table.jsonb("relatives").notNullable();
+      table.string("first_name").notNullable();
+      table.string("middle_name").nullable();
+      table.string("last_name").notNullable();
+      table.string("status").notNullable();
+      table.timestamp("created_at").defaultTo(knex.fn.now());
+      table.timestamp("updated_at").defaultTo(knex.fn.now());
+      table.index("onerep_scan_id");
+      table.index("onerep_scan_result_id");
+    });
+
+  const scanRows = await knex("onerep_scans")
+    .select("id", "onerep_scan_id", "onerep_scan_results");
+
+  const scanResultRows = scanRows.map(scan => {
+    return scan.onerep_scan_results.data.map(scanResult => {
+      const rowToInsert = {
+        onerep_scan_result_id: scanResult.id,
+        onerep_scan_id: scan.onerep_scan_id,
+        link: scanResult.link,
+        age: typeof scanResult.age === "string" ? Number.parseInt(scanResult.age, 10) : null,
+        data_broker: scanResult.data_broker,
+        data_broker_id: scanResult.data_broker_id,
+        emails: JSON.stringify(scanResult.emails),
+        phones: JSON.stringify(scanResult.phones),
+        addresses: JSON.stringify(scanResult.addresses),
+        relatives: JSON.stringify(scanResult.relatives),
+        first_name: scanResult.first_name,
+        middle_name: scanResult.middle_name,
+        last_name: scanResult.last_name,
+        status: scanResult.status,
+      };
+      return rowToInsert
+    });
+  }).flat();
+
+  if (scanResultRows.length > 0) {
+    await knex("onerep_scan_results")
+      .insert(scanResultRows);
+  }
+
+
+  await knex.schema.alterTable("onerep_scans", table => {
+    table.dropColumn('onerep_scan_results');
+  });
+}
+
+/**
+ * @param { import("knex").Knex } knex
+ * @returns { Promise<void> }
+ */
+export async function down(knex) {
+  await knex.schema.alterTable("onerep_scans", table => {
+    table.jsonb('onerep_scan_results');
+  });
+  await knex.schema.dropTable("onerep_scan_results")
+  await knex.schema.alterTable("onerep_scans", table => {
+    table.dropUnique("onerep_scan_id");
+  });
+}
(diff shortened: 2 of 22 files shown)


src/db/migrations/20230907143204_add_onerep_scanresults_table.js
```diff
@@ -0,0 +1,85 @@
...
+}
+
+/**
+ * @param { import("knex").Knex } knex
+ * @returns { Promise<void> }
+ */
+export async function down(knex) {
+  await knex.schema.alterTable("onerep_scans", table => {
+    // TODO: Restore values from onerep_scan_results, or fetch anew from OneRep API?
+    //       Probably not worth it though, since we never wrote to that column in production.
```
Vinnl, line 78: Callout: if you migrate and then rollback, you lose existing scan results. I figured that writing the code to preserve it probably wasn't worth it, but let me know if you disagree.

(I also can't _completely_ restore it, as I'm not copying over OneRep's `created_at` and `updated_at` fields.)

src/db/migrations/20230907143204_add_onerep_scanresults_table.js
```diff
@@ -0,0 +1,87 @@
...
+export async function up(knex) {
+  await knex.schema.alterTable("onerep_scans", table => {
+    table.unique("onerep_scan_id");
+  });
+
+  await knex.schema
+    .createTable("onerep_scan_results", table => {
+      table.increments("id").primary();
+      table.integer("onerep_scan_result_id").notNullable();
+      table.integer("onerep_scan_id").references("onerep_scans.onerep_scan_id").notNullable();
```
Vinnl, line 18: One thing I'm not storing here is the `onerep_profile_id`, since that can already be obtained by joining with `onerep_scans`, so this saves us an extra column. I will have to do that join (and then another join to the `subscribers` table) in my follow-up PR. I'm not really sure on the trade-offs here, i.e. whether an extra join in a query outside of the hot path is preferable compared to having a redundant extra integer column.

src/db/tables/onerep_scans.ts
```diff
@@ -72,35 +57,58 @@ async function setOnerepManualScan(
...
+        age:
+          typeof scanResult.age === "string"
+            ? Number.parseInt(scanResult.age, 10)
+            : undefined,
+        data_broker: scanResult.data_broker,
+        data_broker_id: scanResult.data_broker_id,
+        emails: JSON.stringify(scanResult.emails),
+        phones: JSON.stringify(scanResult.phones),
+        addresses: JSON.stringify(scanResult.addresses),
+        relatives: JSON.stringify(scanResult.relatives),
```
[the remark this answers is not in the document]
rhelmer: No, I don't think it's important. IIRC this was just done because that's what the https://docs.onerep.com/#operation/getScanResult structure looks like, but I agree it's redundant in our case if there's nothing except the `data` key that we need.

src/db/tables/onerep_profiles.ts
```diff
@@ -18,7 +19,11 @@ export async function setProfileDetails(
     last_name: profileData.last_name,
     city_name: profileData.addresses[0]["city"],
     state_code: profileData.addresses[0]["state"],
-    date_of_birth: profileData.birth_date,
+    // TODO: Validate input:
```
rhelmer, line 22: Can you link a ticket to this TODO?

src/knex-tables.d.ts
```diff
@@ -148,10 +149,75 @@ declare module "knex/types/tables" {
     id: number;
     onerep_profile_id: number;
     onerep_scan_id: number;
-    onerep_scan_results: ScanResult;
+    onerep_scan_reason: "manual" | "initial" | "monitoring";
```
rhelmer, line 152: Can this be `Scan["reason"]` (from `onerep.ts`)

src/db/migrations/20230907143204_add_onerep_scanresults_table.js
rhelmer: I'm not worried about rolling back (since we haven't put this in front of real users yet. Can we remove the TODO? Or link to a ticket (but I agree it's not worth it so let's just remove the comment)

rhelmer (changes requested):
This looks great overall! Thanks for unpacking that JSON result into columns with types etc, I know that was a ton of work. I can test this locally, just a few minor things in the meantime.

src/db/migrations/20230907143204_add_onerep_scanresults_table.js
Vinnl: Oh cool, yeah I agree: 12acd68f2

src/db/tables/onerep_profiles.ts
Vinnl: Ah yes, good callout: 284e100f2

src/knex-tables.d.ts
Vinnl: Oh nice, yes it should be: dc2c0cf43

src/db/migrations/20230907143204_add_onerep_scanresults_table.js
mansaj: Choosing not to store is the right decision here. This is a classic tradeoff between normalization / performance and joins definitely have a bad rep. However, we should always start with the simplest approach and stick with the core principles of relational database (ACID), and iterate if necessary ONLY when we have metrics that support performance bottlenecks. By duplicating data, we essentially nullifies consistency, which is the C there.

src/db/migrations/20230907143204_add_onerep_scanresults_table.js
```diff
@@ -0,0 +1,85 @@
...
+      table.timestamp("created_at").defaultTo(knex.fn.now());
+      table.timestamp("updated_at").defaultTo(knex.fn.now());
+      table.index("onerep_scan_id");
+      table.index("onerep_scan_result_id");
+    });
+
+  const scanRows = await knex("onerep_scans")
+    .select("id", "onerep_scan_id", "onerep_scan_results");
+
+  const scanResultRows = scanRows.map(scan => {
```
mansaj, line 40: Thanks for doing the migration here as well.
One weird issue we *might* encounter when we run this in heroku... which isn't your fault.. but unfortunately is that we might already have duplicate `onerep_scan_id` in the db.. which will stop this migration. We'll need to check heroku's db and probably manually clean up the data before pushing this to main

src/db/migrations/20230907143204_add_onerep_scanresults_table.js
```diff
@@ -0,0 +1,85 @@
...
+    return scan.onerep_scan_results.data.map(scanResult => {
+      const rowToInsert = {
+        onerep_scan_result_id: scanResult.id,
+        onerep_scan_id: scan.onerep_scan_id,
+        link: scanResult.link,
+        age: typeof scanResult.age === "string" ? Number.parseInt(scanResult.age, 10) : null,
```
mansaj, line 46: So from the test data, we get `null` for all ages, but I wonder if this is going to be problematic when we get some real data that has `age: number`.. and this would nullify that

src/knex-tables.d.ts
```diff
@@ -148,10 +150,75 @@ declare module "knex/types/tables" {
     id: number;
     onerep_profile_id: number;
     onerep_scan_id: number;
-    onerep_scan_results: ScanResult;
+    onerep_scan_reason: Scan["reason"];
     created_at: Date;
     updated_at: Date;
```
mansaj, line 155: I wonder if it's still necessary to have `updated_at` as a col. Seems like an append-only table now

src/knex-tables.d.ts
```diff
@@ -148,10 +150,75 @@ declare module "knex/types/tables" {
...
+      street?: string;
+      zip?: string;
+    }>;
+    relatives: string[];
+    first_name: string;
+    middle_name?: string;
+    last_name: string;
+    status: RemovalStatus;
+    created_at: Date;
+    updated_at: Date;
```
mansaj, line 188: same as above re append-only

src/db/tables/onerep_scans.ts
```diff
@@ -6,20 +6,9 @@ import initKnex from "knex";
...
-): Promise<ScanResult[]> {
-  return (await knex("onerep_scans")
-    .select("onerep_scan_results")
-    .where("onerep_profile_id", onerepProfileId)
-    .andWhere("onerep_scan_id", onerepScanId)) as unknown as Promise<
-    ScanResult[]
-  >;
-}
-
 export interface GetLatestOnerepScanResult {
```
mansaj, line 12: This is now obsolete no?

src/db/tables/onerep_scans.ts
```diff
@@ -28,28 +17,24 @@ export interface GetLatestOnerepScanResult {
...
-    onerep_scan_reason: "manual" | "initial" | "monitoring";
-  }>;
+): Promise<{ scan: OnerepScanRow | null; results: OnerepScanResultRow[] }> {
+  const scan = await knex("onerep_scans")
+    .first()
+    .where("onerep_profile_id", onerepProfileId)
+    .orderBy("created_at", "desc");
+
+  const results =
+    typeof scan === "undefined"
```
mansaj, line 29: nit: is there a diff between `scan ?` vs checking for undefined here?

src/db/tables/onerep_scans.ts
```diff
@@ -72,35 +57,58 @@ async function setOnerepManualScan(
     onerep_profile_id: onerepProfileId,
     onerep_scan_id: onerepScanId,
     onerep_scan_reason: "manual",
+    // @ts-ignore knex.fn.now() results in it being set to a date,
+    // even if it's not typed as a JS date object:
     created_at: knex.fn.now(),
   });
 }
 
 async function setOnerepScanResults(
```
mansaj, line 66: nit: perhaps the word `set` isn't the best.. it usually implies update rather than append / add. 
Same with `setOnerepManualScan`

src/apiMocks/mockData.ts
```diff
@@ -43,17 +42,20 @@ export type RandomScanOptions = Partial<{
...
   return {
     id: faker.number.int(),
-    profile_id: faker.number.int(),
+    onerep_scan_result_id: faker.number.int(),
+    onerep_scan_id: faker.number.int(),
     first_name: faker.person.firstName(),
     last_name: faker.person.lastName(),
     middle_name: faker.person.middleName(),
-    age: faker.number.int({ min: 14, max: 120 }).toString(),
+    age: faker.number.int({ min: 14, max: 120 }),
```
mansaj, line 58: don't we have a check for `=="string"` for this?

src/db/migrations/20230907143204_add_onerep_scanresults_table.js
Vinnl: Hmm, alternatively, I can just drop the latest versions of those? I'm assuming, especially since it's just Heroku, that losing the later versions isn't a big issue? Added in 10e700cd1, but happy to revert if you prefer going into the Heroku DB itself. (But if so, what would you prefer me to do with it?)

src/db/migrations/20230907143204_add_onerep_scanresults_table.js
Vinnl: Yeah, that's fair. For what it's worth, [the OneRep API docs](https://docs.onerep.com/#operation/getScanResults) do say they return a string. I can add another check for `number` if you prefer, though I'm also a bit wary of handling cases that shouldn't happen (and thus can't be tested).

src/knex-tables.d.ts
Vinnl: Hmm, I'm not familiar enough with Scans to be able to tell with certainty. Don't scans go through several states (i.e. [there's](https://docs.onerep.com/#operation/getScan) a `status` property of which one possible value is at least `in_progress`) - or are we only inserting new ScanResults as they come in, and leave the Scan as-is, since we don't seem to be storing that status? /cc @rhelmer

src/knex-tables.d.ts
Vinnl: That's a good point. Removed the column and blocked update queries in the type in ca99aa2c7.

src/db/tables/onerep_scans.ts
Vinnl: Yep, thanks for flagging! 200e2f8f0

src/db/tables/onerep_scans.ts
Vinnl: The former does implicit type casting, so `""` would also be falsy, but in practice, there shouldn't be a difference here. I just default to explicit matching on what I expect out of habit, but if there's a reason to prefer the other version, happy to change.

src/apiMocks/mockData.ts
Vinnl: The OneRep API returns a string, which we check for, and then parse it into an integer to store it in our database. Since this mocks the data returned by the DB, it returns an integer. (Although now that you mention it, maybe there's a reason OneRep kept it a string? In which case, maybe we should keep it a string in our DB as well?)

src/knex-tables.d.ts
Vinnl: Oh actually, I removed this again: 1f4681b34

The entire point of this change was to prepare https://github.com/mozilla/blurts-server/pull/3385, where I _do_ update scan result rows when they are manually resolved :sweat_smile:

rhelmer (approved):

src/db/migrations/20230907143204_add_onerep_scanresults_table.js
mansaj: We can go with the API then, good call

src/apiMocks/mockData.ts
mansaj: Oh true, long day of PR reviewing 😓

src/knex-tables.d.ts
mansaj: Ok that makes sense

src/knex-tables.d.ts
mansaj: We might update these as a part of the webhook events we will be receiving. Right now it's kind of hazy as to what those events will be like (whether it's an update or a new insert). 
So I guess we can leave it the way it is for now.

mansaj (approved):

src/knex-tables.d.ts
Vinnl: Ah OK, I'll leave that for a future followup then, thanks!

PR #3385, opened 2023-09-08
Add API to manually mark scan results as resolved

Vinnl:
# References: 
Jira: MNTOR-2117
Figma: N/A

# Description

Note: this builds on https://github.com/mozilla/blurts-server/pull/3383.

This adds a boolean column `manually_resolved` to the table of scan results, and an API that allows the user to mark that as true.

# How to test

The client doesn't properly call it yet, but you can replace the `letsFixItBtn` inside `<ScanResultCard>` with the following, and then click the "Let's fix it" button on broker cards:

```jsx
  const letsFixItBtn = (
    <span className={styles.fixItBtn}>
      <Button variant={"primary"} onClick={(event) => {
        event.preventDefault();

        void fetch(
          `/api/v1/user/scan-result/${props.scanResult.onerep_scan_result_id}/resolution`,
          {
            method: "POST",
            credentials: "same-origin",
          }
        );
      }}>{l10n.getString("exposure-card-cta")}</Button>
    </span>
  );
```

- [ ] I've added a unit test to test for potential regressions of this bug. - N/A (mostly DB and API code)
- [ ] All acceptance criteria are met. - N/A

### DIFF
--- a/src/app/api/v1/user/scan-result/[onerepScanResultId]/resolution/route.ts
+++ b/src/app/api/v1/user/scan-result/[onerepScanResultId]/resolution/route.ts
@@ -22,7 +22,7 @@ export async function POST(
   { params }: { params: { onerepScanResultId: string } }
 ): Promise<NextResponse<ResolveScanResultResponse>> {
   const session = await getServerSession(authOptions);
-  if (!session || !session.user || !session.user.subscriber) {
+  if (!session?.user?.subscriber) {
     return new NextResponse<ResolveScanResultResponse>(
       JSON.stringify({ success: false, message: "Unauthenticated" }),
       { status: 401 }
--- a/src/app/api/v1/user/scan-result/[onerepScanResultId]/resolution/route.ts
+++ b/src/app/api/v1/user/scan-result/[onerepScanResultId]/resolution/route.ts
@@ -1,0 +1,67 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import { getServerSession } from "next-auth";
+import { NextRequest, NextResponse } from "next/server";
+
+import { authOptions } from "../../../../../utils/auth";
+import {
+  isOnerepScanResultForSubscriber,
+  markOnerepScanResultAsResolved,
+} from "../../../../../../../db/tables/onerep_scans";
+
+export type ResolveScanResultResponse =
+  | {
+      success: true;
+    }
+  | { success: false; message: string };
+
+export async function POST(
+  req: NextRequest,
+  { params }: { params: { onerepScanResultId: string } }
+): Promise<NextResponse<ResolveScanResultResponse>> {
+  const session = await getServerSession(authOptions);
+  if (!session || !session.user || !session.user.subscriber) {
+    return new NextResponse<ResolveScanResultResponse>(
+      JSON.stringify({ success: false, message: "Unauthenticated" }),
+      { status: 401 }
+    );
+  }
+
+  const scanResultId = Number.parseInt(params.onerepScanResultId, 10);
+  if (typeof scanResultId !== "number" || Number.isNaN(scanResultId)) {
+    return new NextResponse<ResolveScanResultResponse>(
+      JSON.stringify({ success: false, message: "Invalid scan result ID" }),
+      { status: 400 }
+    );
+  }
+
+  const isAllowedToResolve = await isOnerepScanResultForSubscriber({
+    onerepScanResultId: scanResultId,
+    subscriberId: session.user.subscriber.id,
+  });
+  if (!isAllowedToResolve) {
+    return new NextResponse<ResolveScanResultResponse>(
+      JSON.stringify({ success: false, message: "Invalid scan result ID" }),
+      { status: 403 }
+    );
+  }
+
+  try {
+    await markOnerepScanResultAsResolved(scanResultId);
+    return new NextResponse<ResolveScanResultResponse>(
+      JSON.stringify({ success: true }),
+      { status: 200 }
+    );
+  } catch (e) {
+    console.error(e);
+    return new NextResponse<ResolveScanResultResponse>(
+      JSON.stringify({
+        success: false,
+        message: "Something went wrong, please try again.",
+      }),
+      { status: 500 }
+    );
+  }
+}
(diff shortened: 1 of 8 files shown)


src/knex-tables.d.ts
```diff
@@ -183,12 +183,13 @@ declare module "knex/types/tables" {
...
     last_name: string;
     status: RemovalStatus;
+    manually_resolved: boolean;
     created_at: Date;
     updated_at: Date;
   }
   type OnerepScanResultOptionalColumns = Extract<
     keyof OnerepScanResultRow,
-    "middle_name"
+    "manually_resolved" | "middle_name"
```
rhelmer, line 192: Hm, I don't understand why this type would be `"manually_resolved" | "middle_name"`, would you mind elaborating on this?

rhelmer, line 192: Should this be optional? We could make this a required column (with a default of `false`) if that makes more sense. Is the thinking more that the calling code won't always care about this value?

src/db/migrations/20230908154315_add_manual_resolution_column.js
```diff
@@ -0,0 +1,23 @@
...
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+/**
+ * @param { import("knex").Knex } knex
+ * @returns { Promise<void> }
+ */
+export async function up(knex) {
+  return knex.schema.alterTable("onerep_scan_results", table => {
+    table.boolean("manually_resolved").notNullable().defaultTo(false);
```
rhelmer, line 11: 👍

src/db/tables/onerep_scans.ts
```diff
@@ -111,6 +111,39 @@ async function setOnerepScanResults(
...
+
+  return typeof result?.onerep_scan_result_id === "number";
+}
+
+async function markOnerepScanResultAsResolved(
+  onerepScanResultId: number
+): Promise<void> {
+  await knex("onerep_scan_results")
+    .update("manually_resolved", true)
+    .where("onerep_scan_result_id", onerepScanResultId);
```
pdehaan, line 144: Q: does this need to set any `updated_at` columns, or is that done automatically via magic?

src/app/api/v1/user/scan-result/[onerepScanResultId]/resolution/route.ts
```diff
@@ -0,0 +1,66 @@
...
+      success: true;
+    }
+  | { success: false; message: string };
+
+export async function POST(
+  req: NextRequest,
+  { params }: { params: { onerepScanResultId: string } }
+): Promise<NextResponse<ResolveScanResultResponse>> {
+  const session = await getServerSession(authOptions);
+  if (!session || !session.user || !session.user.subscriber) {
```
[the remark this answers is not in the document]
Comment (Vinnl) [reply src/app/api/v1/user/scan-result/[onerepScanResultId]/resolution/route.ts]: Ah yes, I think so. Good example of how patterns proliferate across the codebase, since I based this file on another one: e0aa152e5

src/db/tables/onerep_scans.ts
Vinnl: Oh, good catch! It should, and actually TypeScript would have warned me about this if I had used the object syntax: ba55bebe9

src/app/api/v1/user/scan-result/[onerepScanResultId]/resolution/route.ts
```diff
@@ -0,0 +1,66 @@
...
+  markOnerepScanResultAsResolved,
+} from "../../../../../../../db/tables/onerep_scans";
+
+export type ResolveScanResultResponse =
+  | {
+      success: true;
+    }
+  | { success: false; message: string };
+
+export async function POST(
```
Comment (mansaj) [inline src/app/api/v1/user/scan-result/[onerepScanResultId]/resolution/route.ts:20]: Hmm, so if I'm understanding correctly, here's how manual resolution works:
1) the frontend would probably request the latest scan results for a profile at some point
2) with the list of scan results (one per data broker), we can essentially get the list of `onerepScanResultId` on the frontend
3) in the UI, we can resolve manually per data broker, with this POST call, we simply pass in the id

mansaj (approved):
pending question about how manual resolution works.

I think I also have an edge case for us to consider, which may alter how to get the latest scan results

src/app/api/v1/user/scan-result/[onerepScanResultId]/resolution/route.ts
Comment (Vinnl) [reply src/app/api/v1/user/scan-result/[onerepScanResultId]/resolution/route.ts:20]: Yep, that's exactly it! Specifically, [this UI](https://www.figma.com/file/CaEKIhvSJqf6KNIMzSkt40/Concepts-for-Monitor-MVP-Redesign?type=design&node-id=1290-62939&mode=dev) has "Mark as fixed" buttons that would send this POST call.

What's the edge case that you mentioned?

src/app/api/v1/user/scan-result/[onerepScanResultId]/resolution/route.ts
```diff
@@ -0,0 +1,66 @@
...
+      { status: 200 }
+    );
+  } catch (e) {
+    return new NextResponse<ResolveScanResultResponse>(
+      JSON.stringify({
+        success: false,
+        message: "Something went wrong, please try again.",
+      }),
+      { status: 500 }
+    );
```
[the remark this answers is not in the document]
Comment (rhelmer) [reply src/app/api/v1/user/scan-result/[onerepScanResultId]/resolution/route.ts:64]: I don't think so, only things that are either caught by the top-level error handler, or explicitly use `captureMessage` / `captureException` from Sentry.

These might appear in GCP logs, but I'm not sure if the message is, just the HTTP status we returned to client.

src/db/tables/onerep_scans.ts
rhelmer: I think we could have these default to the current timestamp (or whatever resolution is appropriate) and then purposely not set this column when updating, but we should do whatever is clearest for someone reading the code.

rhelmer (approved):
lgtm! I think as @pdehaan mentions, those error messages being returns in HTTP 500s likely won't be logged anywhere, so I'd suggest at minimum doing a `console.error` (which we can find easily in GCP logs), or using one of Sentry's "capture" functions.

src/app/api/v1/user/scan-result/[onerepScanResultId]/resolution/route.ts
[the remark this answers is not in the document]
Comment (Vinnl) [reply src/app/api/v1/user/scan-result/[onerepScanResultId]/resolution/route.ts:65]: I'm going to play my "I'm a lowly UI engineer!" card for not thinking about logging :sweat_smile: 

I added a `console.error` in 6710e67b7 - is it OK to just dump the error, or should I be adding more context (and if so, in what form)?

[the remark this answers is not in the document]
Comment (Vinnl) [reply src/app/api/v1/user/scan-result/[onerepScanResultId]/resolution/route.ts:65]: Going to merge now to unblock other PRs that have been waiting for a while, but happy to submit a follow-up PR that adds more details.

PR #3389, opened 2023-09-10
verifying behavior and look of breaches dashboard -  header

mozrokafor:
# References: 
Jira: [MNTOR-2132](https://mozilla-hub.atlassian.net/browse/MNTOR-2132)

# Description
- automates https://testrail.stage.mozaws.net/index.php?/cases/view/2095101, https://testrail.stage.mozaws.net/index.php?/cases/view/2095102 and https://testrail.stage.mozaws.net/index.php?/cases/view/2095103

### DIFF
--- a/src/e2e/pages/dataBreachPage.js
+++ b/src/e2e/pages/dataBreachPage.js
@@ -6,16 +6,50 @@ export class DataBreachPage {
   constructor (page) {
     this.page = page
 
-    this.dataBreachesHeader = page.locator('.breaches-header')
+    // navbar
+    this.dataBreachesNavbar = page.locator('.header-wrapper')
+    this.dataBreachesLogo = page.getByAltText("Firefox Monitor")
+    this.dataBreachesLogoLink = page.locator('.header-wrapper a')
+    this.dataBreachesNavbarProfile = page.locator('.user-menu-button')
+    this.dataBreachesNavbarProfileButton = page.locator('.nav-toggle')
+
+    // profile menu items
+    this.dataBreachesNavbarProfileMenu = page.locator('.nav-wrapper menu')
+    this.dataBreachesNavbarProfileMenuHeader = page.locator('.user-menu-header')
+    this.dataBreachesNavbarProfileMenuHeaderSubtitle = page.locator('a .user-menu-subtitle')
+    this.dataBreachesNavbarProfileMenuSettings = page.getByLabel('Account menu').getByRole('link', { name: 'Settings' })
+    this.dataBreachesNavbarProfileMenuHelpAndSupport = page.getByRole('link', { name: 'Help and support', exact: true })
+    this.dataBreachesNavbarProfileMenuSignOut = page.getByRole('button', { name: 'Sign out' })
+
+    // dashboard
     this.breachesFilter = page.locator('.breaches-filter')
+    this.dataBreachesHeader = page.locator('.breaches-header')
     this.breachesResolvedRadio = page.locator('input#breaches-resolved')
     this.breachesUnresolvedRadio = page.locator('input#breaches-unresolved')
     this.breachRows = page.locator('.breach-row')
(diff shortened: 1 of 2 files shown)


src/e2e/pages/dataBreachPage.js
```diff
@@ -6,16 +6,50 @@ export class DataBreachPage {
...
+      helpAndSupportNavButtonLink: "https://support.mozilla.org/kb/firefox-monitor",
+    }
   }
 
   async open () {
     await this.page.goto('/user/breaches')
   }
+
+  async profileMenuExpanded(){
+    return await this.dataBreachesNavbarProfileMenu.getAttribute("aria-expanded") === "true"
```
Vinnl, line 53: Thanks for checking this attribute - makes sure the expanded state will also stay visible for screen reader users!

src/e2e/specs/breachResolution.spec.js
```diff
@@ -59,7 +59,7 @@ test.describe('Breach Resolution', () => {
     await expect(async () => {
       await expect(updatedResolvedBreachTab).toBeTruthy()
     }).toPass({
-      timeout: 5000
+      timeout: 2000
```
Vinnl, line 62: Why do we have magic numbers for the timeout here and below? Ideally, the default timeout suffices for all tests, and if there is a specific reason a particular test takes longer (e.g. a data broker scan is always set to wait a minute before polling for results), that reason is mentioned in a comment next to the magic number.

src/e2e/pages/dataBreachPage.js
```diff
@@ -6,16 +6,50 @@ export class DataBreachPage {
...
+    this.dataBreachesNavbar = page.locator('.header-wrapper')
+    this.dataBreachesLogo = page.getByAltText("Firefox Monitor")
+    this.dataBreachesLogoLink = page.locator('.header-wrapper a')
+    this.dataBreachesNavbarProfile = page.locator('.user-menu-button')
+    this.dataBreachesNavbarProfileButton = page.locator('.nav-toggle')
+
+    // profile menu items
+    this.dataBreachesNavbarProfileMenu = page.locator('.nav-wrapper menu')
+    this.dataBreachesNavbarProfileMenuHeader = page.locator('.user-menu-header')
+    this.dataBreachesNavbarProfileMenuHeaderSubtitle = page.locator('a .user-menu-subtitle')
```
Vinnl, line 19: Possibly these can also be found using `nav` and `menu` roles - the [accessibility inspector](https://firefox-source-docs.mozilla.org/devtools-user/accessibility_inspector/index.html) can be helpful to find what roles are available. Checking for those ensures that we're also correctly exposing what the different buttons are for e.g. screen reader or TTS users.

Vinnl (approved):
Two small suggestions for improvements, but overall no biggies I don't think.

src/e2e/specs/breachResolution.spec.js
mozrokafor: this is added for a quicker fail of the assertion(instead of the default timeout). This is essentially a keep trying every i believe 200ms up until the 2000ms. This allows the script room for any UI/CI weirdness

mansaj (approved):
pending @Vinnl's comments

Also @mozrokafor, since we will be deprecating these pages `/user/...`, perhaps we can prioritize the `/redesign/user` pages instead?

src/e2e/specs/breachResolution.spec.js
Vinnl: OK, then I'd add that in a comment - though in general, I'd love to see "UI/CI weirdness" addressed the first time it shows up, because that has a habit of accumulating over time and making our tests flaky, while getting harder and harder to debug.

src/e2e/pages/dataBreachPage.js
mozrokafor: I am assuming this will be part of the redesign, will deprioritize

PR #3397, opened 2023-09-12
MNTOR-2120 - add Sentry logging for publisher, and skip pubsub for Heroku

rhelmer:
# References: 
Jira: MNTOR-2120

# Description

- The `api/v1/hibp/notify` route is a publisher, and if it drops messages they will be lost. Use Sentry to log these, since the messages are quite minimal.
- Heroku has no access to pubsub, detect this environment via `APP_ENV` and skip pubsub in that env.

# How to test

- Run locally w/o access to PubSub emulator and ensure that Sentry logging happens.
- Ensure that Heroku can process breach alerts.


src/app/api/v1/hibp/notify/route.ts
```diff
@@ -2,43 +2,84 @@
...
+  }
+  if (!topicName) {
+    const ex = "GCP_PUBSUB_TOPIC_NAME env var not set, message lost";
+    captureMessage(JSON.stringify(breachAlert));
+    captureException(ex);
+    throw new Error(ex);
+  }
+
+  const { PubSub } = await import("@google-cloud/pubsub");
   const pubsub = new PubSub({ projectId });
```
rhelmer, line 83: I was discussing this with @bbangert earlier and he suggested that we cache this and the topic after lookup instead of re-creating the connection every time this function is called ... that'll be a little awkward with this new Heroku "short-circuit" approach but probably worth it.

package.json
```diff
@@ -127,6 +127,7 @@
     "nodemon": "^2.0.20",
     "prettier": "2.8.8",
     "react-intersection-observer": "^9.5.2",
+    "request": "^2.88.2",
```
pdehaan, line 130: 👀?
This sneak in again?

rhelmer: It turns out we do need it - I opened a separate PR. I don't know why it's not listed as a dependency, NextJS fails trying to build it: https://github.com/mozilla/blurts-server/pull/3417

pdehaan: Now I'm 2&times; confused, since #3417 adds a [not very popular] [**requests**](https://www.npmjs.com/package/requests) (plural) instead of a [very popular, but deprecated, and generally vulnerable] [**request**](https://www.npmjs.com/package/request) (singular) dependency:

PACKAGE@VERSION | WEEKLY<br>DOWNLOADS | LAST<br>PUBLISH | NOTES
----|----:|----|----
request@2.88.2 | 10,914,14 | 4 years ago | DEPRECATED
requests@0.3.0 | 3,404 | 4 years ago

rhelmer:
FYI putting into draft mode for the moment - I want to do this still, but need to merge in some more immediate changes that landed in the meantime.

PR #3402, opened 2023-09-14
MONTOR-2045: Premium modal metrics and links

flozia:
# References: 
Jira: [MNTOR-2045](https://mozilla-hub.atlassian.net/browse/MNTOR-2045) & [MNTOR-1696](https://mozilla-hub.atlassian.net/browse/MNTOR-1696)

# Description

This PR is adding:
1. Metrics for opening and closing the Premium upsell modal. Metrics for the subscription flow itself are being added separately.
2. The correct subscription links depending on the plan a user selects. We’ll still need to add the yearly plan to staging ([MNTOR-2162](https://mozilla-hub.atlassian.net/browse/MNTOR-2162)).

[MNTOR-2045]: https://mozilla-hub.atlassian.net/browse/MNTOR-2045?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[MNTOR-2162]: https://mozilla-hub.atlassian.net/browse/MNTOR-2162?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[MNTOR-1696]: https://mozilla-hub.atlassian.net/browse/MNTOR-1696?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/app/(proper_react)/layout.tsx
```diff
@@ -3,15 +3,22 @@
...
+import { getNonce } from "../(nextjs_migration)/functions/server/getNonce";
 
 export default function Layout({ children }: { children: ReactNode }) {
   const l10nBundles = getL10nBundles();
 
   return (
     <L10nProvider bundleSources={l10nBundles}>
+      <Script
+        type="module"
+        src="/nextjs_migration/client/js/analytics.js"
```
Vinnl, line 19: Hey @flozia, I haven't looked at this PR in detail yet, but I'm a bit worried about this line. In any case I'd want to avoid references to `nextjs_migration` (I'd like to delete that dir after we've fully transitioned), but also I'd like to avoid vendoring our dependencies, and having implicit global variables.

Then again, this might only be temporary, what with Glean and all, @rhelmer?

But otherwise, maybe we could look into using something like [react-ga4](https://www.npmjs.com/package/react-ga4), or update [react-ga](https://www.npmjs.com/package/react-ga) ourselves (since that's a Mozilla package) to work with GA4?

flozia: We can use one of the proposed packages. I’m not sure about what events we’ll be using Glean and GA for. Before pushing this PR up [I asked about this](https://mozilla.slack.com/archives/C04BLQVN151/p1694452369102829)and it sounded like we are going to utilize both @rhelmer?

Vinnl: Yeah, I believe both at first, and then we can remove GA later after we've verified that they report similar data? Probably worth it to create a wrapper that reports to both, so we only need to update it there down the road.

src/app/(proper_react)/redesign/(authenticated)/user/dashboard/Dashboard.test.tsx
```diff
@@ -112,7 +117,7 @@ it("shows the premium upgrade cta if the user is not a premium subscriber", () =
   expect(premiumCtas.length).toBe(2);
 });
 
-it("opens and closes the premium upsell dialog", async () => {
+it("opens and closes the premium upsell dialog (premium upsell badge)", async () => {
```
Vinnl, line 120: I'm not sure I follow this description. Do you mean 

```suggestion
it("opens and closes the premium upsell dialog via the Premium upsell badge", async () => {
```

? (And similarly to the other test below?)

Vinnl (commented):
Overall approach looks good, pending not vendoring GA, and maybe adding a `sendEvent` wrapper that we can more easily add Glean to.

src/app/(proper_react)/redesign/(authenticated)/user/dashboard/Dashboard.test.tsx
flozia: Renamed the tests in [7051172](https://github.com/mozilla/blurts-server/pull/3402/commits/7051172e477127a734f530c2dfd0fb02c1797383).

src/app/(proper_react)/layout.tsx
rhelmer: Yes I think we'll use both at least for the foreseeable future - having one tool that did event collection both on the back-end and front-end and let us look at it as a single event stream during analysis (and even better had easy-to-use analysis tools) would be the ideal, but we don't have anything like that available to us currently. Glean is a requirement for Nimbus so we also need it for that reason, but I think the analysis tools GA (even 4) gives us is pretty hard to do without.

I like the idea of having a wrapper that reports to both - there's an extra step with Glean where we need to record and have approved the metrics we're collecting, but I don't think that's necessarily a bad thing.

Vinnl: Yeah if the events here are only going to be sent to one of the two, creating an abstraction over them seems out of scope.

flozia: Replaced with the `gaHook` that has been introduced by PR #3440 in https://github.com/mozilla/blurts-server/pull/3402/commits/4673a21723c9c723585b6f9b08628b7b7126d65d.

src/app/(proper_react)/layout.tsx
```diff
@@ -3,26 +3,15 @@
...
-  const ga4MeasurementId =
-    process.env.NEXT_PUBLIC_GA4_MEASUREMENT_ID || "G-CXG8K4KW4P";
-
   return (
     <L10nProvider bundleSources={l10nBundles}>
-      {navigator.doNotTrack !== "1" && (
-        <Script
-          src={`https://www.googletagmanager.com/gtag/js?id=${ga4MeasurementId}`}
-          nonce={headers().get("x-nonce") ?? ""}
-        />
```
Vinnl, line 24: Woah, did we already have this in here? :open_mouth:

src/app/(proper_react)/redesign/Shell.tsx
```diff
@@ -12,87 +12,92 @@ import { MobileShell } from "./MobileShell";
...
-        <div className={styles.content}>
-          <div className={styles.page}>{props.children}</div>
-          <footer className={styles.footer}>
-            <a
-              href="https://www.mozilla.org"
-              className={styles.mozillaLink}
-              target="_blank"
-            >
+    <>
+      <GaScript nonce={props.nonce} />
```
Vinnl, line 29: I'm assuming the only change in the huge diff below is indentation, right?

src/app/functions/universal/getPremiumSubscriptionUrl.ts
```diff
@@ -0,0 +1,24 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+interface GetPremiumSubscriptionUrlParams {
+  type: "monthly" | "yearly";
+}
+
+function getPremiumSubscriptionUrl({
```
Vinnl, line 9: Nice job factoring this out :clap:

Vinnl (approved):

src/app/(proper_react)/redesign/Shell.tsx
flozia: Yes, that’s right!

src/app/(proper_react)/layout.tsx
flozia: Just since PR #3402.

Vinnl: Oh haha, right

src/app/(proper_react)/redesign/GaScript.test.tsx
```diff
@@ -0,0 +1,33 @@
...
+import { GaScript } from "./GaScript";
+
+afterEach(() => {
+  // Clean up script elements
+  [...document.getElementsByTagName("script")].forEach((element) => {
+    element.remove();
+  });
+});
+
+it("renders the GA script tag", () => {
```
rhelmer, line 16: 👍

rhelmer (approved):
I have not tested locally but it looks OK to me (I can help test on stage) - thanks for adding tests!

Vinnl (approved):
Heh, I also just now realise that some of the updated `c8` comments simply _expand_ the scope that is ignored :sweat_smile:

src/app/(proper_react)/redesign/GaScript.test.tsx
```diff
@@ -0,0 +1,33 @@
...
+  const { container } = render(<GaScript nonce="" />);
+  expect(container).toBeInTheDocument();
+
+  expect(document.getElementsByTagName("script").length).toBe(1);
```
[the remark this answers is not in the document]
flozia: I was wondering the same and checked the scripts that were found and logged them while running unit tests: I’ve only caught the GA scripts.

flozia:
> Heh, I also just now realise that some of the updated `c8` comments simply _expand_ the scope that is ignored 😅

@Vinnl Thanks for checking in on the newly added changes. I only really removed one `start` and `stop` comment to address an issue where the last line with the `}` was marked as not covered leading to only expanding the scope to ignore [three more empty lines](https://github.com/mozilla/blurts-server/pull/3402/files#diff-636a780d9fc5548e2c0223ebfa71985792102b4c088cf4d311f3a2be86712d97L59-L63).

src/app/(proper_react)/redesign/GaScript.test.tsx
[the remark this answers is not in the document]
flozia: To be save in case anything changes in the future or is platform specific I added your suggestion in [2e8821b](https://github.com/mozilla/blurts-server/pull/3402/commits/2e8821b1f7493c5dc4b2c308cb6103ad71cc77e7).

PR #3405, opened 2023-09-14
Function to determine next step in the guided resolution flow

Vinnl:
# References: 
Jira: MNTOR-2015
Figma: N/A

# Description

The most important part in this PR is the new `getRelevantGuidedSteps` function, implemented in 342a31ccc894e0de687d40166daee2b3bb522e32. You pass it in a bunch of data, and it will tell you what the first step is that the user hasn't resolved yet, and which step they should end up at if they choose to skip that step.

As an example of how to use it, I added 9b73e9e8f707604fb89ad1230bb4e1c6cde11e50, which uses the function to determine where the CTA on the dashboard should point. If this PR gets merged, we can also use it for the CTAs, skip buttons, and right arrows throughout the guided flow.

I did not add a way to detect what the left arrow should point, because I'm not sure what the logic would be there. Presumably, it should either always go to the previous step (regardless of whether the user would have had anything to do there - i.e. they might not have seen that step before), potentially showing a zero state, or back to the dashboard - but in either of those cases, no special logic is needed. Alternatively, it could require more complicated logic, e.g. go to the last step where something was found, but was resolved, but if so, Product can file a separate ticket for that.

# How to test

Verify that the CTA on the dashboard points to the step you would expect for your different accounts.

### DIFF
--- a/src/app/(proper_react)/redesign/(authenticated)/user/dashboard/Dashboard.stories.tsx
+++ b/src/app/(proper_react)/redesign/(authenticated)/user/dashboard/Dashboard.stories.tsx
@@ -4,12 +4,12 @@
 
 import type { Meta, StoryObj } from "@storybook/react";
 
-import { OnerepScanResultRow } from "knex/types/tables";
+import { OnerepScanResultRow, OnerepScanRow } from "knex/types/tables";
 import { View as DashboardEl } from "./View";
 import { Shell } from "../../../Shell";
 import { getEnL10nSync } from "../../../../../functions/server/mockL10n";
 import {
-  createRandomScan,
+  createRandomScanResult,
   createRandomBreach,
   createUserWithPremiumSubscription,
 } from "../../../../../../apiMocks/mockData";
@@ -63,17 +63,26 @@ const BreachMockItem4: SubscriberBreach = createRandomBreach({
   ],
 });
 
+const scanSample: OnerepScanRow = {
+  created_at: new Date(1998, 2, 31),
+  updated_at: new Date(1998, 2, 31),
+  id: 0,
+  onerep_profile_id: 0,
+  onerep_scan_id: 0,
+  onerep_scan_reason: "initial",
+};
+
 const scannedResultsArraySample: OnerepScanResultRow[] = [
-  createRandomScan({ status: "removed" }),
-  createRandomScan({ status: "waiting_for_verification" }),
-  createRandomScan({ status: "optout_in_progress" }),
-  createRandomScan({ status: "new" }),
-  createRandomScan(),
+  createRandomScanResult({ status: "removed" }),
+  createRandomScanResult({ status: "waiting_for_verification" }),
+  createRandomScanResult({ status: "optout_in_progress" }),
+  createRandomScanResult({ status: "new", manually_resolved: false }),
+  createRandomScanResult({ status: "new", manually_resolved: true }),
 ];
 
 const scannedResolvedResultsArraySample: OnerepScanResultRow[] = Array.from(
   { length: 5 },
-  () => createRandomScan({ status: "removed" })
+  () => createRandomScanResult({ status: "removed" })
 );
 
 const breachItemArraySample: SubscriberBreach[] = [
@@ -180,9 +189,10 @@ export const DashboardWithScan: Story = {
   render: () => (
     <Shell l10n={getEnL10nSync()} session={mockSession}>
       <DashboardEl
+        countryCode="nl"
         user={mockSession.user}
         userBreaches={breachItemArraySample}
-        userScannedResults={scannedResultsArraySample}
+        userScanData={{ scan: scanSample, results: scannedResultsArraySample }}
         isEligibleForFreeScan={false}
         locale={"en"}
         bannerData={dashboardSummaryWithScan}
@@ -203,7 +213,7 @@ export const DashboardWithScanUserFromUs: Story = {
         countryCode="us"
         user={mockSession.user}
         userBreaches={breachItemArraySample}
-        userScannedResults={scannedResultsArraySample}
+        userScanData={{ scan: scanSample, results: scannedResultsArraySample }}
         isEligibleForFreeScan={true}
         locale={"en"}
         bannerData={dashboardSummaryWithScan}
@@ -221,9 +231,10 @@ export const DashboardWithoutScan: Story = {
   render: () => (
     <Shell l10n={getEnL10nSync()} session={mockSession}>
       <DashboardEl
+        countryCode="nl"
         user={mockSession.user}
         userBreaches={breachItemArraySample}
-        userScannedResults={[]}
+        userScanData={{ scan: null, results: [] }}
         isEligibleForFreeScan={false}
         locale={"en"}
         bannerData={dashboardSummaryNoScan}
@@ -241,9 +252,10 @@ export const DashboardWithoutScanUserFromUs: Story = {
   render: () => (
     <Shell l10n={getEnL10nSync()} session={mockSession}>
       <DashboardEl
+        countryCode="us"
         user={{ email: "example@example.com" }}
         userBreaches={breachItemArraySample}
-        userScannedResults={[]}
+        userScanData={{ scan: null, results: [] }}
         isEligibleForFreeScan={true}
         locale={"en"}
         bannerData={dashboardSummaryNoScan}
@@ -260,9 +272,10 @@ export const DashboardEmptyListState: Story = {
   render: () => (
     <Shell l10n={getEnL10nSync()} session={mockSession}>
       <DashboardEl
+        countryCode="us"
         user={mockSession.user}
         userBreaches={breachItemArraySample}
-        userScannedResults={[]}
+        userScanData={{ scan: null, results: [] }}
         isEligibleForFreeScan={true}
         locale={"en"}
         bannerData={dashboardSummaryNoScan}
@@ -282,7 +295,7 @@ export const DashboardFreeUser: Story = {
         countryCode="us"
         user={{ email: "example@example.com" }}
         userBreaches={breachItemArraySample}
-        userScannedResults={scannedResultsArraySample}
+        userScanData={{ scan: scanSample, results: scannedResultsArraySample }}
         isEligibleForFreeScan={true}
         locale={"en"}
         bannerData={dashboardSummaryWithScan}
@@ -302,7 +315,10 @@ export const DashboardFreeUserAllResolved: Story = {
         countryCode="us"
         user={{ email: "example@example.com" }}
         userBreaches={[]}
-        userScannedResults={scannedResolvedResultsArraySample}
+        userScanData={{
+          scan: scanSample,
+          results: scannedResolvedResultsArraySample,
+        }}
         isEligibleForFreeScan={true}
         locale={"en"}
         bannerData={dashboardSummaryWithScan}
@@ -327,7 +343,39 @@ export const DashboardPremiumUser: Story = {
           countryCode="us"
           user={userData}
           userBreaches={breachItemArraySample}
-          userScannedResults={scannedResultsArraySample}
+          userScanData={{
+            scan: scanSample,
+            results: scannedResultsArraySample,
+          }}
+          isEligibleForFreeScan={true}
+          locale={"en"}
+          bannerData={dashboardSummaryWithScan}
+          featureFlagsEnabled={{
+            FreeBrokerScan: true,
+            PremiumBrokerRemoval: true,
+          }}
+        />
+      </Shell>
+    );
+  },
+};
+
+export const DashboardPremiumUserAllResolved: Story = {
+  render: () => {
+    const userData = createUserWithPremiumSubscription();
+    return (
+      <Shell
+        l10n={getEnL10nSync()}
+        session={{ ...mockSession, user: userData }}
+      >
+        <DashboardEl
+          countryCode="us"
+          user={userData}
+          userBreaches={[]}
+          userScanData={{
+            scan: scanSample,
+            results: scannedResolvedResultsArraySample,
+          }}
           isEligibleForFreeScan={true}
           locale={"en"}
           bannerData={dashboardSummaryWithScan}
@@ -348,7 +396,7 @@ export const DashboardNoSession: Story = {
         countryCode="us"
         user={{ email: "example@example.com" }}
         userBreaches={breachItemArraySample}
-        userScannedResults={scannedResultsArraySample}
+        userScanData={{ scan: scanSample, results: scannedResultsArraySample }}
         isEligibleForFreeScan={false}
         locale={"en"}
         bannerData={dashboardSummaryWithScan}
--- a/src/app/(proper_react)/redesign/(authenticated)/user/dashboard/Dashboard.test.tsx
+++ b/src/app/(proper_react)/redesign/(authenticated)/user/dashboard/Dashboard.test.tsx
@@ -16,6 +16,7 @@ import Meta, {
   DashboardPremiumUser,
   DashboardNoSession,
   DashboardFreeUserAllResolved,
+  DashboardPremiumUserAllResolved,
 } from "./Dashboard.stories";
 
 function enablePremium() {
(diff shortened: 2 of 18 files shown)


src/app/(proper_react)/redesign/(authenticated)/user/dashboard/Dashboard.stories.tsx
```diff
@@ -180,9 +189,10 @@ export const DashboardWithScan: Story = {
   render: () => (
     <Shell l10n={getEnL10nSync()} session={mockSession}>
       <DashboardEl
+        countryCode="nl"
         user={mockSession.user}
         userBreaches={breachItemArraySample}
-        userScannedResults={scannedResultsArraySample}
+        userScanData={{ scan: scanSample, results: scannedResultsArraySample }}
```
Vinnl, line 195: Country code is now required, and in addition to the scan results we also need to know whether a scan was ran in the first place - those are needed to determine which next step is relevant.

src/app/(proper_react)/redesign/(authenticated)/user/dashboard/Dashboard.stories.tsx
```diff
@@ -327,7 +343,39 @@ export const DashboardPremiumUser: Story = {
...
+            FreeBrokerScan: true,
+            PremiumBrokerRemoval: true,
+          }}
+        />
+      </Shell>
+    );
+  },
+};
+
+export const DashboardPremiumUserAllResolved: Story = {
```
Vinnl, line 363: This use case (a Premium has that has resolved everything) has its own top banner content, hence a new story (and test).

src/app/(proper_react)/redesign/(authenticated)/user/dashboard/Dashboard.test.tsx
```diff
@@ -227,3 +228,26 @@ it("shows returned free user who has resolved all tasks premium upsell and all f
...
+  const bannerPremiumCta = screen.queryAllByRole("button", {
+    name: "See what’s fixed",
+  });
+  expect(bannerPremiumCta.length).toBe(1);
+
+  // click on cta
+  await user.click(bannerPremiumCta[0]);
+
+  const fixedTab = screen.getByRole("tab", { name: "Fixed" });
+  expect(fixedTab).toHaveAttribute("aria-selected", "true");
```
Vinnl, line 252: This highlighted a bug that I fixed in `TabList.tsx` that resulted in the tab not actually getting selected when switching tabs via the top banner CTA.

Also, the existing tests highlighted two logic errors in `getRelevantGuidedSteps`, so thanks for writing those @codemist and @flozia :)

src/app/(proper_react)/redesign/(authenticated)/user/dashboard/DashboardTopBanner.tsx
```diff
@@ -24,222 +28,261 @@ export type BannerContent =
...
+      /* c8 ignore next 2 */
+    }
+
+    /* c8 ignore start */
+    // If the user has run a data broker scan, but their data was not found,
+    // congratulate them and encourage them to see their results.
+    // TODO: Check with PM - this state is currently never shown, because users
+    //       are directed to their data breaches if they have no unresolved data
+    //       brokers. Should we be keeping track of the user "acknowledging"
+    //       their scan results or something?
```
Vinnl, line 184: Actually, I was typing out this question for PM, and then realised when this state should be shown. Update incoming.

src/app/functions/server/onerep.ts
```diff
@@ -136,6 +144,25 @@ export async function createProfile(profileData: ProfileData): Promise<number> {
   return savedProfile.id;
 }
 
+export async function getProfile(
```
Vinnl, line 147: Oh, I think I thought I needed this function, but then ended up not using it. Should I remove it, or shall I leave it in - it might come in handy later?

src/knex-tables.d.ts
```diff
@@ -59,22 +65,41 @@ declare module "knex/types/tables" {
...
+      | ({
+          useBreachId: boolean;
+        } & Record<
+          SubscriberEmail.email,
+          Record<
+            BreachRow.id,
+            {
+              isResolved: boolean;
+              resolutionsChecked: Array<
+                (typeof BreachDataTypes)[keyof typeof BreachDataTypes]
```
Vinnl, line 92: I wasn't aware of this previously, but apparently we do keep track for each breach specifically what data classes the user has resolved - in the `subscribers` table. So this is also the column we'll have to update when the user resolves something in the guided flow.

src/app/functions/server/getRelevantGuidedSteps.ts
```diff
@@ -0,0 +1,241 @@
...
+    // All steps should have been covered by the above conditions:
+    /* c8 ignore next 4 */
+  }
+
+  return false as never;
+}
+
+function hasCompleted(data: InputData, stepId: StepLink["id"]): boolean {
+  if (stepId === "StartScan") {
+    return data.latestScanData?.scan !== null;
```
Vinnl, line 170: Wait... If `latestScanData` is `undefined`, this function implies that it has been completed?

src/app/functions/server/getRelevantGuidedSteps.ts
```diff
@@ -0,0 +1,241 @@
...
+    return isResolved(BreachDataTypes.Passwords);
+  }
+
+  if (stepId === "LeakedPasswordsSecurityQuestion") {
+    return isResolved(BreachDataTypes.SecurityQuestions);
+  }
+
+  if (stepId === "SecurityTipsPhone") {
+    // TODO: Verify that pressing "Got it" on this screen should actually mark
+    //       phone breaches as resolved.
```
Vinnl, line 222: @codemist is verifying this with UX. It probably should.

src/app/functions/server/getRelevantGuidedSteps.ts
```diff
@@ -0,0 +1,241 @@
...
+    // Anyone can view/resolve their high risk data breaches
+    return true;
+  }
+
+  if (
+    ["LeakedPasswordsPassword", "LeakedPasswordsSecurityQuestion"].includes(
+      stepId
+    )
+  ) {
+    // Anyone can view/resolve their high risk data breaches
```
pdehaan, line 146: copy pasta?
```suggestion
    // Anyone can view/resolve their leaked password breaches
```

src/app/functions/server/getRelevantGuidedSteps.ts
```diff
@@ -0,0 +1,241 @@
...
+  },
+  {
+    href: "/redesign/user/dashboard/fix/security-recommendations/ip",
+    id: "SecurityTipsIp",
+  },
+  {
+    href: "/redesign/user/dashboard",
+    id: "Done",
+  },
+] as const satisfies ReadonlyArray<{ href: string; id: string }>;
```
pdehaan, line 68: Should `id` be "string" or should we be enforcing values against some allow-list?
(Just seeing stuff like `if (stepId === "StartScan") {...}` below and getting scared of typos or mismatches)

src/app/functions/server/getRelevantGuidedSteps.ts
```diff
@@ -0,0 +1,241 @@
...
+    )
+  ) {
+    // Anyone can view security tips
+    return true;
+  }
+
+  if (stepId === "Done") {
+    return true;
+    // All steps should have been covered by the above conditions:
+    /* c8 ignore next 4 */
```
pdehaan, line 162: personal curiosity: why here instead of on L164 closer to the `return` statement?

src/app/functions/server/getRelevantGuidedSteps.ts
```diff
@@ -0,0 +1,241 @@
...
+  }
+
+  if (stepId === "SecurityTipsIp") {
+    return isResolved(BreachDataTypes.IP);
+  }
+
+  if (stepId === "Done") {
+    return false;
+    // All steps should have been covered by the above conditions:
+    /* c8 ignore next 4 */
```
pdehaan, line 237: Same Q here of why not closer to L240?

src/app/functions/server/getRelevantGuidedSteps.ts
```diff
@@ -0,0 +1,241 @@
...
+export type InputData = {
+  user: Session["user"];
+  countryCode: string;
+  latestScanData: LatestOnerepScanData | null;
+  subscriberBreaches: SubscriberBreach[];
+};
+
+// Note: the order is important; it determines in which order the user will be
+//       guided through the pages.
+export const stepLinks = [
```
flozia, line 19: Question: Are there use cases where it would be easier to make this a two-dimensional array and group them per high-level step? Currently, I can only think of marking previous steps as ”done” being easier.
image

src/app/functions/server/getRelevantGuidedSteps.ts
```diff
@@ -0,0 +1,241 @@
...
+    current: current,
+    skipTarget: getNextGuidedStep(data, nextStepsToSkip),
+  };
+}
+
+function getNextGuidedStep(
+  data: InputData,
+  stepsToSkip: Array<StepLink["id"]> = []
+): StepLink | null {
+  // Resisting the urge to add a state machine... ^.^
```
flozia, line 106: … so it’s not off the table? ;)

src/app/functions/server/getRelevantGuidedSteps.ts
```diff
@@ -0,0 +1,241 @@
...
+    return true;
+  }
+
+  if (stepId === "Done") {
+    return true;
+    // All steps should have been covered by the above conditions:
+    /* c8 ignore next 4 */
+  }
+
+  return false as never;
```
flozia, line 165: Question: If we would never return anything here this is basically dead code? In case this is here to catch any logic errors when tweaking the above logic in the future: Would a warning log with a message be more helpful to have?

src/utils/breachResolution.js
```diff
@@ -9,6 +9,8 @@ import { getMessage } from './fluent.js'
  * Equivalent of Typescript "enum"
  * These enum types map to HIBP's breach data types, defined in HIBP's API
  * Always reference enum instead of strings to avoid spelling error / typos (ie. BreachDataTypes.Passwords)
+ *
+ * @deprecated Use the object with the same name from /src/app/functions/universal/breach.ts instead.
```
flozia, line 13: Praise: Thanks for adding this note.

flozia (approved):
Great job — thanks for coming up with and adding the routing logic!

src/app/(proper_react)/redesign/(authenticated)/user/dashboard/DashboardTopBanner.tsx
Vinnl: To deal with this, I created https://github.com/mozilla/blurts-server/pull/3423, and then noticed that there is a whole bunch of other errors. So I'll postpone this until that is merged, after which we can create tickets for all of those.

src/app/(proper_react)/redesign/(authenticated)/user/dashboard/Dashboard.stories.tsx
```diff
@@ -327,7 +343,39 @@ export const DashboardPremiumUser: Story = {
           countryCode="us"
           user={userData}
           userBreaches={breachItemArraySample}
-          userScannedResults={scannedResultsArraySample}
+          userScanData={{
+            scan: scanSample,
+            results: scannedResultsArraySample,
+          }}
+          isEligibleForFreeScan={true}
+          locale={"en"}
```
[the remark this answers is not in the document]
Vinnl: It's rewritten in https://github.com/mozilla/blurts-server/pull/3423 anyway, so to avoid conflicts, I'm leaving it as-is for now.

src/app/functions/server/getRelevantGuidedSteps.ts
Vinnl: :eagle: :eye: 13cde8fb0

src/app/functions/server/getRelevantGuidedSteps.ts
Vinnl: Hmm, not sure. I think I'd rather keep the data structure simple, and potentially add a helper function to operate on the parts that apply to specific steps later, if we do encounter that use case often.

src/app/functions/server/getRelevantGuidedSteps.ts
Vinnl: Yeah, this should be dead code. I was actually hoping that this would trigger TS errors if I forgot to handle a particular step above, but that doesn't seem to happen :/ We could add a log, but I think this would then either never log, or suddenly fill up our logs entirely if we'd forgotten something.

Vinnl:
@flozia @codemist I just added 0ab33b2c6 as per our pairing last week - in addition to just returning the next step, you can now also call `getGuidedStepStatuses()` to get all steps and their statuses, to avoid having to muck about with cookies or something to keep track of which steps were skipped.

src/app/functions/server/onerep.ts
flozia: I don’t have anything against leaving it as a helper function and think it might come in handy later.

src/app/functions/server/getRelevantGuidedSteps.ts
```diff
@@ -0,0 +1,260 @@
...
+
+  return {
+    current: current,
+    skipTarget: getNextGuidedStep(data, current.id),
+  };
+}
+
+function getNextGuidedStep(
+  data: InputData,
+  fromStep?: StepLink["id"]
```
flozia, line 106: Works great! The only thing I can think of right now is that it might be helpful to allow passing `Step["href"]` to `fromStep` alternatively as well. But I guess we can also just match and find the step id from the URL just before passing it in.

src/app/functions/server/getRelevantGuidedSteps.ts
```diff
@@ -0,0 +1,260 @@
...
+    return !data.subscriberBreaches.some((breach) => {
+      return (
+        breach.dataClasses.includes(dataClass) &&
+        !breach.resolvedDataClasses.includes(dataClass)
+      );
+    });
+  }
+
+  // TODO: instead of resolving entire breaches, only resolve specific data classes
+  //       See ADR 0008-data-class-resolution.
```
flozia, line 216: I was not able to find the ADR. Didn’t we already find out that we might not need to restructure in order to be able to resolve specific data classes?

src/app/functions/server/getRelevantGuidedSteps.ts
```diff
@@ -0,0 +1,260 @@
...
+  const fromIndex = stepLinkStatuses.findIndex((step) => step.id === fromStep);
+  const stepLinksToCheck =
+    fromIndex !== -1 ? stepLinkStatuses.slice(fromIndex) : stepLinkStatuses;
+  const nextStep = stepLinksToCheck.find((stepLink) => {
+    return stepLink.eligible && !stepLink.completed;
+  });
+
+  // There should be a relevant next step for every user (even if it's just
+  // going back to the dashbord), so we can't hit this line in tests (and
+  // shouldn't be able to in production either):
```
flozia, line 119: Is it worth it to add Glean pings or server logs in the future to make sure dead code paths are really not being hit in production? This could help us detect if there is a significant number of people who unintentionally speed-run the breach resolution.

flozia (approved):

src/app/functions/server/getRelevantGuidedSteps.ts
Vinnl: Yeah, the reason I added the IDs in the first place is that links are so fiddly (do the include a trailing slash? are they relative? do they include the origin?). And since most calling code will know which step they're currently in, I figured it'd be safer to have them use the ID.

src/app/functions/server/getRelevantGuidedSteps.ts
Vinnl: Whoops. I actually have one written out, but then I found out that we don't actually need to resolve entire breaches (see `isResolved` above, which _does_ look at the specific data class), and so it was no longer needed. I just forgot to remove this comment. Removed in 74b759d1d.

src/app/functions/server/getRelevantGuidedSteps.ts
Vinnl: Yeah good idea, added in 804575af0.

Also made me realise that this specific line should actually be coverable in tests, since it's also called to determine the _next_ step - and when the user is done, there might not be one.

Which, in turn, made me realise that there was actually a bug: when calculating the next relevant step, instead of evaluating all steps after the current step, I was evaluating all steps _starting with_ the current step - which would, of course, always match the current step. So that is now fixed too: 83669d8e4

PR #3407, opened 2023-09-14
Make session prop non-optional on dashboard

flozia:
# Description

Remove `null` as a possible type for `session` on the dashboard and a unit test that covers a code path we won’t be encountering.


Vinnl (approved):
Thanks for taking this on!

PR #3417, opened 2023-09-15
MNTOR-2120 - fix build problems with HIBP publisher route:

rhelmer:
# References: 
Jira: MNTOR-2120

# Description

I noticed that the NextJS build is complaining about this (although it doesn't fail the CI build, and the route seems to work?):

1. force dyanmic https://nextjs.org/docs/app/api-reference/file-conventions/route-segment-config#dynamic
2. provide unlisted `request` dependency


package.json
```diff
@@ -84,6 +84,7 @@
     "react-cookie": "^4.1.1",
     "react-dom": "^18.2.0",
     "react-stately": "^3.23.0",
+    "requests": "^0.3.0",
```
mansaj, line 87: is this package related?

pdehaan, line 87: Commented in https://github.com/mozilla/blurts-server/pull/3397#discussion_r1327645030, but should this be request singular or plural? It was the singular "request" dependency in the other PR.

Vinnl:
Do you have a link to and/or capture of the build error, so we can check how this fixes them?

package.json
rhelmer: Sorry, as @pdehaan points out, `request` is related, corrected.

rhelmer:
> Do you have a link to and/or capture of the build error, so we can check how this fixes them?

Sorry, meant to include it. Right now builds on `main` complain about a missing `request` module (which seems to be an undeclared dependency of the pubsub client), that needs to be fixed to show this error.

I temporarily reverted the fix on this branch, I'm seeing a similar error for a different file now but I don't see `hibp` in the stack trace now: https://github.com/mozilla/blurts-server/actions/runs/6230577278/job/16910807504?pr=3417#step:11:57

The build does fail though. It's odd, I see more `staticGenerationBailout`s than this doing a local build too.

pdehaan:
Ah, I think we might be seeing this: https://github.com/googleapis/retry-request/issues/103

```sh
npm ls retry-request

monitor@1.0.0 /Volumes/Dev/github/mozilla/blurts-server
└─┬ @google-cloud/pubsub@4.0.3
  └─┬ google-gax@4.0.3
    └── retry-request@6.0.0
```

Looks like they have **requests** as a devDependency (and not a dependency or peerDependency):

```sh
npm info retry-request devDependencies dependencies

devDependencies = {
  '@types/request': '^2.48.8',
  async: '^3.0.1',
  gts: '^3.1.1',
  jsdoc: '^4.0.0',
  'jsdoc-fresh': '^2.0.0',
  'jsdoc-region-tag': '^2.0.0',
  linkinator: '^4.0.0',
  'lodash.range': '^3.2.0',
  mocha: '^9.2.2',
  request: '^2.87.0',
  typescript: '^5.1.6'
}
dependencies = { debug: '^4.1.1', extend: '^3.0.2' }
```

Vinnl:
> I temporarily reverted the fix on this branch, I'm seeing a similar error for a different file now but I don't see hibp in the stack trace now: https://github.com/mozilla/blurts-server/actions/runs/6230577278/job/16910807504?pr=3417#step:11:57
>
> The build does fail though.

I just reran that Playwright test, and it looks like it was flaky (/FxA was having issues on stage? Raphael asked me about that yesterday too). It succeeded now, even with the revert, so it looks like that didn't actually cause an error? (Which isn't to say that it's not a good addition.)

rhelmer:
> > I temporarily reverted the fix on this branch, I'm seeing a similar error for a different file now but I don't see hibp in the stack trace now: https://github.com/mozilla/blurts-server/actions/runs/6230577278/job/16910807504?pr=3417#step:11:57
> > The build does fail though.
> 
> I just reran that Playwright test, and it looks like it was flaky (/FxA was having issues on stage? Raphael asked me about that yesterday too). It succeeded now, even with the revert, so it looks like that didn't actually cause an error? (Which isn't to say that it's not a good addition.)

Yes I think you're right, I assumed this was causing the build failure but I think it was just a coincidence. We should clean these up, but I think it makes it less urgent.

PR #3421, opened 2023-09-16
MNTOR-2022 - add basic Cirrus integration

rhelmer:
# References: 
Jira: MNTOR-2022

# Description

Add basic Cirrus integration:

1. derive user ID from server-side session object
2. send user ID to Cirrus, obtain enabled features
3. send user ID to client-side Glean JS component

# How to test

1. Load any page in the current production `(nextjs_migration)` set of pages
2. Observe in browser console and [Glean Debug Ping Viewer](https://debug-ping-preview.firebaseapp.com/pings/fx-monitor-local-dev) that pings are being sent
3. Verify that pings from unauthenticated sessions have a `user_id` that begins with `guest-`, and are different from authenticated sessions

- [x] Playwright needs access to new `NIMBUS_UUID_NAMESPACE` and `NIMBUS_SIDECAR_URL` env vars


src/app/(nextjs_migration)/(guest)/page.tsx
```diff
@@ -14,10 +14,25 @@ import MailImage from "../../../client/images/landing-mail@2x.webp";
...
 
-export default function Home() {
+export default async function Home() {
   const l10n = getL10n();
 
+  const session = await getServerSession(authOptions);
+  const userId = session?.user?.subscriber?.id; // FIXME what if the user is not logged in? Need a "device ID" or equivalent (cookie, etc)
+
+  // @see https://github.com/mozilla/experimenter/tree/main/cirrus
+  const serverUrl = process.env.SERVER_URL ?? "http://localhost:6060";
```
pdehaan, line 28: Q: Should this be `SERVER_URL` or `NEXT_PUBLIC_SERVER_URL` since it looks like client front end code?

rhelmer: This is a server component actually - it needs to be in order to communicate with Cirrus. It then passes userId as a prop to the client component (where it's used by Glean).

rhelmer: (I'm actually not totally sure which URL we need to use here to talk to the Cirrus sidecar... I think it might need to be hardcoded to `localhost` with a particular port)

src/app/(nextjs_migration)/(authenticated)/user/layout.tsx
```diff
@@ -27,6 +29,28 @@ const MainLayout = async (props: Props) => {
     return <SignInButton autoSignIn />;
   }
 
+  const accountId = session?.user?.subscriber?.fxa_uid;
+
+  let userId;
+  if (accountId) {
```
Vinnl, line 35: I'm assuming `accountId` can't be `0`, but to be sure, it's probably better to do something like:

```suggestion
  if (typeof accountId === "number") {
```

(Or a `string` if that's what it is, of course.)

src/app/(nextjs_migration)/(authenticated)/user/layout.tsx
```diff
@@ -27,6 +29,28 @@ const MainLayout = async (props: Props) => {
...
+    if (!process.env.NIMBUS_UUID_NAMESPACE) {
+      throw new Error("env var NIMBUS_UUID_NAMESPACE not set");
+    }
+    userId = uuidv5(accountId.toString(), process.env.NIMBUS_UUID_NAMESPACE);
+  }
+
+  // @see https://github.com/mozilla/experimenter/tree/main/cirrus
+  const serverUrl = process.env.SERVER_URL ?? "http://localhost:6060";
+
+  //@ts-ignore TODO this tells us which features to enable.
```
Vinnl, line 48: Could you clarify in the comment what error we're suppressing here? It's not clear to me from reading the diff.

src/app/(nextjs_migration)/(guest)/layout.tsx
```diff
@@ -2,23 +2,65 @@
...
+      // @see https://nextjs.org/docs/app/api-reference/functions/cookies#cookiessetname-value-options
+      // cookies().set("userId", uuid);
+      userId = `guest-${randomUUID()}`;
+    }
+  }
+
+  // @see https://github.com/mozilla/experimenter/tree/main/cirrus
+  const serverUrl = process.env.SERVER_URL ?? "http://localhost:6060";
+
+  //@ts-ignore TODO this tells us which features to enable, for A/A testing we do nothing.
```
Vinnl, line 55: Same as other comment.

src/app/(nextjs_migration)/components/client/PageLoadEvent.tsx
```diff
@@ -7,16 +7,29 @@
...
+  userId: string;
+};
 
 // Empty component that records a page view on first load.
-export const PageLoadEvent = () => {
+export const PageLoadEvent = (props: Props) => {
+  const [cookies, setCookie] = useCookies(["userId"]);
+  const userId = props.userId;
+
+  // TODO decide if we want to do this here or on the server side
```
Vinnl, line 21: If here, then presumably inside the `useEffect`, rather than on every render (which might happen very frequently).

Vinnl (commented):
Not sure if this is supposed to be reviewed already, given the draft status, but added a few comments with things that struck me.

src/app/(nextjs_migration)/(authenticated)/user/layout.tsx
rhelmer: The error is that `features` is unused - I'll expand on the comment here. It will be used, but for the "A/A" test we don't need it.

src/app/(nextjs_migration)/(guest)/layout.tsx
rhelmer: I copy/pasted this but I will try to make this shared, same answer as above :)

src/app/(nextjs_migration)/(authenticated)/user/layout.tsx
```diff
@@ -27,6 +29,32 @@ const MainLayout = async (props: Props) => {
...
+    }
+    userId = uuidv5(accountId.toString(), process.env.NIMBUS_UUID_NAMESPACE);
+  }
+
+  if (!userId) {
+    throw new Error("No user ID for Nimbus telemetry");
+  }
+
+  // @see https://github.com/mozilla/experimenter/tree/main/cirrus
+  const serverUrl = process.env.SERVER_URL ?? "http://localhost:6060";
```
pdehaan, line 50: I almost wish there was a config wrapper layer around env vars or something so we wouldn't need to have fallbacks or cast from string to int peppered in the codebase.
```sh
git grep -Pn "localhost:6\d{3,}" -- ":(exclude)*.md" | cat

.env-dist:7:SERVER_URL=http://localhost:6060
.env-dist:10:NEXTAUTH_URL=http://localhost:6060
playwright.config.js:58:    // baseURL: 'http://localhost:6060',
src/app/(nextjs_migration)/(authenticated)/user/layout.tsx:50:  const serverUrl = process.env.SERVER_URL ?? "http://localhost:6060";
src/app/(nextjs_migration)/(guest)/layout.tsx:53:  const serverUrl = process.env.SERVER_URL ?? "http://localhost:6060";
src/e2e/utils/helpers.js:14:  local: 'http://localhost:6060',
```

But not sure if something like [**convict**](https://www.npmjs.com/package/convict) is overkill (or if Next.js has a better built-in solution). But the other benefit of convict is that we could potentially [disallow process.env.* usage](https://eslint.org/docs/latest/rules/no-process-env) via ESLint to try and force everybody to go through the config layer.

src/app/(nextjs_migration)/(authenticated)/user/layout.tsx
```diff
@@ -27,6 +29,32 @@ const MainLayout = async (props: Props) => {
     return <SignInButton autoSignIn />;
   }
 
+  const accountId = session?.user?.subscriber?.fxa_uid;
+
+  let userId;
+  if (accountId) {
+    // If the user is logged in, use a UUID based on the user's subscriber ID.
+    // Note: we may want to use the FxA UID here, but we need approval for that first.
```
pdehaan, line 37: Q: DO we have a ticket for this? If so, can we drop it here?

src/app/(nextjs_migration)/(authenticated)/user/layout.tsx
```diff
@@ -36,6 +64,7 @@ const MainLayout = async (props: Props) => {
         src="/nextjs_migration/client/js/nav.js"
```
pdehaan, line 64: Unrelated, but I do love a good sidebar.
Do we have planned work to tweak all these URLs in the future before release? Having paths at launch like /nextjs_migration/ feels&hellip; wrong?

src/app/(nextjs_migration)/(authenticated)/user/layout.tsx
```diff
@@ -27,6 +29,32 @@ const MainLayout = async (props: Props) => {
...
   }
 
+  const accountId = session?.user?.subscriber?.fxa_uid;
+
+  let userId;
+  if (accountId) {
+    // If the user is logged in, use a UUID based on the user's subscriber ID.
+    // Note: we may want to use the FxA UID here, but we need approval for that first.
+    // No need to store in a cookie - this is derived from subscriber ID.
+    if (!process.env.NIMBUS_UUID_NAMESPACE) {
```
pdehaan, line 39: Below (L68) we used `NEXT_PUBLIC_*` prefix, does this need to be `NEXT_PUBLIC_NIMBUS_UUID_NAMESPACE`?
We also seem to be using `process.env.SERVER_URL` but not `process.env.NEXT_PUBLIC_SERVER_URL` so maybe not. 🤷 

https://github.com/mozilla/blurts-server/blob/bb08ba7f28649e0ff2dfc4e57491409ad0ca4e01/src/app/(nextjs_migration)/(authenticated)/user/layout.tsx#L68

src/app/hooks/useGlean.ts
```diff
@@ -17,16 +17,25 @@ export const useGlean = () => {
...
     });
 
     // Glean debugging options can be found here:
     // https://mozilla.github.io/glean/book/reference/debug/index.html
     if (process.env.NEXT_PUBLIC_APP_ENV === "local") {
       // Enable logging pings to the browser console.
       Glean.setLogPings(true);
+      // Tag pings for the Debug Viewer
+      // @see https://debug-ping-preview.firebaseapp.com/pings/fx-monitor-local-dev
+      Glean.setDebugViewTag("fx-monitor-local-dev");
```
pdehaan, line 38: Imma confused.
is... do we need a different value for stage/prod, or does everything get stored in a Glean bucket named `*-local-dev`?
If it is different, should we toss this into an env var and add to .env-dist?

src/app/(nextjs_migration)/(guest)/layout.tsx
rhelmer: I'll also file a Jira ticket and link it which will provide more complete context. The real TODO action here is:

- pass targeting info (locale, geo, etc) to Nimbus
- process `feature` response from Nimbus, and enable any features that the user is eligible for

Nimbus knows which experiments are active and will do the sampling and determine eligibility based on the targeting criteria, Monitor just needs to enable the proper feature for the current user (conceptually similar to the way we do feature-flag allow-lists now, but I don't think we need to use the database for this - we can ask Nimbus as-needed).

rhelmer: I'll also moved the shared code to a server function, I didn't realize when I started that I'd need to implement this separately for guest/authenticated.

src/app/(nextjs_migration)/(guest)/page.tsx
rhelmer: OK I've moved this to a shared function, and I think I've figured out how to call the Cirrus sidecar.

src/app/(nextjs_migration)/(authenticated)/user/layout.tsx
```diff
@@ -27,6 +30,26 @@ const MainLayout = async (props: Props) => {
...
+    userId = uuidv5(accountId, process.env.NIMBUS_UUID_NAMESPACE);
+  }
+
+  if (!userId) {
+    throw new Error("No user ID for Nimbus telemetry");
+  }
+
+  // TODO For initial A/A testing `features` is unused.
+  // @ts-ignore TODO features is declared but its value is never read
+  const features = await getExperiments(userId);
```
yashikakhurana, line 51: should we consider logging it to test end to end?

src/app/(nextjs_migration)/(authenticated)/user/layout.tsx
rhelmer: @pdehaan would you mind putting this on agenda for tech meeting?

src/app/(nextjs_migration)/(guest)/layout.tsx
```diff
@@ -2,23 +2,70 @@
...
-const GuestLayout = (props: Props) => {
+const GuestLayout = async (props: Props) => {
   const l10n = getL10n();
 
+  // If the user is logged in, use UUID derived from FxA UID as Nimbus user ID.
+  const session = await getServerSession(authOptions);
+  const accountId = session?.user?.subscriber?.fxa_uid;
+  let userId;
+
+  if (accountId) {
```
pdehaan, line 32: Q: Above we had an extra check for `typeof accountId === "string"`. Do we not need that here?

```
if (accountId && typeof accountId === "string") { ... }
```

src/app/functions/server/getExperiments.ts
```diff
@@ -0,0 +1,34 @@
...
+ * Call the Cirrus sidecar, which returns a list of eligible experiments for the current user.
+ *
+ * @see https://github.com/mozilla/experimenter/tree/main/cirrus
+ * @param userId Persistent ID for user, either guest or authenticated
+ * @returns {Promise<any>}
+ */
+export async function getExperiments(
+  userId: string | undefined
+): Promise<unknown> {
+  const serverUrl = process.env.NIMBUS_SIDECAR_URL ?? process.env.SERVER_URL;
```
pdehaan, line 17: We should possibly confirm w/ SRE that this will work. I think we had an issue w/ some similar logic (although it was wrapped w/ **convict** so maybe not _exactly_ the same) where SRE was setting falsy env vars like `NIMBUS_SIDECAR_URL=""` in some build configs and if we're doing something similar in this project it wouldn't work with the nullish coalescing operator. https://mozilla-hub.atlassian.net/browse/FXA-7621

Looks like this might be the first instance of nullish fallback to another env var, so worth a quick ping perhaps:

```sh
git grep -Pn " \?\? process\.env\." | cat

src/app/functions/server/getExperiments.ts:17:  const serverUrl = process.env.NIMBUS_SIDECAR_URL ?? process.env.SERVER_URL;
```

Although, counterpoint, we seem to do similar things elsewhere already too:

<details>
<summary><kbd>git grep -n " ?? " | grep -E "(process\.env|AppConstants)\."</kbd></summary>

```sh
docs/adr/0005-feature-flags.md:124:  return (process.env.ENABLED_FEATURE_FLAGS ?? '').split(',').includes(flag);
src/app/(nextjs_migration)/(authenticated)/admin/emails/[template]/page.tsx:47:  const admins = process.env.ADMINS?.split(",") ?? [];
src/app/(nextjs_migration)/(authenticated)/admin/emails/page.tsx:17:  const admins = process.env.ADMINS?.split(",") ?? [];
src/app/(nextjs_migration)/(authenticated)/admin/layout.tsx:17:  const admins = process.env.ADMINS?.split(",") ?? [];
src/app/(proper_react)/redesign/(authenticated)/user/dashboard/fix/leaked-passwords/leakedPasswordsData.tsx:51:  const blockList = (process.env.HIBP_BREACH_DOMAIN_BLOCKLIST ?? "").split(",");
src/app/api/utils/auth.ts:224:  const admins = AppConstants.ADMINS?.split(",") ?? [];
src/app/api/v1/premium/route.ts:19:      `${process.env.SERVER_URL ?? ""}/user/breaches`,
src/app/api/v1/user/verify-email/route.ts:16:      (process.env.SERVER_URL ?? "") + "/user/settings",
src/app/functions/server/breachResolution.ts:102:      const blockList = (AppConstants.HIBP_BREACH_DOMAIN_BLOCKLIST ?? "").split(
src/app/functions/server/getExperiments.ts:17:  const serverUrl = process.env.NIMBUS_SIDECAR_URL ?? process.env.SERVER_URL;
src/app/functions/server/isUserSubscribed.ts:28:      process.env.OAUTH_API_URI ?? ""
src/e2e/utils/helpers.js:20:  process.env['E2E_TEST_ENV'] = process.env.E2E_TEST_ENV ?? 'local'
src/e2e/utils/helpers.js:22:  process.env['E2E_TEST_BASE_URL'] = ENV_URLS[process.env.E2E_TEST_ENV] ?? 'https://stage.firefoxmonitor.nonprod.cloudops.mozgcp.net'
src/utils/breachResolution.js:119:      const blockList = (AppConstants.HIBP_BREACH_DOMAIN_BLOCKLIST ?? '').split(',')
src/views/guestLayout.js:16:  const pageUrl = `${AppConstants.SERVER_URL}${data.pathname ?? '/'}`
src/views/mainLayout.js:16:  const pageUrl = `${AppConstants.SERVER_URL}${data.pathname ?? '/'}`
src/views/partials/floatingBanner.js:31:    <div tabindex='-1' class='floating-banner' data-type='${AppConstants.FLOATING_BANNER_TYPE ?? ''}' data-delay='${AppConstants.FLOATING_BANNER_DELAY ?? '0'}'  hidden>
```
</details>

---

```js
// env.test.js
const { inspect } = require("node:util");

console.log(`TEST=${inspect(process.env.TEST)}`);
```

```sh
TEST= node env.test.js
```

#### OUTPUT

```text
TEST=''
```

So looks like if SRE doesn't set empty vars it might work, if they set `process.env.NIMBUS_SIDECAR_URL=` then it might fail.

src/app/(nextjs_migration)/(authenticated)/user/layout.tsx
rhelmer: Added to agenda for next tech meeting

src/app/(nextjs_migration)/(authenticated)/user/layout.tsx
```diff
@@ -27,6 +30,32 @@ const MainLayout = async (props: Props) => {
...
+    }
+  }
+
+  if (!userId) {
+    console.error("No user ID for Nimbus telemetry");
+  }
+
+  try {
+    // TODO For initial A/A testing `features` is unused. https://mozilla-hub.atlassian.net/browse/MNTOR-2182
+    const features = await getExperiments(userId);
```
pdehaan, line 52: I know this is still early fetch-and-log code, but do we have a way to force a user into an experiment?
For example, if Softvision or us are testing flows, is there a special flag or ripcord we can pull to force ourselves into a certain state for manual/automated testing?
Acceptable answers include: "yes", "no", "not yet", "what?", `<other>`, or colorful profanity if your choosing.

src/app/functions/server/getExperiments.ts
```diff
@@ -0,0 +1,34 @@
...
+  const serverUrl = process.env.NIMBUS_SIDECAR_URL;
+  if (!serverUrl) {
+    throw new Error("env var NIMBUS_SIDECAR_URL not set");
+  }
+
+  let features;
+  try {
+    features = await fetch(`${serverUrl}/v1/features/`, {
+      method: "POST",
+      body: JSON.stringify({ client_id: userId }),
```
yashikakhurana, line 26: `context `key is also needed, you can set `{}`, this is open API schema for the reference- https://github.com/mozilla/experimenter/blob/main/cirrus/server/cirrus/docs/openapi.json 
```javascript
{
  "client_id": "4a1d71ab-29a2-4c5f-9e1d-9d9df2e6e449",
  "context": {}
}

**Review by yashikakhurana — CHANGES_REQUESTED** [review]
overall looks good, one little thing I noticed to supply` context` key as well as part of sending a request

File: src/app/(nextjs_migration)/(guest)/layout.tsx [hunk 16]
```diff
@@ -2,23 +2,70 @@
...
+  let userId = "";
+
+  if (accountId && typeof accountId === "string") {
+    // If the user is logged in, use a UUID based on the user's subscriber ID.
+    // Note: we may want to use the FxA UID here, but we need approval for that first.
+    if (process.env.NIMBUS_UUID_NAMESPACE) {
+      userId = uuidv5(accountId, process.env.NIMBUS_UUID_NAMESPACE);
+    } else {
+      console.error("NIMBUS_UUID_NAMESPACE env var not set");
+    }
```
pdehaan, line 39: Not sure if this is helpful or makes sense, but my eyes are telling me this might be very close (or copypaste) of src/app/(nextjs_migration)/(authenticated)/user/layout.tsx#L33-44.
I dont know if it makes sense to abstract that out, or if that is just overoptimizing.

I think the only differences between the files/blocks is that this has an `else` block while the authenticated layout doesn't.

```js
function userIdFromSession(session) {
  const accountId = session?.user?.subscriber?.fxa_uid;
  let userId = "";

  if (accountId && typeof accountId === "string") {
    // If the user is logged in, use a UUID based on the user's subscriber ID.
    // Note: we may want to use the FxA UID here, but we need approval for that first.
    if (process.env.NIMBUS_UUID_NAMESPACE) {
      userId = uuidv5(accountId, process.env.NIMBUS_UUID_NAMESPACE);
    } else {
      console.error("NIMBUS_UUID_NAMESPACE env var not set");
    }
  }
  return userId;
}
```

which might change this to:
```js
let userId = userIdFromSession(session);
if (!userId) {
    // if the user is not logged in, use a cookie with a randomly-generated Nimbus user ID.
    const cookie = cookies().get("userId");
    if (cookie) {
      userId = cookie.value;
    } else {
      // TODO Cookies can only be set in server action or route handler
      // @see https://nextjs.org/docs/app/api-reference/functions/cookies#cookiessetname-value-options
      // cookies().set("userId", uuid);
      userId = `guest-${randomUUID()}`;
    }
}
```
I honestly don't know if that is any better, or if the code is slightly different.

src/app/(nextjs_migration)/(authenticated)/user/layout.tsx
rhelmer: We do have allow-lists for the current feature-flag implementation - you can use the UI at `/redesign/admin/feature-flags`. However that works by storing flag state and allow-list in the database, so would work for this "force a user into experiment" use case, however we should add more specific support for Nimbus (there's no need to use the database for this, it'll just be for the current session) https://mozilla-hub.atlassian.net/browse/MNTOR-2182

src/app/(nextjs_migration)/(guest)/layout.tsx
rhelmer: Yes they are pretty similar, but different enough I didn't feel like it was worth abstracting just that one piece - I also think that we'll probably change this if we get approval to use FxA UID (filed a ticket to track that)

pdehaan:
LGTM, but I'll let @yashikakhurana be The Decider™ re:Cirrus bits.

yashikakhurana (approved):
Awesome, looks good to me! I just wanted to make sure sidecar is available at `NIMBUS_SIDECAR_URL=http://localhost:6060` As soon as it gets deployed it should be able to make a call and we should start seeing some logs even without A/A, in this case since no experiment is available it should return default FML, let's just make sure once it is released to see the logs before launching A/A. Great job and thank you for working on this @rhelmer 👏 👏 👏

pdehaan (approved):

Vinnl (approved):
Thanks for linking (and creating) all the followup tickets!

Also, though it's probably too early at this point in time, it would be great to get some docs on e.g. how to debug Glean events, where to view them, etc.

src/app/(nextjs_migration)/(guest)/layout.tsx
```diff
@@ -2,23 +2,70 @@
...
   const l10n = getL10n();
 
+  // If the user is logged in, use UUID derived from FxA UID as Nimbus user ID.
+  const session = await getServerSession(authOptions);
+  const accountId = session?.user?.subscriber?.fxa_uid;
+  let userId = "";
+
+  if (accountId && typeof accountId === "string") {
+    // If the user is logged in, use a UUID based on the user's subscriber ID.
+    // Note: we may want to use the FxA UID here, but we need approval for that first.
```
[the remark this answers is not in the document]
rhelmer: I filed a ticket! I'll link it here.

src/app/(nextjs_migration)/(guest)/layout.tsx
```diff
@@ -2,23 +2,70 @@
...
+    } else {
+      console.error("NIMBUS_UUID_NAMESPACE env var not set");
+    }
+  } else {
+    // if the user is not logged in, use a cookie with a randomly-generated Nimbus user ID.
+    const cookie = cookies().get("userId");
+    if (cookie) {
+      userId = cookie.value;
+    } else {
+      // TODO Cookies can only be set in server action or route handler
```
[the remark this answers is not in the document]
rhelmer: Ah yeah that makes sense, thanks.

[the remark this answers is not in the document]
rhelmer: Added a TODO w/ link to https://mozilla-hub.atlassian.net/browse/MNTOR-2180

rhelmer:
> Awesome, looks good to me! I just wanted to make sure sidecar is available at `NIMBUS_SIDECAR_URL=http://localhost:6060` As soon as it gets deployed it should be able to make a call and we should start seeing some logs even without A/A, in this case since no experiment is available it should return default FML, let's just make sure once it is released to see the logs before launching A/A. Great job and thank you for working on this @rhelmer 👏 👏 👏

Thanks! Yes I am going to confirm the host/port for NIMBUS_SIDECAR_URL, the `.env-dist` is only an example which we'll need to override in the server environment, so I'll make sure this gets the correct setting. I'll let you know as soon as this hits our staging environment so we can look for logs.

rhelmer:
> Thanks for linking (and creating) all the followup tickets!
> 
> Also, though it's probably too early at this point in time, it would be great to get some docs on e.g. how to debug Glean events, where to view them, etc.

Good idea, thanks - I made a note to follow up on that.

PR #3411, opened 2023-09-14
MNTOR-2013 - Exposure Card Refinements

codemist:
# References: 
Jira: MNTOR-2013

# Description
This PR resolves a couple of exposure card refinements.
 - Adds extra padding to the expanded details in the exposure card
 - Increase width of the Fix it Button
 - Disables "let's keep going" CTAs for Foxfooding
 - Update counter for data classes, removing "_ found" after each data class
 

# Screenshot (if applicable)

image
image

# Checklist (Definition of Done)
- [ ] Localization strings (if needed) have been added.
- [ ] Commits in this PR are minimal and [have descriptive commit messages](https://chris.beams.io/posts/git-commit/).
- [ ] I've added or updated the relevant sections in readme and/or code comments
- [ ] I've added a unit test to test for potential regressions of this bug.
- [ ] Product Owner accepted the User Story (demo of functionality completed) or waived the privilege.
- [ ] All acceptance criteria are met.
- [ ] Jira ticket has been updated (if needed) to match changes made during the development process.
- [ ] Jira ticket has been updated (if needed) with suggestions for QA when this PR is deployed to stage.

PR #3424, opened 2023-09-18
MNTOR-2013 - Exposure card refinements (follow up)

codemist:
# References: 
Jira: MNTOR-2013

# Description

I had accidentally merged in [2013](https://github.com/mozilla/blurts-server/pull/3411) too early, this PR resolves some of the post-merge comments.

In this PR, I pulled out the `BreachExposureCategory` and `ScannedExposureCategory` now renamed `BreachDataClass` and `DataBrokerDataClass` so that they could share the same layout (and to generally chomp down on the `ExposureCard` component which was starting to balloon up).

# How to test


src/app/components/client/ExposureCardDataClass.tsx
```diff
@@ -0,0 +1,98 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+"use client";
+
+import { ReactElement } from "react";
+import styles from "./ExposureCard.module.scss";
```
Vinnl, line 8: Not sure how doable it is to extract them now, but the idea behind CSS modules is that one can safely modify those styles without unintentionally affecting styles elsewhere, so ideally, each component has its own CSS module containing only the styles relevant to that module.

src/app/components/client/ExposureCardDataClass.tsx
```diff
@@ -0,0 +1,98 @@
...
+      icon={props.icon}
+      label={props.exposureCategoryLabel}
+      count={emailLength}
+      emailData={emailsList}
+    />
+  );
+};
+
+type ExposureCardDataClassLayoutProps = {
+  type: "breach" | "broker";
```
Vinnl, line 71: Looks like this isn't even used/needed :smile: 

```suggestion
```

src/app/components/client/ExposureCardDataClass.tsx
```diff
@@ -0,0 +1,98 @@
...
+  icon: ReactElement;
+};
+
+export const DataBrokerDataClass = (props: DataBrokerDataClassProps) => {
+  const emailsList = (
+    <ul className={styles.emailsList}>
+      {props.scanResultData.emails.map((email: string, index: number) => (
+        <li key={index}>{email}</li>
+      ))}
+    </ul>
```
Vinnl, line 26: Since this HTML is (as far as I can see) exactly the same between breaches and data brokers, you could even consider moving that to `<ExposureDataClassLayout>`, and just passing it `props.scanResultData.emails` / `props.subscriberBreachData.emailsEffected`.

src/utils/subscriberBreaches.ts
```diff
@@ -27,7 +27,7 @@ export interface SubscriberBreach {
   modifiedDate: Date;
   name: string;
   title: string;
-  emailsEffected?: string[];
+  emailsEffected: string[];
```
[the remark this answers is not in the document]
Vinnl: Sounds good to me! And though it wasn't changed in this PR, a small nit: I think it should be `emailsAffected` :smile:

src/app/components/client/ExposureCardDataClass.tsx
```diff
@@ -0,0 +1,98 @@
...
+      <dt>
+        <span className={styles.exposureTypeIcon}>{props.icon}</span>
+        {l10n.getString("exposure-card-label-and-count", {
+          category_label: props.label,
+          count: props.count,
+        })}
+      </dt>
+      <dd>
+        {props.label === l10n.getString("exposure-card-email") &&
+          props.emailData}
```
Vinnl, line 94: I think this still results in an empty `<dd>` if this isn't displaying email addresses? A description list without descriptions sounds confusing, and [MDN mentions](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dl) under "Permitted content" that every `<dt>` should be followed by one or more `<dd>`s.

Vinnl (approved):
Thanks for cleaning this up, looking really good! I'd still like to see the empty `<dd>` for non-emails resolved - possibly, it's a bit too <abbr title="Don't Repeat Yourself">DRY</abbr> atm, and we just need to display different types of data in a different way?

PR #3430, opened 2023-09-20
Unset engine-strict option

Vinnl:
Dependabot was [logging the following error](https://github.com/mozilla/blurts-server/network/updates/723120103):

    Dependabot uses Node.js v18.17.1 and NPM 9.6.5.
    Due to the engine-strict setting, the update will not succeed.

Since we use Volta locally, and explicitly configure our Node versions elsewhere, I don't think engine-strict is necessary. And it would be good to get JS dependency updates again.

~~We might want to update our npm version to align with Dependabot though.~~ I added e3685adc0 to align our npm version with Dependabot - I expect that that's a relatively safe change, i.e. not as impactful as e.g. [changing our Node version](https://github.com/mozilla/blurts-server/pull/3295).

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,7 +4,7 @@
   "description": "Firefox Monitor",
   "engines": {
     "node": "18.12.x",
-    "npm": "8.x"
+    "npm": "9.6.5"
   },
   "type": "module",
   "scripts": {
@@ -46,7 +46,7 @@
   "license": "MPL-2.0",
   "volta": {
     "node": "18.12.1",
-    "npm": "8.19.3"
+    "npm": "9.6.5"
   },
   "dependencies": {
     "@aws-sdk/client-s3": "^3.360.0",
--- a/.npmrc
+++ b/.npmrc
@@ -1,1 +1,0 @@
-engine-strict=true
--- a/netlify.toml
+++ b/netlify.toml
@@ -18,4 +18,4 @@
   # Default build command.
   command = "npm ci; npm run build-storybook"
 
-  environment = { NODE_VERSION = "18.12.1", NPM_VERSION = "8.19.3" }
+  environment = { NODE_VERSION = "18.12.1", NPM_VERSION = "9.6.5" }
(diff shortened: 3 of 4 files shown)


rhelmer (approved):
Thanks for catching that!

mansaj (approved):

package.json
```diff
@@ -4,7 +4,7 @@
   "description": "Firefox Monitor",
   "engines": {
     "node": "18.12.x",
-    "npm": "8.x"
+    "npm": "9.6.5"
```
pdehaan, line 7: Since `node` above is using a wildcard for patch version, does it make sense to do that here too for npm?
I just `nvm`ed latest 18 and it looks like I get this:

```sh
nvm i 18

Downloading and installing node v18.18.0...
Downloading https://nodejs.org/dist/v18.18.0/node-v18.18.0-darwin-arm64.tar.xz...
Checksums matched!
Now using node v18.18.0 (npm v9.8.1)
```

If I explicitly install `nvm i 18.12` it looks like the latest patch version is 18.12.1 (https://nodejs.org/en/blog/release/v18.12.1 -- circa "04 Nov 2022"):

```sh
nvm i 18.12
v18.12.1 is already installed.
Now using node v18.12.1 (npm v8.19.2)
```

Not sure if we're using 18.12.x (which is  also pinned to 18.12.1 below in the Volta config) because thats what SRE is using on the deployed images or if we should be bumping Node versions too eventually.

Vinnl: Yeah it's a good question. The Dependabot error message mentioned specifically this version. Generally I'd prefer to have everyone on exactly the same version, and ideally, that pinned version is regularly updated - for which, of course, we use Dependabot :sweat_smile: 

But of course, Dependabot doesn't keep track of npm versions. Generally, though, I don't think there's much reason to keep up with the very latest version of npm unless you're running into a specific bug, and even in that case we'll have to explicitly pin to that at least version (because your local npm doesn't update by itself), so I think pinning to a specific version is the lesser of two evils.

And then there's also the fact that the `engines` field doesn't really do all that much for us, since we're not an upstream library, and we control the toolchain versions everywhere - so I'd even be inclined to remove it altogether.

PR #3423, opened 2023-09-18
Bring structure to our dashboard Storybook stories

Vinnl:
# References: 
Jira: MNTOR-2015
Figma: N/A

# Description

Note: this builds on https://github.com/mozilla/blurts-server/pull/3405. Also, 100% coverage should be fixed by https://github.com/mozilla/blurts-server/pull/3407.

This allows us to view (and test) all relevant permutations of user data on the dashboard. Specifically:

- Is the user in the US or outside of it?
- Do they have Premium or not?
- Have they run a data broker scan yet? If yes, did they get results, and if yes, did they resolve them all?
- Did they have breaches and, if yes, did they resolve them all?

There should be a story for every possible combination. And because those permutations can be controlled through the `<DashboardWrapper>`'s props, it should be easier to write unit tests that verify a particular property, and that it holds under every permutation of the other properties.

To make this easier, I also removed the `bannerData` prop from the dashboard, since it can calculate that itself from the other props it gets passed.

It also uncovered a number of issues, but to limit the scope of this PR, I propose filing separate tickets and fixing those individually. With the exception of 799fd223b: scan results with `optout_in_progress` should not be listed under "Needs action".

# Screenshot (if applicable)

image

# How to test

Open the Storybook stories, verify that the data matches. The dashboard will have a bunch of things that do not match the data, but I propose fixing that in future PRs.

# Checklist (Definition of Done)
- [x] Localization strings (if needed) have been added.
- [x] Commits in this PR are minimal and [have descriptive commit messages](https://chris.beams.io/posts/git-commit/).
- [x] I've added or updated the relevant sections in readme and/or code comments
- [ ] I've added a unit test to test for potential regressions of this bug.
- [ ] Product Owner accepted the User Story (demo of functionality completed) or waived the privilege.
- [ ] All acceptance criteria are met.
- [ ] Jira ticket has been updated (if needed) to match changes made during the development process.
- [ ] Jira ticket has been updated (if needed) with suggestions for QA when this PR is deployed to stage.

PR #3431, opened 2023-09-20
MNTOR-2009: Indicators for pending broker scan

flozia:
# References: 
Jira: [MNTOR-2009](https://mozilla-hub.atlassian.net/browse/MNTOR-2009)

# Description

Adds indicators and messaging to the dashboard for when the broker scan is still pending.

# Screenshot

| Dashboard no breaches | Dashboard with breaches |
| --- | --- |
| mntor-2009-no-breaches | mntor-2009-with-breaches |

[MNTOR-2009]: https://mozilla-hub.atlassian.net/browse/MNTOR-2009?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/app/(proper_react)/redesign/(authenticated)/user/dashboard/DashboardTopBanner.tsx
```diff
@@ -54,13 +61,52 @@ export const DashboardTopBanner = (props: DashboardTopBannerProps) => {
   );
 
   const contentData = {
+    ScanInProgressContent: {
```
Vinnl, line 64: Hoo boy, this is going to conflict so hard with https://github.com/mozilla/blurts-server/pull/3405 :cry:

locales-pending/premium.ftl
```diff
@@ -249,16 +263,36 @@ dashboard-top-banner-your-data-is-protected-all-fixed-description =
...
 dashboard-exposures-all-fixed-label = All fixed here!
 
+
+dashboard-exposures-area-headline = View all sites where your info is exposed
+dashboard-fixed-area-headline = View all exposures that are fixed or in-progress
+
+# Variables:
+# $exposures_total_num is the total number of exposures the user has.
+# $data_breach_total_num is the total number of data breaches the user has.
+dashboard-exposures-breaches-scan-progress-description = {
```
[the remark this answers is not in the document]
flozia: Asked that [question and content got back](https://www.figma.com/file/CaEKIhvSJqf6KNIMzSkt40?node-id=5356:908&mode=design#561715562). During the UX/content office hours, they also said that they already iterated through some variations and preferred the current version.

As an additional note: This string is only for Premium and will not be translated for now.

src/app/(proper_react)/redesign/(authenticated)/user/dashboard/DashboardTopBanner.tsx
```diff
@@ -24,9 +26,9 @@ export type BannerContent =
 export type DashboardTopBannerProps = {
   content: BannerContent;
   bannerData: DashboardSummary;
-  hasRunScan: boolean;
   isEligibleForFreeScan: boolean;
   type: TabType;
+  scanStatus?: Scan["status"];
```
Vinnl, line 31: If we're always passing this, why would we make it optional? (Likewise for the same prop to `<View>`.)

locales-pending/premium.ftl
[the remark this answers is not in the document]
flodolo: My problem is with the "for now" part 🤔

src/app/(proper_react)/redesign/(authenticated)/user/dashboard/Dashboard.stories.tsx
```diff
@@ -336,6 +342,61 @@ export const DashboardPremiumUser: Story = {
...
             PremiumBrokerRemoval: true,
           }}
+          scanStatus="finished"
+        />
+      </Shell>
+    );
+  },
+};
+
+export const DashboardNoBreachesScanInProgress: Story = {
```
Vinnl, line 352: Ooh, another permutation option... This is going to conflict with https://github.com/mozilla/blurts-server/pull/3423, where it would be implemented as another entry in `brokerOptions` :cry:

src/app/components/client/Chart.tsx
```diff
@@ -14,16 +16,19 @@ import { Button } from "../server/Button";
...
 export type Props = {
   data: Array<[string, number]>;
-  hasRunScan: boolean;
   isEligibleForFreeScan: boolean;
+  scanStatus?: Scan["status"];
 };
 
-export const DoughnutChart = (props: Props) => {
+export const DoughnutChart = ({
+  data,
```
Vinnl, line 28: Is this a plugin you have that destructures all the props? Because if so, we probably want to standardise on that - it makes it harder to see the actual changes in the PR.

src/app/components/client/Chart.tsx
```diff
@@ -100,14 +105,32 @@ export const DoughnutChart = (props: Props) => {
...
 
-  const prompt = (
-    <div className={styles.prompt}>
-      <p>{l10n.getString("exposure-chart-returning-user-upgrade-prompt")}</p>
-      <Link href="/redesign/user/welcome">
-        {l10n.getString("exposure-chart-returning-user-upgrade-prompt-cta")}
-      </Link>
-    </div>
-  );
+  const scanInProgressPrompt = <div className={styles.prompt}></div>;
```
Vinnl, line 108: Doesn't this mean that this is always truthy, for line 124? And what is the value even used for?

Maybe it makes more sense to make it default to `null`, then extract the function body from `getPrompt` to run anyway and assign to this variable, and then just check this variable instead of running the function twice on line 217?

Vinnl (approved):
From your "no breaches" screenshot, underneath "Scan in progress" it says "Next start your free scan..." - that seems a little confusing or possibly wrong?

Other than that, no major things jump out, except that the implementation is building on things that I reworked in two of my PRs :( They're now unblocked, so hopefully I can merge them soon and avoid more instances of this, but I'm wondering which will cause less hassle if we merge it first :thinking:

src/app/components/client/Chart.tsx
flozia: Seems like I left this line there unintentionally — updated in [746b6a5](https://github.com/mozilla/blurts-server/pull/3431/commits/746b6a5357a5403d969ecf6aab886d18e197359d).

locales-pending/premium.ftl
```diff
@@ -57,6 +57,7 @@ exposure-chart-legend-value-nr = { $nr }×
 exposure-chart-caption = This chart shows how many times your info is actively exposed.
 exposure-chart-returning-user-upgrade-prompt = Home address, family members and more are not yet included.
 exposure-chart-returning-user-upgrade-prompt-cta = Start a free scan
+exposure-chart-scan-in-progress-prompt = <b>Scan in progress:</b> address, family members, and more are not yet included.
```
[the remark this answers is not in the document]
flodolo: Can you explain why you find it confusing?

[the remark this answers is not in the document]
Vinnl: I can try, but it's hard :sweat_smile: Let's try with [a quote](https://en.wikipedia.org/wiki/Colon_(punctuation)):

> A colon often precedes an explanation, a list, or a quoted sentence

This is not a list, nor a quoted sentence. I think it's supposed to function as explanation, but "address, family members, and more are not yet included" does not really explain what "Scan in progress" is or means. Instead, it primarily adds more detail, which I think would usually use a different punctuation mark in regular prose? But then in the context I _do_ understand how it's used - I'm just not sure why.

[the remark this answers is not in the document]
flodolo: Thanks, clearer now. That might be actually a question for content, more than localization. In these cases, I would probably expect a — instead of :

src/app/(proper_react)/redesign/(authenticated)/user/dashboard/DashboardTopBanner.tsx
flozia: Right, it’s not optional anymore.

src/app/(proper_react)/redesign/(authenticated)/user/dashboard/Dashboard.stories.tsx
flozia: Added `scan-in-progress` to the `brokerOptions` in https://github.com/mozilla/blurts-server/pull/3431/commits/745a73bbed958dd92fba3f43520c0c27622cc404.

src/app/components/client/Chart.tsx
flozia: No plugin. After reaching a certain threshold of props — usually more than two or three — I’m starting to destruct props rather than having to access `props` every time. That’s just personal preference and I get that this is adding noise. Reverted in [87f7f85](https://github.com/mozilla/blurts-server/pull/3431/commits/87f7f85f62d78fffee3da4603775f9331c27c872).

Vinnl: Oh haha well, not sure if that was worth a full revert :sweat_smile: Another option for next time would be to make that change in a separate commit and call that out in the PR description, then I can just look at the other commits while reviewing. (Assuming there's not too many of them, otherwise it's a bit of a hassle.)

src/app/(proper_react)/redesign/(authenticated)/user/dashboard/DashboardTopBanner.tsx
flozia: It did conflict very hard. Should be resolved now. :)

src/knex-tables.d.ts
```diff
@@ -176,6 +176,7 @@ declare module "knex/types/tables" {
     onerep_profile_id: number;
     onerep_scan_id: number;
     onerep_scan_reason: Scan["reason"];
+    onerep_scan_status: Scan["status"];
```
Vinnl, line 179: This might be a premature copy from https://github.com/mozilla/blurts-server/pull/3436?

(Though I suppose Git will automatically resolve that, maybe.)

Vinnl (approved):
Thanks for getting this up-to-date with `main`! And also for removing all the `c8` ignore comments, love to see that :tada:

src/knex-tables.d.ts
flozia: Yes, that’s already in anticipation of the changes in #3436 — otherwise the linter will complain [here](https://github.com/mozilla/blurts-server/pull/3431/files#diff-d701455e388547a86562525e090b5bd1be6d2f06cee5aa5187d602f12253640dR72).

src/app/functions/server/getRelevantGuidedSteps.ts
```diff
@@ -191,6 +199,10 @@ function hasCompleted(data: InputData, stepId: StepLink["id"]): boolean {
     return data.latestScanData?.scan !== null;
   }
 
+  if (stepId === "ScanInProgress") {
+    return data.latestScanData?.scan?.onerep_scan_status === "finished";
+  }
```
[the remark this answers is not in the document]
flozia: Thanks for catching that. Just making sure: [d170aa4](https://github.com/mozilla/blurts-server/pull/3431/commits/d170aa4d36b668f1d040ade99605c3126fbc4f19) you suggested?

src/app/(proper_react)/redesign/(authenticated)/user/dashboard/Dashboard.stories.tsx
```diff
@@ -98,12 +109,25 @@ const DashboardWrapper = (props: DashboardWrapperProps) => {
     createRandomScanResult({ status: "new", manually_resolved: true }),
   ];
 
+  const mockedOnlyUnresolvedScanResults: OnerepScanResultRow[] = [
```
Vinnl, line 112: Heh, just for my interest: why does it matter that there are _only_ unresolved scan results, i.e. no resolved scan results? IIRC for the other tests it doesn't really matter, so another option would be to just remove the `...mockedResolvedScanResults` from `mockedUnresolvedScanResults`.

src/app/functions/server/getRelevantGuidedSteps.ts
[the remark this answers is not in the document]
Vinnl: Partially! But also the removal of the whole `ScanInProgress` ID (and thus also its reference in `isEligible`).

src/app/functions/server/getRelevantGuidedSteps.ts
```diff
@@ -21,6 +21,10 @@ export const stepLinks = [
     href: "/redesign/user/dashboard/fix/data-broker-profiles/start-free-scan",
     id: "StartScan",
   },
+  {
+    href: "/redesign/user/dashboard/fix/data-broker-profiles/view-data-brokers",
+    id: "ScanInProgress",
+  },
```
Vinnl, line 27: (From the other comment.)

```suggestion
```

src/app/functions/server/getRelevantGuidedSteps.ts
```diff
@@ -141,6 +145,10 @@ function isEligibleFor(data: InputData, stepId: StepLink["id"]): boolean {
     return data.countryCode === "us";
   }
 
+  if (stepId === "ScanInProgress") {
+    return data.countryCode === "us";
+  }
+
```
Vinnl, line 151: (From the other comment.)

```suggestion
```

src/app/(proper_react)/redesign/(authenticated)/user/dashboard/DashboardTopBanner.tsx
```diff
@@ -141,6 +142,48 @@ const Content = (props: {
     );
   }
 
+  if (relevantGuidedStep.id === "ScanInProgress") {
```
Vinnl, line 145: And so here you wouldn't make this conditional on the step (since it's not a different step in the guided experience flow), but on the actual prerequisite for showing this content:

```suggestion
  if (
    relevantGuidedStep.id === "ScanResult" &&
    props.stepDeterminationData.latestScanData?.scan?.onerep_scan_status ===
      "in_progress"
  ) {
```

Vinnl (approved):

src/app/(proper_react)/redesign/(authenticated)/user/dashboard/DashboardTopBanner.tsx
flozia: Got it — removed the step in [f0213a3](https://github.com/mozilla/blurts-server/pull/3431/commits/f0213a332d5120e489fc10fc112be1f9506ae62d).

src/app/(proper_react)/redesign/(authenticated)/user/dashboard/Dashboard.stories.tsx
flozia: This does not really matter to the broker scan-in-progress stories either. My thinking was that if there are exposures coming in during an initial scan only having unresolved ones is the more likely case. But in that case: Let’s keep it simple — removed in [9860df8](https://github.com/mozilla/blurts-server/pull/3431/commits/9860df837463406ce6ebcb16ec4b0323bbb2bb2f).

locales-pending/premium.ftl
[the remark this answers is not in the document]
flozia: I [ran this by content](https://www.figma.com/file/CaEKIhvSJqf6KNIMzSkt40?node-id=5356:908&mode=design#565954658) and one suggestion was to split this up into two sentences. We decided to go with the current version.

locales-pending/premium.ftl
```diff
@@ -196,6 +197,19 @@ dashboard-exposures-filter-show-results = Show results
 dashboard-exposures-filter-reset = Reset
 
 
+dashboard-top-banner-scan-in-progress-title = Your scan is still in progress
+# Variables:
+# $data_breach_total_num is the total number of data breaches the user has.
+# $data_broker_total_num is the total number of data brokers selling the user’s data.
```
pdehaan, line 203: Is this `$data_broker_total_num` variable used?

flozia: Good catch — it’s not. Removed the comment in [f720dea](https://github.com/mozilla/blurts-server/pull/3431/commits/f720deae382fd2820ea384d686f8586bff84c59c).

locales-pending/premium.ftl
```diff
@@ -196,6 +197,19 @@ dashboard-exposures-filter-show-results = Show results
...
+dashboard-top-banner-scan-in-progress-description =
+  { $data_breach_total_num ->
+      [one] We found <b>{ $data_breach_total_num }</b> exposure so far, but we’re still scanning sites that sell your personal info. This should be done within 3 minutes.
+      *[other] We found <b>{ $data_breach_total_num }</b> exposures so far, but we’re still scanning sites that sell your personal info. This should be done within 3 minutes.
+  }
+dashboard-top-banner-scan-in-progress-fix-now-hint = You can refresh this page then, or start fixing your data breaches now.
+dashboard-top-banner-scan-in-progress-fix-later-hint = You can refresh this page then or come back later.
```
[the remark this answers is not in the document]
flozia: Thanks! These were copy and paste. Added in [2a229a7](https://github.com/mozilla/blurts-server/pull/3431/commits/2a229a7b7dd13c1d0ad20eeaf059f1716c0dca71).

PR #3427, opened 2023-09-19
MNTOR-1698: Welcome to Premium

mansaj:
# References: 
Jira: MNTOR-1698
Figma: https://www.figma.com/file/CaEKIhvSJqf6KNIMzSkt40/Concepts-for-Monitor-MVP-Redesign?type=design&node-id=1290-57329&t=LMpwec8NLhVLuzS7-4

# Description
Welcome to Premium screen

# Screenshot (if applicable)

# How to test
http://localhost:6060/redesign/user/dashboard/fix/data-broker-profiles/welcome-to-premium

# Checklist (Definition of Done)
- [ ] Localization strings (if needed) have been added.
- [ ] Commits in this PR are minimal and [have descriptive commit messages](https://chris.beams.io/posts/git-commit/).
- [ ] I've added or updated the relevant sections in readme and/or code comments
- [ ] I've added a unit test to test for potential regressions of this bug.
- [ ] Product Owner accepted the User Story (demo of functionality completed) or waived the privilege.
- [ ] All acceptance criteria are met.
- [ ] Jira ticket has been updated (if needed) to match changes made during the development process.
- [ ] Jira ticket has been updated (if needed) with suggestions for QA when this PR is deployed to stage.

PR #3432, opened 2023-09-20
Remove re-use of CSS Modules across components

Vinnl:
The idea behind CSS modules is to scope them to single components, so that it is more predictable which styles will be affected when you change the CSS.

Additionally, the <Button> component actually supports rendering <Link>s, so we can just use that.

Note: this is branched off https://github.com/mozilla/blurts-server/pull/3427 since that also used this.


flozia (approved):
Looks good — according to the build failure it seems like we’ll need to convert the `Button` component to a `client component`, though.

mansaj:
I was confused by the link/button diff. Thanks!

Vinnl:
Ah good point @flozia, added in bf03f55b6.

@codemist Use `<Button>` if you want it to look like a primary or secondary button :) It will automatically use a `<Link>` or a `<button>` behind the scenes depending on if you pass it an `href` or not.

PR #3436, opened 2023-09-21
Store scan status in our database

Vinnl:
The website has to display different content (https://github.com/mozilla/blurts-server/pull/3431) depending on whether the scan is already finished, or still in progress, so we need to keep track of that too.

# References: 
Jira: MNTOR-2009

### DIFF
--- a/src/app/(proper_react)/redesign/(authenticated)/user/dashboard/subscribed/page.tsx
+++ b/src/app/(proper_react)/redesign/(authenticated)/user/dashboard/subscribed/page.tsx
@@ -48,13 +48,14 @@ export default async function Subscribed() {
   const scans = await getAllScanResults(profileId);
 
   // In dev mode, record scans every time this page is reloaded.
-  // The webhoook does this in production.
+  // The webhook does this in production.
   if (dev) {
     await addOnerepScanResults(
       profileId,
       latestScan.scan.onerep_scan_id,
       scans,
-      "initial"
+      "initial",
+      latestScan.scan.onerep_scan_status
     );
   }
 
--- a/src/app/(proper_react)/redesign/(authenticated)/user/dashboard/subscribed/page.tsx
+++ b/src/app/(proper_react)/redesign/(authenticated)/user/dashboard/subscribed/page.tsx
@@ -54,7 +54,7 @@ export default async function Subscribed() {
       profileId,
       latestScan.scan.onerep_scan_id,
       scans,
-      "manual",
+      "initial",
       latestScan.scan.onerep_scan_status
     );
   }
--- a/src/app/(proper_react)/redesign/(authenticated)/user/dashboard/subscribed/page.tsx
+++ b/src/app/(proper_react)/redesign/(authenticated)/user/dashboard/subscribed/page.tsx
@@ -54,7 +54,7 @@ export default async function Subscribed() {
       profileId,
       latestScan.scan.onerep_scan_id,
       scans,
-      "initial",
+      "manual",
       latestScan.scan.onerep_scan_status
     );
   }
--- a/src/db/tables/onerep_scans.ts
+++ b/src/db/tables/onerep_scans.ts
@@ -62,18 +62,6 @@ async function setOnerepManualScan(
   });
 }
 
-async function updateOnerepScanStatus(
-  onerepScanId: number,
-  onerepScanStatus: Scan["status"]
-) {
-  await knex("onerep_scans").where("onerep_scan_id", onerepScanId).update({
-    onerep_scan_status: onerepScanStatus,
-    // @ts-ignore knex.fn.now() results in it being set to a date,
-    // even if it's not typed as a JS date object:
-    updated_at: knex.fn.now(),
-  });
-}
-
 async function addOnerepScanResults(
   onerepProfileId: number,
   onerepScanId: number,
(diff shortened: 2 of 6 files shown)


src/app/(proper_react)/redesign/(authenticated)/user/dashboard/subscribed/page.tsx
```diff
@@ -48,13 +48,14 @@ export default async function Subscribed() {
   const scans = await getAllScanResults(profileId);
 
   // In dev mode, record scans every time this page is reloaded.
-  // The webhoook does this in production.
+  // The webhook does this in production.
   if (dev) {
     await addOnerepScanResults(
```
Vinnl, line 53: @rhelmer I don't think this function (or even page) is in use anywhere, but it looks weird to me: we first fetch the scan results from the database using `getLatestOnerepScanResults`, and then insert those back into the database using `addOnerepScanResults`, if I understand it correctly.

src/db/tables/onerep_scans.ts
```diff
@@ -78,6 +81,7 @@ async function addOnerepScanResults(
         onerep_profile_id: onerepProfileId,
         onerep_scan_id: onerepScanId,
         onerep_scan_reason: onerepScanReason,
+        onerep_scan_status: onerepScanStatus,
```
flozia, line 84: When running the scan locally or on dev we would not update the status with the polling method in order to workaround not having the webhooks available in `src/app/api/v1/user/welcome-scan/progress/route.ts`. In order to update the entry `onerep_scan_status` in the table `onerep_scan` we’ll need to update the status after `L77` I think and add something along the lines of:

```
await transaction("onerep_scans")
  .where("onerep_scan_id", onerepScanId)
  .update({
    onerep_scan_status: onerepScanStatus,
    updated_at: knex.fn.now(),
  });
```

Does that look correct @mansaj?

flozia (commented):
Thanks for adding the scan status. For local and dev we’ll the status is currently not being updated after running a scan. Suggested a small addition that should take care of this.

src/db/tables/onerep_scans.ts
Vinnl: Oh hmm... We have `getLatestOnerepScanResults`, which just gets the latest entry, so we could make it an append-only column, which I think was Joe's idea. But I also just realised that I made the `onerep_scan_id` unique, so if it should be append-only, I should probably update that as well.

src/app/(proper_react)/redesign/(authenticated)/user/dashboard/subscribed/page.tsx
rhelmer: This page is linked to from the subscription flow directly right now (subplat is skipped, but it would be the callback page for subplat). However it's not final, it's something I put together a month ago so we could test if removal is working on stage/prod without having to go through subplat yet.

Doesn't this do `const scans = await getAllScanResults(profileId);` above (fetching the results from OneRep's API) and then add these to the database (because dev can't have webhooks)?

rhelmer (commented):
This looks good to me overall, does the dev and subscription page make more sense now? If not let me know, I'm sure it can be improved but I can explain how it currently is supposed to work if it's not obvious.

src/app/(proper_react)/redesign/(authenticated)/user/dashboard/subscribed/page.tsx
Vinnl: Oh! OK, yeah, now I get it - so it does fetch the latest scan from the database, but only to obtain the scan ID, and then inserts the (newly fetched) scan results linked to that ID.

So I guess in that case, it should either:

- update the existing `onerep_scans` row to update its status, or
- add a new `onerep_scans` row that the new results would be linked to.

which I guess is another version of [this question](https://github.com/mozilla/blurts-server/pull/3436#discussion_r1334428132) for @mansaj.

src/db/tables/onerep_scans.ts
mansaj: @Vinnl Ok I think I understand the issue here. In order for us to keep `onerep_scan_id` unique, we cannot make the table append only? Because two scan results would share the same scan ids?

If that's the case.. I think we'd probably still prefer append only for the `scan results` table (thus `onerep_scan_id` is no longer unique). I think it would be a lot easier for insert / retrieval, and our `getLatest` function already gets the latest. We also maintain a better record of the history of changes (if we need that later for audit purpose)... There are some of the pros I can think of

Vinnl: @mansaj 

> In the `onerep_scans` table, the scan ids are still unique though right?

Not if it's append-only, because we need to know the latest status of a scan. So if it changes from `in_progress` to `complete`, we'd need to create a new row.

mansaj: Thinking out loud here.. I guess I'm wondering if it's silly idea to have `onerep_scans` be unique/update, and `onerep_scan_results` be append only?

src/app/(proper_react)/redesign/(authenticated)/user/dashboard/subscribed/page.tsx
```diff
@@ -48,13 +48,14 @@ export default async function Subscribed() {
   const scans = await getAllScanResults(profileId);
 
   // In dev mode, record scans every time this page is reloaded.
-  // The webhoook does this in production.
```
pdehaan, line 51: I, for one, will be sad for this typo to disappear.
I like to imagine it's more of a battle cry, webhoooooook!

src/db/tables/onerep_scans.ts
Vinnl: Yeah that works for me. I'll keep it as unique then :) Anyone want to approve the PR? :P

flozia (approved):
Thanks for adding the scan status!

src/app/(proper_react)/redesign/(authenticated)/user/dashboard/subscribed/page.tsx
rhelmer: > Oh! OK, yeah, now I get it - so it does fetch the latest scan from the database, but only to obtain the scan ID, and then inserts the (newly fetched) scan results linked to that ID.

rhelmer: I was pretty torn on this for the initial/existing implementation - in general I'd prefer to do things immutably (e.g. prefer `INSERT` over `UPDATE`), but `UPDATE` seemed natural for updating an existing scan.

mansaj (approved):
Don’t see any blocking issues. Would like to see if @rhelmer can think of anything else  re the update vs append-only trade off

src/db/tables/onerep_scans.ts
flozia: @Vinnl Merged #3470 in here.

rhelmer (approved):
@mansaj and I discussed this quite a bit yesterday - I think that it makes sense in this case since we're updating our record of an existing scan.

The problem in general with mutation is that it erases history (we know when it was created and last updated, but not every step along the way), but we're going to start emitting events when changes like this happen (starting with https://mozilla-hub.atlassian.net/browse/MNTOR-1404).

PR #3441, opened 2023-09-22
send localized breach alerts

rhelmer:
# References: 
Jira: MNTOR-2120

# Description

The previous localization method wasn't working as expected because it was using [AsyncLocalStorage](https://nodejs.org/api/async_context.html#class-asynclocalstorage), which requires a chain of asynchronous operations. This worked fine when this was all running under an HTTP request (Express, and then tested under Cloud Functions), but as a standalone Node cron job this doesn't work as expected (`getStore()` always returns `undefined`).

Note: I noticed this when testing against stage, the feature flag allowing new breaches on prod is currently paused, I'd like to get this pushed to prod before we re-enable.

# How to test

- create an account in your local instance using browser locale set to something other than English (stored as `signup_language` in `subscribers` table)
- have a known `sha1` hash of some test email addresses in your local database, for instance `test@example.com` is `567159d622ffbb50b11b0efd307be358624a26ee` (but it should be an email address you can check the inbox of)
- run the Pub/Sub emulator per the README
- `curl -v -d '{ "breachName": "000webhost", "hashPrefix": "56715", "hashSuffixes": ["9d622ffbb50b11b0efd307be358624a26ee"] }' -H "Authorization: Bearer ${HIBP_NOTIFY_TOKEN}" http://localhost:6060/api/v1/hibp/notify`
- `node src/scripts/emailBreachAlerts.js` <- this will email `test@example.com`


src/scripts/emailBreachAlerts.js
```diff
@@ -162,33 +163,38 @@ export async function poll(subClient, receivedMessages) {
           : [];
         /* c8 ignore stop */
 
-        const availableLanguages = process.env.SUPPORTED_LOCALES;
+        const availableLanguages = process.env.SUPPORTED_LOCALES.split(",");
```
Vinnl, line 166: Oh, TIL that we have this env var. That's bound to get outdated at some point :sweat_smile: FWIW, on the website, [we check](https://github.com/mozilla/blurts-server/blob/1faee2378f0b9bcc514d08768dd81a18b08fe333/src/app/functions/server/l10n.ts#L29-L33) the subdirectories of `/locales/`, so we should be able to do something similar with `readdir` here. Out of scope for this PR I suppose, but wanted to get that out of my brain :)

Vinnl (commented):
Hmm, I'm not very familiar with `AsyncLocalStorage`, and I don't see the `getStore` call in the previous version of `emailBreachAlerts.js` - so I think to be safe, I'll leave this for @mansaj to review, if you don't mind?

src/scripts/emailBreachAlerts.js
rhelmer: Yes, this is all based on the old Express app code... we really need to refactor it, the email templates are all written in a way to support both, we could surely simplify this.

rhelmer:
> Hmm, I'm not very familiar with `AsyncLocalStorage`, and I don't see the `getStore` call in the previous version of `emailBreachAlerts.js` - so I think to be safe, I'll leave this for @mansaj to review, if you don't mind?

Thanks, I'm not super familiar with it either - previously it was initialized here: https://github.com/mozilla/blurts-server/blob/65792b76b68a1b3d78cb80b36c76e428af13532a/src/app.js#L136-L147

src/scripts/emailBreachAlerts.js
pdehaan: Ah, so `require.context()` isn't really a _Node_ thing, it's a _Webpack_ thing? https://webpack.js.org/guides/dependency-management/#requirecontext

Also, not sure what checking the subdirectories of `/locales/` would do, that'd just tell you available locales, not necessarily locales with 80%+ translations (which I think was the intent of `SUPPORTED_LOCALES`). Previously supported locales with high enough translation %%s were maintained in a separate SRE repo which was a bit awkward to maintain and required somebody to list added/removed locales in deploy tickets.

Vinnl: Yep! And oh, that's a good point! I'd still like for that to be automated some way, but in lieu of that... That means that we are now shipping locales with <80% coverage as well :/

pdehaan: Re: automation,
As with most things in life, I'd say @flodolo is much better equipped to answer that than I am.
But for automation, I built https://github.com/pdehaan/pontoonql a lifetime ago to query Pontoon's GraphQL endpoints for translation stats for a project.

```sh
npx pdehaan/pontoonql firefox-monitor-website 80

[{"locale":{"code":"zh-CN","name":"Chinese (China)"},"totalStrings":783,"approvedStrings":783,"stringsWithWarnings":0,"missingStrings":0,"progress":100}, ...]
```

That might be too verbose to be useful, so I tend to pipe that through `jq` to extract just the `locale.code`:

```sh
npx pdehaan/pontoonql firefox-monitor-website 80 | jq '[.[] | .locale.code]'
[
  "zh-CN",
  "zh-TW",
  "cs",
  "da",
  "nl",
  "en-CA",
  "fr",
  "de",
  "it",
  "pt-BR",
  "ru",
  "sk",
  "sl",
  "es-AR",
  "es-CL",
  "es-ES",
  "sv-SE",
  "uk",
  "cy",
  "fy-NL",
  "hu",
  "el",
  "ia",
  "vi",
  "en-GB",
  "fi",
  "gn",
  "tr",
  "id",
  "sq",
  "es-MX",
  "nn-NO",
  "pt-PT"
]
```

Although probably caveats caveats everywhere since I'm pulling from Pontoon and its probably missing reference locales, etc.

Not sure if there is a good way to add that into the pipeline to keep it up to date, or it might be suboptimal if a translation gets temporarily bumped down to 79.8% on a deploy day and gets removed from the `SUPPORTED_LOCALES` list until the next deploy in 2 weeks. Not sure if manual judgement is better (although equally fallible).

mansaj (approved):

PR #3442, opened 2023-09-22
Enable Premium features on Netlify

Vinnl:
The main way you'll currently notice the lack of env vars is in the Premium badge not being visible on Netlify, because PREMIUM_ENABLED is not true.

Which does bring up the question... Why do we even have this `PREMIUM_ENABLED` var in the first place?

### DIFF
--- a/netlify.toml
+++ b/netlify.toml
@@ -16,6 +16,6 @@
   publish = "storybook-static/"
 
   # Default build command.
-  command = "npm ci; npm run build-storybook"
+  command = "npm ci; cp .env-dist .env; npm run build-storybook"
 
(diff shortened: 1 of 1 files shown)


flozia (approved):
The first time `PREMIUM_ENABLED` was added was showing and testing the [Subscribe to Premium](https://github.com/mozilla/blurts-server/blob/main/src/app/(nextjs_migration)/(authenticated)/user/breaches/page.tsx#L127) button on the current prod site. This was before we had the feature flags.

Vinnl:
OK, cool, sounds like we should be able to remove that now then, right?

PR #3443, opened 2023-09-22
Do not merge: debugging Netlify deploy

Vinnl:
Looks like https://github.com/mozilla/blurts-server/pull/3442 didn't work, so I'm going to push a couple of commits and see if I can add debug logging.

### DIFF
--- a/netlify.toml
+++ b/netlify.toml
@@ -16,6 +16,6 @@
   publish = "storybook-static/"
 
   # Default build command.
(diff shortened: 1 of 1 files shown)


Vinnl:
Ah haha, looks like `PREMIUM_ENABLED` isn't set to true by default in our `.env-dist`. I'll just remove the var altogether.

PR #3445, opened 2023-09-22
Format date according to a user's locale

Vinnl:
As called out in https://github.com/mozilla/blurts-server/pull/3427#issuecomment-1729707101, we hardcoded `en-US` here, rather than relying on the user's locale. d500cfac7b72e192480233ea85ca758463faa50d makes that dependent on the user's locale - the main downside is that that locale might not match the language of the website, since this part of the website is forced in English.

This PR also includes 4721768add9bcbfcde33ee3289040f1edab3fd89, which allows calling `getLocale` in client components - but does now require passing in the `l10n` instance.


flodolo (approved):
This is more a r+ on the idea than the implementation, since JS code is really not my area.

Vinnl:
@flodolo That was primarily what I was hoping for, especially w.r.t. the part where the date format might not align with the language :)

flozia (approved):
Looks good and works as expected!

PR #3446, opened 2023-09-22
bump max email worker jobs to 10k, split out to env var

rhelmer:
# References: 
Jira: MNTOR-2120

# Description

The email breach alert worker is taking a little over a minute to process 1k no-op jobs (they are valid breaches, but in the spam category and not eligible for sending email). I'd like to try bumping up the max number of messages the workers process from 1k to 10k.

Once we are able to measure how long it takes when there are actual emails to construct and send, we could consider upping the frequency (it runs once per hour right now), at the current rate it won't process the current backlog before it expires so we should do this before Thursday Sept 28 (and/or increase the retention period).

### DIFF
--- a/src/scripts/emailBreachAlerts.js
+++ b/src/scripts/emailBreachAlerts.js
@@ -42,7 +42,9 @@ const checkInId = Sentry.captureCheckIn({
 });
 
 // Only process this many messages before exiting.
+/* c8 ignore start */
 const maxMessages = parseInt(process.env.EMAIL_BREACH_ALERT_MAX_MESSAGES || 10000);
+/* c8 ignore stop */
 const projectId = process.env.GCP_PUBSUB_PROJECT_ID;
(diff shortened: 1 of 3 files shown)


mansaj (approved):

src/scripts/emailBreachAlerts.js
```diff
@@ -42,7 +42,7 @@ const checkInId = Sentry.captureCheckIn({
 });
 
 // Only process this many messages before exiting.
-const maxMessages = 1000;
+const maxMessages = process.env.EMAIL_BREACH_ALERT_MAX_MESSAGES;
```
pdehaan, line 45: IIUC this is now a string and not a number, but not sure if that matters (maybe `subClient.pull()` converts it behind the scenes anyways). 🤷

rhelmer: Hm yeah good point, I'll can explicitly convert it.

PR #3447, opened 2023-09-22
MNTOR-2022/add content type header for cirrus

rhelmer:
# References: 
Jira: MNTOR-2022

# Description

The Cirrus container requires `Content-Type: application/json` (or it returns HTTP 422), and additionally `context` may not be empty.

# How to test

I ran the Cirrus container in https://github.com/mozilla/experimenter/ locally, but I think we can just deploy to stage and test it there at this point.

### DIFF
--- a/src/app/functions/server/getExperiments.ts
+++ b/src/app/functions/server/getExperiments.ts
@@ -26,7 +26,10 @@ export async function getExperiments(
         "Content-Type": "application/json",
       },
       method: "POST",
-      body: JSON.stringify({ client_id: userId, context: {} }),
+      body: JSON.stringify({
+        client_id: userId,
+        context: { key: "example-key" },
+      }),
     });
   } catch (ex) {
     console.error(`Could not connect to Cirrus on ${serverUrl}`, ex);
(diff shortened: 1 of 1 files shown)


yashikakhurana (approved):
🚢  :shipit:  thanks @rhelmer 👏

src/app/functions/server/getExperiments.ts
```diff
@@ -22,8 +22,14 @@ export async function getExperiments(
...
   try {
     features = await fetch(`${serverUrl}/v1/features/`, {
+      headers: {
+        "Content-Type": "application/json",
+      },
       method: "POST",
-      body: JSON.stringify({ client_id: userId, context: {} }),
+      body: JSON.stringify({
+        client_id: userId,
+        context: { key: "example-key" },
```
pdehaan, line 31: Q: should this be an env var? Or is this not really _"used"_ (as long as it is non-empty)?

mansaj (approved):
Looks like it's just to test a basic post req and see if we can get a 200 back

src/app/functions/server/getExperiments.ts
rhelmer: Right it's just an example, but it will be populated by session-specific info that Nimbus can use for targeting (language, geo-location, etc etc)

PR #3468, opened 2023-09-26
Fix date input placeholder for iOS

flozia:
# References: 
Jira: [MNTOR-1962](https://mozilla-hub.atlassian.net/browse/MNTOR-1962)

# Description

Fixes the visual appearance on iOS devices.

# Screenshot

**Before**
| Empty | Filled |
| --- | --- |
| dob-before-placeholder | dob-before-filled |

**After**
| Empty | Filled |
| --- | --- |
| dob-empty-after | dob-filled |

# How to test

- Visit https://deploy-preview-3468--fx-monitor-storybook.netlify.app/?path=/story/pages-onboarding--onboarding on an iOS device
- Make sure the “date of birth” input looks like on the screenshots above

[MNTOR-1962]: https://mozilla-hub.atlassian.net/browse/MNTOR-1962?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


locales-pending/onboarding.ftl
```diff
@@ -47,6 +47,7 @@ onboarding-enter-details-placeholder-last-name = Enter last name
 onboarding-enter-details-label-location = City and state
 onboarding-enter-details-placeholder-location = Enter city and state
 onboarding-enter-details-label-date-of-birth = Date of birth
+onboarding-enter-details-placeholder-date-of-birth = Enter date of birth
```
Vinnl, line 50: Hmm, so I'm not sure if there's a better option in this particular case, since we don't know what the exact format of the displayed date is going to be, but at least for educational value: the `placeholder` attribute [is supposed to](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input#placeholder) hint to what type of data should go in there, rather than explaining what the field is.

>  So for example if a field is expected to capture a user's first name, and its label is "First Name", a suitable placeholder might be "e.g. Mustafa".

It also doesn't mention `date` as one of the valid elements it can go on though, so that's an extra argument for why it doesn't matter much in this specific case :)

src/app/(proper_react)/redesign/(authenticated)/user/welcome/EnterInfo.tsx
```diff
@@ -126,7 +126,9 @@ export const EnterInfo = ({ onScanStarted, onGoBack }: Props) => {
       label: l10n.getString("onboarding-enter-details-label-date-of-birth"),
       key: "dateOfBirth",
       type: "date",
-      placeholder: "",
+      placeholder: l10n.getString(
+        "onboarding-enter-details-placeholder-date-of-birth"
+      ),
```
Vinnl, line 131: Oh, actually, setting a placeholder explicitly removes the placeholder that Firefox adds by default for date elements:

image

So I think maybe we should just do without the placeholder, or alternatively see if we can do feature detection?

src/app/(proper_react)/redesign/(authenticated)/user/welcome/EnterInfo.module.scss
```diff
@@ -85,6 +85,47 @@
...
+  input[type="date"] {
+    background: transparent;
+
+    &::-webkit-date-and-time-value {
+      text-align: left;
+    }
+
+    &::before {
+      position: absolute;
+    }
```
Vinnl, line 102: Could you add some comments explaining what these are for? I have no idea right now :sweat:

Vinnl (commented):
It currently breaks on at least Firefox, which we probably don't want :)

locales-pending/onboarding.ftl
flozia: That makes sense. Yeah, showing the exact format as a placeholder gets tricky. I guess if we are OK with not showing a placeholder that should be the easiest solution for now.

src/app/(proper_react)/redesign/(authenticated)/user/welcome/EnterInfo.module.scss
flozia: Added comments in [e691fcc](https://github.com/mozilla/blurts-server/pull/3468/commits/e691fccfd105002a6ef7f1d8ba9c57ac12e12298).

src/app/(proper_react)/redesign/(authenticated)/user/welcome/EnterInfo.tsx
flozia: Oh. So I guess as already mentioned not showing the placeholder on iOS (which it already decides not to show) would be the easiest and just fix the visual representation of that input.

Vinnl (approved):
Well, that's a much easier review with so little code left :P

src/app/(proper_react)/redesign/(authenticated)/user/welcome/EnterInfo.module.scss
```diff
@@ -85,6 +85,23 @@
   @media screen and (max-width: $screen-lg) {
     grid-template-columns: 1fr;
   }
+
+  input {
+    min-height: 3rem;
+
+    &::placeholder {
+      // overwrite default opacity for placeholders in Firefox
+      opacity: 1;
```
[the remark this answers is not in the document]
flozia: I don’t have strong feelings about this either. From a design and implementation perspective, I find it confusing that opacity is applied by default. I’d like to overwrite it here because otherwise in Firefox this is leading to the text inputs not having the same text color as for the date input.

image

src/app/(proper_react)/redesign/(authenticated)/user/welcome/EnterInfo.module.scss
```diff
@@ -85,6 +85,23 @@
...
+    }
+  }
+
+  input[type="date"] {
+    // remove grey background and align date input value on iOS
+    background: transparent;
+    &::-webkit-date-and-time-value {
+      text-align: left;
```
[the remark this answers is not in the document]
flozia: Change in [000fde9](https://github.com/mozilla/blurts-server/pull/3468/commits/000fde973fabdd2ada29865a3c69943906411622).

src/app/(proper_react)/redesign/(authenticated)/user/welcome/EnterInfo.module.scss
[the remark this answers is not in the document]
Vinnl: Ha OK, yeah that sounds like a Firefox issue then. I guess maybe the reasoning is that in the DoB field it's not a placeholder, but sub-inputs that you can click? Either way fine to work around it here I suppose.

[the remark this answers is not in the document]
flozia: But yeah, good point that it is probably there for a reason. I find it just really confusing that setting the placeholder explicitly to a color will not be shown as intended. I went ahead and merged the main issue can be fixed for now and hope it’s not too bad of a praxis.

PR #3469, opened 2023-09-26
Add story for the manual broker resolution view

Vinnl:
This moves all the async data fetching from `ManualRemoveView.tsx` into the `page.tsx`, making the former just render data that it gets passed as props. That can then also be used as a story.

Note that this builds on https://github.com/mozilla/blurts-server/pull/3464.

### DIFF
--- a/src/app/(proper_react)/redesign/(authenticated)/user/dashboard/fix/data-broker-profiles/manual-remove/01c.ManualRemove.stories.tsx
+++ b/src/app/(proper_react)/redesign/(authenticated)/user/dashboard/fix/data-broker-profiles/manual-remove/01c.ManualRemove.stories.tsx
@@ -1,0 +1,87 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import type { Meta, StoryObj } from "@storybook/react";
+import { OnerepScanRow } from "knex/types/tables";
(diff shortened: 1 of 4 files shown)


src/app/(proper_react)/redesign/(authenticated)/user/dashboard/fix/data-broker-profiles/manual-remove/01c.ManualRemove.stories.tsx
flozia, line 1: Nit: In Storybook the order of stories is determined by the order they are being imported. When we are starting to introduce a naming convention for files that implies an order it would be cool if we could sort the stories. That’s possible with the [storySort](https://storybook.js.org/docs/react/writing-stories/naming-components-and-hierarchy#sorting-stories) option which could be added to `preview.tsx`. From the examples the following would sort alphabetically:

```
options: {
    storySort: (a, b) => a.id === b.id ? 0 : a.id.localeCompare(b.id, undefined, { numeric: true }),
}
```

| Before  | After |
| --- | --- |
| image | image |

flozia (approved):
Looks good!

src/app/(proper_react)/redesign/(authenticated)/user/dashboard/fix/data-broker-profiles/manual-remove/01c.ManualRemove.stories.tsx
Vinnl: https://github.com/mozilla/blurts-server/pull/3464#discussion_r1339914945

PR #3470, opened 2023-09-26
Update scan status

flozia:
When a scan is finished we’ll need to update the status in our database.

### DIFF
--- a/src/app/(proper_react)/redesign/(authenticated)/user/dashboard/subscribed/page.tsx
+++ b/src/app/(proper_react)/redesign/(authenticated)/user/dashboard/subscribed/page.tsx
@@ -54,7 +54,7 @@ export default async function Subscribed() {
       profileId,
       latestScan.scan.onerep_scan_id,
       scans,
-      "manual",
+      "initial",
       latestScan.scan.onerep_scan_status
     );
   }
--- a/src/app/(proper_react)/redesign/(authenticated)/user/dashboard/subscribed/page.tsx
+++ b/src/app/(proper_react)/redesign/(authenticated)/user/dashboard/subscribed/page.tsx
@@ -54,7 +54,7 @@ export default async function Subscribed() {
       profileId,
       latestScan.scan.onerep_scan_id,
       scans,
-      "initial",
+      "manual",
       latestScan.scan.onerep_scan_status
     );
   }
--- a/src/db/tables/onerep_scans.ts
+++ b/src/db/tables/onerep_scans.ts
@@ -62,18 +62,6 @@ async function setOnerepManualScan(
   });
 }
 
-async function updateOnerepScanStatus(
-  onerepScanId: number,
-  onerepScanStatus: Scan["status"]
-) {
-  await knex("onerep_scans").where("onerep_scan_id", onerepScanId).update({
-    onerep_scan_status: onerepScanStatus,
-    // @ts-ignore knex.fn.now() results in it being set to a date,
-    // even if it's not typed as a JS date object:
-    updated_at: knex.fn.now(),
-  });
-}
-
 async function addOnerepScanResults(
   onerepProfileId: number,
   onerepScanId: number,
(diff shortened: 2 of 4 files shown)


src/app/(proper_react)/redesign/(authenticated)/user/dashboard/subscribed/page.tsx
```diff
@@ -54,7 +54,7 @@ export default async function Subscribed() {
       profileId,
       latestScan.scan.onerep_scan_id,
       scans,
-      "initial",
+      "manual",
```
Vinnl, line 57: As I understand it (after having asked Rob about this a couple of days ago), `initial` is the status for the first scan that happens automatically after a profile has been activated - and once it's active, it'll periodically rescan with a `monitoring` status. `manual`, on the other hand, is for a one-time scan when the profile is not activated - i.e. for non-Premium customers.

I'm not sure if this code applies to a Premium or non-Premium customer. The error message referring to a manual scan above, together with the profile activation, is confusing.

src/db/tables/onerep_scans.ts
```diff
@@ -62,6 +62,18 @@ async function setOnerepManualScan(
...
+async function updateOnerepScanStatus(
+  onerepScanId: number,
+  onerepScanStatus: Scan["status"]
+) {
+  await knex("onerep_scans").where("onerep_scan_id", onerepScanId).update({
+    onerep_scan_status: onerepScanStatus,
+    // @ts-ignore knex.fn.now() results in it being set to a date,
+    // even if it's not typed as a JS date object:
+    updated_at: knex.fn.now(),
+  });
```
Vinnl, line 74: I wonder if it might make more sense to instead incorporate this into `addOnerepScanResults`? Since we basically only call that when we get an updated scan with a potentially updated status?

Vinnl (commented):
Thanks for looking into this. Let me know what you think of my suggestion - if you see a reason not to, happy to re-review.

src/app/(proper_react)/redesign/(authenticated)/user/dashboard/subscribed/page.tsx
flozia: The issue here is that we can’t run an `initial` scan over and over again, because we decided to update the scans. One question I’d have for @rhelmer if we actually still need to run this everytime we visit `/subscribed` on dev.

src/db/tables/onerep_scans.ts
flozia: I was wondering about that as well but was hesitant to change too much in that logic at first. I agree that this might be the more intuitive place to update the scan status. I’ve added [cb8a4aa](https://github.com/mozilla/blurts-server/pull/3470/commits/cb8a4aa442946be398e3fa85a5498024e13627f1) for either inserting or updating the scan.

src/app/(proper_react)/redesign/(authenticated)/user/dashboard/subscribed/page.tsx
flozia: Ah, I got it now with the different scan types. I’ve changed how the scans are being updated — see: https://github.com/mozilla/blurts-server/pull/3470#discussion_r1338447632.

src/db/tables/onerep_scans.ts
```diff
@@ -76,16 +76,27 @@ async function addOnerepScanResults(
...
+        })
+        .onConflict("onerep_scan_id")
+        .ignore()
+        .where("onerep_scan_id", onerepScanId)
+        .update({
+          onerep_scan_status: onerepScanStatus,
+          // @ts-ignore knex.fn.now() results in it being set to a date,
+          // even if it's not typed as a JS date object:
+          updated_at: knex.fn.now(),
+        });
```
Vinnl, line 99: Woah, does this work? Does this result in a single query? I'd have expected something [like `.onConflict().merge()`](https://knexjs.org/guide/query-builder.html#onconflict):

```suggestion
        .merge(["updated_at", "onerep_scan_status"]);
```

Vinnl (approved):
Thanks! I noted inline how I expected the query to look, but if this works too then I'm good with it!

src/db/tables/onerep_scans.ts
flozia: Had to update the columns explicitly with the new values — otherwise the linter complained about non-matching types: https://github.com/mozilla/blurts-server/pull/3470/commits/6c597731e2daa5d914bfcfbbf0ad38659a251a75.

Vinnl: That works for me!

mansaj (approved):
Lgtm. Thanks

rhelmer (approved):

PR #3474, opened 2023-09-27
MNTOR-2224 - allow updating FxA profile info from front-end without sign-out

rhelmer:
# References: 
Jira: MNTOR-2224

# Description

This causes `update()` from `useSession` to trigger a profile refresh, to update the list of subscriptions (and any other info that has changed). It should not require the user to re-authenticate, as it uses the stored access token.

As noted in a comment in the code, this makes use of a client-side function as there doesn't seem to be any supported way to do this server-side in Next-Auth. I experimented a bit with this and I think we could do it with some effort, but this i a far smaller change (at the cost of an extra redirect to the user).

# How to test

Positive test:
1. create a new account
2. go through subscription flow
3. observe redirect from welcome-to-premium -> subscription -> back

Negative test:
1. try to go directly to welcome-to-premium page without having subscribed
2. observe redirect to / (landing page)

- [x] Remove demo button


src/app/api/utils/auth.ts
```diff
@@ -92,8 +92,26 @@ export const authOptions: AuthOptions = {
...
+        const access_token = token.subscriber?.fxa_access_token;
+        if (!access_token) {
+          throw new Error("no access token");
+        }
+
+        const result = await fetch(process.env.OAUTH_PROFILE_URI, {
+          headers: {
+            Authorization: `Bearer ${access_token}`,
+          },
+        });
```
rhelmer, line 111: @Vinnl  I noticed there's a `userInfo` property on our provider that does this same fetch basically, I can't find much info about what it's for in the docs though.

rhelmer: Is there a way to call that, or does it serve a totally different purpose? Maybe both should call a utility function? not sure.

src/app/(proper_react)/redesign/(authenticated)/user/dashboard/subscribed/page.tsx
```diff
@@ -2,76 +2,45 @@
...
+      } catch (ex) {
+        console.error(ex);
+        captureException(ex);
+      }
+      window.location.href = "/";
+    }
+    void updateSession();
+    // We only want to run this a single time - the page should not be re-rendered.
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, []);
```
Vinnl, line 43: I think you should be able to have the recommended dependencies here - [the docs do so too](https://next-auth.js.org/getting-started/client#refetching-the-session), so I think `update` only changes when the user e.g. logs out.

```suggestion
  }, [update]);
```

src/app/(proper_react)/redesign/(authenticated)/user/dashboard/fix/data-broker-profiles/welcome-to-premium/page.tsx
```diff
@@ -25,6 +26,11 @@ export default async function WelcomeToPremium() {
     redirect("/redesign/user/dashboard/");
   }
 
+  // The user may have subscribed and just need their session updated - they will be redirected back to try again if it looks valid.
+  if (!hasPremium(session.user)) {
+    redirect(`${process.env.NEXTAUTH_URL}/redesign/user/dashboard/subscribed`);
+  }
```
Vinnl, line 32: Wouldn't we be passing this URL as a `redirect_uri` to subplat or FxA, instead of redirecting the user there ourselves?

Vinnl (commented):
Would be great if this works! Added a couple of questions/suggestions inline, just to make sure I understand everything that's going on.

src/app/(proper_react)/redesign/(authenticated)/user/dashboard/subscribed/page.tsx
rhelmer: Hm I removed this because IIRC it was not working correctly, let me test and report on what happened. I only want this effect to run a single time and not change when properties change in this case FWIW - either we update and the user has a subscription and we redirect them to the welcome page, or something in that chain fails or is false and we redirect them to /.

src/app/(proper_react)/redesign/(authenticated)/user/dashboard/fix/data-broker-profiles/welcome-to-premium/page.tsx
rhelmer: I think this would make more sense, I already provided them the above URL before I realized we couldn't easily update from the server side, but we can have it changed.

src/app/(proper_react)/redesign/(authenticated)/user/dashboard/subscribed/page.tsx
```diff
@@ -2,76 +2,45 @@
...
+export default function Subscribed() {
+  const { update } = useSession();
+
+  useEffect(() => {
+    async function updateSession() {
+      // You can await here
+      try {
+        const result = await update();
+        if (hasPremium(result?.user)) {
+          window.location.href =
```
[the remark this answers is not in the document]
rhelmer: Yes `update` does refresh the cookie - this is the complication of doing the update purely on the server-side, it's necessary to re-send headers etc (as it is an http-only cookie). The client component is `POST`ing to the `/api/auth/session` endpoint, and including cookies in that request. I think it's likely doing to do more of this server-side but I see why it's complicated.

Anyway yes I will useRouter and thanks for pointing out the localhost hardcode, not intentional :)

src/app/(proper_react)/redesign/(authenticated)/user/dashboard/fix/data-broker-profiles/welcome-to-premium/page.tsx
rhelmer: OK one concern I had was if users subscribed first, but I don't think that actually matters, we can just have this "subscribed" endpoint be the place to start for subscription users.

I'm open to changing the URL FWIW, I only chose this one for convenience (it inherits the dashboard styles etc)

src/app/(proper_react)/redesign/(authenticated)/user/dashboard/subscribed/page.tsx
rhelmer: OK it works fine, it does run twice though... I don't understand `useEffect` enough to know why :(

src/app/(proper_react)/redesign/(authenticated)/user/dashboard/subscribed/page.tsx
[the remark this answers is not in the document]
rhelmer: OK thanks this works great, and it replaces history (which is good in this case), thanks!

src/app/(proper_react)/redesign/(authenticated)/user/dashboard/subscribed/page.tsx
```diff
@@ -2,76 +2,49 @@
...
+      } catch (ex) {
+        console.error(ex);
+        captureException(ex);
+        router.replace(`/`);
+      }
+    }
+    void updateSession();
+  }, [update, router]);
+
+  return <div>Please wait...</div>;
```
pdehaan, line 49: nit: `"Please wait..."` should be localized?
And possibly use `…` (`&hellip;`) instead of `...`?

src/app/(proper_react)/redesign/(authenticated)/user/dashboard/subscribed/page.tsx
```diff
@@ -2,76 +2,49 @@
...
+ */
+export default function Subscribed() {
+  const { update } = useSession();
+  const router = useRouter();
+
+  useEffect(() => {
+    async function updateSession() {
+      try {
+        const result = await update();
+        console.debug(result, hasPremium(result?.user));
```
pdehaan, line 32: ✂️

src/app/(proper_react)/redesign/(authenticated)/user/dashboard/subscribed/page.tsx
```diff
@@ -2,76 +2,49 @@
...
+        if (hasPremium(result?.user)) {
+          router.replace(
+            `/redesign/user/dashboard/fix/data-broker-profiles/welcome-to-premium`
+          );
+        } else {
+          router.replace(`/`);
+        }
+      } catch (ex) {
+        console.error(ex);
```
pdehaan, line 41: ✂️ 
Not sure on our policy of `console` logging and if these are being captured somewhere in STDOUT or STDERR or were just temporary testing artifacts.

src/app/api/utils/auth.ts
rhelmer: OK I ended up refactoring this into a common function, it was less work than trying to build something compatible with context (all it really needs is an access token)

src/app/(proper_react)/redesign/(authenticated)/user/dashboard/subscribed/page.tsx
rhelmer: `console.error` is indeed captured by logging

src/app/(proper_react)/redesign/(authenticated)/user/dashboard/subscribed/page.tsx
rhelmer: Yes I'll put it in pending locales at the very least. This wasn't really part of the design.

src/app/(proper_react)/redesign/(authenticated)/user/dashboard/subscribed/page.tsx
```diff
@@ -2,76 +2,49 @@
...
+        captureException(ex);
+        router.replace(`/`);
+      }
+    }
+    void updateSession();
+  }, [update, router]);
```
rhelmer, line 47: @Vinnl OK so the problem is that this `useEffect` runs several times if I have any dependencies here... ideally I'd only have it run a single time, I don't really care if the dependencies change. Wdyt?

src/app/(proper_react)/redesign/(authenticated)/user/dashboard/subscribed/page.tsx
rhelmer: We're working on moving to structured logging, but for now this is actually used. The way Sentry handles and presents it is more useable though.

src/app/(proper_react)/redesign/(authenticated)/user/dashboard/subscribed/page.tsx
Vinnl: Oh yeah that's fine, that's a somewhat contentious React feature if you enable strict mode (which we do, at the moment): [it will render everything twice](https://react.dev/reference/react/StrictMode#fixing-bugs-found-by-double-rendering-in-development) to surface errors resulting from components being impure functions. Strict mode only affects dev mode.

src/app/(proper_react)/redesign/(authenticated)/user/dashboard/subscribed/page.tsx
```diff
@@ -2,76 +2,50 @@
...
+import { useRouter } from "next/navigation";
+
+import { hasPremium } from "../../../../../../functions/universal/user";
+import { captureException } from "@sentry/browser";
+import { useL10n } from "../../../../../../hooks/l10n";
+
+/**
+ * Client-side page to update session info.
+ *
+ * Next-Auth does not have a simple way to do this purely from the client-side, so we
```
Vinnl, line 18: I think you meant this, no?

```suggestion
 * Next-Auth does not have a simple way to do this purely from the server-side, so we
```

src/app/(proper_react)/redesign/(authenticated)/user/dashboard/subscribed/page.tsx
Vinnl: Twice, or more? Twice should be fine, in the sense that that's only in dev mode: https://github.com/mozilla/blurts-server/pull/3474#discussion_r1341193204

What goes wrong if it runs multiple times?

src/app/api/utils/auth.ts
```diff
@@ -92,8 +85,10 @@ export const authOptions: AuthOptions = {
   ],
   callbacks: {
     // Unused arguments also listed to show what's available:
-    // eslint-disable-next-line @typescript-eslint/no-unused-vars
     async jwt({ token, account, profile, trigger }) {
+      if (trigger === "update") {
+        profile = await fetchUserInfo(token.subscriber?.fxa_access_token ?? "");
```
Vinnl, line 90: I think you should also pass this through the `authOptions.providers[0].profile` callback to ensure that the format is consistent. (Possibly also extracting that out into a separate function.)

src/app/(proper_react)/redesign/(authenticated)/user/dashboard/fix/data-broker-profiles/welcome-to-premium/page.tsx
Vinnl, line 1: Question: is my assumption correct that the user always ends up at this route after subscribing to Premium? Or are we supposed to have code like this on every page that might reflect the user's Premium status? (I'm hoping the former, otherwise I foresee a couple potential problems :) )

Vinnl (approved):
One blocking change (using the `profile` callback), but should be pretty easy to resolve, so approving beforehand with that caveat.

src/app/(proper_react)/redesign/(authenticated)/user/dashboard/subscribed/page.tsx
rhelmer: Oh that makes sense, thanks! It shouldn't be harmful to run multiple times, just wanted to make sure we understood. Thanks

src/app/(proper_react)/redesign/(authenticated)/user/dashboard/fix/data-broker-profiles/welcome-to-premium/page.tsx
rhelmer: The former, this is the page that subplat returns users to currently.

Vinnl: Perfect!

src/app/api/utils/auth.ts
Vinnl: Hmm that's really weird, I was pretty sure (see also the comment about the `profile` function) that it parsed the return value. The primary difference I think is that we use the `uid` from the FxA profile as the `id` used by `Next-Auth`. If the object returned by `/userinfo/` does not include a `.uid` property, then you're probably right and we can merge this.

rhelmer: Oh OK yes stepping through the output in the debugger you're right... `/userinfo/` does return an `FxaProfile`. OK I see your point then, I'm not sure why that passed typechecking - I suspect it's just because the `jwt` callback doesn't make use of any of these extra/conflicting fields like `uid` vs. `id` an `amrValues` - it just wants to create a token like:

```
        token.fxa = {
          locale: profile.locale,
          twoFactorAuthentication: profile.twoFactorAuthentication,
          metricsEnabled: profile.metricsEnabled,
          avatar: profile.avatar,
          avatarDefault: profile.avatarDefault,
          subscriptions: profile.subscriptions,
        };
```

I'll go ahead and do what you originally suggested.

rhelmer: That is, we should have a common way to take an `FxaProfile` and convert that to a standard `Profile`.

PR #3477, opened 2023-09-27
Set fallback id for GA

flozia:
Add fallback value for the public env var `NEXT_PUBLIC_GA4_MEASUREMENT_ID`.


rhelmer (approved):
Shouldn't be needed in theory, since it's set as env var on stage/prod, but not showing up there! It is on dev. Anyway thanks, let's go with this for now.

PR #3478, opened 2023-09-28
MNTOR-42 - do not show upsells for premium users

rhelmer:
# References: 
Jira: MNTOR-42

# Description

Premium users should not see upsells. There are quite a few states so it's hard to exhaustively test - in any case we need to use the universal `hasPremium` function to check`session.user` (obtained with `useSession` for client components, and `getServerSession` for server compontents).

# Screenshot (if applicable)

Not applicable.

# How to test

 waived the privilege.

### DIFF
--- a/src/app/(proper_react)/redesign/(authenticated)/user/dashboard/DashboardTopBanner.tsx
+++ b/src/app/(proper_react)/redesign/(authenticated)/user/dashboard/DashboardTopBanner.tsx
@@ -15,7 +15,6 @@ import {
   getRelevantGuidedSteps,
 } from "../../../../../functions/server/getRelevantGuidedSteps";
 import { hasPremium } from "../../../../../functions/universal/user";
-import { useSession } from "next-auth/react";
 
 export type BannerContent =
   | "ScanInProgressContent"
@@ -79,7 +78,6 @@ const Content = (props: {
   onShowFixed: () => void;
 }) => {
   const l10n = useL10n();
-  const { data: session } = useSession();
 
   if (props.isOnFixedTab) {
     return (
@@ -232,8 +230,7 @@ const Content = (props: {
     //       brokers. Should we be keeping track of the user "acknowledging"
     //       their scan results or something?
 
-    console.debug(session?.user);
-    if (!hasPremium(session?.user)) {
+    if (hasPremium(props.stepDeterminationData.user)) {
       return (
         <div className={styles.explainerContent}>
           <h3>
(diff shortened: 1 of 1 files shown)


src/app/(proper_react)/redesign/(authenticated)/user/dashboard/DashboardTopBanner.tsx
```diff
@@ -229,27 +231,32 @@ const Content = (props: {
...
-              data_broker_sites_total_num: dataBrokerCount,
-            }
-          )}
-        </p>
-        <div className={styles.cta}>
-          <Button href={relevantGuidedStep.href} small variant="primary">
-            {l10n.getString("dashboard-top-banner-no-exposures-found-cta")}
-          </Button>
+
+    console.debug(session?.user);
```
mansaj, line 235: leave this in?

mansaj:
This is triggering a bunch of unit test failures..

rhelmer:
> This is triggering a bunch of unit test failures..

Thanks, I'm doing something very simple so it's likely that I just need to change my approach (or make the unit tests aware of subscription status, which would be a fine outcome too).

rhelmer:
> > This is triggering a bunch of unit test failures..
> 
> Thanks, I'm doing something very simple so it's likely that I just need to change my approach (or make the unit tests aware of subscription status, which would be a fine outcome too).

Oh I see, they are all of the form:

```
    [next-auth]: `useSession` must be wrapped in a <SessionProvider />
```

src/app/(proper_react)/redesign/(authenticated)/user/dashboard/DashboardTopBanner.tsx
```diff
@@ -229,27 +229,31 @@ const Content = (props: {
     //       are directed to their data breaches if they have no unresolved data
     //       brokers. Should we be keeping track of the user "acknowledging"
     //       their scan results or something?
```
Vinnl, line 231: Hmm, as described in the TODO on line 228 (that I can't attach a comment to...), I don't think this change actually does anything? Is there a Storybook story I can look at that does something different than it does on `main`? Would it be possible to add a unit test that would fail on `main` but succeeds with this change?

rhelmer: I'll see if I can get it to show in Storybook... I noticed that too but I see it if I don't have any breach nor broker results in my account.

rhelmer:
@flozia is already working on an exhaustive list of such cases, all we should need to do is check use the univeral `hasPremium` function in most cases I think. Feel free to close this PR if it's not useful.

Vinnl:
OK yeah, I think a comprehensive systematic and testable approach is probably safest, so let's await that.

PR #3479, opened 2023-09-28
MNTOR-2193 - use API to check OneRep status as needed, do not depend on webhook for UI

rhelmer:
# References: 
Jira: MNTOR-2193

# Description

- remove "development" gates around current polling
- compare current remote scan results against the DB on every dashboard load, and store any newer results
  - but fall back to database if there are any errors

# How to test

Using the site locally should now be equivalent on stage/prod - we will still process webhooks when they come in, but always check the remote status for UI purposes (and fall back to database if necessary).


src/app/(proper_react)/redesign/(authenticated)/user/dashboard/page.tsx
```diff
@@ -33,7 +40,45 @@ export default async function DashboardPage() {
...
+
+  // Attempt to fetch the current scan results from the provider.
+  // If anything is newer, add it to the database.
+  // Continue if there are any errors.
+  try {
+    const currentScan = await getAllScanResults(profileId);
+    const newScanResults: ScanResult[] = [];
+    currentScan.forEach((current) => {
+      latestScan.results.forEach((latest) => {
+        console.debug(new Date(current.updated_at), latest.updated_at);
```
Vinnl, line 54: Forgot to remove? (Or alternatively, to add more detail to.)

```suggestion
```

src/app/(proper_react)/redesign/(authenticated)/user/dashboard/page.tsx
```diff
@@ -33,7 +40,45 @@ export default async function DashboardPage() {
...
   const latestScan = await getLatestOnerepScanResults(profileId);
+
+  // Attempt to fetch the current scan results from the provider.
+  // If anything is newer, add it to the database.
+  // Continue if there are any errors.
+  try {
+    const currentScan = await getAllScanResults(profileId);
+    const newScanResults: ScanResult[] = [];
+    currentScan.forEach((current) => {
+      latestScan.results.forEach((latest) => {
```
Vinnl, line 53: This might be easier to follow using something like `const newScanResults = currentScan.map(({ results }) => results.filter(/* ... */)).flat()`?

(But that might also just be personal preference.)

src/app/(proper_react)/redesign/(authenticated)/user/dashboard/page.tsx
```diff
@@ -33,7 +40,45 @@ export default async function DashboardPage() {
...
+          new Date(current.updated_at) > latest.updated_at
+        ) {
+          newScanResults.push(current);
+        }
+      });
+    });
+
+    if (newScanResults.length > 0) {
+      const scanId = latestScan?.scan?.id;
+      if (!scanId && typeof scanId === "number") {
```
Vinnl, line 67: I might be misunderstanding this, but doesn't this only trigger if `scanId === 0`? I'd expect:

```suggestion
      if (typeof scanId !== "number") {
```

src/app/(proper_react)/redesign/(authenticated)/user/dashboard/page.tsx
```diff
@@ -33,7 +40,45 @@ export default async function DashboardPage() {
     return redirect("/redesign/user/welcome/");
   }
 
+  // This contains the latest scan results in our database.
   const latestScan = await getLatestOnerepScanResults(profileId);
+
+  // Attempt to fetch the current scan results from the provider.
```
Vinnl, line 46: Might it make sense to extract this into a `refreshStoredScanResults` function? Since I assume the same code needs to run in response to webhook calls? (Or does the Webhook call already include the latest data?)

Vinnl (approved):

src/app/(proper_react)/redesign/(authenticated)/user/dashboard/page.tsx
rhelmer: Yeah forgot to remove, thanks!

src/app/(proper_react)/redesign/(authenticated)/user/dashboard/page.tsx
rhelmer: Hm yeah I don't use the various array functions enough... I wonder if https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/flatMap could simplify it further? Will see what I can come up with.

src/app/(proper_react)/redesign/(authenticated)/user/dashboard/page.tsx
rhelmer: The webhook payload only includes the scan ID that needs update, sadly :/ I'll see if there's any opportunity for more code sharing though: https://github.com/mozilla/blurts-server/blob/main/src/app/api/v1/onerep-events/route.ts

rhelmer: OK I extracted this to a server function and out of the dashboard component for readability - I think it could be used from the onerep-events webhook potentially but there are some differences:

1. the webhook includes the scan ID as well as the profile ID, and we don't try to filter duplicates
2. `refreshStoredScanResults` as-is just refreshes everything for a give profile ID, filtering duplicates

We could have one function that could do both, but they don't have that much in common except for a few function calls to functions in the `onerep.ts` module.

src/app/(proper_react)/redesign/(authenticated)/user/dashboard/page.tsx
rhelmer: OK I can't figure out a way that seems more readable :/ to be clear, what this needs to do is:

1. fetch the latest results from remote and the db and add any new scan IDs
2. compare each element in the array, if any IDs match then only update if `updated_at` is newer

Vinnl: Woah! I did not know that `.flatMap` was actually already standardised and widely supported! I'd have written a more complete diff, but it's merged already, so :shrug:

PR #3480, opened 2023-09-28
Avoid error about undefined `navigator`

Vinnl:
When server-rendering, `navigator` is not defined - but by comparing that variable directly with `undefined`, we're still referring to it and thus still triggering an error message. Passing it into `typeof` avoids that.


flozia (approved):
Thanks!

PR #3483, opened 2023-09-28
MNTOR-1945 - Data broker logos logic

codemist:
# References: 
Jira: MNTOR-1945

# Description
Adds the logic for dynamically importing data broker logos. Sampled a couple for now, tracking this list: https://docs.google.com/spreadsheets/d/10oLIpgfCA1PKLy1wCLaPzo4f_-Zj6f6TCYV5AUytv5k/edit#gid=0

# Screenshot (if applicable)

image

### DIFF
--- a/src/client/css/partials/breaches.css
+++ b/src/client/css/partials/breaches.css
@@ -221,6 +221,7 @@
 .breach-row .breach-company .breach-logo,
 .breach-row .breach-company img {
   height: 1.5rem;
+  width: 1.5rem;
   display: inline-block;
   vertical-align: bottom;
   margin-right: var(--padding-xs);


src/app/components/client/ExposureCard.tsx
```diff
@@ -44,15 +44,44 @@ export type ExposureCardProps = {
...
-  );
+  const [dataBrokerImage, setDataBrokerImage] = useState<ReactNode>();
+
+  useEffect(() => {
+    const loadDataBrokerImage = async () => {
+      if (isScanResult(exposureData)) {
+        const cleanedName = exposureData.data_broker.split(".")[0];
+        try {
+          const logo = await import(
+            `../client/assets/data-brokers/${cleanedName}.jpg`
```
pdehaan, line 55: Interesting. Is this a temp thing until these move to a/the CDN?

src/app/components/client/ExposureCard.tsx
```diff
@@ -44,15 +44,44 @@ export type ExposureCardProps = {
...
+          setDataBrokerImage(<Image src={logo.default} alt="" />);
+        } catch (error) {
+          // Default to circle logos if logo is not found
+          setDataBrokerImage(<FallbackLogo name={cleanedName} />);
+        }
+      }
+    };
+
+    loadDataBrokerImage().catch((error) => {
+      console.error("Error loading logo:", error);
```
pdehaan, line 66: Q: Where does this `console.error()` go?
Should we log it to Sentry?

src/app/components/client/ExposureCard.tsx
```diff
@@ -151,15 +180,7 @@ const ScanResultCard = (props: ScanResultCardProps) => {
...
-                exposureImg ? (
-                  <Image
-                    className={styles.exposureImage}
-                    alt=""
-                    src={exposureImg}
-                  />
-                ) : (
-                  <FallbackLogo name={scanResult.data_broker} />
-                )
+                exposureImg ?? <FallbackLogo name={scanResult.data_broker} />
```
pdehaan, line 183: Not sure this matters, but I noticed we used a `cleanedName` version above.
```js
const cleanedName = exposureData.data_broker.split(".")[0];
...
setDataBrokerImage(<FallbackLogo name={cleanedName} />);
```

Although maybe `cleanedName` is irrelevant if we're only looking at the first character anyways and this comment is delightfully moot.
```js
props.name.substring(0, 1)
```

src/app/components/client/assets/data-brokers/arivify.jpg
pdehaan, line 1: @maxxcrawford Any guidance here on whether we were better off w/ transparent PNGs for the old, bespoke breach logos?
Not sure of the status of "dark mode" version of Monitor and if it matters, but otherwise the logos all seemed to be 120&times;30px and looked 👌.

src/app/components/client/ExposureCard.tsx
```diff
@@ -45,15 +45,44 @@ export type ExposureCardProps = {
...
-    <ScanResultCard {...props} scanResult={exposureData} />
-  ) : (
-    <SubscriberBreachCard {...props} subscriberBreach={exposureData} />
-  );
+  const [dataBrokerImage, setDataBrokerImage] = useState<ReactNode>();
+
+  useEffect(() => {
+    const loadDataBrokerImage = async () => {
+      if (isScanResult(exposureData)) {
+        const cleanedName = exposureData.data_broker.split(".")[0];
```
Vinnl, line 53: It's probably a good idea to leave the TLD on, both to make it easier to find the website the logo was from, as well as to avoid problems if by any chance two instances share the same name but a different TLD (e.g. `truthfinder.com` vs `truthfinder.net`).

src/app/components/client/ExposureCard.tsx
```diff
@@ -45,15 +45,44 @@ export type ExposureCardProps = {
...
+        try {
+          const logo = await import(
+            `../client/assets/data-brokers/${cleanedName}.jpg`
+          );
+          setDataBrokerImage(<Image src={logo.default} alt="" />);
```
Vinnl, line 58: While I assume this works, React has [built-in affordances](https://react.dev/reference/react/lazy) to support lazily loading components, and using that presumably allows it to optimise the render order and be more robust to future updates.

I believe you would implement that something like this:

```tsx
export const DataBrokerImage = (props: { name: string }) => {
  const DataBrokerLogo = lazy(() => import(`../client/asserts/data-brokers/${name}.jpg`).then(logo => ({
    default: <Image src={logo.default} alt="" />,
  }));

  return (
    <Suspense fallback={<FallbackLogo name={name}/>}>
      <DataBrokerLogo/>
    </Suspense>
  );
};
```

And then you should be able to pass `<DataBrokerImage name={cleanedName} />` into `ScanResultCard`'s `exposureImg`.

Vinnl (commented):
The general approach seems fine, but I'd like to outsource the lazy-loading of the image of React's idiomatic `lazy` API, if possible.

And yeah, I guess it's time to add a ticket to the backlog to discuss with SRE on how to use a CDN?

src/app/components/client/ExposureCard.tsx
```diff
@@ -45,15 +45,44 @@ export type ExposureCardProps = {
...
-  );
+  const [dataBrokerImage, setDataBrokerImage] = useState<ReactNode>();
+
+  useEffect(() => {
+    const loadDataBrokerImage = async () => {
+      if (isScanResult(exposureData)) {
+        const cleanedName = exposureData.data_broker.split(".")[0];
+        try {
+          const logo = await import(
+            `../client/assets/data-brokers/${cleanedName}.jpg`
```
Vinnl, line 56: Oh, and one more thing: wouldn't `.png` be a more logical format for logos? As I understood it, JPEGs are good for photos, whereas PNGs are more optimised for computer graphics?

src/app/components/client/ExposureCard.tsx
rhelmer: It goes to the GCP logs - if it's something we expect might happen occasionally (and there's not much need to investigate and have a stack trace etc) then going through Sentry isn't necessary.

To send to Sentry, you could use `captureException(error)` from Sentry module. Sentry also catches unhandled exceptions, but that means an error page will be displayed to the user, so if we have a fallback in place then that's likely not what we want.

src/app/components/client/ExposureCard.tsx
Vinnl: Yeah that's right, that's why in my example code the function I passed into `lazy` didn't just the `import` of the image, but then also wrapped it in a component and returned _that_ on the `default` property. Maybe it's easier to follow if I split it up:

```ts
async function importImageInComponent(name: string) => {
  const logo = await import(`../client/asserts/data-brokers/${name}.jpg`);
  return {
    default: <Image src={logo.default} alt="" />,
  };
};

const DataBrokerLogo = lazy(() => importImageInComponent(name));
```

And yeah, the `lazy` you're seeing in the DOM is a different one: [this one](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/Img#lazy) rather than [this one](https://react.dev/reference/react/lazy). They do similar things in different situations, hence the similar names.

src/app/components/client/ExposureCard.tsx
```diff
@@ -44,31 +44,51 @@ export type ExposureCardProps = {
...
+    return {
+      default: ImageComponent,
+    };
+  } catch {
+    const FallBackLogo = () => <FallbackLogo name={name} />;
+    return {
+      default: FallBackLogo,
+    };
+  }
+}
```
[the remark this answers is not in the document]
Vinnl: Ooh yeah, would be nice to be able to see in Sentry which logos people are most often not seeing logos for (and e.g. if OneRep has added new data brokers we do not have logos for yet).

Vinnl (approved):
The change looks good. Some unit tests are failing though, because React thinks you might want to test the loading process. Since it's just loading images without user interaction, it's not that interesting to test, so I think a fine solution would be to just extract the `<DataBrokerImage>` into a separate file, and then mock that in a test.

Which would be a neat thing to practice, so you may want to consider doing that yourself :) I can also imagine you not feeling like that after the hassle with lazy loading, though, so if so, I've got a PR for you here: https://github.com/mozilla/blurts-server/pull/3511

PR #3484, opened 2023-09-29
Add docs on Jira iupdate after stage deploy

Vinnl:


docs/release_process.md
```diff
@@ -133,6 +133,12 @@ After you push the tag to GitHub, you should also
 5. Check the pre-release box.
 6. Click "Publish release"
 
+### Update Jira
+
+On [our Jira board](https://mozilla-hub.atlassian.net/jira/software/c/projects/MNTOR/boards/447), take a look at the tickets listed under "Merged tot main". If those were included in the release you just created, drag those tickets to the "Done" column, in the "Merged" section (sections will appear as you start dragging). This will notify QA that they can verify the behaviour on stage.
```
rhelmer, line 138: If you don't like dragging or are on a small screen etc. you can also just change the status using the pulldown menu. But up to you how you want to document, thanks for writing it up!

rhelmer (approved):

docs/release_process.md
Vinnl: Haha yes, the thing with that is that the statuses in the drop-down don't match with the columns on the board (and the names don't reflect what actually happened), so I wanted to avoid introducing more confusion. Also, I believe Betty is looking into that issue, so then the docs might get outdated soon too.

PR #3485, opened 2023-09-29
Make scan resolution skip link work

Vinnl:
This also involved merging the scan steps into a single one, since we don't want the "Skip" button on the scan step to skip to the scan results step.

I also fixed some numbers that were hardcoded to 0, and the link to the credit card breach resolution.

# References: 
Jira: MNTOR-2116
Figma: N/A

# Description

This makes the "Skip for now" link on the manual broker resolution page work.

# How to test

Visit http://localhost:6060/redesign/user/dashboard/fix/data-broker-profiles/manual-remove, then click "Skip for now".

### DIFF
--- a/src/app/(proper_react)/redesign/(authenticated)/user/dashboard/fix/FixView.tsx
+++ b/src/app/(proper_react)/redesign/(authenticated)/user/dashboard/fix/FixView.tsx
@@ -20,6 +20,7 @@ import stepSecurityRecommendationsIcon from "./images/step-counter-security-reco
 import { usePathname } from "next/navigation";
 import { GuidedExperienceBreaches } from "../../../../../../functions/server/getUserBreaches";
 import { useL10n } from "../../../../../../hooks/l10n";
+import { stepLinks } from "../../../../../../functions/server/getRelevantGuidedSteps";
 
 export type FixViewProps = {
   children: ReactNode;
@@ -96,6 +97,15 @@ export const FixView = (props: FixViewProps) => {
     );
   };
 
+  const currentStepIndex = stepLinks.findIndex((stepLink) =>
+    stepLink.href.startsWith(pathname)
+  );
+  const prevStepHref =
+    currentStepIndex <= 0
+      ? "/redesign/user/dashboard"
+      : stepLinks[currentStepIndex - 1].href;
+  const nextStepHref = stepLinks[currentStepIndex + 1].href;
+
   return (
     <div className={styles.fixContainer}>
       <div
@@ -111,7 +121,7 @@ export const FixView = (props: FixViewProps) => {
         <section className={styles.fixSection}>
           <Link
             className={`${styles.navArrow} ${styles.navArrowBack}`}
-            href="/redesign/user/dashboard"
+            href={prevStepHref}
             aria-label={l10n.getString("guided-resolution-flow-back-arrow")}
           >
             <Image alt="" src={ImageArrowLeft} />
@@ -119,7 +129,7 @@ export const FixView = (props: FixViewProps) => {
           <div className={styles.viewWrapper}>{props.children}</div>
           <Link
             className={`${styles.navArrow} ${styles.navArrowNext}`}
-            href="/redesign/user/dashboard"
+            href={nextStepHref}
             aria-label={l10n.getString("guided-resolution-flow-next-arrow")}
           >
             <Image alt="" src={ImageArrowRight} />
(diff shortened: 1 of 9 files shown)


src/app/(proper_react)/redesign/(authenticated)/user/dashboard/fix/FixView.tsx
```diff
@@ -96,6 +97,15 @@ export const FixView = (props: FixViewProps) => {
     );
   };
 
+  const currentStepIndex = stepLinks.findIndex((stepLink) =>
```
Vinnl, line 100: I figured I'd just let the previous and next arrows link to the previous and next step in the flow, but that doesn't quite work well, because I can't determine the current step based on the URL (the Scan step has multiple URLs, for example, depending on if the scan already ran).

Thus, I think I'll do a followup PR where I'll remove this component from the layout, and move it into every individual step, and just make the next step use the same `getRelevantGuidedSteps` function.

Vinnl: Yeah at least those, but I'm also weary of potential other pages where this is the case - either current ones, that I'm not aware of, or future ones. e.g. the celebration steps that we're eventually/already adding? Hence the followup PR to re-use the step determination logic.

src/app/(proper_react)/redesign/(authenticated)/user/dashboard/fix/data-broker-profiles/start-free-scan/page.tsx
```diff
@@ -2,16 +2,44 @@
...
 import { Button } from "../../../../../../../../components/server/Button";
+import { getLatestOnerepScanResults } from "../../../../../../../../../db/tables/onerep_scans";
+import { redirect } from "next/navigation";
+import { getServerSession } from "next-auth";
+import { authOptions } from "../../../../../../../../api/utils/auth";
+import { getOnerepProfileId } from "../../../../../../../../../db/tables/subscribers";
+import { getCountryCode } from "../../../../../../../../functions/server/getCountryCode";
+import { getL10n } from "../../../../../../../../functions/server/l10n";
+
+export default async function StartFreeScan() {
```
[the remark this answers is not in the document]
Vinnl: Which logic do you mean? I think Florian's PR has extracted a lot of the DashboardTopBanner logic into a separate function, but I think it needs to deal with way more permutations, and also do different things in response to them than I think we're doing here - but I must also admit that I don't have the full picture, so I might just need more clarification.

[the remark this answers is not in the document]
flozia: @codemist I don’t think the logic here is related to the dashboard top banner states. As far as I can tell these conditions here only prevent non-US and unauthenticated users from visiting this view.

src/app/functions/server/getRelevantGuidedSteps.ts
```diff
@@ -187,22 +179,20 @@ function isEligibleFor(data: InputData, stepId: StepLink["id"]): boolean {
...
-        return scanResult.status === "new" && !scanResult.manually_resolved;
-      }) &&
-      data.latestScanData?.scan?.onerep_scan_status === "finished"
-    );
+  if (stepId === "Scan") {
+    const hasRunScan =
+      typeof data.latestScanData?.scan === "object" &&
+      data.latestScanData?.scan !== null;
+    const hasResolvedAllScanResults =
+      data.latestScanData?.scan?.onerep_scan_status === "finished" &&
```
flozia, line 187: I think we also landed on counting scan results that have a status of `in_progress` as “resolved”. I guess that will only be relevant for automatic removal, though.

flozia (approved):
Looking good!

src/app/functions/server/getRelevantGuidedSteps.ts
Vinnl: Yeah I think you're right; once your permutations work lands we can add proper tests that actually verify the right links show up in the right situations (which you already mentioned in your test PR), which should've caught this. Nice :eagle: :eye: ! Fixed in 6f5ee8ba2.

Vinnl: Unit test is failing, but that is because the "scan in progress" notice only gets shown if the current step is `"Scan"`, which it no longer is with this change. I believe the PR you just merged changes that, so after a rebase and fixing the conflicts, it should succeed again.

PR #3486, opened 2023-09-29
MNTOR-2224 - add subscription array to converted FxaProfile

rhelmer:
# References: 
Jira: MNTOR-2224

# Description

In #3474 we made the way profiles are converted from `FxaProfile` to more standard OAuth `Profile` more standard, but I didn't include the `subscriptions` array.

# Screenshot (if applicable)


src/app/api/utils/auth.ts
```diff
@@ -201,8 +193,26 @@ async function fetchUserInfo(accessToken: string) {
       Authorization: `Bearer ${accessToken ?? ""}`,
     },
   });
-  const userInfo = (await response.json()) as Profile;
-  return userInfo;
+  const userInfo = await response.json();
+  return userInfo as Profile;
```
Vinnl, line 197: Actually, come to think of it, isn't this type assertion inaccurate? Which would explain why TypeScript didn't complain. (And relatedly, shouldn't line 82 still pass the return value through `convertFxaProfile`?)

```suggestion
  return userInfo as FxaProfile;
```

Vinnl (approved):
Change looks good, though I think the type assertion of `fetchUserInfo` is wrong, and that's why TS didn't tell you to also use this new function inside the `jwt` callback.

src/app/api/utils/auth.ts
rhelmer: Oh weird, good catch, hm... login is failing with this change though, there must be something wrong with the conversion function. I'm surprised it worked before.

rhelmer: (next-auth is complaining that `id` is missing when I clearly see it on the object)

rhelmer: Here's what it did before:
```
-        request: async (context) => {
-          const response = await fetch(AppConstants.OAUTH_PROFILE_URI, {
-            headers: {
-              Authorization: `Bearer ${context.tokens.access_token ?? ""}`,
-            },
-          });
-          const userInfo = (await response.json()) as Profile;
-          return userInfo;
-        },
```

I think what is happening here is that the next-auth `UserinfoEndpointHandler` that we're passing the `fetchUserInfo` function to ends up calling `profile`, which is where this `FxAProfile` - `Profile` conversion used to live.

I'll see if I can re-implement in a way that makes this clearer...

src/app/api/utils/auth.ts
```diff
@@ -62,32 +62,29 @@ export const authOptions: AuthOptions = {
         },
       },
       token: AppConstants.OAUTH_TOKEN_URI,
+      // This function calls `profile` below, which converts the `FxaProfile` -> `Profile`.
       userinfo: {
         request: async (context) =>
-          fetchUserInfo(context.tokens.access_token ?? ""),
+          fetchUserInfo(
+            context.tokens.access_token ?? ""
+          ) as unknown as Promise<Profile>,
```
rhelmer, line 70: @Vinnl does this clarify things or make it worse? :)

rhelmer: For context, this is what this code did previous to any of this refactoring:

```
request: async (context) => {
  const response = await fetch(AppConstants.OAUTH_PROFILE_URI, {
    headers: {
      Authorization: `Bearer ${context.tokens.access_token ?? ""}`,
    },
  });
  const userInfo = (await response.json()) as Profile;
  return userInfo;
},
```

I'm actually not sure I totally understand why this worked before...

Vinnl: I don't think this is right; if I log the return value of `fetchUserInfo`, it looks like this, which matches `FxaProfile`, not `Profile`:

```
{
    email: 'someaddress@mailinator.com',
    locale: 'en-GB,en-US;q=0.8,en;q=0.5,nl;q=0.3',
    amrValues: [ 'pwd', 'email' ],
    twoFactorAuthentication: false,
    metricsEnabled: true,
    uid: 'da4efbcfe51d466792e894accd119ff4',
    avatar: 'https://profile.stage.mozaws.net/v1/avatar/v',
    avatarDefault: false
  }
```

The type error does indicate that apparently `userInfo.request` does expect to return a `Profile`, so I suppose the return value is supposed to parse it into that as well?

I don't quite understand when `profile` is used then, but :shrug:

src/app/api/utils/auth.ts
Vinnl: Hmm, am I missing something? It looks like you have changed this now, and it still works for me? :)

src/app/api/utils/auth.ts
Vinnl: Hmm OK I see now, the way I defined the `Profile` type initially appears to be wrong - that's supposed to be what I called `FxaProfile`, and then confusing, the return value of the `profile` callback is a `User` object, that I hadn't defined yet. Submitted https://github.com/mozilla/blurts-server/pull/3503 that I think fixes this - this branch is its target branch.

PR #3488, opened 2023-10-02
convert the e2e suite to typescript

mozrokafor:
# References: 
Jira: MNTOR-2179

As most of monitor is now typescript. This PR is also converting our monitor e2e suite to typescript.


src/e2e/utils/helpers.ts
```diff
@@ -0,0 +1,112 @@
...
+  };
+
+  switch (true) {
+    case checkIfTitleContains("Enter your email"):
+      await enterYourEmail(page);
+      break;
+    case checkIfTitleContains("Enter your password"):
+      await enterYourPassword(page);
+      break;
+    // case checkIfTitleContains('Set your password'):
```
mansaj, line 100: What are these commented out code for? Did I do this?

src/e2e/pages/dashBoardPage.ts
```diff
@@ -0,0 +1,23 @@
...
+
+export class DashboardPage {
+  readonly page: Page;
+  readonly dataBreachEmailDropdown: Locator;
+  readonly siteFoundImage: Locator;
+  readonly breachStats: Locator;
+
+  constructor(page: Page) {
+    this.page = page;
+    this.dataBreachEmailDropdown = page.locator("custom-select");
```
mansaj, line 15: Thinking out loud here.. as a part of our smoke tests, we'd want to cover all locators so we can detect any that's returning null?

mansaj: I wonder if there's a way to check for coverage

mansaj (approved):
Run the tests locally, 4 failures but expected (bugs filed)

src/e2e/utils/helpers.ts
mozrokafor: looks like you commented them out 7 months ago, i have a [ticket](https://mozilla-hub.atlassian.net/browse/MNTOR-2246) that'll look into these plus any others that are lingering around

PR #3494, opened 2023-10-02
Add unit tests for dashboard banner content

flozia:
Adds additional unit tests for the changes introduced in PR #3493 for the dashboard banner content.


src/app/(proper_react)/redesign/(authenticated)/user/dashboard/Dashboard.test.tsx
```diff
@@ -541,7 +636,856 @@ it("shows scan in progress indicators on the dashboard with results and unresolv
...
+  const ComposedDashboard = composeStory(DashboardNonUsNoBreaches, Meta);
+  render(<ComposedDashboard />);
+
+  const dashboardTopBannerCta = screen.queryAllByRole("link", {
+    name: "Monitor more emails",
+  });
+  expect(
+    getRelevantBannerContentElement(dashboardTopBannerCta)
+  ).toBeInTheDocument();
+});
```
flozia, line 660: I’m only testing the title and CTA content right now. Still planning to add a check for the `href` of the CTA and paragraph content in a follow-up.

mansaj (approved):
Thanks for adding more tests

src/app/(proper_react)/redesign/(authenticated)/user/dashboard/Dashboard.test.tsx
```diff
@@ -210,7 +305,7 @@ it("shows the 'Start a free scan' CTA to free US-based users who haven't perform
   );
   render(<ComposedDashboard />);
 
-  const freeScanCta = screen.getByRole("link", { name: "Start a free scan" });
+  const freeScanCta = screen.queryByRole("link", { name: "Start a free scan" });
```
Vinnl, line 308: Heh, I'd expect changes the other way around - `getByRole` if you need to check if something exists, `queryByRole` if you're verifying that it _doesn't_ exist. See the docs [here](https://testing-library.com/docs/queries/about/#types-of-queries):

> **queryBy...** (...) This is useful for asserting an element that is not present.

The reason `getBy` is useful is that it throws a descriptive error if it doesn't match, telilng you what other elements of the given role it _did_ find, for example.

src/app/(proper_react)/redesign/(authenticated)/user/dashboard/Dashboard.test.tsx
```diff
@@ -543,55 +638,919 @@ it("shows scan in progress indicators on the dashboard with results and unresolv
...
 
-it("shows scan in progress indicators on the dashboard with resolved scan results and unresolved breaches", () => {
+// Check dashboard banner content for story DashboardNonUsNoBreaches
+it("shows the correct dashboard banner title for non-US users, no breaches", () => {
+  const ComposedDashboard = composeStory(DashboardNonUsNoBreaches, Meta);
+  render(<ComposedDashboard />);
+
+  const dashboardTopBannerTitle = screen.queryAllByText("No exposures found");
+  expect(
+    getRelevantBannerContentElement(dashboardTopBannerTitle)
```
Vinnl, line 648: I'm not sure I understand why `getRelevantBannerContentElement` is needed? Is there no way to get the relevant elements in a way that does [not rely on invisible class names](https://testing-library.com/docs/queries/about/#priority)? (And if there is a good reason, could that be added to the JSDoc of `getRelevantBannerContentElement`?)

Vinnl (commented):
Thanks for the long list of tests! I'm not quite sure I follow what's going on/why it's needed yet though, so a clarifying comment could be helpful to properly review.

(And if it's easy to fix with a search+replace, switching from `queryBy` to `getBy` would be nice, because people are probably going to be copy-pasting from these tests in the future :) )

src/app/(proper_react)/redesign/(authenticated)/user/dashboard/Dashboard.test.tsx
flozia: Thanks! Changed in [b05f5ac](https://github.com/mozilla/blurts-server/pull/3494/commits/b05f5ac82a4e5ec2d0a2319ad92b8690b64a0599).

src/app/(proper_react)/redesign/(authenticated)/user/dashboard/Dashboard.test.tsx
flozia: My thinking was the following: I’d like to be able to get and query the dashboard top banner elements as usual. It’s possible for other elements to match the same text, label and roles outside of the top banner. Adding test IDs to the elements that are being tested would get cucumbersome. That’s why I decided to add `getRelevantBannerContentElement` as a helper to filter out elements that should be not taken into account for certain tests.

Added a comment in [84fd600](https://github.com/mozilla/blurts-server/pull/3494/commits/84fd600c51683baf38f79ccd77aa0116d3ea14a4). Do you have a suggestion on how this could be improved?

flozia:
> Thanks for the long list of tests! I'm not quite sure I follow what's going on/why it's needed yet though, so a clarifying comment could be helpful to properly review.

I’ve mentioned in the description that this PR is adding unit tests for testing the dashboard top banner permutations in https://github.com/mozilla/blurts-server/pull/3493. I opened this PR against it to not add more lines of code there. I tried to add descriptive test names. Would you like me to add more clarifying comments inline?

src/app/(proper_react)/redesign/(authenticated)/user/dashboard/Dashboard.test.tsx
Vinnl: Ah, OK! Yes, that comment clarifies a lot. (Also, sorry, when I mentioned "why it's needed" in the overall review, I was referring to this specifically, not to the tests themselves :sweat_smile: )

However, this actually points out an area where we're currently lacking in accessibility. One way people with e.g. low vision or low dexterity navigate a page, is by first listing the [landmarks](https://developer.mozilla.org/en-US/blog/aria-accessibility-html-landmark-roles/) to determine the overall areas of the page, and then navigating within those landmarks - almost exactly what you're trying to do in these tests. But this top section isn't currently a labelled landmark (and neither is the main content below).

So, if you want to do The Right Thing™, I'd suggest wrapping the top banner in a `<section>` that either has an `aria-labelledby` pointing to the `<h3>` in there (which is a bit of a pain due to how the top banner is split up, and also I'm not sure how descriptive that heading is of its contents anyway), or add a direct `aria-label`. And then the area below the top banner would be the `<main>` part, I think.

You can then select the banner itself using

```tsx
  const banner = screen.getByRole("region", { name: "<the heading or aria-label content>" })
```

and then select elements _inside_ the banner using the generic (i.e. not a property on `screen`) `getByRole` function, with `banner` as its first element:

```tsx
  const dashboardTopBannerCta = getByRole(banner, "link", {
    name: "Monitor more emails",
  });
```

If you feel like that might be too much work and not worth the effort, let me know, then I'll approve anyway, but at least will also have made you feel bad :P (And possibly I'll make the change myself in a followup PR.)

flozia: That is a good point and I’m in favor of doing The Right Thing™. Since the top banners not being accessible is not due to recent changes I’ll follow up on this. The unit tests are valuable for making sure that the changes introduced in #3493 don’t break and it makes debugging them easier we should land them as soon as possible.

PR #3499, opened 2023-10-03
MNTOR-2229: dashboard number logic

mansaj:
# References: 
Jira: MNTOR-2229, MNTOR-2262

# Description
* The “Actions needed” screen would display all the numbers that are total - fixed (in progress + resolved), all the numbers will match and reflect the unresolved state no matter the context of the text etc.. This includes the progress card, chart, text on the left and section under View all sites where your info is exposed
* Once the user flips over to the “fixed” tab, it would reflect everything that’s fixed, all the numbers will also match
* Chart in fixed tab should display `# of exposure "Fixed"`

# Screenshot (if applicable)

Not applicable.

# How to test


mansaj:
The diff will be significantly smaller once https://github.com/mozilla/blurts-server/pull/3493 merges in

src/app/functions/server/dashboard.test.ts
```diff
@@ -472,55 +478,106 @@ describe("getExposureReduction", () => {
...
+      ).toBeGreaterThanOrEqual(0);
+    }
+
+    summary.unresolvedSanitizedExposures.forEach((_, v) => {
+      expect(v).toBeGreaterThanOrEqual(0);
+    });
+
+    summary.inProgressFixedSanitizedExposures.forEach((_, v) => {
+      expect(v).toBeGreaterThanOrEqual(0);
+    });
```
pdehaan, line 527: What does this do?
I thought the second arg of [`.forEach()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/forEach)'s callback function was the `index` which is always 0+:
> `index`
> The index of the current element being processed in the array.

For example, this will log `{v: 0}`, `{v: 1}`, `{v: 2}`:
```js
['a', 'b', 'c'].forEach((_, v) => console.log({v}));
```

So when would `v` less than 0?

src/app/functions/server/dashboard.ts
```diff
@@ -274,16 +357,52 @@ export function getDashboardSummary(
   // count unique breaches
   summary.dataBreachTotalNum = subscriberBreaches.length;
   const isBreachesOnly = summary.dataBrokerTotalNum === 0;
-  summary.sanitizedExposures = sanitizeExposures(
-    summary.allExposures,
-    summary.totalExposures,
+
+  // calculate total exposures
+  // summary.totalExposures = summary.dataBreachTotalExposuresNum + summary.dataBrokerTotalExposuresNum;
```
pdehaan, line 362: Q: is this meant to be commented out?

src/app/functions/server/dashboard.ts
```diff
@@ -311,11 +430,8 @@ function sanitizeExposures(
...
-export function getExposureReduction(
-  summary: DashboardSummary,
-  scanResultItems: OnerepScanResultRow[]
-): number {
-  const countOfDataBrokerProfiles = scanResultItems.length;
-  const totalExposures = summary.totalExposures;
-  return Math.round((countOfDataBrokerProfiles / totalExposures) * 100);
+export function getExposureReduction(summary: DashboardSummary): number {
+  return Math.round(
+    (summary.dataBrokerTotalExposuresNum / summary.totalExposures) * 100
```
pdehaan, line 435: Q: Any chance `totalExposures` can be zero here and cause a rift in the universe?

src/app/functions/server/dashboard.test.ts
mansaj: @pdehaan These elements of the array are key value pairs, so the v would be a `number` and the k would be the key. They should look something like
```
[
{ 'physical-addresses': 2 },
{ 'family-members-names': 10 }
]
```

src/app/functions/server/dashboard.ts
mansaj: Thanks Peter, meant to delete.. deleted.

src/app/functions/server/dashboard.ts
mansaj: Thanks Peter, added edge case and test

src/app/functions/server/dashboard.test.ts
pdehaan: OK, I think I'm still confused, but maybe I need to clone this PR and see if I can run+`console.log()` locally to see if the `v` is 0,1 or 2,10.
From your snippet above it does seem like you're using `Array#forEach()` (and not `Map#forEach()` or `Set#forEach()`), so I'm curious how it's destructing key:value pairs here without something like `Object.entries()` to get the actual values. But it's 8:22PM and I'm just eyeballing it in GitHub, so I'm probably overlooking somethin'.

flodolo (approved):

src/app/functions/server/dashboard.ts
```diff
@@ -269,21 +358,58 @@ export function getDashboardSummary(
...
+  summary.dataBreachUnresolvedNum =
+    summary.dataBreachTotalNum - summary.dataBreachResolvedNum;
   const isBreachesOnly = summary.dataBrokerTotalNum === 0;
-  summary.sanitizedExposures = sanitizeExposures(
-    summary.allExposures,
-    summary.totalExposures,
+
+  // count unresolved exposures
+  summary.unresolvedExposures = Object.keys(summary.allExposures).reduce(
+    (a, k) => {
```
Vinnl, line 373: Not entirely sure, but I believe you can save yourself a couple of type assertions below if you add type parameters for the function arguments:

```suggestion
    (a: Exposures, k: keyof Exposures) => {
```

Vinnl (approved):
I believe you verified the numbers with Tony already, so I didn't check those. As for the code, I didn't see any major roadblocks there :)

src/app/functions/server/dashboard.test.ts
Vinnl: I think Peter's right; if every element is indeed an object with a single property with a single value, then you'll still need to iterate over the entries of that object, e.g.:

```suggestion
    summary.unresolvedSanitizedExposures.forEach((unresolvedSanitizedExposure) => {
      Object.values(unresolvedSanitizedExposure).forEach(count => {
        expect(count).toBeGreaterThanOrEqual(0);
      });
    });

    summary.inProgressFixedSanitizedExposures.forEach((inProgressFixedSanitizedExposure) => {
      Object.values(inProgressFixedSanitizedExposure).forEach(count => {
        expect(count).toBeGreaterThanOrEqual(0);
      });
    });
```

What you could alternatively do, is instead of `type SanitizedExposures = Array<Record<string, number>>`, make it an array of arrays with two elements (aka an array of tuples): `type SanitizedExposures = Array<[string, number]>;`. Then, you should be able to destructure it inline:

```suggestion
    summary.unresolvedSanitizedExposures.forEach(([_exposureType, count]) => {
      expect(count).toBeGreaterThanOrEqual(0);
    });

    summary.inProgressFixedSanitizedExposures.forEach(([_exposureType, count]) => {
      expect(count).toBeGreaterThanOrEqual(0);
    });
```

You should be able to verify that the current method doesn't work by modifying the code to actually set a negative value and see that this test doesn't actually fail.

src/app/(proper_react)/redesign/(authenticated)/user/dashboard/View.tsx
```diff
@@ -162,18 +162,31 @@ export const View = (props: Props) => {
...
 
     if (hasUnresolvedExposures) {
       exposuresAreaDescription = l10n.getString(
         "dashboard-exposures-area-description",
         {
-          exposures_total_num: totalExposures,
-          data_breach_total_num: dataBreachTotalNum,
-          data_broker_total_num: dataBrokerTotalNum,
+          exposures_total_num:
+            totalExposures -
```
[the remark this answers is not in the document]
mansaj: Renamed thanks

src/app/functions/server/dashboard.ts
mansaj: Tried that previously, and this is the error that occurs:
```
No overload matches this call.
  Overload 1 of 3, '(callbackfn: (previousValue: string, currentValue: string, currentIndex: number, array: string[]) => string, initialValue: string): string', gave the following error.
    Argument of type '(a: Exposures, k: keyof Exposures) => Exposures' is not assignable to parameter of type '(previousValue: string, currentValue: string, currentIndex: number, array: string[]) => string'.
      Types of parameters 'a' and 'previousValue' are incompatible.
        Type 'string' is not assignable to type 'Exposures'.
 ...
```

src/app/functions/server/dashboard.test.ts
mansaj: Oh right, thanks @Vinnl @pdehaan, confused myself here

src/app/(proper_react)/redesign/(authenticated)/user/dashboard/DashboardTopBanner/DashboardTopBannerContent.tsx
```diff
@@ -254,13 +273,19 @@ export const DashboardTopBannerContent = (props: DashboardTopBannerProps) => {
...
+                  vars: {
+                    remaining_exposures_total_num:
+                      bannerData.totalExposures -
+                      bannerData.dataBreachFixedExposuresNum -
+                      bannerData.dataBrokerFixedExposuresNum -
+                      bannerData.dataBrokerInProgressExposuresNum,
+                  },
+                  elems: {
+                    b: <strong />,
+                  },
```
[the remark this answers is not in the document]
mansaj: Removed the `elems`, don't see it in `premium.ftl`. Thanks

src/app/functions/server/dashboard.test.ts
pdehaan: Or if you don't want the tuple route, I used this `Array#reduce()` mess to build my test case (which uses `Array#concat()` and `Object.values()` magic to build an array of values; conversely I guess we could flatten all the objects into a single object and then call `Object.values()` once):

```js
const assert = require("node:assert/strict");

const tmp = [
  { 'physical-addresses': 10 },
  { 'family-members-names': -2 }
];

const values = tmp.reduce((acc, item) => acc.concat(Object.values(item)), []);

// [10, -2]
console.log(values);

// "AssertionError: -2 should be >= 0"
values.forEach(value => assert(value >= 0, `${value} should be >= 0`));
```

locales-pending/premium.ftl
```diff
@@ -199,21 +206,21 @@ dashboard-exposures-filter-reset = Reset
...
 dashboard-top-banner-scan-in-progress-cta = See what’s ready now
 
 dashboard-top-banner-protect-your-data-title = Let’s protect your data
 # Variables:
-# $data_breach_total_num is the total number of data breaches the user has.
-# $data_broker_total_num is the total number of data brokers selling the user’s data.
-dashboard-top-banner-protect-your-data-description = We found your data in { $data_breach_total_num } data breaches and { $data_broker_total_num } sites selling your personal info. We’ll guide you step-by-step on how to fix it.
+# $data_breach_unresolved_num is the unresolved number of data breaches the user has.
+# $data_broker_unresolved_num is the unresolved number of data brokers selling the user’s data.
+dashboard-top-banner-protect-your-data-description = We found your data in { $data_breach_unresolved_num } data breaches and { $data_broker_unresolved_num } sites selling your personal info. We’ll guide you step-by-step on how to fix it.
```
pdehaan, line 223: Not a regression, but I did have a drive-by l10n question about plurals...
Is it possible that `$data_breach_unresolved_num` or `$data_broker_unresolved_num` would be singular?

For example:
> "We found your data in _**1 data breaches**_ and _**1 sites**_ selling your personal info. We'll guide you step-by-step on how to fix it."

mansaj: This is a good one, Peter. I think we will address this in another PR (see slack)

PR #3503, opened 2023-10-04
Fix Next-Auth types

Vinnl:
This is re: https://github.com/mozilla/blurts-server/pull/3486#discussion_r1345652068

I think they represent what they should know. This also separates out the FxA OAuth config from the Next-Auth user session management config, which should hopefully also help with the mental model.

So we now have `Profile`, which represents the shape of the data as returned by FxA's `/userinfo/` endpoint. Then we have the `User`, representing the user object we give to Next-Auth.


rhelmer (approved):
This is a good idea, thanks!

PR #3500, opened 2023-10-03
Make right-arrow skip links work

Vinnl:
# References: 
Jira: MNTOR-2116
Figma: N/A

# Description

This builds on https://github.com/mozilla/blurts-server/pull/3485. The unit tests seem to trip over `getL10n` - we haven't written tests for server components before. I'll try to figure that out, but since that issue is probably independent from these changes, I'm already submitting this for review.

Unfortunately, this is a fairly significant rework, since the
`<FixView>` needed access to the data required to determine the next
step, which often was already available in the relevant `page.tsx`,
so to be able to pass that data into the `<FixView>`, I had to move
it from the layout into every individual page.

In addition to the right-arrow skip links, this also sets the
correct target for the skip link in the manual broker resolution.

To make debugging easier, I've added Storybook stories for every page in the guided resolution flow.

# How to test

Use the right arrows through the guided resolution flow. I'm not sure if that's currently actually possible to properly test locally, but the target URLs appear to be right in Storybook.

# Checklist (Definition of Done)
- [x] Localization strings (if needed) have been added.
- [x] Commits in this PR are minimal and [have descriptive commit messages](https://chris.beams.io/posts/git-commit/).
- [x] I've added or updated the relevant sections in readme and/or code comments
- [ ] I've added a unit test to test for potential regressions of this bug.
- [ ] Product Owner accepted the User Story (demo of functionality completed) or waived the privilege.
- [x] All acceptance criteria are met.
- [ ] Jira ticket has been updated (if needed) to match changes made during the development process.
- [ ] Jira ticket has been updated (if needed) with suggestions for QA when this PR is deployed to stage.

PR #3504, opened 2023-10-04
Upgrade Next.js and don’t break e2e tests

flozia:
~Playwright tests are failing since PR #3501: Testing a couple of things to see if they pass in CI.~

We reverted PR #3501 which was upgrading to Next.js. This is another attempt at upgrading and not having the `e2e` tests fail. After the changes in this PR, I was not able to reproduce the Playwright failure anymore. For context: We are using a [non-standard NODE_ENV](https://nextjs.org/docs/messages/non-standard-node-env) and it seems like the previous version of Next.js did not make Playwrights web server stumble. Building and running the application without Playwright still works fine, though.

### DIFF
--- a/.env-dist
+++ b/.env-dist
@@ -1,6 +1,3 @@
-# https://nextjs.org/docs/messages/non-standard-node-env
-# development, production, test
-NODE_ENV=development
 # local, heroku, stage, production
 APP_ENV=local
 NEXT_PUBLIC_APP_ENV=local
--- a/locales/gn/app.ftl
+++ b/locales/gn/app.ftl
@@ -164,7 +164,7 @@ how-fxm-2-blurb =
     ñemboguapyre, oĩhápe marandu’i ojehayhúva.
 how-fxm-3-headline = Erekóta marandu’i kundahárape
 how-fxm-3-blurb =
-    Eipurúramo { -brand-name }, og̃uahẽta ndéve marandu’i eikéramo 
+    Eiporúramo { -brand-name }, og̃uahẽta ndéve marandu’i eikéramo 
     tenda ñembyaipyrépe. Ehecháke eimépara’e pe ñembyaípe ha mba’épa ikatu ejapo péva renondépe.
 wtd-after-website = Mba’e ejapóta ñanduti kuatiarogue oñembyai rire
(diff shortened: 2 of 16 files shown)


rhelmer:
@flozia I reverted the whole commit 452a6827697a69bbee33690016ad3e585bbebe2f - if you want to use this to figure out exactly what's failing (and ideally fix it if it's Monitor app doing something wrong!) that would be welcome!

flozia:
Test failures after upgrading Next.js solved in https://github.com/mozilla/blurts-server/pull/3500#issuecomment-1748954605.

Vinnl:
@flozia That only fixes unit test failures - unfortunately I still don't know how to make the Playwright tests succeed :/

flozia:
@Vinnl It seems like this fixing the issue (for now) and should unblock PR #3504.

.env-dist
```diff
@@ -1,6 +1,6 @@
 # https://nextjs.org/docs/messages/non-standard-node-env
 # development, production, test
-NODE_ENV=development
+NODE_ENV=production
```
pdehaan, line 3: Wait... why this change? Does this affect what tests run? Ah, maybe https://github.com/mozilla/blurts-server/blob/c0aa2eb9cf0ce41b74d1527eadf8a78c26977569/.github/workflows/playwright.yml#L51

flozia: Yes, that must be it! `NODE_ENV=development` is not a valid env variable when building Next.js with `npm run build`. This has been biting us in the past few days. Playwright only started failing on us after upgrading to`next: 13.5.4`. Building standalone without Playwright was still working.

Vinnl (approved):
Great job finding this Florian! As an alternative solution, I think we should probably remove the `NODE_ENV` value altogether - Next.js sets it to `development` for `npm run dev` and `production` for `npm run build` [by default](https://nextjs.org/docs/app/building-your-application/configuring/environment-variables#good-to-know):

> If the environment variable NODE_ENV is unassigned, Next.js automatically assigns development when running the next dev command, or production for all other commands.

By listing it in `.env-dist`, we might imply that it's something we should be setting ourselves, while in general, I think we just want to leave it to Next.js and other relevant tooling (e.g. Jest sets it to `test`).

Vinnl:
https://www.youtube.com/watch?v=_ElWlUr_tw8

PR #3507, opened 2023-10-04
Hide Premium upsell for non-US users

flozia:
# References: 
https://github.com/mozilla/blurts-server/pull/3493#pullrequestreview-1652704266

# Description
This PR adds a context for the country code and hides the Premium upsell button for non-US users.

# How to test
Storybook: [Dashboard](http://localhost:6006/?path=/story/pages-dashboard--dashboard-non-us-no-breaches)


Vinnl (approved):
Hmm, I'm a bit weary of adding more implicit dependencies (context), especially since it can only be accessed in client components anyway. Which is not to say I can't be convinced of the benefit (especially since this value shouldn't really change during the user's session), but maybe you could outline why you decided to go with context rather than passing props?

Otherwise the changes seem to make sense.

flozia:
> Hmm, I'm a bit weary of adding more implicit dependencies (context), especially since it can only be accessed in client components anyway.

Since we are already relying on using `context` from third parties, integral parts of the application (localization and user sessions) rely on it already and it’s native React functionality I don’t have concerns using it.

> Which is not to say I can't be convinced of the benefit (especially since this value shouldn't really change during the user's session), but maybe you could outline why you decided to go with context rather than passing props?

While adding and passing around an increasing number of props I found that it’s starting to get a bit convoluted in components. Being able to get the `countryCode` directly in the client component is helpful in my opinion, since we already have a lot of views that should behave differently to users outside of the US — and it’s likely to become more.

That being said: If there is still any concern about using `context` for storing and accessing the country code from client components I am also on board with just passing it down as a prop. Let me know what you prefer @Vinnl @codemist.

src/app/(proper_react)/redesign/(authenticated)/user/dashboard/Dashboard.stories.tsx
```diff
@@ -153,31 +154,32 @@ const DashboardWrapper = (props: DashboardWrapperProps) => {
...
+            user,
+            countryCode: props.countryCode,
+          })}
+          isEligibleForFreeScan={props.countryCode === "us" && !scanData.scan}
+          featureFlagsEnabled={{
+            FreeBrokerScan: true,
+            PremiumBrokerRemoval: true,
+          }}
+          monthlySubscriptionUrl={""}
+          yearlySubscriptionUrl={""}
```
pdehaan, line 179: Do these need to be set via some Env var or something? (doesn't look like this is a new change, but I see we're proxying it below).

flozia: For the Stories we do not need to set them, I think. In the places that are user-facing we have them set via env variables.

src/app/(proper_react)/layout.tsx
```diff
@@ -3,18 +3,25 @@
...
 import { getLocale } from "../functions/universal/getLocale";
 import { L10nProvider } from "../../contextProviders/localization";
 import { ReactAriaI18nProvider } from "../../contextProviders/react-aria";
+import { CountryCodeProvider } from "../../contextProviders/country-code";
+import { getCountryCode } from "../functions/server/getCountryCode";
 
 export default function Layout({ children }: { children: ReactNode }) {
   const l10nBundles = getL10nBundles();
+  const headersList = headers();
+  const countryCode = getCountryCode(headersList);
```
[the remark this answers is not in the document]
flozia: Eligibility for Premium is determined by country code which is determined by location. In [getCountryCode](https://github.com/mozilla/blurts-server/blob/main/src/app/functions/server/getCountryCode.ts#L10) we use the header `X-Client-Region` which we get from GCP. That works on staging and production and there we are usin the GeoIP. `headers.get("Accept-Language")` is only a fallback for when it’s not available, which is the case locally and in our dev environment.

Vinnl:
> Since we are already relying on using context from third parties, integral parts of the application (localization and user sessions) rely on it already and it’s native React functionality I don’t have concerns using it.

It's not the use of context in general that I'm wary of; I'm more afraid of a potential trend where we're going to have a deep tree of nested context providers and defaulting to context over props for many things, leading to harder-to-follow code and potential performance bottlenecks. In this case I think I could go either way, since this value is used in many leaves of our component tree and doesn't change often, so I can live with this change. I just wouldn't want slapping another context on our app to become a decision we start taking lightly :)

PR #3510, opened 2023-10-05
Don’t show list of emails on breach cards

flozia:
# References: 
Jira: [MNTOR-2222](https://mozilla-hub.atlassian.net/browse/MNTOR-2222)

# Description

Don’t show a list of emails on breach cards. See Story: http://localhost:6006/?path=/story/pages-dashboard--dashboard-us-premium-resolved-scan-unresolved-breaches.

# Screenshot

| Before | After |
| --- | --- |
| mntor-2222-before | mntor-2222-after |


flozia:
> Aren't we only hiding emails from data broker cards (free user), not data breaches?

You are correct! I also named the branch `MNTOR-2222-Email-on-data-broker-cards` only to pursue with doing the exact opposite. Fixed!

rhelmer (approved):

PR #3522, opened 2023-10-09
MNTOR-1404/data removal analytics

rhelmer:
# References: 
Jira: MNTOR-1404

# Description

Add Winston logger with optional GCP Cloud Logging support. Running locally should just print to console, running in stage/prod will print to console and also emit structured GCP logs.

I've added a few simple metrics to start:

1. Server-side `page_view` with pathname (see Screenshot)
2. All actions from the OneRep module (scan creation and completion etc)

# Screenshot (if applicable)

Structured logging allows us to easily route messages to BigQuery and write simple queries and dashboards:

image

image

Not applicable.

# How to test

I tested this by running in a GCP sandbox/dev environment.


src/app/layout.tsx
```diff
@@ -7,12 +7,31 @@ import { Metadata } from "next";
...
+
+const logger = createLogger({
+  level: "info",
+  transports: [new transports.Console()],
+});
+
+if (["stage", "production"].includes(process.env.APP_ENV ?? "local")) {
+  logger.transports.push(loggingWinston);
+}
+
```
rhelmer, line 34: We should consider sharing this (similar to how we share the knex config) since I think it'll be the same for every file we want to use this from.

src/app/layout.tsx
```diff
@@ -46,6 +65,12 @@ export default async function RootLayout({
 }: {
   children: ReactNode;
 }) {
+  // @ts-ignore FIXME testing server-side metrics
+  const pathname = children?.props?.childProp?.segment;
```
rhelmer, line 69: @flozia / @Vinnl - is this the best way to get the pathname from a server-side component? Note that it shows something different when using the dev server (e.g. `(nextjs_migration)`), it was the pathname with a real HTTP server when I tested it.

src/app/layout.tsx
```diff
@@ -46,6 +65,12 @@ export default async function RootLayout({
...
+  logger.info("page_view", {
+    pathname,
+    NODE_ENV: process.env.NODE_ENV,
+  });
```
rhelmer, line 73: BTW we should be able to replace `mozlog` with this, and it emitting events should continue to be this simple I think.

src/app/layout.tsx
Vinnl: Yeah, I'd expect to be able to import `logger` from a module and start logging right away?

package.json
```diff
@@ -85,7 +86,9 @@
     "react-cookie": "^4.1.1",
     "react-dom": "^18.2.0",
     "react-stately": "^3.23.0",
-    "uuid": "^9.0.0"
+    "request": "^2.88.2",
```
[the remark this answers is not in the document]
Vinnl: Possibly related: the upstream bug that I think originally led us to add it is now also closed: https://github.com/googleapis/retry-request/issues/103#event-10621035485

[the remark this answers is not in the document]
rhelmer: Thanks - I keep adding it back because building with various GCP-related modules complain about it (even though it's not listed as a dependency) - that PR makes me not want to take it! I'll make sure our deps are up to  date and see if it goes away, thanks.

src/app/layout.tsx
rhelmer: OK cool I'll move this into a shared location. Would `functions/server` be appropriate or is there a better place? It's definitely a server-side-only thing (although we could proxy messages from the client -> API -> server anyway to make our data collection first-party, right now Glean.js submits to the `incoming.telemetry.mozilla.org` endpoint)

src/app/layout.tsx
rhelmer: I'd like a more reliable way to capture things like impressions and clicks, and the server is a pretty reliable way to do that. We already have HTTP access logs that are in a structured format from nginx that we could use here if it doesn't make sense to use NextJS for it, that might be a better way to go.

rhelmer: I'd like to just move as much to first-party and server-side collection as possible, and for things where client events are useful (and the user isn't blocking or opting out) we can send those and join on them.

There is a line somewhere between "server-side logs" and "tracking user behavior that must be deleted if users request" which is something on my mind as well. Somewhat ironically, in order to support deletion requests, we need to be able to associate data with a particular user, which may be something we need NextJS to log in a structured way that gets stored such that we can act on things like deletion requests.

We don't really need to be logging this here though, I'll go ahead and remove it and we can tackle impressions and clicks separately, that really deserves its own discussion.

src/app/functions/server/logging.ts
```diff
@@ -0,0 +1,22 @@
...
+const loggingWinston = new LoggingWinston({
+  labels: {
+    name: "monitor-stats",
+    version: "0.1.0",
+  },
+});
+
+export const logger = createLogger({
+  level: "info",
+  transports: [new transports.Console()],
```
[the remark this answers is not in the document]
rhelmer: I think `test` or `development`, so you can see them in local development?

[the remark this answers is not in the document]
Vinnl: I don't have a strong opinion on showing them in `development` - it could get noisy, but it could also be useful when debugging. But in tests I think they're very likely to be noise indeed.

PR #3524, opened 2023-10-10
MNTOR-2283: premium in progress scans changes

mansaj:
# References: 
Jira: MNTOR-2283
Figma: 

# Description
Added a function for counting existing scans for a user/profile. The dashboard will use this count to determine if a user is still "in progress":
* a returning user with data broker scans already is no longer possible to be "in progress"
* a new user without data broker scans ever may be considered "in progress"

# Screenshot (if applicable)

Not applicable.

# How to test


src/db/tables/onerep_scans.ts
```diff
@@ -168,6 +168,18 @@ async function getScansCount(
     .andWhere("onerep_scan_reason", scanReason);
 }
 
+async function getScansCountForProfile(
+  onerepProfileId: number
+): Promise<number> {
+  return parseInt(
```
pdehaan, line 174: Why are we casting as string and then calling `parseInt()` to convert back to a number?
Wouldn't knex's `.count()` be returning a number directly, or is the count value so high it's converting to a string in some cases here?

mansaj: The .count() can return number or string (empty) I believe. So we cannot just infer `as number`

flozia (approved):

PR #3525, opened 2023-10-10
Fetch all enabled feature flags in one go

Vinnl:
This allows us to pass the enabled feature flags around in the application without needing to introduce asynchronicity and new database queries everywhere we want to check a particular flag. (This is particularly relevant if we want to refer to them in client components.)

Note that this does come with a slight behavioural change: if a flag has a non-empty allow_list, not passing a user will no longer mark that flag as enabled. I expect that the previous behaviour was actually unintended.


src/app/functions/server/featureFlags.ts
```diff
@@ -1,75 +0,0 @@
...
-    return false;
-  }
-
-  if (!flag.isEnabled) {
-    warn("Flag is not enabled:", flag.name);
-    return false;
-  }
-
-  if (!user) {
-    return true;
```
Vinnl, line 68: Just a note: as I understand it, this resulted in `isFlagEnabled` being `true` even if there was an `allow_list` for the flag, as long as you did not pass in a user. In my implementation (in `/db/tables/featureFlags.ts`), the equivalent of not passing in a user is to call `getFlagsEnabledForEveryone`, but it will return `false` if `allow_list` is a non-empty array. Is assume that that's the actual intended behaviour of the allowlist.

rhelmer: That was what I was thinking originally, but I think in practice it ended up being based on the way `isFlagEnabled` is called - if the `session` argument is passed then the allow list is considered, and if it is not then the allow list is not.

I'd like to figure out how to be more explicit. Maybe we should just have a boolean column that signifies this is globally enabled or not?

rhelmer (approved):

src/app/functions/server/featureFlags.ts
Vinnl: That sounds really unintuitive to me - that means you can add email addresses to an allowlist, and then accounts with different email addresses could still have features behind that flag show up.

But I updated the PR to align with that older behaviour, hopefully calling that out somewhat explicitly in the API by calling it `ignoreAllowlist`: 6b7315cb6

But yes, an explicit boolean that enables it for everyone (and clears the allowlist) would work too. Python's Waffle e.g. has an `--everyone` switch in addition to allowing enabling flags for specific users (or e.g. all authenticated users).

PR #3533, opened 2023-10-11
MNTOR-2226: CTA for exposure cards (breaches)

mansaj:
# References: 
Jira: MNTOR-2226
Figma: 

# Description
Making sure CTA works for breaches (maybe even data brokers)

* user hasn’t done their free scan yet: takes user to free scan prompt
* user is a free user: takes user to the data broker results
* user is a premium user: takes them to their first data breach resolution card that is needed to be resolved out of all their breaches
* exposure card auto closes

# Screenshot (if applicable)

Not applicable.

# How to test


src/app/(proper_react)/redesign/(authenticated)/user/dashboard/View.tsx
```diff
@@ -115,6 +116,32 @@ export const View = (props: Props) => {
   const tabSpecificExposures = getTabSpecificExposures(selectedTab);
   const filteredExposures = filterExposures(tabSpecificExposures, filters);
 
+  // MNTOR-2226
+  const getExposureCardLink = (exposure: Exposure) => {
+    let link = "/redesign/user/dashboard";
```
pdehaan, line 121: Q: is this default value used? I see some `if..else if..{if..else}..else` below, but I think every branch is overwriting this default value. Unless... there is a state where you're not eligible or a free scan (L125) _or_ eligible for premium (L128)... 🤔 

Conversely, do we need a temp var and should we just `return` early? (so instead of `link = 'abc'` just do `return 'abc'` and short circuit everything).

src/app/(proper_react)/redesign/(authenticated)/user/dashboard/View.tsx
```diff
@@ -118,28 +118,23 @@ export const View = (props: Props) => {
...
   // MNTOR-2226
   const getExposureCardLink = (exposure: Exposure) => {
-    let link = "/redesign/user/dashboard";
-
     if (!isScanResult(exposure)) {
       // if it's a data breach exposure
       if (props.isEligibleForFreeScan) {
-        link =
-          "redesign/user/dashboard/fix/data-broker-profiles/start-free-scan";
+        return "redesign/user/dashboard/fix/data-broker-profiles/start-free-scan";
```
pdehaan, line 124: Semi-global to this fn, but do these paths need leading `/` for absolute URLs (similar to L137 below)?

src/app/(proper_react)/redesign/(authenticated)/user/dashboard/fix/data-broker-profiles/manual-remove/ManualRemoveView.tsx
```diff
@@ -2,6 +2,9 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
+"use client";
```
mansaj, line 5: Screenshot 2023-10-15 at 8 28 40 PM
@Vinnl getting this client component error... is the only way to fix this to expose it as a server component?

Vinnl: Ah yes, so now that you're using client-side state, this needs to be a client component.

However, Fluent's `ReactLocalization` class is derived from the user's `Accept-Language`, which we need to access in different ways on the client and on the server. So for server components, we use `getL10n`, but if you're turning this into a client component, you'll have to replace that with the `useL10n` React hook, which should fix that error.

src/app/(proper_react)/redesign/(authenticated)/user/dashboard/View.tsx
```diff
@@ -47,6 +47,7 @@ export type Props = {
   userScanData: LatestOnerepScanData;
   isEligibleForFreeScan: boolean;
   isEligibleForPremium: boolean;
+  isPremiumUser: boolean;
```
Vinnl, line 50: Instead of adding yet another parameter, can't we just derive this from `props.user`?

```suggestion
```

src/app/(proper_react)/redesign/(authenticated)/user/dashboard/View.tsx
```diff
@@ -114,30 +116,44 @@ export const View = (props: Props) => {
   const tabSpecificExposures = getTabSpecificExposures(selectedTab);
   const filteredExposures = filterExposures(tabSpecificExposures, filters);
 
+  // MNTOR-2226
+  const getExposureCardLink = (exposure: Exposure) => {
```
Vinnl, line 120: Hmm, I think we should use the `getNextGuidedStep` function here instead. Right now, we're hardcoding e.g. a link to the `high-risk-data-breaches` page, but we'll want to skip that page if the user does not have high-risk breaches.

IIRC the reason we wanted to only have a single card open at all times, is that this CTA would go to the same location for each of them, so I don't think we'd need conditionals here to make it depend on the specific exposure.

(I also think Mel had a different button label in mind because of that, but you might want to check with her on that.)

Vinnl (commented):
The code looks good, but I think the target of the buttons is currently incorrect - unless I'm misunderstanding what they should do?

src/app/(proper_react)/redesign/(authenticated)/user/dashboard/View.tsx
mansaj: Thanks, fixed

src/app/(proper_react)/redesign/(authenticated)/user/dashboard/fix/data-broker-profiles/manual-remove/ManualRemoveView.tsx
mansaj: Oh didn't know we have a hook for this. That makes it easier! Thanks!

src/app/(proper_react)/redesign/(authenticated)/user/dashboard/View.tsx
mansaj: Oh, so I'm used to have a clear separation of controller / view where the view is purely presentation and as simple as possible in terms of not containing any logic or dependency. It doesn't seem that we abide by that principle too much. I'm ok blurring the line here if we don't have that type of convention

src/app/components/client/ExposureCard.tsx
```diff
@@ -59,15 +60,13 @@ export type ScanResultCardProps = {
   locale: string;
   isPremiumBrokerRemovalEnabled: boolean;
   resolutionCta: ReactNode;
-  isExpanded?: boolean;
   isPremiumUser: boolean;
+  isExpanded?: boolean;
```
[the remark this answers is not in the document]
mansaj: done thanks

src/app/(proper_react)/redesign/(authenticated)/user/dashboard/View.tsx
mansaj: Very cool, looked at the function, yea I think you are right, it would be fitting. I skipped the second arg for `getNextGuidedStep`, see L138

src/app/(proper_react)/redesign/(authenticated)/user/dashboard/View.tsx
Vinnl: Oh yeah we've definitely not been doing MVC, though I see how the name "View" is confusing :sweat_smile: The distinction here is more server-side logic vs. client-side logic - it's fine to have logic in the view, as long as it's related to what to display to the user. We could discuss further splitting up the views into [container and presentational components](https://medium.com/@dan_abramov/smart-and-dumb-components-7ca2f9a7c7d0) in the tech meeting, if you're interested - but see the disclaimer at the top of that article :)

src/app/(proper_react)/redesign/(authenticated)/user/dashboard/fix/data-broker-profiles/manual-remove/ManualRemoveView.tsx
```diff
@@ -7,7 +7,8 @@
 import { useState } from "react";
 import { Session } from "next-auth";
 import styles from "./ManualRemoveView.module.scss";
-import { getL10n } from "../../../../../../../../functions/server/l10n";
+import { useL10n } from "../../../../../../../../hooks/l10n";
+// import { getL10n } from "../../../../../../../../functions/server/l10n";
```
Vinnl, line 11:
```suggestion
```

Vinnl (approved):
Looking good, and thanks for the tests!

PR #3541, opened 2023-10-11
MNTOR-2265: Skip first onboarding step when clicking “Start free scan”

flozia:
# References: 
Jira: [MNTOR-2265](https://mozilla-hub.atlassian.net/browse/MNTOR-2265)

# Description

When users are coming from the resolution flow to the onboarding: Directly send them to the “Enter info” step.

# How to test

Note: The steps below will be only fully testable after we implement [MNTOR-2264](https://mozilla-hub.atlassian.net/browse/MNTOR-2264). Instead of navigating to the second step of the “Welcome flow” via the CTAs the essence of this PR can be tested by visiting directly `/redesign/user/welcome/free-scan`.

1. Make sure your user:
    - Has no OneRep profile
    - Has not run a scan
    - Has no scan results
    - Is not subscribed to Premium
    - Is a returning user
3. Visit `/redesign/user/dashboard` and `/redesign/user/dashboard/fix/data-broker-profiles/start-free-scan`
4. Click “Start free scan”
5. You should be directed to the “Enter info” step of the Premium welcome onboarding

[MNTOR-2265]: https://mozilla-hub.atlassian.net/browse/MNTOR-2265?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[MNTOR-2264]: https://mozilla-hub.atlassian.net/browse/MNTOR-2264?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/app/(proper_react)/redesign/(authenticated)/user/welcome/View.module.scss
```diff
@@ -176,6 +176,17 @@
         }
       }
     }
+
+    &.getStartedHidden {
+      li {
+        // If the first step indicator is hidden the space between the other two
+        // indicators grows. We’ll need to increase the length of the line that
+        // connects them:
```
Vinnl, line 184: Excellent comment :clap:

src/app/(proper_react)/redesign/(authenticated)/user/welcome/View.tsx
```diff
@@ -26,16 +27,21 @@ export type Props = {
...
   stepId?: StepId;
+  skipInitialStep?: boolean;
 };
 
 export const View = ({
   user,
   dataBrokerCount,
   breachesTotalCount,
   stepId = "getStarted",
+  skipInitialStep,
```
Vinnl, line 38: Nit, but it might be good if the caller doesn't need to remember what the initial step is (or maybe even why you'd want to skip it). So a more descriptive name, like `skipGetStarted` or `startAtEnterInfo` or something like that could be something to consider?

src/app/(proper_react)/redesign/(authenticated)/user/welcome/View.tsx
```diff
@@ -51,7 +57,9 @@ export const View = ({
         onScanStarted={() => setCurrentStep("findExposures")}
         // TODO: Add unit test when changing this code:
         /* c8 ignore next */
```
Vinnl, line 59: I think you added this test, no? :)

```suggestion
```

src/app/(proper_react)/redesign/(authenticated)/user/welcome/View.tsx
```diff
@@ -51,7 +57,9 @@ export const View = ({
...
-        onGoBack={() => setCurrentStep("getStarted")}
+        onGoBack={() =>
+          skipInitialStep ? router.back() : setCurrentStep("getStarted")
```
Vinnl, line 61: Hmm... I'm a bit wary of relying on whatever the history state is - especially since there might not be a previous location. Don't we know the page the user should be coming from, and if so, could we `router.push` that?

src/app/(proper_react)/redesign/(authenticated)/user/welcome/View.tsx
Vinnl: Actually, can't we just derive this from `stepId` being set to `"enterInfo"` in the first place? With the added benefit that that would no longer be an API in runtime code that's only used in tests :)

src/app/(proper_react)/redesign/(authenticated)/user/welcome/Onboarding.test.tsx
```diff
@@ -210,3 +212,29 @@ it("passes the axe accessibility test suite on step 3", async () => {
...
+it("shows a condensed version of the onboarding skipping step “Get started”", () => {
+  const ComposedOnboarding = composeStory(Onboarding, Meta);
+  render(<ComposedOnboarding skipInitialStep />);
+  const proceedButton = screen.getByRole("button", {
+    name: "Find exposures",
+  });
+  expect(proceedButton).toBeInTheDocument();
+});
+
+it("does not navigate back to step 1 of the onboarding", async () => {
```
Vinnl, line 225:
```suggestion
it("does not navigate back to step 1 of the onboarding when `skipInitialStep` is set", async () => {
```

Vinnl (approved):
Shame I couldn't fully try it out, but http://localhost:6060/redesign/user/welcome/free-scan looks good at least.

Vinnl:
Oh, one more thing I'm wondering: when the user completes this, should they go back to the guided experience, or back to the dashboard? (I assume the latter is what's happening now.)

src/app/(proper_react)/redesign/(authenticated)/user/welcome/View.tsx
flozia: We do not know for sure where users are coming from. We are at least linking here from the dashboard and resolution flow. Alternatively, we could hide the back button if there is no previous history state and/or also add a global hook to track the previous location client-side?

src/app/(proper_react)/redesign/(authenticated)/user/welcome/Onboarding.test.tsx
flozia: Updated in [6d87164](https://github.com/mozilla/blurts-server/pull/3541/commits/6d871645da55f045dded8c1b34133f708f94a102).

src/app/(proper_react)/redesign/(authenticated)/user/welcome/View.tsx
Vinnl: Yes, I was going to suggest that, but I didn't know of a way to check if there's no previous history state :P But if that's possible, that sounds like a good solution too.

(Though are you sure that we're linking straight to the EnterInfo page from the dashboard?)

flozia: We could check the history length, which still might result in false positives because it includes forward items as well. A global hook might be more accurate?

I’m sure about linking from the dashboard directly to the `EnterInfo` page. Checked with Product and UX and added the link [here](https://github.com/mozilla/blurts-server/pull/3541/files#diff-fd7f08410ec922928b6ba022ee3ea2b999b1a0b02c96aea074710c28a65835f9R191).

Vinnl: The main thing I'd be worried about w.r.t. the global hook is needing to remember using it everywhere, and in general it cluttering up the code of every page. It feels like something `useRouter` should be able to support it - might be worth looking in the Next.js issue tracker? Otherwise, the false positives feel like an edge case of an edge case, so is probably fine too.

flozia: Already [checked](https://github.com/vercel/next.js/discussions/36723), but no luck. Ok, sounds good!

src/app/(proper_react)/redesign/(authenticated)/user/welcome/View.tsx
flozia: I’ve added the additional prop to distinguish between setting the initial step of the “long” onboarding vs the shortened version. But since that’s just helpful for writing tests some minor changes there do the same thing. Updated in [efbeb7a](https://github.com/mozilla/blurts-server/pull/3541/commits/efbeb7a01c3378bf9831660b810a5218364665cb).

src/app/(proper_react)/redesign/(authenticated)/user/welcome/View.tsx
flozia: So, it turns out everything I’ve tried first resulted in a potentially frustrating user experience. My preference right now would be to try and keep track of the previous route within Monitor. I’ve implemented a suggestion in PR #3621.

Vinnl: Hmm, one downside of that approach is that React is going to re-render everything in that context provider (i.e. the entire app) on every path change.

One more attempt at an alternative approach: couldn't the previous page just pass a query parameter to indicate what page to link back to? e.g. `?referrer=dashboard`? And then if none is passed, we just leave out the button?

flozia: Yeah, that works — I’ll be following up with this.

flozia: Updated #3621 to use a referrer query param.

PR #3544, opened 2023-10-11
MNTOR-2241 - Refine manual remove exposure cards alignment

codemist:
# References: 
Jira: MNTOR-2241

image

### DIFF
--- a/src/app/(proper_react)/redesign/(authenticated)/user/dashboard/fix/data-broker-profiles/manual-remove/ManualRemoveView.module.scss
+++ b/src/app/(proper_react)/redesign/(authenticated)/user/dashboard/fix/data-broker-profiles/manual-remove/ManualRemoveView.module.scss
@@ -83,6 +83,7 @@
     display: flex;
     flex-direction: column;
     gap: $spacing-md;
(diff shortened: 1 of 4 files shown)


src/app/(proper_react)/redesign/(authenticated)/user/dashboard/fix/fix.module.scss
```diff
@@ -50,10 +50,12 @@
     .navArrow {
       flex: 0 0 auto;
       display: none;
+      position: absolute;
       align-items: center;
       justify-content: center;
       padding: $spacing-md;
       border-radius: $border-radius-sm;
+      right: $spacing-md;
```
Vinnl, line 58: I don't think this element needs to be absolutely positioned, since it's not overlapping anything, so since absolute positioning can be a bit of a headache, I'd like to avoid it. If we want it to be located more to the right, couldn't we just reduce the inline marging of `.fixSection`?

Vinnl (approved):
The only blocker I see is the absolute positioning of the icon, but I'm guessing that that's easy to resolve, so approving already.

PR #3588, opened 2023-10-17
Start redirecting new users to the welcome screen

Vinnl:
# References: 
Jira: MNTOR-2264
Figma: N/A

# Description

Here, "new users" are defined as those who are in the US, have not yet run a data broker scan, and did not create their Monitor account before the release of data broker scanning, as determined by the date set in the `BROKER_SCAN_RELEASE_DATE` environment variable.

# How to test

Set `BROKER_SCAN_RELEASE_DATE` to a date before you created your account, and make sure you haven't run a scan yet. If you open the dashboard, you should be redirected to the Welcome flow. But if your account is older than that date, you should just be able to look at the dashboard.

- [ ] I've added a unit test to test for potential regressions of this bug. - N/A, would involve lots of mocking


src/app/(proper_react)/redesign/(authenticated)/user/dashboard/page.tsx
```diff
@@ -34,9 +35,24 @@ export default async function DashboardPage() {
...
+      Number.parseInt(brokerScanReleaseDateParts[1] ?? "12", 10) - 1,
+      Number.parseInt(brokerScanReleaseDateParts[2] ?? "05", 10),
+    ),
+  );
   if (
     !profileId &&
-    canSubscribeToPremium({ user: session?.user, countryCode: countryCode })
+    canSubscribeToPremium({ user: session?.user, countryCode: countryCode }) &&
+    (parseIso8601Datetime(session.user.subscriber.created_at)?.getTime() ?? 0) >
+      brokerScanReleaseDate.getTime()
```
flozia, line 55: We would also need to check if a user already `hasPremium` and if they did not run a scan we are supposed to redirect as well.

flozia: Also, might want to move that check to the dashboard’s `layout.tsx` and remove the one in https://github.com/mozilla/blurts-server/blob/MNTOR-2264-welcome-new-user/src/app/(proper_react)/redesign/(authenticated)/user/dashboard/fix/layout.tsx#L24.

Vinnl: Added in 75f0d4fae93e35cf3f951d2acdbcb851cd098cfc.

As for the layout, I ended up removing that completely in 0d8f9c45c08698b5c3795f936b6db6f821e5adb1 - I don't think the redirect adds that much value there, since users shouldn't even end up in that flow in that case, and other than that the layout didn't do anything that a parent layout didn't already do. Let me know if I'm missing some obvious reason why I shouldn't have done that :)

flozia: Thanks! I don’t think it does anything we would not already have handled in higher routes.

flozia (approved):
Awesome that we have this in place now!

PR #3591, opened 2023-10-18
MNTOR-2327: Keep track of breaches emails sent in a new table

mansaj:
# References: 
Jira: MNTOR-2327
Figma: 

# Description

# Screenshot (if applicable)

Not applicable.

# How to test


Vinnl:
Just a heads-up: when adding new tables, it's nice to also add its types to `knex-tables.d.ts` so we can get code hints when writing queries: https://knexjs.org/guide/#typescript

[Here](https://github.com/mozilla/blurts-server/commit/e2b3aefa39a2eb4b370253f4224ed32fb65330a9#diff-9c021a9e29941c95a62b900bcf333055b98f1206eed8a12bbf3bd06edd566f71)'s an example.

src/db/tables/email_notifications.js
```diff
@@ -0,0 +1,115 @@
...
+ */
+async function getEmailNotification(
+  subscriberId,
+  breachId,
+  email
+){
+  console.info(
+    `getEmailNotification for subscriber: ${subscriberId}, breach: ${breachId}`,
+  );
+  const res = await knex("email_notifications")
```
rhelmer, line 33: This looks good to me overall - do we want to think about `SELECT .. FOR UPDATE` to handle the case where there may be workers running in parallel (which I think we'll need to do in order to keep up with the backlog)?

rhelmer (changes requested):
Looks great overall, just one comment about guarding against duplicates spread amongst parallel workers.

src/db/tables/email_notifications.js
mansaj: Good call.. Think we should add it for all gets? Like `getNotifiedSubscribersForBreach` below?

We aren't actually using some of these functions at the moment btw

rhelmer (approved):

src/db/tables/email_notifications.js
rhelmer: I think it would be safest, we can check the stats to see if it's causing problems but I doubt it will. If we're getting a lot of duplicates from upstream that is probably something we should follow up on anyway.

src/scripts/emailBreachAlerts.js
```diff
@@ -187,11 +193,34 @@ export async function poll(subClient, receivedMessages) {
...
+              }
+
               const emailTemplate = getTemplate(data, breachAlertEmailPartial);
               const subject = getMessage("breach-alert-subject");
 
               await sendEmail(data.recipientEmail, subject, emailTemplate);
 
+              // mark email as notified in database
+              // if this call ever fails, stop stop the script with an error
+              try{
```
pdehaan, line 218: I'm confused... It looks like Prettier is linting these files, but not sure why it wouldn't format a space after the `try` and before `{` to match L198 above.

mansaj: We are actually ignoring the `src/scripts` folder for some reason. Good catch, turned that on

PR #3596, opened 2023-10-18
MNTOR-2329/switch from mozlog to winston

rhelmer:
# References: 
Jira: MNTOR-2329

# Description

Remove `mozlog` and switch to `winston`.

# How to test

Outside of stage/prod, this should log to the console only. In GCP, it will use cloud logging and parse the payload into `jsonPayload` correctly (which can be used from stackdriver and BigQuery).


src/app/api/utils/auth.ts
```diff
@@ -172,10 +168,10 @@ export const authOptions: AuthOptions = {
   },
   events: {
     signIn(message) {
-      log.debug("fxa-confirmed-profile-data", message.user);
+      logger.debug("fxa-confirmed-profile-data", message.user);
```
mansaj, line 171: PII check?

mansaj: oh since it's debug, prob won't log to gcp

src/app/functions/server/onerep.ts
```diff
@@ -154,7 +153,7 @@ export async function getProfile(
     method: "GET",
   });
   if (!response.ok) {
-    log.info(
+    logger.info(
```
mansaj, line 156: We might want to change this to error or warning?

src/app/functions/server/onerep.ts
```diff
@@ -127,7 +126,7 @@ export async function createProfile(
     body: JSON.stringify(requestBody),
   });
   if (!response.ok) {
-    log.info(
+    logger.info(
```
mansaj, line 129: We might want to change this to error or warning?

mansaj (approved):

src/app/api/utils/auth.ts
rhelmer: Thanks... I think we're capturing debug, I'll make it log something here that we can use for diagnostics but not the whole user object.

src/app/functions/server/onerep.ts
rhelmer: Hm yeah these are definitely errors, thanks

Vinnl:
Woah, great cleanup Rob, thanks for that!

PR #3597, opened 2023-10-18
MNTOR-2374 - do not redirect from new flow to front page on error

rhelmer:
# References: 
Jira: MNTOR-2374

# Description

We would rather have folks testing get an error page and report it, than get silently redirected to the front page and maybe accidentally log into the old flow.

# How to test

1. Try to use `/redesign/user/dashboard/welcome` with/without being in allow list for free scan
2. Try to use `/redesign/user/dashboard/subscribed` with/without being subscribed


src/app/(proper_react)/redesign/(authenticated)/user/dashboard/subscribed/page.tsx
```diff
@@ -36,16 +37,21 @@ export default function Subscribed() {
...
-        router.replace(`/`);
+        console.error(ex);
+        router.replace("/redesign/user/dashboard");
       }
     }
     void updateSession();
-  }, [update, router]);
+    // Only run once, since we will be immediately redirecting no matter what.
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, []);
```
rhelmer, line 54: @Vinnl and I chatted about this one when it landed - I started with `[]` since I want this to run only once (twice in dev mode but that's OK), the problem is that `update` causes this to run over and over, when we only ever want to run this a single time. It will explicitly redirect somewhere no matter what happens (and if it somehow fails we don't want it to keep running in a loop either)

mansaj (approved):
lgtm

PR #3621, opened 2023-10-19
Previous route provider

flozia:
# References: 
GitHub: PR #3541

# Description

This is adding a context provider that keeps track of the previous routes within Monitor. For now, we used it to determine from where users entered the “welcome flow” and to know where we send them when they “go back” or finish their scan.

# How to test

Same steps as for PR #3541.


src/app/functions/server/getReferrerUrl.ts
```diff
@@ -0,0 +1,32 @@
...
+    return referrer.replace(serverUrl, "");
+  }
+
+  switch (referrerParam) {
+    case "dashboard":
+      return "/redesign/user/dashboard";
+    case "fix":
+      return "/redesign/user/dashboard/fix/data-broker-profiles/start-free-scan";
+    default:
+      return "";
```
Vinnl, line 30: Should we consider making this return `null`, and hiding the back button if it is?

Vinnl (approved):
I didn't try it out, but the overall approach looks good - and you didn't [misspell referrer](https://en.wikipedia.org/wiki/HTTP_referer#Etymology) :grin:

src/app/functions/server/getReferrerUrl.ts
flozia: I had it go to `/redesign/user/dashboard` as a fallback before. [f25e84a](https://github.com/mozilla/blurts-server/pull/3621/commits/f25e84a93ee9e5bc3eb725d64f05e8c19b5d80a3) would now hide the “Go back” button if there is no previous route. One downside: After reloading this page there would be also a referrer present and thus we would not show the button.

src/app/(proper_react)/redesign/(authenticated)/user/welcome/[[...slug]]/page.tsx
```diff
@@ -46,18 +50,22 @@ export default async function Onboarding({ params }: Props) {
...
-    </>
+    <View
+      user={session.user}
+      dataBrokerCount={parseInt(
+        process.env.NEXT_PUBLIC_ONEREP_DATA_BROKER_COUNT as string,
+        10,
+      )}
+      breachesTotalCount={allBreachesCount}
+      stepId={firstSlug === FreeScanSlug ? "enterInfo" : "getStarted"}
+      previousRoute={previousRoute || "/redesign/user/dashboard"}
```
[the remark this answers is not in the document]
Comment (flozia) [reply src/app/(proper_react)/redesign/(authenticated)/user/welcome/[[...slug]]/page.tsx]: Adjusted in https://github.com/mozilla/blurts-server/pull/3621/commits/f25e84a93ee9e5bc3eb725d64f05e8c19b5d80a3.

src/app/functions/server/getReferrerUrl.ts
Vinnl: Sorry, I'm not sure I follow - shouldn't the query param still be present, and thus the button as well, when you refresh?

flozia: Sorry, yeah: I forgot to mention I only mean the specific case where the query parameter might get stripped for some reason and it only uses the `referer` value from the header. No big deal I think — just wanted to mention it.

PR #3622, opened 2023-10-19
e2e cron for monitor application

mozrokafor:
# References: 
Jira: [MNTOR-2205](https://mozilla-hub.atlassian.net/browse/MNTOR-2205)

Adding a cron job that will run our all tests in our e2e suite. This will run against stage, allowing the team to catch any issues with the monitor application before it gets to Production.

[MNTOR-2205]: https://mozilla-hub.atlassian.net/browse/MNTOR-2205?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


.github/workflows/playwright.yml
```diff
@@ -4,6 +4,8 @@ on:
     branches: [ main ]
   pull_request:
     branches: [ main ]
+  schedule:
+    - cron: '0 8 * * *'
```
pdehaan, line 8: So every day at 8AM (presumably in whatever timezone the GitHub Actions servers are running)?

pdehaan (approved):

.github/workflows/playwright.yml
pdehaan: I think once a day is plenty. I imagine the risk of stage changing without going through several PRs (which already run the tests) is pretty low.

mansaj: So I don't think this is right.. Right now this `yml` is running e2e for smoke tests. We'd probably want to leave this alone.

The cron is supposed to be for the entire suite

mozrokafor: your're correct joe, we have another ticket that will split this up to its own yml, as we don't need to do the building etc, as we'll use stage/prod env anyway

PR #3639, opened 2023-10-23
MNTOR-2267 - Manual removal exposure cards UI tweaks

codemist:
# References: 
Jira: MNTOR-2267

# Description

The Jira ticket mainly deals with aligning the data classes in the expanded exposure card into one line, but I've gone ahead and added more UI tweaks to the exposure card as well as how they're presented in the manual remove fix view.

# Screenshot (if applicable)

image
image


src/app/components/client/ExposureCard.tsx
```diff
@@ -155,7 +159,10 @@ const ScanResultCard = (props: ScanResultCardProps) => {
             </dt>
             <dd>
               <span
-                className={`${styles.exposureCompanyTitle} ${styles.companyNameArea}`}
+                className={`${styles.exposureCompanyTitle} ${
+                  styles.companyNameArea
+                }
+                ${isCompanyNameTooLong && styles.makeFontSmaller}`}
```
flozia, line 165: If `isCompanyNameTooLong` is falsy this will return `null` so a ternary condition that returns just an empty string would be preferable.

src/app/components/client/ExposureCard.tsx
```diff
@@ -137,6 +137,10 @@ const ScanResultCard = (props: ScanResultCardProps) => {
     );
   }
 
+  const COMPANY_NAME_MAX_CHARACTER_COUNT = 20;
```
flozia, line 140: Question: Since the number of characters highly depends on the container width, would it make sense to take it into account as well?

flozia (approved):
The adjustments are looking good! For a screen width of around `1030px` the layout for the exposed items still looks a bit cluttered, but I’m not sure if there is an easy way for us to avoid that right now.

src/app/components/client/ExposureCard.tsx
flozia: Just FYI: Saw on `main` that `makeFontSmaller` seems is being applied to all viewports.

flozia: No worries — thanks for the follow-up!

PR #3643, opened 2023-10-23
MNTOR-2324 - check for subscription status more often

rhelmer:
# References: 
Jira: MNTOR-2324

# Description

During internal pre-foxfooding testing the last week or so, we've seen the subscription check fail on return from subplat because the `subscription` token hasn't updated in FxA yet, and Monitor only checks a single time.

This change causes Monitor to call the `update` from `useSession` (which fetches the latest profile info from FxA) from the "welcome to premium" page (which is where subplat returns users to), and to no longer redirect users to `/` on error at all.

This also causes the `PremiumBadge` to use the same hook. Note the use of `[]` because we only want the hook to run once, as the comments state `update` always appears to have changed so it will run in a loop and hit FxA constantly otherwise.

# How to test

This is pretty hard to test since it's just resilience against fxa/subplat taking too long to respond, it's a response to what folks are experiencing during testing. We should try to collect data on how often this happens.


rhelmer:
I tried mocking out the `useSession` and `fetch` (since the tests are complaining about this), will see what I can come up with.

src/app/components/client/PremiumBadge.tsx
```diff
@@ -56,6 +57,18 @@ export default function PremiumBadge({
...
 
+  const { update } = useSession();
+
+  useEffect(() => {
+    async function updateSession() {
+      await update();
+    }
+    void updateSession();
+
+    // This should only run once per page load - `update` will always appear to be changed.
```
flozia, line 68: We have two places where we are adding the `<PremiumBadge>` — `MobileShell` and `Toolbar`. I’m wondering if we should create a component (e.g. `<PremiumCheck>`) that is only returning a Fragment `<></>` that is doing the check and adding that to `Shell.tsx`?

flozia (approved):

src/app/components/client/PremiumBadge.tsx
rhelmer: We actually do have that already https://github.com/mozilla/blurts-server/blob/f0e17f84611c0d0a0758c54891c3facac8b644ab/src/app/components/client/SubscriptionCheck.tsx#L11 - I can change to using this approach instead.

rhelmer: Hm OK I gave this a try but test failures started showing up in other tests, and I don't know enough about the test setup to resolve this... I'm going to land as-is because I believe it improves the situation, but moving this out of `PremiumBadge` and into `Shell` directly is worth doing I believe, I'll file a follow-up issue.

rhelmer: Filed https://mozilla-hub.atlassian.net/browse/MNTOR-2396

PR #3646, opened 2023-10-24
Revert "MNTOR-2329/switch from mozlog to winston (#3596)"

rhelmer:
This reverts commit 9ac2c85450cef3e75d4b8e75d04f6d0f84e9378b.

# References: 
Jira: MNTOR-2329

# Description

This caused problems on staging:

```
TypeError: Cannot set property transports of #<Logger> which has only a getter
    at 40066 (/app/.next/server/chunks/1701.js:1:352)
    at __webpack_require__ (/app/.next/server/webpack-runtime.js:1:161)
    at 5183 (/app/.next/server/chunks/5183.js:1:139)
    at __webpack_require__ (/app/.next/server/webpack-runtime.js:1:161)
    at 4205 (/app/.next/server/chunks/5127.js:1:1956)
    at Function.__webpack_require__ (/app/.next/server/webpack-runtime.js:1:161)
    at runNextTicks (node:internal/process/task_queues:60:5)
    at listOnTimeout (node:internal/timers:540:9)
    at process.processTimers (node:internal/timers:514:7)
    at async eq (/app/node_modules/next/dist/compiled/next-server/app-page.runtime.prod.js:107:38102)
```

I'll test this in a GCP sandbox before re-pushing, backing out for now.

# Checklist (Definition of Done)
- [x] Commits in this PR are minimal and [have descriptive commit messages](https://chris.beams.io/posts/git-commit/).
- [x] Jira ticket has been updated (if needed) to match changes made during the development process.

PR #3647, opened 2023-10-24
Do not assign to read-only property

Vinnl:
I think this should resolve the issue of https://github.com/mozilla/blurts-server/pull/3646 without needing a revert.

Without this change, we see the following error on staging:

    TypeError: Cannot set property transports of #<Logger> which has only a getter
        at 40066 (/app/.next/server/chunks/1701.js:1:352)
        at __webpack_require__ (/app/.next/server/webpack-runtime.js:1:161)
        at 5183 (/app/.next/server/chunks/5183.js:1:139)
        at __webpack_require__ (/app/.next/server/webpack-runtime.js:1:161)
        at 4205 (/app/.next/server/chunks/5127.js:1:1956)
        at Function.__webpack_require__ (/app/.next/server/webpack-runtime.js:1:161)
        at runNextTicks (node:internal/process/task_queues:60:5)
        at listOnTimeout (node:internal/timers:540:9)
        at process.processTimers (node:internal/timers:514:7)
        at async eq (/app/node_modules/next/dist/compiled/next-server/app-page.runtime.prod.js:107:38102)

### DIFF
--- a/src/app/functions/server/logging.ts
+++ b/src/app/functions/server/logging.ts
@@ -14,10 +14,8 @@ const loggingWinston = new LoggingWinston({
 
 export const logger = createLogger({
   level: "info",
-  transports: [new transports.Console()],
+  // In GCP environments, use cloud logging instead of stdout.
+  transports: ["stage", "production"].includes(process.env.APP_ENV ?? "local")
+    ? [loggingWinston]
+    : [new transports.Console()],
 });
-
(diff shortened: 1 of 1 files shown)


rhelmer:
Thanks @Vinnl - I'll test this in the GCP sandbox, my initial reaction was to revert so we don't leave `main` broken but happy to fix instead if it's simple enough.

rhelmer (approved):

PR #3648, opened 2023-10-24
Fix: Add session provider to Storybook

flozia:
# References: 
PR #3643

# Description

This adds a session provider to Storybook which should also fix the unit test failures.

### DIFF
--- a/.storybook/preview.tsx
+++ b/.storybook/preview.tsx
@@ -8,6 +8,7 @@ import type { Preview } from "@storybook/react";
 import { action } from "@storybook/addon-actions";
 import { linkTo } from "@storybook/addon-links";
 import "../src/app/globals.css";
+import { SessionProvider } from "../src/contextProviders/session";
 import { L10nProvider } from "../src/contextProviders/localization";
 import { metropolis } from "../src/app/fonts/Metropolis/metropolis";
 import { ReactAriaI18nProvider } from "../src/contextProviders/react-aria";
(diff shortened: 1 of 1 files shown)


.storybook/preview.tsx
```diff
@@ -16,7 +17,7 @@ import { getEnL10nBundlesSync } from "../src/app/functions/server/mockL10n";
 const inter = Inter({ subsets: ["latin"], variable: "--font-inter" });
 
 const AppDecorator: Exclude<Preview["decorators"], undefined>[0] = (
-  storyFn
+  storyFn,
```
flozia, line 20: The added trailing spaces are being added by Prettier due to recent changes in our linting rules.

.storybook/preview.tsx
```diff
@@ -32,7 +33,9 @@ const AppDecorator: Exclude<Preview["decorators"], undefined>[0] = (
 
   return (
     <L10nProvider bundleSources={l10nBundles}>
-      <ReactAriaI18nProvider locale="en">{storyFn()}</ReactAriaI18nProvider>
+      <SessionProvider session={null}>
```
flozia, line 36: Just setting the session to `null` should do here. In places where we mock session and/or user data, we do that in the respective Stories — like here on the Dashboard: https://github.com/mozilla/blurts-server/blob/MNTOR-2324/storybook-session-provider/src/app/(proper_react)/redesign/(authenticated)/user/dashboard/Dashboard.stories.tsx#L152.

PR #3654, opened 2023-10-24
Move subscription check to Shell

flozia:
# References: 
Jira: [MNTOR-2396](https://mozilla-hub.atlassian.net/browse/MNTOR-2396)

# Description

Moves the subscription check from `<PremiumBadge>` to `Shell`.

[MNTOR-2396]: https://mozilla-hub.atlassian.net/browse/MNTOR-2396?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/app/components/client/SubscriptionCheck.tsx
```diff
@@ -4,22 +4,21 @@
...
     async function updateSession() {
-      const session = await getSession();
-      if (!hasPremium(session?.user)) {
-        await update();
-      }
+      await update();
     }
 
-    void updateSession();
+    if (!hasPremium(session?.user)) {
```
flozia, line 19: In `<PremiumBadge>` we did not check if the user already has Premium and did the check either way. Is this something we’d like to keep doing?

src/app/components/client/SubscriptionCheck.tsx
```diff
@@ -4,22 +4,21 @@
...
 "use client";
 
-import { getSession, useSession } from "next-auth/react";
+import { useSession } from "next-auth/react";
 import { useEffect } from "react";
 import { hasPremium } from "../../functions/universal/user";
 
 export const SubscriptionCheck = () => {
-  const { update } = useSession();
+  const { data: session, update } = useSession();
```
flozia, line 12: We can just get the session from `useSession` instead of fetching it via `getSession`.

src/app/components/client/SubscriptionCheck.tsx
rhelmer: No actually now that I think about it I think the `hasPremium` check is explicitly not useful... checking this more often serves two purposes:

1. ensure that we give subscribed users access to features for which they are entitled on all devices as soon as possible
2. remove premium features as soon as cancellation has taken place

src/app/components/client/SubscriptionCheck.tsx
```diff
@@ -4,26 +4,19 @@
...
   useEffect(() => {
     async function updateSession() {
-      const session = await getSession();
-      if (!hasPremium(session?.user)) {
-        await update();
-      }
+      await update();
     }
 
     void updateSession();
```
Vinnl, line 18: Given that we're not actually waiting for the Promise to resolve before doing anything, couldn't we just do:

```suggestion
    update();
```

And also I'm assuming you consciously no longer used the empty dependencies array?

Vinnl (approved):
LGTM!

src/app/components/client/SubscriptionCheck.tsx
flozia: Yes, but I’ve removed it again in  [8addbcd](https://github.com/mozilla/blurts-server/pull/3654/commits/8addbcda5883bad1899f2af29000ea7fc662f3b6).

flozia: Regarding awaiting the session update: No, we do not have to wait for it — updated in [1d60a84](https://github.com/mozilla/blurts-server/pull/3654/commits/1d60a844e4568bf0c620d9430ae78d6652a65278).

PR #3666, opened 2023-10-27
MNTOR-2423 - captureMessage has a length limit, move logging to server

rhelmer:
# References: 
Jira: MNTOR-2423

# Description

~~Sentry's captureMessage has a length limit, move logging to server - I didn't do this initially because this function appears to be both used by client and server code, so I am guarding against server code by putting the import and function call to the server logger behind a check for the `window` global.~~ EDIT: storybook-build is having trouble with this, instead of worrying about it we can just use `console.error` for now, which will work in both GCP (as a stdout textPayload) or client contexts.

### DIFF
--- a/src/app/(proper_react)/redesign/(authenticated)/user/dashboard/getUserDashboardState.tsx
+++ b/src/app/(proper_react)/redesign/(authenticated)/user/dashboard/getUserDashboardState.tsx
@@ -4,7 +4,7 @@
 
 import { ContentProps } from "./DashboardTopBanner/DashboardTopBannerContent";
 import { isGuidedResolutionInProgress } from "../../../../../functions/server/getRelevantGuidedSteps";
-import { captureMessage } from "@sentry/nextjs";
+import { captureException } from "@sentry/nextjs";
 
 export type UserDashboardState =
   | "NonEligiblePremiumUserNoBreaches"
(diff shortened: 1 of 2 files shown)


src/app/(proper_react)/redesign/(authenticated)/user/dashboard/getUserDashboardState.tsx
```diff
@@ -507,7 +507,16 @@ export const getUserDashboardState = (
     return "UsUserScanInProgressResolvedBreaches";
   }
 
-  captureMessage(`InvalidUserState: ${JSON.stringify(contentProps)}`);
+  /* c8 ignore next 9 */
+  if (typeof window === "undefined") {
```
Vinnl, line 511: Possibly the `server-only` package is of interest here: https://nextjs.org/docs/app/building-your-application/rendering/composition-patterns#keeping-server-only-code-out-of-the-client-environment

Vinnl (approved):

mansaj (approved):

mansaj:
Interesting.. Any idea how to make storybook check a part of the CICD? @Vinnl

src/app/(proper_react)/redesign/(authenticated)/user/dashboard/getUserDashboardState.tsx
rhelmer: Yes that is thanks! I'll file a follow to make `logging.ts` work in both contexts, we can use that here. I'd like to make this easier for next time.

PR #3667, opened 2023-10-27
Disable next-auth debug logging

rhelmer:
# Description

This was useful for setting up, but this is more info than we need in stage/prod.


mansaj (approved):
Should we make it conditional per environment?

rhelmer:
Thanks @mansaj yeah that's not a bad idea... we could leave it on for local development?

PR #3686, opened 2023-11-02
MNTOR-2444 - handle scan results updates correctly

rhelmer:
# References: 
Jira: MNTOR-2444

# Description

More detail in https://mozilla-hub.atlassian.net/browse/MNTOR-2444?focusedCommentId=789660 - the short version is that we were not handling new scan IDs correctly, *and* scan results are not attached to scan IDs in the way we thought they would be.

I tried to simplify and reuse the logic in `refreshStoredScanResults` while I was at it.

# How to test

This is unfortunately very difficult to test, as it relies on the behavior of the production scan service, and requires subscribing (which can't be undone). However, I believe it should fix the existing accounts that are stuck with fewer results than they should be showing (one of these is mine).

We don't really have automated testing around this behavior (outside of mocks), but I'll see if I can add it.

### DIFF
--- a/src/app/functions/server/refreshStoredScanResults.ts
+++ b/src/app/functions/server/refreshStoredScanResults.ts
@@ -23,14 +23,14 @@ export async function refreshStoredScanResults(profileId: number) {
     const remoteScans = (await listScans(profileId)).data;
     const localScans = await getAllScansForProfile(profileId);
 
-    const newScan = remoteScans.filter(
+    const newScans = remoteScans.filter(
       (remoteScan) =>
         !localScans.some(
           (localScan) => localScan.onerep_scan_id === remoteScan.id,
         ),
     );
 
-    newScan.forEach((scan) =>
+    newScans.forEach((scan) =>
       logger.info("scan_created_or_updated", {
         onerepScanId: scan.id,
         onerepScanStatus: scan.status,
--- a/src/app/functions/server/refreshStoredScanResults.ts
+++ b/src/app/functions/server/refreshStoredScanResults.ts
@@ -23,16 +23,24 @@ export async function refreshStoredScanResults(profileId: number) {
     const remoteScans = (await listScans(profileId)).data;
     const localScans = await getAllScansForProfile(profileId);
 
-    const newScans = remoteScans.filter(
+    const newScan = remoteScans.filter(
       (remoteScan) =>
         !localScans.some(
           (localScan) => localScan.onerep_scan_id === remoteScan.id,
         ),
     );
 
+    newScan.forEach((scan) =>
+      logger.info("scan_created_or_updated", {
+        onerepScanId: scan.id,
+        onerepScanStatus: scan.status,
+        onerepScanReason: scan.reason,
+      }),
+    );
+
     // Record any new scans.
     await Promise.all(
-      newScans.map(async (scan) => {
+      remoteScans.map(async (scan) => {
         await setOnerepScan(profileId, scan.id, scan.status, scan.reason);
       }),
     );
--- a/src/app/functions/server/refreshStoredScanResults.ts
+++ b/src/app/functions/server/refreshStoredScanResults.ts
@@ -38,7 +38,7 @@ export async function refreshStoredScanResults(profileId: number) {
       }),
     );
 
-    // Record any new scans.
+    // Record any new scans, or change in existing scan status.
     await Promise.all(
       remoteScans.map(async (scan) => {
         await setOnerepScan(profileId, scan.id, scan.status, scan.reason);
--- a/src/app/functions/server/refreshStoredScanResults.ts
+++ b/src/app/functions/server/refreshStoredScanResults.ts
@@ -25,7 +25,6 @@ export async function refreshStoredScanResults(profileId: number) {
 
     const newScans = remoteScans.filter(
       (remoteScan) =>
-        // @ts-ignore FIXME
         !localScans.some(
           (localScan) => localScan.onerep_scan_id === remoteScan.id,
         ),
@@ -37,17 +36,9 @@ export async function refreshStoredScanResults(profileId: number) {
     }
 
     // Refresh results for all scans, new and existing.
+    // The database will ignore any attempt to insert duplicate scan result IDs.
     const allScanResults = await getAllScanResults(profileId);
     await addOnerepScanResults(profileId, allScanResults);
-
-    /*
-
-
-    const newScanResults = remoteScanResults.filter((remoteScanResult) => {
-    });
-
-    await addOnerepScanResults(profileId, newScanResults);
-    */
   } catch (ex) {
     logger.warn("Could not fetch current OneRep results:", ex);
   }
--- a/src/app/functions/server/refreshStoredScanResults.ts
+++ b/src/app/functions/server/refreshStoredScanResults.ts
@@ -4,9 +4,10 @@
 
 import {
   addOnerepScanResults,
-  getLatestOnerepScanResults,
+  getAllScansForProfile,
+  setOnerepScan,
 } from "../../../db/tables/onerep_scans";
-import { ScanResult, getAllScanResults, getScanDetails } from "./onerep";
+import { getAllScanResults, listScans } from "./onerep";
 import { logger } from "./logging";
 
 /**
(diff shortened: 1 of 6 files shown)


src/db/tables/onerep_scans.ts
```diff
@@ -69,47 +92,9 @@ async function setOnerepManualScan(
 
 async function addOnerepScanResults(
   onerepProfileId: number,
-  onerepScanId: number,
   onerepScanResults: Array<ScanResult>,
-  onerepScanReason: Scan["reason"],
-  onerepScanStatus: Scan["status"],
 ) {
   await knex.transaction(async (transaction) => {
```
Vinnl, line 97: I think this no longer needs to be in a transaction, since it's just a single query?

src/app/functions/server/refreshStoredScanResults.ts
```diff
@@ -18,59 +19,26 @@ import { logger } from "./logging";
...
+    const newScans = remoteScans.filter(
+      (remoteScan) =>
+        !localScans.some(
+          (localScan) => localScan.onerep_scan_id === remoteScan.id,
+        ),
+    );
+
+    // Record any new scans.
+    for (const scan of newScans) {
+      await setOnerepScan(profileId, scan.id, scan.status, scan.reason);
```
Vinnl, line 35: As I understand it, this inserts the scans sequentially (i.e. `await`ing the previous insert before starting the next one). Might be better to run the inserts in parallel:

```suggestion
    // Record any new scans.
    await Promise.all(newScans.map(scan => {
      await setOnerepScan(profileId, scan.id, scan.status, scan.reason);
    }));
```

Vinnl (approved):
Let's assume that this is indeed what's needed for now, but we'll really have to test this carefully with real data, and verify with OR that this is indeed how it's supposed. I'm quite nervous about it.

src/db/tables/onerep_scans.ts
```diff
@@ -13,26 +13,48 @@ import { OnerepScanResultRow, OnerepScanRow } from "knex/types/tables";
...
   const results =
-    typeof scan === "undefined"
+    typeof scanIds === "undefined"
       ? []
       : await knex("onerep_scan_results")
           .select()
-          .where("onerep_scan_id", scan.onerep_scan_id);
+          .whereIn("onerep_scan_id", scanIds);
   return {
-    scan: scan ?? null,
```
flozia, line 35: There are a couple of places where we rely on the scan returned by this function. We’ll still need a way to get the most recent scan to determine the correct status within the app — for example:

- https://github.com/mozilla/blurts-server/blob/main/src/app/(proper_react)/redesign/(authenticated)/user/dashboard/View.tsx#L85
- https://github.com/mozilla/blurts-server/blob/main/src/app/functions/server/getRelevantGuidedSteps.ts#L170

There are also some stories that need to be updated to conform to the new structure:
- https://github.com/mozilla/blurts-server/blob/main/src/app/(proper_react)/redesign/(authenticated)/user/dashboard/Dashboard.stories.tsx#L118
- https://github.com/mozilla/blurts-server/blob/main/src/app/(proper_react)/redesign/(authenticated)/user/dashboard/fix/data-broker-profiles/view-data-brokers/ViewDataBrokers.stories.tsx#L62

rhelmer: Thanks - I couldn't find where that was actually used from a quick look.

src/db/tables/onerep_scans.ts
```diff
@@ -13,26 +13,48 @@ import { OnerepScanResultRow, OnerepScanRow } from "knex/types/tables";
...
+    .select("onerep_scan_id")
+    .from("onerep_scans")
     .where("onerep_profile_id", onerepProfileId)
     .orderBy("created_at", "desc");
 
+  const scanIds = result.map((scan_id) => scan_id["onerep_scan_id"]);
+
   const results =
-    typeof scan === "undefined"
+    typeof scanIds === "undefined"
```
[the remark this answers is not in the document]
rhelmer: I just simplified this ... yes it could be done with a join instead, I'm not sure it would be much more efficient (it wouldn't require reading into the app but we wouldn't expect the result set of scan IDs to be very large). Whether it's more readable is a separate question :) I'll try doing it with a `join`.

src/db/tables/onerep_scans.ts
```diff
@@ -13,26 +13,48 @@ import { OnerepScanResultRow, OnerepScanRow } from "knex/types/tables";
...
       ? []
       : await knex("onerep_scan_results")
           .select()
-          .where("onerep_scan_id", scan.onerep_scan_id);
+          .whereIn("onerep_scan_id", scanIds);
   return {
-    scan: scan ?? null,
     results: results,
   };
```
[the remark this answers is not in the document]
rhelmer: As @flozia points out, some parts of the code do expect to get the latest scan result (so they can inspect the `status` for instance), I've added this back.

src/db/tables/onerep_scans.ts
rhelmer: Good point, it shouldn't be needed now.

src/app/functions/server/refreshStoredScanResults.ts
rhelmer: sgtm, committed with a small change:
```
    // Record any new scans.
    await Promise.all(
      newScans.map(async (scan) => {
        await setOnerepScan(profileId, scan.id, scan.status, scan.reason);
      }),
    );
```

src/db/tables/onerep_scans.ts
```diff
@@ -69,84 +97,45 @@ async function setOnerepManualScan(
...
-        onerep_scan_id: onerepScanId,
-        onerep_scan_reason: onerepScanReason,
-        onerep_scan_status: onerepScanStatus,
-        // @ts-ignore knex.fn.now() results in it being set to a date,
-        // even if it's not typed as a JS date object:
-        created_at: knex.fn.now(),
-      })
-      .onConflict("onerep_scan_id")
-      .merge({
-        onerep_scan_status: onerepScanStatus,
```
flozia, line 109: When a scan is finished and we are setting the scan results we should still update the the latest scan status. Otherwise, we are reverting the change introduced by https://github.com/mozilla/blurts-server/pull/3470 — which we rely on for showing the “scan in progress” dashboard state.

rhelmer: Thanks - will add that rn.

rhelmer: OK I did this in a different place, `setOnerepScan`, which I think is clearer than doing it as part of `addOnerepScanResults` ... now we're back in the "set" vs. "add" sort of discussion :) I think we should revisit soon whether we want to just make this an append operation.

src/app/functions/server/refreshStoredScanResults.ts
```diff
@@ -23,16 +23,24 @@ export async function refreshStoredScanResults(profileId: number) {
...
 
-    const newScans = remoteScans.filter(
+    const newScan = remoteScans.filter(
       (remoteScan) =>
         !localScans.some(
           (localScan) => localScan.onerep_scan_id === remoteScan.id,
         ),
     );
 
+    newScan.forEach((scan) =>
```
flozia, line 33: Do we expect `newScan` to ever contain more than one item? The renaming of the variable `newScans` -> `newScan` indicates we do not?

src/app/functions/server/refreshStoredScanResults.ts
```diff
@@ -38,7 +38,7 @@ export async function refreshStoredScanResults(profileId: number) {
       }),
     );
 
-    // Record any new scans.
+    // Record any new scans, or change in existing scan status.
```
flozia, line 41: Thanks for updating the comment! I was just typing a question about it while the [commit](https://github.com/mozilla/blurts-server/pull/3686/commits/2c8f4116889bf186c7917a9a19abb94f3fb44409) came in. ⚡

flozia (approved):
Thanks @rhelmer for reworking how the scan result updates should work!

There are no potential issues that I can anticipate anymore with the parts of the application that I’m familiar with.

src/app/functions/server/refreshStoredScanResults.ts
rhelmer: Oh it might, sorry that was just at typo. I'll make it plural `newScans` just to make it self-documenting

PR #3693, opened 2023-11-03
MNTOR-2451: log getAllEmailsAndBreaches invalid inputs

mansaj:
# References: 
Jira: MNTOR-2451

# Description
Attempt to capture / log errors from invalid inputs for `getAllEmailsAndBreaches`


src/utils/breaches.js
```diff
@@ -2,6 +2,7 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
+
```
pdehaan, line 5: ✂️

src/utils/breaches.js
```diff
@@ -24,10 +25,25 @@ import { filterBreachDataTypes } from './breachResolution.js'
  * @param {*} allBreaches
  * @returns {Promise<AllEmailsAndBreaches>}
  */
-async function getAllEmailsAndBreaches (user, allBreaches) {
-  const monitoredEmails = await getUserEmails(user.id)
+async function getAllEmailsAndBreaches(user, allBreaches) {
+  
+  // @ts-ignore: function will be deprecated
   const verifiedEmails = []
+  // @ts-ignore: function will be deprecated
```
pdehaan, line 32: Is this `@ts-ignore` comment saying the `getallEmailsAndBreaches()` function will be deprecated?
Also, curious why we need these two ignore statements (plus L37 and L42).

src/utils/breaches.js
```diff
@@ -24,10 +25,25 @@ import { filterBreachDataTypes } from './breachResolution.js'
...
   const unverifiedEmails = []
+  
+  if (!user) {
+    console.error("subscriber cannot be undefined");
+    // @ts-ignore: function will be deprecated
+    return { verifiedEmails, unverifiedEmails };
+  }
+  if (!allBreaches || allBreaches.length === 0) {
+    console.error("all breaches object cannot be empty");
```
pdehaan, line 41: Not sure if this is better or clearer, but now matches the argument name.
```suggestion
    console.error("allBreaches object cannot be empty");
```

rhelmer: Could you do Sentry `captureError()` here as well? Then it'll show up when we're doing triage.

src/db/tables/emailAddresses.js
```diff
@@ -69,6 +69,7 @@ async function getEmailAddressRecordByEmail (email) {
 // Not covered by tests; mostly side-effects. See test-coverage.md#mock-heavy
 /* c8 ignore start */
 async function addSubscriberUnverifiedEmailHash (user, email) {
+  const lowerCaseEmail = email.toLowerCase()
```
rhelmer, line 72: This will help in the future, but we still have email addresses in the DB that are not lower case that won't work right now correct?

Should we migrate these? Might be a big migration... could we file a separate ticket to investigate this? We'd have to recalculat  their sha1 hashes (I am pretty sure HIBP lower-cases so they aren't matching anything anyway).

rhelmer (approved):
This looks good to me overall but wondering what to do about existing email addresses with this problem. We should also confirm that HIBP is lower-casing before hashing (I think I recall that they do this but we should double-check).

src/db/tables/emailAddresses.js
mansaj: @rhelmer thanks! Yes. I'm thinking we should just use a one-off script to migrate all the existing records. I will get a ticket going

PR #3704, opened 2023-11-07
WIP: Use manual resolution in stats calculation

Vinnl:
I thought this was going to be a quick fix, but I'm finding it very hard to follow what is going on in the touched code — so I'm sharing this as in-progress work rather than a solution.

The problem was that manually resolving a data broker scan result would not result in the stats at the top of the dashboard being updated. Additionally, I saw the number of fixed items on the "Fixed" tab show 0, even though multiple exposures were listed there.

What was suspicious, is that the `sanitizeExposures` functions appeared to iterate over `sanitizedExposures` to create `other`, and then added all items in there *to that same* object.

However, with my changes, I still see a couple of suspicious things:

- The number of items "Resolved by you" on the fixed tab doesn't match the number in the doughnut chart.
- Likewise, the number of exposures mentioned as left to fix in the top banner content on the "Action needed" tab doesn't match the number in the doughnut chart.
- There are *a lot* of entries in the "Other" category now.

All of which is to say: this probably needs more work. Sharing this PR so I can hopefully hand over this ticket.

### DIFF
--- a/src/app/components/client/Chart.tsx
+++ b/src/app/components/client/Chart.tsx
@@ -34,13 +34,13 @@ export const DoughnutChart = (props: Props) => {
     { type: "dialog" },
     explainerDialogState,
   );
-  const sumOfFixedExposures = props.data.reduce(
+  const sumOfShownExposures = props.data.reduce(
     (total, [_label, num]) => total + num,
     0,
   );
 
   const percentages = props.data.map(([label, num]) => {
-    return [label, num / sumOfFixedExposures] as const;
+    return [label, num / sumOfShownExposures] as const;
   });
 
   const diameter = 100;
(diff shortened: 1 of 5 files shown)


src/app/components/client/Chart.tsx
```diff
@@ -34,13 +34,13 @@ export const DoughnutChart = (props: Props) => {
...
   );
-  const sumOfFixedExposures = props.data.reduce(
+  const sumOfShownExposures = props.data.reduce(
     (total, [_label, num]) => total + num,
     0,
   );
 
   const percentages = props.data.map(([label, num]) => {
-    return [label, num / sumOfFixedExposures] as const;
+    return [label, num / sumOfShownExposures] as const;
```
pdehaan, line 43: Can `sumOfShownExposures` ever be zero and cause this to implode?

pdehaan (approved):

PR #3705, opened 2023-11-07
Retry OneRep profile activation if needed

flozia:
# References: 
Jira: [MNTOR-2393](https://mozilla-hub.atlassian.net/browse/MNTOR-2393)

# Description

Retry OneRep profile activation if needed after subscribing or running a scan if needed.

[MNTOR-2393]: https://mozilla-hub.atlassian.net/browse/MNTOR-2393?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/app/(proper_react)/redesign/(authenticated)/user/dashboard/page.tsx
```diff
@@ -72,6 +74,14 @@ export default async function DashboardPage() {
 
   await refreshStoredScanResults(profileId);
 
+  // If the current user is a subscriber and their OneRep profile is not
+  // activated: Most likely we were not able or failed to kick-off the
```
rhelmer, line 78: How are we detecting if the user is activated or not now? It should be in the profile so we could check that with an API call, but how do we know if `opt-out` didn't work? The lack of `initial` scan would work I think... I don't see where the code is doing any checks like this though?

rhelmer (changes requested):
lgtm and I think makes sense, but question on whether it is actually checking to see if user is already activated and opted-out

src/app/(proper_react)/redesign/(authenticated)/user/dashboard/page.tsx
flozia: I’ve added a check for the profile status here: https://github.com/mozilla/blurts-server/pull/3705/files#diff-88d1915de743c76564f89f25f640df3935f835fc8db3609ae1763c893a1dfdf1R223-R224.

rhelmer (approved):

src/app/(proper_react)/redesign/(authenticated)/user/dashboard/fix/data-broker-profiles/welcome-to-premium/page.tsx
```diff
@@ -35,6 +36,12 @@ export default async function WelcomeToPremiumPage() {
     user: session.user,
   };
 
+  // If the current user is a subscriber and their OneRep profile is not
+  // activated: Most likely we were not able or failed to kick-off the
+  // auto-removal process.
+  // Let’s make sure the users OneRep profile is activated:
+  await activateAndOptoutProfile(profileId);
```
pdehaan, line 43: Does this need to be wrapped in a `try..catch` block?
IIUC, the `activeAndOptoutProfile()` fn has this:
```js
  if (profileStatus === "inactive") {
    await activateProfile(profileId);
    await optoutProfile(profileId);
  } else {
```

and both the `activateProfile()` and `optoutProfile()` functions will potentially throw errors.

mansaj (approved):

src/app/(proper_react)/redesign/(authenticated)/user/dashboard/page.tsx
flozia: Also added the check to only `opt-out` a profile when there is no existing `initial` scan in [24f3b68](https://github.com/mozilla/blurts-server/pull/3705/commits/24f3b6848a0d03540baa79f72b09bb7187056cbc).

src/app/(proper_react)/redesign/(authenticated)/user/dashboard/fix/data-broker-profiles/welcome-to-premium/page.tsx
flozia: Good point! Added in [24f3b68](https://github.com/mozilla/blurts-server/pull/3705/commits/24f3b6848a0d03540baa79f72b09bb7187056cbc).

flozia:
@rhelmer @mansaj After some additional feedback on the PR I’ve added a check to only attempt to `opt-out` of the profile if there is no existing initial scan and other small changes in https://github.com/mozilla/blurts-server/pull/3705/commits/24f3b6848a0d03540baa79f72b09bb7187056cbc.

src/app/functions/server/onerep.ts
```diff
@@ -216,6 +219,28 @@ export async function optoutProfile(profileId: number): Promise<void> {
     );
   }
 }
+
+export async function activateAndOptoutProfile(
+  profileId: number,
+): Promise<void> {
+  try {
+    const { status: profileStatus } = await getProfile(profileId);
```
Vinnl, line 227: Hmm, so this means we're doing a call to OneRep every time we load the dashboard for Premium users. Is OneRep OK with that? And isn't this going to delay our serving the dashboard quite a bit?

flozia: That’s a good point. We might need to consider either adding the profile status to our `onerep_profiles` table or I can also just move the `hasInitialScan` check up and return early.

flozia: I think [ff91bc7](https://github.com/mozilla/blurts-server/pull/3705/commits/ff91bc7bb832edc5449bd8ea15e346115aa54998) should already avoid the repeated calls to OneRep.

Vinnl (approved):
Might be good to also wait for Rob's approval (I'm assuming the multiple DB queries shouldn't be a problem?), but it looks good to me!

src/app/(proper_react)/redesign/(authenticated)/user/dashboard/fix/data-broker-profiles/welcome-to-premium/page.tsx
```diff
@@ -13,11 +13,12 @@ import { WelcomeToPremiumView } from "./WelcomeToPremiumView";
 import { getSubscriberEmails } from "../../../../../../../../functions/server/getSubscriberEmails";
 import { StepDeterminationData } from "../../../../../../../../functions/server/getRelevantGuidedSteps";
 import { getCountryCode } from "../../../../../../../../functions/server/getCountryCode";
+import { activateAndOptoutProfile } from "../../../../../../../../functions/server/onerep";
 
 export default async function WelcomeToPremiumPage() {
   const session = await getServerSession(authOptions);
 
-  // Ensure user is logged in
+  // Ensure user is logged in and a subscriber
```
rhelmer, line 21: So the term `subscriber` pre-dates the premium feature, I'm worried this is going to cause confusion... if they are logged in then they get a `subscriber.id`, not the fault of your comment but I'm worried that someone might think this means the user subscribed to premium.

rhelmer (approved):
Thanks for taking care of that check to see if the user is inactive before trying to activate!

PR #3706, opened 2023-11-07
MNTOR-2400/add delete to users api

rhelmer:
# References: 
Jira: MNTOR-2400

# Description

This adds the ability to delete via the `admin/users` API, in addition to list and force subscribe/unsubscribe.

Right now this doesn't attempt to share code with the FxA/Subplat webhook handler, but I think it should - we might need to change this as we build out new features (like telemetry which will need to get a deletion request) and as we modify existing features.

I had imagined not making delete so granular, but it's really useful to be able to delete `scan`s and `scan_result`s for a user (which forces them to refresh), which is why I did it this way.

# How to test

See JSDoc:

```
 * Multiple actions may be specified in one request:
 * {
 *   "actions":[
 *     "subscribe",
 *     "unsubscribe",
 *     "delete_onerep_profile",
 *     "delete_onerep_scans",
 *     "delete_onerep_scan_results",
 *     "delete_subscriber"
 *   ]
 * }
 * 
 * Actions will be processed in the order specified.
```

Requires the sha1 hash for the primary email of the subscriber, and a `PUT` by an admin user.


src/app/api/v1/admin/users/[primarySha1]/route.ts
```diff
@@ -109,28 +125,62 @@ export async function PUT(
...
+              await deleteSubscriber(subscriber);
+              logger.info("delete_subscriber", {
+                onerepProfileId,
+                primarySha1,
+              });
+              break;
+            }
+            default: {
+              logger.error("Unknown action:", action);
+              return NextResponse.json({ success: true }, { status: 500 });
```
Comment (mansaj) [inline src/app/api/v1/admin/users/[primarySha1]/route.ts:182]: success true?

mansaj (approved):

src/app/api/v1/admin/users/[primarySha1]/route.ts
Comment (rhelmer) [reply src/app/api/v1/admin/users/[primarySha1]/route.ts]: Yes that's not very useful... I saw another place this was true as well.

PR #3710, opened 2023-11-09
Remove session update from PremiumButton

flozia:
# References: 
Jira: [MNTOR-2396](https://mozilla-hub.atlassian.net/browse/MNTOR-2396)

# Description

Removes the redundant session update from `PremiumBadge` and moves the session update to `Shell` to get the latest subscription status. This ensures that we check the status more consistently and remove the duplicate check from `PremiumBadge`: It is being rendered twice — once in [Toolbar](https://github.com/mozilla/blurts-server/blob/premium-button-remove-session-update/src/app/components/client/toolbar/Toolbar.tsx#L27) and also in [MobileShell](https://github.com/mozilla/blurts-server/blob/premium-button-remove-session-update/src/app/(proper_react)/redesign/MobileShell.tsx#L106).

[MNTOR-2396]: https://mozilla-hub.atlassian.net/browse/MNTOR-2396?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/app/(proper_react)/redesign/Shell.tsx
+++ b/src/app/(proper_react)/redesign/Shell.tsx
@@ -14,6 +14,7 @@ import { PageLink } from "./PageLink";
 import { ExtendedReactLocalization } from "../../hooks/l10n";
 import { GaScript } from "./GaScript";
 import getPremiumSubscriptionUrl from "../../functions/server/getPremiumSubscriptionUrl";
+import { SubscriptionCheck } from "../../components/client/SubscriptionCheck";
 
 export type Props = {
   l10n: ExtendedReactLocalization;
@@ -33,6 +34,7 @@ export const Shell = (props: Props) => {
   return (
     <>
       <GaScript nonce={props.nonce} />
+      <SubscriptionCheck />
       <MobileShell
         session={props.session}
         monthlySubscriptionUrl={monthlySubscriptionUrl}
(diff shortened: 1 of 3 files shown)


src/app/components/client/PremiumBadge.tsx
```diff
@@ -64,29 +63,14 @@ const PremiumLayout = (props: Props) => {
       />
     </>
   );
-};
-
-export function PremiumButton(props: Props) {
-  return <PremiumLayout {...props} />;
```
flozia, line 70: @codemist It does not matter too much, but since I’m touching this file either way: As mentioned [here](https://github.com/mozilla/blurts-server/pull/3699#discussion_r1386504849) this wrapper function is redundant since it does not handle any additional logic or functionality.

src/app/(proper_react)/redesign/Shell.tsx
```diff
@@ -33,6 +34,7 @@ export const Shell = (props: Props) => {
   return (
     <>
       <GaScript nonce={props.nonce} />
+      <SubscriptionCheck />
```
flozia, line 37: This change was supposed to already be in `main`. I assume that it might have been removed while resolving merge conflicts or I’ve never pushed it up.

PR #3711, opened 2023-11-09
Make the doughnut chart edges sharper

Vinnl:
By increasing the size of the SVG, there is more detail per pixel, and thus smoother edges.


Vinnl:
Oh hmm, locally I saw a difference, but comparing this preview deployment with the one in main, I don't anymore :/

PR #3719, opened 2023-11-10
Remove dynamic NEXT_PUBLIC_* environment variables

Vinnl:
# References: 
Jira: MNTOR-2434 and MNTOR-2145
Figma: N/A

# Description

It's fine to use NEXT_PUBLIC_* environment variables that are defined in `.env-dist` and are the same in every environment. However, since they are inserted at build time, they cannot be dynamically inserted at runtime, and trying to do so will instead inject the value they have at build time.

I'll also submit a PR to remove those values from the prod env.

### DIFF
--- a/locales/pt-PT/app.ftl
+++ b/locales/pt-PT/app.ftl
@@ -708,6 +708,8 @@ ad-unit-6-before-you-complete = Antes de concluir a próxima inscrição, utiliz
 -brand-firefox = Firefox
 -brand-fx-monitor = Firefox Monitor
 -brand-mozilla = Mozilla
+-brand-premium = Premium
+-brand-monitor-premium = Monitor Premium
 -brand-mozilla-foundation = Fundação Mozilla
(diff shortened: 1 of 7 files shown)


src/app/components/client/toolbar/UserMenu.tsx
```diff
@@ -105,7 +105,7 @@ export const UserMenu = (props: UserMenuProps) => {
       >
         <a
           className={styles.menuItemCta}
-          href={process.env.NEXT_PUBLIC_EXTERNAL_SUPPORT_URL}
+          href={process.env.NEXT_PUBLIC_MONITOR_SUPPORT_URL}
```
Vinnl, line 108: This ensures it uses the value from https://github.com/mozilla/blurts-server/blob/d1a7ce612f8976d8bfef56a29214ada527914822/.env-dist#L150.

mansaj (approved):

PR #3727, opened 2023-11-13
optimized the e2e suite, separated the yml for cron

mozrokafor:
# References: 
JIRA: [MNTOR-2394](https://mozilla-hub.atlassian.net/browse/MNTOR-2394)

Overall small optimization of the e2e code
- separated the cron yml for e2e tests
- make webserver conditional
- remove tests that were specific to prior UI design
- define the env of tests for each test.describe
- updated element locators

[MNTOR-2394]: https://mozilla-hub.atlassian.net/browse/MNTOR-2394?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/e2e/utils/helpers.ts
```diff
@@ -45,6 +45,28 @@ export const getBaseUrl = () => {
...
+    setTimeout(resolve, time);
+  });
+};
+
+export const waitForUrlOrTimeout = async (
+  page: Page,
+  urlSubstring: string,
+  timeout: number,
+) => {
+  const tries = 0;
```
pdehaan, line 59: Can you explain this bit?
I see this is a `const` and we check if `tries < 10` below on L61, but... we never seem to increment `tries` on each reattempt?

src/e2e/specs/breachResolution.spec.ts
```diff
@@ -110,45 +42,41 @@ test.describe("Breaches Dashboard - Headers", () => {
...
-      ).toBe("/user/breaches");
-    }).toPass();
+    expect(
+      await dataBreachPage.dataBreachesLogoLink.first().getAttribute("href"),
+    ).toBe("/user/breaches");
 
     await page.waitForLoadState("networkidle");
     await dataBreachPage.dataBreachesNavbarProfile.click();
 
     // verify mnanage your ff account link, settings option, help and support option, sign out option
```
pdehaan, line 52:
```suggestion
    // verify manage your Mozilla account link, settings option, help and support option, sign out option
```

.github/workflows/e2e_cron.yml
```diff
@@ -0,0 +1,84 @@
...
+        uses: actions/cache@v3
+        with:
+          path: ~/.cache/ms-playwright
+          key: playwright-browsers-${{ env.PLAYWRIGHT_VERSION }}
+
+      - name: Setup Playwright Browser
+        if: steps.cache-playwright-browsers.outputs.cache-hit != 'true'
+        run: npx playwright install --with-deps
+
+      - name: Run Playwright tests - ${{ inputs.environment != null && inputs.environment || 'stage' }}
```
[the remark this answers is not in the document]
mozrokafor: dev is currently not really stable to my knowledge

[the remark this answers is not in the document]
pdehaan: Then should it not be the default above on L10?

```
    inputs:
      environment:
        description: 'Environment to run the e2e against'
        required: false
        default: 'dev'
```

Disclaimer: I'm not 100% sure what the `choice` and `options` are in the workflow, so its a very ignorant question and this wasn't helpful https://github.blog/changelog/2021-11-10-github-actions-input-types-for-manual-workflows/ (or the linked [workflow inputs](https://docs.github.com/en/actions/learn-github-actions/workflow-syntax-for-github-actions#onworkflow_dispatchinputs) docs).

src/e2e/global-setup.ts
```diff
@@ -30,4 +34,35 @@ async function globalSetup() {
   await browser.close();
 }
 
+async function maybeStartWebServer(): Promise<void> {
+  if (process.env.E2E_START_LOCAL_SERVER === "true") {
+    console.log("[e2e] - Starting the local web server...");
+
+    // define port and timeout
```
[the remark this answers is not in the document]
mozrokafor: there's no need for port here, had it there originally while debugging

.github/workflows/e2e_cron.yml
```diff
@@ -0,0 +1,84 @@
...
+            - ui
+jobs:
+  test:
+    timeout-minutes: 60
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/setup-node@v3
+        with:
+          node-version: 20.5.1
```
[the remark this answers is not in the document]
mozrokafor: I don't think so but not quite sure, @Vinnl might have a better answer for you here

src/e2e/global-setup.ts
pdehaan, line 41: Ah cool, in that case&hellip;

```suggestion
    // define timeout
```

Although I could also argue that it's fairly self explanatory at this point and the comment isn't useful and can be removed entirely (unless you want to clarify that it's a server startup/initialization timeout).

.github/workflows/e2e_cron.yml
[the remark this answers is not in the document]
pdehaan: Fun With Greps&trade;
```sh
git grep -P "20\.\d+\.[\dx]+" -- ":(exclude)package-lock.json" | grep -v "\@types/node"

.circleci/config.yml:            - image: cimg/node:20.8.1
.circleci/config.yml:            - image: cimg/node:20.8.1
.github/workflows/playwright.yml:          node-version: 20.5.1
.github/workflows/unittests.yaml:          node-version: '20.5.x'
package.json:    "node": "20.8.x",
package.json:    "node": "20.8.1",
```

To be honest, I expected more embedded Node versions, so I'm guessing my <kbd>grep</kbd> sucks. Although I never would have caught the version in the Dockerfile, https://github.com/mozilla/blurts-server/pull/3631.

.github/workflows/e2e_cron.yml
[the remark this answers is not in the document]
mozrokafor: yep, thats going to be updated to stage as well

package.json
```diff
@@ -17,6 +17,8 @@
     "e2e": "playwright test src/e2e/",
     "e2e:debug": "playwright test src/e2e/ --ui",
     "e2e:smoke": "playwright test src/e2e/ --grep @smoke",
+    "e2e:ui": "playwright test src/e2e/ --grep @ui",
+    "e2e:all": "playwright test src/e2e/ --grep ''",
```
pdehaan, line 21: Does this need the empty `--grep ''`?
Or, how is this different from the vanilla `e2e` target above on L17?

src/e2e/specs/landing.spec.ts-snapshots/stage-landingFooter-chromium-linux.png
pdehaan, line 1: This diff feels wrong.

src/e2e/specs/landing.spec.ts-snapshots/stage-landingFooter-firefox-linux.png
pdehaan, line 1: This diff feels wrong.

package.json
mozrokafor: completely correct, this slipped

.github/workflows/e2e_cron.yml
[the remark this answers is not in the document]
Vinnl: You can actually leave out this line; [setup-node will automatically detect and use the version defined for Volta](https://github.com/actions/setup-node/issues/576#issuecomment-1250698896). One fewer place to keep up to date!

.github/workflows/e2e_cron.yml
```diff
@@ -0,0 +1,84 @@
...
+    inputs:
+      environment:
+        description: 'Environment to run the e2e against'
+        required: false
+        default: 'stage'
+        type: choice
+        options:
+          - stage
+          - prod
+          - dev
```
Vinnl, line 15: Suggestion: might be good to [align these with `NODE_ENV`](https://nextjs.org/docs/pages/building-your-application/configuring/environment-variables#environment-variable-load-order)?

```suggestion
          - production
          - development
```

Though assuming the latter actually points to Heroku, using `heroku` might be better, since it's easy to confuse `development` with `local`.

.github/workflows/e2e_cron.yml
```diff
@@ -0,0 +1,84 @@
...
+      type:
+        description: 'What Type of Test do you want to run?'
+        required: false
+        default: 'all'
+        type: choice
+        options:
+            - all
+            - smoke
+            - ui
```
Vinnl, line 24: Question: in what situations would we want to skip smoke tests?

.github/workflows/playwright.yml
Vinnl, line 1: Could be useful to rename this `e2e_pr` for alignment with the other file? And also update the name to something like "Monitor e2e Smoke Tests". (And maybe the other job "Monitor e2e Cron Tests" - just so they get placed next to each other when sorting alphabetically.)

.gitignore
```diff
@@ -34,6 +34,8 @@ storybook-static/
 
 # testing
 /coverage
+/src/e2e/test-results
+playwright-report
```
Vinnl, line 38: Aren't these already covered by lines 15-17?

package.json
```diff
@@ -17,6 +17,8 @@
     "e2e": "playwright test src/e2e/",
     "e2e:debug": "playwright test src/e2e/ --ui",
     "e2e:smoke": "playwright test src/e2e/ --grep @smoke",
+    "e2e:ui": "playwright test src/e2e/ --grep @ui",
+    "e2e:all": "playwright test src/e2e/",
```
Vinnl, line 21: I think this is exactly the same as `npm run e2e`? (Alternatively, we could remove that one and rename `e2e:debug` to `e2e:all:debug`, but to me, the "default" one being to run them all feels more intuitive.)

```suggestion
```

src/e2e/specs/breachResolution.spec.ts
```diff
@@ -88,13 +21,12 @@ test.describe("Breaches Dashboard - Headers", () => {
...
-    await expect(async () => {
-      await expect(dataBreachPage.dataBreachesLogo).toBeVisible();
-      await expect(dataBreachPage.dataBreachesNavbarProfile).toBeVisible();
-    }).toPass();
+    await expect(dataBreachPage.dataBreachesLogo).toBeVisible();
+    await expect(dataBreachPage.dataBreachesNavbarProfile).toBeVisible();
   });
 
-  test("Verify that the site header options work correctly for a signed in user", async ({
+  // skip as settings dropdown menu is currently wip for redesign TODO: update for redesign when its done
```
Vinnl, line 28: Thanks for describing why it's skipped!

src/e2e/specs/breachResolution.spec.ts
```diff
@@ -110,45 +42,41 @@ test.describe("Breaches Dashboard - Headers", () => {
...
-      ).toBeVisible();
-
-      // sign out
-      await expect(
-        dataBreachPage.dataBreachesNavbarProfileMenuSignOut,
-      ).toBeVisible();
-    }).toPass();
+    // verify manage your Mozilla account link, settings option, help and support option, sign out option
+    // menu is open
+    expect(await dataBreachPage.profileMenuExpanded()).toBeTruthy();
```
Vinnl, line 54: It's good practice to check for literal booleans, rather than type-casting it to a boolean. One day, it will avoid accidentally interpreting a non-empty string as `true` :grin: 

```suggestion
    expect(await dataBreachPage.profileMenuExpanded()).toBe(true);
```

src/e2e/specs/breachResolution.spec.ts
```diff
@@ -110,45 +42,41 @@ test.describe("Breaches Dashboard - Headers", () => {
...
+
+    // menu header
+    await expect(
+      dataBreachPage.dataBreachesNavbarProfileMenuHeader,
+    ).toBeVisible();
+
+    // head text
+    expect(
+      await dataBreachPage.dataBreachesNavbarProfileMenuHeaderSubtitle.textContent(),
+    ).toEqual("Manage your Mozilla account");
```
Vinnl, line 64: For atomic values, you should be able to use `.toBe`. `toEqual` is used to also check object properties, and in those cases, you'll usually want to use `toStrictEqual` anyway, which is why this version stands out a bit as a red flag to me.

```suggestion
    ).toBe("Manage your Mozilla account");
```

Vinnl: Alternatively, you can even use the more specific matcher [`toHaveText()`](https://playwright.dev/docs/api/class-locatorassertions#locator-assertions-to-have-text), which presumably gives more helpful error messages on test failures:

```ts
    expect(
      dataBreachPage.dataBreachesNavbarProfileMenuHeaderSubtitle
    ).toHaveText("Manage your Mozilla account");
```

src/e2e/specs/breachResolution.spec.ts
```diff
@@ -168,21 +96,19 @@ test.describe("Breaches Dashboard - Headers", () => {
...
-
-      // opens a new tab in which user is redirected to the "Monitor Help" page
-      expect(
-        await dataBreachPage.helpAndSupportNavButton.getAttribute("href"),
-      ).toContain(links.helpAndSupportNavButtonLink);
-    }).toPass();
+    // settings button redirects the user to "Settings" tab
+    expect(await dataBreachPage.settingsNavButton.getAttribute("href")).toBe(
+      links.settingsNavButtonLink,
+    );
```
Vinnl, line 102: For these attribute tests, there's also a more specific matcher [`toHaveAttribute`](https://playwright.dev/docs/api/class-locatorassertions#locator-assertions-to-have-attribute) for (presumably) better error messages on tests failures. Also applies to the instances below and in the other spec files:

```suggestion
    expect(dataBreachPage.settingsNavButton).toHaveAttribute(
      "href",
      links.settingsNavButtonLink,
    );
```

src/e2e/specs/landing.spec.ts
```diff
@@ -23,36 +26,32 @@ test.describe("Landing Page element verification", () => {
...
-        links.FAQUrl,
-      );
-      expect(await landingPage.TermsLink.getAttribute("href")).toContain(
-        links.TermsUrl,
-      );
-      expect(await landingPage.GithubLink.getAttribute("href")).toBe(
-        links.GithubUrl,
-      );
-    }).toPass();
+    expect(await landingPage.MozillaFooterLogoLink.getAttribute("href")).toBe(
```
Vinnl, line 29: As mentioned elsewhere, `.toHaveAttribute` can be used in this file as well.

src/e2e/specs/landing.spec.ts
```diff
@@ -90,30 +85,19 @@ test.describe("Landing Page Functionality Verification", () => {
...
-          breaches: [],
-          total: 0,
-          heading:
-            'We found (mocked) <span class="breach-result-email"></span> exposed in <span class="breach-result-count"></span> data breaches.',
-          dataClassStrings: [],
-          logos: [],
-        }),
-      });
+    await page.goto(process.env.E2E_TEST_BASE_URL as string, {
+      waitUntil: "networkidle",
```
Vinnl, line 89: Question: what's the reason this needs a custom `waitUntil` value?

src/e2e/specs/landing.spec.ts
```diff
@@ -90,30 +85,19 @@ test.describe("Landing Page Functionality Verification", () => {
...
     // generate email
     const randomEmail = `${Date.now()}_auto@restmail.net`;
 
-    // await page.pause()
+    // enter email for free scan
     await landingPage.enterScanEmail(randomEmail);
 
-    // verify scan page items
-    await expect(async () => {
+    const freeScanRedirect = await waitForUrlOrTimeout(page, "/scan", 5000);
```
Vinnl, line 98: If this is necessary, shouldn't it be part of `enterScanEmail`?

src/e2e/specs/landing.spec.ts
```diff
@@ -124,18 +108,23 @@ test.describe("Landing Page Functionality Verification", () => {
...
+      // verify 'have i been pwned' website redirect
       expect(await scanPage.haveIBeenPwnedLink.getAttribute("href")).toBe(
         "https://haveibeenpwned.com/",
       );
-    }).toPass();
+    } else {
+      console.log(
+        "[e2e] - Timeout reached without being redirect to the free scan results page",
+      );
+      expect(page.url().includes("/scan")).toBeTruthy();
```
Vinnl, line 127: Is this actually reason for a successful test?

src/e2e/specs/landing.spec.ts-snapshots/stage-landingFooter-chromium-darwin.png
Vinnl, line 1: These `*-darwin.png` files still are not reproducible by anyone on a different machine or in CI, and thus shouldn't be part of the repository IMO. Screenshots generated in CI would be fine, if accompanied by docs on how to generate them, and if they're actually verified by the cron job.

src/e2e/utils/helpers.ts
```diff
@@ -45,6 +45,29 @@ export const getBaseUrl = () => {
   return ENV_URLS[process.env.E2E_TEST_ENV as ENV] || ENV_URLS.local;
 };
 
+export const delay = async (time: number) => {
+  return new Promise(function (resolve) {
+    setTimeout(resolve, time);
+  });
+};
+
+export const waitForUrlOrTimeout = async (
```
Vinnl, line 54: Isn't this basically [`waitForURL`](https://playwright.dev/docs/api/class-page#page-wait-for-url)? And if not, could you add a comment why it's needed and why `waitForURL` doesn't suffice?

Vinnl (approved):
Have you already run the CI job on this branch? If not, can we do that (and see it run successfully) before merging?

src/e2e/specs/landing.spec.ts
mozrokafor: this fails if the monitor app was not able to successfully redirect to a free scan page

src/e2e/specs/landing.spec.ts
mozrokafor: good point, will have the result be returned in enterFreeScanEmail function

src/e2e/specs/landing.spec.ts
mozrokafor: iirc i think the errors are pretty close in terms of getting an "expected" and "actually" read out, interesting, i actually thought the other syntax was easier to read, but i'll update this

.github/workflows/e2e_cron.yml
mozrokafor: won't ever skip, just want to give stakeholders the ability to choose whether to run all, just smoke, or ui regression for their particular instance

.github/workflows/e2e_cron.yml
mozrokafor: this has been updated to 'stage' and 'prod' and its meant for the test code's base url and just any other test related url needs, i can update to use 'stage' and 'production' strings if needed to match the rest of the code

.github/workflows/e2e_cron.yml
Vinnl: Running UI regression tests skip smoke tests, right? I'm just wondering why anyone would ever want to run _only_ those.

src/e2e/specs/landing.spec.ts
Vinnl: We already know that it wasn't able to sucessfully redirect, because `freeScanRedirect` is `false`, no? So if this tests the same thing, couldn't we just move this to line 98, and skip the whole `if .. else` block?

src/e2e/utils/helpers.ts
mozrokafor: this is an issue with the click of free scan, that sometimes does not redirect to the scan page, will add a comment here

src/e2e/utils/helpers.ts
mozrokafor: ah updating to let, test been accidentally using the timeout comparison

.github/workflows/e2e_cron.yml
mozrokafor: my original thought on this is flexibility, for all stakeholders however, i guess all should work

src/e2e/utils/helpers.ts
pdehaan: I might be confused re: `timeout` vs `tries`+`delay`. Looks like the usage above is:
```js
const freeScanRedirect = await waitForUrlOrTimeout(page, "/scan", 5000);
```

But the max timeout value will maybe, kinda, always be 5000ms regardless. So an explicit `timeout` of 10_000ms would still timeout after 5s because we have a max 10 tries w/ a 500ms delay between checks. If we exceed the specified `timeout` OR the number of attempts is 10+, the `while` loop would exit and the function would return `void`|`undefined`|whatever.

Anyways, not a blocker, and possibly not even interesting/relevant, but just thought I'd throw it out there since it might be a gotcha if we end up using this function more, or somebody tries changing the /scan page timeout in the other file and wonders why it still exits at 5s despite a timeout of 40s... 🤷

mozrokafor: noted, will update this in the following ticket to use the 5000/or whatever is provided

src/e2e/specs/landing.spec.ts-snapshots/stage-landingFooter-chromium-darwin.png
Vinnl: I meant "arbitrary different machine" (i.e. someone not necessarily on a MacOS machine), though even on different MacOS machines, reproducing screenshots can be difficult - things like screen resolutions or installed fonts can all impact what is rendered. Thus, to enable everyone to contribute and avoid false negatives, I'd strongly prefer generating the screenshots in CI or some other remote service so that everyone can reproduce them, and discrepancies actually represent broken layouts.

PR #3751, opened 2023-11-17
Add a one-time bounce animation to the button

Vinnl:
Creating a draft PR to get the deploy preview, so I can show folks what this would look like.


Vinnl:
Oh, looks like it needs to be not-draft to trigger the preview deploy.

Vinnl:
Aargh, and it needs `main` as the target branch. This might trigger a review request for flod, sorry about that - no need to review, I'll close this later.

Vinnl:
Aargh, and it might need to have that target at the time of creation :facepalm:

PR #3760, opened 2023-11-17
Fix scan result cards expanding/collapsing in sync

Vinnl:
# References: 
Jira: MNTOR-2504 (this was about cards not opening in the first place. I can't reproduce that myself, but I'm fairly sure that's the same root cause), MNTOR-2463, MNTOR-2446.
Figma: N/A

# Description

The problem was that getLatestOnerepScanResults would return the result of a joined query of onerep_scan_results with onerep_scans, which would mean the `id`s of the former would get overridden by those of the latter, and thus multiple scan results all appearing to have the same ID.

I applied a couple of redundant fixes:
- `getLatestOnerepScanResults` now only returns data from the `onerep_scan_results` column.
- The dashboard now uses the `onerep_scan_result_id to identify scan results.
- Tests now verify that there is actually only a single card expanded.
- Tests now cover the situation where the dashboard shows multiple unresolved scan results.
- (Not committed because it's no longer representative:) I updated the mock data to give the multiple unresolved scan results the same ID, and verified that that caused the test to fail without the changes to `exposureCardKey`, and to pass _with_ those changes.

# How to test

Open the new dashboard with multiple scan results visible, and open and collapse different cards. There should always be at most one card open.

- [X] I've added or updated the relevant sections in readme and/or code comments
- [X] I've added a unit test to test for potential regressions of this bug.


src/app/(proper_react)/redesign/(authenticated)/user/dashboard/View.tsx
```diff
@@ -116,17 +118,18 @@ export const View = (props: Props) => {
   const tabSpecificExposures = getTabSpecificExposures(selectedTab);
   const filteredExposures = filterExposures(tabSpecificExposures, filters);
   const exposureCardElems = filteredExposures.map((exposure: Exposure) => {
-    const exposureCardKey = `${isScanResult(exposure) ? "scan" : "breach"}-${
-      exposure.id
-    }`;
+    const exposureCardKey = isScanResult(exposure)
+      ? "scan-" + exposure.onerep_scan_result_id
```
mansaj, line 122: Is `id` not enough to distinguish the different OneRep exposures?

src/app/(proper_react)/redesign/(authenticated)/user/dashboard/View.tsx
```diff
@@ -116,17 +118,18 @@ export const View = (props: Props) => {
...
       <li key={exposureCardKey} className={styles.exposureListItem}>
+        {exposure.id}
         <ExposureCard
           exposureData={exposure}
           isExpanded={exposureCardKey === activeExposureCardKey}
-          setExpanded={() => {
+          onToggleExpanded={() => {
             if (exposureCardKey === activeExposureCardKey) {
-              setActiveExposureCardKey("");
+              setActiveExposureCardKey(null);
```
mansaj, line 132: What's the reason behind using `null` instead of `""`?

I know this PR isn't to address the "collapse/expand multiple cards" bug, but I'm thinking this might be the root cause. I think I was checking for `"" == ""`, which would always return `true` for anything that has been expanded and collapsed before. I think `null==null` check would also yield the same result and not fix the issue?

src/app/(proper_react)/redesign/(authenticated)/user/dashboard/View.tsx
rhelmer: Is `exposure.id` the scan ID, and `exposure.onerep_scan_result_id` the scan result ID?

If so then I think `id` isn't enough, if the goal is to have a unique ID for each of the exposure cards, I think scan result ID (or something derived from it) is probably what's needed.

src/app/(proper_react)/redesign/(authenticated)/user/dashboard/View.tsx
Vinnl: Oh yeah, that's technically unrelated to this PR, but it was the first thing I tried and I felt it was probably a good change anyway. When using an empty string, the type system can't distinguish between "the currently expanded card" and "no card is expanded at the moment", which means we won't be warned when we're forgetting a check for the latter.

That said, I think this PR does actually address the "collapse/expand multiple cards" bug, because multiple cards would be considered to have the same ID (i.e. their scan ID, rather than that card's scan result's ID). I'm not sure why the check would be `"" === ""`, since `exposureCardKey` should never be an empty string?

src/app/(proper_react)/redesign/(authenticated)/user/dashboard/View.tsx
Vinnl: Yeah, it should be enough _combined with the other changes in this PR_, specifically [this one](https://github.com/mozilla/blurts-server/pull/3760/files#diff-2f5dac2b258572f0c86ba000438f9fd3cb9fd22320519cbb189df216979ee854R53), which ensures that the ID is now actually the ID of our database row of the scan result, rather than the ID of the database row of the scan.

However, I figured it might be a good idea to do some defense-in-depth: `onerep_scan_result_id` should _also_ be able to distinguish between different exposures, and has the additional advantage of being specific to scan results, and thus avoiding the risk of accidentally referring to a different value (like happens in the current code on `main`).

And since OneRep is the source of truth anyway, it also feels better to rely on their identifiers than on the ones of our DB cache? But I don't have a strong argument for that feeling, in terms of what could go wrong otherwise :)

src/db/tables/onerep_scans.ts
```diff
@@ -49,14 +49,14 @@ async function getLatestOnerepScanResults(
   const results =
     typeof scan === "undefined"
       ? []
-      : await knex("onerep_scan_results")
-          .select()
+      : ((await knex("onerep_scan_results")
+          .select("onerep_scan_results.*")
```
rhelmer, line 53: I think it'd be better to be explicit about the column names that you want to return - from the docs, it looks like you should be able to specify multiple column names here as an array of `table.column`.

rhelmer (changes requested):

src/db/tables/onerep_scans.ts
Vinnl: @rhelmer This function is used all over the website everywhere we want to display scans - I think we'll want to return all of them, regardless of what columns exist, now or in the future? (We were already doing that before this PR, by the way - and also all columns from the joined tables, except where they had the same column names - then they overwrote eachother.)

flozia (approved):
Looks good — thanks!

mansaj (approved):
lgtm thanks!

rhelmer (approved):

PR #3761, opened 2023-11-17
Raise max number of Dependabot PRs

Vinnl:
When a bunch of dependencies update around the same time - for example, when Definitely Typed applies a change to all `@types/*` dependencies - the backlog can get clogged up. And when one dependency update depends on another (e.g. right now, a Next.js update requires a Next-Auth update), this can result in manually retriggering Dependabot updates until you get the one you need.


.github/dependabot.yml
```diff
@@ -7,6 +7,7 @@ updates:
     directory: "/"
     schedule:
       interval: "weekly"
+    open-pull-requests-limit: 10
```
pdehaan, line 10: https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#open-pull-requests-limit

> _**By default, Dependabot opens a maximum of five pull requests for version updates.** Once there are five open pull requests from Dependabot, Dependabot will not open any new requests until some of those open requests are merged or closed._

pdehaan (approved):

rhelmer (approved):

PR #3778, opened 2023-11-20
MNTOR-2143 send page_location with cta_id

rhelmer:
# References: 
Jira: MNTOR-2143

# Description

We're not seeing outbound clicks to `mozilla.org` in GA4, I'm suspecting that we're overriding these with our `click` handler for elements that have `cta_id` data properties.

I think we probably need to be including the page location with `clicked_cta` events anyway, so this would be a "good enough" fix for the problem of being able to count CTA clicks to `mozilla.org`.

### DIFF
--- a/public/nextjs_migration/client/js/analytics.js
+++ b/public/nextjs_migration/client/js/analytics.js
@@ -2,40 +2,47 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
-const bodyDataset = document.body.dataset
+const bodyDataset = document.body.dataset;
 
 async function init({ debugMode, ga4MeasurementId }) {
-  if (navigator.doNotTrack === '1') {
+  if (navigator.doNotTrack === "1") {
     window.gtag = function () {
-      console.debug('Analytics disabled by DNT')
-    }
-    window.gtag()
+      console.debug("Analytics disabled by DNT");
+    };
+    window.gtag();
   } else {
-    await import(`https://www.googletagmanager.com/gtag/js?id=${ga4MeasurementId}`)
+    await import(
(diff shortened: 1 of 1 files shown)


public/nextjs_migration/client/js/analytics.js
```diff
@@ -26,8 +26,16 @@ async function init({ debugMode, ga4MeasurementId }) {
     })
 
     // Instrument CTA clicks for analytics.
-    document.querySelectorAll('[data-cta-id]').forEach(cta =>
-      cta.addEventListener('click', () => window.gtag('event', 'clicked_cta', { cta_id: cta.dataset.ctaId })))
+    document
+      .querySelectorAll("[data-cta-id]")
```
pdehaan, line 30: Q: why the switch from single quotes to double quotes? Or was this a Prettier thing (although if so, odd it didn't reformat the quotes on L8, 10, 21, 22, etc.

public/nextjs_migration/client/js/analytics.js
```diff
@@ -26,8 +26,16 @@ async function init({ debugMode, ga4MeasurementId }) {
...
+      .forEach((cta) =>
+        cta.addEventListener("click", () =>
+          window.gtag("event", "clicked_cta", {
+            cta_id: cta.dataset.ctaId,
+            page_location: location.href,
```
pdehaan, line 35: Also, on L23 above we used `window.location.hostname`. Should this be prefixed w/ `window.`?

public/nextjs_migration/client/js/analytics.js
rhelmer: Thanks - I just reset this branch and added a separate commit for the reformatting, so now it's consistent but the change should still be easy to see.

flozia (approved):

PR #3779, opened 2023-11-20
MNTOR-2431 - create alert for scan and profile limits

rhelmer:
# References: 
Jira: MNTOR-2431

# Description

A simple alert script to watch our scan and profile limits. Using Sentry for now, but ideally I'd like to plug this into Grafana, need to do some research to figure out the best way to do that.


src/scripts/onerepStatsAlert.js
```diff
@@ -0,0 +1,80 @@
...
+});
+
+/**
+ * Fetch the latest usage statistics from OneRep's API.
+ *
+ * @see https://docs.onerep.com/#tag/Statistics
+ */
+export async function checkStats() {
+  const scans = await (await onerepFetch("/stats/scans")).json();
+  const profiles = await (await onerepFetch("/stats/profiles")).json();
```
pdehaan, line 33: Possibly unpopular opinion and refactor, but it looks like `onerepFetch()` always expects/returns JSON. Should `onerepFetch()` just automatically return the `.json()` payload to avoid these parenthetical `await`s?
Or possibly a bad idea since some methods might be using `res.ok` or something to see if requests were successful.

src/scripts/onerepStatsAlert.js
```diff
@@ -0,0 +1,80 @@
...
+    ["Free scans", scans.manual, MAX_MANUAL_SCANS],
+    ["Paid scans", scans.initial, MAX_INITIAL_SCANS],
+    ["Profiles activated", profiles.activated, MAX_PROFILES_ACTIVATED],
+    ["Profiles created", profiles.created, MAX_PROFILES_CREATED],
+  ]) {
+    const [name, current, max] = alert;
+    if (current >= max) {
+      const msg = `Alert: OneRep scans over limit for ${name}. Current: ${current}, max: ${max}`;
+      console.error(msg);
+      Sentry.captureMessage(msg);
```
pdehaan, line 45: Q: will sentry be able to tie this error message to a user ID or anything useful? Or will we just get _x_ reports of "Alert: OneRep scans over limit for Free scans. Current: 100, max: 100"?

Or are these limits system wide and not for a specific user?

src/scripts/onerepStatsAlert.js
```diff
@@ -0,0 +1,80 @@
...
+
+// TODO use the shared version when this is converted to Typescript.
+async function onerepFetch(path, options) {
+  const onerepApiBase = process.env.ONEREP_API_BASE;
+  if (!onerepApiBase) {
+    throw new Error("ONEREP_API_BASE env var not set");
+  }
+  const onerepApiKey = process.env.ONEREP_API_KEY;
+  if (!onerepApiKey) {
+    throw new Error("ONEREP_API_BASE env var not set");
```
pdehaan, line 59: LOL
```suggestion
    throw new Error("ONEREP_API_KEY env var not set");
```

This is a typo in upstream as well: https://github.com/mozilla/blurts-server/blob/7a986fdecea1c70e0185532fc2d5bd6c1c42d704/src/app/functions/server/onerep.ts#L101-L104

src/scripts/onerepStatsAlert.js
```diff
@@ -0,0 +1,80 @@
...
+      const msg = `Alert: OneRep scans over limit for ${name}. Current: ${current}, max: ${max}`;
+      console.error(msg);
+      Sentry.captureMessage(msg);
+      // TODO send this to Grafana as well.
+    }
+  }
+}
+
+// TODO use the shared version when this is converted to Typescript.
+async function onerepFetch(path, options) {
```
pdehaan, line 52: Should `options` default to `{}`?

src/scripts/onerepStatsAlert.js
rhelmer: Ugh, so even worse this function is copy/pasted from typescript code, because we don't have cron jobs converted over to TS yet :/ I'll fix in both places for now, but this is something that we intend to fix.

.env-dist
```diff
@@ -160,3 +160,8 @@ NIMBUS_SIDECAR_URL=http://localhost:8001
 
 # The maximum number of jobs that the email breach alert worker will process.
 EMAIL_BREACH_ALERT_MAX_MESSAGES = 10000
+
+MAX_MANUAL_SCANS=100;
+MAX_INITIAL_SCANS=100;
+MAX_PROFILES_ACTIVATED=100;
+MAX_PROFILES_CREATED=100;
```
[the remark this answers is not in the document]
rhelmer: Heh yeah I actually pasted it from JS... I can't find a decent formatter in VSCode for this kind of file, I'll look again.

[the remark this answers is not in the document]
rhelmer: I've added a comment too

src/scripts/onerepStatsAlert.js
rhelmer: Yes this is a good point... I think a good refactoring would be maybe to make the `onerepFetch` function always check for `res.ok`, and make it `throw` if not.

src/scripts/onerepStatsAlert.js
rhelmer: No I don't think it's very user-specific - it's just a quota on a global resource.

I'm sending it to Sentry now out of convenience, all I really want is to be able to generate alerts if we ever see this. We're going to send any of these events to Grafana.

.env-dist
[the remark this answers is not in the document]
pdehaan: Looks like Prettier don't want to format your dotenv files; `https://github.com/prettier/prettier/issues/6103`

src/scripts/onerepStatsAlert.js
rhelmer: Oh also to be clear, getting any of these alerts means flipping these features off :) So just one is enough to start alerting on, we'll start getting it regularly since it's run from cron once any of these are elevated.

src/scripts/onerepStatsAlert.js
mansaj: Is there a ticket for this refactor or doing it in this PR?

src/db/migrations/20231220015816_onerep_stats.js
```diff
@@ -0,0 +1,23 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+export async function up(knex) {
+  return knex.schema
+    .createTable('onerep_stats', table => {
```
mansaj, line 7: Should this be limited to OneRep only or do we want to open this up for other stats in the future? 

ie. We can distinguish with a `type` col

src/scripts/onerepStatsAlert.js
```diff
@@ -0,0 +1,83 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
```
mansaj, line 1: @rhelmer the issue with relying a cronjob is that if / when it fails, we will have stale data and we can end up letting in more people than needed. 
1) The frequency of the runs need to be in sync with OneRep's schedule of updates (everyday)
2) We need to set up alerting for the cronjob

rhelmer: > @rhelmer the issue with relying a cronjob is that if / when it fails, we will have stale data and we can end up letting in more people than needed.
> 
>     1. The frequency of the runs need to be in sync with OneRep's schedule of updates (everyday)

Does the stats API only update daily? I assumed it was closer to real-time than that.

>     2. We need to set up alerting for the cronjob

I'll add it to Sentry.

mansaj: Yes, I think the API is updated probably near real time.

rhelmer: Looks like we have Sentry `captureCheckIn`, is there any other monitoring you had in mind? And yes I think it makes sense to just use this for monitoring, the pages that need to check the real-time status should just query the stats API independently.

src/scripts/onerepStatsAlert.js
rhelmer: I think it's too much for this PR, I've filed a ticket: https://mozilla-hub.atlassian.net/browse/MNTOR-2686

rhelmer:
@mansaj OK I cleaned this up a bunch and tested it, I think if we just scope this sending data to Grafana at a regular interval, then we can keep it pretty simple.

mansaj (approved):
Thanks for the cleanup. Lgtm

PR #3780, opened 2023-11-21
MNTOR-2543: Bulk Resolve API

mansaj:
# References: 
Jira: MNTOR-2543

# Description
Bulk resolving breaches for an user. The endpoint takes in the following request body
```
{
  dataType: String
}
```

And responds the following:
```
{
  success: Boolean, 
  useBreachId: Boolean,
  [email]: {
    [breachId]: {
      isResolved: Boolean
      resolutionsChecked: [String]
    }
  }
}
```

# Screenshot (if applicable)

Not applicable.

# How to test
Send a `PUT` request to `/api/v1/user/breaches/bulk-resolve` with the above request body


src/app/api/v1/user/breaches/bulk-resolve/route.ts
```diff
@@ -0,0 +1,103 @@
...
+        success: true,
+        breachResolutions: updatedSubscriber.breach_resolution,
+      });
+    } catch (e) {
+      logger.error(e);
+      return NextResponse.json({ success: false }, { status: 500 });
+    }
+  } else {
+    // Not Signed in, redirect to home
+    return NextResponse.redirect(appConstants.SERVER_URL);
```
rhelmer, line 101: It seems a little odd redirecting to home here if `email` is not a string, but above it returns `401` if `subscriber` does not exist? Do we expect this to happen, and if so why would we handle these two cases differently?

src/app/api/v1/user/breaches/bulk-resolve/route.ts
```diff
@@ -0,0 +1,103 @@
...
+export async function PUT(req: NextRequest): Promise<NextResponse> {
+  const session = await getServerSession(authOptions);
+  if (!session?.user?.subscriber) {
+    return new NextResponse(
+      JSON.stringify({ success: false, message: "Unauthenticated" }),
+      { status: 401 },
+    );
+  }
+
+  if (typeof session?.user?.email === "string") {
```
rhelmer, line 30: Nit, but the `?` is not really needed to `session.user` anymore (because it's guarded above and there's a `return`, Typescript compiler should be able to figure that one out).

But I am curious about why `email` not being a string gets a different HTTP response, should this check just be moved up and return a `401`?  Do we expect this to happen? Seems like something we might want to log.

rhelmer (changes requested):
This lgtm overall but I'd like to be more consistent in how we deal with unauthenticated users - since it's an API call I think it's totally fine to just return an HTTP error code with some JSON and not do a redirect. We don't expect users to navigate to this endpoint right?

src/app/api/v1/user/breaches/bulk-resolve/route.ts
mansaj: Oh yea! Good catch, need to get rid of this copy/paste

src/app/api/v1/user/breaches/bulk-resolve/route.ts
mansaj: good catch, wrong copy/paste!

mansaj:
> This lgtm overall but I'd like to be more consistent in how we deal with unauthenticated users - since it's an API call I think it's totally fine to just return an HTTP error code with some JSON and not do a redirect. We don't expect users to navigate to this endpoint right?

Yea sorry pushed this out in a rush. Finished up the PR description etc. I noticed that lots of the existing API endpoints are doing the redirect, which doesn't make sense (copy pasted)

src/app/api/v1/user/breaches/bulk-resolve/route.ts
```diff
@@ -0,0 +1,96 @@
...
+              },
+            },
+          };
+        }
+      }
+    }
+
+    // set useBreachId to mark latest version of breach resolution
+    // without this line, the get call might assume recency index
+    currentBreachResolution.useBreachId = true;
```
flozia, line 81: If we are setting this here do we need the duplicate `currentBreachResolution.useBreachId = true` in `L43`?

src/app/api/v1/user/breaches/bulk-resolve/route.ts
```diff
@@ -0,0 +1,96 @@
...
+          let currentResolutionsChecked =
+            currentBreachResolution[currentEmail]?.[breachId]
+              ?.resolutionsChecked;
+          if (currentResolutionsChecked) {
+            if (!currentResolutionsChecked.includes(dataTypeToResolve)) {
+              currentResolutionsChecked.push(dataTypeToResolve);
+            }
+          } else {
+            currentResolutionsChecked = [dataTypeToResolve];
+          }
```
flozia, line 61: Suggestion: We could simplify the nested conditions a bit and just make this:
```
          const currentResolutionsChecked =
            currentBreachResolution[currentEmail]?.[breachId]
              ?.resolutionsChecked ?? [];

          if (!currentResolutionsChecked.includes(dataTypeToResolve)) {
            currentResolutionsChecked.push(dataTypeToResolve);
          }
```

flozia (approved):
Thanks for adding the bulk resolve!

src/app/api/v1/user/breaches/bulk-resolve/route.ts
mansaj: Thanks for the catch!

src/app/api/v1/user/breaches/bulk-resolve/route.ts
```diff
@@ -0,0 +1,92 @@
...
+  if (!session?.user?.subscriber || typeof session.user?.email !== "string") {
+    return new NextResponse(
+      JSON.stringify({ success: false, message: "Unauthenticated" }),
+      { status: 401 },
+    );
+  }
+
+  try {
+    const subscriber: Subscriber = await getSubscriberByEmail(
+      session?.user?.email,
```
pdehaan, line 31: Should this just be `session.user.email`?
IIUC, we know `session.user` should be valid since we would have already bounced unauthenticated folks above on L23-26.

src/app/api/v1/user/breaches/bulk-resolve/route.ts
```diff
@@ -0,0 +1,92 @@
...
+import { getBreaches } from "../../../../../functions/server/getBreaches";
+import { getAllEmailsAndBreaches } from "../../../../../../utils/breaches";
+import {
+  getSubscriberByEmail,
+  setBreachResolution,
+} from "../../../../../../db/tables/subscribers";
+
+export async function PUT(req: NextRequest): Promise<NextResponse> {
+  const session = await getServerSession(authOptions);
+  if (!session?.user?.subscriber || typeof session.user?.email !== "string") {
```
pdehaan, line 22: Should this be `session?.user?.email` to match `session?.user?.subscriber`?
Not sure how likely `session` is to be undefined.

rhelmer (approved):

src/app/api/v1/user/breaches/bulk-resolve/route.ts
flozia: I’ll go ahead and merge this to unblock another PR, but will add this and the aforementioned change there! :)

src/app/api/v1/user/breaches/bulk-resolve/route.ts
flozia: Yes, that would work as well — see the reply to [your other comment](https://github.com/mozilla/blurts-server/pull/3780#discussion_r1403255773).

PR #3782, opened 2023-11-21
Fix linter for unit tests

flozia:
Fixes the linter in the unit tests.


jest.setup.ts
```diff
@@ -3,21 +3,26 @@
...
 import { setProjectAnnotations } from "@storybook/react";
-import { toHaveNoViolations } from "jest-axe";
-import { expect } from "@jest/globals";
 import { defaultFallbackInView } from "react-intersection-observer";
 import failOnConsole from "jest-fail-on-console";
-
 import * as globalStorybookConfig from "./.storybook/preview";
+import "jest-axe/extend-expect";
+
+declare module "expect" {
```
Vinnl, line 15: Let me know if I've understood this correctly. I also submitted a PR upstream so we hopefully won't need this later.

```suggestion
// `@types/jest-axe` doesn't extend `expect` when imported from `@jest/globals`.
// This manually sets the correct type. Note that it has to extend the
// TestingLibraryMatchers to avoid overriding those.
// Also see
// https://github.com/DefinitelyTyped/DefinitelyTyped/pull/67495
declare module "expect" {
```

Vinnl (approved):
Thanks for dealing with this!

package-lock.json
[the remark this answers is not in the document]
flozia: Reversed the changes. They resulted from updating some dependencies to see if this issue affects only this version  of Jest.

jest.setup.ts
flozia: Nice! Thanks for opening that PR.

PR #3266, opened 2023-07-29
MNTOR-1937: Autocomplete location cronjob

mansaj:
# References: 
Jira: MNTOR-1937

# Description
Move autocomplete location json file fetching/bundling process to a cronjob. 
* The job will execute daily
* The job will create the json file and upload it to s3
* Another script added to simply fetch the file from s3 and put it in the root directory as a part of the build step
* Build time / e2e testing should take significantly less time

# Screenshot (if applicable)

Not applicable.

# How to test

# Checklist (Definition of Done)
- [ ] Localization strings (if needed) have been added.
- [ ] Commits in this PR are minimal and [have descriptive commit messages](https://chris.beams.io/posts/git-commit/).
- [ ] I've added or updated the relevant sections in readme and/or code comments
- [ ] I've added a unit test to test for potential regressions of this bug.
- [ ] Product Owner accepted the User Story (demo of functionality completed) or waived the privilege.
- [ ] All acceptance criteria are met.
- [ ] Jira ticket has been updated (if needed) to match changes made during the development process.
- [ ] Jira ticket has been updated (if needed) with suggestions for QA when this PR is deployed to stage.

PR #3783, opened 2023-11-21
Remove unused dependency ts-node

flozia:
Since PR #3266 we do not use `ts-node` as a dependency.


Vinnl (approved):
Oh heh beat me to it, thanks :)

PR #3784, opened 2023-11-21
emailBreachAlerts: Move sentry cron check-in to end of finally block

rhelmer:
# Description

We're trying to diagnose whether the script is hanging in stage/prod or not, I noticed that the Sentry check-in comes after tearing down Postgres connections. We should have it be the very last thing the script does instead, this will allow Sentry to alert us to hanging cron jobs.

### DIFF
--- a/src/scripts/emailBreachAlerts.js
+++ b/src/scripts/emailBreachAlerts.js
@@ -311,11 +311,6 @@ if (process.env.NODE_ENV !== "test") {
           "env vars not set: GCP_PUBSUB_PROJECT_ID and GCP_PUBSUB_SUBSCRIPTION_NAME",
         );
       }
-      Sentry.captureCheckIn({
-        checkInId,
-        monitorSlug: SENTRY_SLUG,
-        status: "ok",
-      });
     })
     .catch((err) => console.error(err))
     .finally(async () => {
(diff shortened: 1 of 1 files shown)


src/scripts/emailBreachAlerts.js
```diff
@@ -311,18 +311,18 @@ if (process.env.NODE_ENV !== "test") {
...
     .catch((err) => console.error(err))
     .finally(async () => {
       // Tear down knex connection pools
       await knexSubscribers.destroy();
       await knexEmailAddresses.destroy();
       await knexHibp.destroy();
+      Sentry.captureCheckIn({
+        checkInId,
+        monitorSlug: SENTRY_SLUG,
+        status: "ok",
```
pdehaan, line 324: Q: Do we know if the `status` is "ok" (if we hit the `catch` block)?

mansaj (approved):

src/scripts/emailBreachAlerts.js
rhelmer: `finally` block should run regardless of whether the `try` or `catch` blocks ran:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/try...catch#finallystatements

PR #3785, opened 2023-11-21
Revert PR #3744

rhelmer:
# Description

This reverts PR #3744 due to a regression with the city/state dialog in the free scan welcome page.

PR #3789, opened 2023-11-23
Re-apply React-Aria upgrade, updating <LocationAutocompleteInput>

Vinnl:
This PR does two things:

1. Revert https://github.com/mozilla/blurts-server/pull/3785 to re-apply the React Aria upgrade
2. Fix the modifications I made to make `<LocationAutocompleteInput>` compatible with the new version, and add a unit test verifying it.

For #2, the main change here is to leave the async loading of the location
data to `useAsyncList`, as described in the react-aria docs:
https://react-spectrum.adobe.com/react-aria/useComboBox.html#asynchronous-loading

I think the root issue, however, was the ComboBox being a
controlled component, w.r.t. its selection, and the selection
value being set to the state variable instead of the value passed
to the `handleOnSelectionChange` handler, and thus being one render
behind. I've now changed it to both be an uncontrolled component,
and to use the callback parameter as selected value.

With these changes, I think we're closer to idiomatic usage of React-Aria's `<ComboBox>`.

### DIFF
--- a/src/app/(proper_react)/redesign/(authenticated)/user/welcome/Onboarding.test.tsx
+++ b/src/app/(proper_react)/redesign/(authenticated)/user/welcome/Onboarding.test.tsx
@@ -3,7 +3,7 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
 import { it, expect } from "@jest/globals";
-import { fireEvent, render, screen } from "@testing-library/react";
+import { render, screen } from "@testing-library/react";
 import { userEvent } from "@testing-library/user-event";
 import { composeStory } from "@storybook/react";
 import { axe } from "jest-axe";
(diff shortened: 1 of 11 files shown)


src/app/(proper_react)/redesign/(authenticated)/user/welcome/Onboarding.test.tsx
```diff
@@ -169,22 +140,57 @@ it("explainer dialog shows on step 2", async () => {
...
+
+  const firstNameField = screen.getByLabelText("First name*");
+  const lastNameField = screen.getByLabelText("Last name*");
   const proceedButton = screen.getByRole("button", {
     name: "Find exposures",
   });
+  const dobField = screen.getByLabelText("Date of birth*");
+  await user.type(firstNameField, "User");
+  await user.type(lastNameField, "Name");
+  await user.keyboard("[Tab]Tu[ArrowDown][Enter][Tab]");
```
flozia, line 155: That’s cool!

flozia (approved):
Works great and with `useAsyncList` and splitting out `useLocationSuggestions` this is a lot cleaner now!

src/app/(proper_react)/redesign/(authenticated)/user/welcome/Onboarding.test.tsx
Vinnl: Actually it was a major PITA to find out how to get the suggestion list to show up in test renders, and this was the only way I found to do it :sweat_smile:

PR #3813, opened 2023-11-29
Telemetry wrapper for Glean and GA

flozia:
# Description

This is a proposal to add a `useTelemetry` hook as a wrapper for reporting to both Glean and GA for the upcoming [analytics instrumentation work](https://mozilla-hub.atlassian.net/browse/MNTOR-2335).

# How to test

Load `/` and you should see outgoing Glean and GA ping.


.env-dist
```diff
@@ -1,5 +1,6 @@
 # local, heroku, stage, production
 APP_ENV=local
+NEXT_PUBLIC_APP_ENV=local
```
flozia, line 3: We’ve reverted most of our usage of `NEXT_PUBLIC_*` env variable. In this case, I think it might be nice to have this one it’s very straightforward to add `useTelemetry` to any client component without having to pass around the props needed to initialize Glean.

src/app/hooks/useGlean.ts
```diff
@@ -4,46 +4,116 @@
...
+  };
+};
+
+type GleanCategories = {
+  [key: GleanCategoryKey]: GleanEvent;
+};
+
+type GleanMetrics = {
+  $schema: string;
+} & GleanCategories;
```
flozia, line 42: All of the types in this file are not final and we might be able to get some from `@mozilla/glean`.

src/app/hooks/useGlean.ts
```diff
@@ -4,46 +4,116 @@
...
 
 // Custom hook that initializes Glean and returns the Glean objects required
 // to record data.
-export const useGlean = (channel: string, appEnv: string) => {
-  // Initialize Glean only on the first render
-  // of our custom hook.
+export const useGlean = () => {
+  const [gleanEvents, setGleanEvents] = useState<GleanMetricModules>(null);
+
+  const loadGleanModules = () => {
```
flozia, line 59: Load Glean modules dynamically from `metrics.yaml` so we do not have to import and export them manually from `useGlean`.

src/app/hooks/useGlean.ts
```diff
@@ -4,46 +4,116 @@
...
+          (modules, moduleName, moduleIndex) => ({
+            ...modules,
+            [moduleName]: importedModules[moduleIndex],
+          }),
+          {},
+        );
+        setGleanEvents(exportModules);
+      })
+      .catch((error) => {
+        console.log("Failed to load Glean modules:", error);
```
flozia, line 79: I assume we might want to report those to Sentry.

src/app/hooks/useTelemetry.ts
```diff
@@ -0,0 +1,47 @@
...
+
+  const record = (eventName: string, params: RecordParams) => {
+    if (!glean) {
+      return;
+    }
+
+    const { action, ...otherParams } = params;
+
+    // Record event via Glean
+    glean[eventName]?.[action]?.record({ path, ...otherParams });
```
flozia, line 30: That’s what I’m mostly unsure about: Since the generated Glean modules are being exported by `useGlean` we can record events in a more generic way than having to explicitly call them. I’m hoping to nail the types down and have a way to get them from `metrics.yaml`.

.env-dist
Vinnl: We didn't remove them because we didn't like them, but because they get compiled in at build time, and thus can't differ per environment. So they're fine to use when we're hardcoding them into `.env-dist`, but for a use case like this, it unfortunately won't work.

src/app/hooks/useGlean.ts
```diff
@@ -4,46 +4,116 @@
 
 "use client";
 
-import { useEffect } from "react";
+import { useEffect, useState } from "react";
 import Glean from "@mozilla/glean/web";
-import * as pageEvents from "../../telemetry/generated/page";
+// @ts-expect-error We do not have type declarations for metrics.yaml, yet.
```
Vinnl, line 9: If you open `next-env.d.ts`, then command-click `next/image-types/global`, you'll see an example of how you can type this in a `.d.ts` file:

```ts
declare module '*.yarm' {
  type GleanExtraKeys = Record<string, string | boolean | number>;

  type GleanCategoryKey = string | number | symbol;

  type GleanEvent = {
    [key: string]: {
      type: string;
      description: string;
      bugs: string;
      data_reviews: string;
      data_sensitivity: string;
      notification_emails: string;
      expires: string;
      extra_keys: GleanExtraKeys;
      record: (params: GleanExtraKeys) => void;
    };
  };

  type GleanCategories = {
    [key: GleanCategoryKey]: GleanEvent;
  };

  type GleanMetrics = {
    $schema: string;
  } & GleanCategories;

  const content: GleanMetrics;

  export default content;
}
```

Vinnl (commented):
I added a couple of inline comments, but reading this made me think of a potential alternative approach that I'd like to suggest. How about, instead of parsing the YAML at runtime, we add a step to the `build-glean` step that parses the YAML and generates TypeScript code that wraps Glean's `EventMetricType` object and either also handles GA reporting, or otherwise at least makes the YAML data available in code? Then `useTelemetry` could instead import those, with less of a performance impact, and stricter typing. (Essentially, I guess we'd be implementing https://bugzilla.mozilla.org/show_bug.cgi?id=1863946 ourselves.)

src/app/hooks/useGlean.ts
rhelmer: Glean.js supports typescript but our `package.json` has it outputting JS:
https://github.com/mozilla/blurts-server/blob/200db634df253ec020fdc892036826a5d694ed12/package.json#L27

If you change the `-o javascript` to `-o typescript` then it should produce `.ts` files instead which should have type definitions.

src/app/hooks/useGlean.ts
rhelmer: Yes I agree, `captureException(error)` would be appropriate here.

.env-dist
rhelmer: Yes unfortunately the build happens in CI, so any `NEXT_*` values are compiled into the JS and then published inside a docker container which gets reused for both stage/prod.

src/app/hooks/useGlean.ts
flozia: Good to know that we have TS available as an output format. I’ve changed the `build-glean` script accordingly in [3f81092](https://github.com/mozilla/blurts-server/pull/3813/commits/3f81092e160e5765e41c267ac390fdb41e13a02a).

next.config.js
```diff
@@ -135,6 +135,11 @@ const nextConfig = {
       type: "asset/source",
     });
 
+    config.module.rules.push({
+      test: /\.ya?ml$/,
```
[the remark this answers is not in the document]
flozia: With the recent change to import the Glean modules during build time, I removed the loader again in https://github.com/mozilla/blurts-server/pull/3813/commits/3c449905ce69c67c498040e69bb87ec46394a3b0.

.env-dist
flozia: I’ve [190c101](https://github.com/mozilla/blurts-server/pull/3813/commits/190c101f6991140e58894ff95dc227cf7b46ce24) a `<PublicEnvProvider>` to help us out here.

src/app/layout.tsx
```diff
@@ -58,7 +63,9 @@ export default async function RootLayout({
         data-ga4-measurement-id={process.env.NEXT_PUBLIC_GA4_MEASUREMENT_ID}
         data-node-env={process.env.NODE_ENV}
       >
-        <SessionProvider session={session}>{children}</SessionProvider>
+        <PublicEnvProvider publicEnv={PUBLIC_ENV}>
```
flozia, line 66: A provider for public env variables so we can access them from client components.

flozia (commented):
Upon the received feedback I’ve updated this PR to not import Glean modules dynamically anymore.

mansaj (approved):
Thanks for this work @flozia 
I think consolidating useGlean and useGa into one hook makes sense. I don't see a use case where we'd want to call these two hooks separately. In that case, perhaps we can emphasize that in the code comments or readme more clearly about the decision to ONLY use `useTelemetry` and deter people from using either of them individually.

Vinnl (commented):
This brought up some other inspiration, but it's quicker to show that through a PR than through comments here, one sec...

Vinnl:
@flozia See https://github.com/mozilla/blurts-server/pull/3840 and let me know what you think :)

rhelmer (approved):

src/scripts/buildGleanModuleIndex.js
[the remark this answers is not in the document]
flozia: The idea behind generating an `index.ts` is to be able to import the generated Glean modules during build-time and not dynamically as done before.

src/scripts/buildGleanTypes.js
flozia, line 1: Thanks to @Vinnl and https://github.com/mozilla/blurts-server/pull/3840 we are now creating Glean types instead of just a module index, which makes the use of the `useGlean` hock type-save.

src/app/hooks/useGa.ts
```diff
@@ -43,6 +43,8 @@ type Ga4EventOptions = {
   params: object;
 };
 
+// Note: Please refrain from using the `useGA` hook directly.
+// The peferred way to record telemetry is through `useTelemetry`.
```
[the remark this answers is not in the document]
flozia: Updated in [08aae01](https://github.com/mozilla/blurts-server/pull/3813/commits/08aae0135123c984d826a666bf835b8258e2b28d) — thanks!

Vinnl (approved):
Awesome job Florian :clap:

src/app/hooks/useGlean.ts
```diff
@@ -6,13 +6,14 @@
...
 import Glean from "@mozilla/glean/web";
-import * as gleanEvents from "../../telemetry/generated";
+import EventMetricType from "@mozilla/glean/private/metrics/event";
+import type { GleanMetricMap } from "../../telemetry/generated/_map";
 import { PublicEnvContext } from "../../contextProviders/public-env";
 
-export type GleanEvents = typeof gleanEvents;
-
+// Note: Please refrain from using the `useGlean` hook directly.
+// The preferred way to record telemetry is through `useTelemetry`.
```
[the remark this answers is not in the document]
flozia: That’s great, thanks! Added in [976ce0f](https://github.com/mozilla/blurts-server/pull/3813/commits/976ce0f630299e8ee3ac7105a4053febe6f9e01a).

PR #3817, opened 2023-12-01
Prevent reading of user on dashboard if there is no session

flozia:
# References: 
Jira: [MNTOR-2608](https://mozilla-hub.atlassian.net/browse/MNTOR-2608)

# Description

I was not able to reproduce the error we are seeing in Sentry:
```
TypeError: Cannot read properties of null (reading 'user')
  File "app:///_next/server/app/(nextjs_migration)/(authenticated)/user/breaches/page.js", line 47, in B
```

With a noticeable amount of them coming in and the bread crumbs pointing at the `/user/breaches` route, there seem to be situations where the `session` check in `layout.tsx` is not sufficient.

[MNTOR-2608]: https://mozilla-hub.atlassian.net/browse/MNTOR-2608?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/app/(nextjs_migration)/(authenticated)/user/layout.tsx
```diff
@@ -29,7 +28,7 @@ export type Props = {
 
 const MainLayout = async (props: Props) => {
   const session = await getServerSession(authOptions);
-  if (!session) {
+  if (!session || !session.user?.subscriber) {
```
Vinnl, line 31:
```suggestion
  if (!session?.user?.subscriber) {
```

Vinnl: Also I think we can then remove some `?`s from line 35 :grin:

src/app/(nextjs_migration)/(authenticated)/user/breaches/page.tsx
```diff
@@ -68,14 +69,15 @@ declare global {
 
 export default async function UserBreaches() {
   const session = await getServerSession(authOptions);
+  if (!session || !session.user?.subscriber) {
```
Vinnl, line 72:
```suggestion
  if (!session?.user?.subscriber) {
```

Vinnl (approved):
One of my takeaways so far is to not put too much logic in layouts (if only because TypeScript can't understand the consequences), so this makes sense to me :)

src/app/(nextjs_migration)/(authenticated)/user/layout.tsx
flozia: Yes — updated in [7c024e2](https://github.com/mozilla/blurts-server/pull/3817/commits/7c024e2b4c203d3653fd472e33e0cc529106f67c).

PR #3824, opened 2023-12-01
Optimize knex connection pooling for Postgres

rhelmer:
# Description

This has a few small changes from the preliminary load testing results against stage. The most common bottleneck when we get a spike in traffic right now is to hit the 500 connection limit on Postgres.

Before we try something like PGBouncer or scaling Postgres, we want to see how far we can get optimizing our connection pool settings - currently each Node application uses Knex, which starts a connection pool with a minimum of 2 and maximum of 10 connections.

I suspect that both the min and max are too high. For the min, the docs have this to say:
> Note that the default value of min is 2 only for historical reasons. It can result in problems with stale connections, despite tarn's default idle connection timeout of 30 seconds, which is only applied when there are more than min active connections. It is recommended to set min: 0 so all idle connections can be terminated.

For the max, it's a bit hard to measure directly, but I'd like to try halving it as a first step, and see if we can find a direct measurement to see if app instances are starving for connections. An indirect measurement would be that requests that depend on the database (which is many of them) are slower on average, but this could be caused by other things as well.

It looks like we could also be making better use of caching (e.g. `select * from breaches` is our top query by far, that could be put into Redis) but I'd like to tackle that in a separate PR.

Also, I've made the email breach alert worker script use the connection pool more effectively, so we might be able to run fewer of them.


src/scripts/emailBreachAlerts.js
```diff
@@ -71,199 +71,201 @@ const subscriptionName = process.env.GCP_PUBSUB_SUBSCRIPTION_NAME;
...
+                notificationType: "incident",
+              });
+
+              const emailTemplate = getTemplate(data, breachAlertEmailPartial);
+              const subject = getMessage("breach-alert-subject");
+
+              await sendEmail(data.recipientEmail, subject, emailTemplate);
+            } catch (e) {
+              console.error("Failed to add email notification to table: ", e);
+              setTimeout(process.exit, 1000);
```
pdehaan, line 234: Should this exit with a non-zero error code?

src/scripts/emailBreachAlerts.js
```diff
@@ -71,199 +71,201 @@ const subscriptionName = process.env.GCP_PUBSUB_SUBSCRIPTION_NAME;
...
-    /* c8 ignore stop */
+
+    subClient.acknowledge({
+      subscription: formattedSubscription,
+      ackIds: [message.ackId],
+    });
+    /* c8 ignore start */
+  } catch (error) {
+    console.error(`Notifying subscribers of breach failed: ${error}`);
   }
```
pdehaan, line 267: nit: Not sure if this should be `catch (e)` to be consistent w/ L232 and L245.

rhelmer:
After discussing this recently and looking at the system stats for Postgres, I think making the connection pool config changes is the more urgent fix, and it's simpler than the changes needed to the email worker, so I reverted those for now.

src/db/knexfile.js
```diff
@@ -25,6 +25,7 @@ if (APP_ENV === "heroku") {
 const RUNTIME_CONFIG = {
   client: "postgresql",
   connection: connectionObj,
+  pool: { min: 0, max: 5 }
```
pdehaan, line 28: LGTM, per https://knexjs.org/guide/#pool docs:
> _"It is recommended to set `min: 0` so all idle connections can be terminated."_

mansaj (approved):

PR #3838, opened 2023-12-05
Do not use worker for confetti

flozia:
# References: 
Jira: [MNTOR-2616](https://mozilla-hub.atlassian.net/browse/MNTOR-2616): Celebration screen animation not triggered on Firefox

# Description

If supported, the confetti animation is rendered by a web worker: For Firefox we need to set the CSP for `worker-src`.

[MNTOR-2616]: https://mozilla-hub.atlassian.net/browse/MNTOR-2616?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Vinnl (approved):
Oh heh, have you been able to verify that this is enough to fix it? I'm a bit surprised we don't need to pass in a nonce or add the worker's hash to the headers.

It might be good though if @rhelmer also takes a look.

flozia:
> Oh heh, have you been able to verify that this is enough to fix it? I'm a bit surprised we don't need to pass in a nonce or add the worker's hash to the headers.

Locally I have been able to verify that this fixes the issue.

rhelmer (changes requested):
Hm yeah I think this works but I don't think it's very safe unfortunately, per https://www.w3.org/TR/CSP2/#source-list-guid-matching

rhelmer:
> Hm yeah I think this works but I don't think it's very safe unfortunately, per https://www.w3.org/TR/CSP2/#source-list-guid-matching
> 
> > As defined above, special URL schemes that refer to specific pieces of unique content, such as "data:", "blob:" and "filesystem:" are excluded from matching a policy of * and must be explicitly listed. Policy authors should note that the content of such URLs is often derived from a response body or execution in a Document context, which may be unsafe. Especially for the [default-src](https://www.w3.org/TR/CSP2/#default_src) and [script-src](https://www.w3.org/TR/CSP2/#script_src) directives, policy authors should be aware that allowing "data:" URLs is equivalent to unsafe-inline and allowing "blob:" or "filesystem:" URLs is equivalent to unsafe-eval.

Is it possible to load this worker via an HTTPS URL instead of `blob:`? If not then I'd consider `nonce` or hash, but loading from URL is probably the simplest if that's possible.

rhelmer:
@flozia are you still working on this one?

flozia:
> Is it possible to load this worker via an HTTPS URL instead of blob:? If not then I'd consider nonce or hash, but loading from URL is probably the simplest if that's possible.

@rhelmer There is currently [no other way to load the worker](https://github.com/catdad/canvas-confetti/issues/139) other than using `blob:` — I suggest we do without the worker for now

rhelmer (approved):

PR #3841, opened 2023-12-05
MNTOR-2469: Script to fix all lowercase emails and replace sha/sub to HIBP

mansaj:
# References: 
Jira: MNTOR-2469

# Description
** Also cleaned up some old scripts **
From MNTOR-2451, we learned that we’ve been inserting uppercase emails into the database, causing an email matching error in production. 

We fixed the signup function so new emails stored in the db will always be lowercased, but the existing emails AND their sha1s need to be migrated / recalculated and resubmitted to HIBP for monitoring

# How to test
`node src/scripts/emailUpperCaseMigration.js`


src/scripts/emailUpperCaseMigration.js
```diff
@@ -0,0 +1,144 @@
+import Knex from "knex";
```
mansaj, line 1: This is the only file that needs to be reviewed FYI

src/scripts/emailUpperCaseMigration.js
```diff
@@ -0,0 +1,146 @@
...
+        );
+      }
+
+      emailPrevMaxId = emailRecord.id;
+      emailRecordsFixed++;
+    }
+  }
+
+  console.log("done.");
+  process.exit();
```
pdehaan, line 145: Should this script exit with a non-zero error code if we hit a `catch` block and rollback a transaction?

src/scripts/emailUpperCaseMigration.js
```diff
@@ -0,0 +1,146 @@
+import Knex from "knex";
+import knexConfig from "../db/knexfile.js";
+const knex = Knex(knexConfig);
```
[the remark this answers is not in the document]
mansaj: Fixed. Though since this is only going to be run once, it might not matter?

src/scripts/emailUpperCaseMigration.js
```diff
@@ -0,0 +1,146 @@
...
+
+import { subscribeHash } from "../utils/hibp.js";
+import { getSha1 } from "../utils/fxa.js";
+
+/**
+ * MNTOR-2469: One-off script to lowercase email addresses of subscribers and emails before hashing
+ */
+async function subscribeLowercaseHashToHIBP(emailAddress) {
```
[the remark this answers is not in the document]
mansaj: Fixed thanks

src/scripts/emailUpperCaseMigration.js
mansaj: Since this will be a one-off script, we will be watching the output and rerun if we run into any rollbacks.

src/scripts/emailUpperCaseMigration.js
```diff
@@ -0,0 +1,145 @@
...
+  return lowerCasedSha1;
+}
+
+(async () => {
+  const chunkSize = 1000;
+  console.log(
+    `Fixing lower-cased emails / sub to hashes in ${chunkSize}-sized chunks`,
+  );
+
+  console.log(`/** Subscribers Table **/`);
```
rhelmer, line 23: Hm, why the multi-line comment markers in the log statement?

src/scripts/emailUpperCaseMigration.js
```diff
@@ -0,0 +1,145 @@
...
+  let subPrevMaxId = 0;
+  while (subRecordsFixed < subsWithUpperCount) {
+    console.log(
+      `working on ${chunkSize}-sized chunk where id > ${subPrevMaxId} ...`,
+    );
+    const subRecordsWithUpperCharsChunk = await knex
+      .select("id", "primary_email")
+      .from("subscribers")
+      .where("id", ">", subPrevMaxId)
+      .whereRaw("primary_email != lower(primary_email)")
```
rhelmer, line 42: Too bad knex doesn't have support for this but I think this is the best approach.

rhelmer (approved):
I don't see anything obviously wrong, left a few comments but up to you.

I'm concerned this might take a very long time to run due to the number of rows we have in production and I doubt we're indexing on `lower(email)` and similar, but I guess it doesn't really matter if it takes a long time as long as we're not making this mistake for new records. I think doing it as a migration might be appropriate, but I'm not sure our infra can handle it if it takes a very long time to run, so having this as a standalone script is probably the best course of action. We should make sure that this is run in `tmux` or similar so it's less likely to get killed it runs for a long time.

I thought a little about optimization:
- Knex will automatically create a connection pool which this script won't take advantage of, you could probably increase the performance by making this more asynchronous (presumably this script is going to spend the vast majority of its time waiting on the database)
- Indexes would help, but I don't think it's worth doing since this is a one-time fix and we're unlikely to run these sorts of queries in normal operation. I'm not totally sure about this though - an index that includes e.g. `lower(email)` might be worthwhile.
- You could probably dispense with the pre-counting part.

src/scripts/emailUpperCaseMigration.js
rhelmer: Yeah, this is where the structured logging really helps :) We should get that working with cron jobs soon.

src/scripts/emailUpperCaseMigration.js
```diff
@@ -0,0 +1,145 @@
...
+      }
+
+      emailPrevMaxId = emailRecord.id;
+      emailRecordsFixed++;
+    }
+  }
+
+  console.log("done.");
+  process.exit();
+})();
```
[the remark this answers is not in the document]
mansaj: Hmm true, I was actually thinking about deleting all the one-off scripts and only keeping the cron ones. Doing a bit of that with this PR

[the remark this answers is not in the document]
mansaj: I think for scripts (especially simple ones), using IIFE seems the most intuitive for me

mansaj:
> I don't see anything obviously wrong, left a few comments but up to you.
> 
> I'm concerned this might take a very long time to run due to the number of rows we have in production and I doubt we're indexing on `lower(email)` and similar, but I guess it doesn't really matter if it takes a long time as long as we're not making this mistake for new records. I think doing it as a migration might be appropriate, but I'm not sure our infra can handle it if it takes a very long time to run, so having this as a standalone script is probably the best course of action. We should make sure that this is run in `tmux` or similar so it's less likely to get killed it runs for a long time.
> 
> I thought a little about optimization:
> 
>     * Knex will automatically create a connection pool which this script won't take advantage of, you could probably increase the performance by making this more asynchronous (presumably this script is going to spend the vast majority of its time waiting on the database)
> 
>     * Indexes would help, but I don't think it's worth doing since this is a one-time fix and we're unlikely to run these sorts of queries in normal operation. I'm not totally sure about this though - an index that includes e.g. `lower(email)` might be worthwhile.
> 
>     * You could probably dispense with the pre-counting part.

@rhelmer re 
pool: are you talking about adjusting the `pool` config? I think the knex connection pool is used with explicitly committing and rolling back transactions?

pre-counting: I think you are talking about splitting the work up to 1) finding the array of IDs where we need to adjust the email addresses 2) parallelize the work with jobs = `connection pool max`

PR #3842, opened 2023-12-05
automation to verify fixed tab tooltips and popups

mozrokafor:
# References: 
Jira: [MNTOR-2208](https://mozilla-hub.atlassian.net/browse/MNTOR-2208)

Handling automation to verify the the fixed tab's tooltips and popups

[MNTOR-2208]: https://mozilla-hub.atlassian.net/browse/MNTOR-2208?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/e2e/pages/dashBoardPage.ts
pdehaan, line 1: tiny nit: Unrelated, but why "dashBoardPage" filename, but the exported class is "DashboardPage"?
Not sure if it's worth renaming the file.

src/e2e/specs/dashboard.spec.ts
```diff
@@ -61,6 +61,34 @@ test.describe(`${process.env.E2E_TEST_ENV} - Breaches Dashboard - Headers`, () =
...
+      description:
+        "https://testrail.stage.mozaws.net/index.php?/cases/view/2301532",
+    });
+
+    // verify fixed tab's tooltips and popups
+    await dashboardPage.fixedTab.click();
+    await expect(dashboardPage.heresWhatsFixedCardTitle).toHaveText(
+      "Here’s what we fixed",
+    );
+    await expect(dashboardPage.fixedHeading).toBeVisible();
```
pdehaan, line 80: Also unrelated, but noticed in the earlier tests (L61) we did something like this:

```js
expect(await dashboardPage.actionNeededTab.isVisible()).toBeTruthy();
```

But here we're using a slightly different (albeit prettier) syntax:
```js
await expect(dashboardPage.fixedHeading).toBeVisible();
```

mozrokafor: good catch, will be moving away from truthy, will update

pdehaan (approved):

mansaj (approved):

PR #3850, opened 2023-12-08
Decrease padding in plan cards

Vinnl:
This makes the `?` icon not drop to a new line on Tony's device, but also in general gives the text some more room on smaller screens.


src/app/(proper_react)/redesign/(public)/PlansTable.tsx
```diff
@@ -227,6 +227,7 @@ export const PlansTable = (props: Props) => {
                       elems: { b: <b /> },
                     },
                   )}
+                  &nbsp;
```
Vinnl, line 230: I added these because the trigger buttons were flush with the text when they were on the same line.

flozia (approved):

PR #3869, opened 2023-12-12
Align Node CI version with local version

Vinnl:
This updates CI to use Node 20.10, like we use locally since 7bf8ad8e286bfe92543162f1e315b85f150bc0aa.

That new version does result in a change in coverage measuring, which caused our coverage to go down. I'll see if I can fix that in another PR.


rhelmer (approved):

.circleci/config.yml
```diff
@@ -35,7 +35,7 @@ executors:
                 description: "Which cimg/postgres tag to use"
                 type: string
         docker:
-            - image: cimg/node:20.9
+            - image: cimg/node:20.10
```
pdehaan, line 38: nit: why specific `20.10.0` above on L6 and L13, but drop the `.0` patch version here?
Although minorest of nits since it looks like it was that way before too.

.github/workflows/unittests.yaml
```diff
@@ -12,7 +12,7 @@ jobs:
       - name: Use Node.js
         uses: actions/setup-node@v3
         with:
-          node-version: '20.5.x'
+          node-version: '20.10.x'
```
pdehaan, line 15: Same nit as above, looks like we used generic "20.10" above in e2e_cron.yml#L24 and e2e_pr.yml#L35, but now adding `.x` patch semver operator.

Not sure if we want to pin everything to an exact major.minor.patch so half our CI doesn't update with patches and the other half doesn't.

netlify.toml
```diff
@@ -18,4 +18,4 @@
   # Default build command.
   command = "npm ci; cp .env-dist .env; npm run build-storybook"
 
-  environment = { NODE_VERSION = "18.12.1", NPM_VERSION = "9.6.5" }
+  environment = { NODE_VERSION = "20.10.0", NPM_VERSION = "10.2.4" }
```
pdehaan, line 21: ❤️ This made me actually write a wrapper around https://nodejs.org/download/release/index.json to see what the latest versions were, their matching node+npm versions, release dates, and whether it's an LTS and/or security release.

Looks like Node v20.10.0 includes npm 10.2.3. Why a sliightly different npm version here? Although I guess it does match our slightly mismatched Volta config. 🤔 

```sh
npx pdehaan/node-versions-md 20 | head -8
```

version  | npm      | date       | lts        | security
--------:|---------:|------------|------------|------------
v20.10.0 | 10.2.3   | 11/21/2023 | Iron       |
v20.9.0  | 10.1.0   | 10/23/2023 | Iron       |
v20.8.1  | 10.1.0   | 10/12/2023 |            | (SECURITY)
v20.8.0  | 10.1.0   | 9/27/2023  |            |
v20.7.0  | 10.1.0   | 9/17/2023  |            |
v20.6.1  | 9.8.1    | 9/6/2023   |            |

Looks like current package.json in `main` has:
https://github.com/mozilla/blurts-server/blob/26ed0928de671eac749ed27ff1c6fc3cc3c88177/package.json#L5-L8
https://github.com/mozilla/blurts-server/blob/26ed0928de671eac749ed27ff1c6fc3cc3c88177/package.json#L46-L49

.circleci/config.yml
Vinnl: Yeah, the only reason I did it was because that's how we did it before - I don't know the original reason. Personally I think it's pretty safe to leave off the patch version, for places that Dependabot can't update for us.

Vinnl:
Superseded by https://github.com/mozilla/blurts-server/pull/4031.

PR #3875, opened 2023-12-12
MNTOR-2467 - move Monitor to mozilla.org domain

rhelmer:
# References: 
Jira: MNTOR-2467

# Description

Allow new domain name in CSP and other configuration that lives in-tree.


mansaj (approved):
So the understanding is that we are only changing the domain names at this stage, leaving the name `Firefox Monitor` in tact for now. Because we have a lot of those in the code base, ie:
`"description": "Firefox Monitor",` in `package.json`

sentry.client.config.ts
```diff
@@ -49,7 +49,8 @@ Sentry.init({
 function getEnvironment() {
   if (
     document.location.origin === "https://monitor.firefox.com" ||
-    document.location.origin === "https://monitor.mozilla.com"
+    document.location.origin === "https://monitor.mozilla.com" ||
+    document.location.origin === "https://monitor.mozilla.org"
```
pdehaan, line 53: why `.com` _and_ `.org` TLDs? 

- https://vpn.mozilla.org/ redirects to https://www.mozilla.org/en-US/products/vpn/
- https://vpn.mozilla.com/ seems to just fail outright.

rhelmer: `.com` was already in-place, I just added `||` - I wanted to make this PR as additive as possible. If we don't end up using `.com` we can remove.

rhelmer:
> So the understanding is that we are only changing the domain names at this stage, leaving the name `Firefox Monitor` in tact for now. Because we have a lot of those in the code base, ie: `"description": "Firefox Monitor",` in `package.json`

That's my understanding, confirming with product that this is their expectation.

src/e2e/utils/helpers.ts
```diff
@@ -30,7 +30,7 @@ export const ENV_URLS = {
   local: "http://localhost:6060",
   heroku: "https://fx-breach-alerts.herokuapp.com",
   stage: "https://stage.firefoxmonitor.nonprod.cloudops.mozgcp.net",
-  prod: "https://monitor.firefox.com",
+  prod: "https://monitor.mozilla.org",
```
Vinnl, line 33: I imagine this changed line will make our prod E2E tests fail until we've fully transitioned? (Though I'm not actually sure if they succeed today...)

Possibly it should wait until we're ready to fully transition? (Similarly for the link in `<AppPicker>`, which has not been updated yet.)

Vinnl (approved):

Vinnl:
Oh, one more remark: it might be good to do a prod deploy with just this change and keeping an eye on that as it happens, since we can't test it in advance on stage.

rhelmer:
> Oh, one more remark: it might be good to do a prod deploy with just this change and keeping an eye on that as it happens, since we can't test it in advance on stage.

Yes agreed, that's why I didn't remove anything related to the current live domain - in case we need to switch back.

src/e2e/utils/helpers.ts
rhelmer: Yes good call-out - I'm planning for this change to roll out as soon as the dependencies like DNS, TLS, and the FxA redirect URI are ready to go.

rhelmer:
> > So the understanding is that we are only changing the domain names at this stage, leaving the name `Firefox Monitor` in tact for now. Because we have a lot of those in the code base, ie: `"description": "Firefox Monitor",` in `package.json`
> 
> That's my understanding, confirming with product that this is their expectation.

Yes confirmed, none of the strings or assets like the logo will change until later, but we're OK to start rolling out the domain change when ready (as mentioned in another comment, there are DNS, TLS, and FxA changes this depends on that SRE is working on).

PR #3876, opened 2023-12-13
Instrument Premium purchase dialog

Vinnl:
# References: 
Jira: MNTOR-2343
Figma: https://www.figma.com/file/LGjPMlfA7Li3xkS5YZ6VLY/Attribution-Documentation?type=whiteboard&node-id=156-31792&t=6qaaSpfRCWMPPiHA-4

# Description

The instrumentation for the "Get yearly plan" link doesn't work yet, because that's a link, and the page navigates away (and unloads the JavaScript) before the event could be sent. (Edit: [this should get dealt with in upstream Glean](https://matrix.to/#/!SCdsJdSTaQHjzEVrAE:mozilla.org/$muLULIgsOMaLwe3HR6HI_oJbMkyD5gZBoRN3GmDL8Ko).) I figured I'd submit the rest for review already to catch potential other issues.

Some other things of note are:
- The Figma refers to `clicked_cta`, but for things other than the actual purchase button, I used `button`, since they're not actually CTAs.
- I also stuck to the available parameters from `metrics.yml`, which means I'm not using `cta_name`, the event names are different, and I'm using `path` rather than `page_location`.
- I do have to obtain the path and session information in the calling code. Feels like it might be good to move those hooks into `useGlean`?
- I didn't know how I should see the data going into GA.

# How to test

Open the dashboard with a non-Premium account that has run their first scan, in a browser that does not send DoNotTrack. Click "Upgrade to Premium" in the top bar, then open the Network Console and click the year/month toggle or dialog close button. You should see telemetry requests being sent (though I don't think you can inspect the data that is sent).

- [ ] All acceptance criteria are met. - Some IDs are different

### DIFF
--- a/src/app/(proper_react)/redesign/(authenticated)/user/dashboard/Dashboard.test.tsx
+++ b/src/app/(proper_react)/redesign/(authenticated)/user/dashboard/Dashboard.test.tsx
@@ -563,6 +563,11 @@ it("counts in Glean how often people click the upgrade CTA to purchase the yearl
     name: "Upgrade to ⁨Premium⁩",
   });
   await user.click(premiumCtas[0]);
+  // Switch to the monthly tab by clicking it...
+  const productTabMonthly = screen.getByRole("tab", { name: "Monthly" });
+  await user.click(productTabMonthly);
+  // ...then back to the yearly tab by pressing the left arrow on the keyboard.
+  await user.keyboard("[ArrowLeft]");
   const confirmButton = screen.getByRole("link", {
     name: "Select yearly plan",
   });
--- a/src/app/(proper_react)/redesign/(authenticated)/user/dashboard/Dashboard.test.tsx
+++ b/src/app/(proper_react)/redesign/(authenticated)/user/dashboard/Dashboard.test.tsx
@@ -46,11 +46,13 @@ import Meta, {
   DashboardUsPremiumScanInProgressResolvedBreaches,
   DashboardInvalidPremiumUserNoScanResolvedBreaches,
 } from "./Dashboard.stories";
+import { useTelemetry } from "../../../../../hooks/useTelemetry";
 
 jest.mock("next/navigation", () => ({
   useRouter: jest.fn(),
   usePathname: jest.fn(),
 }));
+jest.mock("../../../../../hooks/useTelemetry");
 
 jest.mock("../../../../../components/client/DataBrokerImage", () => {
   return {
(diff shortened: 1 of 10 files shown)


src/app/(proper_react)/redesign/(authenticated)/user/dashboard/Dashboard.test.tsx
```diff
@@ -513,6 +515,72 @@ it("toggles between the product offerings in the premium upsell dialog", async (
...
+  expect(mockedRecord).toHaveBeenCalledWith(
+    "ctaButton",
+    "click",
+    expect.objectContaining({
+      button_id: "intent_to_purchase_monthly_plan_nav_modal",
+    }),
+  );
+});
+
+it("counts in Glean how ofen people click the upgrade CTA to purchase the monthly plan", async () => {
```
pdehaan, line 552:
```suggestion
it("counts in Glean how often people click the upgrade CTA to purchase the monthly plan", async () => {
```

src/app/(proper_react)/redesign/(authenticated)/user/dashboard/Dashboard.test.tsx
```diff
@@ -513,6 +515,72 @@ it("toggles between the product offerings in the premium upsell dialog", async (
   expect(productMontlyCta).toBeInTheDocument();
 });
 
+it("counts in Glean how ofen people click the upgrade CTA to purchase the yearly plan", async () => {
```
pdehaan, line 518:
```suggestion
it("counts in Glean how often people click the upgrade CTA to purchase the yearly plan", async () => {
```

pdehaan: Is the title correct? It seems to say _"&hellip; purchase the **yearly** plan"_ but then below on L532-536 we seem to be selecting the **_monthly_** plan.

src/app/(proper_react)/redesign/(authenticated)/user/dashboard/Dashboard.test.tsx
pdehaan: Same Q as above re: test name being "purchase the monthly plan" and the test selecting the yearly plan.

src/app/components/client/PremiumUpsellDialog.tsx
```diff
@@ -139,6 +157,21 @@ function PremiumUpsellDialogContent({
       <Button
         className={styles.productCta}
         href={isMonthly ? monthlySubscriptionUrl : yearlySubscriptionUrl}
+        onPress={() => {
+          // Note: This doesn't currently work; the event is now never sent to
+          //       the back-end because the page unloads before we can do so.
+          //       This will be dealt with in upstream Glean:
+          //       https://matrix.to/#/!SCdsJdSTaQHjzEVrAE:mozilla.org/$muLULIgsOMaLwe3HR6HI_oJbMkyD5gZBoRN3GmDL8Ko
```
rhelmer, line 164: 👍

rhelmer (approved):
lgtm - to see the events in GA4, make sure that `gtag` is configured with `debug_mode` set to `true` and check out `DebugView`: https://support.google.com/analytics/answer/7201382?hl=en

`analytics.js`, the predecessor to `gtag.js`, used to have an easy way to log to the browser console but I haven't been able to find a simple way for `gtag.js`... if we can figure this out it'd be really nice to make that the default when running with `APP_ENV=local`.

src/app/(proper_react)/redesign/(authenticated)/user/dashboard/Dashboard.test.tsx
```diff
@@ -513,6 +515,72 @@ it("toggles between the product offerings in the premium upsell dialog", async (
...
+  const confirmButton = screen.getByRole("link", {
+    name: "Select monthly plan",
+  });
+  // jsdom will complain about not being able to navigate to a different page
+  // after clicking the link; suppress that error, as it's not relevant to the
+  // test:
+  jest.spyOn(console, "error").mockImplementationOnce(() => undefined);
+  await user.click(confirmButton);
+
+  expect(mockedRecord).toHaveBeenCalledWith(
```
rhelmer, line 543: Nice to see tests! If it's helpful at all, Glean has a library to make writing tests easier: https://mozilla.github.io/glean/book/reference/general/initializing.html#testing-api

src/app/(proper_react)/redesign/(authenticated)/user/dashboard/Dashboard.test.tsx
Vinnl: Ugh... End-of-day committing, thanks for catching :facepalm: 237105aca

src/app/(proper_react)/redesign/(authenticated)/user/dashboard/Dashboard.test.tsx
Vinnl: 237105aca

src/app/components/client/PremiumUpsellDialog.tsx
```diff
@@ -95,7 +101,19 @@ function PremiumUpsellDialogContent({
...
+          onSelectionChange={(selectedKey) => {
+            record("button", "click", {
+              button_id:
+                selectedKey === "monthly"
+                  ? "selected_monthly_plan"
+                  : "selected_yearly_plan",
+              // Mocking the entire session object is too messy in tests:
+              /* c8 ignore next */
+              user_id: session.data?.user.subscriber?.fxa_uid ?? undefined,
+              path: currentPath,
```
flozia, line 113: I think it’s not needed to obtain and pass `path` on each `record` call. For Ga we are already passing it by default and we could change [this line](https://github.com/mozilla/blurts-server/blob/MNTOR-2343-premium-instrumentation/src/app/hooks/useTelemetry.ts#L38) in `useTelemetry` to:
```
void recordGlean(eventModule, event, { ...data, path });
```

Getting and passing `path` in `useGlean` and `useGa` hook is also a good alternative for instances where we might use them standalone.

flozia (approved):
Looks good to me and praise for the additional tests! I think sticking to the events and parameters in `metrics.yml` if possible is a great idea.

src/app/components/client/PremiumUpsellDialog.tsx
Vinnl: Right, the only reason I didn't do that is because that would mean that every Glean event then needs to define a `path` property - but I realise now that it actually does, so TypeScript shouldn't complain about it. In fact, it will then warn us if we ever add something to `metrics.yml` that _doesn't_ accept a `path`, which is probably a good idea - we'll probably always want to record it.

And in fact, I suppose the same holds for `session_id`. Thus, I added those to `useTelemetry` in ffb396d74.

Vinnl:
Thanks @flozia for pointing me to [this extension](https://chromewebstore.google.com/detail/google-analytics-debugger/jnkmfdileelhofjcijamephohjechhna?pli=1). With that, I was able to verify that pings were indeed sent to GA, including one too many - when the dialog opened, the toggle would send an "I'm toggled" signal. Fixed that in 502fa9e64.

src/app/hooks/useTelemetry.ts
```diff
@@ -35,7 +37,11 @@ export const useTelemetry = () => {
     > = [Glean, Ga],
   ) => {
     if (platforms.includes(Glean)) {
-      void recordGlean(eventModule, event, data);
+      void recordGlean(eventModule, event, {
+        path: path,
+        user_id: session.data?.user.subscriber?.fxa_uid ?? undefined,
```
flozia, line 42: Nice!

flozia (approved):

PR #3877, opened 2023-12-13
Determine subscribe links on the server-side

Vinnl:
On client-side renders, the environment variables won't be available, and the links will include `undefined`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -83,6 +83,7 @@
     "react-cookie": "^6.1.1",
     "react-dom": "^18.2.0",
     "react-stately": "^3.28.0",
+    "server-only": "^0.0.1",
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -83,6 +83,7 @@
     "react-cookie": "^6.1.1",
     "react-dom": "^18.2.0",
     "react-stately": "^3.28.0",
+    "server-only": "^0.0.1",
```
Vinnl, line 86: I caved in and finally added this, as per Next.js's instructions: https://nextjs.org/docs/app/building-your-application/rendering/composition-patterns#keeping-server-only-code-out-of-the-client-environment

flozia (approved):

PR #3880, opened 2023-12-14
MNTOR-2594 - Log error if unsubscribed user loads welcome-to-premium …

rhelmer:
…page

# References: 
Jira: MNTOR-2594

# Description

Small bit of server-side logging, trying to determine how long it's taking for FxA to begin showing subscriptions. We may need to get more proactive about checking this status.


src/app/(proper_react)/redesign/(authenticated)/user/dashboard/fix/data-broker-profiles/welcome-to-premium/WelcomeToPremiumView.tsx
```diff
@@ -32,6 +35,18 @@ export function WelcomeToPremiumView(props: Props) {
   );
   const dataPointReduction = getDataPointReduction(summary);
 
+  // MNTOR-2594 - log any users that are on welcome-to-premium page but not subscribed.
+  getServerSession(authOptions)
+    .then((session) => {
+      if (!session?.user.fxa?.subscriptions.includes("monitor")) {
+        logger.error("user_not_subscribed", {
+          page: "welcome-to-premium",
+          subscriber_id: session?.user.subscriber?.id,
```
Vinnl, line 44: Don't we want to avoid logging this? Or is this just for foxfooding? (If so, maybe we could add a condition to limit this to foxfooders, or to deactivate after a certain date?)

Vinnl (approved):

src/app/(proper_react)/redesign/(authenticated)/user/dashboard/fix/data-broker-profiles/welcome-to-premium/WelcomeToPremiumView.tsx
rhelmer: I'll go ahead and remove it, the most important thing is that we count how many times and when this occurs. I'll send it to Sentry as well, since it's something we'll want to show up easily in triage.

mansaj (approved):

PR #3881, opened 2023-12-14
MNTOR-2593 - update user session every minute

rhelmer:
# References: 
Jira: MNTOR-2593
Figma: 

# Description

This uses the approach documented at https://next-auth.js.org/getting-started/client#refetching-the-session - I didn't implement the suggestion in the comment to detect if the user is online, I don't think Monitor really works at all offline right now.


src/app/components/client/SubscriptionCheck.tsx
```diff
@@ -10,12 +10,11 @@ import { useEffect } from "react";
...
 
+  // Poll for the session every minute.
   useEffect(() => {
-    void update();
-
-    // This should only run once per page load - `update` will always appear to be changed.
-    // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, []);
+    const interval = setInterval(() => void update(), 1000 * 60);
+    return () => clearInterval(interval);
```
Vinnl, line 16: Unfortunately `setInterval` is notoriously tricky to align with React's rendering cycle. See https://overreacted.io/making-setinterval-declarative-with-react-hooks/

(There's a "Just Show Me The Code" section with a proper implementation.)

Also... Oof, it feels like somehow this shouldn't be necessary :( )

Vinnl (approved):

PR #3823, opened 2023-12-01
Welcome scan telemetry

flozia:
# References: 
Jira:
- [MNTOR-2339](https://mozilla-hub.atlassian.net/browse/MNTOR-2339): US New User Onboarding/Sign Up Instrumentation
- [MNTOR-2340](https://mozilla-hub.atlassian.net/browse/MNTOR-2340): US Existing User Welcome and Onboarding Instrumentation

Figma: https://www.figma.com/file/LGjPMlfA7Li3xkS5YZ6VLY/Attribution-Documentation?type=whiteboard&node-id=111-14239&t=6qaaSpfRCWMPPiHA-4

# Description

This PR is adding telemetry events for the onboarding instrumentation.

# How to test

Go through the onboarding flow and confirm the telemetry events are submitted.

# Checklist (Definition of Done)
- [ ] Localization strings (if needed) have been added.
- [ ] Commits in this PR are minimal and [have descriptive commit messages](https://chris.beams.io/posts/git-commit/).
- [ ] I've added or updated the relevant sections in readme and/or code comments
- [ ] I've added a unit test to test for potential regressions of this bug.
- [ ] Product Owner accepted the User Story (demo of functionality completed) or waived the privilege.
- [ ] All acceptance criteria are met.
- [ ] Jira ticket has been updated (if needed) to match changes made during the development process.
- [ ] Jira ticket has been updated (if needed) with suggestions for QA when this PR is deployed to stage.

[MNTOR-2339]: https://mozilla-hub.atlassian.net/browse/MNTOR-2339?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[MNTOR-2340]: https://mozilla-hub.atlassian.net/browse/MNTOR-2340?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #3883, opened 2023-12-14
Prevent useTelemetry() from breaking Storybook

Vinnl:


flozia (approved):
Thanks for adding that. I was just bringing the telemetry for the welcome scan up to speed where I ran into the issue and added [this](https://github.com/mozilla/blurts-server/pull/3823/files#diff-e201de4e2ea4ee79f492c0495d7f3fce6389034fe26bfb8f3c877f8c03edf914R34-R40).

PR #3886, opened 2023-12-15
Round down the discount percentage to whole number

Vinnl:
This makes it easier to parse, without overpromising. (This change was requested by Tony on Slack.)


flozia (approved):
Looks good!

PR #3889, opened 2023-12-15
Stop automatically sending the session ID

Vinnl:
The `<PageLoadEvent>` component comes up with a deliberate session ID that is also unique for people who aren't logged in, and is gated behind a flag. Thus, we probably want to hold off on setting a session ID until we've come up with the definite strategy and approval there.

### DIFF
--- a/src/app/hooks/useTelemetry.ts
+++ b/src/app/hooks/useTelemetry.ts
@@ -5,7 +5,6 @@
 "use client";
 
 import { usePathname } from "next/navigation";
-import { useSession } from "next-auth/react";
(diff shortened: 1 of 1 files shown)


rhelmer (approved):
Thanks for catching that - I think the `user_id` key was behind a flag before, we could do that if we want to be able to test on dev/stage until we have approval to collect.

Vinnl:
@rhelmer Yep, it's just a bit of a hassle because that all needs to be determined server-side, whereas the hook runs client-side. Which is solvable, but I wasn't sure how important tracking this was, i.e. whether it was worth the effort.

PR #3896, opened 2023-12-18
Keep the ManualRemove View universal

Vinnl:
Although it runs as a server component in production, we also want to render it in Storybook, which runs everything as a client component.

This should make Netlify succeed again.


flozia (approved):
Great catch, thanks!

PR #3906, opened 2023-12-19
MNTOR-2316: US Landing Page Variation when hit 18K OneRep activations

mansaj:
# References: 
Jira: MNTOR-2316

# Description
When we hit 18k OneRep activation limit within 30 days, user will be redirected to the waitlist
This PR is only the API portion of the work. There will be a small follow-on PR for the url redirect

# How to test


src/app/functions/server/onerep.ts
```diff
@@ -439,3 +449,47 @@ export async function fetchAllPages<Data>(
 
   return dataList.flat();
 }
+
+// Local instance map to cache results to prevent excessive API requests
+// Would be nice to share this cache with other pod via Redis in the future
+const profileStatsCache = new Map<string, ProfileStats>();
```
rhelmer, line 455: Hm this would work... I was thinking of having a cron job that did this and populated the database, which we could query from Grafana as well for graphing and alerts. wdyt?

mansaj: What would that table look like? Since telegraf does the time series magic automatically, it would just be one entry in the db?

mansaj: kind of like subscriber count is just a count(*) that telegraf charts overtime by itself

src/app/functions/server/onerep.ts
```diff
@@ -439,3 +449,47 @@ export async function fetchAllPages<Data>(
...
+export async function getProfilesStats(
+  from: Date,
+  to?: Date,
+): Promise<ProfileStats | undefined> {
+  const fromDateString = from.toISOString().substring(0, 10);
```
pdehaan, line 460: Not sure if we want to extract this into some sort of `isoFromDate()` helper since we're doing a `.substring(0, 10)` in at least 3 places here.
```js
function isoFromDate(date = new Date()) {
  const [dateString] = date.toISOString().split("T", 1);
    return dateString;
}
```

I think we could also use `Array#unshift()` if you want a one-liner.
But might even reduce this to something like this, since the default `date` is `new Date()` so we wouldn't need the `to ? to... : new Date()...` bits:
```js
const fromDateString = isoFromDate(from);
const toDateString = isoFromDate(to);
```

Conversely
```js
const toDateString = (to ?? new Date()).toISOString().split("T", 1).shift();
```

src/app/functions/server/onerep.ts
```diff
@@ -439,3 +449,47 @@ export async function fetchAllPages<Data>(
 
   return dataList.flat();
 }
+
+// Local instance map to cache results to prevent excessive API requests
+// Would be nice to share this cache with other pod via Redis in the future
+const profileStatsCache = new Map<string, ProfileStats>();
+export async function getProfilesStats(
```
pdehaan, line 456: Is this called from somewhere yet?

src/app/functions/server/onerep.ts
rhelmer: I put a simple example in https://github.com/mozilla/blurts-server/pull/3779 - I think the query would need to be a slightly more complicated but not much more (`select count(*) where current >= max` or so)

src/app/functions/server/onerep.ts
```diff
@@ -439,3 +449,47 @@ export async function fetchAllPages<Data>(
...
+  from: Date,
```
flozia, line 457: For [MNTOR-2315](https://mozilla-hub.atlassian.net/browse/MNTOR-2315) we need the total number of scans performed and I guess we could just use this function as well. We could make the `from` prop optional as well and in case no time period is provided just return the total — or do you prefer us explicitly passing `from`.

src/app/functions/server/onerep.ts
mansaj: Not yet, but as soon as we get the landing page done, we will

src/app/functions/server/onerep.ts
mansaj: This makes sense, I'll wait till #3779 is merged then

src/app/functions/server/onerep.ts
mansaj: @flozia ah, that makes sense. I can make the `from` optional as well

mansaj: Fixed

rhelmer (approved):

PR #3910, opened 2023-12-19
Mntor 2624 e2e health dashboard data

mozrokafor:
# References: 
Jira: [MNTOR-2624](https://mozilla-hub.atlassian.net/browse/MNTOR-2624)

Creating data for an e2e health dashboard for monitor. This takes it results  from the current test run then generates the appropriates files and places them into the e2e-report branch to be used by the UI later

[MNTOR-2624]: https://mozilla-hub.atlassian.net/browse/MNTOR-2624?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


E2EReporter.ts
```diff
@@ -0,0 +1,207 @@
...
+        this.tempData.testsFlaky++;
+        this.tempData.flakes.push(data);
+        break;
+    }
+
+    // increase test count
+    this.tempData.totalTests++;
+
+    // add duration
+    this.tempData.duration = this.tempData.duration + result.duration;
```
pdehaan, line 149: Not sure if this is any better:
```suggestion
    this.tempData.duration += result.duration;
```

E2EReporter.ts
```diff
@@ -0,0 +1,207 @@
...
+    };
+
+    if (!suiteResult) {
+      this.suites.push(data);
+    }
+  }
+
+  onTestEnd(test: TestCase, result: TestResult) {
+    // find the suite this test belongs to
+    const suiteId = test.parent.title.split(" ").join(" ");
```
pdehaan, line 94: same Q here re `.split(" ").join(" ")`.

src/e2e/global-teardown.ts
```diff
@@ -0,0 +1,132 @@
...
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import https from "https";
+import { readFileSync, writeFileSync } from "fs";
+import { CustomTestResult } from "../../E2EReporter";
+
+async function globalTearDown() {
+  // check env for cron, teardown is only needed for teardown run
+  if (process.env.IS_CRON) {
```
pdehaan, line 11: I might invert your logic here (and L45-47 below) so you check if `IS_CRON` _isn't_ true and just abort early:
```js
if (process.env.IS_CRON !== "true") {
  console.log("[e2e_log] - Not a cron json, json report will not be created");
  return;
}
```
That way we lose a depth of nesting and the `else` bit and keeps is slightly flatter.

src/e2e/global-teardown.ts
```diff
@@ -0,0 +1,132 @@
...
+    // get data from e2e-reports branch files -- these are api calls
+    const {
+      existingFailedReports,
+      existingFlakeReports,
+      existingFullReports,
+      newFullReportJson,
+      newFailReportJson,
+      newFlakeReportJson,
```
pdehaan, line 19: nit: but keeps the `existing*` and `new*` in the same order.
```suggestion
      newFailReportJson,
      newFlakeReportJson,
      newFullReportJson,
```

src/e2e/global-teardown.ts
```diff
@@ -0,0 +1,132 @@
...
+    // merge new and existing reports
+    existingFullReports.push(newFullReportJson as unknown as CustomTestResult);
+    if (newFlakeReportJson.length)
+      existingFlakeReports.push(
+        newFlakeReportJson as unknown as CustomTestResult,
+      );
+    if (newFailReportJson.length)
+      existingFailedReports.push(
+        newFailReportJson as unknown as CustomTestResult,
+      );
```
pdehaan, line 31: Interesting that ESLint didn't trigger something about [`curly`](https://eslint.org/docs/latest/rules/curly) rule, but maybe we don't enforce curly braces for single statements. 🤷

src/e2e/global-teardown.ts
```diff
@@ -0,0 +1,132 @@
...
+    const existingFull = JSON.stringify(existingFullReports);
+    const existingFlake = JSON.stringify(existingFlakeReports);
+    const existingFailed = JSON.stringify(existingFailedReports);
+
+    try {
+      writeFileSync("full_test_report.json", existingFull, "utf-8");
+      writeFileSync("flake_test_report.json", existingFlake, "utf-8");
+      writeFileSync("failed_test_report.json", existingFailed, "utf-8");
+    } catch (err) {
+      console.error("Error writing to json", err);
```
pdehaan, line 43: Do we care about the reports enough to do something like `process.exitCode = 1;` to fail the CI?

src/e2e/global-teardown.ts
```diff
@@ -0,0 +1,132 @@
...
+    return {
+      existingFullReports,
+      existingFlakeReports,
+      existingFailedReports,
+      newFullReportJson,
+      newFlakeReportJson,
+      newFailReportJson,
+    };
+  } catch (err) {
+    console.log("[e2e_log] - error reading report files", err);
```
pdehaan, line 95:
```suggestion
    console.error("[e2e_log] - error reading report files", err);
```

src/e2e/global-teardown.ts
```diff
@@ -0,0 +1,132 @@
...
+            resolve(parsedData);
+          } catch (err) {
+            reject(err);
+          }
+        });
+      })
+      .on("error", (err) => {
+        reject(err);
+      });
+  });
```
pdehaan, line 129: Might just be easier to use `fetch()` API. Something like:

```js
return fetch(url).then(res => res.json());
```

src/e2e/global-teardown.ts
```diff
@@ -0,0 +1,132 @@
...
+      );
+    if (newFailReportJson.length)
+      existingFailedReports.push(
+        newFailReportJson as unknown as CustomTestResult,
+      );
+
+    // convert the merged reports
+    const existingFull = JSON.stringify(existingFullReports);
+    const existingFlake = JSON.stringify(existingFlakeReports);
+    const existingFailed = JSON.stringify(existingFailedReports);
```
pdehaan, line 36: Not sure where the best place for this code is, but it might also be good to do some data/log rotation so we don't end up with 24MB JSON files hidden in a branch.
Can't imagine we need more than 90 or 180 days worth of historical history and can maybe do `.slice(-180)` here.

E2EReporter.ts
```diff
@@ -0,0 +1,207 @@
...
+    };
+
+    // prepare the different json report
+    const reportJson = JSON.stringify(report, null, 2);
+
+    // flaky tests
+    const flakeJson = JSON.stringify(this.tempData.flakes, null, 2);
+
+    // failed tests
+    const failedJson = JSON.stringify(this.tempData.failures, null, 2);
```
pdehaan, line 183: not sure if we want the `, null, 2` pretty-print business or if we just use `JSON.stringify(abc)` and write out minified logs and save 20-30% disk space.

mansaj:
Re the point about using a branch to store data vs. having this in `main`.. if we go with a branch (what it seems in the PR), would we constantly have to keep this branch up to date with main?

pdehaan:
> Re the point about using a branch to store data vs. having this in `main`.. if we go with a branch (what it seems in the PR), would we constantly have to keep this branch up to date with main?

I'll let @mozrokafor be the official answererer to that, but&hellip; I don't _think_ so.
I imagine we're just using the branch as a file storage and don't care about any of the other files in the branch [for now]. Semi-related, but I was also curious if we could use Telemetry or Glean to just submit stats to somewhere in a more central pipeline for processing (if we ever wanted to go the route of dashboards in BigQuery or some other service; versus building bespoke dashboards for every project and then having to maintain that infra).

E2EReporter.ts
```diff
@@ -0,0 +1,207 @@
...
+    skipped: [] as CustomTestResult[],
+    flakes: [] as CustomTestResult[],
+  };
+
+  onBegin() {
+    this.startTime = Date.now();
+  }
+
+  onTestBegin(test: TestCase) {
+    const suiteId = test.parent.title.split(" ").join(" ");
```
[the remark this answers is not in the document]
mozrokafor: its turning test blcok string titles to a temp id, so space need to be removed and the words rejoined with _

E2EReporter.ts
```diff
@@ -0,0 +1,207 @@
...
+    const suiteResult = this.suites.find((s) => s.suiteId === suiteId);
+
+    const data: CustomSuiteResult = {
+      title: test.parent.title,
+      fileName: test.parent.parent!.title,
```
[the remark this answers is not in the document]
mozrokafor: title = describe block
filename = text file name

src/e2e/global-teardown.ts
mozrokafor: i believe that would require a third party

mozrokafor:
> Re the point about using a branch to store data vs. having this in `main`.. if we go with a branch (what it seems in the PR), would we constantly have to keep this branch up to date with main?

what @pdehaan said is correct here, the branch will simply be used for data storage and read

PR #3911, opened 2023-12-19
MNTOR-2648 - Walkthrough section LP

codemist:
# References: 
Jira: MNTOR-2648
Figma: https://www.figma.com/file/YqpVpUbSTzZMj9Jbp2po8Q/Mozilla-Monitor-Premium-LP?type=design&node-id=1942-12535&mode=design&t=8rXVLQVugwqt52dS-4

# Screenshot (if applicable)

Non-US: 
image

US:
image

# How to test
Go to http://localhost:6060/redesign

### DIFF
--- a/src/telemetry/metrics.yaml
+++ b/src/telemetry/metrics.yaml
@@ -153,6 +153,8 @@ popup:
       flow_id:
         description: A randomly generated unique identifier for following user flows within the FxA system.
         type: string
+      popup_id:
+        description: The ID of the popup that was shown, or some way to identify where on the page the interaction is located.
 
 button:
   click:


src/app/(proper_react)/redesign/(public)/ValuePropositions.tsx
```diff
@@ -0,0 +1,514 @@
...
+          fontSize="28"
+          fontWeight="bold"
+          letterSpacing="0px"
+        >
+          <tspan x="124" y="319.26">
+            Your glassesdotcom.com{" "}
+          </tspan>
+          <tspan x="124" y="353.26">
+            password was exposed
+          </tspan>
```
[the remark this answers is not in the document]
Vinnl: I don't have full context yet, so one thing you'll need to consider is if this will still work if the localised strings are much shorter (e.g. Japanese, I suppose) or longer (e.g. German or Dutch), but you could potentially split it up like this?

```suggestion
          {l10n.getFragment("string-id", {
              elems: {
                "first-line": <tspan x="124" y="319.26"/>,
                "second-line": <tspan x="124" y="353.26"/>
              },
          })};
```

And then the Fluent string would look like this:

```ftl
string-id = <first-line>Your glassesdotcom.com</first-line> <second-line>password was exposed</second-line>
```

Though also cc @flodolo in case he has other ideas.

locales-pending/landing-all.ftl
```diff
@@ -9,3 +9,17 @@ landing-all-hero-emailform-submit-label = Get free scan
 
 # This is a label underneath a big number "14" - it's an image that demos Monitor.
 landing-all-hero-image-chart-label = exposures
+
+# Value Proposition 
+
+landing-all-value-prop-fix-exposures = We'll help you fix your exposures
+landing-all-value-prop-fix-exposures-description = Our mission is to put control of your personal data back in your hands. We’ll help you resolve data breaches and keep your info private – and we’ll respect your privacy in the process.
```
flodolo, line 16:
```suggestion
landing-all-value-prop-fix-exposures-description = Our mission is to put control of your personal data back in your hands. We’ll help you resolve data breaches and keep your info private — and we’ll respect your privacy in the process.
```

locales-pending/landing-all.ftl
```diff
@@ -9,3 +9,17 @@ landing-all-hero-emailform-submit-label = Get free scan
...
+landing-all-value-prop-info-at-risk = What info could be at risk?
+landing-all-value-prop-info-at-risk-description = Data leaks are unfortunately part of our digital lives. Your passwords, contact details, financial information, and other personal info can be exposed, putting you at risk of identity theft.
+landing-all-value-prop-scanning-for-exposures = Scanning for exposures...
```
flodolo, line 19:
```suggestion
landing-all-value-prop-scanning-for-exposures = Scanning for exposures…
```

locales-pending/landing-all.ftl
```diff
@@ -9,3 +9,17 @@ landing-all-hero-emailform-submit-label = Get free scan
...
+landing-all-value-prop-get-started = Get started
+landing-all-value-prop-enter-info = Enter info
+landing-all-value-prop-find-exposures = Find exposures
+landing-all-value-prop-exposure-example-leaked-pw = Leaked Passwords
+landing-all-value-prop-exposure-example-fix-it = Let's fix it
+landing-all-value-prop-exposure-example = Your glasses.com password was exposed
```
flodolo, line 25: Using a real website in an example sounds like a terrible idea?

Also, please leave an empty line at the end of the file.

flodolo (changes requested):

flodolo (approved):
For what it's worth, I'm not a fan of the approach: the idea of adding localized images for 2 locales doesn't scale, and treats other locales as 2nd class citizens.

src/app/(proper_react)/redesign/(public)/value-prop-images/fr/leaked-password-example-fr.svg
pdehaan, line 1: @flodolo Can i get a set of eyes on this?
The filename is "-fr" which makes me ignorantly think French. And the acute accents almost support that, but then the other nav heading has a tilde and "Contraseñas filtradas" seems to translate as Spanish as "Leaked passwords". So now I'm confused (and kind of echos not doing translated images because this is hard to verify outside of Pontoon).

src/app/(proper_react)/redesign/(public)/value-prop-images/progress-card.svg
pdehaan, line 1: Does this image not get localized?
Also, should the reference "en" strings be in an "en" folder (to match /de/ and /fr/)?

src/app/(proper_react)/redesign/(public)/value-prop-images/fr/leaked-password-example-fr.svg
flodolo: This content is not localized, so it's really confusing.

But yes, that text at the top is Spanish, not French (good catch!).

src/app/(proper_react)/redesign/(public)/WalkthroughImages.tsx
```diff
@@ -0,0 +1,80 @@
...
+export const ScanningForExposuresIllustration = (props: Props) => {
+  let imageSrc;
+
+  if (props.countryCode === "de") {
+    imageSrc = ScanningForExposuresImageDe;
+  } else if (props.countryCode === "fr") {
+    imageSrc = ScanningForExposuresImageFr;
+  } else {
+    imageSrc = ScanningForExposuresImage;
+  }
```
pdehaan, line 49: Not sure if we just want to default the `imageSrc` to "en" string  on L41 and then do overrides for "de" and "fr":

```suggestion
  let imageSrc = ScanningForExposuresImage;

  if (props.countryCode === "de") {
    imageSrc = ScanningForExposuresImageDe;
  } else if (props.countryCode === "fr") {
    imageSrc = ScanningForExposuresImageFr;
  }
```

src/app/(proper_react)/redesign/(public)/WalkthroughImages.tsx
```diff
@@ -0,0 +1,80 @@
...
+export const LeakedPasswordExampleIllustration = (props: Props) => {
+  let imageSrc;
+
+  if (props.countryCode === "de") {
+    imageSrc = LeakedPasswordExampleImageDe;
+  } else if (props.countryCode === "fr") {
+    imageSrc = LeakedPasswordExampleImageFr;
+  } else {
+    imageSrc = LeakedPasswordExampleImage;
+  }
```
pdehaan, line 70: Same as above:

```suggestion
  let imageSrc = LeakedPasswordExampleImage;

  if (props.countryCode === "de") {
    imageSrc = LeakedPasswordExampleImageDe;
  } else if (props.countryCode === "fr") {
    imageSrc = LeakedPasswordExampleImageFr;
  }
```

src/app/(proper_react)/redesign/(public)/LandingView.tsx
```diff
@@ -40,6 +48,80 @@ export const View = (props: Props) => {
...
+              {props.l10n.getString("landing-all-value-prop-fix-exposures")}
+            </h2>
+            <p>
+              {props.l10n.getFragment(
+                "landing-all-value-prop-fix-exposures-description",
+                {
+                  elems: {
+                    privacy_link: (
+                      <a
+                        href="https://www.mozilla.org/en-US/firefox/privacy/"
```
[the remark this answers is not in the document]
Vinnl: Also, if we set `target="_blank"`, maybe we need the `<OpenInNew>` icon following it as well?

(Which is a bit of a hassle, but could be done something like this:

```ts
                 elems: {
                    privacy_link: (
                      ({ children }) => <a
                        href="https://www.mozilla.org/en-US/firefox/privacy/"
                        target="_blank"
                      >{props.children} <OpenInNew alt={l10n.getString("open-in-new-tab-alt")}/></a>
                    ),
                  },
```

)

src/app/(proper_react)/redesign/(public)/SignUpForm.module.scss
```diff
@@ -0,0 +1,38 @@
+@import "../../../tokens";
+
+.form {
+  display: flex;
+  flex-wrap: wrap;
+  gap: $spacing-xl;
+
+  &.isHero {
```
Vinnl, line 8: I don't see this class on the form element?

(But it might still be nice to add it, instead of adding it directly on the `input` and `button`, and then move the hero-specific styles for those elements into this block - then all hero-specific styles are together. Although I also see the argument for keeping input- and button-related styles together, so up to you.)

src/app/(proper_react)/redesign/(public)/SignUpForm.tsx
```diff
@@ -44,16 +47,25 @@ export const SignUpForm = (props: Props) => {
...
         {l10n.getString("landing-all-hero-emailform-submit-label")}
       </Button>
-      <label htmlFor={emailInputId}>
-        {l10n.getString(
-          props.eligibleForPremium
-            ? "landing-premium-hero-emailform-input-label"
-            : "landing-all-hero-emailform-input-label",
-        )}
-      </label>
+      {props.isHero ? (
```
Vinnl, line 58: [All inputs need a label](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input#accessibility_concerns):

> When including inputs, it is an accessibility requirement to add labels alongside. This is needed so those who use assistive technologies can tell what the input is for.

Ideally, they're also visible, to provide a larger click target. But since the design doesn't include them, and the input fields are pretty big anyway, we can probably just wrap them into a `<VisuallyHidden>` if it's not in the hero.

src/app/(proper_react)/redesign/(public)/WalkthroughImages.tsx
```diff
@@ -0,0 +1,80 @@
...
+  testId: string;
+  countryCode: string;
+}) => (
+  <Image
+    src={image}
+    alt={l10n.getString(
+      "landing-all-value-prop-scanning-for-exposures-illustration-alt",
+    )}
+    data-testid={testId}
+    country-code={countryCode}
```
Vinnl, line 31: `country-code` isn't a valid HTML attribute, I think?

What _is_ a valid attribute is [`lang`](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/lang), so presumably that would be more appropriate here?

src/app/(proper_react)/redesign/(public)/LandingView.test.tsx
```diff
@@ -38,14 +43,50 @@ describe("When Premium is not available", () => {
...
     });
   });
+
+  it("shows the scanning for exposures illustration in the fix your exposures section", () => {
+    const ComposedDashboard = composeStory(LandingNonUs, Meta);
+    render(<ComposedDashboard />);
+
+    const scanningForExposuresIllustration = screen.getByTestId(
+      "scanning-for-exposures-image",
+    );
```
Vinnl, line 63: As mentioned on Slack, [this](https://testing-library.com/docs/queries/about/#priority) is order of preference when it comes to selectors, with the ones closest to how a user interacts with the page being most preferred. Users can't see test IDs, which is why that option is listed last :)

So this matcher could have been:

```ts
    const scanningForExposuresIllustration = screen.getByRole(
      "img",
      { name: "Scanning for exposures progress bar" },
    );
```

And then the test would have caught a bug: both images have the same alt text :)

But as I suggested elsewhere, these images probably don't need alt texts. Which means that the user probably doesn't really interact with the image, which means that... We're back to finding it through a test ID :sweat_smile:

src/app/(proper_react)/redesign/(public)/LandingView.test.tsx
```diff
@@ -306,4 +347,12 @@ describe("When Premium is available", () => {
 
     expect(signIn).toHaveBeenCalledTimes(1);
   });
+
+  it("shows the progress card illustration in the fix your exposures section", () => {
+    const ComposedDashboard = composeStory(LandingUs, Meta);
+    render(<ComposedDashboard />);
+
+    const progressCardIllustration = screen.getByTestId("progress-card-image");
+    expect(progressCardIllustration).toBeInTheDocument();
```
Vinnl, line 356: Thinking a bit about what bugs might actually occur... This is currently primarily verifying that an image with the right test ID exists in the document. Which is OK, but requires us to do some extra work just for the sake of the test (make sure we use the correct test ID everywhere), and those test IDs don't actually affect the user.

So one thing we could also check for, is whether the Premium-specific content is actually shown (using [`'.toHaveTextContent`](https://github.com/testing-library/jest-dom#tohavetextcontent), and whether non-Premium specific content is _not_ shown (`.not.toHaveTextContent`).

Vinnl (commented):
It's looking really good, and it's great to see you writing tests :) I've got a couple of relatively small remarks, and there's some good input from Peter and flod as well.

src/app/(proper_react)/redesign/(public)/WalkthroughImages.tsx
pdehaan: Or `data-country-code=` if we want to use custom attributes.

locales-pending/landing-all.ftl
```diff
@@ -16,5 +16,7 @@ landing-all-value-prop-fix-exposures = We’ll help you fix your exposures
 landing-all-value-prop-fix-exposures-description = Our mission is to put control of your personal data back in your hands. We’ll help you resolve data breaches and keep your info private — and we’ll <privacy_link>respect your privacy</privacy_link> in the process.
 landing-all-value-prop-info-at-risk = What info could be at risk?
 landing-all-value-prop-info-at-risk-description = Data leaks are unfortunately part of our digital lives. Your passwords, contact details, financial information, and other personal info can be exposed, putting you at risk of identity theft.
-landing-all-value-prop-scanning-for-exposures-illustration-alt = Scanning for exposures progress bar
-landing-all-value-prop-leaked-password-illustration-alt = Leaked password example breach resolution step
+
+# Quote
```
Vinnl, line 20: I think these are [supposed to](https://projectfluent.org/fluent/guide/comments.html) have two `#`s (same goes for other instances):

```suggestion
## Quote
```

locales-pending/landing-all.ftl
```diff
@@ -16,5 +16,7 @@ landing-all-value-prop-fix-exposures = We’ll help you fix your exposures
...
+
+landing-all-quote = <data_breaches>Data breaches</data_breaches> happen every 11 minutes, exposing your private information — but don’t worry, we can help.
```
Vinnl, line 22: It might be worth adding a comment on what the `<data_breaches>` tag does, so localisers understand what text it should apply to?

locales-pending/landing-premium.ftl
```diff
@@ -103,5 +103,11 @@ landing-premium-plans-cards-feature-guidance = <b>Guided help</b> to fix high-ri
 landing-premium-plans-cards-feature-monitoring = Continuous monitoring
 
 # Value proposition
+
+landing-premium-value-prop-fix-exposures-description = We provide steps to follow when you’ve been affected by a data breach and can even remove your data from more than 190 sites trying to sell it — and we <privacy_link>respect your privacy</privacy_link> in the process.
```
Vinnl, line 107: Don't we use a variable for the number `190` elsewhere, since the number might go up in the future? (In which case this would also become a pluralised string.)

src/app/(proper_react)/redesign/(public)/SignUpForm.tsx
Vinnl: Huh interesting, indeed I would expect axe to catch that. So I did some digging, and [it does have a rule for this](https://dequeuniversity.com/rules/axe/4.7/label), but apparently it says:

>  Lastly a placeholder attribute may be used to give text inputs an accessible name. This is _not_ a recommended solution as the visual label (the placeholder text) will be removed once the user enters text into the input, causing them to not know what the input is for.

So the `placeholder` caused the test not to fail, but it's still recommended to add an actual label.

src/app/(proper_react)/redesign/(public)/value-prop-images/fr/leaked-password-example-fr.svg
Vinnl: Ahora está en francés :)

src/app/(proper_react)/redesign/(public)/LandingView.tsx
```diff
@@ -42,37 +43,79 @@ export const View = (props: Props) => {
...
+            }}
           />
         </div>
         <div className={styles.heroImage}>
           <HeroImage {...props} />
         </div>
       </header>
 
-      <div className={styles.valuePropositionWrapper}>
+      <section className={styles.quoteWrapper}>
```
Vinnl, line 57: `<main>` already serves as a section, so I think the `<section>`s in this file should just be `<div>`s.

src/app/(proper_react)/redesign/(public)/LandingView.tsx
```diff
@@ -42,37 +43,79 @@ export const View = (props: Props) => {
...
+        <div className={styles.quote}>
+          <h2>
+            {props.eligibleForPremium
+              ? props.l10n.getFragment("landing-premium-quote", {
+                  elems: {
+                    data_brokers: <span className={styles.emphasis} />,
```
Vinnl, line 63: We have an element for emphasis :) (Likewise on line 68)

```suggestion
                    data_brokers: <em className={styles.emphasis} />,
```

src/app/(proper_react)/redesign/(public)/LandingView.tsx
```diff
@@ -42,37 +43,79 @@ export const View = (props: Props) => {
...
         </div>
         <div className={styles.heroImage}>
           <HeroImage {...props} />
         </div>
       </header>
 
-      <div className={styles.valuePropositionWrapper}>
+      <section className={styles.quoteWrapper}>
+        <div className={styles.quote}>
+          <h2>
```
Vinnl, line 59: Since this isn't a heading accompanying some content, I think a `<b>` is probably more appropriate than an `<h2>`. (This is relevant because some people navigate pages by first pulling up a list of all heading elements, and then jumping to the relevant part of the page - and this one would just be noise.)

Vinnl (approved):
Great job!

src/app/(proper_react)/redesign/(public)/LandingView.test.tsx
Vinnl: Yep, that sounds good :+1:

src/app/(proper_react)/redesign/(public)/LandingView.tsx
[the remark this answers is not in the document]
Vinnl: You can remove the `en-US/` in any case though, that's always correct. Worst case, people will get redirected to that anyway, but best case, they'll be sent to a locale-appropriate version.

src/app/(proper_react)/redesign/(public)/LandingView.tsx
```diff
@@ -40,6 +48,80 @@ export const View = (props: Props) => {
...
+                )}
+                data-testid="progress-card-image"
+              />
+            ) : (
+              <ScanningForExposuresIllustration {...props} />
+            )}
+          </div>
+        </div>
+
+        <div className={`${styles.item} ${styles.reverseRow}`}>
```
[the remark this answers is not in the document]
Vinnl: Ah OK. Still, this is what it currently looks like for me:

Screen Shot 2023-12-29 at 14 33 31

whereas Figma says this:

image

(Note the second image being _below_ the form.)

PR #3919, opened 2023-12-21
MNTOR-2647 - Quote LP

codemist:
# References: 
Jira: MNTOR-2467
Figma: https://www.figma.com/file/YqpVpUbSTzZMj9Jbp2po8Q/Mozilla-Monitor-Premium-LP?type=design&node-id=1942-12273&mode=design&t=XjY9OW0DT9CcXqox-4

image

# How to test

### DIFF
--- a/locales-pending/landing-all.ftl
+++ b/locales-pending/landing-all.ftl
@@ -18,3 +18,7 @@ landing-all-value-prop-info-at-risk = What info could be at risk?
 landing-all-value-prop-info-at-risk-description = Data leaks are unfortunately part of our digital lives. Your passwords, contact details, financial information, and other personal info can be exposed, putting you at risk of identity theft.
 landing-all-value-prop-scanning-for-exposures-illustration-alt = Scanning for exposures progress bar
 landing-all-value-prop-leaked-password-illustration-alt = Leaked password example breach resolution step
+
+# Quote
+
+landing-all-quote = <data_breaches>Data breaches</data_breaches> happen every 11 minutes, exposing your private information — but don’t worry, we can help.
--- a/locales-pending/landing-premium.ftl
+++ b/locales-pending/landing-premium.ftl
@@ -105,3 +105,7 @@ landing-premium-plans-cards-feature-monitoring = Continuous monitoring
 # Value proposition
 landing-premium-value-prop-info-at-risk-description = Details like your <exposure_type_list>home address, family members’ names, financial info</exposure_type_list> and more can be exposed when a website is hacked — or sold on data broker sites to anyone looking for you. Knowing what info is out there is the first step in protecting yourself.
 landing-premium-value-prop-progress-card-illustration-alt = Progress card delineating exposures that are fixed, in progress or manually fixed 
+
+# Quote
+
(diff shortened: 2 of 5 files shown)


locales-pending/landing-all.ftl
```diff
@@ -18,3 +18,7 @@ landing-all-value-prop-info-at-risk = What info could be at risk?
 landing-all-value-prop-info-at-risk-description = Data leaks are unfortunately part of our digital lives. Your passwords, contact details, financial information, and other personal info can be exposed, putting you at risk of identity theft.
 landing-all-value-prop-scanning-for-exposures-illustration-alt = Scanning for exposures progress bar
 landing-all-value-prop-leaked-password-illustration-alt = Leaked password example breach resolution step
+
+# Quote
```
flodolo, line 22: This is a stand-alone comment. What's the goal of this comment? It should probably be expanded a bit and tied to the following string (no empty line in-between). Same comment applies to the other string.

flodolo (approved):

locales-pending/landing-premium.ftl
```diff
@@ -105,3 +105,7 @@ landing-premium-plans-cards-feature-monitoring = Continuous monitoring
 # Value proposition
 landing-premium-value-prop-info-at-risk-description = Details like your <exposure_type_list>home address, family members’ names, financial info</exposure_type_list> and more can be exposed when a website is hacked — or sold on data broker sites to anyone looking for you. Knowing what info is out there is the first step in protecting yourself.
 landing-premium-value-prop-progress-card-illustration-alt = Progress card delineating exposures that are fixed, in progress or manually fixed 
+
+# Quote
+
+landing-premium-quote = There’s a $240 billion industry of <data_brokers>data brokers</data_brokers> selling your private information for profit. It’s time to take back your privacy.
```
pdehaan, line 111: Nit: Do we want to include a currency here? (ie: $240 billion USD?)

src/app/(proper_react)/redesign/(public)/LandingView.tsx
```diff
@@ -49,7 +49,25 @@ export const View = (props: Props) => {
         </div>
       </header>
 
-      <div className={styles.valuePropositionWrapper}>
+      <section className={styles.quoteWrapper}>
```
Vinnl, line 52: I believe this is already wrapped in `<main>`, [which is an alternative for `<section>` for the main content](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/section#usage_notes) - so I think a `<div>` for a child of `<main>` was fine.

src/app/(proper_react)/redesign/(public)/LandingView.tsx
```diff
@@ -49,7 +49,25 @@ export const View = (props: Props) => {
...
+        <div className={styles.quote}>
+          <h2>
```
Vinnl, line 54: I think that maybe this isn't really a heading, since there's no content underneath it. (e.g. some tools make it easy to list all headings on a page and then to jump to a particular one, but there's not much reason to jump to this one).

Maybe the `div.quote` should just be a `<p>`, and then the `<h2>` a `<b>`?

Vinnl (approved):
Looks good, just two small nits about which HTML tags to use.

PR #3920, opened 2023-12-21
Walkthrough telemetry LP

codemist:
This adds telemetry to the [walkthrough section](https://www.figma.com/file/LGjPMlfA7Li3xkS5YZ6VLY/Attribution-Documentation?type=whiteboard&node-id=156-31790&t=rKb1P8ZJxlf7KwqJ-4) of the landing page. 

### DIFF
--- a/src/app/(proper_react)/redesign/(public)/LandingView.tsx
+++ b/src/app/(proper_react)/redesign/(public)/LandingView.tsx
@@ -54,6 +54,24 @@ export const View = (props: Props) => {
         </div>
       </header>
 
+      <section className={styles.quoteWrapper}>
+        <div className={styles.quote}>
+          <h2>
+            {props.eligibleForPremium
+              ? props.l10n.getFragment("landing-premium-quote", {
+                  elems: {
+                    data_brokers: <span className={styles.emphasis} />,
+                  },
+                })
+              : props.l10n.getFragment("landing-all-quote", {
+                  elems: {
+                    data_breaches: <span className={styles.emphasis} />,
+                  },
+                })}
+          </h2>
+        </div>
+      </section>
+
       <section className={styles.valuePropositionWrapper}>
         <div className={`${styles.item} ${styles.grayBg}`}>
           <span>


src/app/(proper_react)/redesign/(public)/HyperlinkTelemetry.tsx
```diff
@@ -0,0 +1,31 @@
...
+    record("button", "click", {
+      button_id: "privacy_information",
+    });
+  };
+  return l10n.getFragment("landing-all-value-prop-fix-exposures-description", {
+    elems: {
+      privacy_link: (
+        <a
+          onClick={onClick}
+          href="https://www.mozilla.org/en-US/firefox/privacy/"
```
pdehaan, line 25:
```suggestion
          href="https://www.mozilla.org/firefox/privacy/"
```

src/app/(proper_react)/redesign/(public)/SignUpForm.tsx
```diff
@@ -40,7 +48,12 @@ export const SignUpForm = (props: Props) => {
         className={props.isHero ? styles.isHero : ""}
         name={emailInputId}
         id={emailInputId}
-        onChange={(e) => setEmailInput(e.target.value)}
+        onChange={(e) => {
+          setEmailInput(e.target.value);
+          record("field", "focus", {
```
pdehaan, line 53: is "focus" the right event/thing here?
doesn't `onChange` trigger this event for every keystroke?

rhelmer: Hm yeah I think the name is confusing... it sounds like it would only be for `onFocus`, I don't think it matters too much.

GA4 calls any interaction with the form `form_start` and it also has a `form_submit` event https://support.google.com/analytics/answer/9216061?hl=en, we should consider changing the Glean name to either be the same or maybe something like `form_interaction`.

I don't think we should do that in this PR though, @codemist would you mind using it the way you are for now and file a ticket in the backlog to change the name of this later?

rhelmer: @codemist I think the idea is to just generate an event if the user interacts at all with the form, if this does generate a new event on every keypress that's probably way more than we need. 

Is it possible to have this fire when the form is focused, vs. the input field? Do we use auto focus at all now?

rhelmer (changes requested):
lgtm, I would see if we could fire the `field.focus` element on focus if onChange fires every keypress, we only need one of these events per field.

src/app/(proper_react)/redesign/(public)/HyperlinkTelemetry.tsx
Vinnl, line 1: Instead of having to make a new component for every place you want to add a link, you could also create a component that wraps arbitrary links to add click telemetry:

```ts
export const TelemetryLink = ({
  eventData,
  ...props
}: {
  eventData: GleanMetricMap["button"]["click"];
} & HTMLAttributes<HTMLAnchorElement>) => {

  return (
    <a
      {...props}
      onClick={(event) => {
        record("button", "click", eventData);

        props.onClick?.(event);
      }}
    />
  );
};
```

Then, you can replace usages of `<a>` with `<TelemetryLink eventData={{ /* ... */ }}>`.

src/app/(proper_react)/redesign/(public)/LandingView.tsx
```diff
@@ -79,7 +79,20 @@ export const View = (props: Props) => {
...
             <p>
-              <FixExposuresDescription />
+              {props.l10n.getFragment(
+                "landing-all-value-prop-fix-exposures-description",
+                {
+                  elems: {
+                    privacy_link: (
+                      <TelemetryLink
+                        eventData={{ button_id: "privacy_information" }}
+                        href="https://www.mozilla.org/en-US/firefox/privacy/"
```
pdehaan, line 89: Nit: remove the /en-US/ path and let Bedrock handle best possible client languages.
```suggestion
                        href="https://www.mozilla.org/firefox/privacy/"
```

rhelmer (approved):

src/app/(proper_react)/redesign/(public)/TelemetryLink.tsx
```diff
@@ -0,0 +1,34 @@
...
+import { GleanMetricMap } from "../../../../telemetry/generated/_map";
+import { HTMLAttributes } from "react";
+
+// Telemetry link is shown in a fluent getFragment (which does not get rendered in tests)
+/* c8 ignore start */
+export const TelemetryLink = ({
+  eventData,
+  ...props
+}: {
+  eventData: GleanMetricMap["button"]["click"];
```
pdehaan, line 17: I'm a bit late to the party here, but I might be confused.
The file name is TelemetryLink, but this is using "button" as the event.

Do we need separate Telemetry<i>Link</i> and Telemetry<i>Button</i>? Or should the "button" event and `GleanMetricMap["button"]` be configurable?

For example, take these ~three~ four random results for <kbd>git grep -Pn -A2 "record(Telemetry)?\("</kbd>:

```js
record("link", "click", {
   link_id: "returning_user_info_scan_learn_more",
});
--
record("ctaButton", "click", {
  button_id: "intent_to_start_free_scan",
});
--
record("button", "click", {
  button_id: "edit_free_scan",
});
--
record("upgradeIntent", "click", {
  button_id: isMonthly
    ? "intent_to_purchase_monthly_plan_nav_modal"
```

PR #3921, opened 2023-12-22
MNTOR-2517 - automatically opt-out any monitoring scans

rhelmer:
# References: 
Jira: MNTOR-2517

# Description

Simple change but this is pretty hard to test, I've been doing it by modifying the database manually but I think we need to mock out the external server to automate this and to really test that opt-out is called appropriately.

# How to test

TODO


mansaj:
@rhelmer Ok, so to understand this.. This is just for testing right? Is this to be reverted once testing is done?

src/app/api/v1/onerep-events/route.ts
```diff
@@ -32,49 +32,35 @@ interface OnerepWebhookRequest {
...
-
   try {
     const result: OnerepWebhookRequest = JSON.parse(finalBuffer.toString());
     logger.debug("OneRep Webhook Request received:", result);
 
     if (result.type !== "scan.completed") {
-      logger.debug("Unexpected OneRep webhook type received:", result.type);
-      return;
+      logger.info("ignored_onerep_webhook_type", result.type);
+      return NextResponse.json({ success: false }, { status: 202 });
```
pdehaan, line 58: Do these error responses need messages?
Probably solid in the "nit" category, but not sure if it'd make it easier to debug L58 errors vs L63 errors since they both have a `success:false`/`status:202` responses.

src/app/api/v1/onerep-events/route.ts
```diff
@@ -97,9 +86,27 @@ export async function POST(req: NextRequest) {
 
     return NextResponse.json({ success: true }, { status: 200 });
   } catch (ex) {
-    logger.error(ex);
+    logger.error("onerep_webhook_exception", ex);
     captureException(ex);
 
     return NextResponse.json({ success: false }, { status: 500 });
```
pdehaan, line 92: Possibly same nit here w/ `success:false`/`status:500` on L92 vs L50.

src/app/api/v1/onerep-events/route.ts
```diff
@@ -97,9 +86,27 @@ export async function POST(req: NextRequest) {
...
+  if (!process.env.ONEREP_WEBHOOK_SECRET) {
+    throw new Error("env var ONEREP_WEBHOOK_SECRET must be set");
+  }
+
+  const expectedSignature = crypto
+    .createHmac("sha256", process.env.ONEREP_WEBHOOK_SECRET)
+    .update(buffer)
+    .digest("hex");
+
+  console.debug("test:", signature, expectedSignature);
```
pdehaan, line 109: ✂️

src/app/api/v1/onerep-events/route.ts
```diff
@@ -32,49 +32,35 @@ interface OnerepWebhookRequest {
...
   }
-
   try {
     const result: OnerepWebhookRequest = JSON.parse(finalBuffer.toString());
     logger.debug("OneRep Webhook Request received:", result);
 
     if (result.type !== "scan.completed") {
-      logger.debug("Unexpected OneRep webhook type received:", result.type);
-      return;
+      logger.info("ignored_onerep_webhook_type", result.type);
```
pdehaan, line 57: Any differences between `logger.debug()` and `logger.info()`?

rhelmer: Currently we don't log `debug` level messages in GCP, just for local development. Also FWIW the reason I'm changing the message name is because it's easier to handle these in BigQuery if they don't have spaces.

src/app/api/v1/onerep-events/route.ts
rhelmer: Good point... I would like to leave the `{ success }` responses in for now just for consistency, I agree these should be `true`. I think returning the `202` might be overkill, the caller probably won't ever care if we don't process the response.

rhelmer:
> @rhelmer Ok, so to understand this.. This is just for testing right? Is this to be reverted once testing is done?

@mansaj I have more unit tests in place now (and split up the code a bit to make it easier to test), but I think the fix i the same, we need to call `optOut` every time we see `monitoring` status on an incoming scan.

I considered doing this from `refreshStoredScanResults` in order to fix any existing users and in case we miss any webhooks, but I'm not sure it's safe to call `optOut` too often. I think we could do it from here if we had a way to check to see if it had already been run in response to an incoming `monitoring` scan. I think this would be a good enhancement but I'd like to do it as a followup to this, I feel like the webhook is the most urgent.

Another followup is that we may want to treat `monitoring` scans differently in the UI - they will get auto-fixed if currently subscribed, but they won't if the current customer isn't a subscriber. So they should probably just be moved to the `Fixed` tab automatically if the user is currently a subscriber, but show up in `Action needed` if their subscription has lapsed.

mansaj (approved):

PR #3907, opened 2023-12-19
MNTOR-2342: dashboard telemetry

mansaj:
# References: 
Jira: MNTOR-2342
Figma: https://www.figma.com/file/LGjPMlfA7Li3xkS5YZ6VLY/Attribution-Documentation?type=whiteboard&node-id=0%3A1&t=mk5Mp4FZG6MjWMaG-1

# Description
Part 1 of dashboard telemetry. This PR should include everything except the permutations (dashboard Top Banner CTA)

# Screenshot (if applicable)

Not applicable.

# How to test

# Checklist (Definition of Done)
- [x] Commits in this PR are minimal and [have descriptive commit messages](https://chris.beams.io/posts/git-commit/).
- [x] I've added or updated the relevant sections in readme and/or code comments
- [x] I've added a unit test to test for potential regressions of this bug.
- [x] All acceptance criteria are met.
- [x] Jira ticket has been updated (if needed) to match changes made during the development process.
- [x] Jira ticket has been updated (if needed) with suggestions for QA when this PR is deployed to stage.

PR #3922, opened 2023-12-22
Disallow loose equality operators

Vinnl:
As [observed](https://github.com/mozilla/blurts-server/pull/3907#discussion_r1434448403) by @pdehaan, apparently [eqeqeq](https://eslint.org/docs/latest/rules/eqeqeq) is [not in ESLint's recommended ruleset](https://eslint.org/docs/latest/rules/). I'm guessing it's because it might overwhelm folks, or because people still often rely on it, but there's not really a good reason to not check for the expected types explicitly.


flozia (approved):
Great! Thanks for adding this rule, and also thanks @pdehaan.

PR #3934, opened 2023-12-28
MNTOR-1833/mozilla monitor rebrand

rhelmer:
# References: 
Jira: MNTOR-1833
Figma: 

# Description

# Screenshot (if applicable)

Not applicable.

# How to test

# Checklist (Definition of Done)
- [x] Localization strings (if needed) have been added.
- [x] Commits in this PR are minimal and [have descriptive commit messages](https://chris.beams.io/posts/git-commit/).
- [ ] I've added or updated the relevant sections in readme and/or code comments
- [x] I've added a unit test to test for potential regressions of this bug.
- [x] Product Owner accepted the User Story (demo of functionality completed) or waived the privilege.
- [x] All acceptance criteria are met.
- [x] Jira ticket has been updated (if needed) to match changes made during the development process.
- [ ] Jira ticket has been updated (if needed) with suggestions for QA when this PR is deployed to stage.

PR #3936, opened 2023-12-28
Port settings page to proper React

Vinnl:
# References: 
Jira: MNTOR-1366
Figma: https://www.figma.com/file/CaEKIhvSJqf6KNIMzSkt40/Concepts-for-Monitor-MVP-Redesign?type=design&node-id=8251-5299&mode=dev (only used as inspiration - functionality is based on the current page)

# Description

(Sorry, this is a chonker, and it spans both frontend and backend, so I added everyone as reviewer...)

This ports the settings page to proper React. One major callout is that I have used [Server Actions](https://nextjs.org/docs/app/building-your-application/data-fetching/server-actions-and-mutations) for adding and removing email addresses. This has two main advantages:

- Next,js and React are trying to make this the idiomatic approach, which means that...
- it integrates well with the server-side rendering we do, giving us [a way to update the server-rendered data](https://nextjs.org/docs/app/building-your-application/data-fetching/server-actions-and-mutations#revalidating-data) without having to do a full page reload.

I'm not sure how I would have been able to do the latter without switching to fetching the data client-side and adopting something like [SWR](https://swr.vercel.app/), which we use on Relay.

However, there are also a couple of downsides:

- The biggest one: it was only [released recently](https://nextjs.org/blog/next-14#server-actions-stable) (Oct 26th). Thus, patterns are still solidifying, and there might still be bugs. Such as...
- ~~The call to `revalidatePath` after adding an email doesn't seem to do anything at the moment, and hence, the list of email addresses is not updated. This might be fixed in future Next.js releases.~~ This was a stupid mistake of mine, fixed in 2ddedd13d.
- Server Actions should run on the server, and thus Storybook (which runs client-side) trips over it. They [plan to provide ergonomic module mocking in a future version of Storybook](https://storybook.js.org/blog/storybook-react-server-components/#mocked-and-loaded) which should hopefully allow us to isolate `actions.ts`, but that doesn't buy us anything today. Which means that:
- I can't use a story to write tests.

To solve the latter, I'm just rendering the component directly, and I added a `<TestComponentWrapper>` to do what we used to do in Storybooks `preview.tsx`. (I'm not sure about the file location of that component, so feel free to suggest another place for it.)

# Screenshot (if applicable)

image

# How to test

You should be able to reach the settings page via the menu (user menu and main menu - also on mobile). Everything on that page should work and have the same functionality as the existing page.


locales/en/add-email.ftl
```diff
@@ -19,4 +19,8 @@ add-email-address-input-label = Email address
 add-email-send-verification-button = Send verification link
 
 # $email is the newly added email address. $settings-href is the URL for the Settings page. HTML tags should not be translated, e.g. `<a>`
+# This string will be deprecated when the new Plus plan is live.
 add-email-verify-the-link = Verify the link sent to { $email } to add it to { -brand-fx-monitor }. Manage all email addresses in <a { $settings-href }>Settings</a>.
+# Variables:
+#   $email (string) - An email address submitted by the user for monitoring, e.g. `example@example.com`
+add-email-verify-the-link-2 = Verify the link sent to <b>{ $email }</b> to add it to { -brand-fx-monitor }.
```
flodolo, line 26:
```suggestion
add-email-verify-the-link-2 = Verify the link sent to <b>{ $email }</b> to add it to { -brand-mozilla-monitor }.
```

Since this string is new, it might make sense to start using the new term? Obviously, #3934 would be a blocker for this, or you'd need to add the term here.

With that said, not a blocker, we can use the string with the previous term.

flodolo (approved):

locales/en/add-email.ftl
Vinnl: That works for me, I added it in 734e350c5. Since this won't be visible until the redesign launches and hence #3934 is merged, I'm OK with it being broken for a bit.

Also, a heads-up that I apparently hadn't hit save, so I just added two more strings in 1c74aaeb6.

Vinnl: Although I guess [this comment](https://github.com/mozilla/blurts-server/pull/3934#discussion_r1437433081) implies that maybe I should be using `fx-monitor` still?

src/app/(proper_react)/redesign/(authenticated)/user/(dashboard)/settings/View.tsx
```diff
@@ -0,0 +1,99 @@
...
+            })}
+          </ul>
+          <EmailAddressAdder />
+          <hr />
+          <AlertAddressForm
+            defaultSelected={
+              props.user.subscriber?.all_emails_to_primary
+                ? "primary"
+                : "affected"
+            }
```
flozia, line 76: Not sure if this is only an issue with my local setup: After reloading the page the `primary` email option is always selected by default. Looking into my local database `all_email_to_primary` always is set to `false`.

.storybook/preview.tsx
```diff
@@ -30,19 +27,7 @@ const AppDecorator: Preview["decorators"] = (storyFn) => {
...
-          PUBLIC_APP_ENV: "storybook",
-        }}
-      >
-        <SessionProvider session={null}>
-          <ReactAriaI18nProvider locale="en">{storyFn()}</ReactAriaI18nProvider>
-        </SessionProvider>
-      </PublicEnvProvider>
-    </L10nProvider>
-  );
+  return <TestComponentWrapper>{storyFn()}</TestComponentWrapper>;
```
flozia, line 30: Praise: Nice cleanup!

src/app/(proper_react)/redesign/(authenticated)/user/(dashboard)/settings/actions.ts
```diff
@@ -0,0 +1,153 @@
...
+      `Tried to delete email [${email.id}] without an active session.`,
+    );
+    return;
+  }
+  const subscriber = (await getSubscriberByEmail(
+    session.user.email,
+  )) as SubscriberRow | null;
+  if (email.subscriber_id !== subscriber?.id) {
+    logger.error(
+      `Subscriber [${subscriber?.id}] tried to delete email [${email.id}], which belongs to another subscriber.`,
```
flozia, line 146: Question: We are only logging this on the server so it does not matter as much, but wondering in general if it might be preferable to try and not log conclusions that relate to other user info and rather say: “…, which does not belong to the current subscriber.”?

rhelmer (approved):
lgtm overall - I believe the old settings page uses tooltips too, do we plan to have that or some other kind of indication that the change worked?

src/app/(proper_react)/redesign/(authenticated)/user/(dashboard)/settings/AlertAddressForm.tsx
```diff
@@ -0,0 +1,114 @@
...
+
+export type Props = {
+  defaultSelected: AlertAddress;
+};
+
+const AlertAddressContext = createContext<RadioGroupState | null>(null);
+
+export const AlertAddressForm = (props: Props) => {
+  const l10n = useL10n();
+  const state = useRadioGroupState({
```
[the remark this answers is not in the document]
Vinnl: It would make sense, but I think we probably want a completely new design here that don't involve form elements that do things without submitting a form. Since I'm mostly aiming for feature parity here to unblock the release, I think it's best to iterate on the design in a later stage.

src/app/(proper_react)/redesign/(authenticated)/user/(dashboard)/settings/View.tsx
Vinnl: Oh that's a very good catch. So besides setting the wrong value (fixed in c271c1623), I also read the existing pref from the session token - which didn't get refreshed after the save. Fixed in dbb73f5f1.

src/app/(proper_react)/redesign/(authenticated)/user/(dashboard)/settings/actions.ts
Vinnl: Yes, I considered that too (and it is relevant on the server too, since the people who can access logs are not the same as the people who can access the database). In this case, however, I figured that this attempt is likely to be malicious, and thus warrants the increased detail? /cc @rhelmer

Vinnl:
@codemist 

> We should probably document somewhere when dictating the use of TestComponentWrapper vs using Storybook components, since it's a new pattern.

That's an excellent callout. I added some docs in 9c42ac092; let me know if that's in any way unclear.

@flozia!

> The all_emails_to_primary column in the subscribers table does not get updated in my local database.

Does it not get updated, or did it simply get updated to the value it already had? Looks like I mixed up the API parameters... (I blame the undescriptive `"0"` and `"1"`...) c271c1623

Well spotted by the way!

> the radio inputs are cut off and the delete icon shrinks

Ah yes, nice catch: ad9f22b2e

@rhelmer 

> lgtm overall - I believe the old settings page uses tooltips too, do we plan to have that or some other kind of indication that the change worked?

I agree that feedback on an action would be good (radio buttons aren't supposed to do anything until a form submits anyway - toggle buttons would probably be more appropriate), but as far as I can see, the current version doesn't give feedback either?

[Screencast from 2024-01-03 16-37-01.webm](https://github.com/mozilla/blurts-server/assets/4251/aee12932-4e29-47a2-82c3-b61e3d68345f)

src/app/(proper_react)/redesign/(authenticated)/user/(dashboard)/settings/AlertAddressForm.tsx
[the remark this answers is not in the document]
flozia: Getting up to feature parity sounds good to me.

src/app/api/v1/user/update-comm-option/route.ts
```diff
@@ -14,7 +14,7 @@ import {
 } from "../../../../../db/tables/subscribers";
 
 export interface EmailUpdateCommOptionRequest {
-  /** "0" to send breach alerts to the primary email address, "1" to send them to the affected address */
+  /** "1" to send breach alerts to the primary email address, "0" to send them to the affected address */
```
flozia, line 17: Dang — that’s indeed very easy to confuse.

src/app/api/utils/auth.ts
```diff
@@ -67,6 +67,14 @@ export const authOptions: AuthOptions = {
       if (trigger === "update") {
         // Refresh the user data from FxA, in case e.g. new subscriptions got added:
         profile = await fetchUserInfo(token.subscriber?.fxa_access_token ?? "");
+        if (token.email) {
+          const updatedSubscriberData = await getSubscriberByEmail(token.email);
+          // MNTOR-2599 The breach_resolution object can get pretty big,
+          // causing the session token cookie to balloon in size,
```
flozia, line 73: 🎈

flozia (approved):
Looks great and works as expected!

PR #3960, opened 2024-01-04
Also suppress non-error logs in tests

Vinnl:
This makes it easier to debug issues when they do arise.


mansaj (approved):
Thanks, I'll create a PR to take out the rest of spies for console log

PR #3962, opened 2024-01-04
Respect the SUPPORTED_LOCALES env var

Vinnl:
This means we won't localise the UI to locales that aren't listed in SUPPORTED_LOCALES. This avoids showing languages with only a few localised strings.

I'm not sure if we decided not to use this for now, but it was so easy to implement that I figured I'd just submit a PR and we'd see.

### DIFF
--- a/src/app/functions/server/l10n.ts
+++ b/src/app/functions/server/l10n.ts
@@ -80,13 +80,14 @@ export function getL10nBundles(): LocaleData[] {
 
   const languages = acceptLangHeader ? acceptedLanguages(acceptLangHeader) : [];
   const supportedLocales = process.env.SUPPORTED_LOCALES?.split(",");
-  const currentLocales = negotiateLanguages(
-    languages,
-    Object.keys(bundleSources).filter(
-      (locale) => supportedLocales?.includes(locale),
-    ),
-    { defaultLocale: "en" },
-  );
+  const availableLocales = Object.keys(bundleSources);
+  const filteredLocales =
+    process.env.APP_ENV === "heroku"
+      ? availableLocales
+      : availableLocales.filter((locale) => supportedLocales?.includes(locale));
+  const currentLocales = negotiateLanguages(languages, filteredLocales, {
+    defaultLocale: "en",
+  });
 
   const relevantBundleSources = currentLocales.map((relevantLocale) => ({
     locale: relevantLocale,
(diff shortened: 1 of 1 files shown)


flodolo (approved):

flodolo:
One consideration though: we want the ability to filter locales in production, but in a staging environment we need all locales, independently of their completion level. Can we still achieve this here?

Vinnl:
@flodolo With this change, we can achieve that by adding them all to the `SUPPORTED_LOCALES` environment variable for that environment specifically. Alternatively, I could modify the code to only filter the list by `SUPPORTED_LOCALES` in production.

flodolo:
> @flodolo With this change, we can achieve that by adding them all to the `SUPPORTED_LOCALES` environment variable for that environment specifically. Alternatively, I could modify the code to only filter the list by `SUPPORTED_LOCALES` in production. Whichever you prefer.

Ideally, we want to be able to pick up locales without making changes on DEV (which currently is https://fx-breach-alerts.herokuapp.com/), so not a list that needs to be manually maintained by someone.

DEV = all locales
PROD = specific locales (this list)

I suppose this could be achieved by either assigning a specific value to the list ("all"), or by having a separate ENV variable that identifies the server as DEV?

Vinnl:
@flodolo We have a specific env var that identifies dev; I added dda3fa364 to only apply the filter on non-dev environments.

flodolo (approved):

src/app/functions/server/l10n.ts
```diff
@@ -78,11 +79,15 @@ export function getL10nBundles(): LocaleData[] {
...
-  );
+  const supportedLocales = process.env.SUPPORTED_LOCALES?.split(",");
+  const availableLocales = Object.keys(bundleSources);
+  const filteredLocales =
+    process.env.APP_ENV === "heroku"
+      ? availableLocales
+      : availableLocales.filter((locale) => supportedLocales?.includes(locale));
+  const currentLocales = negotiateLanguages(languages, filteredLocales, {
+    defaultLocale: "en",
+  });
```
pdehaan, line 90: Not sure if it'd be possible to add `filteredLocales` or `currentLocales` to the Dockerflow <https://monitor.mozilla.org/__version__> endpoint as well so we could confirm that the expected locales are supported in the different environments; although that's probably lower on the priority than a bunch of other backlog bugs.

rhelmer (approved):

src/app/functions/server/l10n.ts
Vinnl: Yeah that would be nice, but we'll first have to figure out how to make that endpoint work in the first place - it's currently always showing the same static value :cry:

PR #3963, opened 2024-01-04
MNTOR-2650 - FAQ Section

codemist:
# References: 
Jira: MNTOR-2650
Figma: 

image

### DIFF
--- a/src/app/(proper_react)/redesign/(public)/Faq.tsx
+++ b/src/app/(proper_react)/redesign/(public)/Faq.tsx
@@ -26,7 +26,6 @@ const FaqItem = (props: FaqItemProps) => {
   const { buttonProps } = useButton(
     {
       onPress: props.onExpandAnswer,
-      elementType: "dt",
     },
     buttonRef,
   );


.env-dist
```diff
@@ -166,3 +166,6 @@ NIMBUS_SIDECAR_URL=http://localhost:8001
 
 # The maximum number of jobs that the email breach alert worker will process.
 EMAIL_BREACH_ALERT_MAX_MESSAGES = 10000
+
+NEXT_PUBLIC_LEARN_MORE_ABOUT_MONITOR_PLUS_URL=https://support.mozilla.org/kb/how-does-monitor-plus-work
+NEXT_PUBLIC_FAQ_MONITOR=https://support.mozilla.org/en-US/kb/firefox-monitor-faq
```
pdehaan, line 171:
```suggestion
NEXT_PUBLIC_FAQ_MONITOR=https://support.mozilla.org/kb/firefox-monitor-faq
```

src/app/(proper_react)/redesign/(public)/Faq.tsx
```diff
@@ -0,0 +1,131 @@
...
+      <b className={styles.faqTitle}>
+        {l10n.getString("landing-all-faq-title")}
+      </b>
+      <Link
+        className={styles.faqCta}
+        href={process.env.NEXT_PUBLIC_FAQ_MONITOR as string}
+        target="_blank"
+      >
+        {l10n.getString("landing-all-faq-see-all")}
+      </Link>
```
pdehaan, line 61: I thought [Next.js `<Link>`](https://nextjs.org/docs/pages/api-reference/components/link) was mainly for internal routing and not external links.
https://upmostly.com/nextjs/how-to-link-to-an-external-page-in-nextjs

src/app/(proper_react)/redesign/(public)/Faq.tsx
```diff
@@ -0,0 +1,140 @@
...
+};
+
+const FaqItem = (props: FaqItemProps) => {
+  const { question, answer, isExpanded, onExpandAnswer } = props;
+
+  return (
+    <>
+      <dt className={styles.faqQuestion} onClick={onExpandAnswer}>
+        {question}
+        <CloseBigIcon alt="" className={`${isExpanded && styles.expanded}`} />
```
Vinnl, line 27: It's currently impossible to bring focus to this element (and thus expand/collapse the entry) with the keyboard. [In Relay](https://github.com/mozilla/fx-private-relay/blob/6b0bcf7f17af81670a208160f1ed8645e15f2092/frontend/src/components/landing/FaqAccordion.tsx#L64) we fixed that by wrapping it in a button. If you don't like the extra wrapping, you could also use react-aria's `useButton` hook to turn `<dt>` into a button.

Vinnl: Oh, and one thing I see we didn't do on Relay, but which would be good, is to add an [`aria-expanded`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-expanded) and an `aria-controls` on the `<dt>`/`<button>`.

And since the docs say

> The accessible name of the controlling object should reflect this change.

I think that the most sensible way to do this is probably by adding an alt text to `<CloseBigIcon>` , something like `(Open)` / `(Closed)`, I suppose?

src/app/(proper_react)/redesign/(public)/Faq.tsx
```diff
@@ -0,0 +1,140 @@
...
+                },
+              },
+            )}
+            isExpanded={expandedQuestion === "premium-continuous-data-removal"}
+            onExpandAnswer={() =>
+              handleExpandAnswer("premium-continuous-data-removal")
+            }
+          />
+        )}
+        {/* Ignore the remaining tests as testing the first two FAQ questions are sufficient */}
```
Vinnl, line 116: Thanks for adding a comment, that makes sense!

src/app/(proper_react)/redesign/(public)/Faq.tsx
Vinnl: The `<dl>`  is great though, btw!

src/app/(proper_react)/redesign/(public)/LandingView.test.tsx
```diff
@@ -388,4 +388,55 @@ describe("When Premium is available", () => {
...
+      "What kinds of websites sell my personal information?",
+    );
+    await user.click(faqQuestion);
+    const faqAnswer = screen.getByText(
+      "Certain websites are in the business of collecting and selling people’s personal information without their consent, which is unfortunately legal in the US.",
+      { exact: false },
+    );
+    expect(faqAnswer).toHaveClass("expanded");
+    await user.click(faqQuestion);
+    expect(faqAnswer).not.toHaveClass("expanded");
```
Vinnl, line 406: After adding `[aria-expanded]`, it would be even nicer to check for that using `toHaveAttribute()`, to ensure that doesn't regress.

(Though I think that would be on `faqQuestion`, so you might still also want to add a `toBeInTheDocument()` for `faqAnswer`.)

src/app/(proper_react)/redesign/(public)/LandingView.test.tsx
```diff
@@ -388,4 +388,55 @@ describe("When Premium is available", () => {
     const progressCardIllustration = screen.getByTestId("progress-card-image");
     expect(progressCardIllustration).toBeInTheDocument();
   });
+
+  it("opens and closes an FAQ accordion item", async () => {
+    const user = userEvent.setup();
+    const ComposedDashboard = composeStory(LandingUs, Meta);
+    render(<ComposedDashboard />);
+    const faqQuestion = screen.getByText(
```
Vinnl, line 396: And this would be a `getByRole("button"` with the other changes.

Vinnl (commented):
It's looking good so far, but needs a couple of accessibility fixes - most importantly, the questions need to be focusable (and have focus styling, but I expect that that will come for free once they are focusable).

src/app/(proper_react)/redesign/(public)/Faq.tsx
pdehaan: Or native collapse/expand using `<details>` and `<summary>` tags, although it might tweak the UI slightly.

```html
<details>
  <summary><h3>What kinds of websites sell my personal information?</h3></summary>
  <p>Certain websites are in the business of collecting and selling people’s personal information without their consent, which is unfortunately legal in the US. These sites are called data brokers and they make up a $240 billion dollar industry. They use sophisticated methods to collect personal, financial, location, and even health information, often without your consent or even your knowledge. They’ll sell what they’ve collected to third parties, profiting from your information and leaving you open to violations of your privacy and security.</p>
</details>
```

<details>
  <summary><h3>What kinds of websites sell my personal information?</h3></summary>
  <p>Certain websites are in the business of collecting and selling people’s personal information without their consent, which is unfortunately legal in the US. These sites are called data brokers and they make up a $240 billion dollar industry. They use sophisticated methods to collect personal, financial, location, and even health information, often without your consent or even your knowledge. They’ll sell what they’ve collected to third parties, profiting from your information and leaving you open to violations of your privacy and security.</p>
</details>

.env-dist
pdehaan: Semi-related (but probably outside the scope of this PR), but it looks like we have a couple other `/en-US/` references in `/src/app/(proper_react)/redesign/(public)/LandingView.tsx`

```sh
git grep -in "/en-us/" -- ":(exclude)*.md" | grep -Ev "\s+(//|\*)\s+"

.env-dist:171:NEXT_PUBLIC_FAQ_MONITOR=https://support.mozilla.org/en-US/kb/firefox-monitor-faq
src/app/(proper_react)/redesign/(public)/LandingView.tsx:97:                            href="https://www.mozilla.org/en-US/firefox/privacy/"
src/app/(proper_react)/redesign/(public)/LandingView.tsx:111:                            href="https://www.mozilla.org/en-US/firefox/privacy/"
```

https://github.com/mozilla/blurts-server/blob/c5f95f8c0b6967fe418ba16801c7f16cd3311ba1/src/app/(proper_react)/redesign/(public)/LandingView.tsx#L95-L99

https://github.com/mozilla/blurts-server/blob/c5f95f8c0b6967fe418ba16801c7f16cd3311ba1/src/app/(proper_react)/redesign/(public)/LandingView.tsx#L109-L113

src/app/(proper_react)/redesign/(public)/Faq.tsx
```diff
@@ -0,0 +1,155 @@
...
+  return (
+    <>
+      <dt
+        {...buttonProps}
+        {...focusProps}
+        tabIndex={0}
+        className={styles.faqQuestion}
+      >
+        {question}
+        <CloseBigIcon alt="" className={`${isExpanded && styles.expanded}`} />
```
pdehaan, line 42: Q: Wouldn't the close icon need some `alt=` for a11y?

src/app/(proper_react)/redesign/(public)/Faq.tsx
```diff
@@ -0,0 +1,169 @@
...
+
+  return (
+    <>
+      <dt
+        {...buttonProps}
+        {...focusProps}
+        tabIndex={0}
+        className={styles.faqQuestion}
+        aria-expanded={props.isExpanded}
+        aria-controls={`faq-${props.id}`}
```
Vinnl, line 42: I believe this needs to point to another element (in this case, the related `<dd>`) that has the given value as its `id`. If you want to avoid needing to come up with IDs yourself, you can also use React's [`useId`](https://react.dev/reference/react/useId) hook.

src/app/(proper_react)/redesign/(public)/LandingView.test.tsx
Vinnl: Ah, yeah, if you want to do a partial match on text so you can avoid the SVG, you can pass in a regular expression instead of a string. It doesn't need to be a complex regexp; you can still just match on exact characters, e.g. like this:

```ts
    const faqQuestion = screen.getByRole("button",
      { name: /What kinds of websites sell my personal information?/ },
    );
```

And yes, definitely a good idea to verify that `aria-hidden` is set correctly!

Vinnl (commented):
Almost there - just a couple more adjustments and then we're good to go, I think.

src/app/(proper_react)/redesign/(public)/Faq.tsx
Vinnl: I think both are true: you need `aria-expanded` to indicate whether the item that is controlled by _this_ item is visible or not, and `aria-controls` to indicate which item that is. So to adapt an example from MDN:

```html
<button
  aria-expanded="false"
  aria-controls="the-desc">
  Expand
</button>
<p id="the-desc" hidden>
  This content is current hidden
</p>
```

So the `<p>` has an ID, and the `<button>` has both `aria-expanded`, and an `aria-controls` that points to the `<p>`'s ID.

src/app/(proper_react)/redesign/(public)/Faq.tsx
```diff
@@ -0,0 +1,171 @@
...
+};
+
+const FaqItem = (props: FaqItemProps) => {
+  const l10n = useL10n();
+  const buttonRef = useRef(null);
+
+  const { buttonProps } = useButton(
+    {
+      onPress: props.onExpandAnswer,
+      elementType: "dt",
```
Vinnl, line 29: Now you don't need the `elementType` anymore :grin: 

```suggestion
```

Vinnl (approved):
Nice!

PR #3972, opened 2024-01-05
Allow loading Storybook in other locales

Vinnl:
@flodolo With this change, you'll be able to add `&locale=<chosen locale>` to the Storybook URL to load that locale's strings - [example](https://deploy-preview-3972--fx-monitor-storybook.netlify.app/?path=/story/pages-public-landing-page--landing-non-us&locale=cs). Non-submitted English strings will also be loaded.

### DIFF
--- a/src/app/functions/server/mockL10n.ts
+++ b/src/app/functions/server/mockL10n.ts
@@ -17,22 +17,29 @@ import type { LocaleData } from "./l10n";
 // Code in this file is only used in tests and Storybook, not in production:
 /* c8 ignore start */
 
-export function getEnL10nBundlesSync(): LocaleData[] {
+export function getOneL10nBundleSync(locale = detectLocale()): LocaleData[] {
   return process.env.STORYBOOK === "true"
-    ? getEnL10nBundlesInWebpackContext()
-    : getEnL10nBundlesInNodeContext();
+    ? getOneL10nBundleInWebpackContext(locale)
+    : getOneL10nBundleInNodeContext(locale);
 }
 
-export function getEnL10nBundlesInWebpackContext(): LocaleData[] {
-  const referenceStringsContext: { keys: () => string[] } & ((
+export function getOneL10nBundleInWebpackContext(
+  locale = detectLocale(),
+): LocaleData[] {
+  const allStringsContext: { keys: () => string[] } & ((
     path: string,
     // `require` isn't usually valid JS, so skip type checking for that:
     // eslint-disable-next-line @typescript-eslint/no-explicit-any
   ) => string) = (require as any).context(
-    "../../../../locales/en",
+    "../../../../locales/",
     true,
     /\.ftl$/,
   );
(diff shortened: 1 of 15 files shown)


flodolo:
I tried locally but it doesn't seem to work (nor your link above)? I always get English content.

How hard would it be to actually detect the locale via language negotiation, like we do on the actual website?

flodolo:
Correction: something works, but not as it should

If you check Italian, it's fully localized at this point, but a lot of content remains in English
https://deploy-preview-3972--fx-monitor-storybook.netlify.app/?path=/story/pages-public-landing-page--landing-non-us&locale=it

Vinnl:
Ah thanks for checking, I didn't know what was supposed to be localised. Is it better now?

Locale detection would be a bit more of a hassle unfortunately.

flodolo:
> Ah thanks for checking, I didn't know what was supposed to be localised. Is it better now?

Yes, definitely better.

> Locale detection would be a bit more of a hassle unfortunately.

I'd be happy to have that considered as a follow-up.

The challenge is explaining to non-technical folks how to translate this. With locale detection, I could just provide one link.

Vinnl:
> The challenge is explaining to non-technical folks how to translate this. With locale detection, I could just provide one link.

Yeah, I get that. As a kludge workaround, I do see that Storybook preserves the query parameter as you navigate around, so we could e.g. create a wiki page with links to the Storybook front page for every locale?

flodolo:
> Yeah, I get that. As a kludge workaround, I do see that Storybook preserves the query parameter as you navigate around, so we could e.g. create a wiki page with links to the Storybook front page for every locale?

I'm afraid most people will still miss the instructions :-( 

Out of curiosity, I don't understand enough of JS at this level, but I see functions like `getLocale(getL10nBundles())` in code. Are they completely useless for `detectLocale()`?

Vinnl:
No, I'm fairly sure it would be possible to detect the locale, it's just a bit more work, because unfortunately for Storybook, I can't dynamically get a list of available languages at runtime, so I'll have to manually parse the list of locale directories and then find a way to get that to the language selection at the right time. So it's mostly me trying to limit the scope for this for now.

flodolo:
Having this as a GET parameter works well enough for now (also considering it persists when clicking around). As mentioned, it would be a nice to have enhancement to detect the locale, but that can wait.

src/app/functions/server/mockL10n.ts
```diff
@@ -42,10 +49,10 @@ export function getEnL10nBundlesInWebpackContext(): LocaleData[] {
     true,
     /\.ftl$/,
   );
-  const referenceSourceFilenames = referenceStringsContext.keys();
+  // const referenceSourceFilenames = referenceStringsContext.keys();
```
flozia, line 52:
```suggestion
```

flozia (approved):
Great addition!

PR #3990, opened 2024-01-08
Add a "sign in" button to the landing page

Vinnl:
# References: 
Jira: MNTOR-3698
Figma: N/A

### DIFF
--- a/src/app/(proper_react)/redesign/(public)/LandingView.test.tsx
+++ b/src/app/(proper_react)/redesign/(public)/LandingView.test.tsx
@@ -16,6 +16,7 @@ import {
 import { userEvent } from "@testing-library/user-event";
 import { axe } from "jest-axe";
 import { signIn, useSession } from "next-auth/react";
+import { useTelemetry } from "../../../hooks/useTelemetry";
 import Meta, {
   LandingNonUs,
   LandingNonUsDe,
@@ -107,6 +108,27 @@ describe("When Premium is not available", () => {
 
     expect(signInButton).toBeInTheDocument();
     expect(signIn).toHaveBeenCalledTimes(1);
+  });
+
+  it("counts the number of clicks on the sign-in button at the top", async () => {
+    const mockedRecord = useTelemetry();
+    const ComposedDashboard = composeStory(LandingNonUs, Meta);
+    render(<ComposedDashboard />);
+
+    const user = userEvent.setup();
+
+    const signInButton = screen.getByRole("button", {
+      name: "Sign In",
+    });
+    await user.click(signInButton);
+
+    expect(mockedRecord).toHaveBeenCalledWith(
+      "ctaButton",
+      "click",
+      expect.objectContaining({
+        button_id: "sign_in",
+      }),
+    );
   });
 
   it("shows the data breaches quote", () => {
--- a/src/app/(proper_react)/redesign/(public)/LandingView.test.tsx
+++ b/src/app/(proper_react)/redesign/(public)/LandingView.test.tsx
@@ -71,7 +71,7 @@ describe("When Premium is not available", () => {
     });
   });
 
-  it("shows a link to the dashboard in the header if the user is signed in", () => {
+  it("does not show a 'Sign In' button in the header if the user is signed in", () => {
     const mockedUseSession = useSession as jest.Mock<
       ReturnType<typeof useSession>,
       Parameters<typeof useSession>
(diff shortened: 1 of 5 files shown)


src/app/(proper_react)/redesign/(public)/LandingView.test.tsx
```diff
@@ -24,9 +24,24 @@ import Meta, {
...
+      return {};
+    }),
+  };
+});
 jest.mock("../../../hooks/useTelemetry");
 
+beforeEach(() => {
+  // For reasons that are unclear to me, the mock implementation defind in the
+  // call to `jest.mock` above forgets the implementation. I've spent way too
+  // long debugging that already, so I'm settling for this :(
```
Vinnl, line 40: :cry:

src/app/(proper_react)/redesign/(public)/LandingView.test.tsx
```diff
@@ -24,9 +24,24 @@ import Meta, {
...
+  const mockedUseSession = useSession as jest.Mock;
+  mockedUseSession.mockReturnValue({});
```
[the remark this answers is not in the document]
Vinnl: There's a call to `useSession` [here](https://github.com/mozilla/blurts-server/pull/3990/files#diff-73a4c0c3e9533af14ca2a89a97b1e4c8b67e0a4c2b2120b35f472ffc149eec2dR13), which [expects](https://github.com/mozilla/blurts-server/pull/3990/files#diff-73a4c0c3e9533af14ca2a89a97b1e4c8b67e0a4c2b2120b35f472ffc149eec2dR15) the return value to be an object with optionally a `data` property on it.

We're mocking out `next-auth/react`, but if we don't specify a return value for the mocked `useSession`, it will return `undefined`, resulting in this error:

    Error: Uncaught [TypeError: Cannot read properties of undefined (reading 'data')]

Theoretically, [these lines](https://github.com/mozilla/blurts-server/pull/3990/files#diff-ff0a0f27106b116a3c2d0c1216460f3c6d5fd688772ddb7944ec4acff4df31a6R30-R32) should be sufficient to let the `useSession` mock return an actual object (albeit without `.data`).

But for some reason that I can't figure out, that is insufficient :/ And so these lines make that mock return an empty object again before every test...

src/app/components/client/SignInButton.tsx
```diff
@@ -0,0 +1,29 @@
...
+  if (typeof session.data?.user.email === "string") {
+    return null;
+  }
+
+  return (
+    <Button
+      variant="secondary"
+      onPress={() =>
+        void signIn("fxa", { callbackUrl: "/redesign/user/dashboard/" })
+      }
```
pdehaan, line 24: Q: Does this need telemetry, or do we capture clicks in the target page views anyways?

src/app/components/client/toolbar/SignInButton.tsx
```diff
@@ -0,0 +1,34 @@
...
+      <Link href="/redesign/user/dashboard">
+        {l10n.getString("main-nav-link-dashboard-label")}
+      </Link>
+    );
+  }
+
+  return (
+    <Button
+      variant="secondary"
+      onPress={() =>
```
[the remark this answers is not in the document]
Vinnl: Added in ad04d880d.

src/app/components/client/SignInButton.tsx
Vinnl: Yep, Kaitlyn [asked](https://github.com/mozilla/blurts-server/pull/3990#discussion_r1446341879) as much as well! Added in ad04d880d.

src/app/(proper_react)/redesign/(public)/LandingView.test.tsx
flozia: I think the reason might be that we are resetting mocks on every test run: https://github.com/mozilla/blurts-server/blob/MNTOR-3698-sign-in-button/jest.config.cjs#L145?

flozia (approved):

src/app/(proper_react)/redesign/(public)/LandingView.test.tsx
Vinnl: Yeah, but that should be resetting them to the implementation defined in `jest.mock` (and just reset the values for `.mock.calls` and override calls to `.mockImplementation` or `.mockReturnValue`), as I understand it.

PR #3994, opened 2024-01-09
MNTOR-2697: preserving attributions

mansaj:
# References: 
Jira: MNTOR-2697

# Description
We want to preserve attribution for UTMs for logged in users and preserve in cookies for the users that are not logged in


src/db/migrations/20240108008813_marketing_attributions.js
```diff
@@ -0,0 +1,30 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+export async function up(knex) {
```
pdehaan, line 5: nit: Does this need some JSDoc typings like the `down()` defn below?

src/app/components/client/PageLoadEvent.tsx
```diff
@@ -18,25 +19,61 @@ export type Props = {
...
+    recordTelemetry("page", "view", pageViewParams);
+  }, [recordTelemetry, pathname, userId]);
+
+  // record attributions on page load
+  if (window.location.search?.length > 0) {
+    if (!localStorage.getItem("attributionsFirstTouch")) {
+      localStorage.setItem("attributionsFirstTouch", window.location.search);
+    }
+    localStorage.setItem("attributionsLastTouch", window.location.search);
+  }
```
pdehaan, line 61: Does this need to be in a `try..catch` block?
I thinks `localStorage` can potentially throw a `SecurityError` if local storage is disabled. [1]
And looks like a `QuotaExceededError` if the target disk is full or the user declined a request for more space; when calling `setItem()`. [2]

[1] https://developer.mozilla.org/en-US/docs/Web/API/Window/localStorage#exceptions
[2] https://developer.mozilla.org/en-US/docs/Web/API/Storage/setItem#exceptions

src/db/migrations/20240108008813_marketing_attributions.js
pdehaan, line 1: Q: Does anything use this table yet?

src/knex-tables.d.ts
```diff
@@ -11,6 +11,36 @@ import { BreachDataTypes } from "./app/functions/universal/breach";
 
 // See https://knexjs.org/guide/#typescript
 declare module "knex/types/tables" {
+  interface AttributionRow {
+    id: number;
+    subscriber_id: number;
+    type: string; // first, last touch
```
pdehaan, line 17: Noticed this `type` prop isn't in the `AttributionOptionalColumns` _or_ `AttributionAutoInsertedColumns` schemas below.

mansaj: it's neither optional nor auto inserted

src/db/migrations/20240108008813_marketing_attributions.js
```diff
@@ -0,0 +1,30 @@
...
+    .createTable("attributions", table => {
+      table.increments('id').primary()
+      table.integer("subscriber_id").references("subscribers.id").notNullable().onDelete("CASCADE").onUpdate("CASCADE");
+      table.string("type").notNullable()
+      table.string("utm_source")
+      table.string("utm_campaign")
+      table.string("utm_medium")
+      table.string("utm_term")
+      table.string("entrypoint")
+      table.string("flow_id")
```
rhelmer, line 16: Is storing the `flow_id` useful? I thought this was for an individual session, it seems like we could potentially have a lot more rows if we store this.

rhelmer (changes requested):
lgtm overall, wondering if it's useful to store the `flow_id` permanently though.

src/db/migrations/20240108008813_marketing_attributions.js
mansaj: @rhelmer One thing I need to add is clearing the cookies once they are preserved in the db
I'm not sure how permanent these rows should be... I think we can keep them forever, or we can actively delete them once glean events are submitted?

mansaj: Another option I was thinking about is adding a UNIQUE on `subscriber_id + type` so there's a limit of max 2 entries per `subscriber_id`.. instead of appending, we will just update. That's also an option to keep the record count low

rhelmer: I think it's probably worth recording any UTM parameters encountered, it's likely that the attribution model will get more sophisticated in the future, so it might be nice to have the data there. I'm not too worried about the row count but `flow_id` I just don't know if it's worth storing since it's supposed to be a per-session thing.

rhelmer: > @rhelmer One thing I need to add is clearing the cookies once they are preserved in the db I'm not sure how permanent these rows should be... I think we can keep them forever, or we can actively delete them once glean events are submitted?

I think we should keep the data in the database, I'm not worried about the cookies after we've used them,

mansaj: makes sense, I can take that out

mansaj: > > @rhelmer One thing I need to add is clearing the cookies once they are preserved in the db I'm not sure how permanent these rows should be... I think we can keep them forever, or we can actively delete them once glean events are submitted?
> 
> I think we should keep the data in the database, I'm not worried about the cookies after we've used them,

Another related question is, do we want to set a unique on the table for the utm params:
Say we have a match of 
```
"subscriber_id"
      "type"
     "utm_source"
      "utm_campaign"
      "utm_medium"
      "utm_term"
      "entrypoint"
      ```
      
      Would we ever want to record that again in the db when a record already exists in the past?

rhelmer: No I don't think we need to store duplicates... in fact the more I think about it, we're already emitting telemetry events with this data so we probably don't have to worry too much about permanent storage in any case.

rhelmer (approved):

src/db/migrations/20240108008813_marketing_attributions.js
mansaj: I'm setting an unique() on the table: https://github.com/mozilla/blurts-server/pull/3994/commits/543f182963afa11b194ef115ef92ecb87cda3623

PR #4001, opened 2024-01-10
Count page views on the new site

Vinnl:
# References: 
Jira: MNTOR-2335
Figma: N/A

# Description

This fixes the `<PageLoadEvent>` component to actually track the path of the current page (@rhelmer unless I made a big oversight and it was tracked in a way I did not understand), and adds it to the new pages as well.

# How to test

I added a log statement [here](https://github.com/mozilla/blurts-server/blob/a039d08f5356b7fa2ee25182a87141d3af6a7dfe/src/app/components/client/PageLoadEvent.tsx#L50) to verify what was logged. But I assume the Glean realtime viewer should also work, if you remember where that was :sweat_smile: 

- [ ] I've added a unit test to test for potential regressions of this bug. - Hooks in to the Next.js lifecycle, so would involve lots of mocking I think.

### DIFF
--- a/src/app/(nextjs_migration)/(authenticated)/user/layout.tsx
+++ b/src/app/(nextjs_migration)/(authenticated)/user/layout.tsx
@@ -18,7 +18,7 @@ import MozillaLogo from "../../../../client/images/moz-logo-1color-white-rgb-01.
 import { getL10n } from "../../../functions/server/l10n";
 import { authOptions } from "../../../api/utils/auth";
 import { getNonce } from "../../functions/server/getNonce";
-import { PageLoadEvent } from "../../components/client/PageLoadEvent";
+import { PageLoadEvent } from "../../../components/client/PageLoadEvent";
 import { getExperiments } from "../../../functions/server/getExperiments";
 import { getEnabledFeatureFlags } from "../../../../db/tables/featureFlags";
 
--- a/src/app/(nextjs_migration)/(guest)/layout.tsx
+++ b/src/app/(nextjs_migration)/(guest)/layout.tsx
@@ -14,11 +14,10 @@ import { SignInButton } from "../components/client/SignInButton";
 import { getL10n } from "../../functions/server/l10n";
 import { getServerSession } from "next-auth";
 import { authOptions } from "../../api/utils/auth";
-import { cookies } from "next/headers";
-import { randomUUID } from "crypto";
 import { PageLoadEvent } from "../components/client/PageLoadEvent";
 import { getExperiments } from "../../functions/server/getExperiments";
 import { getEnabledFeatureFlags } from "../../../db/tables/featureFlags";
+import { getUserId } from "../../functions/server/getUserId";
 
 export type Props = {
   children: ReactNode;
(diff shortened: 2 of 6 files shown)


mansaj:
Testing locally, I keep getting this error:
```
(Glean.core.metrics.Event) page.view: Unexpected value for extra key user_id: undefined
```

Which seems to suggest that the user_id isn't set correctly

rhelmer (approved):
Hm, we may have accidentally removed `path` from the `nextjs_migration` version of this, thanks for fixing. I did not test locally but this looks like good cleanup and the new ping lgtm, just make sure it shows up on the debug ping dashboard as intended. Thanks!

Vinnl:
@mansaj Do you have the `FxaUidTelemetry` flag enabled? In any case, I added 2d685629f4c36377b2c86fd28526c1e18f4aec6c that avoids that error.

rhelmer:
> @mansaj Do you have the `FxaUidTelemetry` flag enabled? In any case, I added [2d68562](https://github.com/mozilla/blurts-server/commit/2d685629f4c36377b2c86fd28526c1e18f4aec6c) that avoids that error.

@Vinnl I think until we determine whether we're allowed to use the FxA UID (which would allow linking to other Mozilla systems), we need to have some kind of stable UID for Nimbus to use to determine if a user should be exposed to a particular feature or not.

Additionally, even if/when we are allowed to use the FxA UID, we will continue to need some kind of "guest" user ID (that is stored on the client) for users that are not logged in.

rhelmer:
> > @mansaj Do you have the `FxaUidTelemetry` flag enabled? In any case, I added [2d68562](https://github.com/mozilla/blurts-server/commit/2d685629f4c36377b2c86fd28526c1e18f4aec6c) that avoids that error.
> 
> @Vinnl I think until we determine whether we're allowed to use the FxA UID (which would allow linking to other Mozilla systems), we need to have some kind of stable UID for Nimbus to use to determine if a user should be exposed to a particular feature or not.
> 
> Additionally, even if/when we are allowed to use the FxA UID, we will continue to need some kind of "guest" user ID (that is stored on the client) for users that are not logged in.

It looks like you've preserved the above in https://github.com/mozilla/blurts-server/pull/4001/files#diff-b6d2d9f09f26e4c8ef165c304b93f33aa0ba71bcb2077439fa0b7a77637f1231R10-R32, I think what the `FxaUidTelemetry` needs to gate is if the actual UID is transmitted. When the flag is disabled, we should still send some kind of stable user ID.

mansaj (approved):

Vinnl:
@rhelmer I preserved the behaviour that was already present, but I don't think the existing behaviour already sent a stable user ID if the flag is disabled. Here's the existing code that discards the generated user ID and does not set another stable user ID if the flag is not set: https://github.com/mozilla/blurts-server/blob/677dc7696094f2b9db93ea27d1fcaae1f3f452a1/src/app/(nextjs_migration)/components/client/PageLoadEvent.tsx#L36-L38

Shall I create a follow-up ticket that implements that? Not sure what the urgency would be on that.

PR #4003, opened 2024-01-11
MNTOR-2711 - move subscription check to webhook

rhelmer:
# References: 
Jira: MNTOR-2711

# Description

Currently we depend on the FxA profile (refreshed from the client) and the JWT received from signing in to determine subscription status. We have been having intermittent problems with this approach that are difficult to reproduce, after chatting with FxA eng they suggest that we make this part of the webhook.

# Screenshot (if applicable)

Not applicable.

# How to test

The easiest way to test is to send traffic directly to the endpoint.


src/app/api/v1/admin/users/[primarySha1]/route.ts
```diff
@@ -117,15 +123,34 @@ export async function PUT(
       for (const action of actions) {
         switch (action) {
           case "subscribe": {
+            const sub = await getSubscriberByFxaUid(subscriber.fxa_uid!);
+            const currentFxAProfile = sub.fxa_profile_json as FxaProfile;
+            const subscriptions = new Set(currentFxAProfile.subscriptions);
+
+            subscriptions.add(MONITOR_PREMIUM_CAPABILITY);
```
Comment (Vinnl) [inline src/app/api/v1/admin/users/[primarySha1]/route.ts:130]: Does the webhook pass us the FxA profile data? I'm a bit nervous about manually modifying our copy of the data, lest it gets out-of-sync with the source of truth...

src/app/api/v1/admin/users/[primarySha1]/route.ts
```diff
@@ -117,15 +123,34 @@ export async function PUT(
...
+            // update fxa profile data to match subscription status
+            currentFxAProfile.subscriptions = Array.from(subscriptions);
+            await updateFxAProfileData(sub, JSON.stringify(currentFxAProfile));
+
             // activate and opt out profiles
-            await activateAndOptoutProfile(onerepProfileId);
+            await activateProfile(onerepProfileId);
+            await optoutProfile(onerepProfileId);
```
Comment (Vinnl) [inline src/app/api/v1/admin/users/[primarySha1]/route.ts:137]: Any reason to split this up again? Does that reason also hold for other invocations of `activateAndOptoutProfile`? (i.e. should that function be removed?)

src/app/api/v1/admin/users/[primarySha1]/route.ts
```diff
@@ -117,15 +123,34 @@ export async function PUT(
...
             break;
           }
           case "unsubscribe": {
+            const sub = await getSubscriberByFxaUid(subscriber.fxa_uid!);
+            const currentFxAProfile = sub.fxa_profile_json as FxaProfile;
+            const subscriptions = new Set(currentFxAProfile.subscriptions);
+
+            subscriptions.delete(MONITOR_PREMIUM_CAPABILITY);
+            // update fxa profile data to match subscription status
+            currentFxAProfile.subscriptions = Array.from(subscriptions);
```
Comment (Vinnl) [inline src/app/api/v1/admin/users/[primarySha1]/route.ts:151]: Heh, could also be done without converting to a set, mutating that set, and then converting it back to an array, see below.

But I have a similar concern here about it getting out of sync with the FxA data at the source of truth.

```suggestion
            // update fxa profile data to match subscription status
            currentFxAProfile.subscriptions = currentFxAProfile.subscriptions.filter(sub => sub !== MONITOR_PREMIUM_CAPABILITY);
```

src/app/api/v1/fxa-rp-events/route.ts
```diff
@@ -191,7 +191,7 @@ export async function POST(request: NextRequest) {
           event
         ] as ProfileChangeEvent;
         logger.info("fxa_profile_update", {
-          subscriber: subscriber.id,
+          subscriber_id: subscriber.id,
```
Vinnl, line 194: Nice cleanup, this is the kind of thing that can frustrate debugging sessions a lot.

Vinnl (commented):
It mostly looks good to me, but it would be good to have Joe look at this too - I'm not too sure how valid my concerns are.

src/app/api/v1/admin/users/[primarySha1]/route.ts
Comment (rhelmer) [reply src/app/api/v1/admin/users/[primarySha1]/route.ts:137]: I originally used `activateAndOptoutProfile` but it does an early return here:

https://github.com/mozilla/blurts-server/blob/8e6a071250230e2a2a7e90dd7245b86c0e64edc8/src/app/functions/server/onerep.ts#L246-L248

We should rethink how this works and is used by callers, I agree it's confusing having them be separate methods.

src/app/api/v1/admin/users/[primarySha1]/route.ts
Comment (rhelmer) [reply src/app/api/v1/admin/users/[primarySha1]/route.ts:130]: Yes I am concerned about that too, two thoughts I have:

1. FxA is the source of truth, and we synchronize with it via webhook, JWT, or API. The problem here is that we can't re-sync the JWT without sign-out/in, and the profile API doesn't seem to invalidate the cache reliably. 

2. Signing out and back in will cause the JWT to override what's in our database, so I think we won't stay out of sync for too long even if the webhook fails.

I think with the situation around the profile API our only other option here is making users authenticate again (which we could do). There might be a way to force a refresh of the JWT though so it's worth continuing to investigate.

src/app/api/v1/admin/users/[primarySha1]/route.ts
Comment (rhelmer) [reply src/app/api/v1/admin/users/[primarySha1]/route.ts:137]: I can file an issue to refactor, I think this might bite someone in the future.

src/app/api/v1/admin/users/[primarySha1]/route.ts
Comment (Vinnl) [reply src/app/api/v1/admin/users/[primarySha1]/route.ts:130]: OK yeah, if the Webhook doesn't provide the data and the profile API serves outdated data, then we don't have many options left :sweat_smile: But the override at sign-in does alleviate my concerns somewhat.

Comment (rhelmer) [reply src/app/api/v1/admin/users/[primarySha1]/route.ts:130]: The Webhook does send us a JWT payload, sorry for the confusion - you may be thinking of the OneRep webhook which only sends IDs and not an actual payload. However the client JWT may be outdated which is the problem :/

src/app/functions/server/changeSubscription.ts
```diff
@@ -0,0 +1,30 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import { Subscriber } from "../../(nextjs_migration)/(authenticated)/user/breaches/breaches";
+import { updateFxAProfileData } from "../../../db/tables/subscribers";
+
+const MONITOR_PREMIUM_CAPABILITY = "monitor";
+
+export async function changeSubscription(
```
rhelmer, line 10: @Vinnl I moved this to a server function instead of spreading it around the code, hopefully is easier to read now. 

The reason I was thinking the `Set` conversion was easier to read is that the `subscriptions` key may not exist, and we need to create it if it does not.

rhelmer:
Sorry I should amend something I said previously - a JWT is sent by the webhook but it does not include the full `FxaProfile` payload:

```
interface SubscriptionStateChangeEvent {
  capabilities: [string];
  isActive: boolean;
  changeTime: number;
}
```

mansaj (approved):
LGTM

src/app/api/utils/auth.ts
```diff
@@ -66,7 +72,9 @@ export const authOptions: AuthOptions = {
     async jwt({ token, account, profile, trigger }) {
       if (trigger === "update") {
         // Refresh the user data from FxA, in case e.g. new subscriptions got added:
-        profile = await fetchUserInfo(token.subscriber?.fxa_access_token ?? "");
+        const subscriber = await getSubscriberByEmail(token.email ?? "");
+        profile = subscriber.fxa_profile_json as FxaProfile;
```
pdehaan, line 76: Is it possible `getSubscriberByEmail()` can return `null`/`undefined` and then `subscriber` is nullish below?
I'm a bit unclear on what would happen if `token.email` _is_ nullish and falls back to "".

src/app/functions/server/changeSubscription.ts
```diff
@@ -0,0 +1,30 @@
...
+  subscriber: Subscriber,
+  enabled: boolean,
+) {
+  const currentFxAProfile = subscriber?.fxa_profile_json as FxaProfile;
+  let subscriptions = currentFxAProfile.subscriptions ?? [];
```
pdehaan, line 15: if `subscribers` can be nullish, would `currentFxAProfile` be null/undefined and therefore `currentFxAProfile.subscriptions` could fail if we access `.subscriptions` on an undefined object?

src/app/functions/server/changeSubscription.ts
```diff
@@ -0,0 +1,30 @@
...
+
+  if (enabled) {
+    if (!subscriptions?.includes(MONITOR_PREMIUM_CAPABILITY)) {
+      subscriptions?.push(MONITOR_PREMIUM_CAPABILITY);
```
pdehaan, line 19: nit: Should these be `_PLUS_` instead of `_PREMIUM_`?
Also, do we need the `?` in `subscriptions?.` since we default `subscriptions` to an empty array above on L15?

PR #4005, opened 2024-01-11
Remove NEXT_PUBLIC env vars by constants

Vinnl:
We used NEXT_PUBLIC_* env vars for values that:

- were the same across environments, and
- we wanted to easily change.

In other words, they were essentially constants. By turning them into regular, statically-analysable, JS variables, they:

- can easily be renamed,
- are known to be defined and thus don't need non-null or type assertions, and
- can have non-string values, removing the need to e.g. parse integers,
- are more explicitly the same in every environment, avoiding the temptation to attempt (and fail) to set environment-specific env vars to override them.

We'll also no longer have to tell people to copy the values over to their personal `.env` files.

### DIFF
--- a/src/app/hooks/useGa.ts
+++ b/src/app/hooks/useGa.ts
@@ -5,6 +5,7 @@
 "use client";
 
 import { useEffect } from "react";
+import { CONST_GA4_MEASUREMENT_ID } from "../../constants";
 
 declare global {
   interface Window {
@@ -69,8 +70,7 @@ export const useGa = (): {
 
     if (!window.gtag) {
       /* c8 ignore next 2 */
-      const ga4MeasurementId =
-        process.env.NEXT_PUBLIC_GA4_MEASUREMENT_ID || "G-CXG8K4KW4P";
+      const ga4MeasurementId = CONST_GA4_MEASUREMENT_ID || "G-CXG8K4KW4P";
       initGa4({ ga4MeasurementId, debugMode });
     }
   }, [debugMode]);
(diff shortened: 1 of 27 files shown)


src/app/hooks/useGa.ts
```diff
@@ -51,8 +52,7 @@ export const useGa = (): {
   };
 } => {
   const debugMode =
-    process.env.NEXT_PUBLIC_NODE_ENV !== "production" &&
-    process.env.NODE_ENV !== "test";
+    process.env.NODE_ENV !== "production" && process.env.NODE_ENV !== "test";
```
Vinnl, line 55: Calling out that this is the only potentially behaviour-alternating change I made - AFAIK, `NEXT_PUBLIC_NODE_ENV` isn't set anywhere any more. I changed it to regular `NODE_ENV`, but if we also want to enable debug mode on stage, maybe it should be `APP_ENV`? /cc @rhelmer

rhelmer: Hm, I think Next might automatically treat `NODE_ENV` as if it was `NEXT_PUBLIC_NODE_ENV` and compile it in... we should test. I think we might need to pass this as a prop (or alternatively you could have a server-side API that's called at runtime, maybe server actions make this easy enough?)

rhelmer (changes requested):
See my comment in https://github.com/mozilla/blurts-server/pull/4005/files#r1449288829 otherwise lgtm

src/app/hooks/useGa.ts
Vinnl: Oh, that would be problematic indeed. But good point, it's probably best to keep the existing behaviour for now (even if it's "never matching" because it's undefined), then we can always revisit later: 9375c3544.

rhelmer (approved):

PR #4006, opened 2024-01-11
City and state input enhancements (MNTOR-2713)

flozia:
# References: 
Jira: [MNTOR-2713](https://mozilla-hub.atlassian.net/browse/MNTOR-2713)

# Description

This PR adds the following changes to the “City and state” input:
- Move city and state to the last field
- Don’t clear the city and state field `onBlur`
- Show the city and state error state right away, not just when a user clicks find exposures.
- Show an empty state when no location suggestions match the search value.

[MNTOR-2713]: https://mozilla-hub.atlassian.net/browse/MNTOR-2713?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


locales-pending/onboarding.ftl
```diff
@@ -50,6 +50,7 @@ onboarding-enter-details-label-last-name = Last name
 onboarding-enter-details-placeholder-last-name = Enter last name
 onboarding-enter-details-label-location = City and state
 onboarding-enter-details-placeholder-location = Enter city and state
+onboarding-enter-details-placeholder-location-results = No locations found for search
```
flodolo, line 53: This sounds strange. Does it mean that no location was found when searching? If that's the case, wouldn't "No location found" be enough?

flozia: Yes, that’s what it’s supposed to mean. Ran this by content and “No location found” works, thanks!

Vinnl (approved):
A strict improvement! Potentially could do with a unit test, though mocking the autocompletion API might be a bit of a hassle.

flozia:
> A strict improvement! Potentially could do with a unit test, though mocking the autocompletion API might be a bit of a hassle.

I’ll follow-up with additional unit tests. Getting this in now to get the high-priority issues in before the next round of testing.

PR #4033, opened 2024-01-15
Switch to the redesign

Vinnl:
# References: 
Jira: MNTOR-2677

# Description

This makes the redesign the default experience, and uses the final URLs. What I did (see individual commits):

- Renamed the `redesign` directory to `(redesign)`, so it would no longer show up in the URL paths, without having to update relative imports: 0afcee9
- Search-and-replaced `/redesign/` with `/`, to update all links that still referenced it: fd4b215
- Renamed `(nextjs_migration)` in `/src/app` to `deprecated`, so that it wouldn't share URLs paths (such as `/`) with the new site: 16d7cda
- Copied over the `/breaches/` and `/breach-details/<breach name>/` pages, and also the layouts they used, into `(migration_remnants)` so that they _would_ be visible at their existing URLs. Hopefully we can port these over soon as well: ccc6558
- Added redirects from `/redesign/:path` to just `/:path`: c6ebc82

I also did two slightly-unrelated changes:

- Let Storybook know that we renamed the `/ssn` URL: dde9f99
- Added an `<AutoSignIn>` component to replicate the same functionality in the old sign-in button: 0d6dd03

# How to test

Generally just see if everything works as expected I suppose.

### DIFF
--- a/.github/workflows/e2e_cron.yml
+++ b/.github/workflows/e2e_cron.yml
@@ -14,7 +14,7 @@ on:
           - prod
           - heroku
 jobs:
-  test:
+  e2e-tests:
     timeout-minutes: 60
     runs-on: ubuntu-latest
     steps:
--- a/.github/workflows/e2e_pr.yml
+++ b/.github/workflows/e2e_pr.yml
@@ -5,7 +5,7 @@ on:
   pull_request:
     branches: [ main ]
 jobs:
-  test:
+  e2e-tests:
     timeout-minutes: 60
     runs-on: ubuntu-latest
     # Service containers to run with `container-job`
--- a/.github/workflows/reference_linter.yaml
+++ b/.github/workflows/reference_linter.yaml
@@ -4,7 +4,7 @@ on:
   pull_request:
   workflow_dispatch:
 jobs:
-  linter:
+  l10n-lint:
     runs-on: ubuntu-latest
     steps:
       - name: Clone repository
--- a/.github/workflows/unittests.yaml
+++ b/.github/workflows/unittests.yaml
@@ -3,7 +3,7 @@ name: Unit Tests
 on: [push]
 
 jobs:
-  test:
+  unit-tests:
 
     runs-on: ubuntu-latest
 


src/app/(migration_remnants)/layout.tsx
```diff
@@ -0,0 +1,61 @@
...
+  }
+
+  return (
+    <L10nProvider bundleSources={l10nBundles}>
+      {/* This script predates the use of React and thus shouldn’t wait for
+      hydration to adjust the layout. */}
+      {/* eslint-disable-next-line @next/next/no-sync-scripts */}
+      <script
+        type="module"
+        src="/nextjs_migration/client/js/resizeObserver.js"
```
pdehaan, line 42: Should this URL have `/nextjs_migration/` in the final URL.

src/app/deprecated/(authenticated)/user/breaches/page.tsx
pdehaan, line 1: Probably unrelated, but I did a grep based on a previous diff in this PR and noticed L73 has:
```jsx
return <SignInButton autoSignIn />;
```

Not sure if that should be:
```jsx
return <AutoSignIn />;
```

src/app/deprecated/(authenticated)/user/layout.tsx
pdehaan, line 1: Similarly, not sure if L32 should use:
```diff
-return <SignInButton autoSignIn />;
+return <AutoSignIn />;
```

Per:
```sh
git grep -n "<SignInButton " | cat

src/app/(migration_remnants)/(guest)/layout.tsx:66:              <SignInButton />
src/app/(proper_react)/(redesign)/(public)/LandingView.tsx:42:        <SignInButton />
src/app/deprecated/(authenticated)/user/breaches/page.tsx:73:    return <SignInButton autoSignIn />;
src/app/deprecated/(authenticated)/user/layout.tsx:32:    return <SignInButton autoSignIn />;
src/app/deprecated/(guest)/layout.tsx:67:              <SignInButton />
```

src/app/(migration_remnants)/layout.tsx
Vinnl: For now, yes - the current website does this as well. That directory serves up all static (i.e. non-preprocessed-at-build-time) assets that are used by the old website (including the as-yet-unported public breaches pages). Once all pages are ported, we can delete that entire directory.

src/app/(migration_remnants)/(guest)/layout.tsx
```diff
@@ -0,0 +1,121 @@
...
+        <menu>
+          <li>
+            <a href="/breaches">{l10n.getString("footer-nav-all-breaches")}</a>
+          </li>
+          <li>
+            <a
+              href="https://support.mozilla.org/kb/firefox-monitor-faq"
+              target="_blank"
+            >
+              FAQ
```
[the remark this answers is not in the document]
Vinnl: I think that's an `en-GB` string :P 

Leaving in place for now because I'm still hoping to be able to port these pages before launch, so I don't want to add new l10n strings for that short amount of time. (And it's already present on the current website.)

.eslintrc.json
```diff
@@ -117,7 +117,8 @@
         "src/views/**/*.js",
         "src/middleware/**/*.js",
         "src/controllers/**/*.js",
-        "src/app/(nextjs_migration)/**/*",
+        "src/app/deprecated/**/*",
```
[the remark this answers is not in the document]
Vinnl: The problem with that is that the new and old website share URLs (e.g. `/`, and `/user/dashboard`). The current approach allowed me to make our URLs point to the new pages without breaking lots of relative imports. The next step would be to actually delete the entire `/deprecated/` directory.

src/app/deprecated/(authenticated)/user/breaches/page.tsx
Vinnl: Yeah, we currently have two `<SignInButton>`s, once of which is only in use by the old website - that one has an `autoSignIn` prop, whereas I extracted that into a separate component for the new website. They do the same thing, but this version will be deleted together with the rest of the old website - so it's no problem if it's in use on the old website.

src/app/(migration_remnants)/(guest)/layout.tsx
Vinnl, line 1: Note that this file is an exact copy of `/src/app/deprecated/(guest)/layout.tsx`, copied over so it can still be used for the old breaches pages.

src/app/(migration_remnants)/layout.tsx
Vinnl, line 1: Note that this file is an exact copy of `/src/app/deprecated/layout.tsx` (what used to be `/src/app/(nextjs_migration)/layout.tsx`), copied over so it can still be used for the old breaches pages.

next.config.js
```diff
@@ -123,6 +123,21 @@ const nextConfig = {
...
+        permanent: true,
+      },
+      // While we were implementing a redesign, we made it available at the
+      // /redesign subpath. In case we still have lingering links to there
+      // anywhere, this redirect should have people end up at the right place:
+      {
+        source: '/redesign/:path*',
+        destination: '/:path*',
+        permanent: true,
+      },
```
rhelmer, line 140: 💯

rhelmer (approved):
I think this approach makes sense, the fact that it compiles means you probably got all the imports. Do we know why the e2e tests are failing?

Vinnl:
@rhelmer They're still tuned to the old site, and thus e.g. don't take into account that new users get sent to the onboarding flow. See [this Slack thread](https://mozilla.slack.com/archives/C04BLQVN151/p1705343491630329?thread_ts=1705342329.542369&cid=C04BLQVN151).

PR #4035, opened 2024-01-16
Add banner announcing "Mozilla Monitor" rebrand

Vinnl:
# References: 
Jira: MNTOR-2522
Figma: https://www.figma.com/file/51jgudmkBoMMU9twT7UOKB/Monitor-Plus-Launch?type=design&node-id=310-6816&mode=design&t=PAVHr1PiURqF4TxK-0

# Description

(Based on https://github.com/mozilla/blurts-server/pull/4033 to avoid adding this to the wrong directory.)

# Screenshot (if applicable)

image

image

# How to test

First, enabled the `RebrandAnnouncement` feature flag.

Open the landing page. You should see the announcement. After you click the X, it should go away, and stay away after a refresh.

### DIFF
--- a/src/app/(proper_react)/(redesign)/(public)/LandingView.test.tsx
+++ b/src/app/(proper_react)/(redesign)/(public)/LandingView.test.tsx
@@ -12,6 +12,7 @@ import {
   queryByText,
   render,
   screen,
+  within,
 } from "@testing-library/react";
 import { userEvent } from "@testing-library/user-event";
 import { axe } from "jest-axe";
@@ -162,6 +163,39 @@ describe("When Premium is not available", () => {
     await user.click(signInButton);
 
     expect(signIn).toHaveBeenCalledTimes(1);
+  });
+
+  it("displays a banner announcing the rebrand to 'Mozilla Monitor' by default", () => {
+    const ComposedDashboard = composeStory(LandingNonUs, Meta);
+    render(<ComposedDashboard />);
+
+    const rebrandAnnouncement = screen.getByText(
+      "New name, look and even more ways to",
+      { exact: false },
+    );
+
+    expect(rebrandAnnouncement).toBeInTheDocument();
+  });
+
+  it("hides the banner announcing the rebrand to 'Mozilla Monitor' after clicking the dismiss button", async () => {
+    const ComposedDashboard = composeStory(LandingNonUs, Meta);
+    render(<ComposedDashboard />);
+
+    const rebrandAnnouncement = screen
+      .getAllByRole("complementary")
+      .find((el) =>
+        el.textContent?.includes(
+          "⁨Mozilla Monitor⁩: New name, look and even more ways to reclaim your privacy.",
+        ),
+      )!;
+    const dismissButton = within(rebrandAnnouncement).getByRole("button", {
+      name: "OK",
+    });
+
+    const user = userEvent.setup();
+    await user.click(dismissButton);
+
+    expect(rebrandAnnouncement).not.toBeInTheDocument();
   });
 
   it("shows the german scanning for exposures illustration", () => {
--- a/src/app/(proper_react)/(redesign)/(public)/LandingView.test.tsx
+++ b/src/app/(proper_react)/(redesign)/(public)/LandingView.test.tsx
@@ -42,6 +42,14 @@ beforeEach(() => {
   // long debugging that already, so I'm settling for this :(
   const mockedUseSession = useSession as jest.Mock;
   mockedUseSession.mockReturnValue({});
+
+  // Delete all cookies, to make the rebrand announcement banner show up by
+  // default (I hate the document.cookie API ¬_¬):
+  const cookieParts = document.cookie.split(";");
+  const cookieNames = cookieParts.map((part) => part.split("=")[0]);
+  cookieNames.forEach((cookieName) => {
+    document.cookie = `${cookieName}=; Thu, 01 Jan 1970 00:00:00 GMT`;
+  });
 });
 
 describe("When Premium is not available", () => {
@@ -196,6 +204,30 @@ describe("When Premium is not available", () => {
     await user.click(dismissButton);
 
     expect(rebrandAnnouncement).not.toBeInTheDocument();
+  });
+
+  it("counts how often people close the banner announcing the rebrand to `'Mozilla Monitor'", async () => {
+    const mockedRecord = useTelemetry();
+    const ComposedDashboard = composeStory(LandingNonUs, Meta);
+    render(<ComposedDashboard />);
+
+    const rebrandAnnouncement = screen.getByText(
+      "New name, look and even more ways to",
+      { exact: false },
+    );
+    const dismissButton = within(rebrandAnnouncement.parentElement!).getByRole(
+      "button",
+      { name: "OK" },
+    );
+
+    const user = userEvent.setup();
+    await user.click(dismissButton);
+
+    expect(mockedRecord).toHaveBeenCalledWith(
+      "button",
+      "click",
+      expect.objectContaining({ button_id: "rebrand_announcement_dismiss" }),
+    );
   });
 
   it("shows the german scanning for exposures illustration", () => {
--- a/src/app/(proper_react)/(redesign)/(public)/RebrandAnnouncement.module.scss
+++ b/src/app/(proper_react)/(redesign)/(public)/RebrandAnnouncement.module.scss
@@ -1,0 +1,38 @@
+@import "../../../tokens";
+
+.rebrandAnnouncement {
+  position: sticky;
+  bottom: 0;
+  display: flex;
+  flex-direction: row;
+  align-items: center;
+  justify-content: center;
+  gap: $spacing-xs;
+  padding: $spacing-lg;
+  background-color: $color-blue-20;
+  font: $text-body-lg;
+
+  @media screen and (max-width: $screen-md) {
+    // (button width) + (.rebrandAnnouncement gap) + (2 * inline button padding)
+    padding-inline-end: calc(24px + calc($spacing-xs + calc(2 * $spacing-sm)));
+
+    button {
+      position: absolute;
+      inset-block-start: $spacing-sm;
+      inset-inline-end: $spacing-sm;
+    }
+  }
+
+  button {
+    flex: 0 0 auto;
+    width: 24px;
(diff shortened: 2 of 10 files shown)


src/app/hooks/useLocalDismissal.ts
Vinnl, line 1: This hooks was based on [the equivalent hook in Relay](https://github.com/mozilla/fx-private-relay/blob/main/frontend/src/hooks/localDismissal.ts), except using (and working around :unamused:) the cookie library we were already using.

src/app/hooks/useHasRenderedClientSide.tsx
```diff
@@ -0,0 +1,15 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import { useEffect, useState } from "react";
+
+export function useHasRendedClientSide() {
```
pdehaan, line 7: Nit: "Rended" --> "Rendered" (plus the 2&times; other instances where this is imported and used above in `src/app/(proper_react)/(redesign)/(public)/RebrandAnnouncement.tsx`)
```suggestion
export function useHasRenderedClientSide() {
```

src/app/(proper_react)/(redesign)/(public)/RebrandAnnouncement.tsx
```diff
@@ -0,0 +1,42 @@
...
+import styles from "./RebrandAnnouncement.module.scss";
+import { CloseBtn } from "../../../components/server/Icons";
+import { useL10n } from "../../../hooks/l10n";
+import { useLocalDismissal } from "../../../hooks/useLocalDismissal";
+import { useHasRendedClientSide } from "../../../hooks/useHasRenderedClientSide";
+
+export const RebrandAnnouncement = () => {
+  const l10n = useL10n();
+  const dismissal = useLocalDismissal("mozilla-monitor-rebrand", {
+    duration: Number.MAX_SAFE_INTEGER,
```
pdehaan, line 16: What is the `duration` for here? Auto-closing?
Acccctually, per a doc comment below:
> ```js
> /** How long the dismissal should last, in seconds. Note that by default, the cookie will be deleted in 30 days. */
> ```

Where [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/MAX_SAFE_INTEGER) says the value of `Number.MAX_SAFE_INTEGER` is ~9 quadrillion; and Google says 
that [9,007,199,254,740,991 seconds to years](https://www.google.com/search?q=9%2C007%2C199%2C254%2C740%2C991+s+to+years) is approximately 285_616_414 years (give or take a couple months).

src/app/hooks/useLocalDismissal.ts
```diff
@@ -0,0 +1,84 @@
...
+
+  const [isDismissed, setIsDismissed] = useState(
+    hasDismissedCookie(cookieValue, options.duration),
+  );
+
+  const dismiss = (dismissOptions?: DismissOptions) => {
+    const maxAgeInSeconds =
+      typeof options.duration === "number"
+        ? options.duration
+        : 100 * 365 * 24 * 60 * 60;
```
pdehaan, line 46: 100 years?

src/app/hooks/useHasRenderedClientSide.tsx
Vinnl: Thank you! 690b5fccd

src/app/(proper_react)/(redesign)/(public)/RebrandAnnouncement.tsx
Vinnl: Yep correct. I might've gone overboard with `MAX_SAFE_INTEGER` here, not showing it again for a 100 years would probably have been fine too :sweat_smile:

src/app/hooks/useLocalDismissal.ts
Vinnl: "forever", basically. (There were a couple of instances on Relay where we show something again if it was dismissed more than `x` time ago. Maybe I should just remove it here until we find we actually need it.)

src/app/(proper_react)/(redesign)/(public)/LandingView.test.tsx
```diff
@@ -164,6 +173,61 @@ describe("When Premium is not available", () => {
...
+
+  it("hides the banner announcing the rebrand to 'Mozilla Monitor' after clicking the dismiss button", async () => {
+    const ComposedDashboard = composeStory(LandingNonUs, Meta);
+    render(<ComposedDashboard />);
+
+    const rebrandAnnouncement = screen.getByText(
+      "New name, look and even more ways to",
+      { exact: false },
+    );
+    const dismissButton = within(rebrandAnnouncement.parentElement!).getByRole(
```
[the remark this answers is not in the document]
Vinnl: I was actually hoping to not have to use it, but I didn't have a good way to target the complete banner. (I think it might need some accessible role, but I'm not sure which...)

src/app/(proper_react)/(redesign)/(public)/LandingView.test.tsx
```diff
@@ -164,6 +173,61 @@ describe("When Premium is not available", () => {
...
+      "button",
+      { name: "OK" },
+    );
+
+    const user = userEvent.setup();
+    await user.click(dismissButton);
+
+    expect(mockedRecord).toHaveBeenCalledWith(
```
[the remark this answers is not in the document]
Vinnl: This isn't calling `mockedRecord`, it's checking whether it _has been called_ by the code under test. In other words, I've replaced the `useTelemetry` function that the code calls by a test-specific function (see line 37 of this file), which returns a mock function (`mockedRecord`) to replace the function the code would usually call (`recordTelemetry`). The mock function doesn't actually do anything, other than keep track of how often it has been called, and with which parameters.

On this line, I'm checking that it has actually been called, with the parameters that I'd expect for tracking a button click (`"button"`, `"click"` and an object with the property `button_id` set to `"rebrand_announcement_dismiss"`).

So in other words, this test is not verifying that the banner disappears, but that the correct telemetry is sent.

src/app/(proper_react)/(redesign)/(public)/RebrandAnnouncement.module.scss
```diff
@@ -0,0 +1,38 @@
...
+  align-items: center;
+  justify-content: center;
+  gap: $spacing-xs;
+  padding: $spacing-lg;
+  background-color: $color-blue-20;
+  font: $text-body-lg;
+
+  @media screen and (max-width: $screen-md) {
+    // (button width) + (.rebrandAnnouncement gap) + (2 * inline button padding)
+    padding-inline-end: calc(24px + calc($spacing-xs + calc(2 * $spacing-sm)));
```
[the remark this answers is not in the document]
Vinnl: Ah! I suppose, I wasn't aware of what the `FalseDoorBanner` looked like. I felt like our banners got designed ad-hoc whenever one is needed, but if they are similar in designs, a general banner component indeed makes sense to me.

As for the specificity, what I wanted to do was add `$spacing-xs` padding in between the button and the text, but since the button is absolutely positioned, I couldn't just add that as a gap - I had to reproduce the button width and insert empty space on the right-hand side of the text. Since I had to calculate what that width would be anyway, I figured I'd might as well make it explicit in code to avoid a magic number and to make it easier to re-calculate if the button width is ever adjusted.

src/app/hooks/useHasRenderedClientSide.tsx
```diff
@@ -0,0 +1,15 @@
...
+
+export function useHasRenderedClientSide() {
+  const [hasRenderedClientSide, setHasRenderedClientSide] = useState(false);
+
+  useEffect(() => {
+    setHasRenderedClientSide(true);
+  }, []);
+
+  return hasRenderedClientSide;
+}
```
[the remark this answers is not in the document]
Vinnl: No, good point, I should've added that in the first place! Wrote a comment in 7ee9b9d3e; let me know if that makes sense or needs more elaboration.

src/app/hooks/useLocalDismissal.test.tsx
```diff
@@ -0,0 +1,141 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import { act, renderHook } from "@testing-library/react";
+import { useLocalDismissal } from "./useLocalDismissal";
+import { useCookies } from "react-cookie";
```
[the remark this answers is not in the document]
Vinnl: Indeed we'll generally assume that the library is working well (or rather, we'll have vetted that before we add it :sweat_smile: ). Thus, in these tests, we're not testing the library itself, but how our code interacts with it. On line 9 to 13 you can see that we mock the library, i.e. we replace it with our own implementation that doesn't do anything other than keeping track how often it is called, and with what arguments.

Then the tests ensure that we call that library the right way when we should be setting or removing cookies. But we'll leave whether those cookies _actually_ get set and removed up to the library itself, and assume that that's working well.

(Sometimes those assumptions break, or the way we're calling the library is actually wrong. In other words, the way we're integrating our code and the library is wrong. If we consider the risk of that to be high, we could write _integration tests_, or cover them in our end-to-end (Playwright) tests - but that's more of a hassle and [more likely to result in false positives](https://testing.googleblog.com/2017/04/where-do-our-flaky-tests-come-from.html), so we're not really doing that at this time.)

src/app/(proper_react)/(redesign)/(public)/LandingView.test.tsx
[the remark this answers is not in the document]
Vinnl: Maybe, I'm not sure. The general idea is to make our tests perform much the same actions as our users would do, and looking at the string "New name, look, etc" is obviously closer to that than looking at the source code and finding a test ID would be. But: looking for that string, then inspecting the DOM, looking for the parent element, and then finding a button in there of course _also_ isn't really what a user would do.

But with the response I got from the a11y team, I'm now simulating a user using landmark navigation to get to the `<aside>`, and then seeing the text in there: 9cac3055e. That might sort of be better?

PR #4037, opened 2024-01-16
MNTOR-2711 - update database from subscription webhook before onerep

rhelmer:
# References: 
Jira: MNTOR-2711

# Description

When testing on staging, I encountered a situation where the subscription webhook came through but OneRep activation failed. In this case our webhook handler returns an HTTP error so FxA retries, but the badge doesn't show up right away. 

We should update our database first.

# How to test

I ran into this on stage, this can be tested by pointing the `ONEREP_API_BASE` var to an endpoint that always returns `500`.

### DIFF
--- a/src/app/api/v1/fxa-rp-events/route.ts
+++ b/src/app/api/v1/fxa-rp-events/route.ts
@@ -286,8 +286,31 @@ export async function POST(request: NextRequest) {
             }
 
             // activate and opt out profiles
-            await activateProfile(oneRepProfileId);
-            await optoutProfile(oneRepProfileId);
+            try {
+              await activateProfile(oneRepProfileId);
+            } catch (ex) {
+              if (
+                (ex as Error).message ===
+                "Failed to activate OneRep profile: [403] [Forbidden]"
+              )
+                logger.error("profile_already_activated", {
+                  subscriber_id: subscriber.id,
+                  exception: ex,
+                });
+            }
+
+            try {
+              await optoutProfile(oneRepProfileId);
+            } catch (ex) {
+              if (
+                (ex as Error).message ===
+                "Failed to opt-out OneRep profile: [403] [Forbidden]"
+              )
+                logger.error("profile_already_opted_out", {
+                  subscriber_id: subscriber.id,
+                  exception: ex,
+                });
+            }
 
             logger.info("activated_onerep_profile", {
               subscriber_id: subscriber.id,
--- a/src/app/api/v1/fxa-rp-events/route.ts
+++ b/src/app/api/v1/fxa-rp-events/route.ts
@@ -245,44 +245,49 @@ export async function POST(request: NextRequest) {
           subscriber_id: subscriber.id,
         });
 
-        // get profile id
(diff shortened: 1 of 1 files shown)


src/app/api/v1/fxa-rp-events/route.ts
```diff
@@ -286,8 +286,31 @@ export async function POST(request: NextRequest) {
             }
 
             // activate and opt out profiles
-            await activateProfile(oneRepProfileId);
-            await optoutProfile(oneRepProfileId);
+            try {
```
mansaj, line 289: Not in this PR, but I'm wondering why we have two info logging for this case `FXA_SUBSCRIPTION_CHANGE_EVENT`, L244

src/app/api/v1/fxa-rp-events/route.ts
```diff
@@ -245,45 +245,73 @@ export async function POST(request: NextRequest) {
...
-        });
+        try {
+          // get profile id
+          const result = await getOnerepProfileId(subscriber.id);
+          const oneRepProfileId = result?.[0]?.["onerep_profile_id"] as number;
 
-        // MNTOR-2103: if one rep profile id doesn't exist in the db, fail silently
-        if (!oneRepProfileId) {
-          logger.error("onerep_profile_not_found", {
+          logger.info("get_onerep_profile", {
```
mansaj, line 253: already having a logger.info above?

src/app/api/v1/fxa-rp-events/route.ts
```diff
@@ -245,45 +245,73 @@ export async function POST(request: NextRequest) {
           subscriber_id: subscriber.id,
```
mansaj, line 245: already having a logger.info above (l244, l238)?

src/app/api/v1/fxa-rp-events/route.ts
```diff
@@ -293,12 +321,35 @@ export async function POST(request: NextRequest) {
...
+                        )}`),
+              );
+              return NextResponse.json(
+                { success: false, message: "failed_activating_subscription" },
+                { status: 500 },
+              );
+            }
+
             // deactivation stops opt out process
             await deactivateProfile(oneRepProfileId);
```
mansaj, line 351: Should we do the same with deactivate to catch for the 403?

mansaj (approved):
lgtm

src/app/api/v1/fxa-rp-events/route.ts
rhelmer: There isn't anything else to do after deactivation (there's no inverse of the "opt-out" API call), deactivate should also stop the opt-out process.

rhelmer: Oh I see what you're saying though, if it's already de-activated. Yes I think it does make sense then.

src/app/api/v1/fxa-rp-events/route.ts
rhelmer: Yes this one confused me earlier, I think we just need the one that includes the payload.

PR #4047, opened 2024-01-18
MNTOR-2347: fxa / subplat instrumentation and utm params

mansaj:
# References: 
Jira: MNTOR-2347

# Description
For
1. user sign-up (via free scan button in landing) and 
2. various subplat signup flows

monitor would want to preserve the UTM parameters in cookies and pass them to FxA. If no UTM is present, then we will pass some default parameters

For user sign up, we'd want to use first-touch strategy for attribution.
For subplat sign up, we'd want to use last-touch strategy for attribution

# Screenshot (if applicable)

Not applicable.

# How to test
* Go to the landing page, and pass in some utm params: ie`http://localhost:6060/?utm_medium=something`
* Try signing up
* Observe the URL for FxA
* Try signing up for subplat in the pricing table
* Observe the URL for FxA


src/app/(proper_react)/(redesign)/(authenticated)/user/(dashboard)/dashboard/fix/data-broker-profiles/automatic-remove/AutomaticRemove.stories.tsx
```diff
@@ -64,8 +64,8 @@ export const AutomaticRemoveViewStory: Story = {
             href: "/user/dashboard/fix/high-risk-data-breaches/social-security-number",
           }}
           currentSection="data-broker-profiles"
-          monthlySubscriptionUrl=""
-          yearlySubscriptionUrl=""
+          monthlySubscriptionUrl="http://mozilla.com/subscribe/monthly"
+          yearlySubscriptionUrl="http://mozilla.com/subscribe/yearly"
```
Vinnl, line 68: Just to ensure it's clear to everyone that these are mocks (I was confused and visited the URLs :sweat_smile:):

```suggestion
          monthlySubscriptionUrl="https://example.com/subscribe/monthly"
          yearlySubscriptionUrl="https://example.com/subscribe/yearly"
```

src/app/(proper_react)/(redesign)/(authenticated)/user/(dashboard)/dashboard/fix/data-broker-profiles/automatic-remove/AutomaticRemoveView.tsx
```diff
@@ -23,9 +23,52 @@ export function AutomaticRemoveView(props: Props) {
 
   const dataBrokerCount = CONST_ONEREP_DATA_BROKER_COUNT;
 
-  const { monthlySubscriptionUrl, yearlySubscriptionUrl, ...fixViewProps } =
-    props;
+  let { monthlySubscriptionUrl, yearlySubscriptionUrl } = props;
+  const { ...fixViewProps } = props;
```
Vinnl, line 27: I think the destructuring was there to remove the `monthlySubscriptionUrl` and `yearlySubscriptionUrl` props from `fixViewProps`. The destructuring is now a bit idiosyncratic, because `viewFixProps` is essentially just a clone of `props` - why not just pass that then?

As possible refactoring could be to instead just create new local variables (e.g. `normalisedMonthlySubscriptionUrl`) and modify _those_ rather than changing the function's input parameters.

src/db/tables/attributions.ts
```diff
@@ -16,6 +16,28 @@ async function getAttributionsForSubscriber(subscriberId: number) {
...
+    .where("subscriber_id", subscriberId)
+    .andWhere("type", type)
+    .returning([
+      "entrypoint",
+      "utm_source",
+      "utm_medium",
+      "utm_term",
+      "utm_campaign",
+    ]);
+  return res[0] as Partial<AttributionRow>;
```
Vinnl, line 38: If you add the following to the `Tables` interface in `knex-tables.d.ts`, then you don't need the type assertion - TypeScript can automatically derive the correct return value:

```ts

    attributions: Knex.CompositeTableType<
      AttributionRow,
      // On updates, auto-generated columns cannot be set, and nullable columns are optional:
      Omit<
        AttributionRow,
        AttributionAutoInsertedColumns | AttributionOptionalColumns
      > &
        Partial<Pick<AttributionRow, AttributionOptionalColumns>>,
      // On updates, don't allow updating the ID and created date; all other fields are optional, except updated_at:
      Partial<Omit<AttributionRow, "id" | "created_at">> &
        Pick<AttributionRow, "updated_at">
    >;
```

src/app/components/client/UpsellDialog.tsx
```diff
@@ -76,6 +76,51 @@ function UpsellDialogContent({
   const [selectedTab, setSelectedTab] = useState<Key>(defaultSelectedKey);
   const recordTelemetry = useTelemetry();
 
+  // format subscription urls
```
Vinnl, line 79: Am I correct in thinking that this logic is essentially duplicated in a couple of places? If so, might it make sense to extract it into a separate function in `src/app/functions/universal/`? Might also make it easier to understand what is actually being done (and why).

src/app/components/client/PageLoadEvent.tsx
```diff
@@ -56,9 +56,11 @@ export const PageLoadEvent = (props: Props) => {
     // record attributions on page load
     if (window.location.search?.length > 0) {
```
Vinnl, line 57: I'm a bit worried about having this in the same `useEffect` as the page view recording - I'm not sure, but I think it might result in that effect running too often, resulting in over-counting page views. It might be safer to have both the first-touch and last-touch cookie set in their own separate effects?

src/app/(proper_react)/(redesign)/(public)/PlansTable.tsx
```diff
@@ -88,6 +90,31 @@ export const PlansTable = (props: Props & ScanLimitProp) => {
     currencyDisplay: "narrowSymbol",
   });
   const [billingPeriod, setBillingPeriod] = useState<BillingPeriod>("yearly");
+  const searchParam = useRef(new URLSearchParams());
+
+  useEffect(() => {
```
Vinnl, line 95: I don't think this needs a ref and an effect? See also https://react.dev/learn/you-might-not-need-an-effect.

src/app/(proper_react)/(redesign)/(public)/SignUpForm.tsx
```diff
@@ -29,16 +30,35 @@ export const SignUpForm = (props: Props) => {
...
   const onSubmit: FormEventHandler = (event) => {
     event.preventDefault();
     void signIn(
       "fxa",
       { callbackUrl: props.signUpCallbackUrl },
-      // This passes an `?email=` query parameter to FxA, causing it to prefill
-      // the email address in the sign-up form. See
-      // https://mozilla.github.io/ecosystem-platform/relying-parties/reference/query-parameters#email
-      { email: emailInput },
+      attributionSearchParams.toString(),
```
Vinnl, line 61: Looking at the type definition of `signIn`, it looks like it also accepts the `URLSearchParams` object directly? It's probably converting it back to it internall anyway.

```suggestion
      attributionSearchParams,
```

src/app/(proper_react)/(redesign)/(public)/LandingView.test.tsx
```diff
@@ -67,9 +67,11 @@ describe("When Premium is not available", () => {
     await user.click(submitButton[0]);
 
     expect(signIn).toHaveBeenCalledTimes(1);
-    expect(signIn).toHaveBeenCalledWith("fxa", expect.any(Object), {
-      email: "mail@example.com",
-    });
+    expect(signIn).toHaveBeenCalledWith(
+      "fxa",
+      expect.any(Object),
+      expect.stringContaining(`email=`),
```
Vinnl, line 73: Removing the actual email address that was entered kinda neuters the test... I'm assuming this wouldn't break?

```suggestion
      expect.stringContaining(`email=mail@example.com`),
```

(And probably a test for the correct attribution parameters wouldn't harm too :) )

Vinnl (commented):
I feel like I'm missing some context here both w.r.t. existing code as well as w.r.t. the requirements, so I'll leave the approving to someone else, but I added some comments that hopefully are helpful.

src/app/(proper_react)/(redesign)/(authenticated)/user/(dashboard)/dashboard/fix/data-broker-profiles/automatic-remove/AutomaticRemoveView.tsx
```diff
@@ -23,9 +23,52 @@ export function AutomaticRemoveView(props: Props) {
...
 
+  // format subscription urls
+  if (monthlySubscriptionUrl) {
```
rhelmer, line 30: Hm, testing this now it seems to be storing and setting the UTM parameters appropriately, but the subscription URL is now `localhost:6060` instead of the `FXA_SUBSCRIPTIONS_URL` setting I have in `.env`. Did this change or is it overriding it accidentally?

rhelmer (changes requested):

src/app/(proper_react)/(redesign)/(public)/SignUpForm.tsx
mansaj: Tried without `toString()` and it actually doesn't work. Reverting it back

src/app/(proper_react)/(redesign)/(public)/LandingView.test.tsx
mansaj: Change the string match to `email=mail%40example.com` since it's encoded

src/app/(proper_react)/(redesign)/(public)/PlansTable.tsx
mansaj: Good catch! I forgot to take this out thanks

src/app/(proper_react)/(redesign)/(authenticated)/user/(dashboard)/dashboard/fix/data-broker-profiles/automatic-remove/AutomaticRemoveView.tsx
```diff
@@ -26,6 +27,32 @@ export function AutomaticRemoveView(props: Props) {
   const { monthlySubscriptionUrl, yearlySubscriptionUrl, ...fixViewProps } =
     props;
 
+  // format subscription urls
```
pdehaan, line 30: Since these two values are 99% similar, not sure if it's easier to do a partial or some helper to make sure they stay in sync.
```js
const addAttributions = (url) => modifyAttributionsForUrl(url, 
    {
      entrypoint: "monitor.mozilla.org-monitor-in-product-guided-upsell",
      form_type: "button",
    },
    {
      utm_source: "product",
      utm_medium: "monitor",
      utm_campaign: "guided-upsell",
    },
);

const monthlySubscriptionUrlWithAttributions = addAttributions(monthlySubscriptionUrl);
const yearlySubscriptionUrlWithAttributions = addAttributions(yearlySubscriptionUrl);
```

src/app/components/client/UpsellDialog.tsx
mansaj: Done

src/app/(proper_react)/(redesign)/(authenticated)/user/(dashboard)/dashboard/fix/data-broker-profiles/automatic-remove/AutomaticRemoveView.tsx
mansaj: Hmm, good catch. Not sure, but looking into this.

mansaj: This only seems to happen within dashboard, the landing page one is fine

src/app/(proper_react)/(redesign)/(authenticated)/user/(dashboard)/dashboard/page.tsx
```diff
@@ -153,6 +112,63 @@ export default async function DashboardPage() {
...
   const fxaSettingsUrl = process.env.FXA_SETTINGS_URL!;
   const profileStats = await getProfilesStats();
+  let additionalSubplatParams = new URLSearchParams(
+    cookiesList.get("attributionsLastTouch")?.value,
+  );
+
+  // store utm attributions if present in cookies
+  if (cookiesList.get("attributionsFirstTouch")?.value) {
+    const searchParams = new URLSearchParams(
+      cookiesList.get("attributionsFirstTouch")?.value,
```
pdehaan, line 122: nit: Don't think this needs the `?` in `?.value` since we're already gating above on L120.
```suggestion
      cookiesList.get("attributionsFirstTouch").value,
```

src/db/tables/attributions.ts
mansaj: Fixed thanks

src/app/(proper_react)/(redesign)/(authenticated)/user/(dashboard)/dashboard/page.tsx
```diff
@@ -162,8 +178,8 @@ export default async function DashboardPage() {
       userScanData={latestScan}
       userBreaches={subBreaches}
       enabledFeatureFlags={enabledFeatureFlags}
-      monthlySubscriptionUrl={monthlySubscriptionUrl}
-      yearlySubscriptionUrl={yearlySubscriptionUrl}
+      monthlySubscriptionUrl={`${monthlySubscriptionUrl}&${additionalSubplatParams.toString()}`}
+      yearlySubscriptionUrl={`${yearlySubscriptionUrl}&&${additionalSubplatParams.toString()}`}
```
pdehaan, line 182:
```suggestion
      yearlySubscriptionUrl={`${yearlySubscriptionUrl}&${additionalSubplatParams.toString()}`}
```

pdehaan, line 182: Q: Could this result in any duplicated query params, like potentially conflicting `utm_*` parameters?

~And is the `yearlySubscriptionUrl` (and monthly) always guaranteed to have a `?`?~
Ah, I think it does, per https://github.com/mozilla/blurts-server/blob/0b4b93bfca8acb6a3f8df03013ecd0bf25e5c14c/src/app/functions/server/getPremiumSubscriptionUrl.ts#L32

mansaj: Thanks!

src/app/(proper_react)/(redesign)/(authenticated)/user/(dashboard)/dashboard/page.tsx
mansaj: Thanks!

mansaj: Reverting actually, got a lint error

src/app/(proper_react)/(redesign)/(authenticated)/user/(dashboard)/dashboard/fix/data-broker-profiles/automatic-remove/AutomaticRemoveView.tsx
mansaj: Fixed in https://github.com/mozilla/blurts-server/pull/4047/commits/1a2ef7ada756fab93e0943acc1018b104ff6a86a, thanks

src/app/components/client/PageLoadEvent.tsx
mansaj: Hmm, good call. It might be possible that this func runs twice, once on page load, and once when `cookies` change.. I'm not sure either how likely this happens, but sure I'll separate it out into its own `useEffect`

mansaj: Changed in https://github.com/mozilla/blurts-server/pull/4047/commits/72f2d48adcbe0786da722d338d1b3db39e3a5db6

rhelmer (approved):

PR #4049, opened 2024-01-19
Somewhat more resilient selectors for auth page

Vinnl:
It appears that the IDs on the FxA sign-up page have been removed? Ideally I'd use `getByLabel` and `getByRole` to select on markup that matches what users are looking for, but it looks like these selectors are used on different pages (sign in vs. sign up), and thus the labels differ even though the elements are similar. So for now, I've updated the selectors to ones that appear to work for the different pages (even now that some IDs appear to have disappeared), but ideally, we'd have selectors specific to each page, based on user-visible selection characteristics.


mozrokafor (approved):
LGTM

PR #4050, opened 2024-01-19
Make setup headed when any browser is

Vinnl:
It used to be that, even if headless mode was disabled, the setup would still run in headless mode, which can be confusing. This change ensures that the setup is also run in headed mode, if the tests itself are too.


mozrokafor (approved):
Great Idea, LGTM

PR #4081, opened 2024-01-24
Refresh breaches after resolving some

Vinnl:
# References: 
Jira: MNTOR-2464
Figma: 

# Description

This essentially applies the same thing [we already did for data brokers](https://github.com/mozilla/blurts-server/blob/6ce3bcdb02e4e8cf4bf235898699deb58f78eae3/src/app/(proper_react)/(redesign)/(authenticated)/user/(dashboard)/dashboard/fix/data-broker-profiles/manual-remove/RemovalCard.tsx#L39-L42) for data breaches: after we resolve them, we make sure that the next render of the dashboard fetches fresh data. Otherwise, fixed breaches will still be listed under the "active" tab.

# How to test

Open the resolution flow and move to the first breach. Go back to the dashboard and check what data is breached, and then resolve those in the guided flow. With this branch, when you end up at the dashboard at the end, the breach should be moved to the "Fixed" tab, whereas on `main`, it would still be on the "Active" tab until the page is refreshed.

- [ ] I've added a unit test to test for potential regressions of this bug. - Would only test that `router.refresh` is called, which wouldn't really catch bugs.
- [ ] All acceptance criteria are met. - I don't think all bugs of MNTOR-2464 are resolved yet.

### DIFF
--- a/src/app/(proper_react)/(redesign)/(authenticated)/user/(dashboard)/dashboard/fix/high-risk-data-breaches/HighRiskBreachLayout.tsx
+++ b/src/app/(proper_react)/(redesign)/(authenticated)/user/(dashboard)/dashboard/fix/high-risk-data-breaches/HighRiskBreachLayout.tsx
@@ -110,6 +110,13 @@ export function HighRiskBreachLayout(props: HighRiskBreachLayoutProps) {
         );
       }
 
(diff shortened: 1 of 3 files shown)


flozia (approved):
Seems to work as expected!

PR #4083, opened 2024-01-24
Move dummy subscription pricing to env variable (MNTOR-2154)

flozia:
# References: 
Jira: [MNTOR-2154](https://mozilla-hub.atlassian.net/browse/MNTOR-2154)

# Description

Moves the subscription pricing info into env variables.

[MNTOR-2154]: https://mozilla-hub.atlassian.net/browse/MNTOR-2154?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/app/(proper_react)/(redesign)/(authenticated)/admin/feature-flags/page.tsx
```diff
@@ -105,6 +105,14 @@ export default async function FeatureFlagPage() {
             user={session.user}
             monthlySubscriptionUrl={monthlySubscriptionUrl}
             yearlySubscriptionUrl={yearlySubscriptionUrl}
+            subscriptionBillingAmount={{
```
Vinnl, line 108: A bit annoying now after the fact, so feel free to skip, but I wonder if we could do this in a separate function? That way, we have the (technically unsafe) type assertion in a single place, and we can add `import "server-only";` to catch situations where we're accidentally referencing a variable that doesn't exist on the client side.

Vinnl (approved):

src/app/(proper_react)/(redesign)/(authenticated)/admin/feature-flags/page.tsx
flozia: Good point — done in [526db2d](https://github.com/mozilla/blurts-server/pull/4083/commits/526db2db7c47a705d103f2acd49d2c8dae56868e).

src/app/functions/server/getPremiumSubscriptionInfo.ts
```diff
@@ -32,4 +34,17 @@ function getPremiumSubscriptionUrl({
   return `${subscriptionUrl}/products/${productId}?plan=${planId}`;
 }
 
-export default getPremiumSubscriptionUrl;
+type GetSubscriptionBillingAmount = {
+  [key in SubscriptionPeriod]: number;
+};
```
Vinnl, line 39: Useful shorthand:

```suggestion
type GetSubscriptionBillingAmount = Record<SubscriptionPeriod, number>;
```

(Also I'd maybe remove the `Get` from this type, since it's not a function.)

Vinnl (approved):

PR #4085, opened 2024-01-25
Add local admin interface to administer users

Vinnl:
This provides a simple frontend for the API `/api/v1/admin/users/[sha1]/`, allowing the addition and removal of Plus status and scans, to make problems there easier to debug.

It is only accessible when running locally (i.e. when `APP_ENV="local"`), at `/admin/dev/`.


mansaj (approved):
awesome interface

PR #4094, opened 2024-01-26
Don’t always redirect to /user/dashboard after auto sign-in

flozia:
# References: 
Jira: [MNTOR-2820](https://mozilla-hub.atlassian.net/browse/MNTOR-2820)

# Description

Send users to the requested route after authentication that has been triggered by `AutoSignIn`.

# How to test

1. Visit `/` and make sure you are not authenticated.
2. Navigate to a route that triggers auto sign-in: Enter for example `/user/dashboard/fix` into the address bar.
3. You should not get redirected to `/dashboard` anymore.

[MNTOR-2820]: https://mozilla-hub.atlassian.net/browse/MNTOR-2820?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


rhelmer (approved):

src/app/components/client/AutoSignIn.tsx
```diff
@@ -9,7 +9,7 @@ import { ReactNode, useEffect } from "react";
 
 export const AutoSignIn = (): ReactNode => {
   useEffect(() => {
-    void signIn("fxa", { callbackUrl: "/user/breaches" });
+    void signIn("fxa");
```
mansaj, line 12: without the callbackUrl, does it just go back to the original url?
I know we have another sign in button component. Is the goal to keep that one with redirect?

flozia: Yes, without `callbackUrl` it redirects to the original URL. This `AutoSignIn` component is triggering sign-in when visiting a route that needs authentication. The `SignInButton` is an actual button that we use on the landing page and should direct to the dashboard.

PR #4096, opened 2024-01-26
Allow linking directly to the FAQ section

Vinnl:
This was listed in the acceptance criteria of MNTOR-2650 (https://github.com/mozilla/blurts-server/pull/3963), but understandably forgotten.


src/app/(proper_react)/(redesign)/(public)/Faq.tsx
```diff
@@ -85,7 +85,7 @@ export const FaqSection = ({
   };
 
   return (
-    <div className={styles.faqWrapper}>
+    <div id="faq" className={styles.faqWrapper}>
       <b className={styles.faqTitle}>
```
pdehaan, line 89: Unrelated, and delightfully off-topic, but&hellip; I was surprised this is just a vanilla `<b>` instead of an `<h2>`.

For example, this seems to be Metropolis/20px/600:
```html
<b class="Faq_faqTitle__Bwyvr">Frequently asked questions</b>
```

But elsewhere above, the similar headings seem to be Inter/24px/700:
```html
<h2>Here’s how we help protect you</h2>
```

And below seems to be Inter/20px/500:
```html
<p class="LandingView_title__1hAnl">Take back control of your data</p>
```

flozia (approved):

src/app/(proper_react)/(redesign)/(public)/Faq.tsx
Vinnl: Yep, I just noticed that too while doing heading navigation with my screen reader. Hopefully I'll find the time at some point to update them.

PR #4097, opened 2024-01-26
MNTOR-2832/fix ga4 debug mode

rhelmer:
# References: 
Jira: MNTOR-2832

# Description

Set GA4 debug mode at build time - currently it's always on. We could alternatively fetch info from the serve to determine this, but I think compiling it in is fine, it's not something that's going to change at runtime.

# How to test

`Initialize GA4` will be logged to the browser console if debug mode is on, and packets will only show up in the "DebugView" in GA4.


mansaj (approved):

src/app/hooks/useGa.ts
```diff
@@ -51,9 +51,7 @@ export const useGa = (): {
     record: (options: Ga4EventOptions) => void;
   };
 } => {
-  const debugMode =
-    process.env.NEXT_PUBLIC_NODE_ENV !== "production" &&
-    process.env.NODE_ENV !== "test";
+  const debugMode = process.env.NEXT_PUBLIC_GA4_DEBUG_MODE !== "true";
```
pdehaan, line 54: I'm confused, should this be `!==` or `===`?
Seems odd that it's `debugMode` where `_GA4_DEBUG_MODE !== true` since those sound conflicting.

rhelmer: 🤦 thanks

src/app/hooks/useGa.ts
```diff
@@ -52,7 +52,7 @@ export const useGa = (): {
   };
 } => {
   const debugMode =
-    process.env.NEXT_PUBLIC_NODE_ENV !== "production" &&
+    process.env.NEXT_PUBLIC_GA4_DEBUG_MODE === "true" &&
     process.env.NODE_ENV !== "test";
```
rhelmer, line 56: The tests depend on GA not being in debug mode.

PR #4100, opened 2024-01-28
conditionally start webserver for when prompted only

mozrokafor:
adding the ability to conditionally run the webserver in playwright test framework. The webserver is only needed for the PR runs(or if prompted by runner locally).


src/e2e/specs/landing.spec.ts-snapshots/stage-chooseLevelSection-firefox-linux.png
Vinnl, line 1: Were these committed accidentally?

playwright.config.js
```diff
@@ -12,6 +12,16 @@ dotenv.config()
...
  */
+
+const webServerConfig = {
+  command: 'npm run build; npm start',
+  // Building the app can take some time:
+  timeout: 600_000,
+  port: 6060
+}
+
+const shouldStartWebServer = process.env.E2E_TEST_RUN_WEBSERVER !== "false"
```
Vinnl, line 23: Wouldn't this cover all relevant situations? (Possibly also using it to set `reuseExistingServer`.)

```suggestion
const shouldStartWebServer = process.env.E2E_TEST_ENV === "local"
```

Vinnl (approved):
One inline question about an alternative approach, but this looks fine too.

playwright.config.js
mozrokafor: i thought about this but i figured this would also help the cron job run faster if webserver build is not happening

src/e2e/specs/landing.spec.ts-snapshots/stage-chooseLevelSection-firefox-linux.png
mozrokafor: needed to update the screens for linux but i guess it woul probably be better in a separate pr

PR #4114, opened 2024-01-29
address the different redirect after sign up for local and other envs.

mozrokafor:
Chore: the sign up redirect is different for local than it is for other environments, this change should handle addressing it for local vs others

### DIFF
--- a/src/e2e/specs/auth.spec.ts
+++ b/src/e2e/specs/auth.spec.ts
@@ -27,7 +27,11 @@ test.describe(`${process.env.E2E_TEST_ENV} - Authentication flow verification @s
     await authPage.signUp(randomEmail, page);
 
     // assert successful login
-    expect(page.url()).toBe(`${process.env.E2E_TEST_BASE_URL}/user/dashboard`);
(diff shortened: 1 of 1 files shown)


src/e2e/specs/auth.spec.ts
```diff
@@ -27,7 +27,11 @@ test.describe(`${process.env.E2E_TEST_ENV} - Authentication flow verification @s
     await authPage.signUp(randomEmail, page);
 
     // assert successful login
-    expect(page.url()).toBe(`${process.env.E2E_TEST_BASE_URL}/user/dashboard`);
+    const successUrl =
+      process.env.E2E_TEST_ENV === "local"
```
pdehaan, line 31: Why is it redirect different? Is it a recent PR that just hasn't landed on Stage/Prod yet?

mozrokafor: nope, there's some flags we don't have in stage that we have in local(manual managed by the team currently)

pdehaan (approved):
👍

PR #4115, opened 2024-01-29
chore: clear rebrand banner before test run for the screen comparisons

mozrokafor:
chore: clear the rebrand banner on the landing page before image comparison tests are ran for landing page


mansaj (approved):

src/e2e/specs/landing.spec.ts
```diff
@@ -5,6 +5,7 @@
 import { test, expect } from "../fixtures/basePage.js";
 import { defaultScreenshotOpts } from "../utils/helpers.js";
 
+test.describe.configure({ mode: "parallel" });
```
pdehaan, line 8: Q: What benefit does adding parallel mode here have?

src/e2e/specs/landing.spec.ts
```diff
@@ -90,13 +91,24 @@ test.describe(`${process.env.E2E_TEST_ENV} - Verify the Landing Page content`, (
...
     });
 
+    // temp fix to clear rebrand banner before image comparison
+    try {
+      const rebrandBanner = page.getByTitle("dismiss");
+      if (rebrandBanner) {
+        await rebrandBanner.click();
+      }
+    } catch (error) {
+      console.log("[E2E_Log - No Rebrand Banner, continuing...]");
```
pdehaan, line 109: Not sure if we need a `console.log()` polluting logs or if we can just ignore potential errors silently, but I guess the larger question is should this be abstracted higher up so we can use it in more places?
Or "sssh, no" is a very valid answer if this is all just temporary and the rebrand banners will be going away in X weeks/months/years.

src/e2e/specs/landing.spec.ts
```diff
@@ -90,13 +91,24 @@ test.describe(`${process.env.E2E_TEST_ENV} - Verify the Landing Page content`, (
...
       description:
         "https://testrail.stage.mozaws.net/index.php?/cases/view/2463517",
     });
 
+    // temp fix to clear rebrand banner before image comparison
+    try {
+      const rebrandBanner = page.getByTitle("dismiss");
+      if (rebrandBanner) {
+        await rebrandBanner.click();
+      }
```
pdehaan, line 107: Also noticed we have a `.toHaveScreenshot()` below on L125, but not sure if that needs the banner dismiss logic too.
```sh
git grep -Pn "\.toHaveScreenshot\(" | cat

src/e2e/specs/landing.spec.ts:112:    await expect(landingPage.chooseLevelSection).toHaveScreenshot(
src/e2e/specs/landing.spec.ts:125:    await expect(landingPage.faqSection).toHaveScreenshot(
```

mozrokafor: hmm, i believe it'll need it as the banner's position in flow of the script is dynamic thus will fail image pixel comparisons.

src/e2e/specs/landing.spec.ts
mozrokafor: essentially, pw runs tests in same spec file serially, if this flag is added the tests are ran parallel thus speeding up the test

PR #4122, opened 2024-01-30
Don't run a manual scan for a Plus user

Vinnl:
**Submitting for input on the general approach, and on how to test. Will fix the test later, but figured I'd submit this so people know this code exists.**

# References: 
Jira: MNTOR-2865
Figma: N/A

# Description

If the user enters the onboarding while having purchased Plus beforehand, with this change we no longer start a manual scan. Instead, we activate the profile, wait for the `initial` scan to come in, and then start the opt-out process when they come in.

I thought this might have been one of the reasons users would see "Action needed" scans even if they had Plus: those would simply be the results of a manual scan that we didn't request an opt-out for.

The main reason is probably the delay between OneRep calling our webhook and us starting the opt-out process, but I figured dealing with this flow is probably a good idea anyway.

# How to test

Unfortunately I'm not sure how best to test; specifically, we'd need to simulate OneRep calling our webhook while the scan is in progress. Any ideas? Maybe that's something to add to the dev dashboard as well?

### DIFF
--- a/src/TestComponentWrapper.tsx
+++ b/src/TestComponentWrapper.tsx
@@ -21,7 +21,7 @@ export const TestComponentWrapper = (props: { children: ReactNode }) => {
             process.env.STORYBOOK === "true" ? "storybook" : "test",
         }}
       >
-        <SessionProvider session={null}>
+        <SessionProvider session={{ user: { email: "example@example.com" } }}>
           <ReactAriaI18nProvider locale="en">
             {props.children}
           </ReactAriaI18nProvider>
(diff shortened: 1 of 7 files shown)


rhelmer:
> Unfortunately I'm not sure how best to test; specifically, we'd need to simulate OneRep calling our webhook while the scan is in progress. Any ideas? Maybe that's something to add to the dev dashboard as well?

I've started working on a little mock server for scans in https://github.com/mozilla/blurts-server/tree/wip-refactor-and-dev-tools, one thing I considered early on was having buttons on the dashboard to trigger different types of scans. Right now the server doesn't really have any permanent state, but if it did then I think this would be pretty easy to provide.

rhelmer (approved):

rhelmer:
@Vinnl are you still working on this one?

Vinnl:
@rhelmer I was thinking of closing this, since it doesn't seem to be a pressing issue anymore, especially when we move off OneRep - though perhaps it relates to the reports of entries staying stuck in "In progress"? Though I don't think it will do anything for existing users.

I'll close it for now, but let me know if you feel it makes sense to spend some more time on this.

PR #4123, opened 2024-01-30
Mark new scan results as in progress for Plus users

Vinnl:
**Submitting for input on the general approach, and on how to test. Will fix the test later, but figured I'd submit this so people know this code exists.**

# References: 
Jira: MNTOR-2865
Figma: N/A

# Description

Even if the user has Plus, OneRep won't automatically start removing found exposures; it first sends a request to our webhook, and then the webhook sends an opt-out request to OneRep. Meanwhile, however, we're just waiting for the systems to do their thing, and there's no action for the user to take; hence, this change also marks the exposures as being in progress.

# How to test

With a new Plus user, you should not see "Action needed" "info for sale" cards on the dashboard after completing the free scan flow.

### DIFF
--- a/src/app/(proper_react)/(redesign)/(authenticated)/user/(dashboard)/dashboard/Dashboard.test.tsx
+++ b/src/app/(proper_react)/(redesign)/(authenticated)/user/(dashboard)/dashboard/Dashboard.test.tsx
@@ -445,7 +445,7 @@ it("switches between tab panels", async () => {
 
 it("shows consistent counts in the chart on the active tab", () => {
   const ComposedDashboard = composeStory(
-    DashboardUsPremiumUnresolvedScanUnresolvedBreaches,
+    DashboardUsNoPremiumUnresolvedScanUnresolvedBreaches,
     Meta,
   );
   render(<ComposedDashboard />);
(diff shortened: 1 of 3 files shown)


rhelmer:
@Vinnl do you want me to re-assign the Jira ticket? I came up with a similar solution FWIW https://github.com/mozilla/blurts-server/pull/new/MNTOR-2865/mark-new-scans-in-progress-plus

rhelmer (approved):
This is the same approach that I came to - leave the data alone and just show it as `optout_in_progress` in the view. As long as it shows up in "Action Needed" if the user unsubscribed then I think this is good enough for now. We probably want some way in the future to allow manually resolving brokers that never get auto-resolved but that can be its own separate exploration.

Vinnl:
@rhelmer Yeah I guess I'll switch the tickets around, as I haven't been able to reproduce the other behaviour anyway. I don't see any changes in your link though, maybe you forgot to push?

Vinnl:
Suppressed the coverage error with pointers from @flozia in 20d2a6f2c - although I think I shouldn't be referring to the same case twice, so I think I'll have to update one of those pointers.

PR #4124, opened 2024-01-30
Fix description for legacy users in metrics.yaml

rhelmer:
# Description

Correction for the description of the `legacy_user` key for `dashboard.view` in `metrics.yaml`. I went ahead and re-generated the docs too (maybe we should make this a CI job).

@codemist @Vinnl @flozia I don't see this instrumented, except to determine the `utm_campaign` setting here https://github.com/mozilla/blurts-server/blob/main/src/app/(proper_react)/(redesign)/(authenticated)/user/welcome/View.tsx#L56 - do we have this instrumented?


Vinnl (approved):
I don't recall it being instrumented anywhere, though I haven't reviewed all telemetry PRs.

If `metrics.md` is autogenerated, shouldn't we exclude it from the repo?

PR #4130, opened 2024-01-31
Add missing dashboard view event

rhelmer:
# Description

I noticed when reviewing with the data team that we're missing the `dashboard.view` event.


mansaj (approved):

mansaj:
@rhelmer pending build failures FYI

rhelmer:
> @rhelmer pending build failures FYI

Oops, I think `userId` should be optional for the dashboard `View` for now... we can tighten that up later.

PR #4135, opened 2024-02-01
MNTOR-2886/feature flag waitlist

rhelmer:
# References: 
Jira: MNTOR-2886

# Description

This creates a `MonitorPlus` feature flag which forces the waitlist functionality on (the one that normally kicks in automatically when we're at capacity).

We seem to be defining the quota variables both in the environment and also in `constants.ts`, I wonder if we could consolidate. It also seems that subscription from the dashboard still works, I can address that in this PR.

It would be nice to actually disable fetches from the data provider on the server too, but I don't want to try to tackle that one here, I'll file a separate ticket for that.

# Screenshot (if applicable)

Not applicable.

# How to test


src/app/(proper_react)/(redesign)/(authenticated)/user/(dashboard)/dashboard/DashboardTopBanner/DashboardTopBannerContent.tsx
```diff
@@ -223,7 +223,8 @@ export const DashboardTopBannerContent = (props: DashboardTopBannerProps) => {
               )}
             </p>
             <div className={styles.cta}>
-              {typeof props.totalNumberOfPerformedScans === "undefined" ||
+              {!props.enabledFeatureFlags.includes("MonitorPlus") ||
```
mansaj, line 226: I think another name might be more appropriate? Maybe something like `ForceWaitlist`.. `MonitorPlus` might be too generic/confusing

mansaj, line 226: I think another name might be more appropriate? Maybe something like `ForceWaitlist`.. `MonitorPlus` might be too generic/confusing

rhelmer: Yeah good point, especially because we already have "free scan" and "premium removal" flags.

PR #4138, opened 2024-02-02
Switch to third-party GA4

rhelmer:
# Description

GA4 loads, but we're not seeing data flowing. There appears to be a problem with the way we're loading the script. While I was investigating a fix, I noticed there's a `@next/third-parties/google` module that we can just drop-in, it's documented here:

https://nextjs.org/docs/app/building-your-application/optimizing/third-party-libraries#google-analytics

I left the `useGa` hook largely alone, but I did drop a note in there that we could replace a lot of the `window.gtag` setup with the `sendGAEvent` included in the `@next/third-parties/google` module. The one thing I had to do was move the check for `doNotTrack` to the hook, since the script is loaded in a server component now (in `layout.tsx` as recommended by the docs).

With this in place, I now see data flowing in the network tab in browser devtools.


src/app/hooks/useGa.ts
```diff
@@ -20,16 +16,21 @@ interface InitGaProps {
...
+    console.info("doNotTrack enabled, not initializing GA4");
+    return;
   }
 
   // GA4 setup
   window.dataLayer = window.dataLayer || [];
   if (!window.gtag) {
     window.gtag = function (...args: unknown[]) {
-      window.dataLayer.push(args);
+      window.dataLayer?.push(args);
```
pdehaan, line 33: Why the `?.` here if we define it above on L30?

src/app/hooks/useGa.ts
```diff
@@ -81,6 +82,12 @@ export const useGa = (): {
   return {
     gtag: {
       record: (options) => {
+        // Guard against this running server-side.
+        /* c8 ignore next 4 */
+        if (typeof window === "undefined") {
+          console.warn("window is not defined");
```
pdehaan, line 88: Where toes this `console.warn()` go? User's Devtools? Or server logs?
Or more importantly, do we care, or is this just flooding logs?

src/app/hooks/useGa.ts
rhelmer: I'm not sure, the typescript compiler claims that it might be empty. I intend to replace most of the code in here per the TODO though, I just wanted to try to make the current fix as minimal as possible.

src/app/hooks/useGa.ts
rhelmer: It goes to server logs, and it is an issue we want to fix if it shows up. Not enough to throw an exception and crash, but enough that we should figure out why a server component is trying to use this.

flozia (approved):

src/app/layout.tsx
```diff
@@ -69,6 +70,7 @@ export default async function RootLayout({
           <SessionProvider session={session}>{children}</SessionProvider>
         </PublicEnvProvider>
       </body>
+      <GoogleAnalytics gaId={CONST_GA4_MEASUREMENT_ID} />
```
Vinnl, line 73: We can still prevent loading the script in the first place when DoNotTrack is enabled using `next/headers`:

```suggestion
      {headers().get("DNT") !== "1" && (<GoogleAnalytics gaId={CONST_GA4_MEASUREMENT_ID} />)}
```

(And maybe we should also check that the `Sec-GPC` header is not `"1"`.)

Vinnl (approved):

src/app/layout.tsx
rhelmer: Oh good point, I forgot this is set as a header and not just available via `navigator` in JS. Thanks!

rhelmer: I think we have a ticket for GPC, this is very last-minute so I'd like to keep it simple. I'll file one if not

PR #4153, opened 2024-02-05
Prevent mobile shell from causing body to overflow

flozia:
# References: 
Jira: [MNTOR-1975](https://mozilla-hub.atlassian.net/browse/MNTOR-1975)

# Description

I ran into this issue quite a lot while testing on Chrome. This prevents the mobile shell from over-scrolling. Setting `position: relative` ensures that the child elements don’t overflow.

[MNTOR-1975]: https://mozilla-hub.atlassian.net/browse/MNTOR-1975?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/app/(proper_react)/(redesign)/MobileShell.module.scss
```diff
@@ -67,6 +67,7 @@
   // https://developer.mozilla.org/en-US/docs/Web/CSS/position#syntax
   overflow: auto;
   height: 100%;
+  position: relative;
```
Vinnl, line 70: I can't quite follow _why_ this prevents child elements from overflowing. Could you explain that (in a comment)?

Vinnl: @flozia Looking at the latest comment, it still doesn't tell me _why_ - do you have a link to documentation or something how `position: relative` interacts with whatever is causing the bug? (And also, I see you added a `z-index`, which are always good to accompany with a comment describing which elements it should be overlapping, so future readers know whether it can/should be changed.)

flozia: I’m not entirely sure if this is a bug with WebKit based browsers or an invalid combination of styling properties on our side. It seems like somehow the height of `body` takes the children of the elements in `.nonHeader` into account and causes it to get its overflow scroll. When hiding the children of `.nonHeader` with `* { display: none; }` it keeps its size (because of `height: 100%`) and there is no overflow scroll.

Added a comment in [9b08da8](https://github.com/mozilla/blurts-server/pull/4153/commits/9b08da8830ffb3337235bacbb113e3319c4ce5a2).

flozia: Expanded on my previous comment about the use of `position: absolute` in [2d22521](https://github.com/mozilla/blurts-server/pull/4153/commits/2d2252163bfe349fd7773727bfd82699b35cb726).

Vinnl: This is the perfect comment :pinched_fingers: Thank you!

Vinnl (approved):

PR #4169, opened 2024-02-07
Fix failing e2e tests

Vinnl:
The E2E tests are still failing even though auth should be stable again. The following error in the logs is attached to the `waitForNavigation` calls removed in this PR:

> Error: page.waitForNavigation: net::ERR_TOO_MANY_REDIRECTS

Additionally, that API is deprecated as a source of raciness, and it didn't seem to be necessary anyway.

I've also split up the password selectors for signing in and signing up, since they're different forms with different labels, and thus trying to come up with a selector that matched both is a losing battle.

So let's see if this is enough to make them succeed again.

### DIFF
--- a/src/e2e/pages/authPage.ts
+++ b/src/e2e/pages/authPage.ts
@@ -2,66 +2,45 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
-import { Locator, Page } from "@playwright/test";
+import { Page } from "@playwright/test";
 import { getVerificationCode } from "../utils/helpers.js";
 
 export class AuthPage {
   readonly page: Page;
-  readonly emailInputField: Locator;
-  readonly passwordInputField: Locator;
-  readonly passwordConfirmInputField: Locator;
-  readonly continueButton: Locator;
-  readonly ageInputField: Locator;
-  readonly verifyCodeInputField: Locator;
 
   constructor(page: Page) {
     this.page = page;
-    this.emailInputField = page.locator('input[name="email"]');
-    this.passwordInputField = page.locator('[type="password"]').nth(0);
-    this.passwordConfirmInputField = page.locator('[type="password"]').nth(1);
-    this.ageInputField = page.getByLabel("How old are you?");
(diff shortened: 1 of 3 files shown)


src/e2e/pages/authPage.ts
```diff
@@ -2,68 +2,46 @@
...
-  readonly verifyCodeInputField: Locator;
 
   constructor(page: Page) {
     this.page = page;
-    this.emailInputField = page.locator('input[name="email"]');
-    this.passwordInputField = page.locator('[type="password"]').nth(0);
-    this.passwordConfirmInputField = page.locator('[type="password"]').nth(1);
-    this.ageInputField = page.getByLabel("How old are you?");
-    this.continueButton = page.locator('[type="submit"]').first();
-    this.verifyCodeInputField = page.locator("div.card input");
```
mozrokafor, line 24: QQ: whats the thought behind the removals of the locators here?

Vinnl: See the PR description:

> I've also split up the password selectors for signing in and signing up, since they're different forms with different labels, and thus trying to come up with a selector that matched both is a losing battle.

The idea was that these locators were being re-used across both sign-in and sign-up, even though the pages looked pretty different. So we had a locator like `[type="submit"]` that just _happened_ to match primary buttons, even though those buttons were all different ("Sign up or sign in", "Confirm", "Sign in", ...). So they looked like the wrong abstraction, and I split them up when trying to figure out why authentication was failing.

mozrokafor: agreed. as part of the clean up ticket will be to move to types of locators, i guess i'm trying pick your brain about whether the locators list was useful for you when reading the test code in terms of having the elements in one area and being able to access it outside of the class

Vinnl: Ah, if you're asking for my opinion specifically (i.e. with the caveat that it's just an opinion), then I'd prefer not having to jump around between the locators and the test code. Especially when selectors like `getByRole` and `getByLabel` are used (i.e. with the actual text that's shown on the page), they're pretty easy to read anyway.

Functions for specific, repeated steps, like `signIn` and `signUp`, are pretty useful though.

PR #4172, opened 2024-02-07
e2e smoke redirect fix

mozrokafor:
last piece of the e2e-smoke fix, removing the waitfornavigation method

### DIFF
--- a/src/e2e/specs/auth.spec.ts
+++ b/src/e2e/specs/auth.spec.ts
@@ -3,7 +3,6 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
 import { test, expect } from "../fixtures/basePage.js";
-import { setEnvVariables } from "../utils/helpers.js";
 
 test.describe(`${process.env.E2E_TEST_ENV} - Authentication flow verification @smoke`, () => {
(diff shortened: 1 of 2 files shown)


src/e2e/specs/auth.spec.ts
```diff
@@ -41,12 +38,8 @@ test.describe(`${process.env.E2E_TEST_ENV} - Authentication flow verification @s
     // wait for scanning to complete
     let wait = 0;
     let scanning = page.url() !== successUrl;
-    while (wait < 60 && scanning) {
-      try {
-        scanning = page.url() !== successUrl;
-      } catch (error) {
-        console.log("line 45");
-      }
+    while (wait < 300 && scanning) {
```
pdehaan, line 41: What's this `wait < 300`?
I see we init to `0` above on L39, and increment on L44, but with a 1s pause, that's 300s or 5 minutes of potential pausing before the test fails?
Why the 5&times; increase from 60s timeout in the old code?

mozrokafor: this PR is pretty much a debugging exercise, there's a redirect issue that i'm looking at, will clean this up once we get to a solid solution

mozrokafor:
handled in another pr

PR #4174, opened 2024-02-08
Fix e2e after Plus launch

flozia:
Fixes the `e2e` test for now by replacing the deprecated Playwright method `waitForNavigation` and setting `BROKER_SCAN_RELEASE_DATE` temporarily to a future date until we update the test cases to reflect the new flow.

### DIFF
--- a/.env-dist
+++ b/.env-dist
@@ -139,7 +139,7 @@ SUBSCRIPTION_BILLING_AMOUNT_MONTHLY_US=42.42
 # was released to the welcome flow. Users who had signed up before and thus
 # have seen data breach results before, will be able to see their known breaches
 # first:
-BROKER_SCAN_RELEASE_DATE=2024-02-06
+BROKER_SCAN_RELEASE_DATE=2025-02-06
 
 MONTHLY_SUBSCRIBERS_QUOTA=
 MONTHLY_SCANS_QUOTA=
--- a/.env-dist
+++ b/.env-dist
@@ -139,7 +139,7 @@ SUBSCRIPTION_BILLING_AMOUNT_MONTHLY_US=42.42
 # was released to the welcome flow. Users who had signed up before and thus
 # have seen data breach results before, will be able to see their known breaches
 # first:
-BROKER_SCAN_RELEASE_DATE=2025-02-06
+BROKER_SCAN_RELEASE_DATE=2024-02-06
 
 MONTHLY_SUBSCRIBERS_QUOTA=
 MONTHLY_SCANS_QUOTA=
--- a/src/e2e/pages/authPage.ts
+++ b/src/e2e/pages/authPage.ts
@@ -36,7 +36,7 @@ export class AuthPage {
 
   async enterEmail(email: string) {
     await this.emailInputField.fill(email);
-    await this.continue({ waitForURL: "**/accounts.stage.mozaws.net/**" });
+    await this.continue({ waitForURL: "**/oauth/**" });
   }
 
   async enterPassword() {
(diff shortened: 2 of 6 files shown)


.env-dist
```diff
@@ -139,7 +139,7 @@ SUBSCRIPTION_BILLING_AMOUNT_MONTHLY_US=42.42
 # was released to the welcome flow. Users who had signed up before and thus
 # have seen data breach results before, will be able to see their known breaches
 # first:
-BROKER_SCAN_RELEASE_DATE=2024-02-06
+BROKER_SCAN_RELEASE_DATE=2025-02-06
```
Vinnl, line 142: This will make it the default for our actual deployments as well. The variable should be set there, but it's still riskier than I would like. Could we set this in just for the E2E test runs? i.e. like [here](https://github.com/mozilla/blurts-server/pull/4169/files#diff-6175e3066c9d2ba5a9bd6846a8aaf039cbe9ba5554558e06ac273bc295a8c627R82-R84) (and maybe also in `e2e_cron.yml`? Though that might need more work, and the tests should be updated to not need this variable change anyway.).

flozia:
@mozrokafor We adopted the relevant changes that are also resolving the test failures in #4169 as well — up to you what you prefer for now.

cc @Vinnl

src/e2e/pages/authPage.ts
```diff
@@ -62,7 +60,7 @@ export class AuthPage {
       process.env.E2E_TEST_ACCOUNT_PASSWORD as string,
     );
     await this.ageInputField.type("31");
-    await this.continue();
+    await this.continue({ waitForURL: "**/accounts.stage.mozaws.net/**" });
```
Vinnl, line 63: From my testing, this is the only place that actually needs a `waitForURL`, so maybe we could remove it from the `continue`? It's just the bit where we get redirected back to the Monitor that is hard for Playwright.

We can also avoid hardcoding the FxA stage URL here by waiting for an additional redirect, like this:

```js
await this.page.waitForURL(`${process.env.E2E_TEST_BASE_URL}/**`);
```

src/e2e/specs/auth.spec.ts
```diff
@@ -27,10 +27,7 @@ test.describe(`${process.env.E2E_TEST_ENV} - Authentication flow verification @s
     await authPage.signUp(randomEmail, page);
 
     // assert successful login
-    const successUrl =
-      process.env.E2E_TEST_ENV === "local"
-        ? "/user/dashboard"
-        : "/user/welcome";
+    const successUrl = "/user/dashboard";
```
Vinnl, line 30: Is this still necessary? Eventually, the tests will need to be updated to account for the welcome page as well, I imagine.

Actually, I guess we'll want tests for both the "pre-redesign user" and "new user" experience.

Vinnl (approved):
Great job Florian, so happy to see them working again! Left a couple of things that I think would be good to change, and then let's let Raphael decide whether to merge this one or https://github.com/mozilla/blurts-server/pull/4169.

.env-dist
flozia: In that case, that’s too risky indeed. Updated to apply the same change you linked in [2f2344c](https://github.com/mozilla/blurts-server/pull/4174/commits/2f2344c7c1a4a8cf8e03fd813bf064cd6b64a927)

src/e2e/pages/authPage.ts
flozia: That works! I put them there to make sure this is the source of failure while debugging in CI — updated in [eb173c9](https://github.com/mozilla/blurts-server/pull/4174/commits/eb173c9828365e8f15dc7a42f030655a77317545).

src/e2e/specs/auth.spec.ts
flozia: I think this needs to be updated either way but should be good to leave this as is for now. Also updated in [eb173c9](https://github.com/mozilla/blurts-server/pull/4174/commits/eb173c9828365e8f15dc7a42f030655a77317545).

src/e2e/pages/authPage.ts
flozia: Had to revert and add those back in to have them work in CI.

Vinnl: Huh, that's weird; I only have them when redirecting back to Monitor, and the CI succeeds for my PR :thinking:

PR #4178, opened 2024-02-08
MNTOR-2925 - add a function for checking the session and logging more…

rhelmer:
… diagnostics

# References: 
Jira: MNTOR-2925 

# Description

Start to move to a common function to check if the session is valid, and log a more detailed reason if it is not. I've implemented in the least-invasive way, but I'd like to move all of these checks to a common function as a next step once we determine the problem.


mansaj (approved):
LGTM, reminder to revert this after diagnosis

PR #4180, opened 2024-02-09
MNTOR-2934/set UPLOAD_SENTRY_SOURCEMAP correctly

rhelmer:
# References: 
Jira: MNTOR-2934

# Description

When tags are applied, upload sourcemaps to Sentry using GH Action.

# How to test

We want this only to run for releases, we can verify this in the "Source Maps" settings of Sentry's web UI. 


next.config.js
```diff
@@ -9,8 +9,10 @@ import { withSentryConfig } from "@sentry/nextjs";
 const nextConfig = {
   productionBrowserSourceMaps: true,
   sentry: {
-    disableServerWebpackPlugin: process.env.UPLOAD_SENTRY_SOURCEMAPS ?? "false",
-    disableClientWebpackPlugin: process.env.UPLOAD_SENTRY_SOURCEMAPS ?? "false",
+    disableServerWebpackPlugin:
+      process.env.UPLOAD_SENTRY_SOURCEMAPS === "true" ? false : true,
```
pdehaan, line 13: Guessing we wouldn't want something like `process.env.UPLOAD_SENTRY_SOURCEMAPS !== "true"` and avoiding the ternery.

<s>And probably a temp variable is overkill too if you wanted
```js
const uploadSentrySourcemaps = process.env.UPLOAD_SENTRY_SOURCEMAPS === "true";

disableServerWebpackPlugin: !uploadSentrySourcemaps,
disableClientWebpackPlugin: !uploadSentrySourcemaps,
```
</s>

pdehaan: PS: let me know if you want me to quickly R+ this if you want to get it onto Stage/Prod right meow and just want a blind 👍 to appease the CI Gods.

rhelmer: Good idea, thanks! I figured out a way to test this before landing (I could test some it locally, but I want to make sure it's working when run from CircleCI), the feedback cycle is so long if we have to push to stage to test: https://github.com/mozilla/blurts-server/pull/4179

rhelmer: (so no rush to stage but I appreciate the offer!)

pdehaan: No problem. Let the record show I have poor ethics and am willing to blind R+ at 9PM for the sake of :yolo: based development (and the LOLs). 🤷

rhelmer:
I can tell from https://github.com/mozilla/blurts-server/pull/4179 this won't work as-is, and we shouldn't pass secrets this way either (`ARG`s get preserved in the docker layer that gets pushed). [ Docker build secrets](https://docs.docker.com/build/building/secrets/) are pretty difficult to set up in this environment as well.

I think it might be easier to upload the symbols from the host environment, which has the sentry auth token set in a CircleCI environment variable. It looks like Sentry can be configured to do that.

rhelmer:
> I can tell from #4179 this won't work as-is, and we shouldn't pass secrets this way either (`ARG`s get preserved in the docker layer that gets pushed). [ Docker build secrets](https://docs.docker.com/build/building/secrets/) are pretty difficult to set up in this environment as well.
> 
> I think it might be easier to upload the symbols from the host environment, which has the sentry auth token set in a CircleCI environment variable. It looks like Sentry can be configured to do that.

I've researched this and while it's definitely possible, it requires more changes (e.g. `Sentry.init` needs to have the release passed to it) - I'd like to let Sentry handle this as part of the build with its webpack plugin if at all possible. I think we might be able to use the GitHub Action runner to do this, although I don't know if there's any potential issues if the sourcemaps come from a different build than what was actually shipped - it seems to work fine in my testing but it feels wrong to be doing multiple builds like this.

.github/workflows/build.yaml
```diff
@@ -18,3 +18,13 @@ jobs:
       - run: cp .env-dist .env
       # Verify that the build (incl. type-checking) succeeds
       - run: npm run build
+        env:
+          UPLOAD_SENTRY_SOURCEMAPS: false
+        if: ${{ !startsWith(github.ref, 'refs/tags/') }}
```
pdehaan, line 23: Hypothetically speaking, if I didn't know what exactly this was doing/checking... what is this doing/checking?
Only uploading sourcemaps in stage/prod but not local/Heroku?

next.config.js
```diff
@@ -190,7 +190,7 @@ const sentryWebpackPluginOptions = {
 
   org: "mozilla",
   project: "firefox-monitor",
-  silent: true, // Suppresses all logs
+  silent: false, // Suppresses all logs
```
pdehaan, line 193: Also, why this change? Was this related to sourcemaps, or was for debugging?

.github/workflows/build.yaml
rhelmer: Yes that's right... I'll add a comment.

next.config.js
rhelmer: I think it makes sense to leave this on, it only runs when tags are applied (or if `UPLOAD_SENTRY_SOURCEMAPS` is true for some other reason). It's a lot more work to verify if it's working otherwise.

.github/workflows/build.yaml
```diff
@@ -18,3 +18,15 @@ jobs:
       - run: cp .env-dist .env
       # Verify that the build (incl. type-checking) succeeds
       - run: npm run build
+        env:
+          UPLOAD_SENTRY_SOURCEMAPS: false
+        # Do not upload sourcemaps for non-release (tag) builds.
+        if: ${{ !startsWith(github.ref, 'refs/tags/') }}
```
mansaj, line 24: I think it's possible that this approach only works for `on: release`

Ref: https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#release

mansaj (approved):

.github/workflows/build.yaml
rhelmer: Hm, I think we can either do that, or use `push`. I'd like to avoid duplicating all the steps if possible, I think https://docs.github.com/en/actions/using-workflows/reusing-workflows might be the way to do this?

PR #4182, opened 2024-02-09
Deprecate useGa

Vinnl:
Instead, we use the `sendGAEvent` function from (our fork of) `@next/third-parties`, so that it works with its GA setup.


rhelmer (approved):
Looks like what I was expecting - we should be able use the `DebugView` in GA4 UI to confirm that our custom events are making it in.

rhelmer:
I pushed a commit to this branch after testing to call `window.gtag()` (which is already set up) vs. pushing to `window.dataLayer` directly - it turns out to be tricky to get this just right, but as-is I'm able to see our custom events in the browser network tab, and also in the DebugView dashboard in GA4. So I think we're good to go, thanks @Vinnl for working to get all this upstreamed!

PR #4184, opened 2024-02-09
redesigned landing page free scan with and without existing account

mozrokafor:
automation script to confirm behavior for free scan in new landing page with and without an existing account

# References: 
Jira: [MNTOR-2935](https://mozilla-hub.atlassian.net/browse/MNTOR-2935)

### DIFF
--- a/src/e2e/pages/dashBoardPage.ts
+++ b/src/e2e/pages/dashBoardPage.ts
@@ -77,9 +77,7 @@ export class DashboardPage {
       name: "View all sites where your info is exposed",
     });
 
-    this.exposuresCardHeading = page.getByRole("heading", {
-      name: "Let’s keep protecting your data",
-    });
+    this.exposuresCardHeading = page.getByRole("link", { name: "Home" });
 
     this.fixedHeading = page.getByRole("heading", {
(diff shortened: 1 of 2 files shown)


src/e2e/pages/dashBoardPage.ts
```diff
@@ -77,9 +77,7 @@ export class DashboardPage {
       name: "View all sites where your info is exposed",
     });
 
-    this.exposuresCardHeading = page.getByRole("heading", {
-      name: "Let’s keep protecting your data",
-    });
+    this.exposuresCardHeading = page.getByRole("link", { name: "Home" });
```
mansaj, line 80: How is this the exposure card heading?

mozrokafor: its not, this is to confirm a successful redirect/login by verifying certain element in the dashboard

pdehaan: So is this still a `*CardHeading` or is a `*NavButton` or `*Link` now?

mozrokafor: good catch, updated the name to be more appropriate here

pdehaan (approved):

PR #4185, opened 2024-02-09
MNTOR-2929 - avoid server error and add logging, with followup

rhelmer:
# References: 
Jira: MNTOR-2929 

# Description

We can see from Sentry that `getSubscriberByEmail` sometimes returns `undefined`, which is causing server-side errors for some customers. Add temporary logging to diagnose, with a followup ticket filed to actually fix the problem (linked in FIXME comment).

# How to test

We're not sure how to reproduce this yet, this will temporarily add extra logging in production (server only).

### DIFF
--- a/src/app/functions/server/getUserBreaches.ts
+++ b/src/app/functions/server/getUserBreaches.ts
@@ -19,6 +19,7 @@ import {
 } from "../../../utils/subscriberBreaches";
 import { EmailRow } from "../../../db/tables/emailAddresses";
 import { HibpLikeDbBreach } from "../../../utils/hibp";
+import { logger } from "./logging";
 
 //TODO: deprecate with MNTOR-2021
 export type UserBreaches = {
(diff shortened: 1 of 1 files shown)


src/app/functions/server/getUserBreaches.ts
```diff
@@ -96,7 +97,12 @@ export async function getUserBreaches({
 export async function getSubscriberBreaches(
   user: Session["user"],
 ): Promise<SubscriberBreach[]> {
+  // FIXME this does not always return a subscriber https://mozilla-hub.atlassian.net/browse/MNTOR-2936
   const subscriber = await getSubscriberByEmail(user.email);
```
mansaj, line 101: Do we know if `user.email` is always present and not the problem?

Might as well log that if not

mansaj: oh sorry, i misread..  that makes sense

mansaj (approved):

PR #4190, opened 2024-02-10
MNTOR-2938 - switch from looking up by email to fxa_uid wherever possible

rhelmer:
# References: 
Jira: MNTOR-2938

# Description

We've been trying to fix problems with email mixed-case in a one-off way, but I think at the end of the day email addresses are a poor choice of identifier. The `fxa_uid` is a stable identifier that represents a Mozilla Accounts user, we should switch to this wherever possible. Email addresses are very complex and notoriously difficult to validate, since they are more about what servers accept historically and aren't really intended to be a stable identifier.

I've left `getSubscriberByEmail()` in a few places where it makes sense (e.g. looking up and removing an email address that is bouncing), but when we have a JWT from Mozilla Accounts, we should prefer the UID over the email.

# How to test

This should entirely eliminate the class of bugs that happen when registering with a mixed-case email address. Otherwise, this should be exactly the same. We should ensure that indexes are in place so user lookups are still fast.

### DIFF
--- a/src/app/functions/server/onerep.ts
+++ b/src/app/functions/server/onerep.ts
@@ -361,7 +361,7 @@ export async function isEligibleForFreeScan(
   }
 
   if (!user?.subscriber?.id) {
-    throw new Error("No session");
+    throw new Error("No fxa_uid found in session");
   }
 
   const enabledFlags = await getEnabledFeatureFlags({ email: user.email });
--- a/src/app/functions/server/onerep.ts
+++ b/src/app/functions/server/onerep.ts
@@ -361,7 +361,7 @@ export async function isEligibleForFreeScan(
   }
 
   if (!user?.subscriber?.id) {
-    throw new Error("No fxa_uid found in session");
+    throw new Error("No session with a known subscriber found");
   }
 
   const enabledFlags = await getEnabledFeatureFlags({ email: user.email });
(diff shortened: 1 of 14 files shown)


rhelmer:
Looks like we already have an index on this column \o/ https://github.com/mozilla/blurts-server/blob/1008bfee2ebcf64f9881480892d1e7fa2eaf04bd/src/db/migrations/20191118100718_add-fxa-uid-index.js

src/app/functions/server/onerep.ts
```diff
@@ -361,7 +361,7 @@ export async function isEligibleForFreeScan(
   }
 
   if (!user?.subscriber?.id) {
-    throw new Error("No session");
+    throw new Error("No fxa_uid found in session");
```
Vinnl, line 364: This error message looks wrong now, as this is checking for the ID of the user in our own DB table, not the ID of the FxA user.

Vinnl (approved):
The changes seem straightforward enough and I see no issues locally, though I'm still a bit wary since its impact is all over the codebase - I'm glad you didn't deploy this late on Friday with half the team out :sweat_smile:

src/app/functions/server/onerep.ts
Vinnl: Changed myself since you asked me to merge this: 4eebbaedd.

PR #4201, opened 2024-02-12
`getSubscriberByFxaUid` and `getSubscriberByEmail` can return undefined

Vinnl:
# References: 
Jira: MNTOR-2938
Figma: N/A

# Description

This should only change type-level code, except where it adds exceptions to deal with no subscriber being found. This should help us detect when we run into an issue that causes us not to be able to load subscriber data.

# How to test

Basically, no behaviour should have changed.

- [ ] I've added a unit test to test for potential regressions of this bug. -> Should be covered by existing unit tests, insofar they exist


src/db/tables/subscribers.js
```diff
@@ -65,6 +65,7 @@ async function getSubscriberById (id) {
 
 /**
  * @param {string} uid
+ * @returns {Promise<undefined | import("knex/types/tables").SubscriberRow & { email_addresses: Array<{ id: import("knex/types/tables").EmailAddressRow["id"]; email: import("knex/types/tables").EmailAddressRow["email"]; }> }>}
```
Vinnl, line 68: I started out by adding the `@returns` annotation here (in one commit, and on line 82 in another), and then following up the TypeScript warnings. The other changes should only be type-level changes (i.e. they don't affect how our code runs), except where it explicitly handles the case where this returns `undefined`, logging an error. Hopefully, that will make it easier to debug issues down the road.

(Also: I wish this had been covered by TypeScript in from the start, but alas...)

src/app/deprecated/(authenticated)/user/breaches/breaches.d.ts
```diff
@@ -82,6 +82,9 @@ export interface Breach {
   Title: string;
 }
 
+/**
+ * @deprecated Use {@see SubscriberRow} instead
```
rhelmer, line 86: I'm a little on the fence about this @Vinnl - I landed this PR with the comment, but I think there's at least one use case where it's required and that's when AWS can't deliver a message, we need to be able to look up by email.

Vinnl: Yeah maybe the comment is a bit too stern, though I think marking it as deprecated is still fine - it won't prevent you from actually using it, but some editors will at least highlight it as something to give a second though if you can, and maybe add a comment on why it's necessary in that specific case.

(I am assuming here you meant to comment on https://github.com/mozilla/blurts-server/blob/6ed8f194c668c3bf073418bbe8bfb01d0f4b6d96/src/db/tables/subscribers.js#L84.)

PR #4207, opened 2024-02-13
MNTOR-2957 - avoid loading client-side analytics on server

rhelmer:
# References: 
Jira: MNTOR-2957

# Description

Next is attempting to load client-side analytics scripts on the server-side. This is generating spurious logs and HTTP 500s that don't seem to be impacting the product experience, but are most likely causing SSR not to work and impacting performance.

This is a workaround while we investigate.

### DIFF
--- a/src/app/hooks/useGlean.ts
+++ b/src/app/hooks/useGlean.ts
@@ -15,6 +15,11 @@ export const useGlean = () => {
 
   // Initialize Glean only on the first render of our custom hook.
   useEffect(() => {
+    if (typeof window === "undefined") {
+      console.warn("Glean should only be used on the client");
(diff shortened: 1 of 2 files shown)


src/app/hooks/useGlean.ts
```diff
@@ -24,6 +29,11 @@ export const useGlean = () => {
       throw new ErrorEvent("No PUBLIC_APP_ENV provided for Glean");
     }
 
+    if (typeof window === "undefined") {
```
mansaj, line 32: Wouldn't L18-22 cover this? Is it not the same function?

mansaj (approved):

PR #4227, opened 2024-02-19
Include stripe.js for fraud resistance (MNTOR-1769)

flozia:
# References: 
Jira: [MNTOR-1769](https://mozilla-hub.atlassian.net/browse/MNTOR-1769)

# Description

Include `stripe.js` for fraud resistance as requested in [MNTOR-1769](https://mozilla-hub.atlassian.net/browse/MNTOR-1769).

# How to test

Make sure `<script src="https://js.stripe.com/v3"></script>` is present and `stripe.js` is being loaded without any CSP errors.

### DIFF
--- a/src/app/components/client/StripeScript.tsx
+++ b/src/app/components/client/StripeScript.tsx
@@ -1,0 +1,13 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+"use client";
+
+// Load stripe.js for fraud resistance.
+// https://stripe.com/guides/radar-rules-101#importance-of-using-stripejs
+import "@stripe/stripe-js";
+
+export default function StripeScript() {
+  return <></>;
+}
--- a/package-lock.json
+++ b/package-lock.json
@@ -23,6 +23,7 @@
         "@sentry/nextjs": "^7.100.1",
         "@sentry/node": "^7.58.1",
         "@sentry/tracing": "^7.100.1",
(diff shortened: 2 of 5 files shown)


src/app/components/client/StripeScript.tsx
```diff
@@ -0,0 +1,13 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+"use client";
```
Vinnl, line 5: Could you explain why this needs to be a client component, and why it can't just be imported directly in the layout?

flozia: I opted for including `stripe.js` [as a module](https://docs.stripe.com/js/including), this is also how [MDN does it as well](https://github.com/mdn/yari/blob/2f4a09d0021817207f1dc3a1dea368267ab49d28/client/src/plus/offer-overview/offer-overview-subscribe/stripe.tsx).

An alternative would be to directly add `<script src="https://js.stripe.com/v3" async></script>`.  When it’s [imported as a side effect](https://github.com/stripe/stripe-js?tab=readme-ov-file#import-as-a-side-effect), like it is in this PR, the `<script>` is being [injected client side](https://github.com/stripe/stripe-js/blob/master/src/shared.ts#L38) and expects the DOM to be present — thus the `"use client"`.

Vinnl: But client components are still pre-rendered on the server, so there might still not be a DOM present. Possibly you need to do a dynamic `import("@stripe/stripe-js")` inside a `useEffect` to make it only run on the client side?

Vinnl (approved):
Thank you!

PR #4230, opened 2024-02-21
add logging when free scan is created

rhelmer:
# References: 
Jira: MNTOR-2686

# Description

It's pretty difficult to generate reports right now about how long free scans are taking. This PR just adds a bit of logging when scans are created that we can get in ASAP, I have a branch that refactors `onerep.ts` but it's a much larger change that we should spend some time testing on stage.

### DIFF
--- a/src/app/functions/server/onerep.ts
+++ b/src/app/functions/server/onerep.ts
@@ -161,6 +161,9 @@ export async function createProfile(
     updated_at: ISO8601DateString;
     url: string;
   } = await response.json();
+
+  logger.info("onerep_profile_created", { onerepProfileId: savedProfile.id });
+
   return savedProfile.id;
 }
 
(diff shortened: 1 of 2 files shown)


src/app/functions/server/onerep.ts
```diff
@@ -161,6 +161,9 @@ export async function createProfile(
     updated_at: ISO8601DateString;
     url: string;
   } = await response.json();
+
+  logger.info("onerep_profile_created", { onerepProfileId: savedProfile.id });
```
flozia, line 165: Is it fine for us to log the OneRep profile id in clear text or is this PII?

rhelmer: We've been only logging IDs for reporting purposes (and not the actual data it refers to) which is personally identifiable, but not without credentials. I don't think we *need* this one, but in order to track some things like how long scans take we do need the scan ID. I'll remove it for now and follow up to make sure this is OK to do.

mansaj (approved):

PR #4233, opened 2024-02-21
Also check for updates of Actions

Vinnl:
See https://mozilla.slack.com/archives/C04BLQVN151/p1708515489471449?thread_ts=1708510608.086919&cid=C04BLQVN151:

image


rhelmer (approved):
Oh that's a nice find, thanks!

PR #4234, opened 2024-02-21
Unblock Node upgrade by adding arbitrary c8 ignores

Vinnl:
With the upgrade to Node 20.10, suddenly it started reporting code as uncovered that was previously not, and most of which I'm fairly sure _is_ actually covered by unit tests. I think it's probably caused by [this issue](https://github.com/istanbuljs/v8-to-istanbul/issues/236), but it's fairly hard to debug. By running `npm test -- --runInBand --no-cache` (thanks @flozia) I was able to get consistent reports of uncovered lines, and thus was able to at least get back to a point where it's meeting the thresholds again - but hopefully this won't make it harder to find which new code we still need to write tests for in the future.

I did leave some more diagnostic information in the upstream issue, and it refers to a number of other people running into what seems to be the same issue, so hopefully they'll find a fix there before we run into it again.

### DIFF
--- a/src/app/(proper_react)/(redesign)/(authenticated)/user/(dashboard)/dashboard/filterExposures.ts
+++ b/src/app/(proper_react)/(redesign)/(authenticated)/user/(dashboard)/dashboard/filterExposures.ts
@@ -13,6 +13,9 @@ export function filterExposures(
   filters: FilterState,
 ): Exposure[] {
   return exposures.filter((exposure) => {
+    /* c8 ignore start */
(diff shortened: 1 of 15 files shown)


flozia (commented):
As already discussed: Ideally, we would need to understand why we suddenly see more uncovered lines after the Node upgrade. Best case: It’s fixed by the next version. Thanks for unblocking the Node upgrade!

flozia (approved):

PR #4240, opened 2024-02-22
MNTOR-2983: Subscription changes without first user scan throws 500

mansaj:
# References: 
Jira: MNTOR-2983

# Description
Subscription changes without first user scan currently throws 500. We want to change that to 200 to avoid unnecessary fxa retries

# Screenshot (if applicable)

Not applicable.

# How to test


src/app/api/v1/fxa-rp-events/route.ts
```diff
@@ -310,15 +310,19 @@ export async function POST(request: NextRequest) {
               });
 
               captureException(
-                new Error(`No OneRep profile Id found, subscriber: ${
+                new Error(`User subscribed but no OneRep profile Id found, user: ${
```
rhelmer, line 313: I know you didn't introduce it, but it seems odd to construct an `Error` just to pass to `captureException`... I wonder if `captureMessage` is a better option here.

src/app/api/v1/fxa-rp-events/route.ts
```diff
@@ -310,15 +310,19 @@ export async function POST(request: NextRequest) {
...
             Event: ${event}\n
             updateFromEvent: ${JSON.stringify(updatedSubscriptionFromEvent)}`),
               );
               return NextResponse.json(
-                { success: false, message: "failed_activating_subscription" },
-                { status: 500 },
+                {
+                  success: false,
+                  message:
+                    "failed_activating_subscription, user hasn't ran scan yet",
```
rhelmer, line 323: It feels wrong to return 200 OK status with `{ success: false }`.

I think the `message` should either be all underscore-style or not, mixing it up is probably harder to parse by the API caller (not that they particularly care in this case, but it might be helpful for debugging).

rhelmer: I think this should return `success: true` and the message should be worded more like a warning... this is frustrating. I wonder if we should just leave activation up to the UI entirely? It will never work until the scan has been performed.

rhelmer (changes requested):

src/app/api/v1/fxa-rp-events/route.ts
mansaj: Good call

mansaj: Fixed a bunch of them

src/app/api/v1/fxa-rp-events/route.ts
mansaj: It's a weird one, but I made it `success:true` and the message more consistent with snake case

rhelmer (approved):

PR #4241, opened 2024-02-23
MNTOR-2985/fix deploy, remove pip from Dockerfile

rhelmer:
# References: 
Jira: MNTOR-2985

# Description

The latest version of Alpine enforces `externally-managed-environment` so trying to use `pip` to modify the distro-installed version of Python will fail unless overridden. Maybe this was needed in the past, but nowadays I think we can just install Python3 and let the Glean build do its thing in a virtualenv.

I noticed a few unused variables from our Sentry experiments, I removed these as well.


Dockerfile
```diff
@@ -7,8 +7,6 @@ RUN rm -rf /tmp/*
 # Install Python
 ENV PYTHONUNBUFFERED=1
 RUN apk add --update --no-cache python3 py3-pip && ln -sf python3 /usr/bin/python
-RUN python3 -m ensurepip
-RUN pip3 install --no-cache --upgrade pip setuptools
```
mansaj, line 11: Tried building it, but I think it gets stuck at `line 30`: `RUN GLEAN_PYTHON=python GLEAN_PIP=pip npm run build`

mansaj:
```
Dockerfile:30
--------------------
  28 |     ENV S3_BUCKET=$S3_BUCKET
  29 |     
  30 | >>> RUN GLEAN_PYTHON=python GLEAN_PIP=pip npm run build
  31 |     
  32 |     ARG SENTRY_RELEASE
--------------------
ERROR: failed to solve: process "/bin/sh -c GLEAN_PYTHON=python GLEAN_PIP=pip npm run build" did not complete successfully: exit code: 1
```

rhelmer: @mansaj Hm, this seemed to work in CircleCI (using the SSH debugging feature), is this locally that it's happening for you?

mansaj (approved):

Dockerfile
rhelmer: For posterity: we're both having different local docker build issues, but it works in CircleCI so landing as-is to unblock shipping.

PR #4257, opened 2024-02-23
Merge hotfix

Vinnl:
# References: 
Jira: MNTOR-2986


src/app/(proper_react)/(redesign)/(authenticated)/user/(dashboard)/settings/actions.ts
```diff
@@ -156,8 +156,8 @@ export async function onRemoveEmail(email: EmailRow) {
   }
 
   try {
-    await removeOneSecondaryEmail(email.id);
-    await deleteResolutionsWithEmail(email.subscriber_id, email.email);
+    await removeOneSecondaryEmail(email.id, subscriber.id);
+    await deleteResolutionsWithEmail(subscriber.id, email.email);
```
pdehaan, line 160: I guess one risk here might be that the `subscriber` object is nullish (per L146, and the `subscriber?.id` chaining on L146+149). But I guess if that happens, we're still in a `try..catch` block and it'll return the "user-delete-unknown-error" event below on L163-167.

Q: Do we need an explicit check for a truthy `subscriber` object somewhere around L146.5 (similar to what we do below in blurts-server/src/app/api/v1/user/remove-email/route.ts:32-34?
https://github.com/mozilla/blurts-server/blob/1298645aaa386dde9068bd45ca5a7821203a4ac1/src/app/api/v1/user/remove-email/route.ts#L32-L34

Vinnl: Heh, I was wondering why TypeScript wouldn't warn us of that, but I think it's actually already covered. If `subscriber` is `null`, then

```
email.subscriber_id !== subscriber?.id
```

will be `true` (because `email.subscriber_id` will never be `null`), and then we'll hit the return on line 151.

rhelmer (approved):

PR #4287, opened 2024-03-01
Delete Monitor (but not FxA) account

Vinnl:
# References: 
Jira: MNTOR-1371
Figma: https://www.figma.com/file/CaEKIhvSJqf6KNIMzSkt40/Concepts-for-Monitor-MVP-Redesign?type=design&node-id=8356-55605&mode=dev

# Description

The way this works:

- On the settings page, there's a <DeleteAccountButton> (inside a dialog): a client component that, when pressed, calls the `onDeleteAccount` server action.
- The server action deletes the account (using the same code as on the `fxa-rp-events` route, now extracted into a separate function), sets a cookie `justDeletedAccount` that expires in five minutes, then returns.
- After the action returns, the client signs the user out and redirects them to the homepage. I'd have liked to have done this in the server action, but Next-Auth doesn't allow signing the user out from the server side yet.
- Then on the homepage, we check for the `justDeletedAccount` cookie, delete it, then display a confirmation that deleting the account was successful.

(~~I just noticed I still have to update some telemetry, but I'm submitting it already since I got pretty far in the submission process and telemetry shouldn't materially change the review. I'll add them as fixup commits later so existing reviews don't get borked.~~ Added: fa4b61b40)

# Screenshot (if applicable)

image

image

image

image

# How to test

Enable the `MonitorAccountDeletion`, then visit the settings page with both a free and a Plus user.


src/app/(proper_react)/(redesign)/(authenticated)/user/(dashboard)/settings/EmailAddressAdder.tsx
```diff
@@ -134,7 +134,6 @@ const EmailAddressAddForm = () => {
             setHasPressedButton(true);
           }}
           isLoading={hasPressedButton}
-          aria-live="polite"
```
Vinnl, line 137: I copied this (and the changes to `<Button>`) from https://github.com/mozilla/blurts-server/pull/4203 because it was also relevant to this PR.

src/app/(proper_react)/(redesign)/(public)/LandingView.tsx
```diff
@@ -43,226 +44,229 @@ export type Props = {
...
-              isHero
-              eligibleForPremium={props.eligibleForPremium}
-              signUpCallbackUrl={`${process.env.SERVER_URL}/user/dashboard/`}
-              eventId={{
-                cta: "clicked_get_scan_header",
-                field: "entered_email_address_header",
-              }}
+    <>
+      <AccountDeletionNotification />
+      <main className={styles.wrapper}>
```
Vinnl, line 49: This looks like it's a huge change, but it only increases the indentation level to wrap it in `<></>`, and adds `<AccountDeletionNotification />` to the front.

flodolo (approved):

flozia (commented):
This is working as expected and nothing jumps out to me — great job! I noticed that the respective Mozilla account still lists Monitor under “connected services”. Is this something we need to address or is there even a way for us to notify FxA and remove it from the list?

flozia (approved):

locales/en/settings.ftl
```diff
@@ -59,6 +59,26 @@ settings-deactivate-account-title = Deactivate account
...
+settings-delete-monitor-free-account-dialog-title = Your { -brand-monitor } account will be permanently deleted
+settings-delete-monitor-free-account-dialog-lead = All of your { -brand-monitor } account information will be deleted and we’ll no longer monitor for new data breaches. This will not delete your { -brand-mozilla } account.
+settings-delete-monitor-free-account-dialog-cta-label = Delete account
+settings-delete-monitor-free-account-dialog-cancel-button-label = Never mind, take me back
+settings-delete-monitor-plus-account-title = Delete { -brand-monitor } account
+settings-delete-monitor-plus-account-description = This will permanently delete your { -brand-monitor } account and immediately end your paid { -brand-monitor-plus } subscription.
+settings-delete-monitor-plus-account-cta-label = Delete account
+settings-delete-monitor-plus-account-dialog-title = Your { -brand-monitor } account will be permanently deleted
+settings-delete-monitor-plus-account-dialog-lead-p1 = All of your { -brand-monitor } account information will be deleted and we’ll no longer monitor for new data breaches or data broker exposures. This will not delete your { -brand-mozilla } account.
```
flodolo, line 75: Note to self, I should have caught this 😭 

`Mozilla account` is `{ -brand-mozilla-account }`, not a combination of Mozilla brand and account. Since I've already exposed these strings for translation, we should keep this in at this point.

Vinnl: Oh yes, good point :/ Maybe a good addition to the linter too, to catch instances of `{ -brand-mozilla } account`? Theoretically that could run into false positives, but probably not in practice, and false negatives are the same as the status quo.

PR #4288, opened 2024-03-01
MNTOR-2686/refactor onerep module

rhelmer:
# References: 
Jira: MNTOR-2686

# Description

This centralizes error checking and logging in the private `onerepFetch` function.

# How to test

Nothing should change from the calling code perspective, this is refactoring internal to the `onerep` module to be more consistent with logging and error checking.

- [x] Jira ticket has been updated (if needed) to match changes made during the development process..


src/app/functions/server/onerep.ts
```diff
@@ -117,12 +117,42 @@ async function onerepFetch(
...
     throw new Error("ONEREP_API_KEY env var not set");
   }
-  const url = new URL(path, onerepApiBase);
-  const headers = new Headers(options.headers);
-  headers.set("Authorization", `Bearer ${onerepApiKey}`);
-  headers.set("Accept", "application/json");
-  headers.set("Content-Type", "application/json");
-  return fetch(url, { ...options, headers });
+  try {
+    const url = `${onerepApiBase}${path}`;
```
mansaj, line 121: Is there a reason we'd want to not use `new URL`?

mansaj (approved):
Thanks for getting to this, great refactor

src/app/functions/server/onerep.ts
```diff
@@ -170,72 +189,34 @@ export async function createProfile(
...
-      `Failed to activate OneRep profile: [${response.status}] [${response.statusText}]`,
-    );
-    throw new Error(
-      `Failed to activate OneRep profile: [${response.status}] [${response.statusText}]`,
-    );
-  }
+  await onerepFetch(`/profiles/${profileId}/activate`, {
+    method: "PUT",
+  });
+  logger.info("activated_onerep_profile", { onerepProfileId: profileId });
```
pdehaan, line 205: Not sure if it matters or if we have an established style guide, but should these labels be suffixed or prefixed:
```
onerep_profile_xyz
```

or

```
xyz_onerep_profile
```

This isn't super-conclusive, but also, it's an admittedly kinda trash <kbd>grep</kbd>:
```sh
git grep -Pino "\blogger\.\w+\(\"[\w_-]+\"" src/ | grep "onerep_profile"

src/app/api/v1/admin/users/[primarySha1]/route.ts:187:logger.info("delete_onerep_profile"
src/app/api/v1/fxa-rp-events/route.ts:202:logger.info("deactivated_onerep_profile"
src/app/api/v1/fxa-rp-events/route.ts:294:logger.info("get_onerep_profile"
src/app/api/v1/fxa-rp-events/route.ts:312:logger.error("onerep_profile_not_found"
src/app/api/v1/fxa-rp-events/route.ts:359:logger.info("activated_onerep_profile"
src/app/api/v1/fxa-rp-events/route.ts:375:logger.error("onerep_profile_not_found"
src/app/api/v1/fxa-rp-events/route.ts:406:logger.info("deactivated_onerep_profile"
src/app/functions/server/onerep.ts:184:logger.info("onerep_profile_created"
src/app/functions/server/onerep.ts:205:logger.info("activated_onerep_profile"
src/app/functions/server/onerep.ts:212:logger.info("deactivated_onerep_profile"
src/app/functions/server/onerep.ts:219:logger.info("optout_onerep_profile"
```

src/app/functions/server/onerep.ts
```diff
@@ -376,7 +342,7 @@ export async function isEligibleForFreeScan(
   const scanResult = await getLatestOnerepScanResults(profileId);
 
   if (scanResult.scan) {
-    logger.warn("User has already used free scan");
+    logger.info("User has already used free scan");
```
pdehaan, line 345: Q: Do we have a convention on when we log slugs vs strings?

Re: 
```js
logger.info("deactivated_onerep_profile", {...});
logger.info("User has already used free scan");
```

Not sure if it really matters.
PROS: Consistency
CONS: ? Might break existing metrics for a while?

src/app/functions/server/onerep.ts
```diff
@@ -117,12 +117,42 @@ async function onerepFetch(
...
+        status: response.status,
+        statusText: response.statusText,
+      };
+      logger.error("onerep_response_not_ok", status);
+      throw new Error(`OneRep fetch failed: ${JSON.stringify(status)}`);
+    }
+
+    logger.info("onerep_fetch", { path, options });
+
+    return (await response.json()) as Promise<unknown>;
```
Vinnl, line 140: I think (?) that TypeScript infers the `Promise` from it being an async function, so this annotation might result in a `Promise<Promise<unknown>>`?

```suggestion
    return (await response.json()) as unknown;
```

src/app/functions/server/onerep.ts
```diff
@@ -117,12 +117,42 @@ async function onerepFetch(
...
+  } catch (e) {
```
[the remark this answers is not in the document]
Vinnl: Which would be a `Promise<undefined>`, which technically is also a `Promise<unknown>`. But yeah, this would swallow errors and make calling code unable to handle it. We could add a `throw e` after the `else {}`?

src/app/functions/server/onerep.ts
```diff
@@ -117,12 +117,42 @@ async function onerepFetch(
...
+    headers.set("Authorization", `Bearer ${onerepApiKey}`);
+    headers.set("Accept", "application/json");
+    headers.set("Content-Type", "application/json");
+
+    const response = await fetch(url, { ...options, headers });
+
+    if (!response.ok) {
+      const status = {
+        status: response.status,
+        statusText: response.statusText,
```
Vinnl, line 132: Maybe it would be a good idea to include the URL here as well? Otherwise it's a bit hard to determine _which_ OneRep fetch failed.

```suggestion
        statusText: response.statusText,
        url: url,
```

src/app/functions/server/onerep.ts
```diff
@@ -429,7 +388,8 @@ export async function getAllDataBrokers() {
     const response = await onerepFetch(
       "/data-brokers?per_page=100&page=" + page.toString(),
     );
-    const data: OneRepResponse<
+    const data = await response;
```
Vinnl, line 391: This `await` seems superfluous as well.

src/app/functions/server/onerep.ts
```diff
@@ -479,25 +438,16 @@ export async function getProfilesStats(
...
-  const profileStats: ProfileStats = await response.json();
+  )) as Promise<ProfileStats>;
 
   // cache results in map, with a flush hack to keep the size low
   if (profileStatsCache.size > 5) profileStatsCache.clear();
-  profileStatsCache.set(queryParamsString, profileStats);
+  profileStatsCache.set(queryParamsString, await profileStats);
   return profileStats;
 }
+export { getLatestOnerepScanResults };
```
Vinnl, line 453: Why are we re-exporting this here? Doesn't it make sense to keep a distinction between modules that fetch data from the OneRep API, and those that fetch data from our own database?

src/app/functions/server/onerep.ts
```diff
@@ -479,25 +438,16 @@ export async function getProfilesStats(
...
-    logger.error(
-      `Failed to fetch OneRep profile: [${response.status}] [${response.statusText}]`,
-    );
-    throw new Error(
-      `Failed to fetch OneRep profile: [${response.status}] [${response.statusText}]`,
-    );
-  }
-
-  const profileStats: ProfileStats = await response.json();
+  )) as Promise<ProfileStats>;
```
Vinnl, line 446: I think this type assertion is incorrect (because the return value is `await`'ed), and so is the `await` on line 450.

```suggestion
  )) as ProfileStats;
```

src/app/functions/server/onerep.ts
Vinnl, line 140: Are we actually certain that all OneRep's APIs will always return JSON? Even e.g. in case of error responses?

src/app/functions/server/onerep.ts
```diff
@@ -2,7 +2,7 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
-import type { Session } from "next-auth";
+import { type Session } from "next-auth";
```
Vinnl, line 5: Asking for my own education: what does this do differently?

Vinnl (approved):
I have two main potential concerns; if those are unnecessary, then the overall approach looks fine:

- Aren't we swallowing the errors in `onerepFetch` and, if so, isn't that problematic?
- Are we actually sure the OneRep API always returns JSON?

rhelmer:
> I have two main potential concerns; if those are unnecessary, then the overall approach looks fine:
> 
>     * Aren't we swallowing the errors in `onerepFetch` and, if so, isn't that problematic?

Sorry I should make this clearer - one of the motivations here is that I noticed that the dashboard throws errors if OneRep is down or returning bad results, instead of falling back to our local cache. I'll try to make this clearer, in a comment if nothing else.

>     * Are we actually sure the OneRep API always returns JSON?

The intent is to catch and log this if it doesn't. Some APIs return nothing, AFAIK everything else returns JSON.

Vinnl:
> the dashboard throws errors if OneRep is down or returning bad results, instead of falling back to our local cache.

Shouldn't we solve that by catching the errors in the dashboard and then fetching them from the local cache? As I understand it, this change will make the OneRep functions return `undefined` instead of throwing an error, which I don't imagine (/hope?) result in us serving results from a local cache?

(By "local cache" you mean scan results stored in the database, right?)

src/app/functions/server/onerep.ts
rhelmer: @mansaj and I have discussed the need for it - I think we should go underscore style in general since it makes it better for metrics (it can be treated as an event name with parameters, vs. just a string)

src/app/functions/server/onerep.ts
rhelmer: Pretty sure this was the linter not me doing it purposely :) There's no functional difference AFAIK just style

rhelmer:
> > the dashboard throws errors if OneRep is down or returning bad results, instead of falling back to our local cache.
> 
> Shouldn't we solve that by catching the errors in the dashboard and then fetching them from the local cache? As I understand it, this change will make the OneRep functions return `undefined` instead of throwing an error, which I don't imagine (/hope?) result in us serving results from a local cache?
> 
> (By "local cache" you mean scan results stored in the database, right?)

I was trying to keep it minimal, but I think what you're suggesting makes sense. I am going to put this in draft and go a little further.

src/app/functions/server/onerep.ts
pdehaan: Mostly for my own future reference&hellip; but also, sorry for the _far_ too deep chain of <kbd>grep</kbd>  commands. It sickens me and brings me great shame. 😞 

```
git grep -Pino "\blogger\.\w+\([\"'][\w\s_-]+[\"']" src/ | grep -Eo "['\"].*['\"]" | grep -v "_" | grep -Ev "^[a-z\"]+$" | sort | uniq -c

   2 "No user ID for Nimbus telemetry"
   1 "No user ID for telemetry"
   1 "Successfully queued breach notification"
   1 "User has already used free scan"
   2 "addAttributionForSubscriber"
   1 "addFeatureFlag"
   1 "deleteAttributionById"
   1 "deleteAttributionsForSubscriber"
   1 "deleteFeatureFlagByName"
   1 "disableFeatureFlagByName"
   1 "enableFeatureFlagByName"
   1 "fxa-confirmed-profile-data"
   3 "fxaRpEvents"
   1 "getAttributionsForSubscriber"
   1 "getFeatureFlagByName"
   1 "getJwtPubKey"
   1 "getLatestAttributionForSubscriberWithType"
   1 "updateAllowList"
   1 "updateAttribution"
   1 "updateDependencies"
   1 "updateOwner"
   1 "updateWaitList"
```

rhelmer:
Putting on hold for the moment, may propose some more fundamental changes.

PR #4303, opened 2024-03-05
MNTOR-2945: add del subscription api and unsub before deleting an user

mansaj:
# References: 
Jira: MNTOR-2945

# Description
Implement the unsubscribe feature correctly as a part of deletion, so the user
1) won't be billed again when signing up again
2) won't have any pre-existing records of Plus account

# Screenshot (if applicable)

Not applicable.

# How to test
In heroku, use a new account to subscribe to Plus. Delete that account. Watch for `unsubscribe` email to come through from FxA


src/app/functions/server/deleteAccount.ts
```diff
@@ -37,6 +39,21 @@ export async function deleteAccount(
...
+    // try to unsubscribe from subplat
+    try {
+      const subscriptionId = "monitor";
+      await deleteSubscription(subscriptionId);
+      logger.info("unsubscribe_from_subplat", {
+        subscriber_id: subscriber.id,
+        success: true,
+      });
+    } catch (ex) {
+      logger.error("unsubscribe_from_subplat", {
```
pdehaan, line 52: Should L47 and L52 use the same `unsubscribe_from_subplat` key/string? Or should this be something like `unsubscribe_from_subplat_failed`?

src/utils/fxa.js
```diff
@@ -155,6 +155,29 @@ async function sendMetricsFlowPing (path) {
 }
 /* c8 ignore stop */
 
+/**
+ * @param {string} subscriptionId 
+ */
+// TODO: Add unit test when changing this code:
```
Vinnl, line 161: I don't think it's realistic that we'll ever write unit tests for it - they wouldn't be very useful either, I don't think :)

```suggestion
// Not covered by tests; mostly side-effects. See test-coverage.md#mock-heavy
```

src/utils/fxa.js
```diff
@@ -155,6 +155,29 @@ async function sendMetricsFlowPing (path) {
...
+/* c8 ignore start */
+async function deleteSubscription (subscriptionId) {
+  const url = `${AppConstants.OAUTH_ACCOUNT_URI}/oauth/subscriptions/active/${subscriptionId}`
+  try {
+    const response = await fetch(url, {
```
Vinnl, line 166: Woah, does this really work? Shouldn't we need to pass in a token or something?

Vinnl: I get the following when testing it:

```
delete_fxa_subscription {
  stack: 'Internal Server Error: bad response: 401\n'
    '    at deleteSubscription (webpack-internal:///(action-browser)/./src/utils/fxa.js:182:33)\n' +
...
```

src/app/functions/server/deleteAccount.ts
```diff
@@ -37,6 +39,21 @@ export async function deleteAccount(
     logger.info("deactivated_onerep_profile", {
       subscriber_id: subscriber.id,
     });
+
+    // try to unsubscribe from subplat
```
Vinnl, line 43: This function is also called from the `fxa-rp-events` route, in which case we probably don't want to run this. So maybe this needs to be called in the `onDeleteAccount` action separately, or `deleteAccount` could take a parameter to skip it?

Vinnl (commented):
Unfortunately it appears like I'm getting an error when trying it locally - shouldn't the request to FxA be authenticated somehow?

And I think we also don't want to try unsubscribing if we just got a notification from FxA that the user has been unsubscribed?

(And a small question: won't this a bit too noisy for our logs?)

src/utils/fxa.js
mansaj: Redid the logic after testing in heroku. I misunderstood the API. The subscription ID needs to be fetched first.

src/app/functions/server/deleteAccount.ts
mansaj: I think with the updated logic, there's an `if` statement checking for whether subscription is present from fxa side, which is a more fool-proof way of checking if we'd want to skip

src/app/functions/server/deleteAccount.ts
mansaj: I think the `logger.error` would show up differently enough in our logs that this won't be misinterpretted

src/utils/fxa.js
Vinnl: I think you changed this on a different line (though I guess it applies there too) :)

src/app/functions/server/deleteAccount.ts
Vinnl: That's perfect, thanks.

Vinnl (commented):
Thanks, the code makes sense to me now. But when I try it out locally, I'm getting

```
deleteSubscriber error: delete from "subscribers" where "fxa_uid" = $1 returning "id" - update or delete on table "subscribers" violates foreign key constraint "onerep_scans_onerep_profile_id_foreign" on table "onerep_scans"
```

I wonder if we're missing a cascade somewhere to also delete scans and scan results when we delete a subscriber?

rhelmer (changes requested):
I don't understand how we're sure the subscription ID is for the `monitor` subscription. Have you tested this with an account that is subscribed to multiple services?

We should really convert this to typescript and use the newer formatting rules, is that already in our tech debt epic? That part isn't a blocker for this PR, but not being able to use structured logging here is annoying.

src/utils/fxa.js
```diff
@@ -155,12 +155,78 @@ async function sendMetricsFlowPing (path) {
...
+/**
+ * @param {string} bearerToken
+ * @returns {Promise<boolean>}
+ */
+// Not covered by tests; mostly side-effects. See test-coverage.md#mock-heavy
+/* c8 ignore start */
+async function deleteSubscription(bearerToken) {
+  try {
+    const subs = await getSubscriptions(bearerToken)
+    const subscriptionId = subs?.[0]?.subscriptionId
```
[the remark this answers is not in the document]
mansaj: Oh shoot, good call!! Hmm let me rework this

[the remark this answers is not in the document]
mansaj: @rhelmer fixed, i'll test it on heroku

src/utils/fxa.js
```diff
@@ -155,12 +155,78 @@ async function sendMetricsFlowPing (path) {
...
+        method: "DELETE",
+        headers: {
+          Accept: 'application/json',
+          Authorization: `Bearer ${bearerToken}`
+        }
+      })
+      if (!response.ok) {
+        throw new InternalServerError(`bad response: ${response.status}`)
+      } else {
+        console.info(`delete_fxa_subscription: success - ${JSON.stringify(await response.json())}`)
```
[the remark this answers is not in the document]
mansaj: i think we tried this before and it wouldn't work. Everything has to be `ts`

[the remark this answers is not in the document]
rhelmer: Yeah no worries, we'll convert it and it'll be fine.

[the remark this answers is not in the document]
rhelmer: It should be possible to set our project up such that we can compile ts->js and then have imports from js work, but it's probably not worth the effort. I'd rather just go all ts

rhelmer (approved):

mansaj:
@rhelmer Thanks for catching that, verified the fix in heroku just now. The unsub works 

> I don't understand how we're sure the subscription ID is for the `monitor` subscription. Have you tested this with an account that is subscribed to multiple services?
> 
> We should really convert this to typescript and use the newer formatting rules, is that already in our tech debt epic? That part isn't a blocker for this PR, but not being able to use structured logging here is annoying.

PR #4304, opened 2024-03-06
MNTOR-2765 - Find exposures button should not announce collapsed

codemist:
# References: 
Jira: MNTOR-2765 and potentially MPP-2766


Vinnl (commented):
Thanks for looking at this. I think the first think to look into is whether this is actually needed - I left a comment in the ticket that there might be a misunderstanding here.

But if it is needed, I'm running into a couple of issues:

- If I get the "City and state" field into an invalid value (e.g. I tab into it, then out of it again without selecting anything), I can no longer get it into a valid state using my keyboard (i.e. focusing it again, typing "Tul", then down and <kbd>Enter</kbd> to select `Tulsa, OK, USA`).
- I can no longer close the dialog using the <kbd>Esc</kbd> key.
- I can no longer close the dialog by clicking outside it.
- There is no announcement when the dialog opened, but that is already the case on `main`.

Possibly the second and third item can be fixed by adding an `onDismiss` to the `<Dialog>` inside `<ConfirmInfoDialog>`? I'm not sure what could be causing the former - I wonder if it's something specific to my setup, maybe you could check whether you're seeing it?

Alternatively, maybe just adding `aria-expanded={undefined}` to the `<Button>` would be sufficient too.

(And sorry for taking so long to review!)

src/app/(proper_react)/(redesign)/(authenticated)/user/welcome/EnterInfo.tsx
```diff
@@ -282,32 +284,41 @@ export const EnterInfo = ({
...
-        <Button
+        <TelemetryButton
           variant="secondary"
           onPress={() => confirmDialogState.close()}
           className={styles.startButton}
+          event={{
+            module: "button",
+            name: "click",
+            data: {
+              button_id: "edit_free_scan",
```
Vinnl, line 295: I think this is already handled in the `onOpenChange` handler on line 95, no?

(Though props for the eagle eye in checking that Tony wanted this event!)

src/app/(proper_react)/(redesign)/(authenticated)/user/welcome/EnterInfo.tsx
```diff
@@ -403,7 +400,8 @@ export const EnterInfo = ({
             </Button>
           )}
           <Button
-            {...confirmDialogTrigger.triggerProps}
+            onPress={() => setConfirmDialogIsOpen(true)}
+            aria-haspopup="dialog"
```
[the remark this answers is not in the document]
Vinnl: Weird that `triggerProps` doesn't set it then! Since this is somewhat surprising (i.e. the expectation would be that react-aria sets the correct `aria-*` attributes as needed), it might be a good idea to add a comment so that someone coming across this in the future knows if that exception still holds. And the docs you so helpfully linked are helpful to future readers too :)

```suggestion
            // react-aria sets `aria-expanded`, which causes this to be
            // announced as "collapsed", rather than opening a dialog. By
            // setting aria-haspopup, this interaction is clearer. See also:
            // https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-haspopup
            aria-haspopup="dialog"
```

src/app/(proper_react)/(redesign)/(authenticated)/user/welcome/EnterInfo.tsx
```diff
@@ -402,14 +418,23 @@ export const EnterInfo = ({
...
             variant="primary"
             type="submit"
             className={styles.startButton}
+            event={{
+              module: "popup",
+              name: "view",
+              data: {
+                popup_id: "enter_scan_info_confirmation_modal",
+              },
+            }}
```
Vinnl, line 434: And this I think is already covered by the `onOpenChange` handler on line 91?

src/app/(proper_react)/(redesign)/(authenticated)/user/welcome/Onboarding.test.tsx
```diff
@@ -201,6 +201,19 @@ it("doesn't allow proceeding without typing a valid location", async () => {
...
+  render(<ComposedOnboarding stepId="enterInfo" />);
+
+  expect(screen.queryByRole("dialog")).not.toBeInTheDocument();
+  const firstNameField = screen.getByLabelText("First name*");
+
+  await user.click(firstNameField);
+  await user.keyboard("{Tab}");
+  expect(firstNameField.getAttribute("aria-invalid")).toBe("true");
+});
+
```
Vinnl, line 216: :clap: :tada:

Vinnl: Just wanted to emphasise some more, this is a great thing to test :)

src/app/(proper_react)/(redesign)/(authenticated)/user/welcome/EnterInfo.tsx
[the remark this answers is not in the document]
Vinnl: I'm even slightly tempted to file a bug at react-aria, but I'm not confident enough about what the pattern should be :sweat_smile:

Vinnl (approved):
Awesome job, works just as on `main` for me now, but now with the clearer announcement :)

src/app/(proper_react)/(redesign)/(authenticated)/user/welcome/EnterInfo.tsx
```diff
@@ -404,6 +413,12 @@ export const EnterInfo = ({
           )}
           <Button
             {...confirmDialogTrigger.triggerProps}
+            // react-aria sets `aria-expanded`, which causes this to be
+            // announced as "collapsed", rather than opening a dialog. By
+            // setting aria-haspopup, this interaction is clearer. See also:
+            // https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-haspopup
+            aria-haspopup="dialog"
+            aria-expanded={undefined}
```
Vinnl, line 421: Feeding back [from Jira](https://mozilla-hub.atlassian.net/browse/MNTOR-2765?focusedCommentId=859950) in case someone (probably me :sweat_smile: ) looks at this PR in the future: the problem was that when the form had errors, no dialog would pop up when clicking the button, which makes the "collapsed" announcement confusing.

Another solution would have been to simply disable the button until there were no errors in the form.

PR #4305, opened 2024-03-06
More arbitrary c8 ignores

Vinnl:
As we add code, more existing code suddenly gets marked as uncovered by unit tests intermittently. This adds c8 ignore comments for everything that turned up when running

   npm test -- --no-cache --runInBand

But I really hope this gets fixed upstream soon, because this isn't sustainable.

For more context, see:
https://github.com/mozilla/blurts-server/pull/4234

And the (hopefully correct) upstream issue:
https://github.com/istanbuljs/v8-to-istanbul/issues/236

Ugh. I feel dirty.


flozia (approved):
Thanks for the additions! It is very unfortunate that there seems to be no obvious pattern as to why there are more and more uncovered lines appearing.

PR #4335, opened 2024-03-14
Only run glean-probe-scraper if we modified metrics

Vinnl:
Probably a good idea to also have someone from Glean take a look at this - @rhelmer, do you know who to tag?

(Note: the easiest way to test this is first merging it to `main`, otherwise we'd have to change the default branch IIRC.)

We'll also need to make the scraper job non-blocking, since it might not have run.


pdehaan:
Looks like the original [committers](https://github.com/mozilla/blurts-server/commits/main/.github/workflows/glean-probe-scraper.yml) of this file were @rhelmer and @badboy.

If we want to assign _ownership_ of this file (for future PR reviewer-ing), we could also add some GitHub usernames in [docs/CODEOWNERS](https://github.com/mozilla/blurts-server/blob/main/docs/CODEOWNERS) like we do for L10n and SRE for DB migrations (which now makes me wonder if that actually works since the db/migrations/* seems to be under the src/db/... folder now).

badboy:
Oh, I definitely do not need ownership of this :D

Vinnl:
Haha I get that @badboy, but does the approach in this PR make sense? Just wondering if e.g. the list of files to watch is not complete, or if it _needs_ to run even if nothing changed.

badboy (approved):
List looks good!
Restricting that to only when those change sounds reasonable (We should mention that in our docs too)

rhelmer (approved):

PR #4336, opened 2024-03-15
MNTOR-2789 - Add exposure details aria-label to exposure card chevron

codemist:
# References: 
Jira: MNTOR-2789


src/app/components/client/exposure_card/ScanResultCard.tsx
```diff
@@ -246,15 +246,12 @@ export const ScanResultCard = (props: ScanResultCardProps) => {
             className={styles.chevron}
             onClick={() => props.onToggleExpanded()}
             aria-expanded={props.isExpanded}
+            aria-label={`${l10n.getString("chevron-alt")} ${props.isExpanded ? l10n.getString("chevron-expanded") : l10n.getString("chevron-collapsed")}`}
```
Vinnl, line 249: As I understood the ticket, we don't need to manually add "Expanded" or "Collapsed" to this string - the screen reader will announce that by itself, thanks to the `aria-expanded` attribute. So currently when I use my screen reader, this announces "Exposure details collapsed collapsed".

(I'd also expect this label to be used as the `alt` for the image as we did before, as `aria-label` is a last resort for when there is no content, but I think in practice the effect will be the same.)

```suggestion
            aria-label={l10n.getString("chevron-alt")}
```

src/app/components/client/exposure_card/SubscriberBreachCard.tsx
```diff
@@ -209,18 +209,14 @@ export const SubscriberBreachCard = (props: SubscriberBreachCardProps) => {
             // TODO: Add unit test when changing this code:
             /* c8 ignore next */
             onClick={props.onToggleExpanded}
+            aria-expanded={props.isExpanded}
+            aria-label={`${l10n.getString("chevron-alt")} ${props.isExpanded ? l10n.getString("chevron-expanded") : l10n.getString("chevron-collapsed")}`}
```
Vinnl, line 213: (Same as above)

```suggestion
            aria-label={l10n.getString("chevron-alt")}
```

Vinnl (commented):
I get the "Expanded" or "Collapsed" announcement twice now in my screen reader, unfortunately. Though probably a simple fix? (And would allow the removal of two strings?)

src/app/components/client/exposure_card/ScanResultCard.tsx
Vinnl: You can just select the button with the same string. If you want to verify whether the button indicates expanded/collapsed state, you could e.g. do

```
expect(expandButton).toHaveAttribute("aria-expanded", "true")
```

flodolo (approved):

Vinnl (approved):
Thank you!

PR #4348, opened 2024-03-18
MNTOR-2790 - Do not announce visually hidden elements

codemist:
# References: 
Jira: MNTOR-2790

### DIFF
--- a/.github/workflows/docker_build_deploy.yml
+++ b/.github/workflows/docker_build_deploy.yml
@@ -3,8 +3,6 @@ name: Build Docker image and publish
 on:
   push:
     branches: [ main ]
-    tags: '*'
-
 jobs:
   push_to_registry:
     name: Push Docker image to Docker Hub
--- a/locales/vi/dashboard.ftl
+++ b/locales/vi/dashboard.ftl
@@ -26,7 +26,6 @@ exposure-chart-caption-fixed = Biểu đồ này hiển thị tổng số dữ l
 exposure-chart-returning-user-upgrade-prompt = Chưa bao gồm địa chỉ nhà, thành viên gia đình và nhiều thông tin khác.
 exposure-chart-returning-user-upgrade-prompt-cta = Bắt đầu quét miễn phí
 exposure-chart-scan-in-progress-prompt = <b>Đang quét:</b> chưa bao gồm địa chỉ, thành viên gia đình và nhiều thông tin khác.
-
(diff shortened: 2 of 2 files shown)


Vinnl (commented):
There should be a way easier way to do this; instead of `@include visually-hidden`, set `display: none` in `hideOnMobile`. The entire point of `visually-hidden` is to make it invisible to everyone except screen readers, which I suppose we thought was better in this situation - but if it _should_ also be hidden for screen reader users, we don't need to use it.

(Sorry if you spent a lot of time on this already :/ )

Vinnl:
@codemist So wouldn't adding a `hideOnMobile` class to those `dt` elements (in addition to `visuallyHidden`) solve that?

src/app/components/client/exposure_card/ScanResultCard.tsx
```diff
@@ -202,7 +203,10 @@ export const ScanResultCard = (props: ScanResultCardProps) => {
       <div className={styles.exposureCard}>
         <div className={styles.exposureHeader}>
           <dl className={styles.exposureHeaderList}>
-            <dt className={styles.visuallyHidden}>
+            <dt
+              className={styles.visuallyHidden}
+              aria-hidden={props.isWideScreen ? false : true}
```
pdehaan, line 208: Since `props.isWideScreen` is already Boolean, does this need the ternery, or could we do something like:
```suggestion
              aria-hidden={!props.isWideScreen}
```

Not sure if that is better in any way, but it is shorter by about 14 chars. 🤷

Vinnl (approved):
Sorry for the delay in reviewing, but this works well for me now - the hidden elements no longer get announced!

src/app/components/client/exposure_card/ExposureCard.module.scss
```diff
@@ -48,6 +48,10 @@
     justify-content: space-between;
     height: 30px; // fixed height to standardize image/icon heights
 
+    .hideOnMobile {
+      display: none;
+    }
```
Vinnl, line 53: Oh sorry, actually, I just did notice another problem: `.hideOnMobile` is not limited to mobile (i.e. the media query that shows it on desktop does not apply to the `dt`), and so it no longer gets announced when the value _is_ visible.

Vinnl (changes requested):
Sorry, encountered one more problem :(

Vinnl (approved):
Works as expected now, thank you!

PR #4351, opened 2024-03-20
Remove some more unused code and dependencies

Vinnl:
Followup to #4332. There's more to delete after we've migrated the emails though :)


src/app/functions/server/changeSubscription.ts
```diff
@@ -28,3 +28,20 @@ export async function changeSubscription(
   currentFxAProfile.subscriptions = subscriptions;
   await updateFxAProfileData(subscriber, JSON.stringify(currentFxAProfile));
 }
+
+/**
+ * See https://github.com/mozilla/fxa/blob/564949dfc69f0f675ebb4e5f267282c2546a5767/packages/fxa-profile-server/lib/routes/profile.js#L63-L77
+ */
+type FxaProfile = {
```
Vinnl, line 35: I moved this from the now-deleted `customTypes.d.ts`, since this didn't need to be an implicitly-available type - it's only used in this file.

rhelmer (approved):

PR #4353, opened 2024-03-20
Add example of how to use Fluent in emails

Vinnl:
**Note: there's lots of noise due to renaming the l10n function; important changes are in 08323b839395cbc21f013c50d090f386b3560cb8.**

With this approach, individual emails consist of two parts: an MJML
template, and a render function that binds data to variables inside
those templates, and then returns an HTML string that's ready to
send.

Note that for this approach, I did not use EJS, for two reasons.
First, we already have two other templating systems in our codebase
(specifically, MJML and React), and I didn't want to add another
syntax — especially considering that we'd use them inside the same
files, which no current syntax highlighter understands.
And second, we *also* already have a library that can inject
variables: Fluent.

So instead, I just wrote a small regular expression that matches
`<%= strings-like-this %>` and replaces them by Fluent strings, which
have their variables already injected into them.

The main open question is whether we also want to support using JSX
inside the Fluent strings, which will make things easier for
localisers (who can just use `<custom-tags>like this</custom-tags>`,
without having to worry about e.g. HTML attributes), but will
surely make generating the templates at runtime more costly, and
the related toolchain more complex.

A possible compromise would be to add a small regular expression to
the Fluent instance that replaces `<attributeless-tags>` with full
tags with the relevant attributes, as strings.


rhelmer:
I think this is a reasonable approach overall... I still think using EJS is better than trying to do it ourselves though, I'm concerned about any kind of user-controlled data that may be used in templates (template authors shouldn't be expected to have to do this, I don't think).

This isn't really an issue in our proposed templates, but feels like a footgun to have around (e.g. someone wanted to use the name or email address in the body of the email). The other part of this is hooking up our backend to it, I can do that in my PR. Thanks for dealing with all the Fluent parts in this one! I might merge and make a few small changes and add the backend parts in my PR this is branched from.

rhelmer (approved):
I'd like to see if @flodolo has any thoughts on this, I'll work on the backend part in the other PR and merge this in if this is OK.

flodolo:
I don't think I have a strong opinion about this, as it seems more relevant for developers than localizers.

If you want a more technical opinion, looping in @eemeli might help.

src/emails/templates/example/renderExampleEmail.tsx
```diff
@@ -0,0 +1,50 @@
...
+  const options: Intl.DateTimeFormatOptions = {
+    year: "numeric",
+    month: "long",
+    day: "numeric",
+  };
+  const dateFormatter = new Intl.DateTimeFormat(getLocale(l10n), options);
+
+  return renderMjml(template, {
+    "{data-breach-header-1}": l10n.getString("data-breach-header-1"),
+    "{data-breach-body-1}": l10n.getString("data-breach-body-1", {
```
eemeli, line 30: I find using the full message id in the template rather odd, as it suggests to a reader that they might _need_ to be the same, but that's just a name that's local to the template and this function.

src/emails/templates/example/renderExampleEmail.tsx
```diff
@@ -0,0 +1,50 @@
...
+    "{subscriberName}":
+      params.subscriber.fxa_profile_json?.name ?? "mr. person",
+    // TO BE DECIDED: Supporting `getFragment` realistically means parsing JSX
+    //                when rendering emails. Is that worth it, or do we just
+    //                want to hardcode "real" tags with variables in
+    //                attributes in our .ftl files?
+    //                (i.e. do we want to make life harder for localisers, or
+    //                for the email server?)
+    //                Possibly, we could consider getting the same effect with
+    //                regular expressions, returning strings instead of JSX.
```
eemeli, line 43: Won't you need to either parse JSX or do something comparable just to sanitize user inputs?

You want to make life harder for the email server.

Why would you use regular expressions to parse an XML-ish language?

src/emails/templates/example/index.mjml
```diff
@@ -0,0 +1,18 @@
+<mjml>
+  <mj-body>
+    <mj-section>
+      <mj-column>
+        <mj-text>
+          <!-- Rendering a Fluent string: -->
+          <h1>{data-breach-header-1}</h1>
+          <!-- Rendering a variable directly, without going through Fluent: -->
+          <p>Hello {subscriberName},</p>
```
eemeli, line 9: Showing this unlocalised string concatenation in the example is probably a bad idea, as it suggests that doing this would be ok.

src/emails/templates/example/renderExampleEmail.tsx
Vinnl: Good point, that can be confusing. Renamed in 757251a86.

src/emails/templates/example/renderExampleEmail.tsx
Vinnl: Yes, I wrote this comment before I added that :) Removed in 0ab80e833.

Vinnl:
I'm going to close this and create a new PR including all the changes for `main`, as I'm taking over from Rob to get this over the finish line.

PR #4355, opened 2024-03-21
Remove unused strings

Vinnl:
# References: 
Jira: MNTOR-2682 and MNTOR-1562

It's finally happening! I wrote a quick script to find unused strings (present in Git history beefcdda755aa84cbabe8f9341b286fdad2a6890, but removed again because it's a one-time script only), and then mostly manually inspected those strings and removed them if I thought they were unused (dffb215931f349855e69f17bfa8201ff98c77312).

I think there were some false negatives, so there might be more to clean up later, but this should already help a lot. False positives might exist too, primarily if we don't use the literal message IDs, but instead obtain them in code (e.g. via string concatenation). I caught the ones in `data-classes.ftl`, but hopefully there weren't others that I missed. If there are, I think they'll show up in Sentry soon, and then I'll restore them.


Vinnl:
Btw @pdehaan just noting that it's [a new day today](https://github.com/mozilla/blurts-server/pull/4332#issuecomment-2010187192).

flodolo (approved):
Thanks for doing this.

Just to be on the safe side, I'll wait a few days before merging this one in the l10n repo, to make sure we have time to look at logs.

Vinnl:
@flodolo Is that something you do separately, or should I hold off on merging this PR?

flodolo:
> @flodolo Is that something you do separately, or should I hold off on merging this PR?

It's separate (a workflow in the l10n repo). I've just manually disabled the [workflow](https://github.com/mozilla-l10n/monitor-website-l10n/actions/workflows/import.yaml), which would otherwise run every morning, as I just exposed the new strings today (hopefully there won't be new content soon).

pdehaan:
Another 10/10 PR! `+1 −959`

Vinnl:
@flodolo I just realised:

> to make sure we have time to look at logs.

I don't think we'll see anything in the logs until we merge it back into the l10n repo - if the strings still exist in the localised files, then the website will display those rather than logging an error about strings that it couldn't find.

flodolo:
> I don't think we'll see anything in the logs until we merge it back into the l10n repo - if the strings still exist in the localised files, then the website will display those rather than logging an error about strings that it couldn't find.

How so? It should show errors for en-US, since the files in this repo are already missing the strings you removed.

Vinnl:
> How so? It should show errors for en-US, since the files in this repo are already missing the strings you removed.

I thought (though I'd have to check to be sure) that if the strings for `en` aren't found, Fluent will fall back to one of the other `en`s (`en-GB` or `en-CA`)?

flodolo:
Oh, I didn't think about the fallback 🤔 

If you set your accept-languages to just en-US? (from about:config, edit directly `intl.accept_languages`)

flodolo:
I gues one drastic option would be to remove all the other English variants locally?

I suppose we can merge and see what happens. I would still wait for when I'm back (April 2nd), as I might have to do some heavy lifting to recover translations in case we removed string still in use.

Vinnl:
That sounds good to me!

PR #4359, opened 2024-03-22
Write MJML templates directly in JSX

Vinnl:
@rhelmer Here's another, even simpler, approach to MJML templates - write them in React, like we do with our HTML. It will make sure everything is escaped, and we can use the full power of JS to load strings, run loops, format dates, etc, without the indirection of having to pass variables into the MJML.


src/emails/templates/mjml.d.ts
```diff
@@ -2,8 +2,72 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
-declare module "*.mjml?raw" {
```
rhelmer, line 5: Would be nice if we didn't have to maintain this... there is https://www.npmjs.com/package/@types/mjml but that's for the (server-side) function definitions. I guess from https://github.com/mjmlio/mjml/issues/930 there is probably not much interest in them maintaining this upstream, but we'll need to make sure to keep it up to date - the TS compiling will not guarantee that the template will be rendered to HTML.

src/emails/templates/boilerplate/BoilerplateEmail.tsx
```diff
@@ -0,0 +1,26 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
```
rhelmer, line 4: We discussed already that this can't `use client` or `use server` since for Storybook we need client support but the emails will always be rendered by the server. Email is unusual in that it's just a static page we're generating, there is no client or server runtime, so just including an existing component used by the website is unlikely to work.

Could we have a comment here to that effect? Too bad there is no `use email` ;)

rhelmer (approved):
I left a suggestion to comment in the boilerplate email component about why this cannot `use client` or `use server` and explaining the context a little. Otherwise looks great, I like the idea to use JSX instead of introducing yet another templating language.

src/emails/templates/mjml.d.ts
Vinnl: Yeah, I was also looking at that. There is also [mjml-react](https://www.npmjs.com/package/mjml-react) for which (2-years-old) type definitions are available, but that does the whole thing where it creates specialised component wrappers that convert camelCased attributes to kebab-cased ones, which makes it more of a blocking dependency when it's not updated.

So I don't have a _strong_ opinion, but I would be leaning towards just adding our own type definitions. I think I can cover basically the entire API in half an hour (I just didn't do it yet since I didn't know we'd use this), and we'd only really have to update it if we want to use a changed API in the future - which would be a matter of seconds to do, and we'd be in control ourselves.

src/emails/templates/boilerplate/BoilerplateEmail.tsx
Vinnl: Good callout, added in e6b8d96c6.

src/emails/templates/mjml.d.ts
Vinnl: Took that half-hour just now to finish these up: 1e8ae4eaf

I think that should allow us to create all the templates we want for the foreseeable future.

PR #4369, opened 2024-03-26
Set up email template infrastructure

Vinnl:
# References: 
Jira: MNTOR-3036

This bundles up the MJML work we've done so far in atomic commits (5b95d9cad7e77401e7b60decebe105ed3f3f05bd for the email-specific bits, rather than the preparatory Fluent work) to merge into `main`. I'll submit a followup PR to start using this for an actual email (behind a flag).

See [an example render in Storybook](https://deploy-preview-4369--fx-monitor-storybook.netlify.app/?path=/story/emails-email-boilerplate--boilerplate-email-story).


src/emails/getEmailL10n.ts
```diff
@@ -0,0 +1,20 @@
...
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import { SubscriberRow } from "knex/types/tables";
+import { acceptedLanguages, negotiateLanguages } from "@fluent/langneg";
+import { ExtendedReactLocalization } from "../app/hooks/l10n";
+import { getSpecificL10nSync } from "../app/functions/server/mockL10n";
+
+export function getEmailL10n(
+  subscriber: SubscriberRow,
```
rhelmer, line 11: Not critical, but I wonder if instead of using `SubscriberRow` we should have more constrained types - e.g. we could `Pick` as a way to allow-list. There are properties in `SubscriberRow` we would never want to appear in an email.

rhelmer, line 11: Could we use a more specific type, e.g. using `Pick`? `SubscriberRow` is fairly broad and contains data we wouldn't want to appear in an email.

src/emails/templates/boilerplate/BoilerplateEmail.tsx
```diff
@@ -0,0 +1,48 @@
...
+
+import { SubscriberRow } from "knex/types/tables";
+import { getEmailL10n } from "../../getEmailL10n";
+
+export type Props = {
+  subscriber: SubscriberRow;
+};
+
+/**
+ * Minimal boilerplate for email templates
```
rhelmer, line 13: Excellent comment, thanks!

src/emails/templates/boilerplate/BoilerplateEmail.stories.tsx
```diff
@@ -0,0 +1,30 @@
...
+  title: "Emails/Email boilerplate",
+  component: (props: Props) => (
+    <StorybookEmailRenderer>
+      <BoilerplateEmail {...props} />
+    </StorybookEmailRenderer>
+  ),
+  args: {
+    subscriber: {
+      signup_language: "en",
+    } as SubscriberRow,
```
rhelmer, line 21: Same as above re: `SubscriberRow`, I know we tend to pass this around but I think allow-listing the properties would be prevent someone from accidentally exposing something in the email that shouldn't be there.

rhelmer (approved):
This looks great, thanks! I have a suggestion to scope down the `SubscriberRow` type but I don't think it's blocking, would like to hear what you think about that.

src/emails/getEmailL10n.ts
Vinnl: That's a great point. I've implemented that in the followup PR at https://github.com/mozilla/blurts-server/pull/4370.

PR #4370, opened 2024-03-27
Use MJML template for verification email for users with the `RedesignedEmails` flag enabled

Vinnl:
# References: 
Jira: MNTOR-3038
Figma: N/A

# Description

This PR does a couple of things:

- It uses the MJML template when sending the verification email if the user has the `RedesignedEmails` flag enabled (6b8f67a34a2681b8b0c3a51d4a48ee823dd3a784)
- Updates the verification email to use the same Fluent setup as the website (364e9559fa9cebee2a5a3f928ae0353fb9ac40b2)
- Guards against exposing sensitive data in emails, as suggested in https://github.com/mozilla/blurts-server/pull/4369#discussion_r1539814657 (d1b333c0fc1aff8f31f38f8f7d4ac664de237798)

# How to test

Give a user the `RedesignedEmails`, then add an email address on the settings page. The verification email (also when clicking "resend") should look different. (The [specific look](https://deploy-preview-4370--fx-monitor-storybook.netlify.app/?path=/story/emails-verify-email-address--verify-email-address-email-story) is not relevant yet; this just demonstrates that we can use the template.)

# Checklist (Definition of Done)
- [x] Localization strings (if needed) have been added.
- [x] Commits in this PR are minimal and [have descriptive commit messages](https://chris.beams.io/posts/git-commit/).
- [x] I've added or updated the relevant sections in readme and/or code comments
- [x] I've added a unit test to test for potential regressions of this bug.
- [ ] Product Owner accepted the User Story (demo of functionality completed) or waived the privilege.
- [x] All acceptance criteria are met.
- [x] Jira ticket has been updated (if needed) to match changes made during the development process.
- [x] Jira ticket has been updated (if needed) with suggestions for QA when this PR is deployed to stage.

PR #4385, opened 2024-04-02
Exit early if we can't parse the email ID

Vinnl:
As split off from https://github.com/mozilla/blurts-server/pull/4370#discussion_r1544812661.


mansaj (approved):
lgtm

PR #4394, opened 2024-04-05
MNTOR-3080: change the API to accept both instant alerts and monthly report

mansaj:
# References: 
Jira: MNTOR-3080
Figma: https://www.figma.com/file/CaEKIhvSJqf6KNIMzSkt40/Concepts-for-Monitor-MVP-Redesign?type=design&node-id=2094-87130&mode=design&t=KYeYj6gb0KmNkKGO-4

# Description
Modify the existing `update comms` API to take in two attributes:
1. instant alerts
2. monthly monitor report

Where the first variable can potentially have 3 values: `null, true, false`

# Screenshot (if applicable)

Not applicable.

# How to test


mansaj:
First part of the API changes, to accept `-1` as a value for `null` @codemist

src/app/api/v1/user/update-comm-option/route.ts
```diff
@@ -33,9 +33,13 @@ export async function POST(req: NextRequest) {
...
         throw new Error("No subscriber found for current session.");
       }
+      // -1 = Set as null, do not send instant notifications. Newly added in MNTOR-1368
       // 0 = Send breach alerts to the corresponding affected emails.
       // 1 = Send all breach alerts to user's primary email address.
-      const allEmailsToPrimary = Number(communicationOption) === 1 ?? false;
+      const allEmailsToPrimary =
+        Number(communicationOption) === -1
+          ? null
+          : Number(communicationOption) === 1 ?? false;
```
pdehaan, line 42: I think the `?? false` is a but suspect here. The `Number(value) === 1` should already cast out to a Boolean value, which should mean the nullish coalescing shouldn't be reached since any invalid value to the `Number()` constructor would return as `NaN` instead of `null|undefined`.

```js
const options = [null, "-1", "0", "1", 2, "socks"];

for (const opt of options) {
  const toPrimary = Number(opt) === -1 ? null : Number(opt) === 1 ?? "nullish thing";
  console.log({ opt, cast: Number(opt), toPrimary });
}

// OUTPUT:
{opt: null, cast: 0, toPrimary: false}
{opt: '-1', cast: -1, toPrimary: null}
{opt: '0', cast: 0, toPrimary: false}
{opt: '1', cast: 1, toPrimary: true}
{opt: 2, cast: 2, toPrimary: false}
{opt: 'socks', cast: NaN, toPrimary: false}
```

pdehaan: Not convinced this is any better than the double ternery. Meh.
```js
const options = [null, "-1", "0", "1", 2, "socks"];

for (const opt of options) {
  const toPrimary = opt === "-1" ? null : !!Number(opt);
  console.log({ opt, cast: Number(opt), toPrimary });
}

// OUTPUT:
{ opt: null, cast: 0, toPrimary: false }
{ opt: '-1', cast: -1, toPrimary: null }
{ opt: '0', cast: 0, toPrimary: false }
{ opt: '1', cast: 1, toPrimary: true }
{ opt: 2, cast: 2, toPrimary: true } // SLIGHTLY DIFFERENT BEHAVIOR DUE TO `!!` bool casting
{ opt: 'socks', cast: NaN, toPrimary: false }
```

mansaj: changed it to ternery

src/db/tables/subscribers.js
```diff
@@ -193,6 +193,27 @@ async function setAllEmailsToPrimary (subscriber, allEmailsToPrimary) {
 }
 /* c8 ignore stop */
 
+/**
+ * @param {import("knex/types/tables").SubscriberRow} subscriber
+ * @param {boolean | null} monthlyMonitorReport
```
Vinnl, line 198: Not sure I understand - why should it be possible for this value to be `null`? (And related: why do we need `|| false` on line 206? i.e. if the reason is that we want to be able to distinguish between users who have actively chosen whether to receive the email, and those who haven't made a choice, then shouldn't we also store the `null` in the DB?)

```suggestion
 * @param {boolean} monthlyMonitorReport
```

src/app/api/v1/user/update-comm-option/route.ts
```diff
@@ -19,24 +20,34 @@ import {
...
+        instantBreachAlerts,
+        monthlyMonitorReport,
+      }: EmailUpdateCommOptionRequest = await req.json();
       const subscriber = await getSubscriberByFxaUid(token.subscriber?.fxa_uid);
       if (!subscriber) {
         throw new Error("No subscriber found for current session.");
       }
+      // -1 = Set as null, do not send instant notifications. Newly added in MNTOR-1368
       // 0 = Send breach alerts to the corresponding affected emails.
       // 1 = Send all breach alerts to user's primary email address.
```
Vinnl, line 42: Would it be an idea to use string constants here, instead of numbers? (Possibly with a temporary fallback to also accept `0` and `1` to accommodate pods running different code during deployment, though that's probably overkill.)

Vinnl (commented):
I've mainly got a question about why the API is taking `null` - I think I'm misunderstanding what this needs to support.

src/app/api/v1/user/update-comm-option/route.ts
```diff
@@ -19,24 +20,34 @@ import {
...
-      const allEmailsToPrimary = Number(communicationOption) === 1 ?? false;
+      const allEmailsToPrimary =
+        Number(instantBreachAlerts) === -1
+          ? null
+          : Number(instantBreachAlerts) === 1
+            ? true
+            : false;
```
rhelmer, line 48: I find this double-ternary kind of difficult to read... you might consider defining `Number(instantBreachAlerts)` to a short variable once. I suspect the `? true : false` part is probably not necessary since `Number(instantBreachAlerts) === 1` the `===` will always evaluate to `true` or `false`

rhelmer (changes requested):

src/app/api/v1/user/update-comm-option/route.ts
mansaj: this whole number thing seems silly (re @Vinnl's comment above).. let's just change it to string enums

src/app/api/v1/user/update-comm-option/route.ts
mansaj: Good idea, changed to strings

src/db/tables/subscribers.js
```diff
@@ -193,6 +193,27 @@ async function setAllEmailsToPrimary (subscriber, allEmailsToPrimary) {
...
+ * @param {import("knex/types/tables").SubscriberRow} subscriber
+ * @param {boolean} monthlyMonitorReport
+ */
+// Not covered by tests; mostly side-effects. See test-coverage.md#mock-heavy
+/* c8 ignore start */
+async function setMonthlyMonitorReport (subscriber, monthlyMonitorReport) {
+  const updated = await knex('subscribers')
+    .where('id', subscriber.id)
+    .update({
+      monthly_monitor_report: monthlyMonitorReport || false,
```
Vinnl, line 206: If this parameter inadvertently doesn't get passed, we don't necessarily want to overwrite whatever's in there already, right?

```suggestion
      monthly_monitor_report: monthlyMonitorReport,
```

src/app/api/v1/user/update-comm-option/route.ts
```diff
@@ -11,32 +11,45 @@ import AppConstants from "../../../../../appConstants";
...
+        instantBreachAlerts,
+        monthlyMonitorReport,
+      }: EmailUpdateCommOptionRequest = await req.json();
       const subscriber = await getSubscriberByFxaUid(token.subscriber?.fxa_uid);
       if (!subscriber) {
         throw new Error("No subscriber found for current session.");
       }
+      // -1 = Set as null, do not send instant notifications. Newly added in MNTOR-1368
       // 0 = Send breach alerts to the corresponding affected emails.
       // 1 = Send all breach alerts to user's primary email address.
```
Vinnl, line 39:
```suggestion
      // "null"     = Do not send instant notifications. Newly added in MNTOR-1368
      // "affected" = Send breach alerts to the corresponding affected emails.
      // "primary"  = Send all breach alerts to user's primary email address.
```

src/knex-tables.d.ts
```diff
@@ -91,7 +91,7 @@ declare module "knex/types/tables" {
     // TODO: Find unknown type
     // eslint-disable-next-line @typescript-eslint/no-redundant-type-constituents
     breaches_last_shown: null | unknown;
-    all_emails_to_primary: boolean;
+    all_emails_to_primary: boolean | null; // added  null in MNTOR-1368
```
Vinnl, line 94: Depending on how big of a lift the migration would be: maybe we should rename this to `alert_emails_to_primary`? Since other emails (e.g. the monthly activity email) will always be sent to the primary address, regardless of the value.

(And if the migration is too big of a lift, maybe we want to call that out in a JSDoc comment at least?)

Vinnl (approved):
Thanks for the changes! Some small suggestions inline.

locales/en/settings.ftl
```diff
@@ -8,9 +8,16 @@ settings-page-title = { -product-short-name } Settings
 
 ## Breach alert preferences
 
+# Obsolete
```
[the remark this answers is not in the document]
mansaj: I branched this off of https://github.com/mozilla/blurts-server/tree/mntor-1368 and just pulled the latest from there.

src/knex-tables.d.ts
mansaj: Renaming a column can be more complicated than it seems.. this thread gives some practical steps to do so in production: 

https://stackoverflow.com/questions/72278957/how-to-practically-rename-tables-and-columns-in-postgresql-on-a-production-syste

I think in this case, since it doesn't seem like the name is too far off, I'd say the risk outweighs the benefits

I'll add a comment

PR #4399, opened 2024-04-08
MNTOR-3079: add monthly report column

mansaj:
# References: 
Jira: MNTOR-3079

# Description
Separate out the db migration part for ease of prod deployment / cherry picking


src/db/migrations/20240408161125_add_monthly_report_column.js
```diff
@@ -0,0 +1,15 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+export function up (knex) {
+  return knex.schema.table('subscribers', table => {
+    table.boolean('monthly_monitor_report').defaultTo(false)
```
rhelmer, line 7: How is this different from `monthly_email_optout` column?

rhelmer (changes requested):

src/db/migrations/20240408161125_add_monthly_report_column.js
mansaj: The `monthly_email_optout` column was used for a different email campaign. It's used by a script via `getSubscribersWithUnresolvedBreaches`. 

What we building here is a completely different feature, which should not inherit the values that exist in that column

mansaj: We *can* delete that column along with `getSubscribersWithUnresolvedBreaches` and the script `send-email-to-unresolved-breach-subscribers.js` but that should be a separate issue

rhelmer: Well I ask because I think @Vinnl is using the old column in https://github.com/mozilla/blurts-server/pull/4400 :)

mansaj: Ah yes, @Vinnl I think if that's the case, we should probably clear out the column first (see my point above about existing values not representing this selection)

Also I quite dislike the idea of a column with names like `opt_out`, which forces us to reverse the boolean :(.. this probably also applies to the existing associated timestamp column

Vinnl: It's probably a good idea to delete that column and the script for this exact reason :) But I'm happy to update my code to use a different column with a different value, just let me know which.

mansaj: @Vinnl @rhelmer updated with the removal of the two old columns, https://github.com/mozilla/blurts-server/pull/4400#discussion_r1558276218

src/db/migrations/20240408161125_add_monthly_report_column.js
```diff
@@ -0,0 +1,21 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+export function up (knex) {
+  return knex.schema.table('subscribers', table => {
+    table.dropColumn("monthly_email_optout")
+    table.dropColumn("monthly_email_at")
```
rhelmer, line 8: These drops might take a while to run.

Are we sure that they don't need to be migrated at all, are there existing accounts that have opted-out that will be surprised to receive this? Or are we already accounting for that because we are restricting the date that the email is sent?

src/db/migrations/20240408161125_add_monthly_report_column.js
Vinnl: I have no insight in running time of the migration, so I'll leave the full review to Rob, but two remarks:

- Does Tony actually want this to default to `false`? Or should we set it to `false` for existing or non-Plus subscribers, but default to `true` for new ones?
- Don't forget to update the type definitions in `knex-tables.d.ts` :) Ideally in the same commit as the one adding the migration script, so it's easy to find the relevant migration by running `git blame`.

src/db/migrations/20240408161125_add_monthly_report_column.js
mansaj: I'm pretty sure these are from the old PM who's no longer here. I can ask Tony about it

mansaj: Also good thinking about the scale concern re `DROP COLUMN`. 

See the postgres `alter table` doc:
```
The DROP COLUMN form does not physically remove the column, but simply makes it invisible to SQL operations. Subsequent insert and update operations in the table will store a null value for the column. Thus, dropping a column is quick but it will not immediately reduce the on-disk size of your table, as the space occupied by the dropped column is not reclaimed. The space will be reclaimed over time as existing rows are updated.
```

So I don't think it's an issue in practice

src/db/migrations/20240408161125_add_monthly_report_column.js
rhelmer: lgtm overall @mansaj but see these comments from @Vinnl. If the default of `false` is find, I think all we need is the type def updates (would be nice if the build caught this somehow).

rhelmer (changes requested):

src/db/migrations/20240408161125_add_monthly_report_column.js
mansaj: @rhelmer just fyi, we actually want it to default to `true` now after talking to prod. 
updated the type def and added some comments for functions that need to be deleted when we tackle tech debt

rhelmer (approved):

mansaj:
@Vinnl merged

PR #4400, opened 2024-04-08
Monthly activity email cron job

Vinnl:
# References: 
Jira: MNTOR-3067
Figma: N/A

# Description

This adds a `MonthlyActivityEmail` flag and a script that will send the monthly activity email to 10 (or `$MONTHLY_ACTIVITY_EMAIL_BATCH_SIZE`) Plus subscribers who have that flag enabled, whose `monthly_email_optout` is not `true` (@mansaj - is this column actually boolean, or is this the one where I need to check for `0` or `1` or `-1` or something) and to whom we've not sent such an email in the last 30 days.

Importantly, it also sets up the build infrastructure to compile JSX and TypeScript for cron job scripts (though note that I haven't updated our cron jobs configuration yet to call the script - which will be at `dist/cronjobs/monthlyActivity.js`). The main thing here to realise is that this means that some code is now built by Next.js (using Webpack and SWC) when shipped as part of the website, and by ESBuild when shipped as part of the cron job - and that these have their own peculiarities. For example, JSX files that are also used in cron jobs need to explicitly import React, need to use a wrapper around `import "server-only"`, need to load Fluent differently (see also [my followup PR](https://github.com/mozilla/blurts-server/pull/4401)), and don't have access to variables from `.env` (possibly we should run `dotenv` for cron jobs like we do for the e2e tests).

I'll call out some more specific things inline.

# How to test

Enable the `MonthlyActivityEmail` flag for your user, then run `npm run dev:cronjobs`. You may want to reset the value of their `monthly_email_at` column afterwards if you want to retry.

- [ ] I've added a unit test to test for potential regressions of this bug. - Somewhat, but there'd also be lots of mocking involved for more.
- [ ] All acceptance criteria are met. - Still need to actually have the cron job executed.


src/app/functions/server/notInClientComponent.ts
Vinnl, line 1: This file is needed because, while `import "server-only"` prevents us from loading a module in a client component, it also breaks when running in plain Node.

src/cronjobs/monthlyActivity.tsx
```diff
@@ -0,0 +1,104 @@
...
+import { getLocale } from "../app/functions/universal/getLocale";
+
+void run();
+
+async function run() {
+  const batchSizeEnvVar = Number.parseInt(
+    process.env.MONTHLY_ACTIVITY_EMAIL_BATCH_SIZE ?? "",
+    10,
+  );
+  const batchSize = Number.isNaN(batchSizeEnvVar) ? 10 : batchSizeEnvVar;
```
Vinnl, line 28: I figured I'd start with (configurable) small batches so that we can just run the cron job more often and not overload the server with a single big one, since we're doing multiple queries per user. Since this is limited to Plus users for now, there's not that many accounts we need to get through anyway. Also makes it easier to recover from errors.

src/cronjobs/monthlyActivity.tsx
```diff
@@ -0,0 +1,104 @@
...
+  const countryCodeGuess = locale.split("-")[1] ?? "us";
+
+  const latestScan = await getLatestOnerepScanResults(
+    subscriber.onerep_profile_id,
+  );
+  const subscriberBreaches = await getSubscriberBreaches({
+    fxaUid: subscriber.fxa_uid,
+    countryCode: countryCodeGuess,
+  });
+  const data = getDashboardSummary(latestScan.results, subscriberBreaches);
```
Vinnl, line 67: I still have to check with Tony whether we want to show totals, or only specifics for the past month.

src/db/tables/subscribers.js
```diff
@@ -271,6 +272,59 @@ async function deleteResolutionsWithEmail (id, email) {
...
+    // Note: This will only match people who had a Plus subscription the last
+    //       time they signed in. We'd have to check with SubPlat directly to
+    //       find out if they have a subscription right now.
+    query = query.andWhereRaw(
+      `EXISTS (
+        SELECT 1
+        FROM jsonb_array_elements(fxa_profile_json->'subscriptions') AS subscription
+        WHERE subscription::text = '"${MONITOR_PREMIUM_CAPABILITY}"'
+      )`
+    );
```
Vinnl, line 315: I know pretty little of SQL, so I'm not sure if this is the best way to check for rows where the `fxa_profile_json` column has an object with a `subscriptions` property containing an array of strings, of which one element is `"monitor"`?

src/emails/getEmailL10n.node.ts
Vinnl, line 1: It's probably best to ignore this file for now - it's basically the existing Next.js `getL10n`, but with changes to make it compatible with plain Node. I've got a followup PR that untangles the five different runtimes that l10n currently has to take into account, and abstracts away the common parts.

src/emails/templates/EmailFooter.tsx
```diff
@@ -2,6 +2,7 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
+import React from "react";
```
Vinnl, line 5: Next.js auto-imports React in JSX files, but esbuild doesn't (at least no by default, maybe there's an option somewhere).

Vinnl (commented):
Called out some things to pay attention to.

esbuild.cronjobs.mjs
```diff
@@ -0,0 +1,26 @@
...
+// import specifiers for us.
+build({
+  entryPoints: ["./src/cronjobs/**/*.tsx"],
+  tsconfig: "tsconfig.cronjobs.json",
+  bundle: true,
+  platform: "node",
+  format: "esm",
+  outdir: "dist/cronjobs",
+  sourcemap: true,
+  target: "node20.12",
```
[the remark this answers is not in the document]
rhelmer: I think it can be just `node` or `node20` if you want to be more specific - do you think it's worth specifying it here? The environment running the cron job (k8s `CronJob` using the container we build in `Dockerfile`) will be what's actually available in the environment.

src/cronjobs/monthlyActivity.tsx
```diff
@@ -0,0 +1,104 @@
...
+    sanitizedSubscriber.primary_email,
+    subject,
+    renderEmail(
+      <MonthlyActivityEmail
+        subscriber={sanitizedSubscriber}
+        data={data}
+        l10n={l10n}
+      />,
+    ),
+  );
```
rhelmer, line 103: So neat to be able to do this!

src/db/tables/subscribers.js
```diff
@@ -8,6 +8,7 @@ import AppConstants from '../../appConstants.js'
 
 const knex = createDbConnection();
 const { DELETE_UNVERIFIED_SUBSCRIBERS_TIMER } = AppConstants
+const MONITOR_PREMIUM_CAPABILITY = "monitor";
```
pdehaan, line 11: Was going to ask if this should be `MONITOR_PLUS_CAPABILITY` but looks like we use this in 3&times; places.
Although not sure if it is worth extracting into somewhere central and exporting:
```sh
git grep -Pn "\bMONITOR_[A-Z_]+\b =" | cat

src/app/api/v1/fxa-rp-events/route.ts:36:const MONITOR_PREMIUM_CAPABILITY = "monitor";
src/app/functions/server/changeSubscription.ts:8:const MONITOR_PREMIUM_CAPABILITY = "monitor";
src/db/tables/subscribers.js:11:const MONITOR_PREMIUM_CAPABILITY = "monitor";
```

rhelmer, line 11: Should this be in `constants.ts`, since it's used in more than one place?

src/db/tables/subscribers.js
```diff
@@ -271,6 +272,59 @@ async function deleteResolutionsWithEmail (id, email) {
...
+  if (Array.isArray(flag.allow_list) && flag.allow_list.length > 0) {
+    // The `.andWhereIn` alias doesn't exist:
+    // https://github.com/knex/knex/issues/1881#issuecomment-275433906
+    query = query.whereIn("primary_email", flag.allow_list)
+  }
+
+  if (options.plusOnly) {
+    // Note: This will only match people who had a Plus subscription the last
+    //       time they signed in. We'd have to check with SubPlat directly to
+    //       find out if they have a subscription right now.
```
rhelmer, line 308: The Monitor database is [updated by webhook](https://github.com/mozilla/blurts-server/blob/288ce9797d87d19c3bf7f1a48ca2de9adb246e60/src/app/api/v1/fxa-rp-events/route.ts) as well, so we don't necessarily need to wait until the user logs in again.

This comment *is* relevant for dev environments that don't have webhook handling set up e.g. `localhost`.

src/db/tables/subscribers.js
rhelmer: I'm not sure if it's any more efficient but there is a `?` operator https://www.postgresql.org/docs/9.4/functions-json.html that can check for existence:

```
WHERE (fxa_profile_json->'subscriptions')::jsonb ? 'monitor'
```

src/db/tables/subscribers.js
```diff
@@ -271,6 +272,59 @@ async function deleteResolutionsWithEmail (id, email) {
...
+    query = query.andWhereRaw(
+      `EXISTS (
+        SELECT 1
+        FROM jsonb_array_elements(fxa_profile_json->'subscriptions') AS subscription
+        WHERE subscription::text = '"${MONITOR_PREMIUM_CAPABILITY}"'
```
rhelmer, line 313: I know it's a `const` in this case but I'd still use a parameterized query here, in case anything changes in the future. You should be able to use it with raw queries as well in knex.

src/db/tables/subscribers.js
```diff
@@ -271,6 +272,59 @@ async function deleteResolutionsWithEmail (id, email) {
...
+    .where("name", featureFlagName);
+
+  if (!flag?.is_enabled) {
+    return [];
+  }
+
+  let query = knex('subscribers')
+    .select()
+    .where((builder) => builder.whereNull("monthly_email_optout").orWhere("monthly_email_optout", false))
+    .andWhere(builder => builder.whereNull("monthly_email_at").orWhereRaw('"monthly_email_at" < NOW() - INTERVAL \'30 days\''));
```
rhelmer, line 297: So this avoids re-sending because `updateMonthlyEmailTimestamp()` is called which sets this, right? I just want to make sure we're thinking through that path if there are any exceptions, so we can't e.g. send duplicate emails because something there fails after the email is sent but before this timestamp is updated.

rhelmer (changes requested):
This looks great, I have some quibbles that are pretty minor. I know it's safe as-is but I really think we should always use parameterized queries for knex queries.

flodolo (approved):

esbuild.cronjobs.mjs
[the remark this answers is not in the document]
Vinnl: What I like about saying `20.12` is that it will show up on a project-wide search for that number, so that we won't forget to update it when updating the other instances. Using just `20` wouldn't be a problem for the coming future, but makes it easy to miss when we upgrade to Node 22.

As for the document/wiki/README, I'd actually like to write a small script that just checks all instances and errors out if they're not aligned, that we can run in CI.

src/db/tables/subscribers.js
Vinnl: Hmm, I think I did try to use that, but it got interpreted as a parameter - and then my PG UI didn't support it either. But you just gave me the push to verify that running it in `psql` does work, and then digging through old Knex.js issues to find out how to escape it: ef2a47c32

esbuild.cronjobs.mjs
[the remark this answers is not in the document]
Vinnl: Ah good point, I'd forgotten that we'd already set our repo to use ES modules. Changed in d23c56f59.

src/cronjobs/monthlyActivity.tsx
```diff
@@ -0,0 +1,104 @@
...
+import { getDashboardSummary } from "../app/functions/server/dashboard";
+import { getLatestOnerepScanResults } from "../db/tables/onerep_scans";
+import { getSubscriberBreaches } from "../app/functions/server/getSubscriberBreaches";
+import { getLocale } from "../app/functions/universal/getLocale";
+
+void run();
+
+async function run() {
+  const batchSizeEnvVar = Number.parseInt(
+    process.env.MONTHLY_ACTIVITY_EMAIL_BATCH_SIZE ?? "",
```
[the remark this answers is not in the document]
Vinnl: Hmm, so I was changing it to your suggestion, but then remembered why I didn't do that in the first place - I did still want to guard against passing in an invalid value for the env var, i.e. check `Number.isNaN`. But that does mean we'd ignore the env var value silently, so I'm now just throwing if it's wrong: 32e5b5406.

src/db/tables/subscribers.js
Vinnl: Oh right, I was actually going to ask you, since I copied this pattern [from](https://github.com/mozilla/blurts-server/commit/3bf2291258c100b1d8097ba5701276cb49e9533c#diff-97bf25d3506c6da45f409e35989ed5b2c1de1f9aed63ae4f26cbfea60c7c8e55R8) [you](https://github.com/mozilla/blurts-server/commit/e80edf72644811bbc8766d2023bcd914a50393bf#diff-836cc645d4daafe3034b9a12b669a35290f6006abf29e4565715a6593ff8ae54R32) :sweat_smile:

src/db/tables/subscribers.js
Vinnl: Oh right, good one - clarified the comment in 9c81d4737.

src/db/tables/subscribers.js
Vinnl: Oh yes absolutely we should, better safe than sorry: 39704eb9f

src/db/tables/subscribers.js
Vinnl: Yes exactly, and that gets set [_before_ we send the email](https://github.com/mozilla/blurts-server/blob/3f63a82c31aacc635b99b88dd10163ee00a543ae/src/cronjobs/monthlyActivity.tsx#L46-L48), so if an error occurs, the email won't get sent. Although given https://github.com/mozilla/blurts-server/pull/4399#discussion_r1556466715, I'll probably need to create a new query once that lands.

esbuild.cronjobs.mjs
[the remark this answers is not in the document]
Vinnl: https://github.com/mozilla/blurts-server/pull/4404

src/db/tables/subscribers.js
mansaj: Thinking through this more, what if we:
1) drop the columns (and data) of `monthly_email_optout` and `monthly_email_at`
2) add new cols `monthly_monitor_report: boolean` and `monthly_monitor_report_at:timestamp`

mansaj: Also I believe the data that exists should be inconsequential (very outdated, at least 6-12 months old)

Vinnl: Yeah that works for me. Wouldn't really change the logic here, just different column names, and a new query to update the timestamp when I send an email.

src/cronjobs/monthlyActivity.tsx
```diff
@@ -0,0 +1,108 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
```
mansaj, line 1: Let's use `src/scripts/cronjobs`

Vinnl: With apologies for rebasing, but I was afraid I wasn't going to be able to get all the path changes correct if I spread this over fixup commits. Fixed in f753a0f952a056c5a4ed2cf0c1611ece8e076bb4 and 93c298a1b29e9e135eb1d2d5014a7892d9d62890.

src/db/tables/subscribers.js
Vinnl: OK, [rebased](https://github.com/mozilla/blurts-server/pull/4400#discussion_r1576104457) on latest `main` that has the changes described above, and added e34b2e57b to make use of them.

rhelmer (approved):

src/db/tables/subscribers.js
```diff
@@ -306,6 +360,31 @@ async function updateMonthlyEmailOptout (token) {
...
+    .update({
+      // @ts-ignore knex.fn.now() results in it being set to a date,
+      // even if it's not typed as a JS date object:
+      monthly_monitor_report_at: knex.fn.now(),
+      // @ts-ignore knex.fn.now() results in it being set to a date,
+      // even if it's not typed as a JS date object:
+      updated_at: knex.fn.now(),
+    })
+    .where("primary_email", subscriber.primary_email)
+    .andWhere("id", subscriber.id)
```
mansaj, line 379: is it necessary to match both `primary_email` and `id`? Is there a case where these two won't match?

src/db/tables/subscribers.js
```diff
@@ -271,8 +272,60 @@ async function deleteResolutionsWithEmail (id, email) {
...
+
+  const rows = await query;
+
+  return rows
+}
+/* c8 ignore stop */
+
 /**
  * @param {string} email
+ * @deprecated Only used by the `send-email-to-unresolved-breach-subscribers.js`, which it looks like might not be sent anymore?
```
mansaj, line 328:
```suggestion
 * @deprecated Delete as a part of MNTOR-3077
```

src/db/tables/subscribers.js
```diff
@@ -296,6 +349,7 @@ async function updateMonthlyEmailTimestamp (email) {
  * Unsubscribe user from monthly unresolved breach emails
  *
  * @param {string} token User verification token
+ * @deprecated
```
mansaj, line 352:
```suggestion
 * @deprecated Delete as a part of MNTOR-3077
```

src/emails/getEmailL10n.node.ts
```diff
@@ -0,0 +1,118 @@
...
+ */
+export function getL10nBundles(userLocales: string): LocaleData[] {
+  const languages = userLocales ? acceptedLanguages(userLocales) : [];
+  const supportedLocales = process.env.SUPPORTED_LOCALES!.split(",");
+  // const availableLocales = Object.keys(bundleSources);
+  // const filteredLocales =
+  //   process.env.APP_ENV === "heroku"
+  //     ? availableLocales
+  //     : availableLocales.filter((locale) => supportedLocales?.includes(locale));
+  // const currentLocales = negotiateLanguages(languages, filteredLocales, {
```
mansaj, line 48:
```suggestion

```

mansaj (approved):

src/db/tables/subscribers.js
Vinnl: This is me being overly cautious after our security issues where people were able to have us email other people by adding their email address - by matching on ID in addition to email address, if we're somehow looking at someone else with the same email address, it won't match. But matching on just the ID should be sufficient, of course. There's no practical downside to matching on both though, is that correct?

src/emails/getEmailL10n.node.ts
Vinnl: I removed the whole file in https://github.com/mozilla/blurts-server/pull/4401, but might as well fix it for a cleaner history :) 49c09b4a2

src/db/tables/subscribers.js
Vinnl: 61ce72686

src/db/tables/subscribers.js
```diff
@@ -271,8 +272,60 @@ async function deleteResolutionsWithEmail (id, email) {
...
+ * @returns {Promise<import("knex/types/tables").SubscriberRow[]>}
+ */
+// Not covered by tests; mostly side-effects. See test-coverage.md#mock-heavy
+/* c8 ignore start */
+async function getSubscribersWaitingForMonthlyEmail (options = {}) {
+  // I'm explicitly referencing the type here, so that these lines of code will
+  // show up as errors when we remove it from the flag list:
+  /** @type {import("./featureFlags.js").FeatureFlagName} */
+  const featureFlagName = "MonthlyActivityEmail";
+  const flag = await knex("feature_flags")
```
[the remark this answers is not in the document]
Vinnl: Oh that's a good point, added that check in a util function in 1c425f71f.

The main reason I initially shied away from the util function, is because I had to know both `is_enabled` and the `allow_list`, and thus was already replicating feature flag-specific logic. I've now created a function that does a check for the `deleted_at`, but otherwise just returns the full row. We'll probably replace it by experiments later on anyway.

PR #4402, opened 2024-04-08
MNTOR-2931 - initialize Glean after logging setup, for auto pageload …

rhelmer:
…events

# References: 
Jira: MNTOR-2931

# Description

@rosahbruno debugged why we weren't seeing the auto pageload events being logged, and determined that it was from calling `initialize` before the debug and logging functions.

# How to test

Check browser devtools console for auto pageload events


Vinnl (approved):
AIUI this change would only affect our local and Heroku environments, right? On stage and prod this wouldn't have any effect (and thus those should already work correctly).

rhelmer:
> AIUI this change would only affect our local and Heroku environments, right? On stage and prod this wouldn't have any effect (and thus those should already work correctly).

Yes that's right, there's just no way to confirm that this is working locally or on stage.

A newish feature of Glean is that you can change these settings in browser devtools with the `window.Glean` interface, but I guess that won't work right for this auto page load measurement, I'll file a bug for the Glean team about that since other "auto" measurements should be coming in the future.

PR #4404, opened 2024-04-09
Node version consistency

Vinnl:
Followup to https://github.com/mozilla/blurts-server/pull/4400#discussion_r1556418422. It's not pretty, and it won't keep our dependencies up-to-date for us, but it will ensure that we don't forget to update any references to the Node version when Dependabot updates the Node version in our `Dockerfile`. This way, everything is aligned with what will be production.

### DIFF
--- a/.github/workflows/lint.yaml
+++ b/.github/workflows/lint.yaml
@@ -12,7 +12,7 @@ jobs:
       - name: Use Node.js
         uses: actions/setup-node@v4
         with:
-          node-version: '20.11.x'
+          node-version: '20.12.x'
       - run: npm ci
       - run: npm run build-glean
       - run: npm run build-nimbus
--- a/.github/workflows/lint.yaml
+++ b/.github/workflows/lint.yaml
@@ -19,3 +19,4 @@ jobs:
       - run: cp .env-dist .env
       # Mirror old linter from CircleCI, verifies that linter succeeds
       - run: npm run lint
+      - run: node scripts/check-node-version-alignment.js
--- a/scripts/check-node-version-alignment.js
+++ b/scripts/check-node-version-alignment.js
@@ -1,0 +1,79 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import { readFile, readdir } from "node:fs/promises";
+import { resolve, dirname } from "node:path";
+import { fileURLToPath } from "node:url";
+
+/**
+ * This script checks whether every place we set a Node version is set to the same version.
+ *
+ * I've opted to avoid adding e.g. TOML and YAML parsers, and just assume
+ * specific line formatting, to write the simplest thing that works in the least
+ * amount of time.
+ *
+ * It is run in `.github/workflows/lint.yaml`.
+ */
+
+const __dirname = dirname(fileURLToPath(import.meta.url));
+const load = (path) => readFile(resolve(__dirname, "../", path), "utf-8");
+
+const packageJson = JSON.parse(await load("package.json"));
+const lockFile = JSON.parse(await load("package-lock.json"));
+const dockerFileNodeVersion = (await load("Dockerfile"))
+  .split("\n")
+  .find((line) => line.startsWith("FROM node:"))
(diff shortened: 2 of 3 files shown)


.github/workflows/lint.yaml
```diff
@@ -12,9 +12,10 @@ jobs:
...
         uses: actions/setup-node@v4
         with:
-          node-version: '20.11.x'
+          node-version: '20.12.x'
       - run: npm ci
       - run: npm run build-glean
       - run: cp .env-dist .env
       # Mirror old linter from CircleCI, verifies that linter succeeds
       - run: npm run lint
+      - run: node scripts/check-node-version-alignment.js
```
rhelmer, line 21: If you want to use typescript for small one-off scripts like this, you could consider the `tsx` package. Or if you want to avoid dependencies, something like:

```sh
tsc --outDir dist/scripts scripts/check-node-version-alignment.ts && node dist/scripts/check-node-version-alignment.js
```

I'm assuming you don't want to have separate build/run steps in this instance.

scripts/check-node-version-alignment.js
```diff
@@ -0,0 +1,79 @@
...
+import { readFile, readdir } from "node:fs/promises";
+import { resolve, dirname } from "node:path";
+import { fileURLToPath } from "node:url";
+
+/**
+ * This script checks whether every place we set a Node version is set to the same version.
+ *
+ * I've opted to avoid adding e.g. TOML and YAML parsers, and just assume
+ * specific line formatting, to write the simplest thing that works in the least
+ * amount of time.
```
rhelmer, line 14: Do you want to file a followup in the backlog to add these? I agree that it's less urgent, might be a good first bug for someone if you think it's valuable (in which case I'd file it via GitHub which will mirror to Jira)

scripts/check-node-version-alignment.js
```diff
@@ -0,0 +1,79 @@
...
+const checkVersion = (source, version) => {
+  const voltaVersion = packageJson.volta.node;
+  if (getMinorOnly(voltaVersion) !== getMinorOnly(version)) {
+    console.error(`Error: version [${version}] in ${source} does not match the Volta version ([${voltaVersion}]).`)
+    process.exit(1);
+  }
+};
+
+checkVersion("the package.json engines", packageJson.engines.node);
+checkVersion("the package-lock.json engines", lockFile.packages[""].engines.node);
```
rhelmer, line 73: Did you mean to check the `[...].engines.npm` version here too? I noticed that it's checked above in `.netlify.toml`

rhelmer (changes requested):
lgtm and works great, I commented out the `esbuild.cronjobs.json` bit (since that hasn't landed yet).

I just have one question about the `packageJson.engines.npm` version, and a few suggestions (but you can take or leave those).

.github/workflows/lint.yaml
Vinnl: Yeah, one advantage of adding it in https://github.com/mozilla/blurts-server/pull/4400 is that the barrier to using it isn't that high when it's already in our dependencies. In this case I don't think it would add that much though, since we'd mostly be manually adding types for things that are only checked at runtime (because JSON, YAML, TOML and Docker files don't have type definitions).

scripts/check-node-version-alignment.js
Vinnl: I'd suggest to maybe let's first see whether the lack of them is actually problematic. If not, the extra dependencies might not be worth it.

scripts/check-node-version-alignment.js
Vinnl: Currently the script is only checking whether the Node versions are aligned everywhere, as I'm not really sure what to do with the NPM versions. I think there's something to be said for not specifying it and going with the version bundled with whatever Node version we're using, but there's also something to be said for disconnecting the two and choosing an explicit npm version ourselves (although I don't think Dependabot can check any of the places we can set the npm version). What do you think?

rhelmer: I don't know how much of a problem it is, but I do see complaints in our CI logs if the NPM version doesn't match. I don't think it's urgent though we can add it later, just having the script to check npm version will be a huge improvement, thanks for doing that.

rhelmer (approved):

flodolo (approved):

PR #4407, opened 2024-04-10
Add last scan date below upsell button

Vinnl:
# References: 
Jira: MNTOR-2701
Figma: https://www.figma.com/file/DTbmXzCQCw2PxXpHQW8yfW/Monitor-MVP-Enhancements?type=design&node-id=24-538&mode=design

# Description

Lets a user know when their last scan took place.

Note that I also moved `UpsellBadge.tsx` into the `toolbar` folder, since it's shown in the toolbar.

# Screenshot (if applicable)

image

image

# How to test

Open the dashboard with a user who has performed a scan.

- [ ] I've added a unit test to test for potential regressions of this bug. - No real business logic added.


Vinnl:
Putting back in draft mode - it needs to be put behind a feature flag.

src/db/tables/onerep_scans.ts
```diff
@@ -40,21 +40,26 @@ async function getScanResults(
...
-    };
+    return null;
   }
 
   const scan = await knex("onerep_scans")
     .first()
     .where("onerep_profile_id", onerepProfileId)
     .orderBy("created_at", "desc");
 
+  return scan ?? null;
```
pdehaan, line 55: Interesting! What does this do?
I thought the nullish coalescing would only return the right-hand side if the left-hand side was already null.

Vinnl: Nullish coalescing returns the right-hand side if the left-hand side is `null` _or `undefined`_. So this ensures we return `null` if there was no result.

Vinnl:
@rhelmer I was going to do that as part of setting up the experiment in the first place, in a separate PR, since that has its own ticket (and is probably a fair bit of (review) work by itself).

rhelmer (approved):

Vinnl:
> LGTM but not sure if we're still rolling this out since I think we're looking to redesign it?

I'm not actually sure if we made a decision on that, but we're going to run an experiment for it, so we'll first have to set that up anyway. Possibly we'll move it over a bit after that.

PR #4431, opened 2024-04-22
add new plan_tier key on all glean events

rhelmer:
# References: 
Jira: MNTOR-3116

# Description

Add `plan_tier` with value of `Free` or `Plus` to all  Glean events.

# How to test

You can see your pings on https://debug-ping-preview.firebaseapp.com/pings/fx-monitor-local-dev or in the browser devtools console.

Use the http://localhost/admin/dev UI to add/remove Plus, and ensure that pings have the correct `plan_tier`.


Vinnl (approved):
I suppose this works, but it's kludgy :/ I take it Glean doesn't have something similar to [GA's user properties](https://support.google.com/analytics/answer/9355671?hl=en), where you set it once and it's associated with the session?

rhelmer:
> I suppose this works, but it's kludgy :/ I take it Glean doesn't have something similar to [GA's user properties](https://support.google.com/analytics/answer/9355671?hl=en), where you set it once and it's associated with the session?

Unfortunately not yet, I think it's planned though.

mansaj (approved):
lgtm

rhelmer:
Thanks - I will open a data review bug and fix the unit tests, going to keep this open until the data review is done.

rhelmer:
DS asked if we could also add these to cta button events, which is doable but more work since we need to drill down the subscription state to a bunch of components that don't currently have that... really the right way to do this would be to have a "Session" event that has all of this info such as locale, subscription state, and so on. Also, we are working on setting up backend events, which will be much more trustworthy subscription events than getting it from the client.

So for now I'll go ahead and set this up, just writing this down since I don't believe it's sustainable to have to pass so many properties to every little component so they can record it :) This is just due to limitations in our analysis environment right now.

rhelmer:
> DS asked if we could also add these to cta button events, which is doable but more work since we need to drill down the subscription state to a bunch of components that don't currently have that... really the right way to do this would be to have a "Session" event that has all of this info such as locale, subscription state, and so on. Also, we are working on setting up backend events, which will be much more trustworthy subscription events than getting it from the client.
> 
> So for now I'll go ahead and set this up, just writing this down since I don't believe it's sustainable to have to pass so many properties to every little component so they can record it :) This is just due to limitations in our analysis environment right now.

Oh and as alluded to above, my understanding is that Glean.js has or will soon have a concept of a "session" that we can hopefully use soon. I'll follow up on that.

src/app/(proper_react)/(redesign)/(authenticated)/user/(dashboard)/dashboard/fix/leaked-passwords/LeakedPasswordsLayout.tsx
```diff
@@ -35,6 +35,7 @@ export interface LeakedPasswordsLayoutProps {
   subscriberEmails: string[];
   data: StepDeterminationData;
   isEligibleForPremium: boolean;
+  isPremium: boolean;
```
pdehaan, line 38: Should this be `isPlus` or is Premium better since we are already committed to `isEligibleForPremium` on previous line?

src/app/(proper_react)/(redesign)/(authenticated)/user/(dashboard)/dashboard/fix/leaked-passwords/LeakedPasswordsLayout.tsx
```diff
@@ -174,8 +175,9 @@ export function LeakedPasswordsLayout(props: LeakedPasswordsLayoutProps) {
           ? "password_exposed"
           : "security_question_exposed",
       utm_source: "guided_experience",
+      plan_tier: props.isPremium ? "Plus" : "Free",
```
pdehaan, line 178: Should the `"Plus"` and `"Free"` be in an enum somewhere?
Looks like Ctrl+F of "Plus" gives 6 hardcoded results and "Free" about 9.

rhelmer:
We don't need data review for this per data team (more detail in Jira).

src/app/(proper_react)/(redesign)/(authenticated)/user/(dashboard)/dashboard/fix/leaked-passwords/LeakedPasswordsLayout.tsx
rhelmer: Thanks, enum is a good idea.

rhelmer:
Per recent check-in with DS, they'd like this on `cta_button` as well. I've added to `metrics.yaml`, will take a little more time to instrument this (these components don't tend to have the subscription state already).

src/app/(proper_react)/(redesign)/(authenticated)/user/(dashboard)/dashboard/fix/leaked-passwords/LeakedPasswordsLayout.tsx
rhelmer: I wish that we could define this in Glean and reference it there... AFAICT from the docs, keys on `Event`s are always `string` types so I don't think there's a great way to specify it.

rhelmer: I think this is a good idea, but the way we need to pass down props to all of these components is temporary and I don't expect to keep this forever. I'd rather not have to import the type just for this in all of these. It might make sense to use a function instead of a boolean or something, but I think passing simpler prop types than function is generally preferred. I'd like to keep it as-is for now

src/app/(proper_react)/(redesign)/(authenticated)/user/(dashboard)/dashboard/fix/leaked-passwords/LeakedPasswordsLayout.tsx
rhelmer: I made it `isPremiumUser` everywhere for consistency, we already have these in quite a few places so I think it'd be better to stay consistent for now (such as it is) and move everything from "Premium" to "Plus" together later.

rhelmer:
@Vinnl @flozia DS asked if we could include `has_tier` in `ctaButton.click` events as well as page view... it's much more invasive having to pass properties to everything. We should be able to stop using this approach, do you think this is OK for now?

Vinnl:
@rhelmer It might be easier to add a [context](https://react.dev/learn/passing-data-deeply-with-context) and then [use that in the `useGlean` hook](https://react.dev/reference/react/useContext)? Not sure pretty, but make it easier to remove later when Glean supports sessions?

rhelmer:
> @rhelmer It might be easier to add a [context](https://react.dev/learn/passing-data-deeply-with-context) and then [use that in the `useGlean` hook](https://react.dev/reference/react/useContext)? Not super pretty, but make it easier to remove later when Glean supports sessions?

Ah thanks, yes I think it might be appropriate in this situation.

src/app/hooks/useGlean.ts
```diff
@@ -62,6 +67,11 @@ export const useGlean = (experimentationId?: string) => {
     // Instead of the specific type definitions we generated in the npm script
     // `build-glean-types`, Glean takes a non-specific "ExtraArgs" type as
     // parameter to `record`.
+
+    // Record the `plan_tier` key on all events.
+    // @ts-expect-error Not present in map, but exists on all events.
```
rhelmer, line 72: @Vinnl I see the `plan_tier` key in the generated types, but I'm not sure how to use convince typescript that this is OK since it's not in the `GleanMetricMap`. Do you have any thoughts?

rhelmer:
> > @rhelmer It might be easier to add a [context](https://react.dev/learn/passing-data-deeply-with-context) and then [use that in the `useGlean` hook](https://react.dev/reference/react/useContext)? Not super pretty, but make it easier to remove later when Glean supports sessions?
> 
> Ah thanks, yes I think it might be appropriate in this situation.

As discussed in Slack, I think it makes sense in this case to have `useGlean` just get the session object and call `hasPremium()`. DS would prefer that we just include the `plan_tier` key on every event.

That makes this patch much smaller and less invasive, it should be easy to find and remove once Glean supports sessions more directly and we can just include plan info there.

src/app/hooks/useGlean.ts
```diff
@@ -62,6 +67,11 @@ export const useGlean = (experimentationId?: string) => {
...
+    data.plan_tier = isPremiumUser ? "Plus" : "Free";
```
Vinnl, line 73: I think it's just too much work for TypeScript to infer that `plan_tier` is present on every entry - it would have to create a type intersection for every event (e.g. `click`) event module (e.g. `button`), and _then_ create a type section for each of those.

It's probably better not to expect an error (because it's really not an error, and if they ever find a way to do the inference performantly, it would cause an error), but to just take one of the event module's type definitions and tell TS that that applies regardless of which event module's actually being set:

```suggestion
    // `plan_tier` is set on every metric, but it's too much work for TypeScript
    // to infer that — hence the type assertion.
    (data as GleanMetricMap["button"]["click"]).plan_tier = isPremiumUser
      ? "Plus"
      : "Free";
```

It won't warn us if we ever add an event without adding `plan_tier`, but :shrug:

Vinnl (approved):

PR #4438, opened 2024-04-25
Mntor 2182/add nimbus feature rollout

rhelmer:
# References: 
Jira: MNTOR-2182

# Description

- Add Nimbus support for enabling Monitor feature flags
  - replace fxa/telemetry ID with Glean-supported `experimentationId`
- Pare down the feature flag admin UI
  - only expose add, delete, and enable
  - adding a flag completely overrides Nimbus, and removing puts Nimbus back in control

# How to test

See the bottom of `getExperiments.ts`,  it returns an example `features` result and logs the targeting info, if the server is run in dev mode.


Vinnl:
@rhelmer How would you feel about us removing the feature flag UI and infra altogether? Instead, to test locally, you could modify the `default` values in `nimbus.yaml`, and on stage and production, change the rollout in Experimenter to 0% or 100% as desired. That would ensure we have a single source of truth, and save us a bunch of code. I've worked that out in https://github.com/mozilla/blurts-server/pull/4443, using many of the changes from this PR.

rhelmer:
> @rhelmer How would you feel about us removing the feature flag UI and infra altogether? Instead, to test locally, you could modify the `default` values in `nimbus.yaml`, and on stage and production, change the rollout in Experimenter to 0% or 100% as desired. That would ensure we have a single source of truth, and save us a bunch of code. I've worked that out in #4443, using many of the changes from this PR.

OK sorry I missed this comment, yes I think we could do this. Ideally I'd like to have a `local` environment in the nimbus config, I worry about changing defaults in local files while testing, these are too easy to accidentally commit.

rhelmer:
> > @rhelmer How would you feel about us removing the feature flag UI and infra altogether? Instead, to test locally, you could modify the `default` values in `nimbus.yaml`, and on stage and production, change the rollout in Experimenter to 0% or 100% as desired. That would ensure we have a single source of truth, and save us a bunch of code. I've worked that out in #4443, using many of the changes from this PR.
> 
> OK sorry I missed this comment, yes I think we could do this. Ideally I'd like to have a `local` environment in the nimbus config, I worry about changing defaults in local files while testing, these are too easy to accidentally commit.

@Vinnl I went ahead and changed the remit of this PR, it's now about deprecating feature flags. I think we need to do it in stages though since we have features shipped right now and also in review that are behind flags, let's land the cleanup and simplification of the UI (plus the delete button), and I can mark the code as deprecated. We already have a tech debt ticket to remove feature flags, so we can finalize it and remove all flags from the DB.

After that's all done, we can remove the feature flag code. We might want to leave the `feature_flags` table as it has a record of what was shipped and when.

rhelmer:
Changing the purpose of this PR slightly, going to wait for https://github.com/mozilla/blurts-server/pull/4443 to land and will rebase on top of that. This will simplify the feature flag UI and mark feature flags as deprecated, and some small related cleanup.

src/app/(proper_react)/(redesign)/(authenticated)/admin/feature-flags/page.tsx
```diff
@@ -114,9 +85,13 @@ export default async function FeatureFlagPage() {
...
       <div className={styles.start}>
-        <h1>All Feature Flags</h1>
+        <h1>Flags added here will override Nimbus features.</h1>
+        <h2>Deleted flags will be controllable by Nimbus.</h2>
+        <br />
+        <h3>All Feature Flags</h3>
         <AllFlagsTable data={featureFlags} />
-        <h1>Add New Feature Flag</h1>
+        <br />
+        <h3>Add New Feature Flag</h3>
```
mansaj, line 94: Since we are changing the UI, if it's not too much work, it might be useful to add a table below to show the `deleted/archived feature flags`, since they are soft deleted

src/db/tables/featureFlags.ts
```diff
@@ -88,7 +110,11 @@ async function addFeatureFlag(flag: FeatureFlag) {
     expired_at: flag.expiredAt,
     owner: flag.owner,
   };
-  const res = await knex("feature_flags").insert(featureFlagDb).returning("*");
+  const res = await knex("feature_flags")
```
mansaj, line 113: ok just to clarify this query:
It attempts to upsert the `featureFlagDb` object, when the name existed, it updates the `deleted_at`

And the assumption here is that the `deleted_at` would be updated to null because `featureFlagDb.deleted_at` would be null? This essentially resets the feature flag from soft delete -> no longer deleted?

rhelmer: Right exactly

src/app/(proper_react)/(redesign)/(authenticated)/admin/feature-flags/page.tsx
rhelmer: Done

Vinnl:
@rhelmer I landed https://github.com/mozilla/blurts-server/pull/4443, let me know when you've rebased :)

rhelmer:
> @rhelmer I landed #4443, let me know when you've rebased :)

It's rebased!

src/app/(proper_react)/(redesign)/(authenticated)/admin/feature-flags/components/DeleteFeatureFlag.tsx
```diff
@@ -0,0 +1,40 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+"use client";
+
+import { useRouter } from "next/navigation";
+
+export const DeleteFeatureFlag = (props: { name: string | undefined }) => {
```
Vinnl, line 9: Nit, but I assume this should always be defined?

```suggestion
export const DeleteFeatureFlag = (props: { name: string }) => {
```

src/app/(proper_react)/(redesign)/(authenticated)/admin/feature-flags/page.tsx
```diff
@@ -99,7 +73,35 @@ export default async function FeatureFlagPage() {
...
+              <td>{item.deleted_at?.toString()}</td>
+            </tr>
+          ))}
+        </tbody>
+      </table>
+    );
+  };
+
   const featureFlags = (await getAllFeatureFlags()) || null;
+  const deletedFeatureFlags = (await getDeletedFeatureFlags()) || null;
```
Vinnl, line 104: Nit, but a good habit to build. Since these aren't returning a boolean:

```suggestion
  const featureFlags = (await getAllFeatureFlags()) ?? null;
  const deletedFeatureFlags = (await getDeletedFeatureFlags()) ?? null;
```

src/db/tables/featureFlags.ts
```diff
@@ -8,6 +8,7 @@ import { FeatureFlagRow } from "knex/types/tables";
 
 const knex = createDbConnection();
 
+/** @deprecated This type should not be used */
```
Vinnl, line 11: Maybe point to the replacement (i.e. experiments)?

Vinnl (approved):
Really nice to have the deletion now too, makes it a bit cleaner. I'd maybe check with QA that they're not using the allowlist still though, just in case we're blocking their work for existing feature-flagged features.

rhelmer:
> Really nice to have the deletion now too, makes it a bit cleaner. I'd maybe check with QA that they're not using the allowlist still though, just in case we're blocking their work for existing feature-flagged features.

I don't think we need the allow list for any features that are in-flight or planned, we needed it for Plus but in general I think we can just enable everything on stage. I'll add it back if I'm wrong, I just removed from UI but left the API and DB parts alone :)

PR #4439, opened 2024-04-25
Add script to generate TypeScript types for the response of the Nimbus experiments API

Vinnl:
# References: 
Jira: MNTOR-3094

This adds a script (`src/scripts/build/nimbusTypes.ts`) that reads the `nimbus.yaml` file, and generates a TypeScript file containing strict type definitions for the response data from the `v1/features` API, so that we can pass that on to the rest of the app to determine which features to enable for the current user.

It also generates an object matching that type definition, set to the default values - this makes it easier to test locally, where the Nimbus sidecar isn't running, as well as being helpful when there is an error calling the API.

(It also moves existing build scripts into `/src/scripts/build`.)

This is what it generates for the current `nimbus.yaml`:

```ts

// AUTOGENERATED src/scripts/build/nimbusTypes.js. DO NOT EDIT. DO NOT COMMIT.

/* Nimbus string aliases */


/* Nimbus object types */


/* Nimbus enum types */


/** Status of experiments, as setup in Experimenter */
export type ExperimentData = {
  "example-feature": {
    "enabled": boolean;
    "something": null | string;
  };
};

export const defaultExperimentData: ExperimentData = {
  "example-feature": {
    "enabled": false,
    "something": null,
  },
};
```


config/nimbus.yaml
```diff
@@ -21,6 +21,3 @@ features:
       - channel: production
         value: { "something": "wicked" }
 
-types:
-  objects: {}
-  enums: {}
```
Vinnl, line 26: @rhelmer The docs show a different syntax [for enums](https://experimenter.info/fml-spec/#enumerations) and [for objects](https://experimenter.info/fml-spec/#object-types). Do you know which of these is outdated? I followed the docs for now.

package.json
```diff
@@ -129,6 +130,7 @@
     "stylelint-config-recommended-scss": "^14.0.0",
     "stylelint-scss": "^6.2.1",
     "ts-jest": "^29.1.2",
-    "typescript": "^5.4.2"
+    "typescript": "^5.4.2",
+    "yaml": "^2.4.1"
```
Vinnl, line 134: This is another dependency at build time, but not only is it widely used, it's [by our colleague Eemeli](https://github.com/eemeli/yaml).

config/nimbus.yaml
rhelmer: I suspect the docs are right, thanks.

rhelmer (approved):

PR #4460, opened 2024-04-29
More strictly match version numbers in regexp

Vinnl:
Currently, `npm ls @playwright/test` outputs the following:

```
├── @playwright/test@1.43.1
└─┬ next@14.2.3
  └── @playwright/test@1.43.1 deduped
```

Since our previous regular expression removed everything that preceded an `@`, the ` deduped` is left in place, and causes an error (see [this CI run](https://github.com/mozilla/blurts-server/actions/runs/8878070008/job/24374057900)). Thus, this new regexps removes everything that doesn't strictly match a normal version number.

It also only looks at the first one - I'm not actually sure why Next is pulling it in too, since I don't see it listed [here](https://github.com/vercel/next.js/blob/canary/packages/next/package.json), but it might've been a temporary bug.


mansaj (approved):

.github/workflows/e2e_cron.yml
```diff
@@ -29,7 +29,9 @@ jobs:
         run: cp .env-dist .env
       - name: Store Playwright's Version
         run: |
-          PLAYWRIGHT_VERSION=$(npm ls @playwright/test | grep @playwright | sed 's/.*@//')
+          # The regular expression matches everything containing
+          # `<number>.<number>.<number>`, and returns only that bit:
+          PLAYWRIGHT_VERSION=$(npm ls @playwright/test | grep @playwright | sed --regexp-extended 's/.*([0-9]+\.[0-9]+\.[0-9]+).*/\1/') | head -1
```
pdehaan, line 34: Q: why not something like <kbd>npx playwright --version</kbd>?

```sh
npx playwright --version
# Version 1.43.1

npm ls @playwright/test | grep @playwright | sed -E 's/.*([0-9]+\.[0-9]+\.[0-9]+).*/\1/' | head -1
# 1.43.1
```

Vinnl: Heh, that's a good question :sweat_smile: It [was originally](https://github.com/mozilla/blurts-server/commit/25dd268c0ef0529be2b831450be5bcb02a41bdba) `npm view @playwright/test version`. The commit message doesn't say much about why it was changed, but maybe @mansaj knows?

pdehaan: I think <kbd>npm view</kbd> might be a bit risky. If I run <kbd>npm outdated</kbd> and filter the very randomly selected "pg" dependency, it says our current version is 8.11.3, but npm latest is 8.11.5.
If I query <kbd>npm info pg version</kbd> it's returning _npm's_ latest version number (8.11.5), not what we have installed locally (8.11.3).

```
npm outdated | grep -E "(Package|pg)"
Package                           Current  Wanted  Latest  Location                                       Depended by
pg                                 8.11.3  8.11.5  8.11.5  node_modules/pg                                blurts-server

npm info pg version
8.11.5
```

PR #4462, opened 2024-04-30
Do not import feature flag module from subscribers

Vinnl:
The featureFlags module is written in TypeScript, and imports logger. Both our email tests and plain-JS cron jobs stumble over this, and while the tests could be fixed by simply mocking out the unused problematic code, setting up a build pipeline for our cron jobs is a larger piece of work that we've only just started. Since we're also going to be phasing out feature flags in general (in favour of experiments), inlining it seems like an acceptable compromise in the short term.


mansaj (approved):
Thanks for getting to this!

PR #4463, opened 2024-04-30
Always build experiment data before npm run dev

Vinnl:
This won't re-build it if `nimbus.yaml` changes, but at least it won't break if you haven't run it before, or if it's changed since the last time you started the dev server.

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,7 +8,7 @@
   },
   "type": "module",
   "scripts": {
-    "dev": "next dev --port=6060",
+    "dev": "npm run build-nimbus && next dev --port=6060",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -8,7 +8,7 @@
   },
   "type": "module",
   "scripts": {
-    "dev": "next dev --port=6060",
+    "dev": "npm run build-nimbus && next dev --port=6060",
```
Vinnl, line 11: I could also run `npm run build-glean` in here, but I think that might take a lot longer the first time? On the other hand, you need to do it anyway I think?

rhelmer (approved):

PR #4472, opened 2024-05-03
Update Fluent linter, flag combination of Mozilla term + account

flodolo:
This should prevent strings from using a combination of Mozilla term + "account", and use the correct term.


Vinnl (approved):
Thanks flod, good catch and neat fix!

PR #4473, opened 2024-05-03
Add npm scripts for remaining cron jobs

Vinnl:
This will make it easier to move these scripts or add a build pipeline (e.g. to support TypeScript).

This is a followup to https://github.com/mozilla/blurts-server/pull/4469 - I noticed there were a couple of other cron jobs.


package.json
```diff
@@ -21,6 +21,10 @@
     "e2e:smoke": "playwright test src/e2e/ --grep @smoke",
     "cron:monthly-activity": "node dist/scripts/cronjobs/monthlyActivity.js",
     "cron:breach-alerts": "node src/scripts/emailBreachAlerts.js",
+    "cron:db-delete-unverified-subscribers": "node scripts/delete-unverified-subscribers.js",
+    "cron:db-pull-breaches": "node src/scripts/syncBreaches.js",
+    "cron:remote-settings-pull-breaches": "node scripts/updatebreaches.js",
```
Vinnl, line 26: Does this one actually run? It's [referencing](https://github.com/mozilla/blurts-server/blob/7725b368fcf9c04b82ae44686495a109841e52e7/scripts/updatebreaches.js#L10) a file that doesn't exist, and I'm not sure if it's still relevant - does Firefox still show breaches anywhere? I don't see anything in `about:protections` anymore.

package.json
```diff
@@ -21,6 +21,10 @@
...
+    "cron:onerep-limits-alert": "node src/scripts/onerepStatsAlert.js",
```
pdehaan, line 27: nit: should the script name match the filename, or should the filename match the script name?
ie:
```js
"cron:onerep-stats-alert": "node src/scripts/onerepStatsAlert.js",
// vs
"cron:onerep-limits-alert": "node src/scripts/onerepLimitsAlert.js",
```

Or I see one script above is in a /cronjobs/ directory, although its in a dist/ folder, so maybe that is transpiled, and these scripts shouldn't move into a /cronjobs/ directory to make their intent a bit clearer (at the sake of complicating the PR).

Vinnl: Which cronjobs we run are configured elsewhere, and they're currently pointing to these script paths directly. After this PR, I'm hoping to point them to these npm scripts, and then we can update the paths to the other cronjobs with independent PRs in just this repo - both to move them into a `/cronjobs/` directory, as well as to enable TypeScript (and thus compilation).

package.json
Vinnl: OK, so from Slack it sounds like it's disabled for now, even though the script exists. I'll leave it in here since the script and the reference to it still exist, so the disabling might not necessarily have been (intended to be) permanent yet.

rhelmer (approved):

PR #4477, opened 2024-05-03
Make the fixed dashboard tab the default for Plus users (MNTOR-3087)

flozia:
# References: 
Jira: [MNTOR-3087](https://mozilla-hub.atlassian.net/browse/MNTOR-3087)

# Description

Plus users should be taken to the ”fixed” dashboard tab by default. The only exception is when a user enters from a breach alert email, in which case, they are taken to the “action needed” tab.

## TODO
- [ ] The activeTab prop in the dashboard story is [set to “action-needed”](https://github.com/mozilla/blurts-server/pull/4477/files#diff-623cc418176995db1fc20296aa326a56e0b20da2c77f009d9b6b8c6905ae57a6R212) for now: If this approach looks good to us we’ll need to update the dashboard story and unit test to reflect the updates in this PR. 
- [x] Update the link target to the dashboard in the breach alert email

[MNTOR-3087]: https://mozilla-hub.atlassian.net/browse/MNTOR-3087?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Vinnl (commented):
I'm not sure if I'm a fan of adding two more URLs for the dashboard, given that design is looking at consolidating the tabs into one (not sure if that was already known to you - I heard it at the work week I believe).

Even if we're adding the two URLs, I think I'd rather add them explicitly, rather than just intercepting everything past `/user/dashboard/...` - I think they could just import the same component? We'd also need to add redirects from the old URL to one of the new ones, to not break bookmarks and old links. And for client-side path manipulation, it's probably safer to use the [router](https://nextjs.org/docs/app/api-reference/functions/use-router) than directly manipulating `window.history`, to avoid breaking assumptions Next.js might be making that could lead to mismatched state.

I'd actually have expected this change to be limited to an `isPremium` check in the `useState` parameter for `selectedTab` / `activeTab`, possibly with a check for a hash fragment or query parameter to directly load a specific tab, for use by emails. Any problems with that that I'm missing?

flozia:
> As a premium user should I stay on the fixed tab by default? I'm still getting take to `Action needed` on refresh?
@codemist Staying on the “fixed” tab is what _should_ happen. There was an [e921dec](https://github.com/mozilla/blurts-server/pull/4477/commits/e921dec02c75a847ef44d040daacc44edee43b50) that caused an error that might be related — is this still happening for you?

flozia:
> I'm not sure if I'm a fan of adding two more URLs for the dashboard, given that design is looking at consolidating the tabs into one (not sure if that was already known to you - I heard it at the work week I believe).

@Vinnl I was not aware of the plan to consolidate the tabs into one. I’ll try to address your comment and give a bit more context on why I arrived at this solution:

> Even if we're adding the two URLs, I think I'd rather add them explicitly, rather than just intercepting everything past `/user/dashboard/...` - I think they could just import the same component? We'd also need to add redirects from the old URL to one of the new ones, to not break bookmarks and old links.

The idea behind using the [optional catch-all segments](https://nextjs.org/docs/pages/building-your-application/routing/dynamic-routes#optional-catch-all-segments) is that this approach would not break bookmarks and old links. If no `slug` is provided we would redirect to the respective [default route](https://github.com/mozilla/blurts-server/pull/4477/files#diff-4e0322e3805222940d6c3e82cabf1b35cf68da991d7bb14d24fe20ea8541e5d6R65) for the user.

> And for client-side path manipulation, it's probably safer to use the [router](https://nextjs.org/docs/app/api-reference/functions/use-router) than directly manipulating `window.history`, to avoid breaking assumptions Next.js might be making that could lead to mismatched state.

Using the router leads to a re-render of the page on the server as well. Since we don’t need a full re-render the idea behind updating the URL is to only reflect the client state changes. Shallow routing is not supported by Next.js 13 as it was for Next.js 12. The [recommended approach](https://github.com/vercel/next.js/discussions/48110#discussioncomment-7563979) at the moment is to use the native history API for that.

> I'd actually have expected this change to be limited to an `isPremium` check in the `useState` parameter for `selectedTab` / `activeTab`, possibly with a check for a hash fragment or query parameter to directly load a specific tab, for use by emails. Any problems with that that I'm missing?

Initially, I was going to use URL search params for the tabs, but I remembered a past discussion around the potential unreliabilities of using query params for navigation and was hesitant about using them. The current approach would allow to directly link to a specific tab, but could be easily switched to just use `searchParams`, though.

src/app/(proper_react)/(redesign)/(authenticated)/user/(dashboard)/dashboard/View.tsx
```diff
@@ -84,6 +86,16 @@ export const View = (props: Props) => {
   const l10n = useL10n();
   const recordTelemetry = useTelemetry(props.experimentationId);
   const countryCode = useContext(CountryCodeContext);
+  const pathname = usePathname();
+
+  const [activeTab, setActiveTab] = useState<TabType>(props.activeTab);
+
+  useEffect(() => {
+    const nextPathname = `/user/dashboard/${activeTab}`;
+    if (pathname !== nextPathname) {
```
Vinnl, line 95:
```suggestion
    if (pathname !== nextPathname) {
      // Directly interacting with the history API is recommended by Next.js to avoid re-rendering on the server. See
      // https://github.com/vercel/next.js/discussions/48110#discussioncomment-7563979
```

Vinnl (approved):
Thanks for the explanation Florian, the reasoning seems solid to me.

src/app/(proper_react)/(redesign)/(authenticated)/user/(dashboard)/dashboard/View.tsx
flozia: Thanks, added the comment in [ae2d1a5](https://github.com/mozilla/blurts-server/pull/4477/commits/ae2d1a5b66d2472dc8069aa79711d567f7fd5b0d).

flozia:
Alright, @Vinnl! Since we decided to go with this approach I’ve addressed the two open to-dos:

1. Updated the link from the breach alert emails, as requested in the acceptance criteria, to always point to the “action needed” tab in [9c00730](https://github.com/mozilla/blurts-server/pull/4477/commits/9c00730d6ef6b708d87d7afa113c6bb52885149f).
2. Added unit test to verify that the dashboard is being rendered with the correct active tab in [22eccd4](https://github.com/mozilla/blurts-server/pull/4477/commits/22eccd48bf77bf659f8ef86ab184b343f107a98a).

src/app/(proper_react)/(redesign)/(authenticated)/user/(dashboard)/dashboard/Dashboard.test.tsx
```diff
@@ -425,6 +425,33 @@ it("counts how often people click the 'Contact us' link", async () => {
...
+  );
+  render(<ComposedDashboard activeTab="action-needed" />);
+
+  const tabActionNeededTrigger = screen.getByRole("tab", {
+    name: "Action needed",
+  });
+  expect(tabActionNeededTrigger.getAttribute("aria-selected")).toBe("true");
+});
+
+it("renders the dashboard with the fixed tab selected", () => {
```
Vinnl, line 441: Nit, but the test descriptions contradict each other otherwise :)

```suggestion
it("renders the dashboard with the action needed tab selected when requested", () => {
  const ComposedDashboard = composeStory(
    DashboardUsNoPremiumNoScanNoBreaches,
    Meta,
  );
  render(<ComposedDashboard activeTab="action-needed" />);

  const tabActionNeededTrigger = screen.getByRole("tab", {
    name: "Action needed",
  });
  expect(tabActionNeededTrigger.getAttribute("aria-selected")).toBe("true");
});

it("renders the dashboard with the fixed tab selected when requested", () => {
```

src/app/(proper_react)/(redesign)/(authenticated)/user/(dashboard)/dashboard/View.tsx
```diff
@@ -74,7 +74,7 @@ export type Props = {
   experimentationId: string;
   elapsedTimeInDaysSinceInitialScan?: number;
   totalNumberOfPerformedScans?: number;
-  activeTab: TabType;
+  activeTab: string;
```
Vinnl, line 77: No longer marking this as `TabType` was surprising - is that because we the path parameter in `page.tsx` could be any string? The safest solution IMHO would be to have an `isValidTabType` [user-defined type guard](https://www.typescriptlang.org/docs/handbook/advanced-types.html#user-defined-type-guards) (essentially, extracting [this if condition](https://github.com/mozilla/blurts-server/blob/c8a4edf1a5bbe238948cb6ca40af791fe2fd3b15/src/app/(proper_react)/(redesign)/(authenticated)/user/(dashboard)/dashboard/%5B%5B...slug%5D%5D/page.tsx#L61-L64) into a function), or otherwise at least to add a type assertion in there, rather than having the type information say that this component is fine with any string.

Vinnl (approved):
Ah right, I didn't look at the description again :) Thanks for the additional changes (and opportunity to review), they look good - and there are even new tests for the older non-TS email code! Just two nits that you can take or leave.

src/app/(proper_react)/(redesign)/(authenticated)/user/(dashboard)/dashboard/Dashboard.test.tsx
flozia: Updated in [4c80f52](https://github.com/mozilla/blurts-server/pull/4477/commits/4c80f52920d11d55fac806ad89bf098331c6c8ef).

src/app/(proper_react)/(redesign)/(authenticated)/user/(dashboard)/dashboard/View.tsx
flozia: I’ve added a type assertion in https://github.com/mozilla/blurts-server/pull/4477/commits/3b355f3ea16e5311659fda27d6daaf494ce34292.

PR #4512, opened 2024-05-08
Add loader if center value hasn't rendered

codemist:
# References: 
Jira: MNTOR-3154

# Description

# Screenshot (if applicable)

https://github.com/mozilla/blurts-server/assets/13066134/51987059-5187-4e87-bbc5-4114505e443a

# How to test


Vinnl:
I got nerd-sniped into trying to find out why the colours wouldn't show up, and found out that it's because our Content Security Policy doesn't allow defining styles via the `style` attribute. One potential solution is to widen our CSP policy, which I've submitted a PR for here: https://github.com/mozilla/blurts-server/pull/4531

I'm not sure I'm able to reproduce what the issue is with the center value, but possibly that PR also solves that?

src/app/components/client/Chart.tsx
```diff
@@ -201,10 +202,30 @@ export const DoughnutChart = (props: Props) => {
...
+
+  const nrText = document.querySelector(`.${styles.headingNr}`);
+  const labelText = document.querySelector(`.${styles.headingLabel}`);
+
+  // Ignoring this, otherwise tests will complain:
+  // > Warning: A suspended resource finished loading inside a test, but the
+  // > event was not wrapped in act(...).
+  // > When testing, code that resolves suspended data should be wrapped into
+  // > act(...)
+  /* c8 ignore start */
```
Vinnl, line 215: OK, so I think I figured out why the center value isn't available on page load: it is rendered by a `getFragment`, which replaces `<tags>` in a string with actual HTML tags. However, to do this safely (i.e. to avoid code injection from variables), it uses browser APIs to instantiate DOM elements for those actual tags - and those APIs aren't available during pre-rendering. This is why [`getFragment` only renders on the client-side](https://github.com/mozilla/blurts-server/blob/e5eab8590299ca554f26c80c6f6fe6bad3ed30c7/src/app/hooks/l10n.ts#L20-L22).

So the loading animation while we await the client-side render sounds like a good idea. However, it should be sufficient to use the `useHasRenderedClientSide` hook to determine whether to show it, rather than have the effect dependent on `props.data`, `nrText` and `labelText`. Also, we can add a comment describing the above to explain why we have it :)

And then with https://github.com/mozilla/blurts-server/pull/4551, you should be able to avoid the test from erroring by adding a `jest.mock` call for the `useHasRenderedClientSide` hook.

Vinnl (commented):
Thanks for picking this up, I think the addition of a loader makes sense. I did spend a bit of effort trying to understand what went wrong, which I've detailed inline - I think there's a small change we can make that then enables not suppressing the test error, but outright avoiding it.

And if you don't mind, could you add a story for the loader? Curious to see what the two different versions look like, which is a bit of a hassle to test now :)

(Oh, and apologies for the delayed review, I got a bit buried in messages.)

rhelmer:
@codemist are you still working on this one?

Vinnl:
I'm assuming no, and that this has stalled/been de-prioritised, so closing for now.

PR #4516, opened 2024-05-10
Show CSAT survey only to active Plus subscribers (MNTOR-3191)

flozia:
# References: 
Jira: [MNTOR-3191](https://mozilla-hub.atlassian.net/browse/MNTOR-3191)

# Description

The CSAT survey is not supposed to be shown to users with an expired Plus subscription.

[MNTOR-3191]: https://mozilla-hub.atlassian.net/browse/MNTOR-3191?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/app/(proper_react)/(redesign)/(authenticated)/user/(dashboard)/dashboard/View.tsx
```diff
@@ -189,7 +190,7 @@ export const View = (props: Props) => {
           isPremiumBrokerRemovalEnabled={props.enabledFeatureFlags.includes(
             "PremiumBrokerRemoval",
           )}
-          isPremiumUser={hasPremium(props.user)}
+          isPremiumUser={isPlusUser}
```
rhelmer, line 193: I do agree we should change this, but I am not sure we should change this to `plus` from `premium` piecemeal like this since it leads to confusing inconsistencies like this where we are setting `isPremiumUser={isPlusUser}`.

I would suggest filing a ticket in the backlog to replace all `premium` with `plus` in one go as it's own PR, and make this current one much more targeted.

src/app/(proper_react)/(redesign)/(authenticated)/user/(dashboard)/dashboard/View.tsx
```diff
@@ -404,6 +405,7 @@ export const View = (props: Props) => {
   };
 
   const showCsatSurvey =
+    isPlusUser &&
```
rhelmer, line 408: This is the actual needed change right? Per above I'd just continue calling it `premium` here just for now, and rename this holistically as a separate changeset.

flozia: Updated in [0322800](https://github.com/mozilla/blurts-server/pull/4516/commits/032280010ec31030dea3b29754b02eb102a4dcde).

Vinnl (approved):

PR #4533, opened 2024-05-13
MNTOR-2934/upload sentry sourcemaps for Docker deploy builds

rhelmer:
# References: 
Jira: MNTOR-2934

# Description

Always upload Sentry sourcemaps when performing Docker deploy builds.

This was restricted to only builds triggered by tag before, now that we've moved over to deploying from `main` and do the Docker push from GH actions, we can simplify things.


.github/workflows/docker_build_deploy.yml
```diff
@@ -31,6 +31,9 @@ jobs:
           echo "{\"commit\":\"$GITHUB_SHA\",\"version\":\"$GITHUB_REF\",\"source\":\"https://github.com/$GITHUB_REPOSITORY\",\"build\":\"$GITHUB_RUN_ID\"}" > version.json
 
       - name: Build Docker image
+        env:
+          UPLOAD_SENTRY_SOURCEMAPS: true
```
mansaj, line 35: Looks like we no longer need the env var `UPLOAD_SENTRY_SOURCEMAPS`?

I believe we can remove it from 
- `.env-dist`
- `next.config.js`

rhelmer: I think it still needs to be set as a variable, we need some way to get it into `next.config.js` right? I'm just setting it in the yaml vs. setting it in the GHA environment and then passing it through the yaml.

We only want it to run in this context, not on every PR and every time someone might run `npm run build` locally right?

mansaj: Ah hmm, I think I misunderstood what's going on here. NVM!

mansaj (approved):

.github/workflows/docker_build_deploy.yml
rhelmer: np, thanks!

PR #4538, opened 2024-05-15
Make alternate place names findable as individual results

Vinnl:
# References: 
Jira: MNTOR-3051
Figma: N/A

# Description

When constructing place names that people could search, we'd stuff alternate names to the end of them, e.g. `St Louis, MO, USA Saint Louis`. This meant that searching for `Saint Louis, USA` would not return a result. We could just move the alternate names forward, but then it'd still match on e.g. `St Saint`.

So in this PR, I've added those alternate entries, and then after finding results, map them back to the original location data. Since the alternate entries can now result in multiple matches for the same entry, I also deduplicate them.

If you look at all the changes of this PR together, it can be a bit hard to see what's changed, because I also moved the function into its own module (313796be7c6510cce42b0311d8bb75500dd4ce3d) so I could add unit tests (9acb80c119d1cbbb5646dc650a89f4d7082e7290). I also uncovered a bug where one search result would get removed (fixed and tested in 163d4358a07f49cc767a47d45e3fd972ccda055c). But the actual fix probably deserves the most scrutiny, for which **66b5c56fad9caafa648d53c23f69cbcc29342141 is the commit you'll want to look at**.

# How to test

Searching for `Saint Louis, MO, USA` should turn up a result.


src/app/api/v1/location-autocomplete/getRelevantLocations.ts
```diff
@@ -0,0 +1,105 @@
...
+    a?.forEach((alternateName) => {
+      nameIndexToLocationMap.set(
+        locationNamesAndAlternateNames.length,
+        location,
+      );
+      locationNamesAndAlternateNames.push(
+        `${alternateName} ${s} ${countryCode}`,
+      );
+    });
+  });
```
flozia, line 40: Great approach to the problem!

src/app/api/v1/location-autocomplete/getRelevantLocations.test.ts
flozia, line 1: Thanks for adding these tests.

flozia (approved):

PR #4544, opened 2024-05-16
MNTOR-3096 - add monitor glean backend

rhelmer:
# References: 
Jira: MNTOR-3096

# Description

# Screenshot (if applicable)

Not applicable.

# How to test


src/telemetry/backend-metrics.yaml
```diff
@@ -0,0 +1,218 @@
...
+        type: string
+      session_id:
+        description: An ID that allows us to track “sessions” of the user experience within the product.
+        type: string
+      monitor_user_id:
+        description: Mozilla Monitor-specific user ID.
+        type: string
+      plan_type:
+        description: Which type of plan the user is on ["1 month", "1 year"]
+        type: string
```
sean-rose, line 86: I thought you were going to start with just `plan_tier` for all metrics (like for the `page.view` metric above) since that's easy to get, and not have `plan_type` initially since that's harder to get (having to talk to the SubPlat API).

rhelmer: This was a mistake on my part, thanks. It's not too hard to talk to the subplat API, but I'd rather be consistent for now, we can add `plan_type` later.

src/telemetry/backend-metrics.yaml
```diff
@@ -0,0 +1,222 @@
...
+        description: Mozilla Monitor-specific user ID.
+        type: string
+      plan_tier:
+        description: Which tier of plan the user is on [Free, Plus]
+        type: string
+
+  cancel:
+    type: event
+    description: |
+      A subscription has been activated for this user.
```
sean-rose, line 91: This description should be revised to be accurate for the `subscription.cancel` event.

Also, "cancel" is a somewhat ambiguous term, as it's sometimes used to indicate a subscription's has been set to no longer auto-renew, or sometimes means that a subscription has been immediately deactivated/terminated.  Whichever it is in this case, it would be good to clarify in the description.

sean-rose (approved):
r+wc

In [bug 1896992](https://bugzilla.mozilla.org/show_bug.cgi?id=1896992#c7) I also asked @akkomar to take a look at this, since he has more expertise on Glean backend telemetry.

src/telemetry/backend-metrics.yaml
rhelmer: Good catch, thanks. I'm pretty sure we will only get this from the FxA webhook when the cancellation is finalized (subscription state changed). I'm not sure how we can be notified about pending cancellations, I know there is a SubPlat API we could poll, but it would be ideal if we could be notified of these via webhook as well. I'll follow up about that separately.

src/telemetry/backend-metrics.yaml
```diff
@@ -0,0 +1,222 @@
...
+# https://mozilla.github.io/glean/book/reference/yaml/metrics.html
+
+---
+$schema: moz://mozilla.org/schemas/glean/metrics/2-0-0
+
+page:
+  view:
+    type: event
+    description: |
+      A web page view event distinguishable by the page’s URL.
```
akkomar, line 15:
```suggestion
    description: |
      A web page view event distinguishable by the page’s URL.
    send_in_pings:
      - events
```
We need to declare which ping this is going to be sent in, otherwise [glean_parser will complain](https://github.com/mozilla/glean_parser/blob/f484585a93793ace324adf08ef9da7890c3d763d/glean_parser/javascript_server.py#L161-L169). Also applies to events below.

sean-rose: Good catch!

@rhelmer I think `send_in_pings` is the one remaining thing that needs to be fixed before merging this to unblock https://github.com/mozilla/probe-scraper/pull/757 (incorporating `glean_parser` into the build process could be done in a separate PR).

sean-rose (approved):
r+wc

The updated `subscription.cancel` event description looks good.  Just the one remaining `send_in_pings` issue that should be fixed.

rhelmer:
> Metrics definitions look good to me, apart from missing `send_in_pings` declaration.
> 
> We'll now want to integrate glean_parser in the build and generate logging code. We have some docs now at https://mozilla.github.io/glean/book/user/adding-glean-to-your-project/server.html#how-to-add-glean-server-side-collection-to-your-service and you can also compare [how FxA did this](https://github.com/mozilla/fxa/blob/257a254305d3a3d72497561f9ae24d141acbf1be/packages/fxa-shared/package.json#L220).

Thanks! I will review the docs and example, just FYI we split this part of the work to a separate ticket to fit better into our sprints.

src/telemetry/backend-metrics.yaml
```diff
@@ -0,0 +1,236 @@
...
+      - https://bugzilla.mozilla.org/show_bug.cgi?id=1896222
+    data_reviews:
+      - https://bugzilla.mozilla.org/show_bug.cgi?id=1896222
+    data_sensitivity:
+      - interaction
+    notification_emails:
+      - rhelmer@mozilla.com
+    expires: never
+    send_in_pings:
+      - monitor-events
```
sean-rose, line 26: I'd recommend using the built-in `events` ping (which requires no definition in a pings config file) rather than a custom ping.
```suggestion
      - events
```
I'm guessing you were following FxA's lead here, but we're now in the process of switching FxA's Glean setup to use the built-in `events` ping ([DENG-2407](https://mozilla-hub.atlassian.net/browse/DENG-2407)).

sean-rose: Yes, all the event metrics should be sent in the default `events` ping, and you don't need to define a `monitor-events` ping.

sean-rose (approved):

akkomar (commented):
After taking another look at this I would strongly recommend moving `user_id`, `session_id`, `monitor_user_id`, and `plan_tier` out of extras and making them string metrics.

This way it would be easier to work with them - they would have specific columns in BigQuery and would be easier to discover in Glean Dictionary.
We don't have any written guidelines for this, but it seems like measures that are sent with all the events and generally stay unchanged across a session (@rhelmer this seems to be the case here?) should be defined as ping level metrics. cc @quiiver with whom I talked about this yesterday.

For comparison, here's how [FxA defined user_id metric](https://github.com/mozilla/fxa/blob/e10d285e61214f05e8dbe0f08d559336dd0f622f/packages/fxa-shared/metrics/glean/fxa-backend-metrics.yaml#L43-L61).

rhelmer:
> After taking another look at this I would strongly recommend moving `user_id`, `session_id`, `monitor_user_id`, and `plan_tier` out of extras and making them string metrics.
> 
> This way it would be easier to work with them - they would have specific columns in BigQuery and would be easier to discover in Glean Dictionary. We don't have any written guidelines for this, but it seems like measures that are sent with all the events and generally stay unchanged across a session (@rhelmer this seems to be the case here?) should be defined as ping level metrics. cc @quiiver with whom I talked about this yesterday.
> 
> For comparison, here's how [FxA defined user_id metric](https://github.com/mozilla/fxa/blob/e10d285e61214f05e8dbe0f08d559336dd0f622f/packages/fxa-shared/metrics/glean/fxa-backend-metrics.yaml#L43-L61).

OK that's fine with me, I did it this way to more closely align with how our front-end metrics work.

rhelmer:
> After taking another look at this I would strongly recommend moving `user_id`, `session_id`, `monitor_user_id`, and `plan_tier` out of extras and making them string metrics.
> 
> This way it would be easier to work with them - they would have specific columns in BigQuery and would be easier to discover in Glean Dictionary. We don't have any written guidelines for this, but it seems like measures that are sent with all the events and generally stay unchanged across a session (@rhelmer this seems to be the case here?) should be defined as ping level metrics. cc @quiiver with whom I talked about this yesterday.
> 
> For comparison, here's how [FxA defined user_id metric](https://github.com/mozilla/fxa/blob/e10d285e61214f05e8dbe0f08d559336dd0f622f/packages/fxa-shared/metrics/glean/fxa-backend-metrics.yaml#L43-L61).

OK I went ahead and made this change - it used `extra_keys` because that's the way the front-end metrics work, which was at the request of DS. As long as we won't have a problem piecing together these things in analysis I think it's fine.

sean-rose:
> > After taking another look at this I would strongly recommend moving `user_id`, `session_id`, `monitor_user_id`, and `plan_tier` out of extras and making them string metrics.
> > This way it would be easier to work with them - they would have specific columns in BigQuery and would be easier to discover in Glean Dictionary. We don't have any written guidelines for this, but it seems like measures that are sent with all the events and generally stay unchanged across a session (@rhelmer this seems to be the case here?) should be defined as ping level metrics. cc @quiiver with whom I talked about this yesterday.
> > For comparison, here's how [FxA defined user_id metric](https://github.com/mozilla/fxa/blob/e10d285e61214f05e8dbe0f08d559336dd0f622f/packages/fxa-shared/metrics/glean/fxa-backend-metrics.yaml#L43-L61).
> 
> OK I went ahead and made this change - it used `extra_keys` because that's the way the front-end metrics work, which was at the request of DS. As long as we won't have a problem piecing together these things in analysis I think it's fine.

@rhelmer, after discussing this with @akkomar some more, we'd recommend going back to the previous `extra_keys` approach to keep things consistent with the frontend events.  Sorry for the runaround!

rhelmer:
> > > After taking another look at this I would strongly recommend moving `user_id`, `session_id`, `monitor_user_id`, and `plan_tier` out of extras and making them string metrics.
> > > This way it would be easier to work with them - they would have specific columns in BigQuery and would be easier to discover in Glean Dictionary. We don't have any written guidelines for this, but it seems like measures that are sent with all the events and generally stay unchanged across a session (@rhelmer this seems to be the case here?) should be defined as ping level metrics. cc @quiiver with whom I talked about this yesterday.
> > > For comparison, here's how [FxA defined user_id metric](https://github.com/mozilla/fxa/blob/e10d285e61214f05e8dbe0f08d559336dd0f622f/packages/fxa-shared/metrics/glean/fxa-backend-metrics.yaml#L43-L61).
> > 
> > 
> > OK I went ahead and made this change - it used `extra_keys` because that's the way the front-end metrics work, which was at the request of DS. As long as we won't have a problem piecing together these things in analysis I think it's fine.
> 
> @rhelmer, after discussing this with @akkomar some more, we'd recommend going back to the previous `extra_keys` approach to keep things consistent with the frontend events. Sorry for the runaround!

No problem, thanks! I'll revert that commit.

sean-rose:
> > @rhelmer, after discussing this with @akkomar some more, we'd recommend going back to the previous `extra_keys` approach to keep things consistent with the frontend events. Sorry for the runaround!
> 
> No problem, thanks! I'll revert that commit.

I hate to flip-flop on this again, but I just realized there's another factor that probably tips the scales back toward having those IDs as string metrics: data access/deletion requests.  It'd be much more straightforward to handle those if the IDs are simple columns rather than being in the array structure the extra keys values end up in.

So now I'm thinking we should use string metrics for the user/session-level properties in both the backend and the frontend (which would involve changing frontend telemetry, with LookML overrides to fall back to the previous location for those values in earlier frontend events), but with the following naming/hierarchy so the final metric names/columns make sense and aren't repeated with different categories/prefixes:
- `mozilla`
  - `account_id` (metric column name `metrics.string.mozilla_account_id`)
- `monitor`
  - `user_id` (metric column name `metrics.string.monitor_user_id`)
  - `session_id` (metric column name `metrics.string.monitor_session_id`)
  - `plan_tier` (metric column name `metrics.string.monitor_plan_tier`)

@akkomar what do you think?

akkomar:
> I hate to flip-flop on this again, but I just realized there's another factor that probably tips the scales back toward having those IDs as string metrics: data access/deletion requests. It'd be much more straightforward to handle those if the IDs are simple columns rather than being in the array structure the extra keys values end up in.

@sean-rose Good point, these request (DSR) would be easier to handle with ping-level metrics.

> So now I'm thinking we should use string metrics for the user/session-level properties in both the backend and the frontend (which would involve changing frontend telemetry, with LookML overrides to fall back to the previous location for those values in earlier frontend events) (...)

Naming/hierarchy looks good to me. Ideally, to make this even easier for DSAR we could have some standardized naming conventions for these identifiers. But I wouldn't split hairs over this as we'll already need to support customizations because of https://bugzilla.mozilla.org/show_bug.cgi?id=1889123.

As for switching the frontend part too - do you know how much effort this requires? It would be great to have these fields consistently in string metrics, but I'm not sure a migration in frontend is worth it. Also it seems to me that unless we rewrite the old data, we'll need to support identfiers in extras for DSRs anyway until this data expires.

FYI @ksiegler1

sean-rose:
> As for switching the frontend part too - do you know how much effort this requires? It would be great to have these fields consistently in string metrics, but I'm not sure a migration in frontend is worth it.

I'd hope most of the additional complexity could be contained in centralized code similar to [how FxA is doing it](https://github.com/mozilla/fxa/blob/85633d9ce22db0bee8c7f3ac98ae53ee3b0ec12e/packages/fxa-content-server/app/scripts/lib/glean/index.ts#L83-L114), in which case it probably wouldn't be a ton of effort, but that's optimistic speculation on my part. @rhelmer what do you think?

> Also it seems to me that unless we rewrite the old data, we'll need to support identfiers in extras for DSRs anyway until this data expires.

It should be pretty easy to populate the new string metric columns for historical records using a simple `UPDATE` query.

rhelmer:
> > As for switching the frontend part too - do you know how much effort this requires? It would be great to have these fields consistently in string metrics, but I'm not sure a migration in frontend is worth it.
> 
> I'd hope most of the additional complexity could be contained in centralized code similar to [how FxA is doing it](https://github.com/mozilla/fxa/blob/85633d9ce22db0bee8c7f3ac98ae53ee3b0ec12e/packages/fxa-content-server/app/scripts/lib/glean/index.ts#L83-L114), in which case it probably wouldn't be a ton of effort, but that's optimistic speculation on my part. @rhelmer what do you think?
> 
> > Also it seems to me that unless we rewrite the old data, we'll need to support identfiers in extras for DSRs anyway until this data expires.
> 
> It should be pretty easy to populate the new string metric columns for historical records using a simple `UPDATE` query.
> 
> IMO it'd be worth some reasonable effort now to put things in a more supportable state going forward.

I think migrating the front-end is doable, I'd need to do it in a separate ticket though. I can convert the `metrics.yaml` back again.

rhelmer:
> So now I'm thinking we should use string metrics for the user/session-level properties in both the backend and the frontend (which would involve changing frontend telemetry, with LookML overrides to fall back to the previous location for those values in earlier frontend events), but with the following naming/hierarchy so the final metric names/columns make sense and aren't repeated with different categories/prefixes:
> 
>     * `mozilla`
>       
>       * `account_id` (metric column name `metrics.string.mozilla_account_id`)
> 
>     * `monitor`
>       
>       * `user_id` (metric column name `metrics.string.monitor_user_id`)
>       * `session_id` (metric column name `metrics.string.monitor_session_id`)
>       * `plan_tier` (metric column name `metrics.string.monitor_plan_tier`)

@sean-rose I reverted back to the previous version of the `backend-metrics.yaml`. Do I need to modify it to account for this too or is that something you're proposing handling on the Glean side?

sean-rose:
> > So now I'm thinking we should use string metrics for the user/session-level properties in both the backend and the frontend (which would involve changing frontend telemetry, with LookML overrides to fall back to the previous location for those values in earlier frontend events), but with the following naming/hierarchy so the final metric names/columns make sense and aren't repeated with different categories/prefixes:
> > 
> > * `mozilla`
> >   * `account_id` (metric column name `metrics.string.mozilla_account_id`)
> > * `monitor`
> >   * `user_id` (metric column name `metrics.string.monitor_user_id`)
> >   * `session_id` (metric column name `metrics.string.monitor_session_id`)
> >   * `plan_tier` (metric column name `metrics.string.monitor_plan_tier`)
> 
> @sean-rose I reverted back to the previous version of the `backend-metrics.yaml`. Do I need to modify it to account for this too or is that something you're proposing handling on the Glean side?

Yes, you'll need to modify `backend-metrics.yaml` to do that.  With the current `backend-metrics.yaml` the resulting Monitor backend Glean events table in BigQuery would have the following set of duplicative metric columns:
- `metrics.string.account_monitor_user_id`
- `metrics.string.account_plan_tier`
- `metrics.string.account_session_id`
- `metrics.string.account_user_id`
- `metrics.string.page_monitor_user_id`
- `metrics.string.page_plan_tier`
- `metrics.string.page_session_id`
- `metrics.string.page_user_id`
- `metrics.string.subscription_monitor_user_id`
- `metrics.string.subscription_plan_tier`
- `metrics.string.subscription_session_id`
- `metrics.string.subscription_user_id`

rhelmer:
OK thanks @sean-rose lmk if this is what you were expecting.

sean-rose:
> OK thanks @sean-rose lmk if this is what you were expecting.

Not exactly:
* The string metrics should be specified once directly under top-level `mozilla` and `monitor` category keys.
* The event metrics should be specified directly under the appropriate top-level category keys (multiple levels of nesting are not supported).

I tried pushing a commit to correct those issues myself, but it turns out I don't have access (which makes sense), so I'll add a separate comment with the changes I'd recommend as one big code suggestion.

src/telemetry/backend-metrics.yaml
```diff
@@ -0,0 +1,344 @@
...
+        - https://bugzilla.mozilla.org/show_bug.cgi?id=1896222
+      data_reviews:
+        - https://bugzilla.mozilla.org/show_bug.cgi?id=1896222
+      data_sensitivity:
+        - interaction
+      notification_emails:
+        - rhelmer@mozilla.com
+      expires: never
+      send_in_pings:
+        - events
```
sean-rose, line 344:
```suggestion
mozilla:
  account_id:
    description: Mozilla Accounts ID.
    type: string
    lifetime: application
    send_in_pings:
      - events
    notification_emails:
      - rhelmer@mozilla.com
    bugs:
      - https://bugzilla.mozilla.org/show_bug.cgi?id=1896222
    data_reviews:
      - https://bugzilla.mozilla.org/show_bug.cgi?id=1896222
    expires: never
    data_sensitivity:
      - interaction

monitor:
  session_id:
    description: An ID that allows us to track “sessions” of the user experience within the product.
    type: string
    lifetime: application
    send_in_pings:
      - events
    notification_emails:
      - rhelmer@mozilla.com
    bugs:
      - https://bugzilla.mozilla.org/show_bug.cgi?id=1896222
    data_reviews:
      - https://bugzilla.mozilla.org/show_bug.cgi?id=1896222
    expires: never
    data_sensitivity:
      - interaction

  user_id:
    description: Mozilla Monitor-specific user ID.
    type: string
    lifetime: application
    send_in_pings:
      - events
    notification_emails:
      - rhelmer@mozilla.com
    bugs:
      - https://bugzilla.mozilla.org/show_bug.cgi?id=1896222
    data_reviews:
      - https://bugzilla.mozilla.org/show_bug.cgi?id=1896222
    expires: never
    data_sensitivity:
      - interaction

  plan_tier:
    description: Which tier of plan the user is on [Free, Plus]
    type: string
    lifetime: application
    send_in_pings:
      - events
    notification_emails:
      - rhelmer@mozilla.com
    bugs:
      - https://bugzilla.mozilla.org/show_bug.cgi?id=1896222
    data_reviews:
      - https://bugzilla.mozilla.org/show_bug.cgi?id=1896222
    expires: never
    data_sensitivity:
      - interaction

page:
  view:
    type: event
    description: |
      A web page view event distinguishable by the page’s URL.
    bugs:
      - https://bugzilla.mozilla.org/show_bug.cgi?id=1896222
    data_reviews:
      - https://bugzilla.mozilla.org/show_bug.cgi?id=1896222
    data_sensitivity:
      - interaction
    notification_emails:
      - rhelmer@mozilla.com
    expires: never
    send_in_pings:
      - events
    extra_keys:
      path:
        description: The path of the page.
        type: string
      utm_campaign:
        description: Attribution campaign.
        type: string
      utm_content:
        description: Attribution content.
        type: string
      utm_medium:
        description: Attribution medium.
        type: string
      utm_source:
        description: Attribution source.
        type: string
      utm_term:
        description: Attribution term.
        type: string
      referrer:
        description: The referring URL, if any.
        type: string

subscription:
  activate:
    type: event
    description: |
      A subscription has been activated for this user.
    bugs:
      - https://bugzilla.mozilla.org/show_bug.cgi?id=1896222
    data_reviews:
      - https://bugzilla.mozilla.org/show_bug.cgi?id=1896222
    data_sensitivity:
      - interaction
    notification_emails:
      - rhelmer@mozilla.com
    expires: never
    send_in_pings:
      - events

  cancel:
    type: event
    description: |
      A subscription cancellation has taken effect, and the user is no longer subscribed to Monitor Plus.
    bugs:
      - https://bugzilla.mozilla.org/show_bug.cgi?id=1896222
    data_reviews:
      - https://bugzilla.mozilla.org/show_bug.cgi?id=1896222
    data_sensitivity:
      - interaction
    notification_emails:
      - rhelmer@mozilla.com
    expires: never
    send_in_pings:
      - events

account:
  create:
    type: event
    description: |
      A new Mozilla Monitor account has been created.
    bugs:
      - https://bugzilla.mozilla.org/show_bug.cgi?id=1896222
    data_reviews:
      - https://bugzilla.mozilla.org/show_bug.cgi?id=1896222
    data_sensitivity:
      - interaction
    notification_emails:
      - rhelmer@mozilla.com
    expires: never
    send_in_pings:
      - events

  remove:
    type: event
    description: |
      An existing Mozilla Monitor account has been deleted.
    bugs:
      - https://bugzilla.mozilla.org/show_bug.cgi?id=1896222
    data_reviews:
      - https://bugzilla.mozilla.org/show_bug.cgi?id=1896222
    data_sensitivity:
      - interaction
    notification_emails:
      - rhelmer@mozilla.com
    expires: never
    send_in_pings:
      - events

  profile_change:
    type: event
    description: |
      An existing Mozilla Accounts profile has been changed, for example the primary email address.
    bugs:
      - https://bugzilla.mozilla.org/show_bug.cgi?id=1896222
    data_reviews:
      - https://bugzilla.mozilla.org/show_bug.cgi?id=1896222
    data_sensitivity:
      - interaction
    notification_emails:
      - rhelmer@mozilla.com
    expires: never
    send_in_pings:
      - events

  password_change:
    type: event
    description: |
      The password for an existing Mozilla Account has been changed, and existing auth tokens have been revoked.
    bugs:
      - https://bugzilla.mozilla.org/show_bug.cgi?id=1896222
    data_reviews:
      - https://bugzilla.mozilla.org/show_bug.cgi?id=1896222
    data_sensitivity:
      - interaction
    notification_emails:
      - rhelmer@mozilla.com
    expires: never
    send_in_pings:
      - events
```

sean-rose (approved):

akkomar (approved):

PR #4548, opened 2024-05-16
Save email preferences changes

codemist:
# References: 
Jira: MNTOR-3168

# Description

# Screenshot (if applicable)

Not applicable.

# How to test


src/app/api/v1/user/update-comm-option/route.ts
```diff
@@ -64,3 +69,9 @@ export async function POST(req: NextRequest) {
     return NextResponse.redirect(AppConstants.SERVER_URL, 301);
   }
 }
+
+function isEmailUpdateCommType(
+  value: string,
+): value is EmailUpdateCommTypeOfOptions {
+  return ["null", "affected", "primary"].includes(value);
```
pdehaan, line 76: Q: Should we create an `enum` for these since we seem to semi-duplicate some values here and on L42+L45?

pdehaan: Also, are there other possible values for `value` that aren't one of these three?

Vinnl: TypeScript `enum`s are incompatible with the enum proposal for JS and changes runtime behaviour, and unions of literal strings (which we use here - `EmailUpdateCommTypeOfOptions`) provide the same type safety, so it's recommended to use those.

That type also indicates that indeed, these are the only three valid values :)

That said... See my other comments.

src/app/api/v1/user/update-comm-option/route.ts
```diff
@@ -48,8 +48,13 @@ export async function POST(req: NextRequest) {
         default:
           allEmailsToPrimary = null;
       }
-      await setAllEmailsToPrimary(subscriber, allEmailsToPrimary);
-      await setMonthlyMonitorReport(subscriber, monthlyMonitorReport);
+
```
Vinnl, line 51: I can't leave a comment on the relevant lines because they're unchanged, but it's best to update the type representing the request body to indicate that one of the properties should now be left empty:

```ts
export interface EmailUpdateCommOptionRequest {
  instantBreachAlerts?: EmailUpdateCommTypeOfOptions;
  monthlyMonitorReport?: boolean;
}
```

And then TS will show you what code to wrap in an `if (typeof <x> !== "undefined")`, and you won't need the custom type guard (`isEmailUpdateCommType`).

src/app/(proper_react)/(redesign)/(authenticated)/user/(dashboard)/settings/AlertAddressForm.tsx
```diff
@@ -74,22 +71,15 @@ export const AlertAddressForm = (props: Props) => {
     defaultValue: commsValue(),
     onChange: (newValue) => {
       const chosenOption = newValue as EmailUpdateCommTypeOfOptions;
-      const body: EmailUpdateCommOptionRequest = {
-        instantBreachAlerts: chosenOption,
-        monthlyMonitorReport: monitorReportAllowed,
-      };
```
Vinnl, line 80: If you leave this in and make the change to the type that I suggested in the other file, then if people modify the backend, TS will warn us that we'll have to update the frontend as well.

(In fact, we probably want to do the same for the call to update `monthlyMonitorReport`, since we have no checks that we're matching the request body type there either.)

```suggestion
      const body: EmailUpdateCommOptionRequest = {
        instantBreachAlerts: chosenOption,
      };
```

Vinnl (commented):
Sorry for another review with just remarks, but I hope these are small ones...

src/app/api/v1/user/update-comm-option/route.ts
```diff
@@ -48,8 +48,13 @@ export async function POST(req: NextRequest) {
...
+      if (typeof instantBreachAlerts !== "undefined") {
+        await setAllEmailsToPrimary(subscriber, allEmailsToPrimary);
+      }
+      if (typeof monthlyMonitorReport === "boolean") {
```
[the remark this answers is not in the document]
Vinnl: Yep!

PR #4554, opened 2024-05-17
Fix e2e test for branch mntor-3154

flozia:
Update the expected dashboard URL in the “Authentication flow verification” e2e test.


mansaj (approved):
Thanks, LGTM

PR #4573, opened 2024-05-21
MNTOR-3155 - look up users by fxa uid not primary sha1 hash of email

rhelmer:
# References: 
Jira: MNTOR-3155

# Description

This is mostly just removing code, since internally this code turned a sha1 hash of the primary email address into a fxa UID for lookup. I also noticed that the error handling wasn't logging exceptions correctly so I corrected that.

# How to test

`GET` of a valid `fxa_uid` from `subscribers` table should return relevant IDs.

`PUT` with the following should take the appropriate action:

```
{
  "actions":[
    "subscribe",
    "unsubscribe",
    "delete_onerep_profile",
    "delete_onerep_scans",
    "delete_onerep_scan_results",
    "delete_subscriber"
  ]
}
```

http://localhost:6060/admin/dev provides a convenient dev-only web UI to manage this.

# Checklist (Definition of Done)
- [x] Commits in this PR are minimal and [have descriptive commit messages](https://chris.beams.io/posts/git-commit/).
- [x] Jira ticket has been updated (if needed) to match changes made during the development process.

PR #4574, opened 2024-05-22
MNTOR-3155 - Add dev-admin-only server action to look up fxa uid by email address

rhelmer:
# References: 
Jira: MNTOR-3155

# Description

PR #4573 changed the `admin/user` API to look up by FxA UID instead of email address, but I didn't make the equivalent change to the local-dev-only `admin/dev` UI. This adds a local-dev-only, admin-only server action to look up the fxa UID for the sha1 hash of an email address.

This isn't the most efficient way to do things since it's introducing another round-trip to the server. We might consider as a follow-up just moving all the server-side work into a server action and not using the API, but on the other hand it's nice to have the client exercise the API.

# How to test

Use the local-dev-only `admin/dev` UI to look up a user and perform actions on the account.


Vinnl (approved):
Yeah, this sounds like a fine solution to me. I think we'll survive the extra roundtrip :)

PR #4577, opened 2024-05-22
Update docs on the l10n process

Vinnl:
Talking to @jespy2 yesterday I realised we didn't have up-to-date docs on how our localization process works, so I figured I'd take a stab at aligning it with our current processes.


jespy2 (approved):
LGTM

PR #4578, opened 2024-05-22
Add documentation on how I test dependency updates

Vinnl:
Note that I was only able to add docs for those that I handle. It would be nice to have docs for the rest as well, but this should at least be sufficient to cover for me.

# References: 
Jira: MNTOR-3195


jespy2 (approved):
So far looks really good.  Comprehensive and detailed!  Looking forward to trying it out.

flozia (approved):
Thank you for documenting this!

PR #4579, opened 2024-05-22
Remove inline styles from server components

Vinnl:
# References: 
Jira: MNTOR-3082
Figma: N/A

# Description

Our Content Security Policy has `style-src 'self'`, which doesn't include `'unsafe-eval'`. This means that inline styles are blocked in server components (though they can be added in Client Components, because our CSP does allow client-side JS).

We may still want to consider loosening that up if it turns out that we end up wasting time a lot of figuring out why our inline styles won't render
(https://github.com/mozilla/blurts-server/pull/4531), but since the CSP is in place now, it'll at least be immediately clear in the future when they don't work, so that might not be necessary.

As for specific issues, this fixes the "exposures" label in the hero image not being centered.

# Screenshot (if applicable)

Old:

image

New:

image

# How to test

Open the non-US landing page. The text in the hero image should look aligned.

- [ ] I've added a unit test to test for potential regressions of this bug. - N/A, visual change


src/app/components/server/Icons.tsx
```diff
@@ -510,9 +510,6 @@ export const AvatarIcon = (
       />
       <mask
         id="mask0_328_1696"
-        style={{
-          maskType: "luminance",
```
Vinnl, line 514: [This is the default value](https://developer.mozilla.org/en-US/docs/Web/CSS/mask-type#formal_definition), so removing this doesn't change anything. (It's the avatar icon at the bottom of the manual resolution page in the guided flow.)

flozia (approved):
Looking good!

PR #4582, opened 2024-05-23
MNTOR-3136 - Verify that the Premium upsell screen is displayed correctly

mozilloid:
# References: 
Jira: MNTOR-3136
Figma: 

# Description
Checks that the upsell screen button works by checking the URL. Checks the bullet points by finding elements containing the text values, and checking that they're visible. Checks toggles by clicking on them and verifying that the price and plan text fields have changed. 

# Screenshot (if applicable)

Not applicable.

# How to test


src/e2e/specs/dashboard.spec.ts
```diff
@@ -302,6 +302,58 @@ test.describe(`${process.env.E2E_TEST_ENV} - Breaches Dashboard  - Payment`, ()
...
+  }) => {
+    test.info().annotations.push({
+      type: "testrail",
+      description:
+        "https://testrail.stage.mozaws.net/index.php?/cases/view/2463625",
+    });
+
+    //checking that the user can reach upsell page
+    await dashboardPage.goToDashboard();
+    const upsellButton = page.locator("section > div a");
```
pdehaan, line 319: Couple ideas here:
1. Can this be any more specific? I tried from Canada and US (via VPN) and both seemed to have a `section > div a`
    - Canada: "Monitor more emails" button
    - US: "Get first scan free" button and "Learn more" link 
2. Can we move the upsellButton locator into the page object model file with the other locators? That way we can keep a lot of the logic together and reuse them a bit easier.

src/e2e/specs/dashboard.spec.ts
```diff
@@ -302,6 +302,58 @@ test.describe(`${process.env.E2E_TEST_ENV} - Breaches Dashboard  - Payment`, ()
...
+    await expect(upsellButton).toBeVisible();
+    await upsellButton.click();
+    await page.waitForURL(/.*\/fix\/.*\/view-data-brokers\/?/);
```
pdehaan, line 322: Interesting, why does this need the optional trailing slash (`\/?`)?
Is this intermittent in the tests?

src/e2e/specs/dashboard.spec.ts
```diff
@@ -302,6 +302,58 @@ test.describe(`${process.env.E2E_TEST_ENV} - Breaches Dashboard  - Payment`, ()
...
+    const removeThemButton = page.getByText("Remove them for me");
+    await removeThemButton.click();
+
+    //checking the bullet points
+    await page.waitForURL(/.*\/fix\/.*\/automatic-remove\/?/);
+    const ulElement = page.locator("div > strong + ul");
```
pdehaan, line 328: Same comment as above, re: adding the custom locators to the page object model (POM) &mdash; if there isn't already an existing selector for this.

src/e2e/specs/dashboard.spec.ts
```diff
@@ -302,6 +302,58 @@ test.describe(`${process.env.E2E_TEST_ENV} - Breaches Dashboard  - Payment`, ()
...
+    expect(await ulElement.isVisible()).toBeTruthy();
```
pdehaan, line 329: instead of `.isVisible()`+`.toBeTruthy()`, can we use [web first assertions](https://playwright.dev/docs/best-practices#use-web-first-assertions); like `.toBeVisible()` here?

src/e2e/specs/dashboard.spec.ts
```diff
@@ -302,6 +302,58 @@ test.describe(`${process.env.E2E_TEST_ENV} - Breaches Dashboard  - Payment`, ()
...
+    const label1 = page.locator("div > label").nth(1);
+
+    await label0.click();
+    const price0 = page.locator("div > strong + span").textContent();
+    const plan0 = page.locator("div > strong + span + a").textContent();
+    await label1.click();
+    const price1 = page.locator("div > strong + span").textContent();
+    const plan1 = page.locator("div > strong + span + a").textContent();
+    const areTextsChanging = price0 !== price1 && plan0 !== plan1;
+    expect(areTextsChanging).toBeTruthy();
```
pdehaan, line 355: not sure if this might be simpler w/ two separate asserts something like this:
```js
expect(price0).not.toEqual(price1);
expect(plan0).not.toEqual(plan1);
```
If nothing else, will give you slightly more helpful error messaging vs "expected true, received false".

src/e2e/specs/dashboard.spec.ts
rafeerahman: In general I believe its best practice to use built-in playwright methods instead of xpaths. 

We could either use a more specific/readable xpath, like `//div[starts-with(@class, "dataBrokerProfiles_featuresList")]/ul`, or grab each li element with getByText() and assert their visibility in this case. 

I'm using [codegen](https://playwright.dev/docs/codegen) to generate playwright locators for me, and I notice theres some invisible char's in the text. This would cause `getByText(Monthly scan of 190 data)` to not work, which you might have tried. You would need to add the invisible char's by copying and pasting it directly from the site, `getByText(Monthly scan of ⁨190⁩ data broker)` (pasting this in vs-code will show the difference).

src/e2e/specs/dashboard.spec.ts
```diff
@@ -302,6 +302,58 @@ test.describe(`${process.env.E2E_TEST_ENV} - Breaches Dashboard  - Payment`, ()
...
+    const expectedItems = [
+      "Monthly scan of ⁨190⁩ data broker sites that may be selling your personal info",
+      "Automatic data removal from sites that are selling your personal info",
+      "Guided experience through high risk data breaches that require manual steps",
+      "Continuous monitoring for new exposures",
+      "Alerts when your data has been breached",
+    ];
+
+    for (const itemText of expectedItems) {
+      const liElement = page.locator(`ul li:has-text("${itemText}")`).first();
```
rafeerahman, line 340: You can also nest locators. For example, you could do `const liElems = ulElement.locator("/li")` I believe to grab all the \<li> elements.

Then you can use things like liElems.count() and go through them one by one similarly to how you did here.

But in my opinion how you did it here achieves the same thing, so I don't see a problem with this.

src/e2e/specs/dashboard.spec.ts
```diff
@@ -302,6 +302,58 @@ test.describe(`${process.env.E2E_TEST_ENV} - Breaches Dashboard  - Payment`, ()
...
+      await expect(liElement).toBeVisible();
+    }
+
+    //testing that toggles work
+    const label0 = page.locator("div > label").nth(0);
+    const label1 = page.locator("div > label").nth(1);
+
+    await label0.click();
+    const price0 = page.locator("div > strong + span").textContent();
```
rafeerahman, line 349: We should put these locators in one of the `page` files, probably dashboardPage.ts in this case.

I believe we can re-use one of the existing properties from the class. 

If not, instead of using an xpath, using something like `page.getByText('Yearly', { exact: true })` should work for the yearly or monthly toggle buttons.

mozilloid: @rafeerahman Do you think it should be in dashboardPage.ts if they're multiple clicks away from being visible?

rafeerahman: I think creating a new page file would be the best option. Something like automaticRemovalPage.ts? You can follow the outline of the dashboardPage.ts class.

When you create a new page file, you also need to add it to /fixtures/basePages.ts so you can use it within the spec files.

src/e2e/specs/dashboard.spec.ts
mozilloid: @pdehaan same question as [here](https://github.com/mozilla/blurts-server/pull/4582#discussion_r1611874394) - do you think it should be in dashboard POM if these locators are multiple clicks away? I don't see a point in creating a separate POM just for these locators as well.

rhelmer (approved):
I don't have anything to add beyond what @pdehaan and @rafeerahman already suggested.

src/e2e/specs/dashboard.spec.ts
mozilloid: I guess it doesn't really matter since the URL resolves to be without a '/',  but I just wanted to be sure.

src/e2e/specs/dashboard.spec.ts
mozilloid: Seems like Locator does not support 'for..of', which makes the code more complicated than it should be. I think I'll put the expected contents inside the automaticRemovePage POM, which will hopefully make it cleaner

mozilloid:
@pdehaan , @rafeerahman I outsourced all locators to their respective POMs. Thus, I added new POMs for databrokersPage and automaticRemovePage. I also put the expected contents into the POM instead of the test. Also, upsell button typically only has either "lets fix it" or "lets keep going" as the text, so I just performed a getByText there. I think this should finalize the PR, lmk!

src/e2e/pages/dashBoardPage.ts
```diff
@@ -218,6 +220,9 @@ export class DashboardPage {
     this.dashboardPageLink = page.getByRole("link", { name: "Dashboard" });
 
     this.faqsPageLink = page.getByTitle("Frequently asked questions").first();
+
+    //upsell button
+    this.upsellScreenButton = page.getByText(/Let’s (keep going|fix it)/);
```
pdehaan, line 225: 👍 Nice use of regex!

pdehaan (approved):
Looks good, @mozilloid, thanks!

PR #4606, opened 2024-05-29
Revert "Import translations from l10n repository (2024-05-28)"

rhelmer:
Reverts mozilla/blurts-server#4600

We are investigating an unexpected spike in out-of-memory errors that started on stage right after this was landed. It seems unlikely this patch would cause the problem by itself, unless there's an existing bug in Fluent or its dependencies but we want to try rolling back things in the regression range.


rhelmer:
Thanks - working on reproducing this on stage before landing.

rhelmer:
OK I see it in the stage logs, let's see if this makes any difference. If not, I'll expand the regression range and try again in the morning. No luck reproducing locally so far.

PR #4608, opened 2024-05-30
Mntor-3130 - Verify that the Apps&Services header options work correctly.

mozilloid:
# References:

Jira: MNTOR-3130
Figma:

# Description
Similarly to mntor-3131, goes to each option and verifies URL. For Apps&Services header, checks the URL based on a single RegExp. 

# Screenshot (if applicable)

Not applicable.

# How to test
npx playwright test -g "Verify that the Apps and Services header options work correctly."


src/e2e/pages/dashBoardPage.ts
```diff
@@ -135,6 +139,13 @@ export class DashboardPage {
     this.actionNeededTab = page.getByRole("tab", { name: "Action needed" });
     this.fixedTab = page.getByRole("tab", { name: "Fixed" });
     this.profileButton = page.getByTitle("Profile").nth(1);
+
+    this.profileSettings = page.locator(
+      'a[title*="Configure"][title*="Mozilla Monitor"]',
+    );
+    this.profileSignOut = page.locator(
+      'button[title*="Sign out of"][title*="Mozilla Monitor"]',
```
pdehaan, line 147: Same Q as above, re: double `[title*=""]` lookup.

```html
<button class="UserMenu_menuItemCta__UDn5d"
     title="Sign out of ⁨Mozilla Monitor⁩">
  <img alt="" loading="lazy" width="24" height="24"
     decoding="async" data-nimg="1"
     src="/_next/static/media/menu-icon-signout.429478a7.svg"
     style="color: transparent;">Sign out</button>
```

src/e2e/pages/dashBoardPage.ts
```diff
@@ -155,6 +167,7 @@ export class DashboardPage {
     this.servicesFirefoxMobile = page.getByRole("link", {
       name: "⁨Firefox⁩ for Mobile",
     });
+    this.servicesMozilla = page.locator('[data-key="mozilla"] > a');
```
pdehaan, line 170: OK, I think this is the bottom link in the DOM of the popup bento box menu:
```html
<li role="menuitem" tabindex="-1"
     data-key="mozilla" class="AppPicker_menuItemWrapper__TpxV9 ">
  <a href="https://www.mozilla.org/?utm_source=monitor.mozilla.org&amp;utm_medium=referral&amp;utm_campaign=bento&amp;utm_content=desktop"
     class="AppPicker_menuLink__O8tJX AppPicker_mozillaLink__LTe4E"
     target="_blank" rel="noopener noreferrer">Made by ⁨Mozilla⁩</a></li>
```

Wonder if something like `page.getByRole("listitem", { name: "Made by Mozilla" })` might work (possibly not since I see you're using the `>` child selector).

src/e2e/specs/dashboard.spec.ts
```diff
@@ -127,6 +133,94 @@ test.describe(`${process.env.E2E_TEST_ENV} - Breaches Dashboard - Headers`, () =
...
+      dashboardPage.manageProfile,
+      /.*accounts.*settings.*/,
+    );
+    await openProfileMenuItem(
+      dashboardPage.profileSettings,
+      /.*\/user\/settings.*/,
+    );
+
+    const base_url = process.env["E2E_TEST_BASE_URL"];
+    expect(base_url).toBeTruthy();
```
pdehaan, line 221: What is this testing?
Don't we fail sooner/elsewhere if the `E2E_TEST_BASE_URL` env var isn't set?

src/e2e/specs/dashboard.spec.ts
```diff
@@ -535,26 +625,6 @@ test.describe(`${process.env.E2E_TEST_ENV} - Breaches Dashboard - Footer`, () =>
         "https://testrail.stage.mozaws.net/index.php?/cases/view/2463570",
     });
 
-    const clickOnATagCheckDomain = async (
```
pdehaan, line 538: For future reviewers (or me in 15 minutes when i forget), this removed function was moved into utils/helpers.ts

src/e2e/specs/dashboard.spec.ts
```diff
@@ -642,3 +718,42 @@ test.describe(`${process.env.E2E_TEST_ENV} - Breaches Dashboard - Navigation`, (
     );
   });
 });
+
+test.describe(`${process.env.E2E_TEST_ENV} - Breaches Dashboard - Data Breaches`, () => {
+  test.beforeEach(async ({ landingPage, page, authPage }) => {
+    await forceLoginAs(
+      "joe@mailinator.com",
+      "TestPass1234",
```
pdehaan, line 726: Interesting... How does this work? If you reuse the same user for every test/run, isn't the profile contaminated?
Do we ever destroy this user?
Do we have steps for if we need a new fake test acct in the future? (just pre-create an account on dev/stage/production with the exact same usernames/passwords so the tests will run everywhere?)

src/e2e/specs/dashboard.spec.ts
```diff
@@ -642,3 +718,42 @@ test.describe(`${process.env.E2E_TEST_ENV} - Breaches Dashboard - Navigation`, (
...
+
+    await expect(dashboardPage.upsellScreenButton).toBeVisible();
+    await dashboardPage.upsellScreenButton.click();
+    await page.waitForURL(/.*\/data-broker-profiles\/view-data-brokers\/?/);
+    await expect(dataBrokersPage.forwardArrowButton).toBeVisible();
+    await dataBrokersPage.forwardArrowButton.click();
+    await page.waitForURL(/.*\/high-risk-data-breaches.*/);
+    const highRiskDataBreachLi = page.locator(
+      'li:has(div:has-text("High risk data breaches"))',
+    );
```
pdehaan, line 752: I'm confused on this locator vs the ... ah, so _this_ locator just matches an `<li>` but it's the sub-sub-locator on L756 that actually gets the child div's text content.

src/e2e/specs/dashboard.spec.ts
```diff
@@ -127,6 +133,94 @@ test.describe(`${process.env.E2E_TEST_ENV} - Breaches Dashboard - Headers`, () =
...
+      "www.mozilla.org",
+      /.*\/products\/vpn\/?.*/,
+    );
+    await clickOnLinkAndGoBack(
+      dashboardPage.servicesRelay,
+      "relay.firefox.com",
+    );
+    await clickOnLinkAndGoBack(
+      dashboardPage.servicesPocket,
+      /getpocket\.com|apps\.apple\.com|app\.adjust\.com/,
```
pdehaan, line 176: Q: Why does this need so many variable `host`s?
Although I'm on macOS and it tries going to add.adjust.com for me, which then tries to redirect me to the apps.apple.com App Store. So I guess it depends on where in the pipeline of `ckickOnLinkAndGoBack()` and `clickOnATagCheckDomain()` we're using the specified host(s) to verify where the link is resolving.

Are you on Windows 10/11? Or in what case(s) would this resolve to getpocket.com? Although I'm not sure if I want to even think about doing OS detection to craft what the matrix of OS=>hostnames would look like.

src/e2e/specs/dashboard.spec.ts
```diff
@@ -127,6 +133,94 @@ test.describe(`${process.env.E2E_TEST_ENV} - Breaches Dashboard - Headers`, () =
...
+      /.*(\/pocket-and-firefox\/?).*|.*about.*|.*pocket-stay-informed.*/,
```
pdehaan, line 177: Similarly here... it's getting a bit confusing since I'm trying to figure out which paths belong to which hosts between these two regexps. 🤔

src/e2e/specs/dashboard.spec.ts
```diff
@@ -127,6 +133,94 @@ test.describe(`${process.env.E2E_TEST_ENV} - Breaches Dashboard - Headers`, () =
...
+        await page.goto(href!);
+      } else {
+        await what.click();
+      }
+      await page.waitForURL(whatUrl);
+    };
+
+    await openProfileMenuItem(
+      dashboardPage.manageProfile,
+      /.*accounts.*settings.*/,
```
pdehaan, line 213: Is this verifying the URL is https://accounts.firefox.com/settings?

src/e2e/specs/dashboard.spec.ts
```diff
@@ -127,6 +133,94 @@ test.describe(`${process.env.E2E_TEST_ENV} - Breaches Dashboard - Headers`, () =
...
+    );
+    await openProfileMenuItem(
+      dashboardPage.profileSettings,
+      /.*\/user\/settings.*/,
```
pdehaan, line 217: Same here re: https://monitor.mozilla.org/user/settings
I wonder if we could get the current project's baseURL and use ``await expect(page).toHaveURL(domainAndPath);``

src/e2e/utils/helpers.ts
```diff
@@ -160,3 +162,63 @@ export function removeUnicodeChars(text: string): string {
...
+  test.slow(
+    true,
+    "this test runs through the welcome scan flow, increasing timeout to address it",
+  );
+
+  // speed up test by ignoring non necessary requests
+  await page.route(/(analytics)/, async (route) => {
+    await route.abort();
+  });
+  await page.context().clearCookies();
```
pdehaan, line 208: Is Playwright persisting cookies/state between tests?

src/e2e/utils/helpers.ts
```diff
@@ -160,3 +162,63 @@ export function removeUnicodeChars(text: string): string {
...
+  } catch {
+    visible = false;
+  }
+  if (visible) {
+    await authPage.useDifferentEmailButton.click();
+    await page.waitForURL(/^(?!.*signin).*/);
+  }
+  await authPage.signIn(email, password);
+  await page.waitForURL("**/user/dashboard");
+  expect(page.url()).toContain("/user/dashboard");
```
pdehaan, line 223: Any benefits to using `expect(page.url()).toContain(...)` vs [`await expect(page).toHaveURL(/.../);`](https://playwright.dev/docs/api/class-pageassertions#page-assertions-to-have-url) for URL assertions?

src/e2e/specs/dashboard.spec.ts
mansaj: commented in the other PR, it's not a good security practice to list out credentials in plain text

src/e2e/pages/dashBoardPage.ts
```diff
@@ -135,6 +139,13 @@ export class DashboardPage {
     this.actionNeededTab = page.getByRole("tab", { name: "Action needed" });
     this.fixedTab = page.getByRole("tab", { name: "Fixed" });
     this.profileButton = page.getByTitle("Profile").nth(1);
+
+    this.profileSettings = page.locator(
+      'a[title*="Configure"][title*="Mozilla Monitor"]',
```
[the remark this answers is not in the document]
mozilloid: There are invisible characters around 'Mozilla Monitor' which for some reason are not properly understood by playwright when I do getByTile. So, I think keeping it this way is the easiest. Also, the sidebar doesn't have a title, so the mapping is unique

src/e2e/pages/dashBoardPage.ts
mozilloid: Same answer as above. Some random characters around Mozilla Monitor

src/e2e/pages/dashBoardPage.ts
mozilloid: Ok! I will chain locators, which should achieve the same behavior

src/e2e/specs/dashboard.spec.ts
mozilloid: Yeah, so its something like: find a li that has a div that has a text.

src/e2e/specs/dashboard.spec.ts
mozilloid: Yeah, so its behaving rather strange. When I access the link normally through my browser, it redirects to apple store, but in tests it goes to the app.adjust. Also, the testrail spec seems to also accept getpocket.com, so i included it as well.

src/e2e/specs/dashboard.spec.ts
mozilloid: Do you think there is a better way to do this if there are so many hosts and so many paths that are thus possible?

src/e2e/specs/dashboard.spec.ts
mozilloid: Yeah

src/e2e/specs/dashboard.spec.ts
mozilloid: I'd still need to call .waitForURL() and pass in that URL. At that point, isn't it redundant?

src/e2e/utils/helpers.ts
mozilloid: It would redirect me to dashboard whenever i'd try to go to the landingPage. If i clear the cookies it would get rid of the auth token, so I can start fresh.

src/e2e/utils/helpers.ts
mozilloid: Sure ill change it.

src/e2e/pages/dashBoardPage.ts
mozilloid: nvm, it has these weird invisible characters here too, and the locator chain becomes rather long... I can change it to be a locator chain, but imo the argument in .locator makes it cleaner.

src/e2e/specs/dashboard.spec.ts
mozilloid: It says that process.env[..] is an option string, so i am just checking that its non-null. You're right that it would've probably failed before that if it was indeed null. I think its presence doesnt make a difference tbh.

src/e2e/specs/dashboard.spec.ts
mozilloid: This user just has some nice properties which are convenient for this test. And i dont destroy the user. What do you mean by contaminated btw?

PR #4612, opened 2024-05-30
MNTOR-3243: rework release retag

mansaj:
# References:

Jira: MNTOR-3243

# Description
Rework the script so that it's using the GH tag event instead of GH release event.
GH release event sometimes confuses branch names with commit sha


.github/workflows/release_retag.yaml
```diff
@@ -0,0 +1,36 @@
...
+      uses: actions/checkout@v4
+
+    - name: Log in to Docker Hub
+      uses: docker/login-action@v3
+      with:
+        username: ${{ secrets.DOCKER_USERNAME }}
+        password: ${{ secrets.DOCKER_PASSWORD }}
+    - name: Extract tag
+      id: extract_tag
+      run: echo "::set-output name=TAG::${GITHUB_REF#refs/tags/}"
```
rhelmer, line 23: Hm, I haven't seen `set-output` before... is it the same thing as https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/ ? Looks like GH deprecated it but postponed it since they see from telemetry folks are using it.

It looks like they have this "environment file" as a replacement, should we do that instead? https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#environment-files

rhelmer (changes requested):

.github/workflows/release_retag.yaml
mansaj: https://github.com/mozilla/blurts-server/pull/4612/commits/2ada006dd064ef457d59dd2c88257d69f3a5db6a

I think this way works better / cleaner

rhelmer (approved):

PR #4617, opened 2024-06-03
Add cronjob for sending first data broker removal fixed email (MNTOR-2700)

flozia:
# References:

Jira: [MNTOR-2700](https://mozilla-hub.atlassian.net/browse/MNTOR-2700)

# Description

Add cronjob script for sending the first data broker removal fixed email introduced in PR #4614.

# How to test

1. Run db migration: `npm run db:migrate`
2. Add feature flag `FirstDataBrokerRemovalFixedEmail`
3. Run cronjob: `npx tsx --tsconfig tsconfig.cronjobs.json src/scripts/cronjobs/firstDataBrokerRemovalFixed.tsx`

[MNTOR-2700]: https://mozilla-hub.atlassian.net/browse/MNTOR-2700?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/db/tables/featureFlags.ts
```diff
@@ -44,7 +44,8 @@ export type FeatureFlagName =
   | "MonthlyActivityEmail"
   | "CsatSurvey"
   | "CancellationFlow"
-  | "ConfirmCancellation";
+  | "ConfirmCancellation"
+  | "FirstDataBrokerRemovalFixedEmail";
```
flozia, line 48: Depending on whether we intend to make this a real experiment (waiting on confirmation) we should consider using the deprecated feature flags. This would spare us from experiment checks for each individual user.

flozia: @rhelmer and I discussed with PM that for the emails it might be better to just use the feature flag here. We don’t intend to run an experiment and having just an on/off switch is handy for the cronjob.

zifiri99: > Depending on whether we intend to make this a real experiment (waiting on confirmation) we should consider using the deprecated feature flags. This would spare us from experiment checks for each individual user.

zifiri99:
> Preview URL :rocket: : https://blurts-server-pr-4617-mgjlpikfea-uk.a.run.app
>

src/db/tables/featureFlags.ts
rhelmer: We might want to re-consider deprecating our own flag system since it's proved to be useful... I think doing slow roll-out is a good general practice so I'm still a bit on the fence about it.

rhelmer: However in this case I agree let's use Monitor's own feature flag system for this feature.

src/db/migrations/20240601114803_add_first_broker_removal_email_sent.js
```diff
@@ -0,0 +1,23 @@
...
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+/**
+ * @param { import("knex").Knex } knex
+ * @returns { Promise<void> }
+ */
+export function up (knex) {
+  return knex.schema.table("subscribers", table => {
+    table.boolean("first_broker_removal_email_sent").defaultTo(false);
```
rhelmer, line 11:
```suggestion
    table.boolean("first_broker_removal_email_sent").defaultTo(false);
    table.index("first_broker_removal_email_sent");
```

Since production has millions of rows on this table and you're querying it regularly below, I suspect you'll want an index on this. We could try to measure beforehand, but GCP has tools that make it pretty easy to see the cost of queries so we can double-check it that way.

src/db/tables/subscribers.js
```diff
@@ -291,6 +291,60 @@ async function deleteResolutionsWithEmail (id, email) {
...
+    // ...with an OneRep account...
+    .whereNotNull("onerep_profile_id")
+    // ...who haven’t received the email...
+    .andWhere("first_broker_removal_email_sent", false)
+    // ...and signed up after the feature flag `FirstDataBrokerRemovalFixedEmail`
+    // has been enabled last.
+    .andWhere("created_at", ">=", flag.modified_at);
+
+  if (Array.isArray(flag.allow_list) && flag.allow_list.length > 0) {
+    // If the feature flag has an allowlist, only send to users on that list.
```
rhelmer, line 336: FYI we removed this from the UI, but the feature still exists in the API and database. We can re-add the ability to change the allow list to the UI if needed.

src/scripts/cronjobs/firstDataBrokerRemovalFixed.tsx
```diff
@@ -0,0 +1,124 @@
...
+
+function isFulfilledResult(
+  result: PromiseSettledResult<SubscriberFirstRemovedScanResult | undefined>,
+): result is PromiseFulfilledResult<SubscriberFirstRemovedScanResult> {
+  return typeof result !== "undefined" && result.status === "fulfilled";
+}
+
+void run();
+
+async function run() {
```
rhelmer, line 32: I think the duplication here is OK for this PR since it's only the second script we've done like this, but we should look for opportunities to share code with `src/scripts/cronjobs/monthlyActivity.tsx`. I am imagining that as soon as we fix a bug or change behavior in one it's going to be easy to forget about the other.

src/scripts/cronjobs/firstDataBrokerRemovalFixed.tsx
```diff
@@ -0,0 +1,124 @@
...
+        data.value.firstRemovedScanResult,
+      );
+    }),
+  );
+  console.log(
+    `[${new Date(Date.now()).toISOString()}] Sent [${subscribersToEmailWithData.length}] first data broker removal fixed emails.`,
+  );
+}
+
+async function sendFirstDataBrokerRemovalFixedActivityEmail(
```
rhelmer, line 97: For example if this function was the only unique thing for each email cron job I think that would be very easy to understand!

src/db/tables/subscribers.js
```diff
@@ -291,6 +291,60 @@ async function deleteResolutionsWithEmail (id, email) {
...
+async function getPotentialSubscribersWaitingForFirstDataBrokerRemovalFixedEmail() {
+  // I'm explicitly referencing the type here, so that these lines of code will
+  // show up as errors when we remove it from the flag list:
+  /** @type {import("./featureFlags.js").FeatureFlagName} */
+  const featureFlagName = "FirstDataBrokerRemovalFixedEmail";
+  // Interactions with the `feature_flags` table would generally go in the
+  // `src/db/tables/featureFlags` module. However, since that module is already
+  // written in TypeScript, it can't be loaded in pre-TypeScript cron jobs,
+  // which currently still import from the subscribers module. Hence, we've
+  // inlined this for now.
```
rhelmer, line 308:
```suggestion
  // inlined this until https://mozilla-hub.atlassian.net/browse/MNTOR-3077 is fixed.
```
Thanks for the comment, I was about to ask why this isn't using the APIs. We have a tech debt ticket in the backlog, I'd suggest linking so we can find and fix this later.

rhelmer (changes requested):
This looks great, I have a few suggestions but the only one that's likely to be very critical is using an index on the `first_broker_removal_email_sent` column, since the `subscribers` table is quite large and only getting larger.

src/scripts/cronjobs/firstDataBrokerRemovalFixed.tsx
flozia: That’s a good point and I was considering this. I’ve refrained from touching anything related to the `monthlyActivity` mail as we are still observing its behavior and I did not want to interfere in any way. As soon as we confirm that we have our cronjobs run stable I’m happy to unify them more.

src/db/tables/subscribers.js
flozia: I’ve taken the comment over from @Vinnl who added it in `getSubscribersWaitingForMonthlyEmail` — can’t take the credit. ;) Updated the comments in both places in

src/db/migrations/20240601114803_add_first_broker_removal_email_sent.js
flozia: Added the index for `irst_broker_removal_email_sent` in [9dc7c02](https://github.com/mozilla/blurts-server/pull/4617/commits/9dc7c028a3dbae1f56e393381089969064684ad3).

src/scripts/cronjobs/firstDataBrokerRemovalFixed.tsx
```diff
@@ -0,0 +1,128 @@
...
+    throw new Error(
+      `Could not send first data broker removal fixed emails, because the env var FIRST_DATA_BROKER_REMOVAL_FIXED_EMAIL_BATCH_SIZE has a non-numeric value: [${process.env.FIRST_DATA_BROKER_REMOVAL_FIXED_EMAIL_BATCH_SIZE}].`,
+    );
+  }
+  const potentialSubscribersToEmail =
+    await getPotentialSubscribersWaitingForFirstDataBrokerRemovalFixedEmail();
+
+  const subscribersToEmailWithData = (
+    await Promise.allSettled(
+      potentialSubscribersToEmail.map(async (subscriber) => {
```
mansaj, line 47: There is no error handling within the `map` function. If any of the asynchronous operations (e.g., refreshStoredScanResults or getLatestOnerepScanResults) fail, the error will not be caught, and the function will fail silently. Is that ok?
Probably best to wrap the whole func in `try/catch`

src/scripts/cronjobs/firstDataBrokerRemovalFixed.tsx
```diff
@@ -0,0 +1,128 @@
...
+        // data from their API. Thus, let's refresh the data in our DB in real-time:
+        if (subscriber.onerep_profile_id !== null) {
+          await refreshStoredScanResults(subscriber.onerep_profile_id);
+        }
+        const latestScan = await getLatestOnerepScanResults(
+          subscriber.onerep_profile_id,
+        );
+
+        let firstRemovedScanResult = null;
+        for (const scanResult of latestScan.results) {
```
mansaj, line 58: This logic can probably be simplified with `.filter.reduce` chain
```
const firstRemovedScanResult = latestScan.results
        .filter(scanResult => !scanResult.manually_resolved && scanResult.status === "removed")
        .reduce((first, scanResult) => {
        ...
```

rhelmer (approved):

src/scripts/cronjobs/firstDataBrokerRemovalFixed.tsx
flozia: This was my first thought as well. I guess it does not matter too much, because the results likely won’t be a huge list, but with the current implementation, we only have to go over them once.

src/scripts/cronjobs/firstDataBrokerRemovalFixed.tsx
flozia: @mansaj For the functionality of sending out the emails that should not matter too much if this would only happen occasionally. The next run should take care of it. In order for us to better notice if we see this failing repeatedly, logging an error would be good. Updated in https://github.com/mozilla/blurts-server/pull/4617/commits/e9a03c6566a164821dd36e646c36e331e0f3d84b.

PR #4627, opened 2024-06-03
disable welcome-scan-optional-info feature on stage for testing Nimbus

rhelmer:
# Description

We're testing roll-out on stage, we need to set `welcome-scan-optional-info` to `false` by default.


pdehaan (approved):
#yolo

PR #4628, opened 2024-06-04
MNTOR-2879 - explicit order-by for getLatestOnerepScanResults

rhelmer:
# References:

Jira: MNTOR-2879

# Description

Duplicate `link` columns are currently filtered out using `DISTINCT` in the `SELECT` statement, but since the `ORDER BY` isn't consistent different the result set can be slightly different, which throws off our calculations.

This might be a OneRep-specific problem, but if it persists in the future we may want to figure out how to merge these sorts of scan results instead of just picking one.

# How to test

Create two or more scan results with the same link, and change the `updated_at` record on them. We also have a test account in production with over 1k results, I think it's just more likely to happen there but could conceivably happen on as few as two scan results in production.


mansaj (approved):
Ah interesting, thanks for catching this

PR #4629, opened 2024-06-04
Cirrus expects ISO-3166 compliant upper-case country code

rhelmer:
# References:

# Description

Make the `countryCode` upper-case for Cirrus, we should consider changing this throughout the code to be more compliant wiht ISO-3166. I'll file a separate ticket if we decide to do that.

### DIFF
--- a/src/app/functions/server/getExperiments.ts
+++ b/src/app/functions/server/getExperiments.ts
@@ -45,7 +45,7 @@ export async function getExperiments(params: {
         context: {
           // Nimbus takes a language, rather than a locale, hence the .split:
           language: params.locale.split("-")[0],
-          region: params.countryCode,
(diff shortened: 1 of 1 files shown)


pdehaan (approved):
R+
But (for my own curiosity) what cases were we seeing the country code not uppercased?

rhelmer:
> R+ But (for my own curiosity) what cases were we seeing the country code not uppercased?

It's explicitly lower-cased here: https://github.com/mozilla/blurts-server/blob/8a4cf7aa23c559f462a3ffbaad744800f4366ed4/src/app/functions/server/getCountryCode.ts#L14

I think maybe it shouldn't be but will need to change all the code and tests, so I'll file a ticket if we decide to do that.

PR #4633, opened 2024-06-05
MNTOR-3271: add table and util functions for subscriber coupon code

mansaj:
# References:

Jira: MNTOR-3271

# Description
For cancellation discounts, we need a way to store the discount for record

# How to test
`npm run db:migrate`

### DIFF
--- a/src/db/tables/subscriber_coupons.ts
+++ b/src/db/tables/subscriber_coupons.ts
@@ -1,0 +1,54 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import createDbConnection from "../connect.js";
+import { logger } from "../../app/functions/server/logging";
+import { SubscriberCouponRow } from "knex/types/tables";
+
+const knex = createDbConnection();
+
+async function checkCouponForSubscriber(
+  subscriberId: number,
+  couponCode: string,
+) {
+  logger.info("checkCouponForSubscriber", subscriberId);
+  return (
+    (
+      await knex("subscriber_coupons")
+        .orderBy("created_at")
+        .where("subscriber_id", subscriberId)
+        .andWhere("coupon_code", couponCode)
+        .returning("*")
+    ).length > 0
+  );
+}
+
+async function addCouponForSubscriber(
+  subscriberId: number,
+  couponCode: string,
+) {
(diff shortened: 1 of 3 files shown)


src/db/tables/subscriber_coupons.ts
```diff
@@ -0,0 +1,54 @@
...
+      })
+      .returning("*");
+  } catch (e) {
+    if ((e as Error).message.includes("violates unique constraint")) {
+      logger.info("addCouponForSubscriber", {
+        subscriberId,
+        error: (e as Error).message,
+      });
+    } else {
+      logger.error(e);
```
rhelmer, line 48: This prints a pretty confusing message sometimes, I think it's better if we do something like:
```
logger.error("could_not_add_coupon", { error: JSON.stringify(e) });
```
Because when you see the message in the logs you'll just see whatever `e.toString()` returns, so it might not be clear exactly where it was thrown.

Do you think it should re-throw the error here? Otherwise the calling code won't have a way to know it didn't work, if that's important.

src/db/tables/subscriber_coupons.ts
```diff
@@ -0,0 +1,54 @@
...
+) {
+  logger.info("checkCouponForSubscriber", subscriberId);
+  return (
+    (
+      await knex("subscriber_coupons")
+        .orderBy("created_at")
+        .where("subscriber_id", subscriberId)
+        .andWhere("coupon_code", couponCode)
+        .returning("*")
+    ).length > 0
```
rhelmer, line 23: Hm, took me a moment to realize this was an existence check... if we use raw you could use [SELECT EXISTS](https://www.postgresql.org/docs/current/functions-subquery.html#FUNCTIONS-SUBQUERY-EXISTS) but up to you. I wish `knex` just had a `.exists()`!

rhelmer (changes requested):
The SQL suggestion is up to you, the logger one I feel more strongly about since it's taking me some extra debugging to track down unexpected errors from these.

src/db/tables/subscriber_coupons.ts
mansaj: I rewrote it in a simpler way without using raw: https://github.com/mozilla/blurts-server/pull/4633/commits/adc67e8645c9cc49b674fbce25b4fdbeba8dd7bd

rhelmer (approved):

src/db/tables/subscriber_coupons.ts
mansaj: Good idea to rethrow. I'll handle that on the caller side

PR #4639, opened 2024-06-06
MNTOR-3273: increment sign in counter for subscriber

mansaj:
# References:

Jira: MNTOR-3273

# Description
Backend function to keep track of # of sign-ins

### DIFF
--- a/src/db/tables/subscribers.js
+++ b/src/db/tables/subscribers.js
@@ -516,9 +516,10 @@ async function deleteOnerepProfileId (subscriberId) {
  */
 // Not covered by tests; mostly side-effects. See test-coverage.md#mock-heavy
 /* c8 ignore start */
-async function incrementSignInCount (fxaId) {
+async function incrementSignInCountForEligibleFreeUser (fxaId) {
   return await knex('subscribers')
     .where('fxa_uid', fxaId)
+    .whereNotNull('onerep_profile_id')
     .increment("sign_in_count", 1)
 }
 /* c8 ignore stop */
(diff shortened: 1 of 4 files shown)


src/db/tables/subscribers.js
```diff
@@ -510,6 +510,33 @@ async function deleteOnerepProfileId (subscriberId) {
...
+/**
+ * @param {string} fxaId
+ */
+// Not covered by tests; mostly side-effects. See test-coverage.md#mock-heavy
+/* c8 ignore start */
+async function incrementSignInCount (fxaId) {
+  return await knex('subscribers')
+    .where('fxa_uid', fxaId)
+    .increment("sign_in_count", 1)
+}
```
flozia, line 523: I’m wondering if it might be enough to only keep track of the count for users with a OneRep profile. As we only need to rely on the count for Plus (eligible) users after their initial sign-up we could also only start counting from then onwards. Or does this make things unnecessarily complicated?

mansaj: Everyone who runs a free scan also has a OneRep profile no? To determine plus status, we'd actually need to fetch that from the fxa subscription array? 

Ah I thought we'd display this for everyone. Let me take a look at the requirements again

flozia: Yes, users who run a free scan also have a OneRep profile. With “Plus (eligible)” users I mean free users that are in the US and can potentially upgrade to Plus. In order to make implementing the current requirements easier I think we only care about the login count after a user runs a scan. Theoretically, users could authenticate multiple times before reaching the dashboard if they don’t run the scan and I think we could ignore those?

flozia (approved):
Looks good to me. Thank you for adding this!

PR #4635, opened 2024-06-05
MNTOR-3057 - Cancel flow discount callout 

codemist:
# References:

Jira: MNTOR-3057
Figma: https://www.figma.com/design/DTbmXzCQCw2PxXpHQW8yfW/Monitor-MVP-Enhancements?node-id=1157-25611&t=CyXrO2sT6Vra9psg-0

# Description

The flags `CancellationFlow`, `ConfirmCancellation` and `DiscountCouponNextThreeMonths` are required. You'll also need the `CURRENT_COUPON_CODE_ID` env var set, which can be found in this ticket: https://mozilla-hub.atlassian.net/browse/MNTOR-3057?focusedCommentId=895742.

 With account that's subscribed to premium, go through the cancellation flow and the 30% off discount state should be visible. Selecting the CTA should output a `fxa_apply_coupon_code_success` in your server logs, and you should be taken to the `all-set` dialog state. 

# Screenshot (if applicable)

https://github.com/mozilla/blurts-server/assets/13066134/07a7601e-d9db-48d2-a056-ca7a4da4d093

# How to test

# Checklist (Definition of Done)

- [ ] Localization strings (if needed) have been added.
- [ ] Commits in this PR are minimal and [have descriptive commit messages](https://chris.beams.io/posts/git-commit/).
- [ ] I've added or updated the relevant sections in readme and/or code comments
- [ ] I've added a unit test to test for potential regressions of this bug.
- [ ] If this PR implements a feature flag or experimentation, the Ship Behind Feature Flag status in Jira has been set
- [ ] Product Owner accepted the User Story (demo of functionality completed) or waived the privilege.
- [ ] All acceptance criteria are met.
- [ ] Jira ticket has been updated (if needed) to match changes made during the development process.
- [ ] Jira ticket has been updated (if needed) with suggestions for QA when this PR is deployed to stage.

PR #4697, opened 2024-06-19
MNTOR-3246- Add error state and limitations apply page

codemist:
# References:

Jira: MNTOR-3246

# Screenshots

Error state:

image

# How to test (if applicable)

Follow the same flags/env setup as https://github.com/mozilla/blurts-server/pull/4635.

Selecting the "Limited time, restrictions apply" link should take you to `/limitations-apply` in a new tab:

image


locales-pending/limitations-apply-premium.ftl
```diff
@@ -2,8 +2,14 @@
...
+# $discount_percentage_num is the amount discounted in percentage per month
+# $discount_duration is the number of month(s) that users will pay the discounted price
+settings-limitations-apply-title = { $discount_percentage_num } off { $discount_duration } months promotion
+settings-limitations-pill = Rules & Restrictions
+settings-limitations-apply-description-one = Discount available for a limited time only.
+settings-limitations-apply-description-two = Redeemable 1 time only.
+# Variables:
+# $discount_percentage_num is the amount discounted in percentage per month
+# $discount_duration is the number of month(s) that users will pay the discounted price
+settings-limitations-apply-description-three = { -brand-mozilla-monitor } monthly subscribers will receive { $discount_percentage_num } off their next { $discount_duration } consecutive months billing.
```
flodolo, line 15:
```suggestion
settings-limitations-apply-description-three = 
    { $discount_duration ->
        [one] { -brand-mozilla-monitor } monthly subscribers will receive { $discount_percentage_num } off their next { $discount_duration } consecutive month billing.    
       *[other] { -brand-mozilla-monitor } monthly subscribers will receive { $discount_percentage_num } off their next { $discount_duration } consecutive months billing.    
    }
```

locales-pending/settings-premium.ftl
flodolo, line 1: Why do we need to repeat the same strings in two files?

locales-pending/limitations-apply-premium.ftl
```diff
@@ -2,8 +2,14 @@
...
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
-rules-and-restrictions-pill = Rules & Restrictions
 # Variables:
-# $discount_percentage_num is the amount discounted in percentage per month.
-# $discount_duration is the number of month(s) that users will pay the discounted price.
-limitations-apply-header = { $discount_percentage_num } off the next { $discount_duration } months
+# $discount_percentage_num is the amount discounted in percentage per month
+# $discount_duration is the number of month(s) that users will pay the discounted price
+settings-limitations-apply-title = { $discount_percentage_num } off { $discount_duration } months promotion
```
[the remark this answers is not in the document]
flodolo: Even if this is English only, and it will always be 1+ months, better to use the plural form in case we end up translating.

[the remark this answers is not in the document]
flodolo: Looks like that commit fixes the other string, not this one?

flodolo (approved):

locales-pending/limitations-apply-premium.ftl
```diff
@@ -2,8 +2,22 @@
 # License, v. 2.0. If a copy of the MPL was not distributed with this
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
-rules-and-restrictions-pill = Rules & Restrictions
 # Variables:
-# $discount_percentage_num is the amount discounted in percentage per month.
-# $discount_duration is the number of month(s) that users will pay the discounted price.
-limitations-apply-header = { $discount_percentage_num } off the next { $discount_duration } months
+# $discount_percentage_num is the amount discounted in percentage per month
```
Vinnl, line 6: The variable and comment seem to imply that this does not include the `%` sign - and maybe it shouldn't? I wouldn't be surprised if there are locales in which it's common to add a space in between the number and the percentage, for example.

Same for percentages in the other strings.

locales-pending/limitations-apply-premium.ftl
```diff
@@ -2,8 +2,22 @@
...
+        [one] { $discount_percentage_num } off { $discount_duration } month promotion
+       *[other] { $discount_percentage_num } off { $discount_duration } months promotion
+}
+settings-limitations-pill = Rules & Restrictions
+settings-limitations-apply-description-one = Discount available for a limited time only.
+settings-limitations-apply-description-two = Redeemable 1 time only.
+# Variables:
+# $discount_percentage_num is the amount discounted in percentage per month
+# $discount_duration is the number of month(s) that users will pay the discounted price
+settings-limitations-apply-description-three =
```
Vinnl, line 19: This might make more sense, given that it's not on the settings page?

```suggestion
limitations-apply-title =
    { $discount_duration ->
        [one] { $discount_percentage_num } off { $discount_duration } month promotion
       *[other] { $discount_percentage_num } off { $discount_duration } months promotion
}
limitations-pill = Rules & Restrictions
limitations-apply-description-one = Discount available for a limited time only.
limitations-apply-description-two = Redeemable 1 time only.
# Variables:
# $discount_percentage_num is the amount discounted in percentage per month
# $discount_duration is the number of month(s) that users will pay the discounted price
limitations-apply-description-three =
```

src/app/(proper_react)/(redesign)/(authenticated)/user/(dashboard)/settings/CancelFlow.module.scss
```diff
@@ -38,6 +38,36 @@
...
+    text-decoration-color: $color-grey-30;
+    .limitationsApplyText {
+      display: flex;
+      gap: $spacing-xs;
+      align-items: center;
+      justify-content: center;
+      text-decoration-color: $color-grey-30;
+
+      svg {
+        width: 13px; // width of open in new tab icon
```
Vinnl, line 55: Thanks for the comment!

src/app/components/client/TelemetryLink.tsx
```diff
@@ -19,6 +19,7 @@ export const TelemetryLink = ({
   target,
   upsell,
   showIcon,
+  className,
```
Vinnl, line 22: Passing classes on breaks the protections that we're hoping CSS modules to give us - which is that deleting or changing CSS in one part of the application breaks it elsewhere. Specifically, if we were to update the `<TelemetryLink>` styling, we'd have to not forget to also check that those updates interact well with the styles applied to it in `<CancelFlow>`.

Ideally, if we re-use a component, we just do not change the styling, but I get that that's sometimes a bit of a hassle. However, rather than potentially making people think that is an established pattern by adding an explicit prop, maybe we could just update the styling in `CancelFlow.module.scss` to target `a` elements inside a particular parent component? And instead of the `.hidden` class you could choose to not show the entire element.

```suggestion
```

Vinnl (approved):
Looks good and appears to do what it should do. I have a few ignorable nits for the l10n strings, and **one blocker**, which is not adding a `className` prop to `<TelemetryLink>` - I proposed an alternative inline.

locales-pending/limitations-apply-premium.ftl
flodolo: That's a valid point

```
            {l10n.getString("limitations-apply-title", {
              discount_percentage_num: "30%",
              discount_duration: 3,
            })}
```

Note that [JS provides ways to format that number](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat/NumberFormat#options)

PR #4725, opened 2024-06-26
Mntor 3031--Error: User is not eligible for feature

jespy2:
# References:

Jira: MNTOR-3031

# Description
Error: User is not eligible for feature
  File "monitor/src/app/api/v1/user/welcome-scan/create/route.ts", line 53, in POST
    throw new Error("User is not eligible for feature");


src/app/(proper_react)/(redesign)/(authenticated)/user/welcome/EnterInfo.tsx
```diff
@@ -46,7 +48,7 @@ const createProfileAndStartScan = async (
 
   const result: WelcomeScanBody = await response.json();
   if (!result?.success) {
-    throw new Error("Could not start scan");
+    logger.error("Could not start scan", { response, status: 403 });
```
rhelmer, line 51: I don't think logging the status code here is necessary. `response` is probably useful, I think it will be logged correctly since `winston` uses structured logging but we should confirm this in stage (e.g. that it doesn't just end up being `toString()`'d to `Object` or something silly)

rhelmer: I would suggest either returning a HTTP `403` to the user here, or redirect them somewhere. If the user isn't eligible we don't want to continue with the `EnterInfo` flow.

rhelmer (changes requested):

src/app/(proper_react)/(redesign)/(authenticated)/user/welcome/EnterInfo.tsx
jespy2: Updated here:
https://github.com/mozilla/blurts-server/pull/4725/commits/5f3f8585817a19c51c47fd6faa11d1775552de6e

src/app/(proper_react)/(redesign)/(authenticated)/user/welcome/EnterInfo.tsx
```diff
@@ -46,7 +50,9 @@ const createProfileAndStartScan = async (
 
   const result: WelcomeScanBody = await response.json();
   if (!result?.success) {
-    throw new Error("Could not start scan");
+    logger.error("Could not start scan", { response });
+    res?.status(403).json({ success: false });
+    return { success: false };
```
rhelmer, line 55: Here you could `return redirect("/")` (or wherever is appropriate). `/` is probably good because it will return to the landing page if there is an authentication problem, or automatically redirect to the dashboard if not.

The only thing to consider is if it's possible to cause a redirect loop, I don't think it should since this user is ineligible and should not have been able to make it here in the first place.

rhelmer (changes requested):

src/app/(proper_react)/(redesign)/(authenticated)/user/welcome/EnterInfo.tsx
```diff
@@ -30,11 +30,15 @@ import { CONST_URL_PRIVACY_POLICY } from "../../../../../../constants";
 
 import styles from "./EnterInfo.module.scss";
 import { TelemetryButton } from "../../../../../components/client/TelemetryButton";
+import winston from "winston";
+import { NextApiResponse } from "next";
```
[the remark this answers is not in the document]
jespy2: Updated here:
https://github.com/mozilla/blurts-server/pull/4725/commits/20fb12b952fc94b9869e8d038f49bfb4a636405e

src/app/(proper_react)/(redesign)/(authenticated)/user/welcome/EnterInfo.tsx
jespy2: Updated here:
https://github.com/mozilla/blurts-server/pull/4725/commits/20fb12b952fc94b9869e8d038f49bfb4a636405e

src/app/(proper_react)/(redesign)/(authenticated)/user/welcome/EnterInfo.tsx
```diff
@@ -46,7 +49,8 @@ const createProfileAndStartScan = async (
 
   const result: WelcomeScanBody = await response.json();
   if (!result?.success) {
-    throw new Error("Could not start scan");
+    logger.error("Could not start scan", { response });
```
rhelmer, line 52: Hm, I'm surprised that this works, `EnterInfo` is a client component right and `winston` is server (Node) right? I think it'd be better to use `Sentry.captureException`, you can get that with `import Sentry from "@sentry/nextjs"` directly in the component then you don't need to pass it as a prop.

I'm curious now if passing a server-side function as a prop like this does work, I *think* that props being passed from React server to client worked via [structured cloning](https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API/Structured_clone_algorithm) or similar.

Alternatively a server action could be used here, but I would just use Sentry it's much simpler and safer than trying to do it ourselves.

rhelmer (changes requested):
Changes lgtm, sorry I didn't notice the use of `winston` in a client component, I think Sentry would be a better fit here and simplify things (no need to pass as props etc)

src/app/(proper_react)/(redesign)/(authenticated)/user/welcome/EnterInfo.tsx
rhelmer: Just checked and https://react.dev/reference/rsc/use-server#serializable-parameters-and-return-values are the rules for what can be passed from server to client components.

src/app/(proper_react)/(redesign)/(authenticated)/user/welcome/EnterInfo.tsx
jespy2: Updated here:
https://github.com/mozilla/blurts-server/pull/4725/commits/e69a817bbc8f2c66908fb3bfc0aa8ea268686e1a

src/app/(proper_react)/(redesign)/(authenticated)/user/welcome/EnterInfo.tsx
jespy2: Updated here:
https://github.com/mozilla/blurts-server/pull/4725/commits/e69a817bbc8f2c66908fb3bfc0aa8ea268686e1a

jespy2: The above commit created breaking changes which are fixed here:
https://github.com/mozilla/blurts-server/pull/4725/commits/9b692572c67661956945f6b2b4cfe2c5b1634b21

jespy2: Updated to use NextJS import of Sentry instead of node.  Reverted changes in previous commit.
https://github.com/mozilla/blurts-server/pull/4725/commits/4861861de8badf0e92d0c812305dfe9cf5fa3992

rhelmer (approved):

PR #4732, opened 2024-06-29
Mntor-3286, Mntor-3285, Mntor-3321, Mntor-3287: Mocks for HIBP, OneRep + E2E test integration

mozilloid:
# References:

Jira: MNTOR-Mntor-3286, Mntor-3285, Mntor-3321, Mntor-3287
Figma:

# Description
- Added mock endpoints for OneRep and HIBP
- Modified the way that entries are stored in the database for mock data
- Added some path modifications to enable communication with mock endpoints
- Edited a few tests to work with local development
- Provided an endpoint to modify the test data used, but disabled it for now

# Screenshot (if applicable)

Not applicable.

# How to test


src/app/api/mock/hibp/data/fakeAllBreaches.json
```diff
@@ -0,0 +1,82 @@
+{
```
rhelmer, line 1: I think providing this mock data is fine for now, if you want to (or we could consider in the future) generating fake data with the right types e.g. the way https://github.com/mozilla/blurts-server/blob/974957027a73df5f0a3091e8d24a46282dd02e6b/src/apiMocks/mockData.ts works, which would provide a little more fuzziness.

src/app/api/mock/hibp/range/search/[hashPrefix]/route.ts
```diff
@@ -0,0 +1,35 @@
...
+  websites: string[];
+}[];
+
+export function GET() {
+  const prodError = errorIfProduction();
+  if (prodError) return prodError;
+
+  // Mock data for test email, can be randomized
+  const userEmail = process.env.E2E_TEST_ACCOUNT_EMAIL;
+  //TODO: getServerSession doesn't work here for some reason
```
Comment (rhelmer) [inline src/app/api/mock/hibp/range/search/[hashPrefix]/route.ts:23]: This would be expected if e.g. this is called by `fetch` without passing credentials such as a `Bearer` token.

src/app/api/mock/onerep/config/config.ts
```diff
@@ -0,0 +1,168 @@
...
+  type typeOfAddr = [{ city: string; state: StateAbbr }];
+
+  return MockUser.ADDRESSES.map((address) => ({
+    city: address.city,
+    state: address.state as StateAbbr,
+  })) as typeOfAddr;
+}
+
+const DEFAULT_NUMBER_BREACHES = 10;
+const MAGIC_NUM_brokerId = 78127;
```
rhelmer, line 74: Does this have a reason for mixing the case style (`UPPER_CASE` vs `camelCase`)?

src/app/api/mock/onerep/profiles/[profileId]/activate/route.ts
```diff
@@ -0,0 +1,21 @@
...
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import { NextRequest, NextResponse } from "next/server";
+import { errorIfProduction } from "../../../../utils/errorThrower";
+
+export function PUT(req: NextRequest) {
+  const prodError = errorIfProduction();
+  if (prodError) return prodError;
+
+  const profileId: number = Number(req.url.match(/profiles\/([0-9]+)/)![1]);
```
Comment (rhelmer) [inline src/app/api/mock/onerep/profiles/[profileId]/activate/route.ts:12]: Since you're using `[profileId]` in the file path, this is already parsed for you: https://nextjs.org/docs/app/building-your-application/routing/dynamic-routes

e.g. if you do:
```
export async function PUT(
  req: NextRequest,
  { params }: { params: { profileId: number } },
) 
```

Then `params.profileId` will already be a parsed `Number` and you don't need to do the extra work here.

src/app/api/mock/onerep/scan-results/route.ts
```diff
@@ -0,0 +1,19 @@
...
+
+import { errorIfProduction } from "../../utils/errorThrower";
+import { MOCK_ONEREP_BROKERS } from "../config/config";
+import { NextRequest, NextResponse } from "next/server";
+
+export function GET(req: NextRequest) {
+  const prodError = errorIfProduction();
+  if (prodError) return prodError;
+
+  const page = req.url.match(/page=([0-9]+)/)![1] || "1";
```
rhelmer, line 13: For HTTP params, there is a convenience method so you get a [URLSearchParams](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams) (similar to a `Map`) interface (https://nextjs.org/docs/app/api-reference/functions/next-request#nexturl), for example:
```
req.nextUrl.searchParams.get("page")
```

src/app/functions/server/onerep.ts
```diff
@@ -38,7 +38,7 @@ export type ShowProfileResponse = CreateProfileRequest & {
   status: "active" | "inactive";
   created_at: ISO8601DateString;
   updated_at: ISO8601DateString;
-  url: `https://api.onerep.com/profiles/${number}`;
+  url: `${string}/profiles/${number}`;
```
rhelmer, line 41: Does `${string}` actually work here, isn't this looking for the variable `string` which (I don't think) is set?

rhelmer: I would expect something like an environment variable or a string containing the host name.

src/app/functions/server/onerep.ts
```diff
@@ -118,6 +118,16 @@ async function onerepFetch(
   if (!onerepApiKey) {
     throw new Error("ONEREP_API_KEY env var not set");
   }
+
+  //If mock, remove the first slash so that it doesn't overwrite the path
```
rhelmer, line 122: I think this is OK, but it can be easier and safer to deal use the [URL interface](https://developer.mozilla.org/en-US/docs/Web/API/URL) (which is standard in both browser and Node JS environments) instead of manipulating URLs as strings.

rhelmer (changes requested):
lgtm overall, I tested it locally and seems to provide reasonable results. I have a few suggestions, the only real problem is I'm not sure the `${string}` in `onerep.ts` for the hostname evaluates to anything so that should be checked.

src/app/api/mock/hibp/data/fakeAllBreaches.json
rhelmer: I realize this makes testing more complicated, no need to do anything here but something to think about.

src/app/functions/server/onerep.ts
mozilloid: I tried to use it, but it requires that the path does not have a leading '/', o/w it overwrites the path in the ONEREP_API_BASE (so, locally it would overwrite to become localhost:port/path, instead of localhost:port/otherPath/path). But the real endpoint won't work w/o leading '/'. This is why I opted for a conditional removal of the slash.

src/app/functions/server/onerep.ts
mozilloid: Since it's in the definition of a type, and template literal types are statically resolved at compile time, I don't think I can place a variable in there. The previous URL links to the real website, but I wanted the URL to link to a mock profile. And string should be interpreted as a type, not a variable (I think). Let me know if its a mistake

src/app/api/mock/hibp/data/fakeAllBreaches.json
mozilloid: This sounds great! I looked over the code, I think a future integration with the existing methods there will be very beneficial.

PR #4748, opened 2024-07-02
Adopt .env.local file

Vinnl:
This makes it clearer what variables actually need to be set locally, and which have been forgotten. It also makes the build simpler, by removing the need to copy the .env-dist file.

This should be safe to apply, since .env-dist already got loaded by default, just like .env now is. And it is still the case that actual environment variables overwrite the ones in the .env file.

For non-Next.js setups (e.g. cron jobs or database migrations), I switched to dotenv-flow. The regular dotenv explicitly avoids inheritance [1], because it wants environment variables to be specific to an environment. That was already not the case with most of our environment variables, so the switch makes sense for us.

If we merge this, everybody will have to rename their `.env` to `.env.local`. No other changes should be required.

Next steps could be to remove unused variables from .env, and possibly moving variables with local/stage-specific values to .env.local.example, though that riskier, since environments might depend on those being present.

[1] https://www.npmjs.com/package/dotenv#should-i-have-multiple-env-files


mansaj:
Thanks for taking this on!!! LGTM, @Vinnl do you know how it would work in this scenario:
Let's say NODE_ENV is set as `production` and I don't have a `.env.production` in src. Will this automatically fall back to using `.env`? Is `.env` the catch-all?

Vinnl:
@mansaj Actual environment variables take precedence, so if `NODE_ENV` is set to `"production"`, none of the `.env` files matter. If `NODE_ENV` is not set, it will look for it in `.env.local`. If that doesn't define it either, it'll fall back to `.env`.

(And for _other_ environment variables, there's also `.env.production`, which comes after `.env.local`, but that's only loaded if `NODE_ENV` is set to `"production"`.)

jespy2 (approved):
Not super familiar with much this, but the `.env.local.example` file is super declarative and the way you have it organized, it reads almost like a newbie checklist.  This will make a huge difference in onboarding!  Great work!

flozia (approved):
Looks great to me.

rhelmer (approved):

rhelmer (approved):

mansaj:
> @mansaj Actual environment variables take precedence, so if `NODE_ENV` is set to `"production"`, none of the `.env` files matter. If `NODE_ENV` is not set, it will look for it in `.env.local`. If that doesn't define it either, it'll fall back to `.env`.
> 
> (And for _other_ environment variables, there's also `.env.production`, which comes after `.env.local`, but that's only loaded if `NODE_ENV` is set to `"production"`. But I don't think we'll use that anyway.)

hmm, yea so in my scenario.. There's only `.env` and `NODE_ENV` is set as `production`

then what?

mansaj (approved):

PR #4757, opened 2024-07-04
MTNOR-3240

jespy2:
# References:

Jira: MNTOR-3240

# Description
All breaches page not showing breaches in descending order by date.

# How to test
Navigate to /breaches.  Observe that breaches cards are now in descending order by date.


src/app/(proper_react)/(redesign)/(public)/breaches/page.tsx
```diff
@@ -30,6 +30,9 @@ export function generateMetadata() {
 
 export default async function Page() {
   const allBreaches = await getBreaches();
+  const breachesNewestFirst = allBreaches.sort((a, b) => {
+    return +new Date(b.BreachDate) - +new Date(a.BreachDate);
```
Vinnl, line 34: Could you add a comment what these pluses do here? It's not directly obvious to me. Is that a way to implicitly convert it to a number? If so, could we be more explicit, so people won't have to know what the type cast happens to do?

```suggestion
    return (new Date(b.BreachDate)).getTime() - (new Date(a.BreachDate)).getTime();
```

src/app/(proper_react)/(redesign)/(public)/breaches/page.tsx
```diff
@@ -30,6 +30,9 @@ export function generateMetadata() {
 
 export default async function Page() {
   const allBreaches = await getBreaches();
+  const breachesNewestFirst = allBreaches.sort((a, b) => {
```
Vinnl, line 33: Nit, but `.sort` modifies the array in place, so by re-assigning it to a new variable, people might be mislead that the original (`allBreaches`) is unmodified when it's not. `.toSorted` is the version that does not modify the original.

(It was only added recently, but since this runs on the server where we're running Node 20, that should be fine. Alternatively, you could just call `.sort` and not re-assign to `breachesNewestFirst`, but just continue working with `allBreaches`.)

Vinnl (approved):
Does what it says on the tin! I added two code style suggestions, but they're not blockers.

src/app/(proper_react)/(redesign)/(public)/breaches/page.tsx
Vinnl, line 34: Oh actually, the acceptance criteria aren't clear, but we might want to show the "newest" breaches in the sense of most recently added - sometimes they only get added years after they happen.

```suggestion
    return +new Date(b.AddedDate) - +new Date(a.AddedDate);
```

Vinnl (commented):
Might need clarification with Tony, but I think we might want to sort them by the date we added them.

src/app/(proper_react)/(redesign)/(public)/breaches/page.tsx
jespy2: Updated here:
[135736436744f0e19afac9a7b4b770ee29f084fc](https://github.com/mozilla/blurts-server/pull/4757/commits/135736436744f0e19afac9a7b4b770ee29f084fc)

src/app/(proper_react)/(redesign)/(public)/breaches/page.tsx
jespy2: Updated here:
https://github.com/mozilla/blurts-server/pull/4757/commits/46032bc138fa154892c610fc9e7fdc6c5ea99247

PR #4763, opened 2024-07-08
Mntor-3355: integration with e2e, support for email hash based mock-data retrieval

mozilloid:
# References:

Jira: MNTOR-3355, MNTOR-3140, MNTOR-3141
Figma:

# Description
Mock Endpoints finalized, with emails as environment variables based data retrieval.

# Screenshot (if applicable)

Not applicable.

# How to test


rhelmer (changes requested):
The stack trace is from minified source but I suspect the error you're seeing here is from: https://github.com/mozilla/blurts-server/blob/2c2c009bc4c44fb6bd0cded3641ceda7c8c714e5/src/utils/hibp.js#L68

The problem is probably that the test server isn't running on the IP address and/or port that is configured. The reason the error message is so confusing here is that:

1. it uses the old deprecated `InternalServerError` object, which in the pre-NextJS app (ExpressJS) translated these into HTTP errors automatically
2. it is attempted to localize the error message, which as a practice I think doesn't make much sense to do because the user should generally never see this
3. (2) is calling the deprecated `getMessage` l10n function but it's failing for some reason 

I'll make sure we have a ticket in the tech debt backlog to fix (1), for the purposes of this bug I'd suggest changing `InternalServerError(getMessage(...)` in `src/utils/hibp.js` to just return `InternalServerError(...)` and see if that gives you better error messages.

We should also move away from `InternalServerError` altogether but I see it is used a ton in the oldest pre-TS/Next code (`src/db` and `src/utis`) which is why I'm filing that in the backlog, feels like a bigger task to refactor all of that.

PR #4777, opened 2024-07-08
MNTOR-3353: add refresh token logic to session management

mansaj:
# References:

Jira: MNTOR-3353

# Description
- Adding refresh token fetching logic to next-auth
- The expiry / tokens are stored in the db

# Test

## Test for refresh token

- run `db:migration` and make sure the `fxa_session_expiry` col exists
- If the column is empty, login and see that the session expiry should be an hour in the future
- Change the time to be some time in the past OR set it to null 
- Go to dashboard, watch the session refresh, come back to the table and check on the session field again, it should have refreshed
- The access token should have also been refreshed

## Test for refresh fail

- empty out the columns `fxa_session_expiry`, `fxa_access_token`, and `fxa_refresh_token` columns in the `subscribers` table
- navigate to either dashboard or settings
- it should automatically log the user out and prompt a relogin


mansaj:
One thing I'm trying to figure out is where to add the client side of the session error to force re-sign in.
I'm thinking maybe in the dashboard (and settings), have a `useSession` and try to detect that error. If I see the error, then redirect?

src/app/api/utils/auth.ts
```diff
@@ -222,7 +226,43 @@ export const authOptions: AuthOptions = {
       }
       if (token.subscriber) {
         session.user.subscriber = token.subscriber;
+
+        // refresh token
+        const dbFxATokens = await getFxATokens(token.subscriber.id);
+        if (
+          !dbFxATokens.fxa_session_expiry ||
+          dbFxATokens.fxa_session_expiry.getTime() < Date.now()
+        ) {
```
pdehaan, line 235: I think the `getFxATokens()` can return `null` which might explode these if we're trying to access props on a null object.

src/db/tables/subscribers.js
```diff
@@ -148,6 +150,49 @@ async function updateFxAData (subscriber, fxaAccessToken, fxaRefreshToken, fxaPr
...
+    .update({
+      fxa_access_token: fxaAccessToken,
+      fxa_refresh_token: fxaRefreshToken,
+      fxa_session_expiry: new Date(sessionExpiredAt),
+      // @ts-ignore knex.fn.now() results in it being set to a date,
+      // even if it's not typed as a JS date object:
+      updated_at: knex.fn.now(),
+    })
+    .returning('*')
+  const updatedSubscriber = Array.isArray(updated) ? updated[0] : null
```
pdehaan, line 176: Is it possible that `updated` can be an empty array?
Although i think accessing `[0]` on an empty array just returns `undefined` and doesn't throw an error.

src/utils/fxa.js
```diff
@@ -73,6 +73,41 @@ async function revokeOAuthTokens(subscriber) {
...
+                  grant_type: "refresh_token",
+                  refresh_token: refreshToken,
+                  ttl: 604800, // request 7 days ttl
+                }),
+                method: "POST",
+              
+    })
+
+    const responseTokens = await postResp.json();
+    if (!postResp.ok) throw responseTokens;
```
pdehaan, line 99: What does this line do? What would `postResp.json()` return that would be `throw`able on a non-`ok` response?

src/utils/fxa.js
```diff
@@ -249,6 +284,7 @@ function getSha1(email) {
 }
 
 export {
+  refreshOauthTokens,
```
pdehaan, line 287: Nit: should be `*OAuth*` to match L288-289.

src/app/api/utils/auth.ts
Vinnl: Ah great eye, and that's another advantage of using `.first()` instead of `.select()` in that query - then TS would've warned us of this.

src/app/api/utils/auth.ts
```diff
@@ -222,7 +226,43 @@ export const authOptions: AuthOptions = {
       }
       if (token.subscriber) {
         session.user.subscriber = token.subscriber;
+
+        // refresh token
+        const dbFxATokens = await getFxATokens(token.subscriber.id);
+        if (
+          !dbFxATokens.fxa_session_expiry ||
+          dbFxATokens.fxa_session_expiry.getTime() < Date.now()
```
Vinnl, line 234: Maybe we should try rotating it if it expires soon, rather than when it has already expired? Since sessions have a lifetime as well in which the token can expire.

```suggestion
          dbFxATokens.fxa_session_expiry.getTime() < (Date.now() - 5 * 60 * 1000)
```

src/app/api/utils/auth.ts
```diff
@@ -209,7 +213,7 @@ export const authOptions: AuthOptions = {
       }
       return token;
     },
-    session({ session, token }) {
+    async session({ session, token }) {
```
Vinnl, line 216: Shouldn't refreshing the access token happen in the `jwt` callback? As I understand it, the session callback gets called pretty often, so hitting the database there can cause pretty heavy load. And the JWT would still contain the expired access token.

Vinnl (commented):
Could you add a "How to test" section? I'm unsure how to tell now whether this works as intended.

The most important thing that I think needs changing is doing this check when constructing the JWT, rather than when validating the session, since the latter happens pretty often.

src/utils/fxa.js
mansaj: I recently saw this as a code example on the next-auth doc. So basically it's a short hand for us to throw the correct error (.json() actually contains the error when the response isn't `ok`). In other places of our code, we'd do this whole dance of checking for `.ok` first and then do like a `.status` or `.text`. This is probably a way cleaner and better approach

src/app/api/utils/auth.ts
```diff
@@ -144,6 +147,7 @@ export const authOptions: AuthOptions = {
             profile.locale,
             account.access_token,
             account.refresh_token,
+            account.expires_at,
```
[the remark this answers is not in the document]
mansaj: The other line would catch it if this is null/undefined. I think it's a bit redundant to have that in two places

src/app/api/utils/auth.ts
mansaj: So there are two approaches to this. The database route is a bit heavy, but it's also a bit safer. It's outlined in the next-auth doc. 

We've traditionally opted for storing the sessions etc the database (prior to moving to next-auth). We were also using redis for session management. I think it should be an either-or situation

src/app/api/utils/auth.ts
mansaj: I simplified it a little bit so it uses first() and null check.

`first()` won't throw an error, it would just return `undefined` if it's not available

src/app/api/utils/auth.ts
mansaj: I don't know if we should artificially set a time (not really a best practice)
I think the point is to let sessions expire and potentially force a refresh or relogin

src/app/api/utils/auth.ts
Vinnl: Oh yeah, I'm not saying not going the database route, but doing it less often? I think the session gets refreshed about every minute when the user has a tab open, at least when an app is using the JWT session strategy - which Monitor does.

We could consider switching to the database session strategy, in which case I think the session callback will be called less often and doing the refresh there makes sense, but if we don't do that, I'd be afraid we'll be hammering our DB too much. But I'm not a back-end engineer, so I'll defer to your judgment :)

src/app/api/utils/auth.ts
Vinnl: I guess the question is: how do we do that? Are we going to refresh the session every time an API request to FxA fails with the relevant status code? Since this callback won't necessarily automatically get called when the session expires.

rhelmer: > I guess the question is: how do we do that? Are we going to refresh the session every time an API request to FxA fails with the relevant status code? Since this callback won't necessarily automatically get called when the session expires.

Yes I am wondering that too... I'm not sure what is going to trigger this `session` callback if e.g. a `fetch` to some API (like subscriptions details) fails.

rhelmer (changes requested):
Sounds like from talking this morning that you're going to investigate switching over to the [database strategy](https://authjs.dev/guides/refresh-token-rotation#database-strategy) - I think that will make things a lot clearer, since we're sort-of using both strategies right now.

src/db/tables/subscribers.js
mansaj: updated it with a length check

src/app/functions/server/applyCoupon.ts
```diff
@@ -67,8 +67,7 @@ export async function checkCurrentCouponCode(
   const currentCouponCode = process.env.CURRENT_COUPON_CODE_ID;
   if (!currentCouponCode) {
     logger.error(
-      "fxa_check_coupon_failed",
-      "Coupon code ID is not set. Please set the env var: CURRENT_COUPON_CODE_ID",
+      "fxa_check_coupon_failed: Coupon code ID is not set. Please set the env var: CURRENT_COUPON_CODE_ID",
```
rhelmer, line 70: Do you know what this looks like in the logging console? Is it possible to still look these up with `jsonPayload.message='fxa_check_coupon_failed'` or does the second argument need to be an object?

src/app/functions/server/checkSession.ts
```diff
@@ -18,6 +18,9 @@ export function checkSession(session: Session | null) {
   } else if (!session.user.subscriber.id) {
     logger.warn("no_subscriber_id_in_session", { session });
     return false;
+  } else if (session.error === "RefreshAccessTokenError") {
+    console.error("Refreshing access token failed... require login again");
```
rhelmer, line 22: Why `console` and not `logger` here? I think structured logging is better in general in newer code unless we can't for some reason (e.g. it doesn't support JS using it directly)

src/utils/fxa.js
```diff
@@ -88,13 +123,12 @@ async function getSubscriptions(bearerToken) {
...
+    const resp = getResp.json()
+    if (!getResp.ok) throw resp;
 
-    if (!getResp.ok) {
-      throw new InternalServerError(`bad response: ${getResp.status}`)
-    } else {
-      console.info(`get_fxa_subscriptions: success`)
-      return await getResp.json()
-    }
+    console.info(`get_fxa_subscriptions: success`)
```
rhelmer, line 129: 💯 FYI I filed a ticket in backlog to find and fix all of these (we have a ton of call sites in old JS code that look like this), but I am in favor of fixing when it's in the way too.

rhelmer (changes requested):

src/app/functions/server/checkSession.ts
mansaj: Copy pasted, will fix

src/app/functions/server/applyCoupon.ts
mansaj: Yea let me rework this

mansaj: I think the second arg needs to be an obj, it wasn't showing. That's why I appended it to the first arg

mansaj:
@rhelmer one thing we'd want to do in stage is to load test this for database connection (we will be checking db pretty frequently with the session changes). I'm optimistic that most of it should just be `get` and not `update` but it'd be nice to see

rhelmer (approved):

PR #4782, opened 2024-07-10
Make sure experiment code exists during tests

Vinnl:
It took quite a bit of tracking down when all my tests failed due to my not having generated the Nimbus experiment types. Since the code generation happens basically instantly, it seems fine to just run it every time before running the tests.


flozia (approved):
Great addition, thanks.

PR #4789, opened 2024-07-10
Convert experiment automatic-removal-csat-survey to feature flag

flozia:
# Description

Converts the experiment `automatic-removal-csat-survey` to use the feature flag `AutomaticRemovalCsatSurvey` instead.


rhelmer (approved):
lgtm, did you want to run this as an experiment though? If the goal is just to keep it behind a flag so we keep `main` shippable that's totally fine.

If you want to do some kind of a/b test or do a phased roll-out then that's where using Nimbus would be valuable.

PR #4790, opened 2024-07-10
Allow overriding Nimbus features locally, and clone before modifying experiment data

rhelmer:
# References:

Jira: MNTOR-3338

# Description

This is a small follow-up after finding bugs testing https://github.com/mozilla/blurts-server/commit/d73950a7d3fe5dc564429a335a5ec80d30a148aa

# How to test

This now works locally, by cloning/modifying/returning a clone of `localExperimentData`. I ended up cloning this because I noticed during testing that changing `localExperimentData` (since is an exported data structure from a module) will be cached for the lifetime of that Node process, which we don't want.

So to test, all you should need to do is have `?nimbus_web_preview=true` set when loading a page, and all features should be enabled. 


src/app/functions/server/getExperiments.ts
```diff
@@ -28,9 +28,22 @@ export async function getExperiments(params: {
...
+  // TODO MNTOR-3380 - until Cirrus implements preview mode, set all Nimbus features to `true` for QA purposes.
+  if (params.previewMode === true) {
+    // Clone the `localExperimentData` object so we don't modify the exported data structure.
+    const overriddenExperimentData = localExperimentData;
+    Object.assign(overriddenExperimentData, localExperimentData);
+    Object.keys(overriddenExperimentData).forEach((key) => {
+      const feature = overriddenExperimentData[key as keyof ExperimentData];
+      feature["enabled"] = true;
+    });
+    return overriddenExperimentData;
```
Vinnl, line 40: I'm not sure how to reproduce the problem you're describing in the PR, but I don't think this does what you want, as the properties still reference the same objects, i.e. even though `overriddenExperimentData !== localExperimentData`, it's still the case that `overriddenExperimentData["last-scan-date"] === localExperimentData["last-scan-date"]`. I think you want something like this:

```suggestion
    const overriddenExperimentData = Object.fromEntries(
      Object.entries(localExperimentData).map(
        ([experimentId, experimentConfig]) => {
          return [experimentId, { ...experimentConfig, enabled: true }];
        },
      ),
    ) as ExperimentData;
    return overriddenExperimentData;
```

Vinnl (commented):
I'm not sure whether I've correctly understood the point of this PR (and thus how to reproduce the problem), but see my inline comment and let me know if that makes sense.

src/app/functions/server/getExperiments.ts
rhelmer: Thanks! I think https://developer.mozilla.org/en-US/docs/Web/API/structuredClone would do the right thing as well.

The problem I noticed is that modifications to the `localExperimentData` export from the generated Nimbus module seem to be cached for the lifetime of the application, but if `Object.assign` is keeping any references to the old object then that's true it's not going to solve much.

Vinnl (approved):
Thanks for the clarification. I verified that if I disabled the `last-scan-date` experiment in my `nimbus.yml` on `main`, adding the query parameter didn't enable it. On this branch, it did.

PR #4791, opened 2024-07-10
Explicitly provide env file path

flozia:
We need to explicitly provide the [option.path](https://github.com/kerimdzhanov/dotenv-flow?tab=readme-ov-file#optionspath) in [appConstants.js](https://github.com/mozilla/blurts-server/blob/main/src/appConstants.js#L12) when running scripts like `npm run db:migrate` which change the `CWD`.


rhelmer (approved):
lgtm

Vinnl (approved):
Huh, weird that the original `dotenv` [didn't need this](https://github.com/mozilla/blurts-server/blob/b44826a17fd56e472573a7897d6ef02e50a1dc84/src/appConstants.js#L6-L7)!

PR #4794, opened 2024-07-12
Add element view telemetry (MNTOR-3358)

flozia:
# References:

Jira: [MNTOR-3358](https://mozilla-hub.atlassian.net/browse/MNTOR-3358)

# Description

Adds the hook `useViewTelementry`, which reports a telemetry event when an item is shown within the viewport for the first time. For this hook, a new Glean module `view` with the event name `enter` has been added.

# How to test

1. (Re)build Glean with `npm run build-glean`
2. Visit the landing page `/` and scroll through the page
3. Observe the telemetry events


src/telemetry/metrics.yaml
```diff
@@ -396,6 +396,40 @@ collapse:
         description: Which tier of plan the user is on [Free, Plus]
         type: string
 
+view:
```
flozia, line 399: @rhelmer We were talking about adding a `view` event to `ctaButton` at first. I’ve opted for adding `view` as a module that works better with the more generic `useViewTelemetry`. This makes it easier to observe other elements or types of view actions if we need to do so in the future.

Does this sound reasonable, and would we need to request a data review for this event type? The type of metrics we record is not different from the other event modules.

src/app/hooks/useViewTelemetry.ts
```diff
@@ -0,0 +1,32 @@
...
+  EventModule extends "view",
+  EventName extends keyof GleanMetricMap[EventModule],
+>(
+  args: TelemetryArgs & GleanMetricMap[EventModule][EventName],
+  options?: IntersectionOptions,
+) {
+  const { experimentationId, ...telemetryArgs } = args;
+  const recordTelemetry = useTelemetry({ experimentationId });
+  const { ref } = useInView({
+    skip: Object.keys(telemetryArgs).length === 0,
```
flozia, line 19: Make sure that the IntersectionObserver is not created if there is no telemetry to report.

src/app/(proper_react)/(redesign)/(public)/SignUpForm.tsx
```diff
@@ -32,6 +34,15 @@ export const SignUpForm = (props: Props) => {
   const l10n = useL10n();
   const [emailInput, setEmailInput] = useState("");
   const record = useTelemetry();
+  const { view } = props.eventId;
+  const refViewTelemetry = useViewTelemetry(
+    {
+      element_id: view,
+    },
+    {
+      skip: typeof view === "undefined",
```
flozia, line 43: The prop `view` can be `undefined` — skip the creation of `IntersectionObserver` in the `useViewTelemetry` hook.

src/telemetry/metrics.yaml
flozia: With our current setup of Glean modules and with respect to the already added `view` events I’ve [adjusted the metrics.yaml](https://github.com/mozilla/blurts-server/pull/4794/files#diff-b07fc1fd4cfa7508bc162a346e98d6c12ef6d3c30ec9bbda3c8e3fe1472b637aR432) to add a `view` event to `ctaButton`.

rhelmer (approved):

rhelmer (approved):
Thanks - this is what I expected, using intersection observer to see when the component actually comes into view. I haven't tested locally but code lgtm.

I would suggest using `Glean.setLogPings(true)` in the browser devtools console when this hits stage and make sure it's generating the pings you expect for all these views.

src/app/(proper_react)/(redesign)/(public)/FreeScanCta.tsx
```diff
@@ -88,12 +82,13 @@ export const FreeScanCta = (
   ) : (
     <div>
       <TelemetryButton
+        buttonRef={refViewTelemetry as unknown as RefObject<HTMLButtonElement>}
```
Vinnl, line 85: I wonder if it maybe would make sense to just add this type assertion (or I suppose `RefObject<any>`) to the return value of `useViewTelemetry`?

Vinnl: Or maybe, clicking through to the type definition of the `ref` prop on `<button>` and the type definition of react-intersection-observer, `RefCallback` would be even more accurate, but in practice, I don't think it matters in practice.

Vinnl (approved):

src/app/hooks/useViewTelemetry.ts
```diff
@@ -0,0 +1,38 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import { IntersectionOptions, useInView } from "react-intersection-observer";
```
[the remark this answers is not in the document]
flozia: Ah, yes! Installed `react-intersection-observer` as dependency in [3d29d1d](https://github.com/mozilla/blurts-server/pull/4794/commits/3d29d1de50acc06b4ccd56a45bc69429cb45e299).

src/app/(proper_react)/(redesign)/(public)/FreeScanCta.tsx
flozia: I moved the type assertion to the `useTelemetryView` view hook in [6618a55](https://github.com/mozilla/blurts-server/pull/4794/commits/6618a558dc6acc841886a6187cc795998b4f7eae). I’ve kept `RefObject`, because otherwise we would need to convert back to `unknown` again when using the return value of the hook.

PR #4814, opened 2024-07-16
Sort-of port update-remote-settings cron to TS

Vinnl:
# References:

Jira: MNTOR-2142
Figma: N/A

# Description

The request was to port it to TypeScript even though this code couldn't run yet, because we might be enabling this again soonish.

Thus, this ports it to TypeScript, recovers the module that it referenced but no longer existed, replaces that module's use of the `got` package with `fetch`, and ensures it builds. I don't have the right access to be able to verify that it actually runs as expected though.

Here's the original `lib/remote-settings.js` that the old `updatebreaches.js` referenced: https://github.com/mozilla/blurts-server/blob/99a6884d83c07eac9726d508d5b3d0f10ed70e41/lib/remote-settings.js (deleted in c727bfe968937e51b0cd42efefd010c7c401aeae)

# How to test

:shrug: 

CI shows that at least the build succeeds, so it compiles.


rhelmer (approved):
Thanks for the lengthy comment, I think this is fine for now.

PR #4821, opened 2024-07-17
MNTOR-null: fix yearly pricing default var

mansaj:
# References:

Jira: MNTOR-null

# Description
fix yearly pricing default var


mansaj:
@codemist This might be the reason you were seeing both yearly and monthly subscriptions in your local testing

PR #4822, opened 2024-07-17
Implement RP metrics for landing page CTA experiment (MNTOR-3341)

flozia:
# References:

Jira: [MNTOR-3341](https://mozilla-hub.atlassian.net/browse/MNTOR-3341)

# Description

Follow-up for PR #4787 that makes sure we are passing the correct parameters to the Mozilla accounts endpoint `/authorization` and is implemented as [suggested by data science](https://mozilla.slack.com/archives/C05TV26C5V4/p1721156802024629?thread_ts=1719862765.251169&cid=C05TV26C5V4).

**1. Fetch `/metrics-flow` endpoint**
The parameters documented [here](https://mozilla.github.io/ecosystem-platform/relying-parties/reference/metrics-for-relying-parties#relying-party-hosted-email-form), plus `service` set to Monitor's OAuth client ID.

**2. Mozilla accounts sign-in/up URL**
- `flow_id`, `flow_begin_time`, and `device_id` are being set dynamically based on the `flowId`, `flowBeginTime`, and `deviceId` values returned from the `/metrics-flow` call in step `1`.
- `entrypoint` set consistently
- `entrypoint_experiment` and `entrypoint_variation` set based on the active experiment (if any).
- UTM parameters set as appropriate (same as described above for the metrics-flow call).

# How to test

1. Enable the Nimbus experiment `landing-page-free-scan-cta`
4. Signing in/up with one of the “Free scan” CTA on the landing page.
5. The relevant URL parameter should be present on the Mozilla accounts auth page.

[MNTOR-3341]: https://mozilla-hub.atlassian.net/browse/MNTOR-3341?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/contextProviders/accounts-metrics-flow.tsx
```diff
@@ -0,0 +1,73 @@
...
+  enabled,
+  metricsFlowParams,
+}: SessionProviderProps) => {
+  const [data, setData] = useState<ContextValues["data"]>(null);
+  const [loading, setLoading] = useState<boolean>(enabled);
+  const searchParams = useSearchParams();
+
+  useEffect(() => {
+    async function fetchMetricsFlowData() {
+      const updatedSearchParams = new URLSearchParams(searchParams.toString());
```
sean-rose, line 44: It seems like this will be passing along any/all URL parameters from the Monitor page request to the Monitor API request and subsequently to the FxA `metrics-flow` request, which runs the risk of FxA balking at some random parameters.

IMO the only URL parameters it would make sense to pass along would be UTM parameters, and ideally the UTM parameters used here would be the same as those being passed to the FxA sign-in (e.g. using similar logic as `getFreeScanSearchParams` to look for any appropriate attribution data and fall back to default UTM parameter values if necessary).

src/app/(proper_react)/(redesign)/(public)/FreeScanCta.tsx
```diff
@@ -88,6 +59,7 @@ export const FreeScanCta = (
   ) : (
     <div>
       <TelemetryButton
+        disabled={metricsFlowContextData.loading}
```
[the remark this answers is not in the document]
flozia: That is a very good point! Let’s not take the risk of potentially blocking the button. I’ve adjusted the behaviour in [f527450](https://github.com/mozilla/blurts-server/pull/4822/commits/f527450c93f8d73f1ffd09fc4aea5888a9dbf81d).

src/contextProviders/accounts-metrics-flow.tsx
```diff
@@ -0,0 +1,73 @@
...
+      const response = await fetch(
+        `/api/v1/accounts-metrics-flow?${updatedSearchParams.toString()}`,
+      );
+      const data: {
+        success: boolean;
+        flowData?: MetricFlowData;
+      } = await response.json();
+
+      setData(data.flowData ?? null);
+      setLoading(false);
```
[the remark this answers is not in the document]
flozia: I’ve removed the `loading` state since we don’t need it anymore to disable the button. Nevertheless, I’ve added a `try...catch` to capture potential exceptions [here](https://github.com/mozilla/blurts-server/pull/4822/files#diff-da0b4522c3531b359c10d240016c86520424e66fa4834709ff747655f80fc2a9R43-R58).

src/app/api/v1/accounts-metrics-flow/route.ts
```diff
@@ -0,0 +1,43 @@
...
+  return data;
+}
+
+export async function GET(request: NextRequest) {
+  try {
+    const { searchParams } = new URL(request.url);
+    const flowData = await fetchMetricsFlowParams(searchParams);
+    return NextResponse.json({ success: true, flowData }, { status: 200 });
+  } catch (e) {
+    return NextResponse.json({ success: false }, { status: 500 });
```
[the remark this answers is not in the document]
flozia: Added reporting here makes sense, yes! We use Sentry and I’ve added logging in https://github.com/mozilla/blurts-server/pull/4822/commits/397891424dc69a0a838b3ae1361555a848521e2f.

src/contextProviders/accounts-metrics-flow.tsx
flozia: Makes sense! I’ve restricted the parameters to the ones that are being explicitly passed here in https://github.com/mozilla/blurts-server/pull/4822/commits/e1460963b05b4da7864f245e2687004140942971.

sean-rose: It looks like that will always pass `utm_source=product&utm_medium=monitor&utm_campaign=get_free_scan` to `metrics-flow`, which is reasonable, but that isn't consistent with what's being passed to the FxA sign-in (where it first looks for `cookies.attributionsFirstTouch` and then falls back to those default UTM values).

IMO it'd be best to be consistent in both places (either both always pass the preset `get_free_scan` UTM values, or both look for `cookies.attributionsFirstTouch`).

src/app/functions/universal/getFreeScanSearchParams.ts
```diff
@@ -0,0 +1,66 @@
...
+        flow_begin_time: metricsFlowData.flowBeginTime.toString(),
+      }),
+      ...(experimentData &&
+        experimentData["landing-page-free-scan-cta"].enabled && {
+          entrypoint_experiment: "landing-page-free-scan-cta",
+          entrypoint_variation:
+            experimentData["landing-page-free-scan-cta"].variant,
+        }),
+    },
+    FREE_SCAN_UTM_SEARCH_PARAMS,
```
sean-rose, line 62: It might be good to only pass the fallback UTM params when no UTM params are found in `cookies.attributionsFirstTouch`, because as currently written the fallback UTM params can potentially be mixed with UTM params present in `cookies.attributionsFirstTouch` if only some of the source/medium/campaign UTM params are in `cookies.attributionsFirstTouch` (it'll essentially fill in any gaps, potentially leading to unexpected results).

src/contextProviders/accounts-metrics-flow.tsx
flozia: Continuing to pass `cookies.attributionsFirstTouch` to the FxA sign-in and not mixing UTM parameters makes sense to me. I’ve adjusted the PR to only pass `utm_source=product&utm_medium=monitor&utm_campaign=get_free_scan` as a fallback if `cookies.attributionsFirstTouch` is not available in [830072b](https://github.com/mozilla/blurts-server/pull/4822/commits/830072bfd147224745e8eec06aa901291c675351).

src/app/(proper_react)/(redesign)/(public)/SignUpForm.tsx
```diff
@@ -35,16 +37,19 @@ export const SignUpForm = (props: Props) => {
   const [emailInput, setEmailInput] = useState("");
   const record = useTelemetry();
   const [cookies] = useCookies(["attributionsFirstTouch"]);
+  const metricsFlowContext = useContext(AccountsMetricsFlowContext);
```
rhelmer, line 40: :100: seems like a good use of context vs. prop drilling

rhelmer (approved):

sean-rose (approved):
This looks like it will work correctly, but I wasn't able to verify that empirically at the preview URL since the CTA experiment isn't enabled by default.  Assuming you've been testing this locally with the experiment enabled and have seen the `metrics-flow` calls being made with the returned flow values getting passed along to the FxA sign-in form, then this should be good to go.

flozia:
> This looks like it will work correctly, but I wasn't able to verify that empirically at the preview URL since the CTA experiment isn't enabled by default. Assuming you've been testing this locally with the experiment enabled and have seen the `metrics-flow` calls being made with the returned flow values getting passed along to the FxA sign-in form, then this should be good to go.

@sean-rose I’ve verified this working locally: I see the `metrics-flow` calls being made and the returned flow values are being passed to the FxA sign-in form. Thank you — I really appreciate your guidance on the initial requirements and implementation for this!

PR #4825, opened 2024-07-18
Re-add allowlist functionality for feature flags

Vinnl:
image


src/knex-tables.d.ts
```diff
@@ -47,8 +47,8 @@ declare module "knex/types/tables" {
     dependencies?: string[];
     allow_list?: string[];
     wait_list?: string[];
-    added_at?: Date;
-    modified_at?: Date;
+    created_at: Date;
+    modified_at: Date;
```
rhelmer, line 51: Not opposed to changing this, but do we need to have a db migration for this then?

rhelmer (changes requested):

src/knex-tables.d.ts
mansaj: No, the col is already called `created_at` in our db.

I'm not sure why it was `added_at` in this type file.. It was wrong

mansaj (approved):

rhelmer (approved):

PR #4855, opened 2024-07-23
Convert js to ts files in src/utils

codemist:
# References:

Jira: MNTOR-3101

# Description

This converts a few smaller functions in `src/utils` into Typescript, namely:

- state.ts
- error.ts
- emailAddress.ts
- recommendations.ts
- parse.ts
- formatDate.ts

### DIFF
--- a/src/app/functions/universal/user.ts
+++ b/src/app/functions/universal/user.ts
@@ -34,7 +34,7 @@ export function hasSetupOnerep(
 const USER_MIN_AGE = 13;
 export function meetsAgeRequirement(dateOfBirth: ISO8601DateString): boolean {
   const dateNow = new Date();
-  const dateBirth = new Date(dateOfBirth as string);
+  const dateBirth = new Date(dateOfBirth);
   const dateDelta = new Date(dateNow.valueOf() - dateBirth.valueOf());
   const unixStartDate = new Date(0);
   const age = dateDelta.getUTCFullYear() - unixStartDate.getUTCFullYear();


src/utils/formatDate.ts
```diff
@@ -2,19 +2,27 @@
...
 // TODO: Add unit test when changing this code:
 /* c8 ignore next 8 */
-function formatDate (date, locales) {
-  const jsDate = new Date(date)
+
+type DateTimeFormatOptions = {
+  year: "numeric";
+  month: "long";
+  day: "numeric";
+};
```
Vinnl, line 12: You don't have to define this type yourself. If you hover over the `Intl.DateTimeFormat` constructor (which is where you get the first error), you can see the type of its second argument, and then use that directly.

[Screencast from 2024-07-24 11-36-40.webm](https://github.com/user-attachments/assets/b623da33-5aee-4aaa-8b55-e71702836df0)

(Sometimes we'll be using a library that has a type definition for a function parameter, but doesn't export it, which is a bit of a pain. In that case, you could do something like `type DateTimeFormatOptions = Parameters<typeof Intl.DateTimeFormat>[1];` as a workaround.)

src/utils/parse.ts
```diff
@@ -0,0 +1,51 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+// Tests are already submitted in https://github.com/mozilla/blurts-server/pull/3359:
+/* c8 ignore start */
+
+type E164PhoneNumberString = string | null;
```
Vinnl, line 8: In the original `.js` file, you can see this [JSDoc comment](https://www.typescriptlang.org/docs/handbook/jsdoc-supported-types.html#typedef-callback-and-param):

```js
/**
 * @typedef {string} ISO8601DateString See https://en.wikipedia.org/wiki/ISO_8601
 * @typedef {`+${string}`} E164PhoneNumberString See https://en.wikipedia.org/wiki/E.164
 */
```

The TypeScript equivalent would be:

```ts
/** See https://en.wikipedia.org/wiki/ISO_8601 */
export type ISO8601DateString = string;
/**  See https://en.wikipedia.org/wiki/E.164 */
export type E164PhoneNumberString = `+${string}`;
```

These types were re-used elsewhere in the codebase, which is causing the current build error:

```
 ./src/app/api/v1/user/welcome-scan/create/route.ts:25:10
Type error: Module '"../../../../../../utils/parse"' has no exported member 'ISO8601DateString'.

  23 | import { setProfileDetails } from "../../../../../../db/tables/onerep_profiles";
  24 | import { StateAbbr } from "../../../../../../utils/states";
> 25 | import { ISO8601DateString } from "../../../../../../utils/parse";
     |          ^
  26 | import { getCountryCode } from "../../../../../functions/server/getCountryCode";
  27 | import { getExperimentationId } from "../../../../../functions/server/getExperimentationId";
  28 | import { getExperiments } from "../../../../../functions/server/getExperiments";
```

So you probably want to re-add those and use those types. (They're not particularly special, but they communicate to the reader that the functions in this file don't just take _any_ string, but strings formatted in a particular way.)

src/utils/recommendations.ts
```diff
@@ -0,0 +1,264 @@
...
+
+// `getAllPriorityDataClasses` contains hardly any logic and can probably do
+// with a refactoring too, given its unused arguments.
+/* c8 ignore start */
+
+// It's unclear why this function has unused parameters, but since they're
+// positional parameters, removing them risks breaking things:
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
+
+type PriorityDataClassArray = {
```
Vinnl, line 13: Nice types! Small nit: this doesn't look like an array, so maybe it should not be called an `-Array` :)

Vinnl (commented):
Added some tips inline, looking good so far!

src/utils/parse.ts
```diff
@@ -0,0 +1,53 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+// Tests are already submitted in https://github.com/mozilla/blurts-server/pull/3359:
```
rhelmer, line 5: Hm, does this mean that we can remove the `c8 ignore` comment below?

rhelmer (approved):
lgtm overall, just a question about one of the comments that mentions tests being added, I wonder if we could stop using `c8 ignore` in that file now.

src/utils/parse.ts
```diff
@@ -0,0 +1,53 @@
...
+/* c8 ignore start */
+
+/** See https://en.wikipedia.org/wiki/ISO_8601 */
+export type ISO8601DateString = string | undefined | null;
+/**  See https://en.wikipedia.org/wiki/E.164 */
+export type E164PhoneNumberString = `+${string}` | null;
```
Vinnl, line 11: It would be surprising for me if I'm using these types anywhere and they could suddenly also be `null` or `undefined`. If the following is causing errors elsewhere, then you probably want to use a type like `E164PhoneNumberString | null` _there_, rather than just adding it here and potentially breaking all code that references this type.

```suggestion
export type ISO8601DateString = string;
/**  See https://en.wikipedia.org/wiki/E.164 */
export type E164PhoneNumberString = `+${string}`;
```

PR #4856, opened 2024-07-23
Petition banner adjustments (MNTOR-3113)

flozia:
# References:

Jira: [MNTOR-3113](https://mozilla-hub.atlassian.net/browse/MNTOR-3113)

# Description

Record a `banner.view` event for the petition banner and show the CSAT survey directly after dismissal.

# How to test

Same steps as for PR https://github.com/mozilla/blurts-server/pull/4775.

### DIFF
--- a/src/app/(proper_react)/(redesign)/(authenticated)/user/(dashboard)/dashboard/Dashboard.test.tsx
+++ b/src/app/(proper_react)/(redesign)/(authenticated)/user/(dashboard)/dashboard/Dashboard.test.tsx
@@ -10,6 +10,7 @@ import {
   queryByRole,
   render,
   screen,
+  waitFor,
   within,
 } from "@testing-library/react";
 import { userEvent } from "@testing-library/user-event";
@@ -3754,7 +3755,7 @@ describe("CSAT survey banner", () => {
     expect(answerButton).not.toBeInTheDocument();
   });
 
-  it("displays the petition CSAT survey for users in the treatment branch after they interacted with the “Data privacy petition banner”", async () => {
+  it("displays the petition CSAT survey for users in the treatment branch after they clicked “No, thank you”", async () => {
     const user = userEvent.setup();
     const ComposedDashboard = composeStory(
       DashboardUsPremiumResolvedScanNoBreaches,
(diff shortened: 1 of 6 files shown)


src/app/(proper_react)/(redesign)/(authenticated)/user/(dashboard)/dashboard/Dashboard.test.tsx
```diff
@@ -3779,16 +3780,44 @@ describe("CSAT survey banner", () => {
...
+      />
+    );
+    render(<ComposedDashboardComponent />);
+
+    const dismissCta = screen.getByRole("link", {
+      name: "Sign petition",
+    });
+    await user.click(dismissCta);
+
+    await waitFor(() => {
```
Vinnl, line 3814: This `waitFor` surprised me - maybe deserves a comment about why it's necessary?

src/app/components/client/PetitionBanner.tsx
```diff
@@ -44,6 +51,14 @@ export const PetitionBanner = (props: { user: Session["user"] }) => {
           <TelemetryButton
             variant="primary"
             className={styles.signButton}
+            onPress={() => {
+              // In order to prevent the banner from being hidden before
+              // the click on the link button is getting registered: Delay the
+              // dismissal until the next event loop execution.
```
Vinnl, line 57: Great comment! (And I presume that that's the reason for the `waitFor` I called out :P)

src/telemetry/metrics.yaml
```diff
@@ -168,6 +168,39 @@ popup:
         description: Which tier of plan the user is on [Free, Plus]
         type: string
 
+banner:
```
Vinnl, line 171: I'm assuming this doesn't need another data review before merging?

Vinnl (approved):

src/telemetry/metrics.yaml
flozia: Since this event module is recording the same telemetry like for example `page` I think we were not sure about that. I’ll ask the Glean team so we have clarity on that before merging.

cc: @rhelmer

src/app/components/client/PetitionBanner.tsx
flozia: Yup! That’s the reason.

src/app/(proper_react)/(redesign)/(authenticated)/user/(dashboard)/dashboard/Dashboard.test.tsx
flozia: Good callout — I see how the use of `waitFor` is unexpected here at first. I’ve added a comment in [caf277b](https://github.com/mozilla/blurts-server/pull/4856/commits/caf277b2f5170338723f2c169f2d2f9cf2d8d42a).

src/telemetry/metrics.yaml
flozia: We are good to go with this event as the intent and use of `banner.view` is already covered by the original data review. Upon [feedback](https://matrix.to/#/!XDoIOuMgPLgJFSLqRP:mozilla.org/$akGWGnvckAaxwiYd9-7Te9wpIr1upGnmeM6nCxoJXV4?via=mozilla.org&via=matrix.org&via=t2bot.io) from a data steward on Matrix, I’ve added a [comment to the original data review](https://bugzilla.mozilla.org/show_bug.cgi?id=1823766#c15) as a future reference.

cc: @rhelmer

PR #4859, opened 2024-07-24
Restore top-level 404 pages

Vinnl:
We worked around not being able to add leading underscores to routes by adding a catch-all handler and then looking at the URL. However, this intercepts _all_ top-level URLs, and then prevents the 404 page from showing up there.

It turns out a different workaround is now posible too: using %5F instead of an underscore at the start. See
https://github.com/vercel/next.js/discussions/50629#discussioncomment-6473908 and https://nextjs.org/docs/app/building-your-application/routing/colocation#private-folders.


src/e2e/specs/landing.spec.ts
```diff
@@ -383,3 +383,12 @@ test.describe(`${process.env.E2E_TEST_ENV} - Verify the Landing Page Functionali
     expect(page.url()).toBe(successUrl);
   });
 });
+
+test("Verify that the 404 page shows up on non-existent pages @smoke", async ({
```
mansaj, line 387: thanks!

mansaj (approved):
Very cool

Vinnl:
> Unrelated, can we use https://nextjs.org/docs/app/building-your-application/configuring/absolute-imports-and-module-aliases also?

We can; the main downside IMHO is that only tools that understand our Next.js config will be able to understand them. For example, the GitHub code browsing interface might not be able to discover the definition of imported functions anymore. So not sure if that's worth the upside.

PR #4865, opened 2024-07-25
Migrate utils/hibp.js to TypeScript

Vinnl:
# References:

Jira: MNTOR-3101
Figma: N/A

# Description

This removes the `Breach` type in functions/universal/breaches, which was created when first introducing TypeScript and the flow of data was still unclear, but by now had overlap with other types and no clear provenance.

Instead, there are now three breach-related types, that represent where the data came from:

- `HibpGetBreachesResponse`: this is an array of breach elements as returned from the HIBP API, unprocessed. Properties are in PascalCase, so are a breach's data classes.
- `BreachRow`: this is a breach's data as stored in our database, along with some data we added to it, such as a favicon URL. Properties are snake_case, and data classes are lowercased and kebab-cased by the formatDataClassesArray function.
- `HibpLikeDbBreach`: this is a breach's data fetched from the database, but stored in an object meant to look like the ones in HibpGetBreachesResponse. In other words, it contains the same data as BreachRow (including lowercased, kebab-cased data classes), but on PascalCase properties.

The latter is somewhat of a historical artefact, because we used to try to load breaches from our database, then if our database didn't contain any breaches yet, fetch them live from the HIBP API and continue working with that.

We no longer do that: now, even after fetching them from the HIBP API, we do a new query to get them from the database and process them into HibpLikeDbBreach, so that we can assume a consisent data structure everywhere we work with breaches.

I also added a new `WritableDateColumns` type helper to `knex-tables.d.ts`, which takes a database row type, and marks `Date` columns as `Date | ISO8601DateString`. This is useful because, while SELECT queries using Knex.js will indeed return Dates, when inserting or updating, an ISO 8601 string works too.

# How to test

There shouldn't be any regressions. I made very few runtime adjustments, I'll go over the diff and call them out in a self-review.

- [ ] I've added a unit test to test for potential regressions of this bug. - Should be covered by the existing ones


src/app/functions/server/breachResolution.ts
```diff
@@ -98,7 +98,9 @@ function appendBreachResolutionChecklist(
   const { verifiedEmails } = userBreachData;
 
   for (const { breaches } of verifiedEmails) {
-    breaches.forEach((b: Breach) => {
+    // Old untyped code, adding type defitions now isn't worth the effort:
+    /* eslint-disable @typescript-eslint/no-explicit-any */
+    breaches.forEach((b: any) => {
```
Vinnl, line 103: `verifiedEmails` is of type `any`. I'm not sure where the data originally comes from (i.e. from the DB or the HIBP API) so I'm not sure what the appropriate type is, but the now-deleted `Breach` type wasn't particularly safer than `any`, since it was just a guess too.

src/app/functions/server/getBreaches.ts
```diff
@@ -5,40 +5,28 @@
...
   // if "breaches" table does not return results, fall back to HIBP request
   if (breaches?.length < 1) {
-    const breachesResponse = (await req("/breaches")) as Breach[];
-    logger.debug(
-      "loadBreachesIntoApp",
-      `loaded breaches from HIBP: ${breachesResponse.length}`,
-    );
-
-    for (const breach of breachesResponse) {
-      breach.DataClasses = formatDataClassesArray(breach.DataClasses);
```
Vinnl, line 36: I moved this formatting into the `upsertBreaches` function, since data classes should always be lowercased and snake-cased when inserted into the database, and that avoids having to create a weird intermediate type that is like the HIBP API response, except with that transformation applied.

src/app/functions/server/getBreaches.ts
```diff
@@ -5,40 +5,28 @@
...
-      breaches.push(breach);
-    }
+    const breachesResponse = await fetchHibpBreaches();
+    logger.debug(`loaded breaches from HIBP: ${breachesResponse.length}`);
 
     // sync the "breaches" table with the latest from HIBP
-    await upsertBreaches(breaches);
+    await upsertBreaches(breachesResponse);
+    breaches = await getAllBreachesFromDb();
```
Vinnl, line 29: This is one of the few changes to runtime behaviour - instead of `breaches` having the data we just fetched from the HIBP API, it's the data we inserted into the database, which had a couple of transformations applied to it.

This should not cause issues, since we _already_ first tried to fetch the data from the database. Only if no breaches had been loaded in the database yet, did we send a request to the HIBP API and use its return value.

src/scripts/cronjobs/emailBreachAlerts.ts
```diff
@@ -121,6 +121,16 @@ export async function poll(
 
     const { breachName, hashPrefix, hashSuffixes } = data;
     const breachAlert = getBreachByName(breaches, breachName);
+    // Check added to old code for type safety, but we've been assuming
+    // getBreachByName will always find a breach here without tests so far, so
+    // apparently that's been working well enough:
+    /* c8 ignore next 6 */
+    if (!breachAlert) {
```
Vinnl, line 128: Technically a change in runtime behaviour, but this would presumably have caused an error before as well, just without the log.

src/scripts/cronjobs/updateBreachesInRemoteSettings.ts
```diff
@@ -99,8 +102,8 @@ if (
 }
 
 (async () => {
-  const allHibpBreaches = (await HIBP.req("/breaches")) as { body: Breach[] };
-  const verifiedSiteBreaches = allHibpBreaches.body.filter((breach) => {
+  const allHibpBreaches = await HIBP.fetchHibpBreaches();
```
Vinnl, line 105: Note that I replaced the generic `req` (which just sent requests to the given endpoint of the HIBP API) with a specialised `fetchHibpBreaches`, which sends a request to `/breaches` specifically, and thus is able to return a typed response.

src/utils/hibp.ts
```diff
@@ -0,0 +1,443 @@
...
+  IsSensitive: boolean;
+  IsSpamList: boolean;
+  IsVerified: boolean;
+  ModifiedDate: ISO8601DateString;
+  PwnCount: number;
+}>;
+
+export async function fetchHibpBreaches(): Promise<HibpGetBreachesResponse> {
+  return (await hibpApiFetch("/breaches")) as HibpGetBreachesResponse;
+}
```
Vinnl, line 121: Looks like the changes were too hard for Git to reconcile across the `.js` and `.ts` file, so let me try to point them out.

Here I renamed `req` to `hibpApiFetch` and made it internal to this module. As called out before, I added a `fetchHibpBreaches` wrapper that specifically calls out to HIBP's `/breaches` API and types the return value.

src/utils/hibp.ts
```diff
@@ -0,0 +1,443 @@
...
+ * HIBP, except that we added a `FaviconUrl` and ID, that AddedDate,
+ * ModifiedDate and BreachDate are Date objects, and that a couple of fields are
+ * not available (because we do not store them in our database, at the time of
+ * writing).
+ *
+ * Presumably, our code used to be set up in such a way that it was supposed to
+ * work with objects that were either returned from the HIBP API, _or_ from our
+ * database - but the latter doesn't capitalise the properties.
+ */
+export type HibpLikeDbBreach = {
```
Vinnl, line 179: Instead of basing this type on the HIBP response type, I explicitly based it on the types of the database row - after all, that's where we're getting these values from.

(That said, if you look at `knex-tables.d.ts`, you'll see that _those_ types are based on the HIBP response, if we don't do any further processing of them.)

src/utils/hibp.js
```diff
@@ -1,424 +0,0 @@
...
-  }
-  // This will be replaced by a map with the breach logos when
-  // `downloadBreachIcons` resolves, but by setting it to an empty Map first,
-  // we don't delay the server start - we just won't have breach logos yet.
-  app.locals.breachLogoMap = new Map()
-  app.locals.breaches = breaches
-  app.locals.breachesLoadedDateTime = Date.now()
-  console.info('done-loading-breaches', 'great success 👍')
-}
-/* c8 ignore stop */
```
Vinnl, line 227: This function wasn't used anywhere, so I deleted it.

src/app/functions/server/breachResolution.ts
rhelmer: Looks like it's from https://github.com/mozilla/blurts-server/blob/319434001d2110ceaf32e4dc259d9c793f12ce9b/src/app/functions/server/getUserBreaches.ts#L35 which has a note that we should deprecate it - @mansaj is that actionable now?

src/app/functions/server/getBreaches.ts
```diff
@@ -5,40 +5,28 @@
...
+export async function getBreaches(): Promise<HibpLikeDbBreach[]> {
   if (breaches) {
     return breaches;
   }
   breaches = await getAllBreachesFromDb();
-  logger.debug(
-    "loadBreachesIntoApp",
-    `loaded breaches from database: ${breaches.length}`,
-  );
+  logger.debug(`loaded breaches from database: ${breaches.length}`);
```
rhelmer, line 20: It'll be easier to find in the logs if you use it like:

```
logger.debug("loaded breaches from database", { breachesLength: breaches.length });
```

`loaded breaches from database` will be a key that is indexed and fast to search in the log console.

src/db/tables/breaches.js
```diff
@@ -60,8 +61,8 @@ async function upsertBreaches(hibpBreaches) {
           modified_date: breach.ModifiedDate,
           pwn_count: breach.PwnCount,
           description: breach.Description,
-          logo_path: breach.LogoPath,
-          data_classes: breach.DataClasses,
+          logo_path: /** @type {RegExpExecArray} */(/[^/]*$/.exec(breach.LogoPath))[0],
```
rhelmer, line 64: Do you know why this regex is necessary?

rhelmer (approved):
lgtm overall, just a small suggestion about `logger`.

One thing I'm a little concerned about is having to build and maintain complex object/relational mapping on our own, I think if we start finding `knex-tables.d.ts` to be too much to maintain over time we might want to consider switching to something that can automate the Postgres schema <-> TypeScript definitions for us. This feels like a place where it's easy to make little mistakes that will only be noticeable at runtime.

src/app/functions/server/breachResolution.ts
mansaj: `verifiedEmails` should come from HIBP API

mansaj: @rhelmer yes it's actionable. We should work on MNTOR-2021 when we have a chance. 

If I remember correctly, we have another function called `getSubBreaches` or something that basically replaces it

src/app/functions/server/getBreaches.ts
mansaj: The convention is that we now use snake case for the first param, so probably something like
`loaded_breaches_from_database`

mansaj (approved):

src/app/functions/server/breachResolution.ts
Vinnl: Thanks, looks like Kaitlyn was right and that this came from `getAllEmailsAndBreaches` from `breaches.js` - updated in d313659b2 (note that I did have to add one other `as any` - comment justifying that added inline).

src/app/functions/server/getBreaches.ts
Vinnl: Heh, I just removed the mention of the unused function, but fixed in 9e4191322.

src/db/tables/breaches.js
Vinnl: Unfortunately I don't; I moved that over from https://github.com/mozilla/blurts-server/blame/642be732f7c8f45c290c7407ade85a395388f113/src/utils/hibp.js#L115, but there's no clarification there.

PR #4866, opened 2024-07-25
[iOS] Wait until dismissing the petition banner for iOS (MNTOR-3432)

flozia:
# References:

Jira: [MNTOR-3432](https://mozilla-hub.atlassian.net/browse/MNTOR-3432)

# Description

Setting a timeout with `0` is not enough for iOS to delay the dismissal of the banner before the click on the link is registered. A lower value `100` should be plenty, but since this is an arbitrary value and undesirable workaround — let’s make extra sure there is enough time.

# How to test

Same steps as for PR https://github.com/mozilla/blurts-server/pull/4775.

[MNTOR-3432]: https://mozilla-hub.atlassian.net/browse/MNTOR-3432?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Vinnl (approved):
Ugh, I kinda suspect there must be a different way to get it to the back of the task queue on Safari, but I don't feel like reading [this](https://jakearchibald.com/2015/tasks-microtasks-queues-and-schedules/) again for this minor thing, so this is probably fine.

flozia:
> this

Thanks for the reference! I’ll look into this but will go ahead and merge to unblock [MNTOR-3113](https://mozilla-hub.atlassian.net/browse/MNTOR-3113).

PR #4892, opened 2024-07-30
MNTOR-3393: first pass of refactoring some util functions

mansaj:
# References:

Jira: MNTOR-3393

# Description
First pass of some refactoring:
* Get rid of InternalServiceError (see ticket)
* Refactored FxA and HIBP util


src/db/tables/emailAddresses.js
```diff
@@ -203,7 +202,7 @@ async function _addEmailHash (sha1, email, signupLanguage, verified = false) {
   } catch (e) {
     // @ts-ignore Log whatever, we don't care
     console.error(e)
-    throw new InternalServerError(getMessage('error-could-not-add-email'))
+    throw new Error(getMessage('error-could-not-add-email'))
```
rhelmer, line 205: Can we get rid of the `getMessage` here too?

src/utils/fxa.ts
```diff
@@ -47,34 +46,47 @@ async function destroyOAuthToken(
 
   try {
     const response = await fetch(tokenUrl, tokenOptions);
-    if (!response.ok)
-      throw new InternalServerError(`bad response: ${response.status}`);
-    await response.json();
-    return;
+    const responseJson = await response.json();
+    if (!response.ok) throw responseJson;
```
rhelmer, line 50: Hm, I think this may be a bit hard to debug from logs... wouldn't this throw `responseJson` (which is a `Response` object but not an `Error`) and then fail the `e instanceof Error` test below and this would end up returning `false` without any insight into why it failed?

src/utils/fxa.ts
```diff
@@ -131,16 +143,17 @@ async function refreshOAuthTokens(
...
       method: "POST",
     });
 
-    const responseTokens: FxaPostOauthTokenResponseSuccessRefreshToken =
-      await postResp.json();
-    if (!postResp.ok) throw responseTokens;
-    return responseTokens;
+    const responseJson: FxaPostOauthTokenResponseSuccessRefreshToken =
+      await response.json();
+    if (!response.ok) throw responseJson;
```
rhelmer, line 154: Same as above re: throwing `responseJson` here

rhelmer (changes requested):
This looks much better! I see a stray `getMessage` but the other pattern I noticed is that throwing `Response` objects that fail the `.ok` check probably aren't logging as much info as you want to be able to diagnose failures

src/utils/fxa.ts
Vinnl: You're right that it'll fail the `instanceof Error` check, but it's not a `Response` object - that's `response`. `responseJson` will be the parsed response body.

However, also note that if the error response body is _not_ valid JSON, that this will also throw an error, in which case we might want to move it until _after_ the `response.ok` check.

(It's also not clear to me why the original code called `await response.json()` in the first place, only to then discard the return value.)

mansaj: I first saw this pattern in the next-auth documentation, but now thinking more about it, perhaps we need some modifications. When failed, there are essentially 3 possible scenarios for line 49 `const responseJson = await response.json();`:

1. Failed call returns a valid JSON (fxa does this pretty well, so I'd say we should be able to expect this when dealing with fxa)
2. Empty response, error thrown for `can't parse` (hopefully not with fxa)
3. Some other response (html maybe), error thrown for `can't parse` (again, not likely with fxa)

It's a good point that `throw json` may not be an `Error` object. I think we should handle everything in the `catch()` below

mansaj:
@flodolo deleting a string here: `error-could-not-add-email` since we are no longer using it. 
Should I do it for all the files or just in `en`?

src/utils/fxa.ts
```diff
@@ -47,34 +46,47 @@ async function destroyOAuthToken(
...
-    if (!response.ok)
-      throw new InternalServerError(`bad response: ${response.status}`);
-    await response.json();
-    return;
+    const responseJson = await response.json();
+    if (!response.ok) throw new Error(responseJson);
+    console.info("destroy_oauth_token_success");
+    return true;
   } catch (e) {
     if (e instanceof Error) {
```
rhelmer, line 54: I noticed elsewhere that you re-threw if it wasn't an `Error`, do you think that's worth doing here? If it's not an instance of `Error` then it will just silently return `false` otherwise right?

rhelmer (changes requested):
lgtm overall but I'm unsure about one part of the error handling.

flodolo (approved):

src/utils/fxa.ts
mansaj: Changed. Though I don't think it's a big deal here. 
It depends on the context, for destroying token, failing silently and logging is probably what we want. We won't be throwing the error back to the user and ask users to do a different action

src/utils/fxa.ts
```diff
@@ -47,34 +46,47 @@ async function destroyOAuthToken(
...
-      console.error("destroyOAuthToken", { stack: e.stack });
+      console.error("destroy_oauth_token", { stack: e.stack });
```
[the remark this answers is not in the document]
mansaj: We want to change everything to `logger.error` but some files that depend on this file is still using `.js`

Once everything is migrated over to `ts` we will be using `logger`

src/db/tables/emailAddresses.js
```diff
@@ -203,7 +201,7 @@ async function _addEmailHash (sha1, email, signupLanguage, verified = false) {
   } catch (e) {
     // @ts-ignore Log whatever, we don't care
     console.error(e)
-    throw new InternalServerError(getMessage('error-could-not-add-email'))
+    throw new Error('error_could_not_add_email')
```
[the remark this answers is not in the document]
mansaj: Oh good catch, errors don't need to be formatted like this, only logs

rhelmer (approved):

PR #4916, opened 2024-08-05
Update docs on react-intersection-observer testing

Vinnl:
@flozia set you as reviewer - I think this is correct, but let me know if not.


flozia (approved):
Sounds good to me.

PR #4919, opened 2024-08-05
Update dependency doc with info on testing backend updates

rhelmer:
# Description

Backend dependency updates can be a little tricky to test locally in some cases, I called this out where appropriate.

I also landed a small load testing script that we've used for HIBP in the past which can be helpful to test PubSub.


src/loadtest/hibp.js
```diff
@@ -0,0 +1,50 @@
...
+ * NOTE: This is a k6 loadtest script, see https://k6.io/ for more information.
+ *
+ * In particular, k6 supports JS but is not a Node package: https://k6.io/docs/get-started/installation/
+ * This is why import/no-unresolved and no-undef are ignored.
+ */
+
+/* eslint-disable import/no-unresolved */
+/* eslint-disable no-undef */
+
+import http from "k6/http";
```
mansaj, line 15: Is there a doc for how to use this script?

rhelmer: It's pretty configurable at runtime, maybe best place to link might be https://grafana.com/docs/k6/latest/get-started/running-k6/#add-vus

rhelmer: Or just https://grafana.com/docs/k6/latest/get-started/running-k6

src/loadtest/hibp.js
Vinnl, line 1: Could we move this into `/src/scripts/`? Or maybe even `/src/scripts/one-offs/` or something like that? I think it would be good to collect scripts that you run locally occasionally in a single place that's separate from our main codebase. (I'd also like to move such scripts from `/scripts/` and `/src/scripts/` there, once I've figured out which ones we actually still use.)

src/loadtest/hibp.js
Vinnl: Do you often use different configurations, or is there a specific set of parameters you usually call it with? If the latter, would be useful to have an example in the docblock above, or maybe even an addition to `scripts` in `package.json`?

package.json
```diff
@@ -145,6 +145,7 @@
     "jest-canvas-mock": "^2.5.2",
     "jest-environment-jsdom": "^29.7.0",
     "jest-fail-on-console": "^3.3.0",
+    "k6": "^0.0.0",
```
Vinnl, line 148: Do we actually need this? npm says this is a

> Dummy package for autocompleting k6 scripts.

...so would I still need to install it separately? If so, what does this package bring us - shell autocompletion?

docs/dependency-updates.md
```diff
@@ -215,31 +215,66 @@ TODO: Describe how the `fxa-rp-events` endpoint uses these packages.
...
+- for email verification tokens in `src/db/tables/emailAddresses.js`
+
+1. Check that `nonce` is present in `content-security-policy` header for HTTP responses.
+2. Check that `experimentationId` is set as a cookie in the form `guest-${uuid}`
+3. Add a secondary email from the Settings screen and ensure that email verification works
 
 ### `@aws-sdk/*`
 
-TODO: Describe how to verify that AWS SDK updates didn't break anything.
+Used for S3, by the `src/scripts/s3.js` cron job.
```
Vinnl, line 231:
```suggestion
Used for S3, by the `npm run cron:db-pull-breaches` cron job.
```

Weird that `s3.js` is in `/src/scripts/`, since it seems to only export a utility function `uploadToS3` - I don't see it being run as a cronjob itself.

So then I suppose the way to verify this is to run `npm run dev:cron:db-pull-breaches` locally and check that it completes without errors?

Vinnl (approved):
Thanks, very helpful instructions! I think the only true blocker is that `/src/scripts/s3.js` doesn't actually run as a cronjob, as far as I can see, so we probably shouldn't say that - but that's an easy fix :)

src/loadtest/hibp.js
rhelmer: Adding this to `scripts` in `package.json` is a great idea - the main part that you generally want to configure at runtime is the number of virtual (concurrent) users.

package.json
rhelmer: On further inspection it looks like this package doesn't really do anything useful, I think `@types/k6` might be worth installing. I'll remove, thanks.

docs/dependency-updates.md
rhelmer: Hm yeah I think `s3.js` is just a utility library essentially, it should probably be moved to `src/utils` maybe...  thanks for suggestion, that sgtm.

src/loadtest/hibp.js
rhelmer: One problem with calling from `package.json` is that `k6` is a native (Go language) binary that needs to be installed separately, so it could be a little confusing. k6 uses Go to load this `hibp.js` script, which makes it unlike our other JS code that expects to run under browser and/or Node.

src/loadtest/hibp.js
rhelmer: I mentioned this in another comment, but one thing that's potentially confusing is that this JS file can only be loaded by `k6` which is a native binary (written in Go), it won't work under Node as you'd probably expect for other files in a `scripts` folder.

rhelmer: I added a section to the README on installation and running, and I moved this to `/src/scripts/loadtest/hibp.js`, wdyt?

mansaj (approved):

src/loadtest/hibp.js
Vinnl: Yeah fair enough, as long as there's an example command somewhere we can use as an example. I'd normally also ask for the longform version of CLI arguments, but given that the longform of `-u` is apparently `--vus`, that's not necessarily more informative :sweat_smile:

docs/dependency-updates.md
Vinnl: If I prefix `dev:` (to not have to compile), it does work. You might need these env vars though (also listed in `.env.local.example`):

```
AWS_ACCESS_KEY_ID=
AWS_SECRET_ACCESS_KEY=
```

PR #4922, opened 2024-08-06
Port the signup report email to the new template

Vinnl:
# References:

Jira: MNTOR-3449
Figma: N/A

# Description

Note that the old template includes code to render data from a variable `emailBreachStats`, but that variable is always unset. Thus, we're never rendering that anymore, so I didn't port it. For what was originally in that variable, see
https://github.com/mozilla/blurts-server/blob/870cca305e2e52ddfe7e0ea5ff5611a58537b70f/template-helpers/emails.js#L8-L33

# Screenshot (if applicable)

Before:

image

and without breaches:

image

After:

image

and without breaches:

image

# How to test

From easy to hard, aka from superficial to more thorough:

1. Open it [in Storybook](https://deploy-preview-4922--fx-monitor-storybook.netlify.app/?path=/story/emails-signup-report)
2. Send it to yourself via `/admin/emails`
3. Enable the `RedesignedEmails` feature flag. Create a new account with an email that has/has not been breached.

### DIFF
--- a/src/app/api/utils/auth.tsx
+++ b/src/app/api/utils/auth.tsx
@@ -30,6 +30,9 @@ import { getL10n } from "../../functions/l10n/serverComponents";
 import { OAuthConfig } from "next-auth/providers/oauth";
 import { SerializedSubscriber } from "../../../next-auth";
 import { record } from "../../functions/server/glean";
+import { getEnabledFeatureFlags } from "../../../db/tables/featureFlags";
+import { renderEmail } from "../../../emails/renderEmail";
+import { SignupReportEmail } from "../../../emails/templates/signupReport/SignupReportEmail";
 
 const fxaProviderConfig: OAuthConfig<FxaProfile> = {
   // As per https://mozilla.slack.com/archives/C4D36CAJW/p1683642497940629?thread_ts=1683642325.465929&cid=C4D36CAJW,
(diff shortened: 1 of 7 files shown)


src/emails/components/BreachCard.tsx
```diff
@@ -0,0 +1,93 @@
...
+            <span style={{ paddingInlineStart: "4px" }}>
+              {props.breach.Title}
+            </span>
+          </mj-text>
+        </mj-column>
+      </mj-section>
+      <mj-section padding="20px">
+        <mj-column>
+          <mj-text align="center" font-size="16px" padding="24px">
+            <p style={{ color: "#5e5e72" }}>
```
Vinnl, line 38: I moved this into its own component because the breach alert email and the signup report email shared it. The only change is that I added the `style` attribute here and on the `` in `<BreachLogo>`, instead of adding a class and a stylesheet to the head - because the component doesn't have access to the head.

src/app/api/utils/auth.tsx
```diff
@@ -193,8 +196,24 @@ export const authOptions: AuthOptions = {
...
+            enabledFlags.includes("RedesignedEmails")
+              ? renderEmail(
+                  <SignupReportEmail
+                    l10n={l10n}
+                    breachedEmailAddress={profile.email}
+                    breaches={unsafeBreachesForEmail}
+                  />,
+                )
+              : emailTemplate,
+          );
```
rhelmer, line 216: I like this approach, do you think that this flag should be checked in the `/admin/email` test tool as well? Or maybe that tool should allow choosing old/new email.

rhelmer (approved):

src/app/api/utils/auth.tsx
Vinnl: It would be neat, but the admin tool doesn't support rendering the old templates yet. And since QA has apparently already tested most of the other emails, I think we're actually pretty close to phasing out the old emails, so it's probably not worth effort of adding that :)

PR #4923, opened 2024-08-06
Use solid background colour for logo in dark mode

Vinnl:
Just a quick change based on [this Slack thread](https://mozilla.slack.com/archives/C0460QALW69/p1722873434592529?thread_ts=1722533360.709509&cid=C0460QALW69) in response to MNTOR-3476 where Tony and Zach indicated that they preferred having a solid background rather than an outline for the logo. This isn't usually visible for the email because it already has a white background, unless the email client forces dark mode and sets a dark background.

I also added a story to the monthly activity email storybook to be able to more easily check what it looks like. (But note that in the preview deployment, email image sources point to the stage environment, so you won't see the new image in there before we merge this.)


src/emails/templates/monthlyActivity/MonthlyActivityEmail.stories.tsx
```diff
@@ -10,20 +10,31 @@ import { SanitizedSubscriberRow } from "../../../app/functions/server/sanitize";
...
+  component: (props: StoryProps) => (
+    <>
+      {props.emulateDarkMode && (
+        <style>{`
+        * {
+          background-color: #1e293b !important;
+          color: white !important;
+        }
+      `}</style>
+      )}
```
flozia, line 25: Question (non-blocking): Would it make sense to move this into `StorybookEmailRenderer` so the prop `emulateDarkMode` can be used in all stories without duplicating the styles?

flozia (approved):

src/emails/templates/monthlyActivity/MonthlyActivityEmail.stories.tsx
Vinnl: That's an excellent idea. Stories will still have to pass on the prop, _but_ I also added support for the `prefers-color-scheme` media query, so you can now set your browser to dark mode for every email story. Added in 24fe12f49.

flozia: Awesome, thanks!

flozia (approved):

src/emails/templates/monthlyActivity/MonthlyActivityEmail.stories.tsx
Vinnl: Haha, and of course now it's complaining that this isn't covered by tests :sweat_smile:

PR #4932, opened 2024-08-08
Removed old email code and RedesignedEmails flag

Vinnl:
# References:

Jira: MNTOR-3453 and MNTOR-3452
Figma:

# Description

I searched for references to `RedesignedEmails`, removed it and the code that activated when it was disabled, code that was only referenced by such code (e.g. the old `fluent.js`), and a couple of strings that were only referenced by that code or other unused code.

This was me today (minus the politics):

image

# How to test

Nothing should have regressed, compared to having the flag enabled.


flodolo (approved):

src/app/api/utils/email.tsx
```diff
@@ -25,32 +23,34 @@ export async function sendVerificationEmail(
...
-    heading: "email-verify-heading",
-    subheading: "email-verify-subhead",
-    partial: { name: "verify" },
-  };
 
-  const enabledFlags = await getEnabledFeatureFlags({
-    email: user.primary_email,
-  });
+  if (!unverifiedEmailAddressRecord.verification_token) {
+    throw new BadRequestError("subscriber has no verification_token");
```
rhelmer, line 28: Not necessarily a blocker for landing this, but I think these custom `Error`s in the past automatically returned as the matching HTTP status code in the old ExpressJS app, but I think they have been logged as rather mysterious-looking errors since we switched from NextJS because the callers don't always handle these consistently.

Would you mind causing an error to throw here to see what it actually looks like when logged, and if the caller is handling it and returning the correct HTTP status code? It might be better to return a plain `Error` and move away from these custom error types but lmk what you think.

rhelmer (changes requested):
This looks great, thanks for finding and removing all of that old code! I have one question about our custom error types, I'm not sure they are working the way it was originally intended.

mansaj:
This is great @Vinnl ! Now we can change a lot of the logging to use `logger` instead of `console`

src/app/api/utils/email.tsx
mansaj: We can get rid of these error classes and just use `NextResponse`

Vinnl: OK, so exceptions are pretty hard to trace through the code, but I at least tracked down the usages of `BadRequestError`, because this is the only place that is thrown. The places that catch it are below.

In any case, I don't actually think this should be a 400 Bad Request, since the email address row that lacks the `verification_token` is coming from the database. We also can't just return a `NextResponse` from this function, since it's just a helper - the callers are using `NextResponse` already. I started reworking those to detect the specific error type, but I wasn't sure if that's actually where we wanted to go (vs. removing the special classes and just adding clearer logs), if there was even value in doing something differently, and given the number of classes in `error.ts`, it started to fail pretty out-of-scope for this PR.

https://github.com/mozilla/blurts-server/blob/a4a5c5bdd086addadbee0417d8f02c13916350fd/src/app/(proper_react)/(redesign)/(authenticated)/user/(dashboard)/settings/actions.ts#L120-L125

This is a server action, i.e. we don't have control over the status code, and wouldn't do anything with it anyway.

https://github.com/mozilla/blurts-server/blob/a4a5c5bdd086addadbee0417d8f02c13916350fd/src/app/api/v1/user/email/route.ts#L107

I'm not sure if this API is still used, but in any case, it catches all errors and returns a 500 error response.

https://github.com/mozilla/blurts-server/blob/a4a5c5bdd086addadbee0417d8f02c13916350fd/src/app/api/v1/user/resend-email/route.ts#L71

The API when adding a secondary email address that needs to be verified - also catches all errors and returns a 500.

rhelmer (approved):

PR #4936, opened 2024-08-09
MNTOR-3248: Stablize HIBP webhook

mansaj:
# References:

Jira: MNTOR-3487, MNTOR-3248

# Description
Resolving some issues related to scaling our postgres database.
We started noticing a lot of connection timeouts for Postgres when we get huge spikes of traffic coming from HIBP when big breaches occur. We (SRE and we) have implemented some workaround via PG Bouncer, but the issue still persist

In this PR, I'm attempting to address it via caching some frequently-used and rarely-changing big objects in Redis.
The `all breaches` object from HIBP is an obvious choice for now and it will be set in Redis the expiration is set to be `1 hour`. 
The appropriate fallback logic follows the below priority:
1. Redis
2. Postgres
3. HIBP API

There will additional PRs to set the object in Redis for the cronjob that fetches the `all breaches` object for Postgres

# Screenshot (if applicable)

Not applicable.

# How to test
* Without Redis, the code will automatically use a mock version of Redis (all the tests, or FE eng, can use this version)
* With Redis installed, set the `REDIS_URL` to something like `redis://localhost:your-port-of-choice` and you should be able to see the breaches object populated in Redis with the key `breaches`


src/app/api/v1/fxa-rp-events/route.ts
```diff
@@ -172,7 +173,7 @@ export async function POST(request: NextRequest) {
     const e = new Error(
       `could not find subscriber with fxa user id: ${fxaUserId}`,
     );
-    logger.error("fxaRpEvents", e);
+    logger.error("fxa_rp_event", { exception: e.message });
```
rhelmer, line 176: 💯 thanks for standardizing these!

src/utils/hibp.ts
```diff
@@ -211,13 +212,41 @@ export type HibpLikeDbBreach = {
...
+    // attempts to fetch breaches from Redis first
+    let redisBreaches;
+    if (rClient) {
+      redisBreaches = JSON.parse(
+        (await rClient.get(redisBreachKey)) || "[]",
+      ) as BreachRow[];
+    }
+
+    if (!redisBreaches || redisBreaches.length < 1) {
+      // if Redis fails, attempt to get breaches from Postgres and set Redis
```
rhelmer, line 228: I think this will happen more often than you expect, because there's a lag between the time the cron job runs that ingests these and the time that breach alerts come in.

I'm not sure what the best way to fix this is, we don't want to have a "thundering herd" problem where this code tries to refresh the cache.

Maybe it'd be best to return `429` if Redis is down, and not do the fallback to Postgres?

rhelmer (changes requested):

src/utils/hibp.ts
mansaj: I don't think this is that big of a deal. Without Redis, all the nodes currently do GET from Postgres. Without the fallback, users will not get the right breaches information here. The top priority for us is to provide the correct information to users.

I'm adding the logic to `set` Redis in another PR via the cronjob, this way the cache should be kept fresh, and it's unlikely this code triggers. But I think it's important to have this safeguard here regardless

mansaj: See the description of the PR for more info / clarification

rhelmer (changes requested):
Sorry I guess the one about logging the password is blocking, unless I'm misreading.

src/utils/hibp.ts
```diff
@@ -211,13 +212,41 @@ export type HibpLikeDbBreach = {
...
+  const rClient = redisClient();
+
   try {
-    dbBreaches = await getAllBreaches();
+    // attempts to fetch breaches from Redis first
+    let redisBreaches;
+    if (rClient) {
+      redisBreaches = JSON.parse(
+        (await rClient.get(redisBreachKey)) || "[]",
+      ) as BreachRow[];
```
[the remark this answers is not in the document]
mansaj: It's much easier to just store it as a string (treating it like an obj). 

Redis handles hash and arrays natively. However, if you look at the ioRedis doc, they provide `lpush` or `hset` but the `GET` is where it gets tricky. It assumes that you want to get a field from the hash or an element from the array. In our case, we simply want the whole object in/out.

src/db/redis/client.ts
```diff
@@ -0,0 +1,21 @@
...
+
+let singleton: Redis;
+
+export const redisClient = () => {
+  if (process.env.REDIS_URL?.includes("redis.mock")) {
+    singleton = createRedisMockInstance();
+    logger.info("redis_mock_client_created_success");
+  } else {
+    singleton = createRedisInstance();
+    logger.info("redis_client_created_success");
```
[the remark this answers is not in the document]
mansaj: changed to `debug`

rhelmer (approved):
lgtm

PR #4964, opened 2024-08-20
Align disabled state with email validation

Vinnl:
# References:

Jira: MNTOR-3496 and MNTOR-3479
Figma:

# Description

Previously, the submit button of the "Add another email address" form would be disabled if an email didn't match the pattern "<something>@<something>.<something>". However, form submit was left to be handled by client-side validation by the browser. Thus, it was possible to submit the form with an email address that was technically valid (e.g. "testtest@mailinator") even though the submit button was still disabled. Thus, this change makes the button's disabled state depending on the browser's form validation.

Note that I also added the check that some data was actually entered - the browser doesn't report on invalid form fields until the user has actually entered some data into that form. This is good UX, but the button can provide feedback earlier.

Also, you can now no longer submit the form if there's a space following the email address.

# How to test

Add a secondary email address without the TLD, and verify that the submit button turns blue and you're able to submit the form. If you add a space to the address, it's no longer possible.

- [ ] I've added a unit test to test for potential regressions of this bug. - No, since this delegates the validation to the browser, so mocking the browser behaviour doesn't leave us with much logic to test.


flozia (approved):
Great enhancement!

PR #4966, opened 2024-08-21
MNTOR-3465: HIBP responses with 429

mansaj:
# References:

Jira: MNTOR-3465

# Description
Respond with 429 instead of 500 for HIBP webhook


rhelmer (approved):

rhelmer (changes requested):
We probably do want to be notified if the topic is not found, should that one report to Sentry or do you think `logger.error` is sufficient? That one is unlikely to fix itself

mansaj:
> We probably do want to be notified if the topic is not found, should that one report to Sentry or do you think `logger.error` is sufficient? That one is unlikely to fix itself

added sentry, it is much easier than combing through the logs at this point.
Do you think 429 or 500 make more sense here

rhelmer (approved):

PR #4972, opened 2024-08-22
MNTOR-1800: Part 1 - new table for subscriber email preferences

mansaj:
# References:

Jira: MNTOR-3583

# Description
The work to migrate email preferences to a new table need to be done in a few steps:
1. create the table
2. handle migration of the data from old table (subscribers) to this new table.
3. Change the business logic to fetch from this new table / update new table


src/db/migrations/20240822032133_subscriber_email_preferences.js
```diff
@@ -0,0 +1,24 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+export async function up(knex) {
+  await knex.schema
+    .createTable('subscriber_email_preferences', function(table) {
+      table.increments('id').primary();
+      table.integer("subscriber_id").references("subscribers.id").notNullable().onDelete('CASCADE').onUpdate('CASCADE');
```
rhelmer, line 9: These always make me nervous 😅 I just double-checked and this should delete the row in `subscriber_email_preferences` when a row in `subscribers` is deleted, but not vice versa, that's the intent here correct? :)

rhelmer (approved):

src/db/migrations/20240822032133_subscriber_email_preferences.js
mansaj: Correct. The cascade is one way

src/knex-tables.d.ts
```diff
@@ -190,6 +190,19 @@ declare module "knex/types/tables" {
...
+    subscriber_id: number;
+    instant_breach_alert: boolean;
+    all_emails_to_primary: boolean;
+    monthly_monitor_report: boolean;
+    monthly_monitor_report_at: Date | null;
+  }
+  type SubscriberEmailPreferencesAutoInsertedColumns = Extract<
+    keyof SubscriberEmailPreferencesRow,
+    "id" | "subscriber_id"
+  >;
```
Vinnl, line 204:
```suggestion
```

Looking at the definitions below, it doesn't look like you're actually using this type? (Which looking back honestly is probably the more readable approach, just explicitly excluding columns (i.e. `id`) rather than adding indirection through another type.)

src/db/migrations/20240822032133_subscriber_email_preferences.js
```diff
@@ -0,0 +1,24 @@
...
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+export async function up(knex) {
+  await knex.schema
+    .createTable('subscriber_email_preferences', function(table) {
+      table.increments('id').primary();
+      table.integer("subscriber_id").references("subscribers.id").unique().notNullable().onDelete('CASCADE').onUpdate('CASCADE');
+      table.boolean('instant_breach_alert').defaultTo(true);
+      table.boolean('all_emails_to_primary').defaultTo(true);
+      table.boolean('monthly_monitor_report').defaultTo(true);
```
Vinnl, line 12: I recall reading (but can't find) the advice to instead make this a nullable datetime column `*_disabled_at`, would that make sense here? It's not a table that's read or written to particularly often I don't think, so not too much of a load, but would e.g. allow us to notice weird unsubscription patterns, and/or correlate unsubscription spikes with certain events? (Alternatively, simple Glean metrics might be fine too.)

Vinnl (approved):

PR #4982, opened 2024-08-26
Add new string for menu entry in the footer

Vinnl:
# References:

Jira: MNTOR-3545


flodolo (approved):
Should there be code using this string?

tumblr_efed1fa72ace34501a27bf9231101924_37b9f01c_640

Vinnl:
@flodolo I'm preparing that separately, but since we're deploying to prod pretty often, I figured I'd submit this first and give the localisers some time to translate it first.

PR #5012, opened 2024-09-03
Breach alert email redesign

Vinnl:
# References:

Jira: MNTOR-3433
Figma: https://www.figma.com/design/Z7pfHXJTSG5Szm3UrkYsVx/Emails-v2.0?node-id=1-697&m=dev

# Description

Implements a new design for the breach alert email. I still have [a couple of open question](https://mozilla.slack.com/archives/C0460QALW69/p1725273059741189?thread_ts=1724679600.596489&cid=C0460QALW69), but I've made some best guesses at what the answers will be.

I've assigned the whole team for review since it's both UI and backend changes.

# Screenshot (if applicable)

Non-US:

image

US free, no scan:

image

US free, scan has run:

image

Plus user:

image

# How to test

Most realistic is using `/admin/emails` to send a message to yourself - I've added a button there for the redesigned email specifically. Storybook has a bunch of the permutations.


public/images/email/footer-bg-shapes.png
Vinnl, line 1: Of note: theoretically we'll never be able to delete these images, since once sent, emails will keep referring them. However, these backgrounds have OK'ish fallbacks in plain colours, so removing them wouldn't be the end of the world.

src/app/(proper_react)/(redesign)/(authenticated)/user/(dashboard)/dashboard/fix/page.tsx
Vinnl, line 1: The changes in this file aren't really necessary; at first I thought I had to drop people into the guided flow, and didn't want to redirect to a specific step in the email in case the user resolves it. Thus, I updated plain `/fix/` (which we're not really using) to redirect to the most relevant step.

Turns out I didn't need that, but figured it didn't really harm anyway.

src/emails/templates/mjml.d.ts
```diff
@@ -186,13 +186,21 @@ declare namespace React.JSX {
         "background-position"?: CSSProperties["backgroundPosition"];
         "background-position-x"?: CSSProperties["backgroundPositionX"];
         "background-position-y"?: CSSProperties["backgroundPositionY"];
-        "background-repeat"?: CSSProperties["backgroundRepeat"];
+        "background-repeat"?: Extract<
```
Vinnl, line 189: MJML threw a couple of errors, such as apparently not supporting `repeat-y` for `background-repeat`. Presumably because some email clients don't support it...

src/scripts/cronjobs/emailBreachAlerts.tsx
```diff
@@ -260,7 +264,30 @@ export async function poll(
...
+                 */
+                const assumedCountryCode = getSignupLocaleCountry(recipient);
+
+                const scanData = await getLatestOnerepScanResults(
+                  recipient.onerep_profile_id,
+                );
+                const allSubscriberBreaches = await getSubscriberBreaches({
+                  fxaUid: recipient.fxa_uid,
+                  countryCode: assumedCountryCode,
+                });
```
Vinnl, line 290: @rhelmer @mansaj I wanted to draw some extra attention to these - I'm not sure how many subscribers we might be sending emails for in a single poll, but the email can now show the number of exposed data points for the user, which means I have to fetch all scan results and breaches for each of them. Just wondering if this won't cause excessive load and, if so, how best to deal with that?

locales/en/email-strings.ftl
```diff
@@ -2,10 +2,19 @@
...
+
+email-header-logo-alt = { -brand-mozilla-monitor }
+email-header-button-sign-in = Sign in
+
+## Email footers
+
+email-footer-support-heading = Questions about { -brand-mozilla-monitor }?
+email-footer-support-content = Visit our <support-link>Support Center</support-link> for help
+email-footer-reason-subscriber = You’re receiving this automated email as a subscriber of { -brand-mozilla-monitor }. If you received it in error, no action is required. For more information, please visit <support-link>{ -brand-mozilla } Support</support-link>.
+email-footer-reason-subscriber-one-time = You’ve received this one-time automated email because you are subscribed to { -brand-monitor-plus }. You won’t receive any further emails like this. For more information, please visit <support-link>{ -brand-mozilla } Support</support-link>.
```
flodolo, line 15: Should this string be exposed in the localizable file, since it talks about Monitor Plus?

flodolo (approved):

locales/en/email-strings.ftl
Vinnl: Heads-up: I updated one more string whose change in Figma I had missed (70bc71cdf), and added two new strings because the footer text changed (57a23294e). This new footer will be applied to every email at some point, so I opted to leave the previous one untranslated again to save the localisers some work.

src/scripts/cronjobs/emailBreachAlerts.tsx
flozia, line 290: Nit: We would only need to fetch these when `BreachEmailRedesign` is enabled as we don’t use the data for the current email template. That shouldn’t matter too much if we decide to flip the flag soon and keep it on, though.

src/emails/templates/EmailHero.tsx
```diff
@@ -0,0 +1,82 @@
...
+
+  return (
+    <mj-wrapper padding="24px 16px">
+      <mj-section
+        padding="10px 12px"
+        background-color="#e4d2ff"
+        background-url={`${process.env.SERVER_URL}/images/email/hero-bg-gradient.png`}
+        background-repeat="repeat"
+        background-position-x={0}
+        border-radius="16px 16px 0 0"
```
flozia, line 31: Nice! 🙌

src/emails/templates/breachAlert/BreachAlertEmail.tsx
```diff
@@ -55,3 +75,351 @@ export const BreachAlertEmail = (props: Props) => {
...
+              padding-right: 4px;
+              padding-inline-end: 4px;
+            }
+            dd {
+              display: inline;
+            }
+          `}
+        </mj-style>
+      </mj-head>
+      <mj-body>
```
flozia, line 139: One minor thing that I came across regarding the layout: From the designs, it looks like the highlighted text aligns with the text beneath as well as the buttons.

image

flozia (approved):
This is looking really good! Just left a few non-blocking comments inline.

src/app/functions/universal/user.ts
```diff
@@ -4,13 +4,21 @@
...
-/* c8 ignore start */
-export function hasPremium(user?: Session["user"]): boolean {
-  return user?.fxa?.subscriptions?.includes("monitor") ?? false;
+export function hasPremium(user?: Session["user"] | SubscriberRow): boolean {
+  const subscriptions =
+    // Simulating subscribers with incomplete FxA profile data
+    // is a bit too much effort for too little gain, hence:
+    /* c8 ignore next */
+    (user as SubscriberRow)?.fxa_profile_json?.subscriptions ??
+    (user as Session["user"])?.fxa?.subscriptions;
```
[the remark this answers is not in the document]
Vinnl: We won't have access to both here - it's just that in some places in the codebase, we'll have a user session, and in others (e.g. in cronjobs), we'll have fetched the user from the database - this change allows us to check whether they have Plus regardless of how we obtained data about the user.

src/emails/templates/EmailFooter.tsx
```diff
@@ -127,3 +127,117 @@ export const EmailFooter = (props: Props) => {
...
   );
 };
+
+export const RedesignedBreachEmailFooter = (props: Props) => {
+  const l10n = props.l10n;
+  const supportLinkUrlObject = new URL(CONST_URL_SUMO_MONITOR_SUPPORT_CENTER);
+  supportLinkUrlObject.searchParams.set("utm_medium", "email");
+  supportLinkUrlObject.searchParams.set("utm_source", "monitor-product");
+  supportLinkUrlObject.searchParams.set("utm_campaign", props.utm_campaign);
+  supportLinkUrlObject.searchParams.set("utm_content", "support-center");
```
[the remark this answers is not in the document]
Vinnl: Ah! Right, I keep misunderstanding how that function works. However, it looks like it also needs a separate `searchParams` object, in which case it wouldn't actually get shorter while adding a layer of indirection:

```suggestion
  const supportLinkUrlObject = new URL(CONST_URL_SUMO_MONITOR_SUPPORT_CENTER);
  supportLinkUrlObject.search = "?" + modifyAttributionsForUrlSearchParams(
    supportLinkUrlObject.searchParams,
    {
      utm_source: "monitor-product",
      utm_campaign: props.utm_campaign,
      utm_content: "support-center",
    },
    {},
  ).toString();
```

I don't think that's more readable or error-proof, but if we already use it elsewhere and people are used to it, then maybe I should just align with that?

src/scripts/cronjobs/emailBreachAlerts.tsx
Vinnl: That's a good idea; if these are gated behind the flag too, then we can easily disable it again [if they cause issues](https://github.com/mozilla/blurts-server/pull/5012#discussion_r1741837930).

Changed in d117dcefdda13eb94d18744146c893014636fcfa.

src/emails/templates/breachAlert/BreachAlertEmail.tsx
Vinnl: I'm _almost_ positive that wasn't the case before :P But well-spotted, fixed in 278c2ba45348b3c28d8670952a5bcd6d6055e6ef. For some reason I had to also make the buttons wider to prevent them from going on two lines :/

(For a short while, working with hardcoded pixel values and all that is pretty nice, but then you quickly run into issues like these...)

src/emails/templates/EmailFooter.tsx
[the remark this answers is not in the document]
flozia: Either is fine. I don’t think it’s more error-proof, but I don’t dislike the idea of aligning how we are appending `searchParams`. The current function as is was mostly used for modifying query params, but also can be very well be used for just adding them — we might need to consider renaming or splitting the functions a bit. `modifyAttributionsForUrl` would work without taking `searchParams` as an argument.

src/scripts/cronjobs/emailBreachAlerts.tsx
flozia: Great, thanks!

rhelmer: Here's some background info, I'm going to add advice as a separate comment but in case this is useful:

It's a bit hard to predict how many subscribers we'll email for a single `poll()` invocation - there's some analysis of the existing HIBP anonymized dataset here https://blog.cloudflare.com/validating-leaked-passwords-with-k-anonymity/

The `receivedMessage` being processed here is the same hash prefix + suffixes described in that blog post. Our system additionally only sends if certain criteria are met, and also of course each hash must match an email in the Monitor database so in practice we send fewer than the average 478 (max 584) cited there.

However the possible number is quite large:
> it is worth noting that as the length of a SHA-1 hash is a total of 40 hexadecimal characters long and 5 characters is utilised by the Hash Prefix, the total number of possible hashes associated with a Hash Prefix is 16^{35} ≈ 1.39E42.

rhelmer: We run the jobs that process these messages in parallel, so it's OK if one takes a while, and it can be safely resumed if interrupted. So I wouldn't worry too much about the edge cases here, although we don't want to have too much of a delay.

If we're using cached data I wouldn't expect it to be a problem, but if we need to refresh it from the provider that might cause more of an issue. Now that it's easy for us to use Redis again I'd suggest we consider caching data, we could have Redis automatically expire data from the cache after a certain amount of time.

I'm not sure we'd get a very good hit rate on that cache though, the main problem I can foresee is if we're refreshing broker data (I'm assuming that's what you mean by "scan results"?)

src/emails/templates/EmailFooter.tsx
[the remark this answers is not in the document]
rhelmer: One approach that could make this simpler for callers would be to refactor the `src/app/functions/universal/attributions.ts` helpers to instead extend the `URL` and `URLSearchParams` interfaces, e.g. `export class AttributedURL extends URL` and encapsulate the functionality that the current `modifyAttributionsForUrl` and `modifyAttributionsForUrlSearchParams` helper functions provide into that class.

Then callers could do something like:
```
const supportLinkUrlObject = new AttributedURL(CONST_URL_SUMO_MONITOR_SUPPORT_CENTER);
```

And equivalent for `AttributedURLSearchParams` (although I think if `AttributedURL.searchParams` was `AttributedURLSearchParams` then it would probably not be necessary to export directly, callers could just modify an existing `AttributedURL`'s `searchParams` and get the behavior they want).

That would allow callers to use the same API but it'd be clear from the class name(s) that these are overridden. These are long-stable interfaces so it seems unlikely to break in some subtle way.

src/scripts/cronjobs/emailBreachAlerts.tsx
```diff
@@ -260,20 +267,62 @@ export async function poll(
...
+                if (enabledFeatureFlags.includes("BreachEmailRedesign")) {
+                  /**
+                   * Without an active user session, we don't know the user's country. This is
+                   * our best guess based on their locale. At the time of writing, it's only
+                   * used to determine whether to count SSN breaches (which we don't have
+                   * recommendations for outside the US).
+                   */
+                  const assumedCountryCode = getSignupLocaleCountry(recipient);
+
+                  const scanData = await getLatestOnerepScanResults(
```
rhelmer, line 282: These may be stale, if that's important then `refreshStoredScanResults` can be `await`ed first.

src/app/(proper_react)/(redesign)/(authenticated)/user/(dashboard)/dashboard/fix/page.tsx
```diff
@@ -3,7 +3,36 @@
...
+    return redirect("/");
+  }
+
+  const countryCode = getCountryCode(headers());
+  const breaches = await getSubscriberBreaches({
+    fxaUid: session.user.subscriber.fxa_uid,
+    countryCode,
+  });
+  const profileId = await getOnerepProfileId(session.user.subscriber.id);
+  const scanData = await getLatestOnerepScanResults(profileId);
```
rhelmer, line 29: Same as above re: `refreshStoredScanResults`

src/app/(proper_react)/(redesign)/(authenticated)/admin/emails/actions.tsx
```diff
@@ -165,16 +172,42 @@ export async function triggerBreachAlert(emailAddress: string) {
...
-    l10n.getString("breach-alert-subject"),
-    <BreachAlertEmail
-      breach={createRandomHibpListing()}
-      breachedEmail={emailAddress}
-      utmCampaignId="breach-alert"
-      l10n={l10n}
-    />,
+  const assumedCountryCode = getSignupLocaleCountry(subscriber);
+
+  const scanData = await getLatestOnerepScanResults(
```
rhelmer, line 177: Same as above re: `refreshStoredScanResults`

rhelmer (changes requested):
As we discussed - if it's important that the scan results are up-to-date then I think you want to call `refreshStoredScanResults` (see the monthly activity report for an example). We know that these are often stale unfortunately so that's the workaround we've been doing so far.

We also talked about using the `./src/scripts/loadtest/hibp.js` k6 script to perform load testing on stage and measure our assumptions about how much impact this will have. The "system" and "query" insights tabs for the database should be especially useful, and we can query logs to get good examples of how many notifications tend to come in for a typical data breach as well as the distribution of hash suffixes per prefix.

src/emails/templates/EmailFooter.tsx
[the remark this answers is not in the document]
Vinnl: Ah right, I hadn't noticed that these are two different functions. I suppose I could use `modifyAttributionsForUrl` for a slightly shorter version - I'd assume I'd pass an empty object for the third parameter (`defaultValues`), but I see many calls to that actually use that for UTM parameters. Does either of you know why those calls don't just override them either? I'm just wondering if we can't just simplify it to a `setUrlParams` function - i.e. why do we even need a `defaultValues` parameter?

src/scripts/cronjobs/emailBreachAlerts.tsx
Vinnl: OK, added d0a6996f6 for the three cases you called out here. One thing I did notice, though, is that we have a lot of calls to `getLatestOnerepScanResults` across our codebase that do not call this function first (e.g. every step in the guided resolution). I guess that might be fine since all of those go through the dashboard, where I assume it should get refreshed, but... Yeah, cache invalidation is hard.

src/app/(proper_react)/(redesign)/(authenticated)/admin/emails/actions.tsx
```diff
@@ -132,6 +137,9 @@ export async function triggerMonthlyActivity(emailAddress: string) {
     month: "long",
   });
 
+  if (typeof subscriber.onerep_profile_id === "number") {
```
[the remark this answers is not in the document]
Vinnl: It does not! But this dashboard is only visible to admins, who want to see what the email looks like. The subject line will also have "Test email:" prefixed to it.

src/scripts/cronjobs/emailBreachAlerts.tsx
Vinnl: I refactored the code to only fetch the above data for free US users who have run a scan, which I think is (at this point in time) a relatively manageable number. I'm still not sure if it was even going to be a problem, but at least this should make it safer to roll out: a7d419d.

src/app/functions/universal/user.test.ts
```diff
@@ -0,0 +1,52 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import { describe, it, expect } from "@jest/globals";
+import { SubscriberRow } from "knex/types/tables";
+import { hasPremium } from "./user";
+import { Session } from "next-auth";
+
+describe("hasPremium", () => {
```
rhelmer, line 10: Thanks for adding a test!

rhelmer (approved):

PR #4924, opened 2024-08-06
Convert subscriber.js into Typescript

codemist:
# References:

Jira: MNTOR-3100

Figma:

# Description

# Screenshot (if applicable)

Not applicable.

# How to test

# Checklist (Definition of Done)

- [ ] Localization strings (if needed) have been added.
- [ ] Commits in this PR are minimal and [have descriptive commit messages](https://chris.beams.io/posts/git-commit/).
- [ ] I've added or updated the relevant sections in readme and/or code comments
- [ ] I've added a unit test to test for potential regressions of this bug.
- [ ] If this PR implements a feature flag or experimentation, the Ship Behind Feature Flag status in Jira has been set
- [ ] Product Owner accepted the User Story (demo of functionality completed) or waived the privilege.
- [ ] All acceptance criteria are met.
- [ ] Jira ticket has been updated (if needed) to match changes made during the development process.
- [ ] Jira ticket has been updated (if needed) with suggestions for QA when this PR is deployed to stage.

PR #5015, opened 2024-09-03
Finish converting subscribers to TS

Vinnl:
Followup to https://github.com/mozilla/blurts-server/pull/4924 (merges into that) to fix unit tests and tighten up a few more types.

### DIFF
--- a/src/app/(proper_react)/(redesign)/(authenticated)/user/(dashboard)/settings/page.tsx
+++ b/src/app/(proper_react)/(redesign)/(authenticated)/user/(dashboard)/settings/page.tsx
@@ -88,6 +88,9 @@ export default async function SettingsPage({ searchParams }: Props) {
   );
 
   const userData = await getSubscriberById(session.user.subscriber.id);
+  if (!userData) {
+    return redirect("/");
+  }
 
   return (
     <SettingsView
--- a/src/app/api/utils/auth.tsx
+++ b/src/app/api/utils/auth.tsx
@@ -242,7 +242,6 @@ export const authOptions: AuthOptions = {
             // MNTOR-2599 The breach_resolution object can get pretty big,
             // causing the session token cookie to balloon in size,
             // eventually resulting in a 400 Bad Request due to headers being too large.
-
             delete updatedUser.breach_resolution;
             token.subscriber = updatedUser;
           } catch (error) {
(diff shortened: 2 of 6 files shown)


src/app/(proper_react)/(redesign)/(authenticated)/user/(dashboard)/settings/page.tsx
```diff
@@ -88,6 +88,9 @@ export default async function SettingsPage({ searchParams }: Props) {
   );
 
   const userData = await getSubscriberById(session.user.subscriber.id);
+  if (!userData) {
```
Vinnl, line 91: A couple of the subscriber-fetching functions (e.g. `getSubscriberById`) assumed that they'd always find one in the DB, but theoretically you could have an invalid ID in here. I thought it was safest to keep that exposed in the types and handle it at the call sites - such as here.

So _theoretically_ this should never get hit, but if someone ends up with a session with an invalid subscriber ID and tries to visit the settings page, this will redirect them to the landing page.

src/app/api/utils/auth.tsx
```diff
@@ -129,13 +129,15 @@ export const authOptions: AuthOptions = {
               account.access_token,
               account.refresh_token,
               account.expires_at ?? 0,
-              JSON.stringify(profile),
+              profile,
```
Vinnl, line 132: I checked, and both calls to `updateFxAData` called `JSON.stringify` on this argument, and then the function itself calls `JSON.parse` again. So I removed both and just passed the full object directly.

(This stringifying and parsing [predates the Next.js codebase by a long time](https://github.com/mozilla/blurts-server/blob/467a892c0b65fbc8889265054a266bd7f8d25433/src/db/tables/subscribers.js#L65), so there probably was a need for it back then.

src/db/tables/subscribers.ts
```diff
@@ -564,21 +616,25 @@ async function getSubscribersWithUnresolvedBreachesCount() {
...
+type WithEmailAddresses = SubscriberRow & {
+  email_addresses: EmailAddressRow[];
+};
+
 /* c8 ignore stop */
 
 // Not covered by tests; mostly side-effects. See test-coverage.md#mock-heavy
 /* c8 ignore start */
-async function joinEmailAddressesToSubscriber(subscriber: SubscriberRow) {
-  if (subscriber) {
```
Vinnl, line 572: I removed the `if` here and just made sure that callers only call this function to query for related email addresses if they've found a subscriber.

PR #5016, opened 2024-09-03
Reduce dashboard axe runs to a single one

Vinnl:
Running the axe accessibility test suite is fairly heavy, and the different storybook scenarios don't differ too much in terms of what the DOM looks like. In other words, running this many tests is unlikely to catch accessibility issues, but it has a quite severe impact on the running time of our unit tests. Thus, it's probably worth it to just run it once instead.


flozia (approved):
I agree that running the axe tests on permutations that are so similar to each other is a bit redundant. Removing those does not seem to shave off so much time — at least in CI — when looking at previous runs, though.

Vinnl:
> Removing those does not seem to shave off so much time — at least in CI — when looking at previous runs, though.

That's fair enough - I guess the main benefit is my own laptop not getting out-of-memory errors as much, with everything else grinding to a halt and some of the tests failing because of that :sweat_smile: CI is probably better able to efficiently parallelise - but the saved CPU cycles at least help some for the environment? :)

PR #5019, opened 2024-09-04
Only require token when unsubscribing

Vinnl:
Since the user is not logged in, there is no extra information given by passing in the email address, but it does add PII to our request logs. Hence, just accepting the unsubscribe token might be enough.

Merges into https://github.com/mozilla/blurts-server/pull/4988. /cc @codemist

### DIFF
--- a/src/app/api/v1/user/unsubscribe-email/route.ts
+++ b/src/app/api/v1/user/unsubscribe-email/route.ts
@@ -5,40 +5,26 @@
 import { NextResponse } from "next/server";
 import type { NextRequest } from "next/server";
 import { logger } from "../../../../functions/server/logging";
-import { verifyUnsubscribeToken } from "../../../utils/email";
-import { unsubscribeMonthlyMonitorReportForEmail } from "../../../../../db/tables/subscriber_email_preferences";
+import { unsubscribeMonthlyMonitorReportForUnsubscribeToken } from "../../../../../db/tables/subscriber_email_preferences";
 
 export async function GET(req: NextRequest) {
   try {
     const { searchParams } = new URL(req.url);
-    const email = searchParams.get("email");
     const unsubToken = searchParams.get("token");
 
-    if (!email || !unsubToken) {
+    if (!unsubToken) {
       return NextResponse.json(
         {
           success: false,
-          message: "email and token are required url parameters.",
+          message: "token is a required url parameter.",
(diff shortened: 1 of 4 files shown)


src/db/tables/subscriber_email_preferences.ts
```diff
@@ -166,9 +166,9 @@ async function getEmailPreferenceForSubscriber(subscriberId: number) {
   return res?.[0];
 }
 
-async function getEmailPreferenceForPrimaryEmail(email: string) {
-  logger.info("get_email_preference_for_primary_email", {
-    email,
+async function getEmailPreferenceForUnsubscribeToken(unsubscribeToken: string) {
```
mansaj, line 169: Since we are just matching tokens, it doesn't have to use join anymore

src/app/api/v1/user/unsubscribe-email/route.ts
```diff
@@ -5,40 +5,26 @@
 import { NextResponse } from "next/server";
 import type { NextRequest } from "next/server";
 import { logger } from "../../../../functions/server/logging";
-import { verifyUnsubscribeToken } from "../../../utils/email";
-import { unsubscribeMonthlyMonitorReportForEmail } from "../../../../../db/tables/subscriber_email_preferences";
+import { unsubscribeMonthlyMonitorReportForUnsubscribeToken } from "../../../../../db/tables/subscriber_email_preferences";
 
 export async function GET(req: NextRequest) {
```
mansaj, line 10: Wil turn this into a POST call to avoid logging token

src/db/tables/subscriber_email_preferences.ts
```diff
@@ -241,6 +248,6 @@ export {
   addEmailPreferenceForSubscriber,
   updateEmailPreferenceForSubscriber,
   getEmailPreferenceForSubscriber,
-  getEmailPreferenceForPrimaryEmail,
-  unsubscribeMonthlyMonitorReportForEmail,
+  getEmailPreferenceForUnsubscribeToken,
+  unsubscribeMonthlyMonitorReportForUnsubscribeToken,
 };
```
mansaj, line 253: Added unique index to the unsub token in the first PR. The index is not necessary anymore

mansaj:
I'll modify this PR and merge it in

PR #5023, opened 2024-09-09
In the flag UI, only show flags that the code uses

Vinnl:
I got some questions from QA about enabling/disabling a flag not working, which was because the flag had been removed from the codebase.

To remove this confusion, I changed the `FeatureFlagName` type to be derived from an array of flag names. I then use this same array in the feature flag UI to determine which flags to actually show, and to also list flags that the codebase knows about, but aren't in the database yet. So in addition to removing the confusion, it also avoids mistyping or even having to memorise the flag name. And hopefully, it also provides another incentive to actually remove the flags from the codebase once they're enabled :)


rhelmer (approved):
This is a great idea, thanks!

PR #5034, opened 2024-09-09
Unused code removal

Vinnl:
Remove some unused code, and some unused devDependencies identified by Knip (d433aeeec859c932fb634fc53b7c47a95da2bd9f).


src/db/tables/emailAddresses.ts
```diff
@@ -326,39 +306,6 @@ async function getUserEmails(userId: number): Promise<EmailAddressRow[]> {
 }
 /* c8 ignore stop */
 
-// This is used by SES callbacks to remove email addresses when recipients
-// perma-bounce or mark our emails as spam
-// Removes from either subscribers or email_addresses as necessary
-// Not covered by tests; mostly side-effects. See test-coverage.md#mock-heavy
-/* c8 ignore start */
-async function removeEmail(email: string) {
```
Vinnl, line 334: I'll just note that this function and the comment sounded like it might still be used, but I tracked it down to getting added in 649d27a8ce94041ee3d5ea71707bfa6bf66f6a05 to support `controllers/ses.js`, which is long gone, so I think we're good there.

package.json
```diff
@@ -123,7 +122,6 @@
     "@types/jest-axe": "^3.5.9",
     "@types/jsonwebtoken": "^9.0.6",
     "@types/jwk-to-pem": "^2.0.3",
-    "@types/k6": "^0.52.0",
```
rhelmer, line 126: Hm, I could swear this was used by `./src/scripts/loadtest/hibp.js`... maybe only for VSCode? I guess as long as it's passing lint and looks OK in VSCode then sure.

src/utils/hibp.ts
```diff
@@ -451,29 +451,6 @@ async function subscribeHash(sha1: string) {
...
- *
- * There is one possible response code that can be returned:
- * HTTP 200: Range subscription successfully deleted
- *
- * @param {string} sha1 sha1 of the email being subscribed
- * @returns 200 response codes
- */
-// TODO: Add unit test when changing this code:
-/* c8 ignore start */
-async function deleteSubscribedHash(sha1: string) {
```
rhelmer, line 466: Hm, yeah this doesn't seem to be how their current API works.

src/db/tables/subscribers.ts
```diff
@@ -475,29 +427,6 @@ async function getSubscribersWaitingForMonthlyEmail(
 }
 /* c8 ignore stop */
 
-/**
- * @param email
- * @deprecated Only used by the `send-email-to-unresolved-breach-subscribers.js`, which it looks like might not be sent anymore? Delete as a part of MNTOR-3077?
- */
-// Not covered by tests; mostly side-effects. See test-coverage.md#mock-heavy
-/* c8 ignore start */
-async function updateMonthlyEmailTimestamp(
```
rhelmer, line 484: 💯

src/db/tables/emailAddresses.ts
rhelmer: This does sound like something we should be implementing... however we do keep an eye on the bounce rate and it hasn't been a problem, so I wonder if this is actually necessary or not. It's probably worth investigating as a followup ticket.

rhelmer (approved):
lgtm overall - as I mentioned though I think the k6 type is or was used by the loadtest script, I thought it was used by our lint and VSCode.

There are a bunch of eslint comments in it which might be why it's working, or maybe we just don't need it. I don't think it's very important either way, k6 doesn't use Node as its JS interpreter so there's a limit to how useful our lint tools are.

package.json
Vinnl: Ah no, you're right, I think that might have been because Knip didn't see a direct import from `k6`? Reverted in 3674b06d1.

I got too many false positives with Knip anyway, so I've shelved it for now. Might re-investigate after everything is TS, perhaps that helps.

PR #5043, opened 2024-09-11
Convert most remaining JS to TS

Vinnl:
# References:

Jira: MNTOR-3077
Figma:

# Description

There were a couple of remaining files in `/src/utils` that were still plain JS, and some in `/src/db/`. This PR converts them to TS, which is possible now that our cron jobs are TS as well.

The remaining JS is:
- `localStorage.js` - this can be removed, but since that conflicts with https://github.com/mozilla/blurts-server/pull/5012, I'm holding off on that for now.
- DB migrations - I don't think Knex supports these as TS, they're fairly trivial anyway, and the existing scripts shouldn't ever change either, so I'm leaving these.
- AppConstants - that has more impact, so I have that as a separate PR in case people have feelings about that: https://github.com/mozilla/blurts-server/pull/5044
- Some config files at the root, which should probably stay JS.
- Scripts in `/src/scripts/loadtest/` and `/src/scripts/build/`; introducing the TS toolchain in there is probably more of a nuisance than a help.

So, give or take a couple of `any`s still lying around the codebase, mostly brings us to full type safety in the entire codebase!


rhelmer (approved):
It's great to finally have this done!

I believe that knex supports TS for migration scripts now, but I agree that it's probably going to be minimally useful, especially for the older migration files that are mostly for record-keeping purposes. However note that we *do* run these still for new installs (mostly dev environments) when `npm run db:migrate` is run for the first time on a new database, so I think it would be OK to convert these if we ever want to.

I agree as well that config files are likely to benefit the least from type safety, I suspect we could move most or all of these but it's not very important.

Vinnl:
> However note that we do run these still for new installs (mostly dev environments) when npm run db:migrate is run for the first time on a new database, so I think it would be OK to convert these if we ever want to.

Yeah, but that's exactly why I'm weary about converting them: we want those to run exactly the same as they did for us, but since we've already run them, if I make any errors while converting, we wouldn't catch that. And since no other code by definition depends on the migrations, there's not really any benefit to converting them anyway. Could be something to consider for future migrations though.

PR #5044, opened 2024-09-11
Remove AppConstants

Vinnl:
To make the migration as safe as possible, in places where it was used to access environment variables that were absolutely required, I replaced it with a function that forces the caller to explicitly enumerate the environment variables they specifically expect, allows them to access them in a type-safe way, and throws if the named environment variables aren't set. (This is slightly different than the previous behaviour, where AppConstants would only log the missing variable without throwing, but I checked that all these variables should be set in production.)

Merges into https://github.com/mozilla/blurts-server/pull/5043.


src/app/api/utils/auth.tsx
```diff
@@ -26,6 +25,15 @@ import { SerializedSubscriber } from "../../../next-auth";
 import { record } from "../../functions/server/glean";
 import { renderEmail } from "../../../emails/renderEmail";
 import { SignupReportEmail } from "../../../emails/templates/signupReport/SignupReportEmail";
+import { getEnvVarsOrThrow } from "../../../envVars";
+
+const envVars = getEnvVarsOrThrow([
```
rhelmer, line 30: Oh that's pretty cool, nice idea. I didn't like how non-standard `AppConstants` was but I really like the behavior of throwing upfront.

src/app/api/v1/user/email/route.ts
```diff
@@ -108,6 +107,6 @@ export async function POST(req: NextRequest) {
     }
   } else {
     // Not Signed in, redirect to home
-    return NextResponse.redirect(AppConstants.SERVER_URL, 301);
+    return NextResponse.redirect(process.env.SERVER_URL ?? "/", 301);
```
rhelmer, line 110: I'm curious why you don't use the `getEnvVarsOrThrow` function here? Seems like you could avoid having to do the ` ?? "/"`, is this fallback ever desirable?

src/app/api/v1/user/remove-email/route.ts
```diff
@@ -50,7 +49,7 @@ export async function POST(req: NextRequest) {
         existingEmail.email,
       );
       return NextResponse.redirect(
-        AppConstants.SERVER_URL + "/user/settings",
+        process.env.SERVER_URL + "/user/settings",
```
rhelmer, line 52: Same question re: `getEnvVarsOrThrow` vs. `process.env`

src/scripts/cronjobs/updateBreachesInRemoteSettings.ts
```diff
@@ -33,19 +33,36 @@
...
+const FX_REMOTE_SETTINGS_WRITER_PASS =
+  process.env.FX_REMOTE_SETTINGS_WRITER_PASS;
+const FX_REMOTE_SETTINGS_WRITER_SERVER =
+  process.env.FX_REMOTE_SETTINGS_WRITER_SERVER;
+
+if (
+  !FX_REMOTE_SETTINGS_WRITER_USER ||
+  !FX_REMOTE_SETTINGS_WRITER_PASS ||
+  !FX_REMOTE_SETTINGS_WRITER_SERVER
+) {
```
rhelmer, line 54: Seems like you could use `getEnvVarsOrThrow` and catch it + `process.exit` (I think throwing from here would the process to exit anyway but being explicit is probably better just in case it's refactored later)

rhelmer (approved):
lgtm, I'm wondering why you don't use the new `getEnvVarsOrThrow` in a few places but I'll leave it up to you whether it's appropriate vs. plain `process.env`.

src/app/api/v1/user/remove-email/route.ts
Vinnl: Same reasoning as https://github.com/mozilla/blurts-server/pull/5044#discussion_r1756579517 - in this case, if it's unset, it would fall back to redirecting to `/user/settings` anyway, I think.

src/scripts/cronjobs/updateBreachesInRemoteSettings.ts
Vinnl: Yeah, I was actually considering that too, but then I figured I'd have to wrap the entire file in a `try {}` block, which... Isn't that big of a deal, but felt a little bit icky, so I stuck with mostly matching the existing code. I hadn't realised yet that throwing would cause an exit anyway, so that could actually be a good compromise too. What do you think?

Vinnl:
@rhelmer Heh, my answers to all your questions are basically: good point, here's my reasoning, but I could go either way, what do you think? So, do you have opinions? :P

rhelmer (approved):

PR #5059, opened 2024-09-16
Make load test script more self-documenting

Vinnl:
It also allows you to pass in a plain email address and have the performance test send breach notifications for that specific email address, without having to manually calculate a hash.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -68,3 +68,7 @@ updates:
     directory: "/"
     schedule:
       interval: "weekly"
+    ignore:
+          - dependency-name: "node"
+            # Ignore latest version of Node, since we prefer LTS. This will need to be updated over time:
+            versions: ["22.x"]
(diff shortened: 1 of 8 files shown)


src/scripts/loadtest/hibp.js
```diff
@@ -11,33 +11,55 @@
...
+  );
+}
 
-const url = `${__ENV.SERVER_URL}/api/v1/hibp/notify`;
+const VIRTUAL_USERS =
+  typeof envVars.K6_VIRTUAL_USERS === "string"
+    ? parseInt(envVars.K6_VIRTUAL_USERS, 10)
+    : typeof envVars.K6_VUS === "string"
+      ? parseInt(envVars.K6_VUS, 10)
+      : 1000;
```
Vinnl, line 32: I arbitrarily took 1000 as a default value here, but I don't know how many requests HIBP sends in practice - that would probably be a better value.

rhelmer: It really depends on the breach... I think that's a good default, especially if running against a less-provisioned environment than our production instance.

rhelmer (approved):

PR #5068, opened 2024-09-16
MNTOR-3029 - guard against un-parseable URLs being passed to next-auth redirect

rhelmer:
# References:

Jira: MNTOR-3029

# Description

Currently next-auth throws an unhandled exception if there are fuzzing attempts against the `redirect` endpoint. We do want to log these, but it being unhandled means it pollutes Sentry and our logs with a lot of useless information.

This modifies the default behavior (per the documentation) to:

1. use the `URL.canParse` static method to try parsing any incoming URLs
2. if it cannot be parsed, log this and return `baseUrl`

We should consider trying to upstream this to the next-auth project, I think it's a better default but one problem is that logging is not very standardized in the JS ecosystem so I'm not sure what a good upstream fix would look like yet.

NOTE: this introduces a slight change in behavior - the way this feature works is that the `next-auth.callback-url` cookie is set to whatever the URL passed in a `POST` to `/api/auth/signin/fxa` via the`callbackUrl` form field is. If the `callbackUrl` starts with a `/` then it is appended to `baseUrl` (relative), otherwise it's treated as an absolute URL.

The URL in the cookie is URL-encoded, but other than that it seems to allow pretty much any path here. For absolute URLs there's an origin check using `new URL(...).origin`, and relative URLs must begin with `/` and always prepend the `baseUrl`.

Before this PR, next-auth throws an unhandled exception from `new URL()` so HTTP 500 is returned to the caller. With this PR, any un-pareseable URLs are logged and `baseUrl` is returned, so the caller gets an HTTP 200 instead.

I'm going to land this after the next release so we have more time for testing, and I also want to file an upstream issue with next-auth because I think this should be better handled by next-auth itself. It shouldn't be possible for callers to just cause an HTTP 500 to be thrown through invalid input - this is a client error and should be HTTP 4xx class (probably `400 Bad Request`)

# How to test

Sign in/out should work as normal and redirect to any valid URL. Attempting to sign in with an invalid redirect URL should be ignored and the request should be redirected to `/`.

See the Jira ticket for an example test case that triggers the HTTP 500 response from next-auth without this PR.

- [ ] File a ticket on next-auth project


src/app/api/utils/auth.tsx
```diff
@@ -81,6 +81,31 @@ export const authOptions: AuthOptions = {
   },
   providers: [fxaProviderConfig],
   callbacks: {
+    redirect({ url, baseUrl }) {
+      // Do not pass un-parseable URLs to next-auth, @see MNTOR-3029
+      // Preserve default behavior otherwise. @see https://next-auth.js.org/configuration/callbacks#redirect-callback
+
+      // Allows relative callback URLs
+      if (url.startsWith("/")) {
+        if (!URL.canParse(`${baseUrl}${url}`)) {
```
Vinnl, line 90: TIL about `URL.canParse`, neat!

src/app/api/utils/auth.tsx
```diff
@@ -81,6 +81,31 @@ export const authOptions: AuthOptions = {
...
+          return baseUrl;
+        }
+        return `${baseUrl}${url}`;
+      }
+      // Allows callback URLs on the same origin
+      else if (new URL(url).origin === baseUrl) {
+        if (!URL.canParse(url)) {
+          logger.warn("Cannot parse redirect url:", { url });
+          return baseUrl;
+        }
```
Vinnl, line 103: I think `new URL` throws when `url` cannot be parsed, so I think this clause is never hit? (And thus we'd still see the parsing errors in the logs?)

src/app/api/utils/auth.tsx
```diff
@@ -81,6 +81,31 @@ export const authOptions: AuthOptions = {
...
   providers: [fxaProviderConfig],
   callbacks: {
+    redirect({ url, baseUrl }) {
+      // Do not pass un-parseable URLs to next-auth, @see MNTOR-3029
+      // Preserve default behavior otherwise. @see https://next-auth.js.org/configuration/callbacks#redirect-callback
+
+      // Allows relative callback URLs
+      if (url.startsWith("/")) {
+        if (!URL.canParse(`${baseUrl}${url}`)) {
+          logger.warn("Cannot parse redirect url:", {
```
Vinnl, line 91: I half remember Joe mentioning that we had a convention to use snake_cased identifiers for our logs? e.g.

```suggestion
          logger.warn("nextauth_relative_redirect_url_parse_error:", {
```

Vinnl (approved):
Looking forward to seeing in our logs what these invalid URLs look like!

src/app/api/utils/auth.tsx
rhelmer: Hm OK well it makes sense why it was throwing before... I'm a little worried about changing the behavior here, I guess next-auth throws an error response here now.

I'm going to try making it explicitly return something like:
```
NextResponse.json({ success: false }, { status: 403 })
```
If the URL passed looks absolute and it can't be parsed.

rhelmer: Hm OK returning `NextResponse.json` won't work here, I just walked through this with a debugger and see what the problem is. I left a minimal `curl` command in the Jira ticket too.

rhelmer: I re-wrote this so `URL.canParse` is called before `new URL(url).origin`.

rhelmer:
> Looking forward to seeing in our logs what these invalid URLs look like!

I put a sample in the Jira ticket, the current exception has it (but it's not using structured logging so it shows up on different lines, it's a pain to dig out currently).

src/app/api/utils/auth.tsx
Vinnl: Did you maybe forget to push it? I think I still see it after `new URL`.

Also, I just learned of the static [`URL.parse`](https://developer.mozilla.org/en-US/docs/Web/API/URL/parse_static) method, which doesn't throw but returns `null` for invalid URLs - that might be relevant to your interests.

src/app/api/utils/auth.tsx
```diff
@@ -81,6 +81,33 @@ export const authOptions: AuthOptions = {
...
+        if (URL.canParse(url)) {
+          if (new URL(url).origin === baseUrl) {
+            return url;
+          }
+        } else {
+          logger.warn("nextauth_absolute_redirect_url_parse_error", { url });
+          return baseUrl;
+        }
+      }
+      return baseUrl;
```
Vinnl, line 109: Reading [the docs](https://next-auth.js.org/configuration/callbacks#redirect-callback), it looks like the default behaviour is to only allow absolute URLs with the same base URL, and the relative URL bit is only to showcase the use of the callback?

```suggestion
      if (URL.canParse(url) && new URL(url).origin === baseUrl) {
        return url;
      }

      logger.warn("nextauth_absolute_redirect_url_parse_error", { url });
      return baseUrl;
```

Vinnl (approved):

src/app/api/utils/auth.tsx
rhelmer: Hm I just took a look at what actually ships in the `next-auth` package and it looks like this:

```
 redirect({
    url,
    baseUrl
  }) {
    if (url.startsWith("/")) return `${baseUrl}${url}`;else if (new URL(url).origin === baseUrl) return url;
    return baseUrl;
  },
```

PR #5070, opened 2024-09-17
Remove localstorage.js

Vinnl:
The last remaining reference was in
`src/scripts/cronjobs/emailBreachAlerts`, which didn't actually use it; it only stored the list of relevant locales there. There's no more code remaining that read that.


src/scripts/cronjobs/emailBreachAlerts.tsx
```diff
@@ -242,133 +240,109 @@ export async function poll(
...
-
-        const availableLanguages = process.env.SUPPORTED_LOCALES!.split(",");
-        const supportedLocales = negotiateLanguages(
-          requestedLanguage,
-          availableLanguages,
-          { defaultLocale: "en" },
-        );
-
-        await localStorage.run(new Map(), async () => {
-          localStorage.getStore().set("locale", supportedLocales);
```
Vinnl, line 262: Note: all the "new code" is just the code inside this `async` function with two fewer levels of indentation, but otherwise unchanged. So the only real change in this PR is the removal of lines 248-262.

flozia (approved):

PR #5101, opened 2024-09-25
add GA4 un/subscribe events

rhelmer:
# References:

Jira: MNTOR-3007

# Description

This change adds a custom `unsubscription` and `subscription` event to GA using the [GA4 Measurement Protocol](https://developers.google.com/analytics/devguides/collection/protocol/ga4).

Note that we explored using the standard  [purchase](https://developers.google.com/analytics/devguides/collection/ga4/set-up-ecommerce) event but we don't have all the required parameters we need such as the `value`, since this is fired in response to a webhook and we can't query SubPlat for more information.

# How to test

There is a new `sendPingToGA()` server function which is called in the Accounts webhook handler when subscription notifications are received. This function can also be called directly (see the unit test for an example).

When running locally, `debug_mode` will be set to `true` and pings can be tested as documented at https://developers.google.com/analytics/devguides/collection/protocol/ga4/verify-implementation?client_type=gtag

It's also possible to send pings to the validation server which checks them for well-formedness and returns JSON, but isn't very useful otherwise since you can't inspect the data: https://developers.google.com/analytics/devguides/collection/protocol/ga4/validating-events?client_type=gtag


src/app/(proper_react)/layout.tsx
```diff
@@ -24,6 +25,24 @@ export default async function Layout({ children }: { children: ReactNode }) {
...
+      await addClientIdForSubscriber(
+        session?.user.subscriber?.id,
+        gaCookieVersion,
+        parseInt(gaCookiePath),
+        gaCookieClientId,
+        parseInt(gaCookieTimestamp),
+      );
+    }
+  }
+
```
rhelmer, line 45: @Vinnl do you think this belongs in a server-only component like `PageLoadEvent` or is it better directly in here? Or is there some other approach I'm missing?

src/app/functions/server/googleAnalytics.test.ts
```diff
@@ -0,0 +1,163 @@
...
+  }
+
+  const logger = new Logging();
+  return {
+    logger,
+  };
+});
+
+beforeEach(async () => {
+  global.fetch = jest.fn().mockResolvedValue({ ok: true });
```
Vinnl, line 31: If you'd like to fix the type errors here:

```suggestion
  global.fetch = jest.fn<typeof global.fetch>().mockResolvedValue({ ok: true } as Response);
```

src/app/functions/server/googleAnalytics.test.ts
```diff
@@ -0,0 +1,163 @@
...
+
+  const { logger } = await import("./logging");
```
Vinnl, line 33: Is this line still necessary? Tests seem to succeed without it. (And if it is required, is assigning it to `logger` required?)

```suggestion
```

src/app/functions/server/googleAnalytics.test.ts
```diff
@@ -0,0 +1,163 @@
...
+    },
+  );
+});
+
+it("sends event name and parameters to GA and receives error response", async () => {
+  global.fetch = jest.fn().mockResolvedValue({
+    ok: false,
+    status: "500",
+    text: jest.fn().mockResolvedValue("failed"),
+  });
```
Vinnl, line 67: Likewise, to fix the type errors:

```suggestion
  global.fetch = jest.fn<typeof global.fetch>().mockResolvedValue({
    ok: false,
    status: 500,
    text: jest.fn<Response["text"]>().mockResolvedValue("failed") as Response["text"],
  } as Response);
```

src/app/api/v1/fxa-rp-events/route.ts
Vinnl, line 1: Nice tests!

src/app/functions/server/googleAnalytics.ts
```diff
@@ -0,0 +1,77 @@
...
+  const gaClientInfo = await getClientIdForSubscriber(subscriberId);
+  if (!gaClientInfo) {
+    throw new Error(`No stored GA cookie for subscriber ${subscriberId}`);
+  }
+
+  const { client_id, cookie_timestamp } = gaClientInfo;
+
+  if (!client_id || !cookie_timestamp) {
+    throw new Error(
+      `No GA client_id found for subscriber ${subscriberId}, cannot send backend events to Google Analytics`,
```
Vinnl, line 38: Small trick I learned from a former colleague: if you wrap the variable in start- and end-delimiters, it's easier to notice when a value is missing or malformed:

```suggestion
      `No GA client_id found for subscriber [${subscriberId}], cannot send backend events to Google Analytics`,
```

Vinnl (approved):
In typical Vincent style I've left a bunch of comments, but I think not a single one is blocking - although I'd really appreciate having type definitions for the new table, if we decide to keep it.

.env
```diff
@@ -162,4 +162,5 @@ SENTRY_AUTH_TOKEN=
 # Whether GA4 sends data or not. NOTE: must be set in build environment.
 NEXT_PUBLIC_GA4_DEBUG_MODE=true
 
-CURRENT_COUPON_CODE_ID=
\ No newline at end of file
+CURRENT_COUPON_CODE_ID=
+GA4_API_SECRET=unsafe-default-secret-for-dev
```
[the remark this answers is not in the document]
rhelmer: I'm only using the value for tests, it's for the backend to be able to send to GA (which can be secured using a secret unlike usual front-end GA code). I can remove it by default that's fine

src/db/migrations/20240920175305_add-ga4-client-to-subscribers.js
```diff
@@ -0,0 +1,26 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+ /**
+  * @param { import("knex").Knex } knex
+  * @returns { Promise<void> }
+  */
+export async function up(knex) {
+  await knex.schema.createTable('google_analytics_clients', (table) => {
```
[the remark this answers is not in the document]
rhelmer: Oh thanks I forgot about that... sure

src/app/functions/server/googleAnalytics.test.ts
rhelmer: Hm pretty sure I tried that, I ended up switching over to `spy` instead which seems to work fine locally with the linter and the build. I'll leave this comment open until tests are fully passing though, I'm pretty sure the unit test failures are mocking problems and not implementation problems.

src/app/functions/server/googleAnalytics.ts
rhelmer: Oh good idea - I didn't commit the suggestion because I found a few other places to fix this and pushed my branch.

src/app/functions/server/googleAnalytics.test.ts
rhelmer: Ah OK I tried again and had just mocked it incorrectly, fixed in my branch thanks!

.env
rhelmer, line 166: @Vinnl your point about not setting this so we throw an error in production if it's unset makes sense... however that'll make the test fail when running locally or in CI (unless it's set to some known value).

What do we usually do in this case? I could leave it unset but handle it different in `test` and `local` environments... that is making the code-under-test behave a bit differently based on environment which complicates things but might make sense in this case.

Vinnl: Hmm, I think if it really doesn't matter what this value is set to, I suppose I'd look into not making the tests fail if it's not set? i.e. I assume that if the tests actually verify GA behaviour, then this would have to be set to a true value anyway?

Agreed that modifying the code-under-test is probably no better than setting this by default.

PR #5121, opened 2024-10-01
MNTOR-3640 - Add footer background image only during light mode

codemist:
# References:

Jira: MNTOR-3640

# Description

# Screenshot (if applicable)

Light mode:
image
Dark mode: 
image

Not applicable.

# How to test


src/emails/templates/EmailFooter.tsx
```diff
@@ -137,107 +137,115 @@ export const RedesignedBreachEmailFooter = (props: Props) => {
...
-            </a>
-            &emsp;
-            <a href={CONST_URL_PRIVACY_POLICY} style={{ color: "#0060DF" }}>
-              {l10n.getString("email-footer-meta-privacy-notice")}
-            </a>
-          </mj-text>
-        </mj-column>
-      </mj-section>
-    </mj-wrapper>
+    <mjml>
```
Vinnl, line 140: I don't think you can nest `<mjml>` tags inside other `<mjml>` tags (the ones in the full email template). So I think the `<mj-style>` will have to be moved into the `<mj-head>` of every individual email :/ 

I also would instead try to hide the image in dark mode, rather than adding it in light mode - with dark mode being the edge case, and media queries (and anything that's not native MJML functionality) being unlikely to work consistently across email clients, that ensures that it's both less likely to go wrong, and when it does go wrong, that only the rare case breaks.

Vinnl (commented):
Unfortunately I don't think this is the way to solve it :/ I've done a bit of searching around for a "good" way to do so, but haven't been able to find one yet. Since it wasn't marked as a blocker, I deferred that work for now.

Adding a media query to the head of every email that hides the images in dark mode could work, but is a bit of a hassle that "pollutes" every email we have and can easily be forgotten, so I'm not sure if that's worth the hassle either.

src/emails/templates/EmailFooter.tsx
Vinnl: Yeah 😢

Vinnl: Yeah up to you if it's worth the hassle, depending on how important product considers this to be :)

src/emails/templates/breachAlert/BreachAlertEmail.tsx
```diff
@@ -34,6 +35,7 @@ export const BreachAlertEmail = (props: Props) => {
     <mjml>
       <mj-head>
         <mj-preview>{l10n.getString("email-spotted-new-breach")}</mj-preview>
+        {HeaderStyles()}
```
Vinnl, line 38: I think it's the same, but any reason not to do this? (Here and everywhere in this PR?)

```suggestion
        <HeaderStyles/>
```

Vinnl (approved):
Curious to see if this'll work!

PR #5104, opened 2024-09-26
Add support for Mozilla Accounts prompt none auth flow

flozia:
# References:

Jira: [MNTOR-3492](https://mozilla-hub.atlassian.net/browse/MNTOR-3492)

# Description

Adds support for Mozilla Accounts `promt=none` authentication flow. The initial use cases will be a promotional link from Mozilla Accounts. We expect them to link to us with the following UTM parameters that we can use to initiate the `promt=none` authentication flow:
- `utm_source=moz-account`
- `utm_campaign=settings-promo`
- `utm_content=monitor-free`

# How to test

- Enable the feature flag `PromptNoneAuthFlow`
- Visit http://localhost:6060/?utm_source=moz-account&utm_campaign=settings-promo&utm_content=monitor-free with and without the user authenticated with FxA

# Checklist (Definition of Done)

- [x] Commits in this PR are minimal and [have descriptive commit messages](https://chris.beams.io/posts/git-commit/).
- [x] I've added a unit test to test for potential regressions of this bug.
- [x] If this PR implements a feature flag or experimentation, the Ship Behind Feature Flag status in Jira has been set
- [x] Product Owner accepted the User Story (demo of functionality completed) or waived the privilege.
- [x] All acceptance criteria are met.
- [x] Jira ticket has been updated (if needed) with suggestions for QA when this PR is deployed to stage.

[MNTOR-3492]: https://mozilla-hub.atlassian.net/browse/MNTOR-3492?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #5150, opened 2024-10-09
Add E2E tests for silent authentication

flozia:
# References:

PR #5104

# Description

This PR adds E2E tests for the silent authentication flow.

### DIFF
--- a/src/e2e/pages/authPage.ts
+++ b/src/e2e/pages/authPage.ts
@@ -96,7 +96,7 @@ export class AuthPage {
 
   async initSilentAuth() {
     await this.page.goto(
-      `${process.env.E2E_TEST_BASE_URL as string}/?utm_source=moz-account&utm_campaign=settings-promo&utm_content=monitor-free`,
+      `${process.env.E2E_TEST_BASE_URL as string}/?feature_flags=PromptNoneAuthFlow&utm_source=moz-account&utm_campaign=settings-promo&utm_content=monitor-free`,
     );
     // FxA can take a while to load on stage:
     await this.page.waitForURL("**/oauth/**");
--- a/src/middleware.ts
+++ b/src/middleware.ts
@@ -12,6 +12,13 @@ export function middleware(request: NextRequest) {
 
   const requestHeaders = new Headers(request.headers);
   requestHeaders.set("x-nonce", nonce);
+
+  if (process.env.E2E_TEST_ENV === "local") {
+    const forcedFeatureFlags =
+      request.nextUrl.searchParams.get("feature_flags");
+    requestHeaders.set("x-forced-feature-flags", forcedFeatureFlags ?? "");
+  }
+
   // Add the CSP to the request headers - that will make Next.js detect it and
   // add it to the inline `<script>` tags that it injects itself, as per
   // https://github.com/vercel/next.js/discussions/51039#discussioncomment-6596642
--- a/src/middleware.ts
+++ b/src/middleware.ts
@@ -13,7 +13,7 @@ export function middleware(request: NextRequest) {
   const requestHeaders = new Headers(request.headers);
   requestHeaders.set("x-nonce", nonce);
 
-  if (process.env.E2E_TEST_ENV === "local") {
+  if (typeof process.env.E2E_TEST_ENV !== "undefined") {
     const forcedFeatureFlags =
       request.nextUrl.searchParams.get("feature_flags");
     requestHeaders.set("x-forced-feature-flags", forcedFeatureFlags ?? "");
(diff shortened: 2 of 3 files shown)


src/middleware.ts
```diff
@@ -12,6 +12,13 @@ export function middleware(request: NextRequest) {
 
   const requestHeaders = new Headers(request.headers);
   requestHeaders.set("x-nonce", nonce);
+
+  if (typeof process.env.E2E_TEST_ENV !== "undefined") {
```
mansaj, line 16: I'm worried that if this env var is set somehow in stage or prod, we would have this feature flag forced? 

Would it be better to have more guards like `&& not in prod && not in stage`?

flozia: I was thinking about this as well and had the condition first only to check for `process.env.E2E_TEST_ENV === "local"`. Now you are mentioning it: Yes, let’s play it safe — changed in [4ae7859](https://github.com/mozilla/blurts-server/pull/5150/commits/4ae78594797e925b6e975b4b6ae4d73f019da948).

rhelmer (approved):

src/e2e/pages/authPage.ts
```diff
@@ -71,4 +71,34 @@ export class AuthPage {
     expect(verificationCode).toBeDefined();
     await this.enterVerificationCode(verificationCode as string);
   }
+
+  async signInToFxA(email: string, password: string) {
```
Vinnl, line 75: It's not clear to me (because I'm not terribly familiar with our existing e2e tests) why this extra method is needed - don't we already have methods to sign in?

src/middleware.ts
```diff
@@ -12,6 +12,13 @@ export function middleware(request: NextRequest) {
 
   const requestHeaders = new Headers(request.headers);
   requestHeaders.set("x-nonce", nonce);
+
+  if (process.env.E2E_TEST_ENV === "local") {
+    const forcedFeatureFlags =
+      request.nextUrl.searchParams.get("feature_flags");
+    requestHeaders.set("x-forced-feature-flags", forcedFeatureFlags ?? "");
+  }
+
```
Vinnl, line 21: Instead of adding runtime middleware code, we could also skip that and just directly call [`.setExtraHTTPHeaders`](https://playwright.dev/docs/api/class-page#page-set-extra-http-headers), I think?

src/e2e/pages/authPage.ts
```diff
@@ -71,4 +71,34 @@ export class AuthPage {
...
+    await this.page.waitForURL(/^(?!.*signin).*/);
+
+    // enter password
+    await this.passwordInputField.fill(password);
+    await this.continue({ waitForURL: process.env.FXA_SETTINGS_URL });
+  }
+
+  async initSilentAuth() {
+    await this.page.goto(
+      `${process.env.E2E_TEST_BASE_URL as string}/?feature_flags=PromptNoneAuthFlow&utm_source=moz-account&utm_campaign=settings-promo&utm_content=monitor-free`,
```
Vinnl, line 99: So as per my other comment, this [would be](https://playwright.dev/docs/api/class-page#page-set-extra-http-headers):

```suggestion
    await this.page.setExtraHTTPHeaders({ "x-forced-feature-flags": "PromptNoneAuthFlow" });
    await this.page.goto(
      `${process.env.E2E_TEST_BASE_URL as string}/?utm_source=moz-account&utm_campaign=settings-promo&utm_content=monitor-free`,
```

Vinnl (commented):
The feature flag override looks good to me, and overall the code looks good too (just a question and an ignorable suggestion). However... The tests appear to fail in CI?

flozia:
There seems to be an issue with the test accounts set up for the E2E tests. When attempting to sign-in to FxA directly rather than via OAuth it results in an application error. The issue is filed under [FXA-10631](https://mozilla-hub.atlassian.net/browse/FXA-10631).

src/e2e/pages/authPage.ts
flozia: This method is for logging explicitly in to the Monitor Accounts application and not Monitor.

src/middleware.ts
flozia: Oh yeah — a lot better: [7f3feea](https://github.com/mozilla/blurts-server/pull/5150/commits/7f3feeaca6ffce562b3ab32e7319ebe97b26fda4).

src/e2e/pages/authPage.ts
flozia: Thanks! Update in [7f3feea](https://github.com/mozilla/blurts-server/pull/5150/commits/7f3feeaca6ffce562b3ab32e7319ebe97b26fda4).

flozia:
> The feature flag override looks good to me, and overall the code looks good too (just a question and an ignorable suggestion). However... The tests appear to fail now in CI?

@Vinnl The tests are passing again. An issue with the account prevented us from logging into Mozilla Accounts.

PR #5154, opened 2024-10-09
Add cronjob for the monthly free email

Vinnl:
# References:

Jira: MNTOR-3606
Figma:

# Description

There's quite a lot in this PR, as I found the need to make more changes as I tested the end-to-end flow - apologies for it getting unwieldy. What I did:

- Remove the need for an explicit `React` import (but since this is broken on `main`, I've also submitted that separately: https://github.com/mozilla/blurts-server/pull/5153)
- Removed the `MonthlyActivityEmail` flag
- Removed the unsubscribe link button from the email admin tool, and added the ability to send the activity email (which contains that link)
- Fixed the unsubscription API call
- Fixed the unsubscription API for cases where the subscription preference isn't set
- Updated Storybook to show MJML errors
- Fix MJML errors in the free monthly email template
- Grouped the monthly emails into `monthlyActivityFree` and `monthlyActivityPlus` folders, i.e. at the same level in the hierarchy
- And of course, added the new cron job

I'll go through the PR to call out some things as well :)

# How to test

You might want to add this to your `.env.local` to only do two at a time, to make it easier to verify in the DB for example.

This should run the cron job directly: `npm run dev:cron:monthly-activity-free`

This should be more realistic for production: `npm run build-cronjobs; npm run cron:monthly-activity-free`

That should send the activity email to free users whose account was created later than 1 January 2022, and whose language was set to `en-US` when signing up (or was unset).

- [ ] I've added a unit test to test for potential regressions of this bug. - Lots of side effects


src/app/(proper_react)/(redesign)/(public)/unsubscribe-email/monthly-report-free/UnsubscribeMonthlyReportView.tsx
```diff
@@ -38,9 +38,12 @@ export const UnsubscribeMonthlyReportView = ({ token }: { token: string }) => {
 
   const handleUnsubscription = async () => {
     try {
-      const response = await fetch(`/api/unsubscribe-email?token=${token}`, {
-        method: "GET",
-      });
+      const response = await fetch(
+        `/api/v1/user/unsubscribe-email?token=${token}`,
```
Vinnl, line 42: I saw this request 404'ing; I'm assuming that this is the correct URL.

src/app/api/utils/email.tsx
```diff
@@ -10,14 +10,6 @@ import { VerifyEmailAddressEmail } from "../../../emails/templates/verifyEmailAd
 import { sanitizeSubscriberRow } from "../../functions/server/sanitize";
 import { getL10n } from "../../functions/l10n/serverComponents";
```
Vinnl, line 11: This import causes problems in cronjobs (which is why there's a separate l10n module for cronjobs), and since this `unsubscribeLinkForSubscriber` is only intended for the monthly free email, I moved it into that module.

src/db/tables/subscriber_email_preferences.ts
```diff
@@ -261,11 +261,14 @@ async function unsubscribeMonthlyMonitorReportForUnsubscribeToken(
   try {
     sub = await getEmailPreferenceForUnsubscribeToken(unsubscribeToken);
 
-    if (sub.id && !sub.monthly_monitor_report_free) {
+    if (
```
Vinnl, line 264: I think the previous code didn't take into account that `sub.monthly_monitor_report_free` could be `null` or `undefined`, hence these changes.

src/db/tables/subscribers.ts
```diff
@@ -397,27 +392,17 @@ async function getSubscribersWaitingForMonthlyEmail(
...
-    .andWhereRaw("\"created_at\" < NOW() - INTERVAL '1 month'");
-
-  if (Array.isArray(flag.allow_list) && flag.allow_list.length > 0) {
-    // If the feature flag has an allowlist, only send to users on that list.
-    // The `.andWhereIn` alias doesn't exist:
-    // https://github.com/knex/knex/issues/1881#issuecomment-275433906
-    query = query.whereIn("primary_email", flag.allow_list);
-  }
-
-  if (options.plusOnly) {
```
Vinnl, line 410: This removes the flag check, and always limits this to Plus users (rather than accepting that as an option).

src/db/tables/subscribers.ts
```diff
@@ -429,6 +414,83 @@ async function getSubscribersWaitingForMonthlyEmail(
...
 
+// Not covered by tests; mostly side-effects. See test-coverage.md#mock-heavy
+/* c8 ignore start */
+async function getFreeSubscribersWaitingForMonthlyEmail(): Promise<
+  SubscriberRow[]
+> {
+  const flag = await getFeatureFlagData("MonthlyReportFreeUser");
+  const accountCutOffDate = parseIso8601Datetime(
+    process.env.MONTHLY_ACTIVITY_FREE_EMAIL_ACCOUNT_CUTOFF_DATE ??
+      "2022-01-01T00:00:00.000Z",
```
Vinnl, line 425: Ticket said:

> Only send to monitor users (...) who have signed up for monitor in the past 3 years.

However, I'm assuming that that should be 4 years in a year, so I just set 1 January 2022 as the cutoff date.

src/emails/components/EmailBanner.tsx
```diff
@@ -46,43 +46,41 @@ export const EmailBanner = (props: {
...
-              <mj-text
-                font-size="16px"
-                line-height="24px"
-                padding="0"
-                color="#000000"
-              >
-                <h3>{props.heading}</h3>
-                <p>{props.content}</p>
-              </mj-text>
+        <mj-group>
```
Vinnl, line 49: @codemist I think the changes to the template are most relevant to you; MJML threw some errors about `<mj-wrapper>`s not being allowed inside sections, and `<mj-group>`s not in wrappers, etc.

(Those errors were not visible in Storybook before; I updated Storybook to also show those errors. I tried to keep the looks the same while fixing the errors.)

src/db/tables/subscribers.ts
```diff
@@ -429,6 +414,85 @@ async function getSubscribersWaitingForMonthlyEmail(
...
+  );
+
+  if (!flag?.is_enabled) {
+    return [];
+  }
+
+  let query = knex("subscribers")
+    .select<SubscriberRow[]>("subscribers.*")
```
Vinnl, line 433: One thing I should point out here: I did not add a LIMIT clause. The reason for this is that we also need to select people who had en-US as their sign up language, but we need to parse the accept-language field to do so. So this'll fetch a lot of records, and then only afterwards filter them down to a small batch. (I should probably add this as a comment, but I'm on my phone now.)

src/emails/components/EmailBanner.tsx
Vinnl: Thanks for checking! I saw them when actually trying to send the email, which errored out at [the call to `renderEmail`](https://github.com/mozilla/blurts-server/blob/076f70e0f7700a5d3cd99a3f059e557ea709053c/src/scripts/cronjobs/monthlyActivityFree.tsx#L101-L108).

Which of course is a bit cumbersome, which is why I updated the `<StorybookEmailRenderer>` to also show those errors, so that next time, we can see them while actually working on the template :)

src/app/(proper_react)/(redesign)/(authenticated)/user/(dashboard)/settings/AlertAddressForm.tsx
```diff
@@ -57,7 +57,7 @@ export const AlertAddressForm = (props: Props) => {
     : props.data.monthly_monitor_report_free;
 
   // TODO: Deprecate this when monthly report for free users has been created
-  const monthlyFreeUserReportEnabled =
+  const canToggleFreeOrPlusMonthlyReport =
     props.enabledFeatureFlags.includes("MonthlyReportFreeUser") ||
     hasPremium(props.user);
```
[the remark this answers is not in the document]
Vinnl: Ah sorry, that's because I was confused, seeing as it also controlled showing the toggle for Plus users (where the old name implied it was only for the free report). Indeed we should be able to remove it once the `MonthlyReportFreeUser` flag is removed, and replace it with just the `hasPremium` call.

src/app/(proper_react)/(redesign)/(authenticated)/user/(dashboard)/settings/AlertAddressForm.tsx
```diff
@@ -191,30 +191,29 @@ export const AlertAddressForm = (props: Props) => {
             </AlertAddressRadio>
           </AlertAddressContext.Provider>
         )}
-        {props.enabledFeatureFlags.includes("MonthlyActivityEmail") &&
```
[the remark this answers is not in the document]
Vinnl: Yep! I removed it as part of this PR, since it's enabled in prod now. We still have `MonthlyReportFreeUser` though.

src/emails/StorybookEmailRenderer.tsx
```diff
@@ -30,12 +34,19 @@ export const StorybookEmailRenderer = (props: Props) => {
...
+          __html:
+            renderResult.errors.length > 0
+              ? `
+                <h1>MJML rendering errors:</h1>
+                <ul style="font-family: monospace;">${renderResult.errors.map((error) => `<li>${error.message}</li>`).join("\n")}</ul>
+                <hr/>
+                ${renderResult.html}
+                `
+              : renderResult.html,
         }}
```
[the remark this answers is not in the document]
Vinnl: Yep! Otherwise there's no way to see that MJML doesn't like something until you get to the part where you're actually sending the email.

src/db/tables/subscribers.ts
```diff
@@ -429,6 +414,85 @@ async function getSubscribersWaitingForMonthlyEmail(
...
+
+  if (Array.isArray(flag.allow_list) && flag.allow_list.length > 0) {
+    // If the feature flag has an allowlist, only send to users on that list.
+    // The `.andWhereIn` alias doesn't exist:
+    // https://github.com/knex/knex/issues/1881#issuecomment-275433906
+    query = query.whereIn("subscribers.primary_email", flag.allow_list);
+  }
+  const rows = await query;
+
+  return rows;
```
rhelmer, line 492: I kind of like this pattern vs. the usual `res[0]`, makes it a bit clearer what's being returned.

src/scripts/cronjobs/monthlyActivityFree.tsx
```diff
@@ -0,0 +1,151 @@
...
+          dataSummary={data}
+          l10n={l10n}
+          unsubscribeLink={unsubscribeLink}
+        />,
+      ),
+    );
+  } catch (e) {
+    console.error("send_monthly_activity_email_free", {
+      exception: e,
+    });
```
rhelmer, line 113: This probably won't show up in GCP error logs the way you want - it'd be better to use `logger.error` (from `/src/app/functions/server/logging.ts`)

src/scripts/cronjobs/monthlyActivityFree.tsx
```diff
@@ -0,0 +1,151 @@
...
+      sub = await updateEmailPreferenceForSubscriber(subscriber.id, true, {
+        unsubscribe_token: newUnsubToken,
+      });
+    }
+
+    return `${process.env.SERVER_URL}/unsubscribe-email/monthly-report-free?token=${sub.unsubscribe_token}`;
+  } catch (e) {
+    console.error("generate_unsubscribe_link", {
+      exception: e as string,
+    });
```
rhelmer, line 147: Same here re: logger (this uses [Winston](https://github.com/winstonjs/winston) and it understands [structured logging](https://cloud.google.com/logging/docs/structured-logging))

src/app/api/utils/email.tsx
rhelmer: Do you know why this is the case? I remember we used to have to do that `localstorage` workaround but that's been removed.

src/db/tables/subscribers.ts
rhelmer: Should this be in `constants.ts`?

src/db/tables/subscribers.ts
rhelmer: Yes please add this in a comment. Let me know if you have problems or need help optimizing this, it might help to look at the SQL generated here. Or we could just use the GCP SQL debugging tools to inspect once it lands on stage (we can also see the actual generated SQL there too).

rhelmer (approved):
I have a few questions and comments, up to you but I strongly suggest using `logger.*` over `console.*` for backend code (both in Node and standalone things like cronjobs).

Two things we run into very frequently are:
1. console messages spanning multiple lines, which makes it more work to piece together
2. Objects (e.g. `Error`) not being serialized to text correctly and being logged as `[Object]` which is not very useful.

src/scripts/cronjobs/monthlyActivityFree.tsx
```diff
@@ -0,0 +1,151 @@
...
+      exception: e,
+    });
+  }
+}
+
+export async function getMonthlyActivityFreeUnsubscribeLink(
+  subscriber: SubscriberRow,
+) {
+  try {
+    const newUnsubToken = randomBytes(64).toString("hex");
```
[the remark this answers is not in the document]
Vinnl: Hmm, I inlined it because it was just a single line and only used here, but I see you've created and merged https://github.com/mozilla/blurts-server/pull/5149 since. Isn't that column supposed to be nullable - i.e. why do we even need to provide a default value there?

src/app/api/utils/email.tsx
Vinnl: Yes, this is the error message:

```
ReferenceError: require is not defined in ES module scope, you can use import instead
    at <anonymous> (/<snip>/blurts-server/src/app/functions/l10n/serverComponents.ts:23:30)
```

Specifically, [we use a Webpack API](https://github.com/mozilla/blurts-server/blob/ba1dfdebe5fefad6edad50a32da78fb838b9494a/src/app/functions/l10n/serverComponents.ts#L21-L23) (which will work both on the server-side as well as on the client-side) to load FTL files in Next.js, but that API is not available in cronjobs. Thus, in cronjobs, we just [use the plain Node APIs](https://github.com/mozilla/blurts-server/blob/ba1dfdebe5fefad6edad50a32da78fb838b9494a/src/app/functions/l10n/cronjobs.ts#L34).

(Differences between runtimes like these are why I used to argue for splitting up the projects, but I think we've got it working pretty well now - but you'll still see symptoms of the way we need to work around the differences here and there, and this is one example.)

src/db/tables/subscribers.ts
Vinnl: Constants in `constants.ts` are not overridable at runtime with an environment variable; I figured this would give us some more flexibility if, after deployment, it suddenly turns out that this cutoff date isn't actually the right one.

src/scripts/cronjobs/monthlyActivityFree.tsx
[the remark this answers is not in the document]
Vinnl: I guess for a more specific question: do you think there's any risk in dc0ac3201?

src/scripts/cronjobs/monthlyActivityFree.tsx
Vinnl: Changed in f65d38b9b.

src/scripts/cronjobs/monthlyActivityFree.tsx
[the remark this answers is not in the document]
mansaj: @Vinnl I think after you resolve the conflict with main, you will see that this code has been changed

It will always generate a token instead of "" or null?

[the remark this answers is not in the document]
Vinnl: @mansaj Right, but I'm asking if that's necessary to do? Or if we can just keep the token-generation limited to the email-sending code where I have it now?

PR #5155, opened 2024-10-09
Add nightly testing job to report "true" coverage

Vinnl:
People would like to see reports of our coverage without ignore statements; this nightly job produces that.


rhelmer (approved):
This seems fine, @Vinnl did you see @pdehaan's suggestion about using `--coverageReporter=babel`? If it makes sense to stick to `c8` for other reasons that's fine with me just wanted to make sure you saw that.

Vinnl:
Yeah, I've seen that too. I'd lean towards this approach, since that doesn't require adding another dependency (babel - assuming we're not already using it elsewhere), and is consistent with how we measure coverage for new code.

(Although given the issues we've had with c8, I wouldn't be opposed to moving everything to the Babel reporter either, but that's a bit of work too. I'd assumed the c8 reported would be more accurate and more actively maintained, given that it's part of v8, but so far that doesn't really seem to be the case.)

Vinnl:
I also added 876cc5dde3895f2e03a3802f4af29fff6a43e07d, so that the `lcov.info` file is always available as [part of the CI job artefacts](https://github.com/mozilla/blurts-server/actions/runs/11330985471?pr=5155), which should be useful if the tool that collates all of those across projects is ready.

PR #5184, opened 2024-10-16
Move unsubscribe link generation into own module

Vinnl:
# References:

Jira: N/A
Figma:

# Description

It's not just used by the cronjob, but also by the admin tool on /admin/emails/. Since that is a Next.js page, if the function is imported from a module that also imports the cronjob l10n module, it breaks. Thus, I've moved it into its own module.

# How to test

Try sending an email using the admin email tool (any email). It should no longer fail.


flozia (approved):
Looks good!

PR #5185, opened 2024-10-16
MNTOR-3648: Only initialise Glean once

Vinnl:
# References:

Jira: MNTOR-3648
Figma: N/A

# Description

Instead of having every component that intends to send pings to Glean (via `useGlean` or `useTelemetry`), only some of which actually passed in an experimentation ID, with this change, we only run the initialisation code once (via the <GleanScript> component), and other components only run the ping-sending code.

@flozia I also tagged you as a reviewer because I touched some code you added - specifically, I removed the `<PublicEnvProvider>` which was only used to provide `APP_ENV` to every instance of `useGlean`, but which really only needed it for the initialisation code. And as a small fix, I updated the call to `getEnabledFeatureFlags` in one of the layouts to not pass an empty string as the user's email address.

# How to test

I'm not quite sure what the exact symptom was that caused the issue to be reported, but if I add a `console.log` to `GleanScript`'s initialisation call, I only see it being called twice (or once if I disable `reactStrictMode` in `next.config.js`), and both have the experimentation ID set correctly.

I also found a bug where different calls to `getExperimentationId` would result in different `guest-*` IDs if no cookie was set yet. You can verify this by adding a `console.log` with the ID there, and then visiting the landing page in Private Browsing. Before ae818d500, you'd see two different experimentation IDs being logged. I'd love to add an automated test for this, but I'm not sure how to, unfortunately.

- [ ] I've added a unit test to test for potential regressions of this bug. - Don't know how to, ideas welcome.
- [ ] All acceptance criteria are met. - I don't see any clear ones in the ticket.

### DIFF
--- a/src/app/components/client/GleanScript.tsx
+++ b/src/app/components/client/GleanScript.tsx
@@ -6,7 +6,6 @@
 
 import { useEffect } from "react";
 import Glean from "@mozilla/glean/web";
-import { useCookies } from "react-cookie";
 
 export type Props = {
   experimentationId: string;
@@ -14,17 +13,6 @@ export type Props = {
 };
 
 export const GleanScript = (props: Props) => {
-  const [cookies, setCookie] = useCookies(["experimentationId"]);
-
-  useEffect(() => {
-    if (
-      props.experimentationId.startsWith("guest") &&
-      !cookies.experimentationId
-    ) {
-      setCookie("experimentationId", props.experimentationId);
-    }
-  }, [cookies.experimentationId, props.experimentationId, setCookie]);
-
   // Initialize Glean only on the first render
   useEffect(() => {
     // Enable upload only if the user has not opted out of tracking.
--- a/src/app/functions/server/getExperimentationId.ts
+++ b/src/app/functions/server/getExperimentationId.ts
@@ -2,11 +2,12 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
-import { cookies } from "next/headers";
-import { UUID, randomUUID } from "crypto";
+import { headers } from "next/headers";
+import { UUID } from "crypto";
 import { Session } from "next-auth";
 import { v5 as uuidv5 } from "uuid";
 import "./notInClientComponent";
+import { logger } from "./logging";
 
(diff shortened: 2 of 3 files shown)


rhelmer:
I've been testing and it's working in Chrome but with this warning (I think this is new, but I'll double check):
```
(Glean.platform.browser.SendBeaconFallbackUploader) `sendBeacon` is not available. Falling back to the `fetch` uploader.
```

However in Firefox it's having some CSP problem:
```
Cross-Origin Request Blocked: The Same Origin Policy disallows reading the remote resource at https://incoming.telemetry.mozilla.org/submit/monitor-frontend/events/1/8efb7730-4556-48b5-887c-73e8e8626769. (Reason: CORS request did not succeed). Status code: (null).
[...]
(Glean.platform.browser.FetchUploader) Network error while attempting to upload ping.
 TypeError: NetworkError when attempting to fetch resource.
```

I do see the Chrome ping and it does have the `experimentationId` in the ping (which is the real bug here): https://debug-ping-preview.firebaseapp.com/pings/fx-monitor-local-dev/d29445ed-cf5a-4fba-b715-b559ceac62d1

@Vinnl is this working for you?

rhelmer (changes requested):
Code changes lgtm, are you seeing the same CSP problems or is it something in my setup?

Vinnl:
@rhelmer I get the same behaviour on `main` - is that the problem that I was supposed to solve in this ticket?

It's not clear to me what I'm supposed to see, and what we're currently seeing on `main`.

rhelmer:
> @rhelmer I get the same behaviour on `main` - is that the problem that I was supposed to solve in this ticket?
> 
> It's not clear to me what I'm supposed to see, and what faulty behaviour we're currently seeing on `main`.

Thanks for checking - maybe it's a problem with the local dev server. No I don't think you need to worry about it, let's just verify on stage.

rhelmer (approved):

Vinnl:
OK yeah, I've just tested in Firefox with maximally permissive preferences and with all content blocking disabled, and then I do see local pings show up in the Glean debug ping viewer. More importantly, I see that on `main`, the events don't include an experimentation ID, whereas on this branch, they do, and they're consistent: when not signed in, they all have the same experimentation ID, and after signing in, they have the same new ID. So I think this works now - though I'm still not sure what to tell QA w.r.t. how to verify this.

PR #5187, opened 2024-10-16
correct transaction isolation level for adding email

rhelmer:
# References:

Jira: MNTOR-1452

# Description

As previously discussed, this sets the isolation level to `SERIALIZABLE` so adding secondary email works as expected.

# How to test

See Jira ticket for test case.

### DIFF
--- a/src/db/tables/emailAddresses.ts
+++ b/src/db/tables/emailAddresses.ts
@@ -11,6 +11,7 @@ import { updateFxAData } from "./subscribers";
 import { ForbiddenError, UnauthorizedError } from "../../utils/error";
 import { EmailAddressRow, SubscriberRow } from "knex/types/tables";
 import { ReactLocalization } from "@fluent/react";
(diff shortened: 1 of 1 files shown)


mansaj (approved):

src/db/tables/emailAddresses.ts
```diff
@@ -43,22 +44,32 @@ async function addSubscriberUnverifiedEmailHash(
...
+            })
+            .returning("*");
+      });
   });
-  return res[0];
+
+  if (res) {
+    return res[0];
+  } else {
+    return false;
```
mansaj, line 71: Looks like you might have to change this for type check

rhelmer: Ah thanks I guess it didn't break the quick check-in hook lint run... I'll fix this up. I think the caller should probably handle it being possibly `false` (or maybe `undefined`), or we should have it throw.

rhelmer: I switched it to throw an exception.

PR #5192, opened 2024-10-17
Update fallback path for silent authentication error

flozia:
# References:

Follow-up to address https://github.com/mozilla/blurts-server/pull/5104#discussion_r1799621794.

# Description

Update the fallback path in case of a silent authentication callback error.

# How to test

Same steps as for PR #5104:
- Enable the feature flag `PromptNoneAuthFlow`
- Visit http://localhost:6060/?utm_source=moz-account&utm_campaign=settings-promo&utm_content=monitor-free with and without the user authenticated with FxA


Vinnl (approved):
Works now, thanks!

PR #5202, opened 2024-10-21
MNTOR-3707: Push source maps to Sentry

Vinnl:
When trying to trace code in Sentry, the UI shows the following:

> Missing source file with a matching Debug ID
> You already uploaded artifacts with Debug IDs but none of the
> uploaded source files had a Debug ID matching this stack frame's
> Debug ID: <snip>

Our build logs currently show the following error:

    [@sentry/nextjs - Edge] Warning: No auth token provided. Will
    not upload source maps. Please set the `authToken` option. You can
    find information on how to generate a Sentry auth token here:
    https://docs.sentry.io/api/auth/

So presumably, making the auth token available to the build job should enable that job to actually upload the source maps to Sentry.

# References:

Jira: MNTOR-3707


rhelmer:
Thanks for getting this one - I think it was previously filed as https://mozilla-hub.atlassian.net/browse/MNTOR-3494 as well.

Vinnl:
Thanks, moved that ticket into the sprint and linked it to the BLE work!

PR #5230, opened 2024-10-22
use structured logging for email utils

rhelmer:
# References:

Jira: MNTOR-3713

# Description

Use the Winston logging library from the email utils so we get [structured logging](https://cloud.google.com/logging/docs/structured-logging) in GCP. Right now log messages are split across several lines which makes it difficult to locate and link to problems.

# How to test

Our logging library automatically falls back to console logging in local environments, so the output can be inspected that way. We'll want to check on stage that the output looks acceptable.

### DIFF
--- a/jest.config.cjs
+++ b/jest.config.cjs
@@ -113,7 +113,10 @@ const customJestConfig = {
   // ],
 
   // A map from regular expressions to module names or to arrays of module names that allow to stub out resources with a single module
-  // moduleNameMapper: {},
+  moduleNameMapper: {
+    // Force module uuid to resolve with the CJS entry point, because Jest does not support package.json.exports. See https://github.com/uuidjs/uuid/issues/451
+    uuid: require.resolve("uuid"),
+  },
 
   // An array of regexp pattern strings, matched against all module paths before considered 'visible' to the module loader
   modulePathIgnorePatterns: ["e2e/"],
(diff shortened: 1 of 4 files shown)


jest.config.cjs
```diff
@@ -113,7 +113,10 @@ const customJestConfig = {
   // ],
 
   // A map from regular expressions to module names or to arrays of module names that allow to stub out resources with a single module
-  // moduleNameMapper: {},
+  moduleNameMapper: {
```
Vinnl, line 116: Weird, https://github.com/uuidjs/uuid/issues/451#issuecomment-1112329871 seems to imply this shouldn't be necessary. Maybe because we're using `jest.config.cjs` (i.e. a CommonJS config)?

Thanks for the comment anyway :)

src/utils/email.test.ts
```diff
@@ -12,6 +12,25 @@ jest.mock("nodemailer", () => {
...
+    error(message: string, details: object) {
+      console.error(message, details);
+    }
+  }
+
+  const logger = new Logging();
+  return {
+    logger,
+  };
+});
```
Vinnl, line 32: Nit, but this can be simplified a bit:

```suggestion
jest.mock("../app/functions/server/logging", () => {
  return {
    logger: {
      info: console.info,
      warn: console.warn,
      error: console.error,
    },
  };
});
```

Vinnl (approved):
Nice!

PR #5257, opened 2024-10-28
Use a singleton for the DB connection object

Vinnl:
# References:

Jira: MNTOR-3606
Figma:

# Description

This is the riskier sibling to https://github.com/mozilla/blurts-server/pull/5254, where I'm trying to figure out what could be causing the monthly-activity-free cronjob to be hanging.

In this PR, I call `destroy()` on the connection object when it's done - but more importantly, I have every `db/tables` module re-use the same connection object, to avoid having to do this: https://github.com/mozilla/blurts-server/blob/0da029654e65dedffe2b3bd169d1160626f2c11b/src/scripts/cronjobs/emailBreachAlerts.tsx#L427-L429

I think this should be safe, and in practice, the connections should be pooled already, but there might always be unforeseen effects, so please take some time to think about potential problems and/or ways to catch any that might occur.

### DIFF
--- a/src/db/connect.ts
+++ b/src/db/connect.ts
@@ -5,6 +5,7 @@
 import knex from "knex";
 import knexConfig from "./knexfile.js";
 
+let connection: knex.Knex;
 export default function createDbConnection(): knex.Knex {
   /* c8 ignore start */
   if (process.env.NODE_ENV === "development") {
@@ -15,9 +16,11 @@ export default function createDbConnection(): knex.Knex {
       );
     }
 
-    return (global as unknown as Record<string, knex.Knex>)[client];
+    connection ??= (global as unknown as Record<string, knex.Knex>)[client];
+    return connection;
   }
   /* c8 ignore stop */
 
-  return knex(knexConfig as knex.Knex.Config);
+  connection ??= knex(knexConfig as knex.Knex.Config);
+  return connection;
 }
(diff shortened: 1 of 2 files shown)


rhelmer (approved):
Yes I think this is a better approach here. Could you please file a ticket to refactor the rest of the code to use this approach?

Vinnl:
@rhelmer 

> Could you please file a ticket to refactor the rest of the code to use this approach?

The risky bit is that this already ensures all `/db/tables/*` files to re-use that singleton (because they call the `createDbConnection` function that I modified), so I think the rest of the code already is adapted. Or do you mean updating the rest of the .destroy() calls?

rhelmer:
> @rhelmer
> 
> > Could you please file a ticket to refactor the rest of the code to use this approach?
> 
> The risky bit is that this already ensures all `/db/tables/*` files to re-use that singleton (because they call the `createDbConnection` function that I modified), so I think the rest of the code already is adapted. Or do you mean updating the rest of the .destroy() calls?

Yes thinking of e.g. https://github.com/mozilla/blurts-server/blob/c2e3402e06bbc64b7e564ebe90e1fee405e49434/src/scripts/cronjobs/emailBreachAlerts.tsx#L427-L429 - I wonder if we could expose this destroy method to the scripts so they have a simpler way to make sure everything is closed.

PR #5258, opened 2024-10-29
MNTOR-3494: Upload Sentry source maps by default

Vinnl:
Alternatively, I could make sure that we're passing in `UPLOAD_SENTRY_SOURCEMAPS` in CI, but since I assume we'll generally want to upload the source maps always, doing so by default seems more natural.

### DIFF
--- a/next.config.js
+++ b/next.config.js
@@ -205,7 +205,7 @@ const sentryOptions = {
   hideSourceMaps: false,
 
   sourcemaps: {
-    disable: process.env.UPLOAD_SENTRY_SOURCEMAPS !== "true",
+    disable: process.env.UPLOAD_SENTRY_SOURCEMAPS === "false",
   },
 };
 


next.config.js
```diff
@@ -205,7 +205,7 @@ const sentryOptions = {
   hideSourceMaps: false,
 
   sourcemaps: {
-    disable: process.env.UPLOAD_SENTRY_SOURCEMAPS !== "true",
```
rhelmer, line 208: I think this is fine, IIRC it will log a warning saying what you need to do (e.g. set the token) but should not break the build.

rhelmer (approved):

next.config.js
Vinnl: Ah right, for people who don't have the token, having it on by default will log the error, that makes sense. But agreed that that's no biggie. If we really want to disable the warning, we can also just trigger `disable` based on whether the token is available - but I think it'd be good to have the warning in our CI logs if something goes wrong with the env var.

PR #5291, opened 2024-11-07
MNTOR-3776: cronjob to pull in the data brokers and refresh (part 2)

mansaj:
# References:

Jira: MNTOR-3776

# Description
Part 2 of the ticket: cronjob to pull in the data brokers and refresh
# How to test
`npm run dev:cron:db-pull-data-brokers`


src/scripts/cronjobs/syncOnerepDataBrokers.ts
```diff
@@ -0,0 +1,60 @@
...
+} from "../../db/redis/client.js";
+
+const SENTRY_SLUG = "cron-sync-onerep-data-brokers";
+
+Sentry.init({
+  dsn: process.env.SENTRY_DSN,
+  tracesSampleRate: 1.0,
+});
+
+const checkInId = Sentry.captureCheckIn({
```
rhelmer, line 27: It's not harmful but we no longer have this service enabled in Sentry FYI

rhelmer (approved):

rhelmer (commented):
I found an issue while testing this, in https://github.com/mozilla/blurts-server/blob/ba9c0e4a10a1f797dc13aae5252f1f4e530ab250/src/app/functions/server/onerep.ts#L427 where `per_page` is hardcoded.

`getAllDataBrokers` function is already using `fetchAllPages` so I think we just need pass the current/total pages correctly. I'll add that to this PR.

rhelmer:
> I found an issue while testing this, in
> 
> https://github.com/mozilla/blurts-server/blob/ba9c0e4a10a1f797dc13aae5252f1f4e530ab250/src/app/functions/server/onerep.ts#L427
> where `per_page` is hardcoded.
> 
> `getAllDataBrokers` function is already using `fetchAllPages` so I think we just need pass the current/total pages correctly. I'll add that to this PR.

Oh actually I take that back, it's just a bit confusing to read here but `per_page` is `100` and `page` is passed correctly, so this should be fine.

PR #5307, opened 2024-11-13
MNTOR-3380 - pass nimbus_preview param to Cirrus

rhelmer:
# References:

Jira: MNTOR-3380/

# Description

The Nimbus team came up with an ADR to support enabling preview mode on specific page loads. We previously had a workaround in place that would simply enable all experiments when an HTTP parameter like `?nimbus_web_preview=true` was passed, this change really just removes that and does two things:

1. change the param name to `nimbus_preview`
2. pass the param on to the Cirrus sidecar

This will instruct Cirrus to check which experiments are in preview mode on the server, and only enable these.

# How to test

See ticket. It's probably easiest to merge this to stage and test there, but we can at least verify locally that `nimbus_preview=true` is appended to the URL set in `process.env.NIMBUS_SIDECAR_URL`.

### DIFF
--- a/src/app/functions/server/getExperiments.ts
+++ b/src/app/functions/server/getExperiments.ts
@@ -35,13 +35,13 @@ export async function getExperiments(params: {
   if (!process.env.NIMBUS_SIDECAR_URL) {
     throw new Error("env var NIMBUS_SIDECAR_URL not set");
   }
-  const serverUrl = new URL(process.env.NIMBUS_SIDECAR_URL);
+  const serverUrl = new URL(`${process.env.NIMBUS_SIDECAR_URL}/v1/features`);
 
   try {
     if (params.previewMode === true) {
       serverUrl.searchParams.set("nimbus_preview", "true");
     }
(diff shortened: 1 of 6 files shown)


src/app/functions/server/getExperiments.ts
```diff
@@ -28,30 +28,21 @@ export async function getExperiments(params: {
...
     throw new Error("env var NIMBUS_SIDECAR_URL not set");
   }
+  const serverUrl = new URL(process.env.NIMBUS_SIDECAR_URL);
 
   try {
-    const response = await fetch(`${serverUrl}/v1/features`, {
+    if (params.previewMode === true) {
+      serverUrl.searchParams.set("nimbus_preview", "true");
+    }
+    serverUrl.pathname = "/v1/features";
```
flozia, line 44: Question/suggestion: Why not constructing the URL directly with the pathname?
```code
const serverUrl = new URL(`${process.env.NIMBUS_SIDECAR_URL}/v1/features`);
```

flozia (approved):
Looks good!

PR #5352, opened 2024-12-02
Upgrade storybook version

flozia:
[Upgrading Storybook](https://storybook.js.org/docs/configure/upgrading) to unblock bumping the Next.js dependency group (PR #5294).


rhelmer (approved):
Thanks!

PR #5366, opened 2024-12-04
MNTOR-3840 - Action needed tab

codemist:
# References:

Jira: MNTOR-
Figma:

# Description

# Screenshot (if applicable)

Not applicable.

# How to test

### DIFF
--- a/src/app/(proper_react)/(redesign)/(authenticated)/user/(dashboard)/dashboard/[[...slug]]/page.tsx
+++ b/src/app/(proper_react)/(redesign)/(authenticated)/user/(dashboard)/dashboard/[[...slug]]/page.tsx
@@ -102,7 +102,10 @@ export default async function DashboardPage({ params, searchParams }: Props) {
     return redirect("/user/welcome");
   }
 
-  const latestScan = await getScanResultsWithBroker(profileId);
+  const latestScan = await getScanResultsWithBroker(
+    profileId,
+    hasPremium(session.user),
+  );
 
   const scanCount =
     typeof profileId === "number"
--- a/src/app/(proper_react)/(redesign)/(authenticated)/admin/emails/actions.tsx
+++ b/src/app/(proper_react)/(redesign)/(authenticated)/admin/emails/actions.tsx
@@ -142,6 +142,7 @@ export async function triggerMonthlyActivityFree(emailAddress: string) {
   }
   const latestScan = await getScanResultsWithBroker(
     subscriber.onerep_profile_id,
+    hasPremium(session.user),
   );
   const data = getDashboardSummary(
     latestScan.results,
@@ -180,6 +181,7 @@ export async function triggerMonthlyActivityPlus(emailAddress: string) {
   }
   const latestScan = await getScanResultsWithBroker(
     subscriber.onerep_profile_id,
+    hasPremium(session.user),
   );
   const data = getDashboardSummary(
     latestScan.results,
@@ -217,7 +219,10 @@ export async function triggerBreachAlert(
   if (typeof subscriber.onerep_profile_id === "number") {
     await refreshStoredScanResults(subscriber.onerep_profile_id);
   }
-  const scanData = await getScanResultsWithBroker(subscriber.onerep_profile_id);
+  const scanData = await getScanResultsWithBroker(
+    subscriber.onerep_profile_id,
+    hasPremium(session.user),
+  );
   const allSubscriberBreaches = await getSubscriberBreaches({
     fxaUid: subscriber.fxa_uid,
     countryCode: assumedCountryCode,
--- a/src/app/(proper_react)/(redesign)/(authenticated)/user/(dashboard)/dashboard/fix/data-broker-profiles/automatic-remove/page.tsx
+++ b/src/app/(proper_react)/(redesign)/(authenticated)/user/(dashboard)/dashboard/fix/data-broker-profiles/automatic-remove/page.tsx
@@ -21,6 +21,7 @@ import {
   getPremiumSubscriptionUrl,
 } from "../../../../../../../../../functions/server/getPremiumSubscriptionInfo";
 import { getAttributionsFromCookiesOrDb } from "../../../../../../../../../functions/server/attributions";
+import { hasPremium } from "../../../../../../../../../functions/universal/user";
 
 const monthlySubscriptionUrl = getPremiumSubscriptionUrl({ type: "monthly" });
 const yearlySubscriptionUrl = getPremiumSubscriptionUrl({ type: "yearly" });
@@ -38,7 +39,10 @@ export default async function AutomaticRemovePage() {
 
   const countryCode = getCountryCode(headers());
   const profileId = await getOnerepProfileId(session.user.subscriber.id);
-  const scanData = await getScanResultsWithBroker(profileId);
+  const scanData = await getScanResultsWithBroker(
+    profileId,
+    hasPremium(session.user),
+  );
   const subBreaches = await getSubscriberBreaches({
     fxaUid: session.user.subscriber.fxa_uid,
     countryCode,
--- a/src/app/(proper_react)/(redesign)/(authenticated)/user/(dashboard)/dashboard/fix/data-broker-profiles/manual-remove/page.tsx
+++ b/src/app/(proper_react)/(redesign)/(authenticated)/user/(dashboard)/dashboard/fix/data-broker-profiles/manual-remove/page.tsx
@@ -23,7 +23,10 @@ export default async function ManualRemovePage() {
 
   const countryCode = getCountryCode(headers());
   const profileId = await getOnerepProfileId(session.user.subscriber.id);
-  const scanData = await getScanResultsWithBroker(profileId);
+  const scanData = await getScanResultsWithBroker(
+    profileId,
+    hasPremium(session.user),
+  );
   const subBreaches = await getSubscriberBreaches({
     fxaUid: session.user.subscriber.fxa_uid,
     countryCode,
--- a/src/app/(proper_react)/(redesign)/(authenticated)/user/(dashboard)/dashboard/fix/data-broker-profiles/removal-under-maintenance/page.tsx
+++ b/src/app/(proper_react)/(redesign)/(authenticated)/user/(dashboard)/dashboard/fix/data-broker-profiles/removal-under-maintenance/page.tsx
@@ -27,7 +27,10 @@ export default async function RemovalUnderMaintenance() {
     redirect("/user/dashboard");
   }
   const profileId = await getOnerepProfileId(session.user.subscriber.id);
-  const latestScan = await getScanResultsWithBroker(profileId);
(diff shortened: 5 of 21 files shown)


src/app/functions/universal/dataBroker.ts
```diff
@@ -0,0 +1,17 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+export type DataBrokerRemovalStatus =
+  | "active"
+  | "on_hold"
+  | "ceased_operation"
+  | "scan_under_maintenance"
+  | "removal_under_maintenance";
```
flozia: Nit: `DataBrokerRemovalStatus` could also be derived from `DataBrokerRemovalStatusMap`:

```suggestion
export type DataBrokerRemovalStatus = (typeof DataBrokerRemovalStatusMap)[keyof typeof DataBrokerRemovalStatusMap];
```

src/app/functions/universal/scanResult.ts
```diff
@@ -4,6 +4,7 @@
 
 // TODO: Move pure functions that operate on scan results to this file
 
+// TODO: removal_under_maintenance does not belong here
```
flozia: Question: Should we remove it then or at least link an issue for us to take care of it?

src/app/components/server/StatusPill.tsx
```diff
@@ -24,71 +24,85 @@ export const StatusPillTypeMap: Record<string, StatusPillType> = {
...
+
+  // Determine pill type and status label based on props
+  const pillType = props.directType
+    ? props.directType
+    : props.exposure
+      ? getExposureStatus(
+          props.exposure,
+          props.enabledFeatureFlags?.includes("AdditionalRemovalStatuses") ??
+            false,
+          props.isRemovalUnderMaintenance || false,
```
flozia, line 48: Rather that using `||` we should explicitly check if the prop is `undefined`:

```suggestion
          typeof props.isRemovalUnderMaintenance !== "undefined",
```

src/app/(proper_react)/(redesign)/(authenticated)/user/(dashboard)/dashboard/[[...slug]]/page.tsx
```diff
@@ -103,6 +104,9 @@ export default async function DashboardPage({ params, searchParams }: Props) {
   }
 
   const latestScan = await getLatestOnerepScanResults(profileId);
+  const dataBrokersWithRemovalUnderMaintenance =
+    await getScanResultsWithBrokerUnderMaintenance(profileId);
```
Comment (flozia) [inline src/app/(proper_react)/(redesign)/(authenticated)/user/(dashboard)/dashboard/[[...slug]]/page.tsx:108]: @mansaj @codemist I might be a bit late to the discussion, but I’m wondering what your thoughts are on this:

The naming `getScanResultsWithBrokerUnderMaintenance` seems to be a bit misleading. We are only fetching the scan results that are under maintenance here. We have to do this in multiple places all over the application and I’m wondering if we should not have an endpoint that gives us the results with the correct status that we are caring about. Otherwise, we need to fetch and check in place which result has a status mismatch.

src/app/(proper_react)/(redesign)/(authenticated)/user/(dashboard)/dashboard/View.tsx
```diff
@@ -171,12 +176,20 @@ export const View = (props: Props) => {
     return timestampB - timestampA;
   });
 
+  const dataBrokersResultsRemovalUnderMaintenance =
+    props.userScanData.dataBrokersRemovalUnderMaintenance || [];
```
flozia, line 180: It is safer to use the `??` to check if `props.userScanData.dataBrokersRemovalUnderMaintenance` is `null` or `undefined`.

```suggestion
    props.userScanData.dataBrokersRemovalUnderMaintenance ?? [];
```

src/app/components/server/StatusPill.tsx
```diff
@@ -24,71 +24,97 @@ export const StatusPillTypeMap: Record<string, StatusPillType> = {
...
```
flozia, line 48: Could we break this out to not have a nested ternary condition? I think introducing the extra prop `directType` should not be needed if we already pass `isRemovalUnderMaintenance` to `getExposureStatus`.

Additionally, I’m wondering why we can’t check `removal_under_maintenance` on the exposure as we did before — or should we remove the status check in the switch statement?

src/app/components/server/StatusPill.tsx
```diff
@@ -24,71 +24,97 @@ export const StatusPillTypeMap: Record<string, StatusPillType> = {
...
 
-type DirectTypeProps = { type: StatusPillType };
-type ExposureProps = { exposure: Exposure };
-export type Props = (DirectTypeProps | ExposureProps) & {
+export type Props = {
+  directType?: StatusPillType;
   enabledFeatureFlags?: FeatureFlagName[];
   note?: string;
+  exposure?: Exposure;
+  isRemovalUnderMaintenance?: boolean;
```
flozia, line 32: Proposal: Instead of changing and adding specific logic to the `StatusPill` component for items under removal: Could we not just leave the component unchanged and pass the type `removal_under_maintenance` for those exposures?

src/app/functions/universal/scanResult.ts
mansaj: Yea it should be removed

mansaj: The status is present in the other file `dataBroker.ts`

src/app/(proper_react)/(redesign)/(authenticated)/user/(dashboard)/dashboard/[[...slug]]/page.tsx
Comment (mansaj) [reply src/app/(proper_react)/(redesign)/(authenticated)/user/(dashboard)/dashboard/[[...slug]]/page.tsx:108]: @flozia are you suggesting we merge this function into the existing GET call for `getLatestOnerepScanResults`?

I don't know if we'd want to do that because we might want to just a "clean" GET for all the scan results for a user

Comment (mansaj) [reply src/app/(proper_react)/(redesign)/(authenticated)/user/(dashboard)/dashboard/[[...slug]]/page.tsx:108]: Alternatively, we can also have a more barebone function that just fetches the list of all data broker statuses from Redis (fallback to db), but then the frontend would have to do even more comparison / logic

src/app/(proper_react)/(redesign)/(authenticated)/admin/qa-customs/onerepConfig.tsx
```diff
@@ -380,9 +380,6 @@ const OnerepConfigPage = (props: Props) => {
                   <option value="new">New</option>
                   <option value="optout_in_progress">In Progress</option>
                   <option value="removed">Removed</option>
-                  <option value="removal_under_maintenance">
```
mansaj, line 383: we are no longer doing this right?

src/app/(proper_react)/(redesign)/(authenticated)/user/(dashboard)/dashboard/[[...slug]]/page.tsx
Comment (mansaj) [reply src/app/(proper_react)/(redesign)/(authenticated)/user/(dashboard)/dashboard/[[...slug]]/page.tsx:108]: Oh, another alternative is to have a function that doesn't do the filtering at all, but just do the JOIN, then just let the frontend have the entire joined scan results (with databroker statuses) and do the filtering whenever/however needed

This might actually be the most flexible (and equally performant) option

@rhelmer @flozia @codemist  wdyt

Comment (flozia) [reply src/app/(proper_react)/(redesign)/(authenticated)/user/(dashboard)/dashboard/[[...slug]]/page.tsx:108]: @mansaj I thought that we could maybe not merge it into the existing `getLatestOnerepScanResults`, but rather have one (additional function) that has the joined scan results that we can rely on to have the relevant statuses for the frontend. I like the suggestion of not filtering the results after fetching them as suggested. What do you think @codemist — would this make things easier or more complicated for how the manual broker removal is intended to work?

src/app/components/server/StatusPill.tsx
flozia: But couldn’t we (instead of adding additional logic to the status pill), keep it unchanged and pass `type={removal_under_maintenance}` for broker results under removal?

src/app/(proper_react)/(redesign)/(authenticated)/user/(dashboard)/dashboard/fix/data-broker-profiles/removal-under-maintenance/RemovalUnderMaintenanceView.tsx
```diff
@@ -46,6 +46,7 @@ export const RemovalUnderMaintenanceView = (props: Props) => {
     "DataBrokerManualRemoval",
   );
 
+  /* c8 ignore start */
```
flozia, line 49: Looks like we have some missing unit tests here …

src/app/(proper_react)/(redesign)/(authenticated)/user/(dashboard)/dashboard/fix/data-broker-profiles/removal-under-maintenance/RemovalUnderMaintenanceView.tsx
```diff
@@ -138,19 +140,26 @@ export const RemovalUnderMaintenanceView = (props: Props) => {
     );
   }
 
+  {
+    /* TODO: Add tests here */
+  }
+  {
+    /* c8 ignore start */
+  }
```
flozia, line 148: … and here. :)

src/app/functions/server/getRelevantGuidedSteps.ts
```diff
@@ -138,10 +137,13 @@ export function isEligibleForStep(
 ): boolean {
   // Only premium users can see the manual data broker removal flow, once they have run a scan
   if (stepId === "DataBrokerManualRemoval") {
-    return (
-      hasPremium(data.user) &&
-      typeof data.user.subscriber?.onerep_profile_id === "number"
-    );
+    const dataBrokersRequireManualRemoval =
+      /* c8 ignore next */
```
flozia, line 141:
```suggestion
```

flozia (commented):
Great job incorporating the new API! There is some commented out code that we should take care of — looks good otherwise.

src/apiMocks/mockData.ts
```diff
@@ -40,17 +41,12 @@ export type RandomScanResultOptions = Partial<{
...
   manually_resolved: boolean;
+  broker_status: DataBrokerRemovalStatus;
 }>;
 
-/**
- * Generates scan result with randomly-generated mock data.
- *
- * @param options
- * @returns A single scan result.
- */
```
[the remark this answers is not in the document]
flozia: Reintroduced [here](https://github.com/mozilla/blurts-server/pull/5395/files#diff-7c4dd2fa921d79056e3e84c00aaa44e9e157649f10f91aa0ba4850920d047f98R47-R52).

flozia (approved):

PR #5369, opened 2024-12-05
ensure that SENTRY_DSN and SENTRY_RELEASE are declared at build time

rhelmer:
# References:

Jira: MNTOR-3853

# Description

While debugging a problem I noticed that Sentry doesn't seem to be sending front-end data from stage/prod, I believe it's because we're not setting `NEXT_PUBLIC_SENTRY_DSN` at build time in the docker environment.

# How to test

I think building the docker image locally is the easiest way to tell, we should be able to tell by inspecting that if this value was set or not. Practically, it's probably easiest to merge and check stage.


rhelmer:
Just to simplify things, I declared `NEXT_PUBLIC_SENTRY_DSN` in the GHA env vars, and just used that name consistently instead of having to juggle `NEXT_PUBLIC_SENTRY_DSN` vs. `SENTRY_DSN`.

mansaj (approved):

PR #5395, opened 2024-12-10
Additional unit test coverage for “Removal under maintenance”

flozia:
Add additional unit test coverage for PR #5366.

### DIFF
--- a/src/utils/fetchWithDelay.test.ts
+++ b/src/utils/fetchWithDelay.test.ts
@@ -19,15 +19,11 @@ describe("fetchWithDelay", () => {
   });
 
   it("throws an error", async () => {
-    global.fetch = jest.fn().mockRejectedValueOnce("fetch failed");
-    try {
-      expect(
-        await fetchWithDelay(`${faker.internet.url()}/api/test`, {
-          delay: 750,
-        }),
-      ).toThrow("fetch failed");
-    } catch (_) {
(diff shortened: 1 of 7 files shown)


src/utils/fetchWithDelay.test.ts
```diff
@@ -0,0 +1,33 @@
...
+
+  it("throws an error", async () => {
+    global.fetch = jest.fn().mockRejectedValueOnce("fetch failed");
+    try {
+      expect(
+        await fetchWithDelay(`${faker.internet.url()}/api/test`, {
+          delay: 750,
+        }),
+      ).toThrow("fetch failed");
+    } catch (_) {
```
rhelmer, line 29: Hm, I'm surprised this is necessary... does the `expect(...).toThrow()` throw in this case? Or is this for some other unexpected error? Might want to log it just in case something unrelated happens.

rhelmer (approved):

src/utils/fetchWithDelay.test.ts
flozia: Now you are mentioning it — I was surprised as well. I’ve updated the test to work as expected: [dce81dd](https://github.com/mozilla/blurts-server/pull/5395/commits/dce81ddc7c9da0b45d42388ebadc53e97f502f42).

PR #5431, opened 2024-12-18
MNTOR-3857 - move locale filtering from TS to SQL for monthly free activity report

rhelmer:
# References:

Jira: MNTOR-3857

# Description

Our cronjob container is running out of memory trying to fetch data and filter in the application side, specifically it wants to extract the country code (if present) from the [locale header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Accept-Language) which can be pretty complex.

We can have the database do this filtering in a much more efficient way, I think it's worth revisiting this to simplify the query but the minimally-invasive fix I could figure out is:

1. use nested `CASE / ELSE` to split the locale into `country_code`
2. use a subquery so we can do a `WHERE IN` on `country_code`

The `batchSize` is now passed as an argument to the DB function and used directly as a `LIMIT` in the query, and the list of acceptable country codes is also passed as a `string[]` argument.

# How to test

Moving this to SQL makes it more difficult to unit test, I ended up modifying my local database and testing that way.

We should continue to explore ways to test our SQL in controlled environments, especially as it gets more complicated. This probably should be refactored into a stored procedure, for instance, which we could more feasibly unit test.

### DIFF
--- a/src/db/tables/subscribers.ts
+++ b/src/db/tables/subscribers.ts
@@ -515,7 +515,7 @@ async function getFreeSubscribersWaitingForMonthlyEmail(
   }
 
   const wrappedQuery = knex
-    // @ts-ignore TODO update our knex types
+    // @ts-ignore TODO MNTOR-3890 Move away from this approach and simplify query.
     .from({ base_query: query }) // Use the existing query as a subquery
     .select("*")
     .whereIn("country_code", countryCodes)
--- a/src/db/tables/subscribers.ts
+++ b/src/db/tables/subscribers.ts
@@ -520,7 +520,6 @@ async function getFreeSubscribersWaitingForMonthlyEmail(
     .whereIn("country_code", countryCodes)
     .limit(batchSize);
 
-  console.debug(wrappedQuery.toQuery());
   const rows = await wrappedQuery;
 
   return rows;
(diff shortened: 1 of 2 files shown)


rhelmer:
Here's an example of the generated SQL:
```
select * from (select "subscribers".*, CASE
        WHEN (fxa_profile_json->>'locale') ~ ',' THEN
          CASE
            WHEN split_part(fxa_profile_json->>'locale', ',', 1) ~ '-' THEN
              split_part(split_part(fxa_profile_json->>'locale', ',', 1), '-', 2) -- Extract country code from first part
            ELSE
              split_part(fxa_profile_json->>'locale', ',', 1) -- Fallback to the language code
          END
        WHEN (fxa_profile_json->>'locale') ~ '-' THEN
          split_part(fxa_profile_json->>'locale', '-', 2) -- Extract country code if present
        ELSE
          fxa_profile_json->>'locale' -- Fallback to the language code
      END AS country_code from "subscribers" left join "subscriber_email_preferences" on "subscribers"."id" = "subscriber_email_preferences"."subscriber_id" where ("monthly_monitor_report_free" is null or "monthly_monitor_report_free" = true) and ("monthly_monitor_report_free_at" is null or "monthly_monitor_report_free_at" < NOW() - INTERVAL '1 month') and subscribers."created_at" < NOW() - INTERVAL '1 month' and "subscribers"."created_at" >= '2021-12-31 16:00:00.000' and (NOT (subscribers.fxa_profile_json)::jsonb ? 'subscriptions' or NOT (subscribers.fxa_profile_json->'subscriptions')::jsonb ? 'monitor')) as "base_query" where "country_code" in ('US') limit 5
```

mansaj (approved):

PR #5432, opened 2024-12-19
MNTOR-3812 - switch to Cirrus v2 output

rhelmer:
# References:

Jira: MNTOR-3812

# Description

Switch to the Cirrus v2 output. This requires changing the code that generates types based on the `nimbus.yaml`.

This is a bit cumbersome to do by hand, we should consider using some pre-existing code-generation tool for
this, but I think it works well enough for now. I can file a followup tech debt ticket for that.

# How to test

Cirrus is used by just about every part of Monitor including the front page, so it's hard to miss. We also have unit test coverage.

### DIFF
--- a/src/app/(proper_react)/(redesign)/(authenticated)/user/(dashboard)/dashboard/View.tsx
+++ b/src/app/(proper_react)/(redesign)/(authenticated)/user/(dashboard)/dashboard/View.tsx
@@ -62,7 +62,7 @@ export type TabType = "action-needed" | "fixed";
 
 export type Props = {
   enabledFeatureFlags: FeatureFlagName[];
-  experimentData: ExperimentData;
+  experimentData: ExperimentData["Features"];
   user: Session["user"];
   userBreaches: SubscriberBreach[];
   userScanData: LatestOnerepScanData;
--- a/src/scripts/build/nimbusTypes.js
+++ b/src/scripts/build/nimbusTypes.js
@@ -4,6 +4,7 @@
 
 import { mkdir, readFile, writeFile } from "node:fs/promises";
 import { parse } from "yaml";
+import * as prettier from "prettier";
 
 run();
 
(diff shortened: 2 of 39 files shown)


src/scripts/build/nimbusTypes.js
```diff
@@ -91,7 +91,19 @@ function getFallbackObject(nimbusConfig) {
...
+      ${featureFallbackDefs.join("\n")}
+    },
+    "Enrollments": {
+      "nimbus_user_id": "-1",
+      "app_id": "-1",
+      "experiment": "-1",
+      "branch": "-1",
+      "experiment_type": "-1",
+      "is_preview": false
+    }`;
```
flozia, line 105: The newly added string literal makes the resulting `experiment.ts` harder to grasp. Would be nice if we could fix the formatting for `"Features"` and `"Enrollments"`. Here is an example of the current output:

```
export const defaultExperimentData: ExperimentData = {

    "Features": {
        "last-scan-date": {
    "enabled": false,
  },

  "welcome-scan-optional-info": {
    "enabled": false,
    "variant": "none",
  },

  "data-privacy-petition-banner": {
    "enabled": false,
  },

  "landing-page-free-scan-cta": {
    "enabled": false,
    "variant": "ctaWithEmail",
  },

  "data-broker-removal-time-estimates": {
    "enabled": false,
  },

    },
    "Enrollments": {
      "nimbus_user_id": "-1",
      "app_id": "-1",
      "experiment": "-1",
      "branch": "-1",
      "experiment_type": "-1",
      "is_preview": false
    }};
```

src/app/(proper_react)/(redesign)/(authenticated)/user/(dashboard)/dashboard/View.tsx
```diff
@@ -62,7 +62,7 @@ export type TabType = "action-needed" | "fixed";
 
 export type Props = {
   enabledFeatureFlags: FeatureFlagName[];
-  experimentData: ExperimentData;
+  experimentData: ExperimentData["Features"];
```
flozia, line 65: We would need to update this across the application. Currently, the linter and build is breaking because of this.

src/scripts/build/nimbusTypes.js
rhelmer: Thanks, yeah I think we should use some smarter tooling for code generation... for now I can fix the formatting by hand.

rhelmer: Looked into this further - we can actually just not worry about formatting on the generation side and call `prettier` programmatically when writing the file.

flozia (approved):
Thanks for updating the formatting as well!

PR #5440, opened 2024-12-20
MNTOR-3814 - use context to fetch experiment data from Cirrus

rhelmer:
# References:

Jira: MNTOR-3814

# Description

This PR introduces an `ExperimentsContext` with a provider and a `useExperiments()` hook that can be called by any client component that needs it. I implemented it just for Glean so far (per the linked Jira ticket) but if this approach is OK I can file a followup to move from prop-drilling to this more generally for features that need access to experiment data (which features are activated, what the active branch is, etc)

# How to test

I've been using the Glean debug dashboard to inspect pings. Note that local pings have `-1` for most `Enrollments` values, you need to use Cirrus (or something that mocks it). I'll see if I can mock this to add some automated tests.

An easy way to inspect Glean pings is to run this in the browser devtools console:
```js
Glean.setLogPings(true)
```
Which should log:
```
Pings will be logged to the console until this tab is closed.
```


src/telemetry/metrics.yaml
```diff
@@ -56,6 +56,24 @@ page:
...
+        type: string
+      nimbus_branch:
+        description: Nimbus branch
+        type: string
+      nimbus_experiment_type:
+        description: Nimbus experiment type
+        type: string
+      nimbus_is_preview:
+        description: Nimbus preview mode enabled
+        type: string
```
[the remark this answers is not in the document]
rhelmer: Yes, so we may not be able to land this until that's done.

src/middleware.ts
```diff
@@ -29,6 +29,14 @@ export function middleware(request: NextRequest) {
     existingExperimentationId?.value ?? `guest-${crypto.randomUUID()}`;
   requestHeaders.set("x-experimentation-id", experimentationId);
 
+  // Check for Nimbus preview mode. Note that this requires a full page reload
+  // to activate: https://nextjs.org/docs/app/api-reference/file-conventions/layout#caveats
+  const nimbusPreviewMode = request.nextUrl.searchParams.get("nimbus_preview");
+  requestHeaders.set(
+    "x-nimbus-preview-mode",
+    nimbusPreviewMode === "true" ? "true" : "false",
```
flozia, line 37: Nit: This could be simplified to:
```suggestion
    (nimbusPreviewMode === "true").toString(),
```

flozia: Thinking about it again: The current ternary operator might be easier to read.

flozia (approved):
Approving with the caveat that we are waiting on directions from a data steward. The code looks good to me and when giving this a spin locally it works as expected.

src/middleware.ts
rhelmer: I took a look at the existing code and I see some examples of doing this same thing as ternary but not `toString()`. I like the latter method since it's a little shorter (and arguably clearer, I think it'd be easy to gloss over the fact that the ternary is *also* returning a string) but I'd rather be consistent for now and switch everything else over too if we're going to go that way. Good suggestion :)

rhelmer: Oh also I don't like that the ternary could easily typo e.g. `treu` or `fales` which to `toString()` approach doesn't have.

PR #4949, opened 2024-08-12
add docs and a cli flag for testing adm-zip dependency updates

rhelmer:
# References:

Jira: MNTOR-3467

# Description

A dependency update for `adm-zip` was opened and I noticed we didn't have docs for this yet. There are two `package.json` scripts related to the auto-complete feature that use this dependency:

1. `create-location-data` - intended to run from cron to download and process location files and upload to S3
2. `get-location-data` - downloads the latest output from (1) from the S3 bucket

Only (1) actually uses `adm-zip`, the S3 part wasn't optional so I added a `--skip-upload` flag.

# How to test

Try following the instructions per the docs :) Note that it takes quite a while to run, at least on my machine.

# Checklist (Definition of Done)

- [x] Commits in this PR are minimal and [have descriptive commit messages](https://chris.beams.io/posts/git-commit/).

PR #5472, opened 2025-01-09
Fix uploadToS3 import path

Vinnl:
This potentially fixes MNTOR-3922 as well, since the icon upload imported from the `.js` file: https://github.com/mozilla/blurts-server/blob/d91f5a5242282c7e65ee9d4b3f5180a4f6f68fc4/src/scripts/cronjobs/syncBreaches.ts#L22

I reverted `s3.ts` back to a JS file because it was still being imported from the JS upload script. Or at least, it was supposed to, as that had the wrong path.

### DIFF
--- a/src/scripts/build/uploadAutoCompleteLocations.js
+++ b/src/scripts/build/uploadAutoCompleteLocations.js
@@ -27,6 +27,7 @@ import os from "os";
 import path from "path";
 import fs from "fs";
 import AdmZip from "adm-zip";
+import { uploadToS3 } from "../../utils/s3.js";
 
 const REMOTE_DATA_URL = "https://download.geonames.org/export/dump";
 const DATA_COUNTRY_CODE = "US";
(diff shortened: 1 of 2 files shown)


src/scripts/build/uploadAutoCompleteLocations.js
```diff
@@ -27,6 +27,7 @@ import os from "os";
 import path from "path";
 import fs from "fs";
 import AdmZip from "adm-zip";
+import { uploadToS3 } from "../../utils/s3.js";
```
Vinnl, line 30: Note the changed path - but also, @rhelmer, you explicitly made this into a dynamic import here:

https://github.com/mozilla/blurts-server/pull/4949/files#diff-86136336e578d7cf634398ffcbdfaf05b5b938e85866325dad589fb4ff98b191R331

Was that just to avoid importing it if not used? Because I think that potentially obscures the path being invalid, so I changed it to a regular import - but let me know if there's another reason.

rhelmer: IIRC it was to be able to run this locally more easily when setting up a new dev env, I'm fine changing it as long as it works. It was added in https://github.com/mozilla/blurts-server/pull/4949

rhelmer (approved):

PR #5479, opened 2025-01-09
MNTOR-3919: db table changes for churn email column (Part 1)

mansaj:
# References:

Jira: MNTOR-3919

# Description
Table column added for churn email boolean 

### DIFF
--- a/src/db/migrations/20250109140803_add_churn_prevention_email_sent.js
+++ b/src/db/migrations/20250109140803_add_churn_prevention_email_sent.js
@@ -8,8 +8,8 @@
  */
 export function up (knex) {
   return knex.schema.table("subscribers", table => {
-    table.boolean("churn_prevention_email_sent").defaultTo(false);
-    table.index("churn_prevention_email_sent");
+    table.timestamp("churn_prevention_email_sent_at");
+    table.index("churn_prevention_email_sent_at");
   });
 }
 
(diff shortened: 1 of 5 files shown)


rhelmer (approved):

src/db/migrations/20250109140803_add_churn_prevention_email_sent.js
```diff
@@ -0,0 +1,25 @@
...
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+/**
+ * @param { import("knex").Knex } knex
+ * @returns { Promise<void> }
+ */
+export function up (knex) {
+  return knex.schema.table("subscribers", table => {
+    table.boolean("churn_prevention_email_sent").defaultTo(false);
```
Vinnl, line 11: Should it maybe be a date instead? Since users may potentially churn again next year? Or are we assuming we'll have a better solution by then? :sweat_smile:

mansaj: Hmm interesting.. If there's another strategy, I don't think it's going to be sending the same email. For consistency sake, I think it's probably best to use boolean but that's an interesting point.

mansaj: Actually, we do have precedent for using ts, so I'll change it to using ts and check for null instead of `false`

PR #5480, opened 2025-01-09
MNTOR-3919: getting setter for churn email state (Part 2)

mansaj:
# References:

Jira: MNTOR-3919
Figma:

# Description
These functions will be used by the cronjob to get/set churn email state for a user. 

### DIFF
--- a/src/db/tables/subscribers.ts
+++ b/src/db/tables/subscribers.ts
@@ -559,7 +559,9 @@ async function markChurnPreventionEmailAsJustSent(
 ) {
   const affectedSubscribers = await knex("subscribers")
     .update({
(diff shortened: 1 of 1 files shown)


rhelmer (approved):
lgtm although I think @Vinnl asked elsewhere: are we likely to want to do this again? If so, a datestamp might be better than a boolean. If we want to wait and refactor at that time I think it's OK too but worth considering.

mansaj:
> lgtm although I think @Vinnl asked elsewhere: are we likely to want to do this again? If so, a datestamp might be better than a boolean. If we want to wait and refactor at that time I think it's OK too but worth considering.

decided to go with the ts suggestion

PR #5485, opened 2025-01-13
Make lack of plaintext emails explicit

Vinnl:
SubPlat apparently sends plaintext emails:
https://storage.googleapis.com/mozilla-storybooks-fxa/commits/d06a74b249ebfe979642fdc94f4086a3ed031312/fxa-auth-server/index.html?path=/story/subplat-emails-layout--layout-with-was-deleted

We currently don't, although technically, we easily could: we'd just have to add a `text` property [to the options object passed to `gTransporter.sendMail`](https://github.com/mozilla/blurts-server/blob/a99b0506c72bcb02570839d9426513f040b01cb2/src/utils/email.ts#L65). Since there's currently discussion about whether to send them ourselves, I figured I'd at least adopt SubPlat's style for Storybook, to make it top-of-mind that we could send them, and to make it apparent where we're not and, in case we do start sending them, to make it ease to compare and see where we include information in the HTML email that we don't in the plaintext version.

### DIFF
--- a/src/emails/StorybookEmailRenderer.tsx
+++ b/src/emails/StorybookEmailRenderer.tsx
@@ -9,6 +9,7 @@ import { renderToStaticMarkup } from "react-dom/server";
 
 export type Props = {
   children: ReactNode;
+  plainTextVersion: string | null;
   emulateDarkMode?: boolean;
 };
 
@@ -18,7 +19,7 @@ export const StorybookEmailRenderer = (props: Props) => {
   const rawMjml = renderToStaticMarkup(props.children);
   const renderResult = mjml2html(rawMjml);
   return (
-    <>
+    <div className="wrapper">
       <style>{`
       @media (prefers-color-scheme: dark) {
         * {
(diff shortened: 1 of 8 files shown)


src/emails/StorybookEmailRenderer.tsx
```diff
@@ -31,22 +32,71 @@ export const StorybookEmailRenderer = (props: Props) => {
...
+                  ${renderResult.html}
+                  `
+                : renderResult.html,
+          }}
+          className={props.emulateDarkMode ? "dark-mode-enforced" : ""}
+        />
+      </section>
+      <section className="plaintext">
+        <div className="badge">Plaintext</div>
+        <pre>{props.plainTextVersion ?? "Not set"}</pre>
```
flozia, line 97: Is it intended that we are showing the string and not the rendered HTML here?

```suggestion
        {props.plainTextVersion ? (
          <div
            dangerouslySetInnerHTML={{
              __html: props.plainTextVersion,
            }}
            className={props.emulateDarkMode ? "dark-mode-enforced" : ""}
          />
        ) : (
          "Not set"
        )}
```

Vinnl: @flozia I'm not sure I follow - the plain text version is an alternative to the HTML version, i.e. there shouldn't be HTML in there?

flozia: @Vinnl In that case I misunderstood. I thought the idea was that is was plain text when looking at the [FXA email story](https://storage.googleapis.com/mozilla-storybooks-fxa/commits/d06a74b249ebfe979642fdc94f4086a3ed031312/fxa-auth-server/index.html?path=/story/subplat-emails-layout--layout-no-product-with-brand-messaging): There are still links in there — how would those be handled?

Vinnl: Ha actually, I think that's a mistake on their side? Generally, it'll be up to the mail client to make spelled-out URLs clickable, or to the user to copy-paste it.

flozia: Makes sense!

PR #5509, opened 2025-01-16
Cirrus has enrolments as an array

Vinnl:
This PR I'm not sure about, and it's probably wrong, but: in MNTOR-3812, @rhelmer showed an example with `Enrollments` being an array, whereas our code says that it's a single object.

I'm not sure if it actually is, and if it is, I'm not sure why and how we're supposed to handle it - in this PR, I'm just always getting the first element, but that's probably not correct. However, I figured I'd submit this PR to at least have this observation in play somewhere.

### DIFF
--- a/src/app/hooks/useGlean.ts
+++ b/src/app/hooks/useGlean.ts
@@ -50,17 +50,17 @@ export const useGlean = () => {
         typeof experimentData["Enrollments"] !== "undefined"
       ) {
         (data as GleanMetricMap["button"]["click"]).nimbus_user_id =
-          experimentData["Enrollments"]["nimbus_user_id"];
+          experimentData["Enrollments"][0]?.nimbus_user_id;
         (data as GleanMetricMap["button"]["click"]).nimbus_app_id =
-          experimentData["Enrollments"]["app_id"];
+          experimentData["Enrollments"][0]?.app_id;
         (data as GleanMetricMap["button"]["click"]).nimbus_experiment =
-          experimentData["Enrollments"]["experiment"];
+          experimentData["Enrollments"][0]?.experiment;
         (data as GleanMetricMap["button"]["click"]).nimbus_branch =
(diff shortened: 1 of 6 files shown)


rhelmer:
I'd like to land the other PRs first and see what's going on with stage, this was definitely working (with the `CirrusV2` flag enabled) before we deployed to prod. It may have broken in some subtle way, but before the holiday break I confirmed that Glean was sending the data it was getting back and experiments seemed to work. It definitely didn't fail the way prod did yesterday.

So I'd like to understand the current state after things are rolled back before we land, I don't doubt that this is a problem though I just want to understand the sequence of what happened.

rhelmer:
Hm maybe there is a problem with the documentation... I did test this and it worked as expected (with V2), I'll re-test before we land this. It might be a bug in Cirrus (or its docs).

rhelmer:
Stage is working now with everything else landed, I flipped the `CirrusV2` flag on and we're getting ```useGlean.ts:63 Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'toString')
    at useGlean.ts:63:55```

So I think it's OK to land this one, I think we might need to dig into this one some more though.

Vinnl:
@rhelmer That does indicate that we need to fix this, but I'm still not sure that this PR is the right fix. (Or actually, I'm fairly sure it's not.) I still have [this open question](https://mozilla.slack.com/archives/C089H39FN9W/p1737102105758719?thread_ts=1737031044.890559&cid=C089H39FN9W) for Yashika about how to pass the enrollments to Glean, given that Glean only accepts single enrollments.

rhelmer:
Waiting for an update to Glean to support arrays in `extra_keys`

Vinnl:
@rhelmer Alright, updated this PR to no longer send the enrollment data to Glean.

rhelmer (approved):

PR #5511, opened 2025-01-17
Use optional chaining to avoid undefined monthly_monitor_report

rhelmer:
# References:

Jira: MNTOR-3848

# Description

Use [optional chaining](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining) to avoid `monthly_monitor_report` being undefined.

I believe that the problem here is that `monthly_monitor_report` can be `null` in the database if the monthly report has not run successfully for this account. I also believe we started getting more of these reports because the cron job that sends this email broke temporarily (it's now fixed), but there's still a possibility for this to happen for e.g. new users.

I did not add a unit test, but I did change the typescript types to reflect that this may be undefined.

# How to test

The easiest way to to test locally is to ensure that `monthly_monitor_report` is `null` in the `subscribers` table.


Vinnl:
The question is: if the value can also be `undefined`, then why didn't TS warn us when we passed that into these components? So I traced that down to a couple of overeager type assertions, which I removed in https://github.com/mozilla/blurts-server/pull/5521.

I'm also not 100% sure what the values of the checkboxes initialised with this value should be if it's undefined, but I would guess that they should actually be checked by default, rather than unchecked as they are now. My PR does that too.

Vinnl (approved):

PR #5512, opened 2025-01-17
MNTOR-3918: churn table, admin dashboard

mansaj:
# References:

Jira: MNTOR-3918
Figma:

# Description
* New table for the churn management
* New admin dashboard to upload and display churn table
* Upsert (create and replace when necessary) new entries into the churn table, just by uploading a new CSV


src/app/(proper_react)/(redesign)/(authenticated)/admin/churn-subscribers/page.tsx
```diff
@@ -0,0 +1,22 @@
...
+import { isAdmin } from "../../../../../api/utils/auth";
+import { ChurnAdmin } from "./ChurnAdmin";
+
+export default async function DevPage() {
+  const session = await getServerSession();
+
+  if (
+    !session?.user?.email ||
+    !isAdmin(session.user.email) ||
+    process.env.APP_ENV !== "local"
```
Vinnl, line 16: We'll want to use this on prod, right?

```suggestion
    !isAdmin(session.user.email)
```

src/db/tables/subscriber_churns.ts
```diff
@@ -32,4 +32,18 @@ async function upsertSubscriberChurns(
   }
 }
 
-export { upsertSubscriberChurns };
+async function getAllSubscriberChurns(): Promise<SubscriberChurnRow[]> {
+  try {
+    const res = await knex("subscriber_churns").select().returning("*");
```
Vinnl, line 37: AFAIK [`returning` is for inserts, updates and deletes](https://knexjs.org/guide/query-builder.html#returning):

```suggestion
    const res = await knex("subscriber_churns").select("*");
```

I think this might work as well somehow and we're already doing it elsewhere, but I just submitted [a PR](https://github.com/mozilla/blurts-server/pull/5521) that removes some instances that have both `select` arguments _and_ `returning`, so it's probably best to only use one?

src/db/tables/subscriber_churns.ts
```diff
@@ -32,4 +32,18 @@ async function upsertSubscriberChurns(
...
+
+    logger.info("get_all_subscriber_churns_success", { count: res.length });
+    return res as SubscriberChurnRow[];
```
Vinnl, line 40: Also, this shouldn't need a type assertion, which can hide issues.

```suggestion
    return res;
```

(I think the `returning` is breaking type inference.)

src/app/(proper_react)/(redesign)/(authenticated)/admin/churn-subscribers/actions.tsx
```diff
@@ -0,0 +1,20 @@
...
+import { SubscriberChurnRow } from "knex/types/tables";
+
+export async function getAllChurns() {
+  return getAllSubscriberChurns();
+}
+
+export async function upsertAllChurns(
+  churningSubscribers: SubscriberChurnRow[],
+) {
+  return upsertSubscriberChurns(churningSubscribers);
```
Vinnl, line 19: I think it's part of a different PR's scope, but could we add an `updated_at` column? Or maybe even not do updates, but just insert new ones and select by the latest `created_at` date?

Just thinking of ways to recover from the inevitable faulty CSV upload, where it would be great if we could just remove the rows inserted by the faulty upload to get back to a working state.

src/app/(proper_react)/(redesign)/(authenticated)/admin/churn-subscribers/ChurnAdmin.tsx
```diff
@@ -0,0 +1,121 @@
...
+    if (!file) return;
+
+    const reader = new FileReader();
+    reader.onload = async (e) => {
+      const text = e.target?.result as string;
+      const rows = text.split("\n");
+
+      console.log({ rows: rows.length });
+      const parsedData: SubscriberChurnRow[] = rows
+        .slice(1)
```
Vinnl, line 40: Maybe do a check here that the headers and number of columns match what we expect, and bail out if not? In case someone selects the wrong file.

src/app/(proper_react)/(redesign)/(authenticated)/admin/churn-subscribers/ChurnAdmin.tsx
```diff
@@ -0,0 +1,121 @@
...
+  >(null);
+
+  // Load data on component mount
+  useEffect(() => {
+    const loadSubscriberData = async () => {
+      const data = await getAllChurns();
+      setSubscriberData(data);
+    };
+
+    void loadSubscriberData();
```
Vinnl, line 26: Hmm, isn't it better to avoid loading the existing subscriber data into the page? It's not super-sensitive I don't think, but just to minimise who has access to what data.

In any case, there's no need to load it in an effect, if you do want to keep it. You can load it on the server side in `page.tsx` and then pass it to this component as a prop.

src/app/(proper_react)/(redesign)/(authenticated)/admin/churn-subscribers/ChurnAdmin.tsx
```diff
@@ -0,0 +1,121 @@
...
+            intervl: values[4],
+            plan_id: values[6],
+            product_id: values[7],
+            current_period_end: new Date(values[8]).toISOString(),
+          };
+        });
+
+      console.log({ parsedData });
+      // Update backend and refresh data
+      await upsertAllChurns(parsedData);
```
Vinnl, line 58: I've got a slight worry that starting the upload the moment the user selects a file is also prone to attract errors from simple misclicks. To be safe, we could wrap it in a form, add a submit button, and only upload it once the form is submitted?

Vinnl (approved):
I didn't try it myself yet, but the overall approach looks good. I do think we could/probably should do a bit more to avoid corrupting the DB, which I've added some comments about, and two easy fixes about type safety.

src/app/(proper_react)/(redesign)/(authenticated)/admin/churn-subscribers/ChurnAdmin.tsx
mansaj: I don't think it matters in this case since the data is uploaded by the same person. It's just a way to sanity check what's updated in the db and make sure nothing weird happened

Vinnl: Oh, I thought it also showed what was already in the database before the person uploaded anything? And this page is visible to _any_ admin, right?

src/app/(proper_react)/(redesign)/(authenticated)/admin/churn-subscribers/actions.tsx
mansaj: we could add a clear all function.

src/app/(proper_react)/(redesign)/(authenticated)/admin/churn-subscribers/ChurnAdmin.tsx
mansaj: added

src/app/(proper_react)/(redesign)/(authenticated)/admin/churn-subscribers/actions.tsx
mansaj: added a red button

src/app/(proper_react)/(redesign)/(authenticated)/admin/churn-subscribers/ChurnAdmin.tsx
mansaj: correct

src/app/(proper_react)/(redesign)/(authenticated)/admin/churn-subscribers/ChurnAdmin.tsx
mansaj: added this

PR #5525, opened 2025-01-22
 Add stories for header and footer components 

Vinnl:
Laura and Wolasi asked for these. (I also moved the header and footer components into the `components/` folder, since they're... You know, components.


src/emails/StorybookEmailRenderer.tsx
```diff
@@ -89,7 +89,7 @@ export const StorybookEmailRenderer = (props: Props) => {
                   `
                 : renderResult.html,
           }}
-          className={props.emulateDarkMode ? "dark-mode-enforced" : ""}
+          className={`body ${props.emulateDarkMode ? "dark-mode-enforced" : ""}`}
```
Vinnl, line 92: I didn't end up needing this class. Could remove it too, I guess :shrug:

PR #5527, opened 2025-01-23
Upgrade to Node 20.18

Vinnl:
This makes sure we have the latest security patches.

I triggered the Dependabot for Docker, but it didn't find any new versions. It looks like it stopped doing new PRs since https://github.com/mozilla/blurts-server/pull/5241#issuecomment-2441482193 - so maybe we can't tell it to stay on v20? This doesn't seem to have helped either: https://github.com/mozilla/blurts-server/blob/2213d78b1d2fee1af5e587aaf56ea19aecd9c47d/.github/dependabot.yml#L78


Vinnl:
Aargh, again newly uncovered lines :facepalm:

rhelmer (approved):
Node changes lgtm. Would be really nice to get to the bottom of why Node version updates change our coverage results though

mansaj (approved):

Vinnl:
The coverage results are inconsistent between different runs now, so I'm giving up on this for now :( Really annoying. Might be worth switching to using Babel for coverage, or even away from Jest altogether, although both options involve lots of rework without tangible results :(

PR #5549, opened 2025-01-28
Fix experiment preview for public layout

flozia:
# References:

Jira: [MNTOR-3981](https://mozilla-hub.atlassian.net/browse/MNTOR-3981)

# Description

Fixes the option to force enable experiments on public pages.

# How to test

- Visit the public landing page
- Add the UTM parameter `/?nimbus_preview=true`

[MNTOR-3981]: https://mozilla-hub.atlassian.net/browse/MNTOR-3981?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/app/(proper_react)/(redesign)/(public)/layout.tsx
```diff
@@ -7,29 +7,50 @@ import { PublicShell } from "./PublicShell";
...
-    countryCode,
-    locale: getLocale(l10n),
-    previewMode: false,
+    experimentationId: experimentationId,
+    countryCode: countryCode,
+    locale: currentLocale,
+    previewMode: nimbusPreviewMode === "true",
   });
 
+  const nimbus_user_id = experimentData["Enrollments"]?.nimbus_user_id;
```
Vinnl, line 44: The code you copied here unfortunately was incorrect; `.Enrollments` is an array, not an object. I have a fix here: https://github.com/mozilla/blurts-server/pull/5509/commits/7d5da75ef7cdfb3b4f6167403393abfba736ba4c

src/app/(proper_react)/(redesign)/(public)/layout.tsx
```diff
@@ -7,29 +7,50 @@ import { PublicShell } from "./PublicShell";
...
-  const headersList = await headers();
-  const countryCode = getCountryCode(headersList);
   const enabledFeatureFlags = await getEnabledFeatureFlags({
     isSignedOut: true,
   });
   const l10n = getL10n(await getAcceptLangHeaderInServerComponents());
-  const experimentationId = await getExperimentationId(null);
+  const currentLocale = getLocale(
+    getL10nBundles(await getAcceptLangHeaderInServerComponents()),
+  );
```
Vinnl, line 28: Can't we just pass in the `l10n` object we already have here?

```suggestion
  const currentLocale = getLocale(l10n);
```

src/app/(proper_react)/(redesign)/(public)/layout.tsx
```diff
@@ -7,29 +7,50 @@ import { PublicShell } from "./PublicShell";
...
+  const experimentationId = await getExperimentationId(session?.user ?? null);
   const experimentData = await getExperiments({
-    experimentationId,
-    countryCode,
-    locale: getLocale(l10n),
-    previewMode: false,
+    experimentationId: experimentationId,
+    countryCode: countryCode,
+    locale: currentLocale,
+    previewMode: nimbusPreviewMode === "true",
```
Vinnl, line 41: Hmm, I might be missing some context here, but it's not clear to me why callers would need to determine the preview mode in different ways, but I see some instances reading the header added by the middleware, and some directly reading request parameters. Can't `getExperiments` just do that work for us directly? Or are we also using it in emails?

Vinnl (approved):

src/app/(proper_react)/(redesign)/(public)/layout.tsx
flozia: In this case, it is because we can’t read request parameters directly from layouts.

src/app/(proper_react)/(redesign)/(public)/layout.tsx
flozia: Yes! Updated in [c30e1f7](https://github.com/mozilla/blurts-server/pull/5549/commits/c30e1f7c4b3752b8f6a256f7fb21aeefa6b24b0c).

src/app/(proper_react)/(redesign)/(public)/layout.tsx
Vinnl: Right, I suppose that's how we got here. But now that we have that middleware adding the header, I suppose we can just use that everywhere, and incorporate it directly into the function? Would've prevented this bug :)

flozia: That is true. I’ve created [MNTOR-4011](https://mozilla-hub.atlassian.net/browse/MNTOR-4011) to unify setting the preview mode and ensure we don’t interfere with upcoming experiments.

PR #5558, opened 2025-01-29
Use Ruby version syntax for Dependabot-Docker

Vinnl:
Looks like https://github.com/mozilla/blurts-server/pull/5546 wasn't enough :unamused: 

According to
https://docs.github.com/en/code-security/dependabot/working-with-dependabot/dependabot-options-reference#versions-ignore this is needed. Unfortunately I can find very little information on "Ruby version syntax", but I think this comment states that removing the `.x` might do it:
https://github.com/dependabot/dependabot-core/issues/2644#issuecomment-709212848


rhelmer (approved):
Thanks... yes this one is a bit hard to test unfortunately, this looks like it's worth a try. Let's continue staying on top of Node releases and double-checking that dependabot is doing the right thing, I don't think there's a non-trivial way to set up a test environment for this otherwise.

PR #5573, opened 2025-01-30
MNTOR-3897: add admin check for actions

mansaj:
# References:

Jira: MNTOR-3897

# Description
Add admin checks for churn dash actions 


src/app/(proper_react)/(redesign)/(authenticated)/admin/churn-subscribers/actions.tsx
```diff
@@ -9,17 +9,34 @@ import {
   deleteSubscriberChurns,
 } from "../../../../../../db/tables/subscriber_churns";
 import { SubscriberChurnRow } from "knex/types/tables";
+import { getServerSession } from "../../../../../functions/server/getServerSession";
+import { isAdmin } from "../../../../../api/utils/auth";
 
 export async function getAllChurns() {
+  const session = await getServerSession();
+  if (!session?.user?.email || !isAdmin(session.user.email)) {
+    return [];
```
rhelmer, line 18: From the docs, I believe that returning `null` is the preferred way to do this currently:
https://nextjs.org/docs/app/building-your-application/authentication#server-actions

I found there is an experimental `forbidden` function that returns an `HTTP 403` which is an even better fit:
https://nextjs.org/docs/app/api-reference/functions/forbidden

I think we should return `null` for now, and maybe file a ticket in the tech debt backlog to switch to using this `forbidden` handler when it becomes stable. wdyt?

src/app/(proper_react)/(redesign)/(authenticated)/admin/churn-subscribers/actions.tsx
```diff
@@ -9,17 +9,34 @@ import {
...
+  }
+
   return getAllSubscriberChurns();
 }
 
 export async function upsertAllChurns(
   churningSubscribers: SubscriberChurnRow[],
 ) {
+  const session = await getServerSession();
+  if (!session?.user?.email || !isAdmin(session.user.email)) {
```
rhelmer, line 28: Is there an advantage to `(!session?.user?.email || !isAdmin(session.user.email))` vs. just `(!isAdmin(session?.user?.email)`?

rhelmer (changes requested):
I mentioned a few things, the only important one is that from the NextJS docs, I believe that returning `null` is the right thing to do for now (but note there's an experimental `forbidden` handler which will be available eventually)

src/app/(proper_react)/(redesign)/(authenticated)/admin/churn-subscribers/actions.tsx
mansaj: moved them to a separate func

PR #5575, opened 2025-02-03
Unify setting the preview mode for getExperiments

flozia:
# References:

Jira: [MNTOR-4011](https://mozilla-hub.atlassian.net/browse/MNTOR-4011)

# Description

At the moment, we have two ways of forcing experiments via `nimbus_preview`:

1. Via the search param `nimbus_preview` in pages
2. Via `x-nimbus-preview-mode` in layouts (which is getting set by the search param `nimbus_preview` in the middleware)

This PR unifies the the Nimbus preview to check `x-nimbus-preview-mode` only and moves the condition into `getExperiments`.

# How to test

Append `/?nimbus_preview` to a URL that has disabled experiments.

[MNTOR-4011]: https://mozilla-hub.atlassian.net/browse/MNTOR-4011?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/app/functions/server/getExperiments.ts
```diff
@@ -53,14 +52,20 @@ export async function getExperiments(params: {
     serverUrl.pathname = "/v1/features";
   }
 
+  const headersList = await headers();
+  // Check for Nimbus preview mode. Note that this requires a full page reload
+  // to activate: https://nextjs.org/docs/app/api-reference/file-conventions/layout#caveats
```
Vinnl, line 57: This comment doesn't quite seem to apply here, though it could possibly reworded to talk about if `getExperiments` is called in a layout.

(That said, I'm not sure what the failure mode here is - surely the query parameter only gets added by manually adjusting the URL, in which case pressing <kbd>Enter</kbd> seems implied? We're not supposed to add `<Link>`s with these query params, right?)

Vinnl (approved):
Looks much more reliable this! If @rhelmer has time, I think it'd be good if he'd take a look as well, but I'm fairly confident that this is safe and right.

rhelmer (approved):

src/app/functions/server/getExperiments.ts
flozia: That is right — I’ve [361b196](https://github.com/mozilla/blurts-server/pull/5575/commits/361b196b7c6d85b1c2a6ed21d9bbe6122123ee4f). No, we are not supposed to add links with query params. This is only meant to force an experiment by visiting the URL directly.

Vinnl: Cool, thanks!

PR #5583, opened 2025-02-03
Fix running E2E tests locally

flozia:
# Description

If `sharp` is not available, running the E2E tests fails with an error. This PR includes `sharp` optionally as an NPM package.

Also: Now after testing the testing of the landing experiment on `stage` is done and we are all set up to run experiments via experimenter again we can have the experiment not enabled by default — this is currently causing the E2E test suite to fail.

# How to test

Run the E2E tests locally with `npm run e2e:smoke`.


Vinnl:
Might want to assign https://mozilla-hub.atlassian.net/browse/MNTOR-3916 to yourself :)

Vinnl (approved):

PR #5587, opened 2025-02-04
MNTOR-4028: Allow renewing after receiving an expiration email

Vinnl:
# References:

Jira: MNTOR-4028
Figma:

# Description

We email people that their subscription is about to expire, then mark them as having received the email. We check whether we have marked someone as having received the email when collecting the list of subscribers to email, which means that we shouldn't use that list when determining whether someone can renew.

I've updated it to use a similar check, but to skip the check for `churn_prevention_email_sent_at`.

# How to test

Create a `.csv` file with the following contents:

```csv
userid,customer,created,nickname,intervl,intervl_count,plan_id,product_id,current_period_end
18b8ea3c9b12427893ded71fda0cb8aa,cus_custom,2/6/2024 10:57:57,Yearly Production Pricing ($8.99 per month: $107.88),year,1,price_1Nv4ODJNcmPzuWtRoYpoFHXd,prod_OiV9RSaatywSRy,2/2/2025 13:37:00
```

Replace `18b8ea3c9b12427893ded71fda0cb8aa` with the `fxa_uid` of the subscriber you want to email. (This subscriber should have Plus; you can give them Plus via `/admin/dev`.) Then upload that file via `/admin/churn-subscribers/`. It should say that there's one customer to email in the next cron run.

Then, run that cron job through `npm run dev:cron:churn-discount`. The customer should receive an email.

If you click the CTA in that email with `main` checked out, you should see "Your ⁨Monitor Plus⁩ subscription is still active", whereas on this branch, you should be able to renew successfully.

- [ ] I've added a unit test to test for potential regressions of this bug. - No, because it involves a DB query, which we don't have a good way to test for.

### DIFF
--- a/src/db/tables/subscriber_churns.ts
+++ b/src/db/tables/subscriber_churns.ts
@@ -59,6 +59,29 @@ async function getAllSubscriberChurns(): Promise<SubscriberChurnRow[]> {
   }
 }
 
+async function getUpcomingChurns(): Promise<Array<SubscriberChurnRow>> {
+  try {
+    const res = await knex("subscriber_churns")
+      .select("*")
+      .where("intervl", "year")
+      .whereNotNull("current_period_end")
+      .where(knex.raw("current_period_end::timestamptz"), ">=", knex.fn.now())
+      .where(
+        knex.raw("current_period_end::timestamptz"),
+        "<=",
+        knex.raw("CURRENT_TIMESTAMP + interval '7 days'"),
+      );
+
+    logger.info("get_upcoming_churns_success", { count: res.length });
+    return res;
+  } catch (e) {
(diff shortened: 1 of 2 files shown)


src/db/tables/subscriber_churns.ts
```diff
@@ -59,6 +59,29 @@ async function getAllSubscriberChurns(): Promise<SubscriberChurnRow[]> {
   }
 }
 
+async function getUpcomingChurns(): Promise<Array<SubscriberChurnRow>> {
```
Vinnl, line 62: This is the same as the `getChurnsToEmail` query, with the join and `churn_prevention_email_sent_at` check removed.

flozia (approved):

PR #5591, opened 2025-02-05
MNTOR-3971: Use last optout date instead of last update

Vinnl:
# References:

Jira: MNTOR-3971
Figma: N/A

# Description

This newly-added API response property accurately reflects the last scan date (which we had incorrectly hoped the last update date to do).

Note that I'm not familiar with the QA customs tool, so I mostly followed existing patterns and added `last_optout_at` wherever I could find `optout_attempts`. I think that covers it, but it uses quite a few custom types and otherwise lives fairly separate from the rest of the codebase, so I'm not particularly confident.

# Screenshot (if applicable)

Not applicable.

# How to test

Make sure the `DataBrokerRemovalAttempts` and `AdditionalRemovalStatuses` flags are on, which in Storybook you can do by [setting it to](https://deploy-preview-5591--fx-monitor-storybook.netlify.app/?path=/story/pages-logged-in-dashboard-us-user-plus--dashboard-us-premium-unresolved-scan-unresolved-breaches&args=enabledFeatureFlags[0]:DataBrokerRemovalAttempts;enabledFeatureFlags[1]:AdditionalRemovalStatuses):

```json
["DataBrokerRemovalAttempts", "AdditionalRemovalStatuses"]
```

(You might have to refresh, because the initial value Storybook sets is `{}`, which is not an array and thus causes an error.)

A Plus user (e.g. in [the above story](https://deploy-preview-5591--fx-monitor-storybook.netlify.app/?path=/story/pages-logged-in-dashboard-us-user-plus--dashboard-us-premium-unresolved-scan-unresolved-breaches&args=enabledFeatureFlags[0]:DataBrokerRemovalAttempts;enabledFeatureFlags[1]:AdditionalRemovalStatuses)) should then see the removal attempt in the Fixed tab for any "Requested removal" scan results. If you want to be sure that it's showing the last optout date as returned from the API, you can set it to some recognisable value in `/src/apiMocks/mockData.ts`.

# Checklist (Definition of Done)

- [x] Localization strings (if needed) have been added.
- [x] Commits in this PR are minimal and [have descriptive commit messages](https://chris.beams.io/posts/git-commit/).
- [x] I've added or updated the relevant sections in readme and/or code comments
- [x] I've added a unit test to test for potential regressions of this bug.
- [x] If this PR implements a feature flag or experimentation, I've checked that it still works with the flag both on, and with the flag off.
- [x] If this PR implements a feature flag or experimentation, the Ship Behind Feature Flag status in Jira has been set
[...]

PR #5586, opened 2025-02-03
Add CSAT banner for 716 days after having a Plus subscription

flozia:
# References:

Jira: [MNTOR-3991](https://mozilla-hub.atlassian.net/browse/MNTOR-3991)

# Description

Add a CSAT survey for users after ~24 months of being a Plus subscriber.

# How to test

**Storybook**

1. CsatSurvey (AutomaticRemoval): `/?path=/story/misc-csatsurvey--csat-survey-automatic-removal`
2. Set `elapsedTimeInDaysSinceInitialScan >= 716`
3. Click on a response
4. The follow-up link `href` should match the pattern: ~~`https://survey.alchemer.com/s3/7714663/*`~~ `https://survey.alchemer.com/s3/8176616/*`

# Checklist (Definition of Done)

- [x] Commits in this PR are minimal and [have descriptive commit messages](https://chris.beams.io/posts/git-commit/).
- [x] I've added a unit test to test for potential regressions of this bug.
- [x] Product Owner accepted the User Story (demo of functionality completed) or waived the privilege.
- [x] All acceptance criteria are met.
- [x] Jira ticket has been updated (if needed) with suggestions for QA when this PR is deployed to stage.

[MNTOR-3991]: https://mozilla-hub.atlassian.net/browse/MNTOR-3991?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #5596, opened 2025-02-06
Remove some feature flags

Vinnl:
Removed the flags `ConfirmCancellation`, `FirstDataBrokerRemovalFixedEmail` and `AdditionalRemovalStatuses`, which were relatively easy to remove and have been on in production for a while now.

Update: also removed `AutomaticRemovalCsatSurvey`, `BreachEmailRedesign`, `MonthlyReportFreeUser` and `UpdatedEmailPreferencesOption`.


src/db/tables/subscribers.ts
```diff
@@ -350,17 +345,8 @@ async function getPotentialSubscribersWaitingForFirstDataBrokerRemovalFixedEmail
     // ...who haven't received the email...
     .andWhere("first_broker_removal_email_sent", false)
     // ...and signed up after the feature flag `FirstDataBrokerRemovalFixedEmail`
-    // has been enabled last.
-    .andWhere("created_at", ">=", flag.modified_at);
```
Vinnl, line 354: @flozia From [this comment](https://mozilla-hub.atlassian.net/browse/MNTOR-2700?focusedCommentId=907914), I think it was around this time that this flag was enabled, and if I'm understanding the query correctly, as long as this date isn't _earlier_ than that the flag was enabled, we should be good, right? I could also move it some more into the future, if we want to be safe.

flodolo (approved):

src/db/tables/subscribers.ts
flozia: @Vinnl Yes that’s right.

src/db/tables/featureFlags.ts
```diff
@@ -38,17 +38,10 @@ export async function getDeletedFeatureFlags() {
 }
 
 export const featureFlagNames = [
-  "UpdatedEmailPreferencesOption",
   "CancellationFlow",
-  "ConfirmCancellation",
-  "FirstDataBrokerRemovalFixedEmail",
   "DiscountCouponNextThreeMonths",
   "LatestScanDateCsatSurvey",
-  "AutomaticRemovalCsatSurvey",
```
flozia, line 47: I would prefer if we could keep all the feature flags related to the CSAT surveys. We only recently landed a PR that extends the functionality of this CSAT PR #5586.

src/db/tables/featureFlags.ts
```diff
@@ -38,17 +38,10 @@ export async function getDeletedFeatureFlags() {
...
-  "AdditionalRemovalStatuses",
```
flozia, line 48: I believe this is also a feature flag we would need to keep. We’ve [recently disabled it](https://mozilla.slack.com/archives/C0460QALW69/p1736794513970349?thread_ts=1736544869.685789&cid=C0460QALW69) with an issue related to the last opt-out removal. Even though, this issue will be addressed by PR #5591 the flag `AdditionalRemovalStatuses`, it depends on the opt-out removal to be activated and to work correctly.

src/db/tables/featureFlags.ts
Vinnl: Just to make sure: when you say "keep all the feature flags", in the context of this PR you only mean undoing the removal of `AutomaticRemovalCsatSurvey`, right? I did that in 2cac1d137. Thanks for noticing.

src/db/tables/featureFlags.ts
Vinnl: Reverted this in 82df2e2e1. Could you create a ticket that specifies when this flag is good to remove?

src/db/tables/featureFlags.ts
flozia: Thanks! Yes, I specifically mean “keep all the feature flags related to the CSAT surveys”.

flozia (approved):
Thank you for cleaning up the feature flags.

PR #5597, opened 2025-02-06
*Actually* fix flaky Plus subscription tests

Vinnl:
# References:

Jira: MNTOR-
Figma:

# Description

They were still failing for me every now and then, in particular when running all tests and my laptop being under heavy load, even after e23ea49b8214efa0d6fcc5cdb6840d3e14e978b0 (https://github.com/mozilla/blurts-server/pull/5584).

I could reliably reproduce that by adding a one-second timeout to every Promise in the tests, and then fix it by wrapping `await`ing those Promises in `act`. Which makes sense; after the Promises resolve, React needs to be given the opportunity to update the state and then the rendered view.

# How to test

Run the unit tests - they should all succeed, consistently. If you want to reproduce, change a Promise (like the `applyCouponPromise` to have a timeout, like this:

```ts
  const applyCouponPromise = new Promise((resolve) => {
    setTimeout(() => {
      resolve({
        success: true,
      });
    }, 1000);
  });
```

On `main`, you should see the test to start to consistently fail, whereas if you do this on this branch, they should succeed.

- [X] If this PR implements a feature flag or experimentation, I've checked that it still works with the flag both on, and with the flag off.
- [ ] Product Owner accepted the User Story (demo of functionality completed) or waived the privilege. N/A


mansaj:
For assertions that rely on eventual changes in the DOM, I think another option is to wrap in waitFor (which might be a little cleaner) It will retry the assertion until it passes or times out:

```
 await waitFor(() =>
     expect(screen.getByText(/Couldn’t renew subscription./)).toBeInTheDocument()
   );
   ```

Vinnl:
In this case I actually think `act` is cleaner, because it's not waiting for an arbitrary amount of time and retrying things in intervals; it's actually pushing the rest of the test code back onto the queue, to resume running when React has run all rendering code. Thus, it runs as fast as possible, but no faster, and doesn't expire.

PR #5619, opened 2025-02-12
Run ESLint on root files

flozia:
# References:

Follow-up for PR #5599

# Description

Run ESLint also on the files in the root directory.

### DIFF
--- a/eslint.config.js
+++ b/eslint.config.js
@@ -39,7 +39,7 @@ const config = [
         project: "tsconfig.json",
       },
     },
-    ignores: ["coverage", "dist", "sentry.*.config.ts"],
+    ignores: ["coverage", "dist"],
     plugins: {
       jsdoc,
       "@typescript-eslint": tsEslint,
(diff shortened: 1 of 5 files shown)


eslint.config.js
```diff
@@ -171,13 +169,19 @@ const config = [
...
-  // Ignore the following `.js` files from linting for now.
+  // Ignore the following files for now.
   {
     files: [
-      "eslint.config.js",
       "src/app/global-error.js",
-      "src/db/**/*.js",
-      "src/scripts/**/*.js",
+      "src/db/migrations/*.js",
+      "src/scripts/build/*.js",
```
Vinnl, line 177: Would the build scripts need lots of ignore comments? Otherwise, it might be good to catch potential bugs in those as well?

Vinnl (approved):

eslint.config.js
flozia: Created PR #5622 which extends TSLint coverage to include these files.

PR #5622, opened 2025-02-12
Increase typescript-eslint coverage

flozia:
# Description

Increases TSLint coverage to include the following files:
- `src/app/global-error.js`
- `src/db/migrations/*.js`
- `src/scripts/build/*.js`
- `src/scripts/loadtest/*.js`

# How to test

Running `npm run lint` should be successful.


src/db/migrations/20221026215921_add-breaches-table.js
```diff
@@ -2,17 +2,31 @@
...
     table.increments("id").primary();
+    // @ts-ignore TODO: Determine if the following line should be changed to
+    // `defaultTo` or remain unchanged as initially deployed.
     table.string("name").default("").unique();
+    // @ts-ignore TODO: Determine if the following line should be changed to
+    // `defaultTo` or remain unchanged as initially deployed.
     table.string("title").default("");
+    // @ts-ignore TODO: Determine if the following line should be changed to
+    // `defaultTo` or remain unchanged as initially deployed.
     table.string("domain").default("");
```
rhelmer, line 20: @flozia was this renamed on the `knex` side? I only see `defaultTo` in their docs, I don't see any reference to `default`

src/db/migrations/20230703010356_add-onerep-profiles-table.js
```diff
@@ -2,16 +2,28 @@
...
+ * @returns { Promise<void> }
+ */
 export function up(knex) {
   return knex.schema.createTable("onerep_profiles", (table) => {
     table.increments("id").primary();
     table
       .integer("onerep_profile_id")
       .references("subscribers.onerep_profile_id")
       .nullable();
+    // @ts-ignore TODO: Determine if the following line should be changed
```
rhelmer, line 16: Curious why this doesn't work, `varchar` should be valid no?

src/db/migrations/20230907143204_add_onerep_scanresults_table.js
```diff
@@ -100,6 +104,8 @@ export async function down(knex) {
   });
   await knex.schema.dropTable("onerep_scan_results");
   await knex.schema.alterTable("onerep_scans", (table) => {
+    // @ts-ignore TODO: Determine if the following line should be changed to
+    // `defaultTo` or remain unchanged as initially deployed.
```
rhelmer, line 108: FYI this is `dropUnique` not `default`, which should be valid https://knexjs.org/guide/schema-builder.html#dropunique

src/db/migrations/20231102024624_add_unique_index_to_scan_results_id.js
```diff
@@ -24,6 +24,8 @@ export async function up(knex) {
  */
 export function down(knex) {
   return knex.schema.table("onerep_scan_results", (table) => {
+    // @ts-ignore TODO: Determine if the following line should be changed to
+    // `defaultTo` or remain unchanged as initially deployed.
```
rhelmer, line 28: Same as above re: `dropUnique`

src/db/migrations/20240423150332_add_suffix_and_middle_name_to_onerep_profile_table.js
```diff
@@ -2,9 +2,17 @@
...
 
+/**
+ * @param { import("knex").Knex } knex
+ * @returns { Promise<void> }
+ */
 export function up(knex) {
   return knex.schema.table("onerep_profiles", (table) => {
+    // @ts-ignore TODO: Determine if the following line should be changed
+    // or remain unchanged as initially deployed.
     table.varchar("name_suffix").after("onerep_profile_id");
```
rhelmer, line 13: Same q as above re: `varchar`

src/scripts/build/getAutoCompleteLocations.js
```diff
@@ -15,6 +15,10 @@ if (!existsSync(dataPath)) {
     const fetchUrl = `https://s3.amazonaws.com/${process.env.S3_BUCKET}/autocomplete/locationAutocompleteData.json`;
     console.debug({ fetchUrl });
     const { body } = await fetch(fetchUrl);
+    if (!body) {
+      throw new Error(`Failed to fetch: ${fetchUrl}`);
+    }
+    // @ts-ignore If body is `null` we are already throwing an error above.
     await finished(Readable.fromWeb(body).pipe(stream));
```
rhelmer, line 22:
```suggestion
    await finished(Readable.fromWeb(body!).pipe(stream));
```

rhelmer: I think this is cleaner than ts-ignore and will result in the same thing at runtime I think, *but* I think the real potential problem it's pointing out is that `body` may be non-null but still not a `ReadableStream`. I'm not sure there's much we can do to handle it in this case though, my suggestion should throw a `TypeError` at runtime which is probably the best we can do.

rhelmer (changes requested):
This is great, thanks! A few questions and suggestions. It's nice that we can use JSDoc-style comments to shore up the older code like migrations.

src/db/migrations/20221026215921_add-breaches-table.js
flozia: I don’t think so — at least I couldn’t find any reference. I was just being cautious about updating existing migration files. Sounds like maybe we should.

src/db/migrations/20231102024624_add_unique_index_to_scan_results_id.js
flozia: Ah, yes. I’ve copied the wrong `ts-ignore` comment over. The line is still getting highlighted by the linter complaining that it should be `table.dropUnique(["onerep_scan_id"]);` it also just takes a `string`.

src/scripts/loadtest/hibp.js
```diff
@@ -67,10 +66,12 @@ export const run = () => {
 
   try {
     const result = res.json();
+    // @ts-ignore TODO: Add type for result.
```
Vinnl, line 69:
```suggestion
    const result = res.json();
    // @ts-ignore TODO: Add `PostHibpNotificationResponseBody` type to `src/app/api/v1/hibp/notify/route`, and use it.
```

Vinnl (approved):
Thanks for being so thorough!

src/db/migrations/20221026215921_add-breaches-table.js
Vinnl: Unless we're nuking our prod database (please don't :P) I think it's better to have the migration files accurately reflect how the prod DB got to its current state.

flozia: Yes, that is why I’ve just added the ignores. What do you think @rhelmer and @mansaj about leaving those as is vs. updating?

src/db/migrations/20230703010356_add-onerep-profiles-table.js
flozia: It looks like `varchar` is implemented, but missing from the type [TableBuilder](https://github.com/knex/knex/blob/master/types/index.d.ts#L2420). I think it is recommended to use `.string()`.

src/db/migrations/20240423150332_add_suffix_and_middle_name_to_onerep_profile_table.js
flozia: See https://github.com/mozilla/blurts-server/pull/5622#discussion_r1954465432.

src/scripts/build/getAutoCompleteLocations.js
flozia: Since this is a `.js` file non-null assertion won’t work here. :/

src/db/migrations/20221026215921_add-breaches-table.js
rhelmer: FWIW, if we want to ensure that changes to existing migration files don't change the schema, we could test it locally:

1. create new db, run "before" migrations, dump schema (`pgdump`)
2. apply changes to migration scripts
3. repeat step 1 and diff the "before"/"after" schema

It's possible for migration scripts to insert/delete data too but I don't think we tend to do that.

src/db/migrations/20230703010356_add-onerep-profiles-table.js
rhelmer: Ah OK, yeah "under the hood" in Postgres I believe TEXT and VARCHAR types are implemented the same way so `.text()` is probably what you want https://knexjs.org/guide/schema-builder.html#text

src/scripts/build/getAutoCompleteLocations.js
rhelmer: Oh shoot right sorry I forgot, yeah that's fine then. Unfortunate to have to turn it off for the whole line that's all. I guess if you care you could explicitly check this and throw.

rhelmer (approved):
lgtm, thanks for slogging through that! I'd say in general you've erred on the safe side rather than changing existing migration code (and risk mutating the schema for new vs. existing DB instances) which seems appropriate.

src/scripts/build/uploadAutoCompleteLocations.js
```diff
@@ -293,10 +337,30 @@ try {
             return false;
           }
 
-          return locationDataPopulated.some((location) => {
+          return locationDataRows.some((location) => {
```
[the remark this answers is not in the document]
flozia: So the funny thing is: The result of this filter is still correct. The goal here is to filter populated locations that are second-order administrative divisions that are also labeled as a populated place. In the dataset we use I’ve only encountered the [5 New York Burrows](https://www.geonames.org/5128581/new-york-city.html) to match this criteria.

Still, the naive update I made here increases the runtime to an unreasonable degree. Since this is not breaking anything and the aim of this PR was not to make optimizations to this script I’ve created [MNTOR-4150](https://mozilla-hub.atlassian.net/browse/MNTOR-4150) for a follow-up to keep this PR contained. [00d0e13](https://github.com/mozilla/blurts-server/pull/5622/commits/00d0e13c24a732c14fc5b2af941b7310525fb089) this for now.

src/scripts/loadtest/hibp.js
```diff
@@ -67,10 +66,12 @@ export const run = () => {
...
     if (result.success !== true) {
       throw new Error(`Non-success result: ${JSON.stringify(result)}`);
     }
   } catch {
+    // @ts-ignore TODO: Add type for result.
```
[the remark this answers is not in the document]
flozia: Addressed in [7c35c2e](https://github.com/mozilla/blurts-server/pull/5622/commits/7c35c2e73a9eb5b10d87e8a7c249eb4d47eb193f).

PR #5623, opened 2025-02-13
Workaround inaccurate Jest coverage reporting

Vinnl:
This directly applies the following Pull Request:
https://github.com/SimenB/collect-v8-coverage/pull/235

It's not entirely clear why it's needed, but it seems reasonable
and harmless enough, and there are no signs that Jest is really
maintained anymore, so a longer-term fix does not appear to be
forthcoming. Thus, this'll at least properly unblock Node upgrades
and other annoying coverage surprises.


flozia (approved):
Thank you for this. So nice to see those `c8 ignore`’s go.

rhelmer (approved):

PR #5639, opened 2025-02-18
Upgrade to Node 20.18.2

Vinnl:
This makes sure we have the latest security patches. Ideally Dependabot would've submitted a PR with this, but I'm not sure why it doesn't - possibly because we've closed PRs to newer (major) versions?


Vinnl:
Nooooo, there's still a coverage issue :cry: But it is consistent across runs, and seems so be a wider manifestation of an existing problem, so I added 9be3f10b0 now in the hope that that is the only change we'll have to do. If it fails again, I'm giving up again for now.

rhelmer (approved):

PR #5640, opened 2025-02-18
MNTOR-3946: State that we don't endorse breached sites

Vinnl:
# References:

Jira: MNTOR-3946
Figma:

# Description

We link to breached sites, but we don't endorse or advertise them, and search engines and other agents shouldn't interpret our links to them as such. That is what [`nofollow`](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/rel#nofollow) does.

# How to test

Visit a public breach page, verify that its link has a `nofollow` attribute.

### DIFF
--- a/src/app/(proper_react)/(redesign)/(public)/breach-details/[breachName]/BreachDetailView.tsx
+++ b/src/app/(proper_react)/(redesign)/(public)/breach-details/[breachName]/BreachDetailView.tsx
@@ -117,7 +117,7 @@ export const BreachDetailsView = (props: Props) => {
               href={`https://${breach.Domain}`}
               eventData={{ link_id: breach.Domain }}
               target="_blank"
-              rel="noopener noreferrer"
+              rel="nofollow noopener noreferrer"
             >
               {breach.Domain}
             </TelemetryLink>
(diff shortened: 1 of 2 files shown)


src/app/(proper_react)/(redesign)/(public)/breach-details/[breachName]/BreachDetailView.tsx
```diff
@@ -117,7 +117,7 @@ export const BreachDetailsView = (props: Props) => {
               href={`https://${breach.Domain}`}
               eventData={{ link_id: breach.Domain }}
               target="_blank"
-              rel="noopener noreferrer"
```
Comment (rhelmer) [inline src/app/(proper_react)/(redesign)/(public)/breach-details/[breachName]/BreachDetailView.tsx:120]: Per https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a#target `target="_blank"` automatically implies `noopener`, do we need to keep `noreferrer`?

rhelmer (approved):
qq about `noreferrer`, I think we might want to keep that but otherwise lgtm.

src/app/(proper_react)/(redesign)/(public)/breach-details/[breachName]/BreachDetailView.tsx
Comment (Vinnl) [reply src/app/(proper_react)/(redesign)/(public)/breach-details/[breachName]/BreachDetailView.tsx:120]: Good call - Amri told me once that they're no longer needed because browsers do it by default, but digging some more, for `noreferrer` it's actually that browsers set [a `strict-origin-when-cross-origin` referrer policy](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Referrer-Policy#strict-origin-when-cross-origin), which only sends the top-level domain. Sending nothing at all is probably even better. I'm just going to restore these `rel`s, because they don't do any harm either.

Comment (Vinnl) [reply src/app/(proper_react)/(redesign)/(public)/breach-details/[breachName]/BreachDetailView.tsx:120]: 6b0ea6965

PR #5641, opened 2025-02-18
MNTOR-3947 Tell search engines which pages make sense to list

Vinnl:
# References:

Jira: MNTOR-3947
Figma:

# Description

Essentially, only our public pages are relevant to surface in search results.

See https://nextjs.org/docs/app/api-reference/file-conventions/metadata/robots and https://nextjs.org/docs/app/api-reference/file-conventions/metadata/sitemap.

# How to test

Visit http://localhost:6060/sitemap.xml and http://localhost:6060/robots.txt and check that you don't see anything unexpected.

- [ ] I've added a unit test to test for potential regressions of this bug. - N/A, trivial


src/app/robots.ts
```diff
@@ -0,0 +1,16 @@
...
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import type { MetadataRoute } from "next";
+
+export default function robots(): MetadataRoute.Robots {
+  return {
+    rules: {
+      userAgent: "*",
+      allow: "/",
+      disallow: ["/user/dashboard/", "/user/settings/", "/user/welcome/"],
```
flozia, line 12: Question: Is it safer to disallow the paths explicitly or why not just do `/user/*`?

flozia (approved):

src/app/robots.ts
Vinnl: Yeah I did consider that, but I was worried that I couldn't be confident that we wouldn't in the future put any public-facing pages under `/user/`, and then not realise that they wouldn't be crawled. That failure mode seemed worse to me than a loginwalled page inadvertently being indexed.

rhelmer (approved):

Vinnl:
@rhelmer Do we all have access to that? If not, that page also says that merely linking to the sitemap in `robots.txt` should be enough to find it when they next crawl the site, which I expect to happen pretty frequently.

PR #5643, opened 2025-02-19
Logging refinements

Vinnl:
While debugging breach sending there was a bit of info that I would've wanted to have, so I added that here. I also saw some information that I'd rather not see, so I removed that.


rhelmer (approved):
lgtm, thanks!

PR #5645, opened 2025-02-19
Add notifications table

codemist:
# References:

Jira: MNTOR-3951
Figma:

# Description

# Screenshot (if applicable)

Not applicable.

# How to test

### DIFF
--- a/src/app/(proper_react)/(redesign)/(authenticated)/admin/feature-flags/components/FlagEditor.tsx
+++ b/src/app/(proper_react)/(redesign)/(authenticated)/admin/feature-flags/components/FlagEditor.tsx
@@ -16,9 +16,13 @@ import { Button } from "../../../../../../components/client/Button";
 import { CreateFeatureFlagRequestBody } from "../../../../../../api/v1/admin/feature-flags/route";
 import { FeatureFlagName } from "../../../../../../../db/tables/featureFlags";
 
-export const NewFlagEditor = (props: { flagName: FeatureFlagName }) => {
+export const NewFlagEditor = (props: {
+  flagName: FeatureFlagName;
+  adminOnly: boolean;
+}) => {
   return (
     <FlagEditor
+      adminOnly={props.adminOnly}
       flagName={props.flagName}
       isEnabled={false}
       onToggleEnable={async (isEnabled) => {
@@ -45,9 +49,13 @@ export const NewFlagEditor = (props: { flagName: FeatureFlagName }) => {
   );
 };
 
-export const ExistingFlagEditor = (props: { flag: FeatureFlagRow }) => {
+export const ExistingFlagEditor = (props: {
+  flag: FeatureFlagRow;
+  adminOnly: boolean;
+}) => {
   return (
     <FlagEditor
+      adminOnly={props.adminOnly}
       flagName={props.flag.name}
       isEnabled={props.flag.is_enabled}
       onToggleEnable={async (isEnabled) => {
@@ -82,6 +90,7 @@ type Props = {
   isEnabled: boolean;
   onToggleEnable: (isEnabled: boolean) => Promise<void>;
   allowList: string[];
+  adminOnly: boolean;
   onUpdateAllowlist: (allowList: string[]) => Promise<void>;
 };
 const FlagEditor = (props: Props) => {
@@ -114,8 +123,9 @@ const FlagEditor = (props: Props) => {
             variant="secondary"
             onPress={() => void setIsEnabled(true)}
             small
+            disabled={props.adminOnly}
           >
-            Enable for everyone
+            {props.adminOnly ? "Allow list only" : "Enable for everyone"}
           </Button>
         )}
       </div>
--- a/src/app/(proper_react)/(redesign)/(authenticated)/admin/feature-flags/page.tsx
+++ b/src/app/(proper_react)/(redesign)/(authenticated)/admin/feature-flags/page.tsx
@@ -8,6 +8,7 @@ import {
   FeatureFlagName,
   featureFlagNames,
   getAllFeatureFlags,
+  isFeatureFlagAdminOnly,
 } from "../../../../../../db/tables/featureFlags";
 import { isAdmin } from "../../../../../api/utils/auth";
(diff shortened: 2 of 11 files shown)


src/db/migrations/20250219182508_add_notifications_table.js
```diff
@@ -0,0 +1,32 @@
...
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+/**
+ * @param { import("knex").Knex } knex
+ * @returns { Promise<void> }
+ */
+export function up(knex) {
+  return knex.schema.createTable("notifications", (table) => {
+    table.increments("notification_id").primary();
+    table.string("label").nullable();
```
rhelmer, line 12: I believe `nullable` is the default, it's fine to include but I expect you can exclude if you want: https://knexjs.org/guide/schema-builder.html#nullable

src/db/migrations/20250219182508_add_notifications_table.js
```diff
@@ -0,0 +1,32 @@
...
+    table.string("title").nullable();
+    table.string("description").nullable();
+    table.string("small_image_path").nullable();
+    table.string("big_image_path").nullable();
+    table.string("cta_label").nullable();
+    table.string("cta_link").nullable();
+    table.string("string").nullable();
+    table.string("audience").nullable();
```
rhelmer, line 20: Could any of these be over 255 characters? I'd consider `text` instead of `string`, I believe knex defaults to a `varchar(255)` https://knexjs.org/guide/schema-builder.html#string which `text` is unlimited.

The distinction is a little confusing, in Postgres `varchar` and `text` are both unlimited by default, but you can specify a limit with `varchar(n)` which I suspect knex is doing from their docs.

rhelmer (changes requested):
lgtm overall, a few questions and suggestions. If you know what your query might look like (e.g. if you're going to filter on a particular column), you can tell Postgres you want to index it with `.index()` as well, but since the table isn't going to grow that fast you could skip that for now. Just something to keep in mind if the table gets large, if there's an index then lookups (e.g. `WHERE` clause) will much faster.

src/db/migrations/20250219182508_add_notifications_table.js
Vinnl: Personally, since it's such an important property, I'd prefer being explicit. If only because it triggers the relevant question: should these actually all be nullable? Are there notifications that don't have a title, for example?

src/db/migrations/20250219182508_add_notifications_table.js
```diff
@@ -0,0 +1,32 @@
...
+  return knex.schema.createTable("notifications", (table) => {
+    table.increments("notification_id").primary();
+    table.string("label").nullable();
+    table.string("title").nullable();
+    table.string("description").nullable();
+    table.string("small_image_path").nullable();
+    table.string("big_image_path").nullable();
+    table.string("cta_label").nullable();
+    table.string("cta_link").nullable();
+    table.string("string").nullable();
```
Vinnl, line 19: Is this an actual column that's needed? :sweat_smile:

src/db/migrations/20250219182508_add_notifications_table.js
```diff
@@ -0,0 +1,32 @@
...
+ * @returns { Promise<void> }
+ */
+export function up(knex) {
+  return knex.schema.createTable("notifications", (table) => {
+    table.increments("notification_id").primary();
+    table.string("label").nullable();
+    table.string("title").nullable();
+    table.string("description").nullable();
+    table.string("small_image_path").nullable();
+    table.string("big_image_path").nullable();
```
Vinnl, line 16: So... Where are these paths supposed to point? It sounds like they might need to be in our repository, which means they would need an engineer to add, which means... That this might actually get more of a hassle than it would be if an engineer could just add the whole notification in one go, like we did on Relay?

Vinnl (commented):
Could you also add the type definition of what a `NotificationsRow` might look like to `knex-tables.d.ts`? Then we get autocompletion when writing queries. Specifically, you'll have to add a `notifications` property to [the `Tables` interface](https://github.com/mozilla/blurts-server/blob/05721742caaa245f2bda70d5eeb9bcf7d04b6456/src/knex-tables.d.ts#L377), and can look at the other properties for how that's done (and ask me if you need more help).

And _ideally_, you'd make that part of the same commit that added the migration, so that it's easy to find the migration through `git blame` later.

src/db/migrations/20250219182508_add_notifications_table.js
```diff
@@ -0,0 +1,32 @@
...
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+/**
+ * @param { import("knex").Knex } knex
+ * @returns { Promise<void> }
+ */
+export function up(knex) {
+  return knex.schema.createTable("notifications", (table) => {
+    table.increments("notification_id").primary();
```
mansaj, line 11: nit: we usually just use "id" as the incremental primary key for a table

We would use names like "xxx_id" to signal a foreign key (an ID that comes from another table but referenced in this one, ie subscriber_id)

src/db/migrations/20250219182508_add_notifications_table.js
mansaj: I also don't think every field being `nullable` make sense here. I'm assuming that majority of these fields are actually required no?

src/db/migrations/20250219182508_add_notifications_table.js
Vinnl: I'm assuming this wasn't a reply to my question about having a column called `string`, but to my question about the type definitions.

So just for context, the reason for that is that we can then roll back the unrelated changes without having to remove the migrations (because Knex complains if there are migrations logged in the DB for which there are no files in the repo). Since the type definitions are related to the migrations (i.e. they define what the DB look like after the migrations), it's fine to include them with the migration PR/commit.

src/db/migrations/20250219182508_add_notifications_table.js
Vinnl: I moved this discussion to chat, since it'll be out of scope for this PR.

PR #5659, opened 2025-02-25
Fix qa_custom_brokers bugs

codemist:
# References:

Jira: MNTOR-
Figma:

# Description

- Introduces a parameter of `onerepScanId` into `getAllMockedScanResults` and `getMockedScanResultsWithBrokerUnderMaintenance` to filter for different results for different users.
- Fixes the format for `last_optout_at` 
- Removes mock logic data in `getLatestOnerepScanResults` 
- Adds requested removal option to the config page

# Screenshot (if applicable)

Not applicable.

# How to test


src/db/tables/onerep_scans.ts
```diff
@@ -159,43 +159,20 @@ async function getLatestOnerepScanResults(
...
-
-    const qaBrokers = !showCustomBrokers ? [] : await getAllMockedScanResults();
-    if (!showRealBrokers) {
-      logger.info("get_latest_results_custom_brokers", {
-        onerepProfileId,
-        onerepScanId: scan?.onerep_scan_id,
-        qaBrokers,
-      });
-      results = qaBrokers;
-    } else {
```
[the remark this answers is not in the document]
mansaj: so without the data switch, will QAs be able to get to the mock data?

src/db/tables/qa_customs.ts
```diff
@@ -135,12 +135,17 @@ async function addQaCustomBroker(
...
 }
 
-async function getAllMockedScanResults(): Promise<
-  OnerepScanResultDataBrokerRow[]
-> {
-  const res = (await knex("qa_custom_brokers").select(
-    "*",
-  )) as OnerepScanResultDataBrokerRow[];
+async function getAllMockedScanResults(
+  onerepScanId: number | null,
```
mansaj, line 139: Why is this scan id instead of profile ID?
Was this using scan ID before?

src/db/tables/onerep_scans.ts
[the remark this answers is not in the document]
mansaj: Oh is this the part where we are mixing the QA data and real data?
Yea we wanted that gone right? So there's no mixing of the data

src/db/tables/qa_customs.ts
mansaj: Is this the QA table or the real table?

This is really confusing. If this is for the QA table, we can either rename or add the right column.

src/app/(proper_react)/(redesign)/(authenticated)/admin/qa-customs/onerepConfig.tsx
```diff
@@ -401,8 +407,12 @@ const OnerepConfigPage = ({
                   className={styles.input}
                   type="date"
                   name="last_optout_at"
-                  placeholder={new Date().toISOString().split("T")[0]}
-                  value={newBroker.last_optout_at?.toDateString()}
+                  // placeholder={new Date().toISOString().split("T")[0]}
```
mansaj, line 410: what was this placeholder for?

src/db/tables/qa_customs.ts
```diff
@@ -135,12 +135,21 @@ async function addQaCustomBroker(
...
+): Promise<OnerepScanResultDataBrokerRow[]> {
+  if (onerepProfileId === null) {
+    logger.error(`onerepProfileId not set`);
+  }
+
+  // Using the onerep_scan_id  as a placeholder for the profile ID
+  const res = (await knex("qa_custom_brokers")
+    .select("*")
+    .where(
+      "onerep_scan_id",
```
mansaj, line 149: This is very confusing for maintainability. We should change the name in the table `qa_custom_brokers` from `scan_id` to `profile_id` if that's the place we are storing this:

Specifically using: `table.renameColumn`

We can find examples of similar changes in: `20190328111900_add_email_addresses_table.js`

mansaj:
LGTM, just one comment about the table

src/db/tables/qa_customs.ts
rhelmer: @codemist can add a TODO to the comment with a link to a ticket to fix this? I'm a little confused about the issue, if it's just too much to change right now, or if there's a good idea to use the same value for both IDs. Even in the latter case I think we should spend some more time looking for a better solution here, but that can be its own separate ticket that someone can work on.

rhelmer (approved):
lgtm overall, with a caveat about adding a followup ticket to clean up the use of `scan_id` vs. `profile_id` (using the same value for both is OK but we should make the code more explicit so it's clear)

PR #5662, opened 2025-02-25
Make feature flags auditable

Vinnl:
# References:

Jira: MNTOR-4062
Figma:

# Description

The main change here is deprecating the `feature_flags` table (I wasn't brave enough to drop it yet), and replacing it by a `feature_flag_events` table that is append-only. This will allow us to have a full audit trail of when feature flags get modified.

To keep a similar experience as the `feature_flags` table, I added a `feature_flag_view` view.  Note that I did not preserve the wait list, description, dependencies, expiration, deletion, and owner, since we haven't used those.

The feature flag UI exposes the last date a flag was modified, and who modified it - but the full history is available in the DB if we ever need it.

I also updated the feature flag API to list all enabled flags when the user is not logged in. This will allow us to compare stage's and prod's enabled feature flags before deploying in the future, and warn the BLE to flip back a flag in stage if they don't match.

I can create a PR with just c76fc434a616438aecdc57e5e83c3484a6be9fb4 if the overall approach looks OK.

# How to test

Use feature flags!

Also, the migration is more complex than usual, also moving over data, so doing the migration, verifying that there is no data loss, modifying some flags, and rolling back and again verifying that there is no data loss, is probably a good idea.

- [ ] I've added a unit test to test for potential regressions of this bug. - No, side-effect heavy (just DB)
- [ ] All acceptance criteria are met. - N/A
- [ ] Jira ticket has been updated (if needed) to match changes made during the development process. - N/A
- [ ] Jira ticket has been updated (if needed) with suggestions for QA when this PR is deployed to stage. - N/A

### DIFF
--- a/src/app/api/v1/admin/feature-flags/route.ts
+++ b/src/app/api/v1/admin/feature-flags/route.ts
@@ -10,9 +10,6 @@ import {
   addFeatureFlag,
   enableFeatureFlagByName,
   disableFeatureFlagByName,
-  updateAllowList,
-  updateWaitList,
-  deleteFeatureFlagByName,
 } from "../../../../../db/tables/featureFlags";
 
 import { isAdmin } from "../../../utils/auth";
@@ -46,7 +43,10 @@ export type CreateFeatureFlagRequestBody = {
 
 export async function POST(req: NextRequest) {
   const session = await getServerSession();
-  if (isAdmin(session?.user?.email || "")) {
+  if (
+    isAdmin(session?.user?.email || "") &&
+    typeof session?.user?.subscriber?.id === "number"
+  ) {
     // Signed in
     try {
       const newFlag = (await req.json()) as CreateFeatureFlagRequestBody;
@@ -56,12 +56,8 @@ export async function POST(req: NextRequest) {
       const resp = await addFeatureFlag({
         name: newFlag.name,
         is_enabled: newFlag.isEnabled,
-        description: newFlag.description,
-        dependencies: newFlag.dependencies,
         allow_list: newFlag.allowList,
-        wait_list: newFlag.waitList,
-        expired_at: newFlag.expiredAt,
-        owner: newFlag.owner,
+        last_updated_by_subscriber_id: session.user.subscriber.id,
       });
       return NextResponse.json(resp);
     } catch (e) {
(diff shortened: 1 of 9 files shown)


src/db/migrations/20250225111207_add_feature_flag_events.js
```diff
@@ -0,0 +1,90 @@
...
+    .first();
+  if (!fakeFirstuser) {
+    throw new Error(
+      "Couldn't find a subscriber with an @mozilla.com email address to attribute pre-existing feature flags to.",
+    );
+  }
+  await knex("feature_flag_events").insert(
+    existingFeatureFlags.map((flag) => ({
+      ...flag,
+      created_by_subscriber_id: fakeFirstuser.id,
```
Vinnl, line 67: Good to call out: we're currently not tracking who last touched a feature flag, so I just pretended that the first subscriber with an `@mozilla` email address did (this is probably Luke I suppose). Since this is for our eyes only, it doesn't really matter, I don't think.

Vinnl: Oh, I guess this doesn't work on an empty DB, like in our e2e tests. I'd really prefer not to make the column nullable though...

mansaj (approved):
LGTM, thanks for taking out all the tech debt.

Re deprecating types and functions, are there tickets to get rid of these? If so should we make a note in the comments

src/app/api/v1/admin/feature-flags/route.ts
```diff
@@ -28,7 +27,11 @@ export async function GET() {
       return NextResponse.json({ success: false }, { status: 500 });
     }
   } else {
-    return NextResponse.json({ success: false }, { status: 401 });
+    const flagNamesInDb = await getEnabledFeatureFlags({ isSignedOut: true });
+    const availableFlags = featureFlagNames;
+    return NextResponse.json(
+      availableFlags.filter((flag) => flagNamesInDb.indexOf(flag) !== -1),
```
rhelmer, line 33: I think this function could use a comment to explain the difference if logged in as an admin or not.

Should we consider making the `isAdmin` function a little easier to use, and maybe have some way to ensure that it's used totally consistently? I kind of wish everything denied access by default unless we allow-listed it, instead of the other way around, at least in the `admin` directory.

rhelmer (approved):
lgtm, nice use of views! I left a small suggestion, and a complaint about how I wish it was easier to use authorization checks.

src/app/api/v1/admin/feature-flags/route.ts
Vinnl: Hmm, so I added a comment in 65b034ed5, but you're right, it's pretty weird to vary the response structure based on whether the user is an admin or not. Of course this is not a public API, and I don't think there are any places that call this API yet, so it's not a pressing issue, but it's not great either way.

We could add middleware that intercepts `/api/v1/admin/` requests and checks whether the user is an admin, and then move this API outside of that namespace. But not super high priority IMO.

Vinnl:
@mansaj Good callout, just created the ticket (MNTOR-4191) and added the comments (e1a7345c708fa2b96a3f7dd893904b3ff47e5812).

PR #5660, opened 2025-02-25
Add option to add and update additional scan profile details

flozia:
# References:

Jira: [MNTOR-4065](https://mozilla-hub.atlassian.net/browse/MNTOR-4065)
Figma: https://www.figma.com/design/2iCgADpFXKOFZTEgCPkQsR/Settings?node-id=597-10052&t=c2UkGyKPdbvneuz5-4

# Description

Adds additional scan profile details and the option to update them:
1. The migration file for the changes in this PR is split out to PR #5664.
2. Updates the `/admin/dev` panel to add the option to test the backend changes.

# How to test

1. Run the migrations: `npm run db:migrate`
2. Run an initial welcome scan
3. Enable the feature flag `EditScanProfileDetails`
4. Make sure you can still run an initial scan
5. Visit `/admin/dev`
6. Make changes to the scan profile
7. Confirm any profile updates are applied to the table `onerep_profiles` and also show for the remote profile

# Checklist (Definition of Done)

- [x] Commits in this PR are minimal and [have descriptive commit messages](https://chris.beams.io/posts/git-commit/).
- [x] I've added or updated the relevant sections in code comments
- [x] I've added a unit test to test for potential regressions of this bug.
- [x] All acceptance criteria are met.
- [x] Jira ticket has been updated with suggestions for QA when this PR is deployed to stage.

[MNTOR-4065]: https://mozilla-hub.atlassian.net/browse/MNTOR-4065?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #5664, opened 2025-02-26
Migration for PR #5660

flozia:
# References:

PR: #5660
Jira: [MNTOR-4065](https://mozilla-hub.atlassian.net/browse/MNTOR-4065)

# Description

Migration file for PR #5660 that is adding new optional columns to `onerep_profiles`.

# Screenshot (if applicable)

Not applicable.

# How to test

- Run `npm run db:migrate`: The table `onerep_profiles` should contain the applied changes.
- Run `npm run db:rollback`: The table `onerep_profiles` should have its initial state.

[MNTOR-4065]: https://mozilla-hub.atlassian.net/browse/MNTOR-4065?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/db/migrations/20250218094907_alter_and_add_profile_details_to_onerep_profiles_table.js
```diff
@@ -0,0 +1,63 @@
...
+
+    await knex.raw(`
+      UPDATE "onerep_profiles"
+      SET "addresses" = jsonb_build_array(
+        jsonb_build_object(
+          'city', "city_name",
+          'state', "state_code"
+        )
+      )
+      WHERE "city_name" IS NOT NULL AND "state_code" IS NOT NULL;
```
Vinnl, line 37: I take it these are always _both_ null or not null at the same time?

src/db/migrations/20250218094907_alter_and_add_profile_details_to_onerep_profiles_table.js
```diff
@@ -0,0 +1,63 @@
...
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+/**
+ * @param { import("knex").Knex } knex
+ * @returns { Promise<void> }
+ */
+export function up(knex) {
+  return knex.schema.alterTable("onerep_profiles", async (table) => {
+    table
+      .specificType("first_names", "character varying(255)[]")
```
Vinnl, line 12: With backend/DB design not being my speciality, I don't want to have a strong opinion here, but I suppose the following are reasons why it's OK to use array types and JSON rather than the more cumbersome foreign tables:

- We don't care about DBMS portability,
- we won't be selecting rows based on the contents of these columns,
- we'll rarely write to these columns, and
- when we do write to them, we'll often update all elements in one go.

Writing this down more for myself, but happy to learn if anyone has corrections or additions.

src/db/migrations/20250218094907_alter_and_add_profile_details_to_onerep_profiles_table.js
```diff
@@ -0,0 +1,63 @@
...
+      .defaultTo(knex.raw(`ARRAY[]::character varying(255)[]`));
+    table.jsonb("addresses").notNullable().defaultTo("[]");
+
+    await knex.raw(`
+      UPDATE "onerep_profiles"
+      SET "addresses" = jsonb_build_array(
+        jsonb_build_object(
+          'city', "city_name",
+          'state', "state_code"
+        )
```
Vinnl, line 35: Shouldn't we also update the `first_names`, `middle_names`, etc. columns as well to copy the existing data?

src/db/migrations/20250218094907_alter_and_add_profile_details_to_onerep_profiles_table.js
```diff
@@ -0,0 +1,63 @@
...
+    `);
+
+    table.dropColumns(
+      "first_names",
+      "last_names",
+      "middle_names",
+      "phone_numbers",
+      "addresses",
+    );
+  });
```
Vinnl, line 62: Rolling back the migration doesn't actually drop these columns for me...

Vinnl: They do if I remove the `knex.raw` above. Possibly that should not be part of the `alterTable` function?

src/db/migrations/20250218094907_alter_and_add_profile_details_to_onerep_profiles_table.js
flozia: I’ve considered adding a new table, but that might be too much overhead. We are already using `jsonb` in a couple of places. That means:
- we don’t have DBMS portability either way
- I don’t anticipate that we need to select rows based on these columns
- writing to those columns happens only if users update their profile info
- we only update the elements in those columns all at once

src/db/migrations/20250218094907_alter_and_add_profile_details_to_onerep_profiles_table.js
flozia: `first_name` and `middle_name` stay as separate columns and only alternate first and last names will be written to `first_names` and `middle_names`. That matches the schema for [OneReps profiles](https://docs.onerep.com/#operation/getProfile).

src/db/migrations/20250218094907_alter_and_add_profile_details_to_onerep_profiles_table.js
flozia: Yes, that’s right!

src/db/migrations/20250218094907_alter_and_add_profile_details_to_onerep_profiles_table.js
Vinnl: Ah gotcha.

src/db/migrations/20250218094907_alter_and_add_profile_details_to_onerep_profiles_table.js
Vinnl: Yep, works now!

Vinnl (approved):
I was now able to run this migration, and on rollback I saw the data getting carried over from the new columns to the old ones :shipit:

src/db/migrations/20250218094907_alter_and_add_profile_details_to_onerep_profiles_table.js
```diff
@@ -0,0 +1,63 @@
...
+/**
+ * @param { import("knex").Knex } knex
+ * @returns { Promise<void> }
+ */
+export async function down(knex) {
+  await knex.raw(`
+    UPDATE "onerep_profiles"
+    SET "city_name" = COALESCE(("addresses"->0->>'city'), ''),
+        "state_code" = COALESCE(("addresses"->0->>'state'), '')
+    WHERE jsonb_array_length(COALESCE("addresses", '[]'::jsonb)) > 0;
```
rhelmer, line 51: Have you tested this?

rhelmer (approved):
I wish we didn't need so much `knex.raw` here, but I agree that we're not aiming to be RDBMS-agnostic at all, and arrays are a very useful feature of Postgres.

The code looks fine but my main concern is how this is tested, we should have a plan for how to test this on stage.

I'm a little worried about the rollback `down` function, we should make sure to test that thoroughly. It would be far easier to roll back than to restore a database backup, but I really don't want to have to use it.

src/db/migrations/20250218094907_alter_and_add_profile_details_to_onerep_profiles_table.js
Vinnl: I have tested this locally at least, so there's that. (i.e. I modified an address locally, then rolled back, and saw the original fields updated.)

flozia: Yes, this has been tested locally.

flozia:
@rhelmer @Vinnl I’m moving the migration of the `addresses` column to the separate PR https://github.com/mozilla/blurts-server/pull/5688 so https://github.com/mozilla/blurts-server/pull/5660 is not blocked.

Vinnl (approved):
:+1:

PR #5684, opened 2025-03-04
MNTOR-4151: Use higher-res images for the logo in emails

Vinnl:
# References:

Jira: MNTOR-4151
Figma:

# Description

This might make the logo look sharper on hi-res screens.

# Screenshot (if applicable)

There's no change for the regular header, but apparently the logo was a bit further away in the hero header with the purple background, and I think the font is a bit different.

Old:

image

New:

image

# How to test

Use the admin tool to send both the free and Plus monthly activity emails to a Mailinator address. The image URLs will then be served from localhost, so you can run the server to see what they look like in Mailinator.

- [ ] I've added a unit test to test for potential regressions of this bug. - N/A, visual


flozia (approved):
Right now the logo with the background has an additional padding of `10px`. This matches the previous images, but wouldn’t the images ideally have the same size?

Vinnl:
@flozia Yeah, we added the burnt-in padding when we added the background, to make it slightly less jarring in dark mode, but that does result in some extra padding for that image. We should probably add a negative margin to that image in the regular header, but... Email is scary, it's not a huge issue, and apparently we've been living with it for a while, so... It's probably the best use of our time to leave it like this.

PR #5685, opened 2025-03-05
MNTOR-4180: Gracefully handle cronjob interruptions

Vinnl:
# References:

Jira: MNTOR-4180
Figma:

# Description

Since the SIGINT handling of https://github.com/mozilla/blurts-server/pull/5658 seems to have resolved the issues there, this PR implements that for all cronjobs that close email pools or database connections. Instead of just terminating when they receive a SIGINT, I added a handler that first does the closing before exiting.

I also removed the long-running task code from the `emailBreachAlerts` job, since that wasn't necessary to resolve the issue.

# How to test

For the jobs that take a while to complete, you can see if they terminate gracefully if you run `kill -SIGINT <process ID>`. For the ones that complete almost instantaneously, I'm not sure if there's a good way to test the interrupt handling, but you can at least run them and see if normal functionality isn't impacted.

- [ ] I've added a unit test to test for potential regressions of this bug. - N/A, just side effects.


src/scripts/cronjobs/firstDataBrokerRemovalFixed.tsx
```diff
@@ -145,3 +146,13 @@ async function sendFirstDataBrokerRemovalFixedActivityEmail(
     ),
   );
 }
+
+function isFulfilledResult(
```
rhelmer, line 150: I know this was just moved and not new, just curious why this is needed? Does it do the same thing as [Promises.allSettled()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/allSettled)?

src/scripts/cronjobs/emailBreachAlerts.tsx
```diff
@@ -393,19 +393,6 @@ async function pullMessages() {
...
   return [subClient, response.receivedMessages] as const;
 }
-
-async function run() {
-  const [subClient, receivedMessages] = await pullMessages();
-  await poll(subClient, receivedMessages ?? []);
-
-  // We're transitioning this into a long-running process,
-  // but to ensure a smooth transition, allow keeping the cronjob behaviour:
-  if (process.env.LOOP === "true") {
```
rhelmer, line 403: You mentioned in the PR top-level comment that this wasn't necessary to resolve the issue, do we still want this behavior? I thought this was moved from a cron job to a long-running task, is there a shell script or something wrapping this?

Just FYI, I believe that Node will eventually run out of stack space running recursively like this, you probably want a `while` or other type of iterative loop (because JS doesn't have tail call optimization).

rhelmer (approved):
lgtm overall, a few questions on bits I'm confused about but shouldn't be blockers I don't think. @Vinnl do you need help testing?

src/scripts/cronjobs/emailBreachAlerts.tsx
```diff
@@ -393,19 +393,6 @@ async function pullMessages() {
 
   return [subClient, response.receivedMessages] as const;
 }
```
mansaj, line 395: Just to understand here, transitioning from task back to cron means that SRE will need to switch back to running `cron:breach-alerts`

src/scripts/cronjobs/firstDataBrokerRemovalFixed.tsx
```diff
@@ -19,24 +19,25 @@ import { sanitizeSubscriberRow } from "../../app/functions/server/sanitize";
...
-  );
-}
+process.on("SIGINT", () => {
+  logger.info("SIGINT received, exiting...");
+  tearDown();
+});
 
 void run();
 
+function tearDown() {
```
mansaj, line 36: nit: the teardown can be pulled into a shared function.

mansaj (approved):

src/scripts/cronjobs/emailBreachAlerts.tsx
Vinnl: As I understand it, [Brandon never actually activated](https://mozilla.slack.com/archives/C08C8P5RTBN/p1741106222206889?thread_ts=1741027363.416669&cid=C08C8P5RTBN) the long-running task, since handling the SIGINT by itself was enough to resolve the issue. So that's why I removed it.

(And good call about the stack issue, I didn't think of that.)

src/scripts/cronjobs/firstDataBrokerRemovalFixed.tsx
Vinnl: You mean across the different cronjobs? I think what specifically needs to be teared down differs per cron job, i.e. some will need to close down the email pool, but others will never send email. So then it's probably best to keep the cron job in control?

Vinnl:
I'm going to merge this since it's just after a production release; it can't hurt to let this cook on stage for a bit :)

Vinnl:
Oh crap, looks like forgot :facepalm:

PR #5717, opened 2025-03-20
Add support for SubPlat 3

flozia:
# References:

Jira: [MNTOR-4013](https://mozilla-hub.atlassian.net/browse/MNTOR-4013)

# Description

Adds support for the new SubPlat 3 URLs.

# How to test

- Enable the feature flag `SubPlat3`
- Subscription CTAs should point to the updated URLs

[MNTOR-4013]: https://mozilla-hub.atlassian.net/browse/MNTOR-4013?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/emails/templates/breachAlert/BreachAlertEmail.tsx
```diff
@@ -62,7 +62,10 @@ export const BreachAlertEmail = (props: BreachAlertEmailProps) => {
   }
 
   const premiumSubscriptionUrlObject = new URL(
-    getPremiumSubscriptionUrl({ type: "yearly" }),
+    getPremiumSubscriptionUrl({
+      type: "yearly",
+      enabledFeatureFlags: [],
```
flozia, line 67: Thinking about handling the updated URLs to support emails that have been sent out best. Most likely we’ll need to add a redirect to properly support toggling the flag between SubPlat 2 and SubPlat 3.

src/app/(proper_react)/(redesign)/(public)/LandingView.stories.tsx
```diff
@@ -15,6 +15,8 @@ const meta: Meta<typeof View> = {
   component: (props: ViewProps) => {
     const experimentData =
       props.experimentData ?? defaultExperimentData["Features"];
+    const { enabledFeatureFlags: enabledFeatureFlagsProp } = props;
+    const enabledFeatureFlags = enabledFeatureFlagsProp ?? [];
```
Vinnl, line 19: Heh, am I misreading this, or is this a somewhat convoluted way of doing

```suggestion
    const enabledFeatureFlags = props.enabledFeatureFlags ?? [];
```

? :sweat_smile:

src/emails/templates/breachAlert/BreachAlertEmail.tsx
Vinnl: Oh yeah that will be problematic - but the existing link is not to the Monitor site, right? So could we have FxA do the redirect? Or at least discuss the problem with them, see if they have a solution?

Vinnl: (I assume this is the reason you're not actually passing the feature flags to the email? It might be a good idea to do it anyway, or at least call it out in a comment, because future contributors could see the `enabledFeatureFlags` parameter and expect that they can use it.

src/app/functions/server/getPremiumSubscriptionInfo.test.ts
Vinnl, line 1: I will note that these tests depend on the correct values being set in the `.env`. I think they're being set by default in the main `.env`, and there's little reason to change them, so it's probably fine, but thought I'd call it out in case I overlooked a risk.

Vinnl (approved):
LGTM!

src/scripts/cronjobs/emailBreachAlerts.tsx
```diff
@@ -108,6 +109,7 @@ export async function poll(
   );
 
   const breaches = await getAllBreachesFromDb();
+  const subPlatFeatureFlag = await getFeatureFlagData("SubPlat3");
```
Vinnl, line 112: Any reason not to enable `getEnabledFeatureFlags`? We already changed the allowlist logic once, so there is some risk to not re-using the logic, and potentially future contributors will modify `sendEmail` and not realise that not all feature flags show up in that parameter?

Vinnl (approved):
And a commit added just before I submitted my review :)

src/scripts/cronjobs/emailBreachAlerts.tsx
flozia: We can also use `getEnabledFeatureFlags` here. I thought there might have been a reason to be more precise in getting the feature flag as it has been also used in [cronjobs/churnDiscount](https://github.com/mozilla/blurts-server/blob/main/src/scripts/cronjobs/churnDiscount.tsx#L34). Or was this pre-changing the allowlist logic?

src/app/(proper_react)/(redesign)/(public)/LandingView.stories.tsx
flozia: I just enjoy destructering too much and went overboard here. Updated in [0b448ba](https://github.com/mozilla/blurts-server/pull/5717/commits/0b448baa87b18d4a760fcf37d9731865b8313fc1).

src/scripts/cronjobs/emailBreachAlerts.tsx
Vinnl: Ah yes, there is indeed a reason, I should've added a comment - that way, I only needed a single query for all recipients.

Hmm... I imagine that might apply here too. It would probably be better to split up `getEnabledFeatureFlags` into a DB function that fetches all the feature flag data, and then another one that filters it down to a single user, but that's out of scope for this PR.

Then maybe at least put a comment above the `enabledFeatureFlags` to call out that this is a trap that future contributors might run into.

flozia: Yes, I was just about to update to using `getEnabledFeatureFlags` and this would need a call for each subscriber instead of one for the cronjob.

PR #5735, opened 2025-03-27
SVCSE-2596: Log how many emails we've sent at a timeout

Vinnl:
We don't know for sure how many emails we manage to send before our cronjobs get terminated; this adds logs so that we have a better picture.

# How to test

Run `kill -SIGTERM <process ID>`. I added a timeout myself so that I had enough time to run that.


rhelmer (approved):
lgtm, @Vinnl do we have an idea of the total emails to send? Is this already logged somewhere?

Vinnl:
@rhelmer We don't, we're [not running a query that loads the full list](https://github.com/mozilla/blurts-server/blob/6fb4b8461a0bd521b338284aada81ce2e26e0bee/src/db/tables/subscribers.ts#L488). However, it looks like the list is significantly smaller than what we thought for a while, so assuming this new logs shows a reasonably high number, we should be able to set the batch size just a bit below that and get through the full (smaller) list.

PR #5737, opened 2025-03-29
Update enabled locales (add Korean, Thai)

flodolo:
I received a message asking why Thai wasn't enabled, turns out the process is manual, and hasn't been checked in forever.

This PR adds Thai (94.8%) and Korean (100%), drops `cak` (39.76%), `nb-NO` (38.69%), and `ro` (36.85%).


flodolo:
Is there more that needs to be done here? Also, could we automate it somehow? [This](https://gist.github.com/flodolo/f41e0d7073df458a7fb30a854d0da126) is the script that I created eons ago.

flozia (approved):

Vinnl:
I'll merge this, but will note for posterity that this only affects local instances - work is in progress to also update the actual environment variables in production.

PR #5753, opened 2025-04-01
Add OpenTelemetry instrumentation

Vinnl:
# References:

Jira: MNTOR-4268
Figma:

# Description

I'm not actually sure if this is a good idea, because I have hardly any idea what I'm doing, let alone do I know how to view the traces locally. That said, it doesn't seem to break anything locally, so maybe we could test on stage? Or even better: maybe you do know what I'm doing, and have sensible input? :sweat_smile: 

I'm not sure if GCP will automatically pick this up, or if we also need [`@google-cloud/opentelemetry-cloud-trace-exporter`](https://www.npmjs.com/package/@google-cloud/opentelemetry-cloud-trace-exporter), but there has just been a major release of the OpenTelemetry SDK, and [that package isn't compatible yet](https://github.com/GoogleCloudPlatform/opentelemetry-operations-js/issues/790).


src/app/functions/server/logging.ts
Vinnl, line 1: Changes here based mostly on the docs in https://www.npmjs.com/package/@opentelemetry/instrumentation-winston

src/instrumentation.node.ts
```diff
@@ -0,0 +1,35 @@
...
+  });
+
+  // Initialize OpenTelemetry
+  const sdk = new NodeSDK({
+    resource: resourceFromAttributes({
+      [ATTR_SERVICE_NAME]: "firefoxmonitor",
+    }),
+    spanProcessor: new SimpleSpanProcessor(new OTLPTraceExporter()),
+  });
+  sdk.start();
```
Vinnl, line 34: These additions are based on these docs: https://nextjs.org/docs/app/building-your-application/optimizing/open-telemetry

src/instrumentation.node.ts
```diff
@@ -0,0 +1,35 @@
...
+      : 0.1,
+
+    // Setting this option to true will print useful information to the console while you're setting up Sentry.
+    debug: false,
+  });
+
+  // Initialize OpenTelemetry
+  const sdk = new NodeSDK({
+    resource: resourceFromAttributes({
+      [ATTR_SERVICE_NAME]: "firefoxmonitor",
```
Vinnl, line 30: I'm not sure if this service name needs to match any existing value, and if so, which? And if that value differs per environment, how do I get access to it?

rhelmer:
@kschelonka do you mind taking a look at this? We are looking into enabling log tracing in GCP using opentelemetry in the NextJS app.

kschelonka:
yay otel! yes, you will need a collector -- there are 3 components (receiver, processor, exporter). I am looking into getting a collector set up and forwarding to managed prometheus in the gcpv2 environment. Will look more closely at the instrumentation setup in a bit

Vinnl:
Closing since there seems to be no availability for getting this over the finish line at the moment. We can re-open later.

PR #5754, opened 2025-04-01
MNTOR-4084: Update auth with Moscary

mansaj:
# References:

Jira: MNTOR-4084

# Description

Update auth to using Moscary auth. 

# How to test
* feature flag has to be enabled to for the flip to happen
* base url needs to be updated
* new environment var for the JWT token
* PM me for clarification if needed


rhelmer (changes requested):
I don't have any huge concerns with this approach overall, but it should have a test.

Have you chatted w/ fxa team about this? They might be able to provide someone to review, if they don't have a better alternative.

.env.local.example
```diff
@@ -41,6 +41,9 @@ APP_ENV=local
 SERVER_URL=http://localhost:6060
 PORT=6060
 NEXTAUTH_URL=http://localhost:6060
+MOSCARY_URL=http://localhost:3000
+MOSCARY_JWT_SECRET=2ba0b54934944fb1a200b11ab26ba4b869e331efb25c486977b558bc27443b7c
```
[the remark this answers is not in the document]
mansaj: Hmm I thought we usually use a burner key for these. But I can also just use a string like `unsafe-secret-for-dev`

Should I also change the other ones (or at least prefix it)

mansaj:
> I don't have any huge concerns with this approach overall, but it should have a test.
> 
> Have you chatted w/ fxa team about this? They might be able to provide someone to review, if they don't have a better alternative.

not possible via fxa at the moment

mansaj:
Updated the PR to use priv/pub key pair

src/app/functions/server/generateJWT.ts
```diff
@@ -0,0 +1,52 @@
...
+import { redisClient, REDIS_JWT_KEY_PREFIX } from "../../../db/redis/client";
+import { isAdmin } from "../../api/utils/auth";
+import { logger } from "./logging";
+const rClient = redisClient();
+
+// We'll need a key ID to identify the key used for signing
+const KEY_ID = process.env.MOSCARY_JWT_KEY_ID || "moscary-jwt-key";
+
+export async function getMoscaryJWT(fxaUid: string, email: string) {
+  const REDIS_KEY = `${REDIS_JWT_KEY_PREFIX}${fxaUid}`;
```
kschelonka, line 15: Recommend using session identifier here since otherwise there's no way to invalidate the cache before TTL ends. For example if we make someone an admin, they can log out and log in to get access to their new role immediately versus having to wait for the timeout period (or otherwise implementing cache invalidation)

src/app/functions/server/generateJWT.ts
```diff
@@ -0,0 +1,52 @@
...
+  }
+  token = jwt.sign(
+    {
+      sub: fxaUid,
+      email: email,
+      isAdmin: isAdmin(email),
+    },
+    privateKey, // Use the private key here
+    {
+      expiresIn: "1h",
```
kschelonka, line 40: nit: you can use a numeric const here (3600 works also) so that it can be reused in line 49

src/app/.well-known/jwks.json/route.ts
```diff
@@ -0,0 +1,64 @@
...
+    logger.error("jwks_endpoint_error", {
+      message: "MOSCARY_JWT_PUBLIC_KEY is not set",
+    });
+    return NextResponse.json(
+      { error: "Server configuration error" },
+      { status: 500 },
+    );
+  }
+
+  const jwk = pemToJwk(publicKeyPem, keyId);
```
kschelonka, line 44: if the pemfile itself is coming from an env var, then there's no need to ever call this method after startup (since env vars require redeploy to change). recommend saving in memory

src/app/functions/server/generateJWT.ts
mansaj: Good catch! Fixed, thanks

src/app/.well-known/jwks.json/route.ts
mansaj: Good catch! Fixed, thanks

src/app/functions/server/generateJWT.ts
```diff
@@ -0,0 +1,57 @@
...
+export const EXPIRATION_TIME = 3600;
+
+export async function getMoscaryJWT(
+  fxaUid: string,
+  email: string,
+  sessionId: string,
+) {
+  const REDIS_KEY = `${REDIS_JWT_KEY_PREFIX}${fxaUid}:${sessionId}`;
+
+  // if token is already exists in redis, return it
```
rhelmer, line 22:
```suggestion
  // if token already exists in redis, return it
```

rhelmer (approved):

PR #5756, opened 2025-04-02
Add setting views to Storybook and mock Server Actions

flozia:
# Description

For the [settings redesign](https://www.figma.com/design/2iCgADpFXKOFZTEgCPkQsR/Settings?node-id=597-10052&t=c2UkGyKPdbvneuz5-4), we are about to introduce more permutations that would be useful for reviewing and testing in Storybook. This PR is only adding the default views to focus on resolving an issue with adding components and views that rely on Server Actions:

With the Server Actions we use for the setting views, Storybook currently breaks when importing Actions that have node dependencies, and we need a way to mock those. To mock modules [Storybook suggests](https://storybook.js.org/docs/writing-stories/mocking-data-and-modules/mocking-modules) to use Node.js native [subpath imports](https://nodejs.org/api/packages.html#subpath-imports).

### DIFF
--- a/.storybook/main.ts
+++ b/.storybook/main.ts
@@ -37,6 +37,17 @@ const config: StorybookConfig = {
       type: "asset/source",
     });
 
+    // Defining `paths` in tsconfig.json breaks Storybook’s path resolution.
+    // Removing `TsconfigPathsPlugin` ensures that Storybook uses the default
+    // module resolution behavior.
+    config.resolve = {
+      ...config.resolve,
+      plugins:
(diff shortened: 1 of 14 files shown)


Vinnl (approved):
I'm approving this, but have also proposed https://github.com/mozilla/blurts-server/pull/5762 as an alternative. I'll leave it up to you to decide which to go with (unless you don't want that responsibility) :)

flozia:
Closing in favour of https://github.com/mozilla/blurts-server/pull/5762#pullrequestreview-2783250689.

PR #5762, opened 2025-04-07
Add setting views to Storybook and mock Server Actions using props

Vinnl:
This builds on #5756, but uses props to pass server actions into our settings components, rather than mocking the module using Node's subpath imports. This is just so we can compare the two approaches.

I reused everything from the `stories` folder for the settings page, except moving the mocks from the mock file to the stories wrapper.


flozia (approved):
Thank you for updating the Server Actions to use props @Vinnl. I do like the idea to be able to import actions in directly from client components, but as discussed I’m also worried about path resolutions and its implications we might run into with using subpath imports as proposed in https://github.com/mozilla/blurts-server/pull/5756.

Let’s move forward with this approach.

PR #5701, opened 2025-03-13
Rename notifications to announcements

codemist:
# References:

Jira: MNTOR-
Figma:

# Description

# Screenshot (if applicable)

Not applicable.

# How to test

# Checklist (Definition of Done)

- [ ] Localization strings (if needed) have been added.
- [ ] Commits in this PR are minimal and [have descriptive commit messages](https://chris.beams.io/posts/git-commit/).
- [ ] I've added or updated the relevant sections in readme and/or code comments
- [ ] I've added a unit test to test for potential regressions of this bug.
- [ ] If this PR implements a feature flag or experimentation, I've checked that it still works with the flag both on, and with the flag off.
- [ ] If this PR implements a feature flag or experimentation, the Ship Behind Feature Flag status in Jira has been set
- [ ] Product Owner accepted the User Story (demo of functionality completed) or waived the privilege.
- [ ] All acceptance criteria are met.
- [ ] Jira ticket has been updated (if needed) to match changes made during the development process.
- [ ] Jira ticket has been updated (if needed) with suggestions for QA when this PR is deployed to stage.

PR #5783, opened 2025-04-08
Fix sidebar notifications link

flozia:
# References:

PR #5701

# Description

In PR #5701, the “Set notifications” sidebar link was mistakenly updated.


Vinnl (approved):
Oh heh, and these were referring to _email_ notifications, not announcements, which was the precise reason for the rename 😅

PR #5809, opened 2025-04-15
Move getSha1 to actions

flozia:
# References:

Jira: [MNTOR-4297](https://mozilla-hub.atlassian.net/browse/MNTOR-4297)

# Description

Move `getSha1` to the settings actions.

### DIFF
--- a/src/app/(proper_react)/(redesign)/(authenticated)/admin/dev/actions.tsx
+++ b/src/app/(proper_react)/(redesign)/(authenticated)/admin/dev/actions.tsx
@@ -114,3 +114,13 @@ export async function triggerManualProfileScan(onerepProfileId: number) {
     console.error("Manual scan triggered by admin failed:", error);
   }
 }
+
+export async function getSha1(source: string): Promise<string> {
+  const msgUint8 = new TextEncoder().encode(source);
+  const hashBuffer = await crypto.subtle.digest("SHA-1", msgUint8);
+  const hashArray = Array.from(new Uint8Array(hashBuffer));
+  const hashHex = hashArray
+    .map((b) => b.toString(16).padStart(2, "0"))
+    .join("");
+  return hashHex;
+}
--- a/src/app/(proper_react)/(redesign)/(authenticated)/admin/dev/UserAdmin.tsx
+++ b/src/app/(proper_react)/(redesign)/(authenticated)/admin/dev/UserAdmin.tsx
@@ -19,6 +19,7 @@ import {
   updateOnerepProfile,
   triggerManualProfileScan,
   getAllProfileScans,
+  getSha1,
 } from "./actions";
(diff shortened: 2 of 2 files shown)


Vinnl (approved):
No objection to this change, but I'm wondering why you needed to make it? Anything we can learn? :)

flozia:
I wrongly suspected the function caused an error on `stage` – the frontend logging was a bit cryptic. I’ve updated the PR to update the guarding conditions so they also correctly work in the `stage` environment.

src/app/(proper_react)/(redesign)/(authenticated)/admin/dev/actions.tsx
```diff
@@ -79,9 +79,7 @@ export async function getAllProfileScans(onerepProfileId: number) {
   if (
     !session?.user?.email ||
     !isAdmin(session.user.email) ||
-    // only allow admins to trigger a scan for their own profile in production
-    (process.env.APP_ENV === "production" &&
-      session.user.subscriber?.onerep_profile_id !== onerepProfileId)
```
Vinnl, line 84: Hmm, this seems unrelated to the description of the PR? Do we want admins to be able to trigger scans for _anyone_ on _prod_?

src/app/(proper_react)/(redesign)/(authenticated)/admin/dev/page.tsx
```diff
@@ -10,11 +10,7 @@ import { UserAdmin } from "./UserAdmin";
 
 export default async function DevPage() {
   const session = await getServerSession();
-  if (
-    !session?.user?.email ||
-    !isAdmin(session.user.email) ||
-    process.env.APP_ENV === "production"
```
Vinnl, line 16: This also doesn't seem limited to stage, and would expose this page in prod?

src/app/(proper_react)/(redesign)/(authenticated)/admin/dev/actions.tsx
flozia: I’ve made this condition stricter so that `getAllProfileScans` always returns a `404` in `prod` on the line below.

src/app/(proper_react)/(redesign)/(authenticated)/admin/dev/page.tsx
flozia: All the actions are guarded and return a `404` when `process.env.APP_ENV === "production"`, except `triggerManualProfileScan` (callable for admin accounts only on production). We’ll need a way to trigger manual scans. I can keep the condition here for now (readded in [939fd31](https://github.com/mozilla/blurts-server/pull/5809/commits/939fd31f8d138bf0976a7c718a273fbdc5cd4fe9)) as a redundant safe-guard and add a new view with just a button for that (will do that in a follow-up after this has been tested on `stage`).

PR #5850, opened 2025-04-30
Add admin panel for triggering manual scans on prod

flozia:
# References:

Jira: [MNTOR-4297](https://mozilla-hub.atlassian.net/browse/MNTOR-4297)

# Description

To verify that [MNTOR-3567](https://mozilla-hub.atlassian.net/browse/MNTOR-3567) is working as intended without us having to wait a month, we need a way to trigger manual scans on production. This PR adds a boiled-down version of the production admin panel that only lists profile scans and lets them trigger a manual scan. As an additional safeguard, admins can only look up users with primary `@mozilla.com` emails on production.

# How to test

- Be authenticated as an admin
- Visit `/admin/prod`

[MNTOR-4297]: https://mozilla-hub.atlassian.net/browse/MNTOR-4297?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[MNTOR-3567]: https://mozilla-hub.atlassian.net/browse/MNTOR-3567?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/app/(proper_react)/(redesign)/(authenticated)/admin/dev/UserAdminProduction.tsx
+++ b/src/app/(proper_react)/(redesign)/(authenticated)/admin/dev/UserAdminProduction.tsx
@@ -1,0 +1,135 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+"use client";
+
+import { useState } from "react";
+import { useSession } from "next-auth/react";
+import styles from "./UserAdmin.module.scss";
+import { Button } from "../../../../../components/client/Button";
+import {
+  lookupFxaUid,
+  triggerManualProfileScan,
+  getAllProfileScans,
+} from "./actions";
+import { OnerepScanRow } from "knex/types/tables";
+import { InputField } from "../../../../../components/client/InputField";
+import { FeatureFlagName } from "../../../../../../db/tables/featureFlags";
+import { DataTable, getSha1 } from "./UserAdmin";
+
+export const UserAdminProduction = ({
+  enabledFeatureFlags,
+}: {
+  enabledFeatureFlags: FeatureFlagName[];
+}) => {
+  const session = useSession();
+  const [isLoading, setIsLoading] = useState(false);
+  const [emailInput, setEmailInput] = useState("");
+  const [status, setStatus] = useState<null | string>(null);
+  const [onerepProfileId, setOnerepProfileId] = useState(null);
+  const [oneRepProfileScans, setOneRepProfileScans] = useState<
+    OnerepScanRow[] | null
+  >(null);
+
+  const setProfileScans = async (onerepProfileId: number) => {
+    const profileScans = await getAllProfileScans(onerepProfileId);
+    if (profileScans) {
+      setOneRepProfileScans(
+        profileScans.sort(
+          (a, b) =>
+            new Date(a.created_at).getTime() - new Date(b.created_at).getTime(),
+        ),
+      );
+    }
+  };
+
+  const triggerScanAction = async () => {
+    if (!enabledFeatureFlags.includes("EditScanProfileDetails")) {
+      return;
+    }
+
+    try {
+      if (onerepProfileId) {
+        await triggerManualProfileScan(onerepProfileId);
+        setProfileScans(onerepProfileId);
+        setStatus(`Running manual scan for [${emailInput}] succeeded.`);
+      }
+    } catch (error) {
+      setStatus(`[Running manual scan failed: [${error}].`);
+    }
+  };
+
+  return (
+    <main className={styles.wrapper}>
+      <header className={styles.header}>
+        Logged in as <b>{session.data?.user.email}</b>
+      </header>
+      <form
+        className={styles.form}
+        onSubmit={(event) => {
+          event.preventDefault();
+
+          void getSha1(emailInput)
+            .then(async (emailHash) => {
+              setIsLoading(true);
+              const fxaUid = await lookupFxaUid(emailHash);
+              const response = await fetch(`/api/v1/admin/users/${fxaUid}`);
+              if (!response.ok) {
+                setIsLoading(false);
+                return;
+              }
+              const data = await response.json();
+              setOnerepProfileId(data.onerepProfileId);
+              setProfileScans(data.onerepProfileId);
+
+              setIsLoading(false);
+            })
+            .catch(() => {
+              setOnerepProfileId(null);
+              setOneRepProfileScans(null);
+            })
+            .finally(() => {
+              setIsLoading(false);
+            });
+        }}
+      >
+        <div className={styles.userPicker}>
+          <InputField
+            type="email"
+            value={emailInput}
+            onChange={(email) => {
+              setOnerepProfileId(null);
+              setOneRepProfileScans(null);
+              setEmailInput(email);
+            }}
+            placeholder="Subscriber email"
+            label={`Find user by email address ${isLoading ? "…" : ""}`}
+            description="Works for @mozilla.com emails only"
+          />
+        </div>
+        <Button variant="primary" type="submit">
+          Lookup email
+        </Button>
+      </form>
+      {status && <p className={styles.status}>{status}</p>}
+      <section>
+        <h2>OneRep profile</h2>
+        {oneRepProfileScans ? (
+          <>
+            <DataTable
+              header="Current profile scans"
+              data={oneRepProfileScans}
+              open
+            />
+            <Button variant="primary" onPress={() => void triggerScanAction()}>
+              Trigger manual scan
+            </Button>
+          </>
+        ) : (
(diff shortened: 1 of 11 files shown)


src/app/api/v1/admin/users/[fxaUid]/route.ts
```diff
@@ -63,7 +61,11 @@ export async function GET(
 
       const fxaUid = params.fxaUid;
       const subscriber = await getSubscriberByFxaUid(fxaUid);
-      if (!subscriber) {
+      if (
+        !subscriber ||
+        (process.env.APP_ENV === "production" &&
+          !isMozMail(subscriber.primary_email))
```
Comment (Vinnl) [inline src/app/api/v1/admin/users/[fxaUid]/route.ts:67]: I think I'd separate this out into a separate `if` clause (that returns a 403 Forbidden, if you want to be pure about it), just to reduce the complexity of the disjunction, to make sure we're absolutely explicit about who's allowed to call this when, and to avoid us making mistakes about that in the future.

(Oh, and maybe instead of comparing to `"production"`, check if it's not `"stage"` or `"local"`, i.e. such that it fails if it's not set or set to an invalid value.)

src/app/(proper_react)/(redesign)/(authenticated)/admin/dev/actions.tsx
```diff
@@ -17,18 +17,21 @@ import {
...
-    !isAdmin(session.user.email) ||
-    process.env.APP_ENV === "production"
-  ) {
+  if (!session?.user?.email || !isAdmin(session.user.email)) {
     return notFound();
   }
 
   const subscriber = await getSubscribersByHashes([emailHash]);
+  if (
+    process.env.APP_ENV === "production" &&
```
Vinnl, line 30: As suggested elsewhere:

```suggestion
    // On production, only allow looking up Mozilla email addresses
    process.env.APP_ENV !== "stage" && process.env.APP_ENV !== "local" &&
```

src/app/(proper_react)/(redesign)/(authenticated)/admin/dev/actions.tsx
```diff
@@ -76,11 +79,7 @@ export async function updateOnerepProfile(
 
 export async function getAllProfileScans(onerepProfileId: number) {
   const session = await getServerSession();
-  if (
-    !session?.user?.email ||
-    !isAdmin(session.user.email) ||
-    process.env.APP_ENV === "production"
-  ) {
+  if (!session?.user?.email || !isAdmin(session.user.email)) {
```
Vinnl, line 82: Since this potentially exposes PII as well (I think?), maybe we should do a check that this profile ID is associated with a Mozilla email address as well?

src/app/(proper_react)/(redesign)/(authenticated)/admin/dev/actions.tsx
```diff
@@ -93,13 +92,7 @@ export async function getAllProfileScans(onerepProfileId: number) {
...
 export async function triggerManualProfileScan(onerepProfileId: number) {
   const session = await getServerSession();
-  if (
-    !session?.user?.email ||
-    !isAdmin(session.user.email) ||
-    // only allow admins to trigger a scan for their own profile in production
-    (process.env.APP_ENV === "production" &&
-      session.user.subscriber?.onerep_profile_id !== onerepProfileId)
-  ) {
+  if (!session?.user?.email || !isAdmin(session.user.email)) {
```
Vinnl, line 95: And maybe here a check on email address as well? Although I think this one maybe doesn't expose PII and can be helpful for customer support?

src/app/functions/universal/isMozMail.ts
```diff
@@ -0,0 +1,8 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+export const isMozMail = (email: string) => {
+  const emailParsed = email.trim().toLowerCase();
+  return emailParsed.endsWith("@mozilla.com");
```
Vinnl, line 7: In this case we probably _want_ this restriction, but might be good to point out in a comment perhaps: not all colleagues have an `@mozilla.com` email address (e.g. there's also Pocket and MoFo addresses). None of those need the access granted here, but maybe someone makes assumptions in the future based on the function name.

src/app/(proper_react)/(redesign)/(authenticated)/admin/dev/UserAdminProduction.tsx
```diff
@@ -0,0 +1,135 @@
...
+              const response = await fetch(`/api/v1/admin/users/${fxaUid}`);
+              if (!response.ok) {
+                setIsLoading(false);
+                return;
+              }
+              const data = await response.json();
+              setOnerepProfileId(data.onerepProfileId);
+              setProfileScans(data.onerepProfileId);
+
+              setIsLoading(false);
```
Vinnl, line 86: Nit: Redundant with `.finally`?

```suggestion
```

Vinnl (approved):
Looks good, suggested some simple changes, and as mentioned on Slack, I'd like a separate route for the prod version. (I'm just really don't want us to accidentally overexpose this :) )

src/app/functions/universal/isMozMail.ts
flozia: Added a comment in [8c1f676](https://github.com/mozilla/blurts-server/pull/5850/commits/8c1f6764e2088fc88472874714c23d59a9870ed4).

src/app/api/v1/admin/users/[fxaUid]/route.ts
Comment (flozia) [reply src/app/api/v1/admin/users/[fxaUid]/route.ts]: Good idea. I’m explicitly returning a `403` in https://github.com/mozilla/blurts-server/commit/0ac6234b96f03678a0751e6bfe9c6e3656deeb63.

src/app/(proper_react)/(redesign)/(authenticated)/admin/dev/actions.tsx
flozia: Update in [7b598da](https://github.com/mozilla/blurts-server/pull/5850/commits/7b598dab3a98bfcab7e7fe0d43a4186da03ede98).

src/app/(proper_react)/(redesign)/(authenticated)/admin/dev/actions.tsx
flozia: I was thinking about this as well. Right now, we are not calling this action without checking the email address of the subscriber which is checked before. I’ve added a check to the function directly as an additional safeguard in [7b598da](https://github.com/mozilla/blurts-server/pull/5850/commits/7b598dab3a98bfcab7e7fe0d43a4186da03ede98).

src/app/(proper_react)/(redesign)/(authenticated)/admin/dev/actions.tsx
flozia: I’ve added the additional check for now. See https://github.com/mozilla/blurts-server/pull/5850#discussion_r2068606190.

flozia (commented):
Thanks @Vinnl. I’ve broken the panel out to `/admin/prod` and added/updated some additional conditions.

src/db/tables/subscribers.ts
```diff
@@ -55,6 +55,22 @@ async function getSubscriberByFxaUid(
...
+): Promise<undefined | (SubscriberRow & WithEmailAddresses)> {
+  const [subscriber] = await knex("subscribers").where({
+    onerep_profile_id,
+  });
+  if (!subscriber) {
+    return;
+  }
+  const subscriberAndEmails = await joinEmailAddressesToSubscriber(subscriber);
+  return subscriberAndEmails;
+}
```
Vinnl, line 71: I don't think we need the associated email addresses, right? And outside of the admin use case, we probably don't want to use this function.

```suggestion
async function getSubscriberByOnerepProfileId(
  onerep_profile_id: SubscriberRow["onerep_profile_id"],
): Promise<undefined | SubscriberRow> {
  const subscriber = await knex("subscribers").first().where({
    onerep_profile_id,
  });
  return subscriber;
}
```

Vinnl (approved):
Thanks for making the updates!

src/app/(proper_react)/(redesign)/(authenticated)/admin/dev/actions.tsx
Vinnl: Right, but keep in mind that "calling the action" is an API call that the user can simulate without doing the email check beforehand.

src/db/tables/subscribers.ts
flozia: That’s right: [7c425dc](https://github.com/mozilla/blurts-server/pull/5850/commits/7c425dcc9eff9b318062e399944d896e38d948c4).

PR #5854, opened 2025-05-01
MNTOR-4365, MNTOR-4389 - Pass user session object into announcements init function to use checkUserHasMonthlySubscription()

codemist:
Reverts mozilla/blurts-server#5847

Fixes https://mozilla-hub.atlassian.net/browse/MNTOR-4389 

I made these changes to improve the resilience of the `initializeUserAnnouncements` function, particularly for new or incomplete user accounts. In theory, the dashboard was breaking because the code assumed that certain `user.subscriber` fields were always present. However, for brand-new users or edge cases, these fields can be `null` or undefined, causing the logic to fail or skip critical checks. Without proper handling, the frontend tried to load announcement data before the backend had finished inserting the necessary rows.

To fix this, I added early defensive checks (redirecting if the subscriber object is missing), and safely handled the audience filtering logic.


rhelmer (changes requested):
Thanks for the detailed description and inline review comments! Makes sense.

If I understand this correctly, this is a race condition between the user session being created and the `subscriber` row in the database being available?

One suggestion I'd make is to log when we encounter a session that has no `subscriber` row, so we have observe this in stage/prod and see how prevalent it is.

As for detecting this kind of problem earlier (e.g. at build/test time), I think the e2e test could likely flush out these kinds of race conditions by injecting randomness (chaos/fuzz testing), but that's a pretty expensive and slow approach, and it will be relatively difficult to decipher the output of such tests and figure out exactly which component failed. It's worth seriously considering, but I think for this particular problem we should have a smaller target.

If we can figure out which component(s) are the problem here, we might be able to reproduce it with a unit or integration test that's relatively quick to run, and will fail in a more obvious way. The first step is that we need to reproduce this behavior, I'm assuming we don't have reliable steps to reproduce right now?

In any case please add logging and file a follow-up on this underlying issue, if we're observing cases where we have a user session but no subscriber row, this is almost certainly a problem in some lower-level component that we might redesign and/or add more tests around.

rhelmer (approved):

PR #5864, opened 2025-05-05
fix: Force UTC timezone for DOB on /user/settings/edit-profile screen

rhelmer:
# Description

Small followup for https://github.com/mozilla/blurts-server/pull/5859 — the date of birth is stored as UTC but it's displayed on the `edit-profile` screen in the local timezone, which can cause it to look off by one day.

The previous PR fixed this for the `edit-info` but not the `edit-profile` screen. It's the same one-line fix to the client JS otherwise.

# How to test

Navigate to "Update scan info" on sidebar, check DoB, then press "Add details" button and check DoB. I'm seeing different values when my system is in PST.

Changing system to UTC and reloading the page causes the problem to go away.


EMMLynch (approved):
Low risk change that matches the implementation on Friday in the other section of code.

PR #5885, opened 2025-05-15
Make explicit that this isn't a package to install

Vinnl:
Recreated from https://github.com/mozilla/blurts-server/pull/5882 because the preview deploy was failed and couldn't be recreated.


Vinnl:
Merging since it was approved in https://github.com/mozilla/blurts-server/pull/5882.

PR #5940, opened 2025-05-28
Highlight feature flags that are active in prod

Vinnl:
This will make it easier to identify (local/stage) flag settings that will cause discrepancies with production.

image

Flags with a purple outline are also active on prod.

(I quickly whipped this up to check something locally, figured we could use this in stage as well.)


flozia (approved):
That’s a nice addition to the feature flags!

PR #5953, opened 2025-06-03
Add feature flag disable landing page redirect

flozia:
# Description

To test the bundle features on the landing page, we need a way to disable the redirect to the dashboard for authenticated users. This also fixes the feature flag allow list for `PrivacyProductsBundle` on the bundle onboarding page.

# How to test

**Landing page redirect**
- Login
- Toggle and/or add email to feature flag `DisableLandingToDashboardRedirect`
- Visit the landing page

**Bundle onboarding**
- Navigate to `/bundle/onboarding`
- Toggle email from `PrivacyProductsBundle`
- Revisit `/bundle/onboarding`


src/app/(bundle)/bundle/onboarding/page.tsx
```diff
@@ -12,10 +12,14 @@ import {
...
+import { getServerSession } from "../../../functions/server/getServerSession";
 
 export default async function Page() {
+  const session = await getServerSession();
+  const isAuthenticated = typeof session?.user.subscriber?.fxa_uid === "string";
   const enabledFeatureFlags = await getEnabledFeatureFlags({
-    isSignedOut: true,
+    isSignedOut: !isAuthenticated,
+    email: session?.user?.email ?? "",
   });
```
Vinnl: This relies on the email address `""` not being on the allowlist, which happens to work, but since it doesn't exactly codify the intention, I'm still wary of doing this - just in case someone breaks that assumption in the future. There's an easy safer way, I think:

```suggestion
  const enabledFeatureFlags = await getEnabledFeatureFlags(
    typeof session?.user.subscriber?.fxa_uid === "string" && typeof session.user.subscriber.email === "string"
      ? {
        isSignedOut: false,
        email: session.user.email,
      }
      : { isSignedOut: true },
  );
```

Vinnl (commented):
Just a small suggestion - let me know if that doesn't work.

src/app/(bundle)/bundle/onboarding/page.tsx
flozia: That’s a good point. Being more intentional about the condition is probably a good idea: Updated in https://github.com/mozilla/blurts-server/pull/5953/commits/e197258beddb75f1e0ebea0ae71db1c0b885e6c6.

Vinnl (approved):
Thanks for the changes!

PR #5961, opened 2025-06-04
MNTOR-4437 - Add support for monthly, yearly, and bundle user segmentation in announcements

codemist:
# References:

Jira: MNTOR-4437

# Description

This PR deprecates `checkUserHasMonthlySubscription` in favor of the more flexible `getUserSubscriptionType`, which returns "monthly", "yearly", or "bundle" based on the user’s current active plan. The `getBillingAndSubscriptions()` function only ever returns a single active subscription with one price ID, plans are mutually exclusive, so users upgrading (e.g., from yearly to bundle) will only reflect the latest plan. This simplifies audience targeting logic and avoids cumulative checks across multiple plan types.

I also added support for selecting `"bundle_user"` and` "premium_non_bundle"` in the admin announcements panel.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -77,4 +77,4 @@ updates:
         # to indicate any matching version until the next major one..
         # See https://docs.github.com/en/code-security/dependabot/working-with-dependabot/dependabot-options-reference#versions-ignore
         # and https://guides.rubygems.org/patterns/#pessimistic-version-constraint
-        versions: ["~> 23", "~> 25", "~> 27", "~> 29"]
+        versions: ["~> 23", "<= 24.10", "~> 25", "~> 27", "~> 29"]
--- a/src/app/(bundle)/bundle/onboarding/BundleOnboardingView.test.tsx
+++ b/src/app/(bundle)/bundle/onboarding/BundleOnboardingView.test.tsx
@@ -13,7 +13,6 @@ import HeaderMeta, { BundleHeaderDefault } from "./BundleHeader.stories";
 import {
   bundleQueryParamsExternalProducts,
   bundleQueryParamsMonitor,
-  promptNoneAuthParams,
 } from "./BundleOnboardingView";
 
 jest.mock("../../../hooks/useTelemetry");
@@ -70,10 +69,7 @@ it("opens a new tab when cta is selected", async () => {
   render(<ComposedBundlePage />);
 
   const monitorLink = screen.getByRole("link", { name: "Go to ⁨Monitor⁩" });
-  expect(monitorLink).toHaveAttribute(
-    "href",
-    `/?prompt=none&${bundleQueryParamsMonitor}`,
-  );
+  expect(monitorLink).toHaveAttribute("href", `/?${bundleQueryParamsMonitor}`);
 
   await user.click(monitorLink);
 
@@ -101,9 +97,7 @@ it("opens a new tab when cta is selected", async () => {
   const relayHref =
     process.env.FIREFOX_RELAY_LANDING_URL +
     "/accounts/fxa/login?process=login&" +
-    bundleQueryParamsExternalProducts +
-    "&auth_params=" +
-    encodeURIComponent(promptNoneAuthParams);
+    bundleQueryParamsExternalProducts;
 
   expect(relayLink).toHaveAttribute("href", relayHref);
 
--- a/src/app/(bundle)/bundle/onboarding/BundleOnboardingView.tsx
+++ b/src/app/(bundle)/bundle/onboarding/BundleOnboardingView.tsx
@@ -29,19 +29,13 @@ export const bundleQueryParamsMonitor = new URLSearchParams({
   utm_content: "launch-us",
 }).toString();
 
-export const promptNoneAuthParams = new URLSearchParams({
-  prompt: "none",
-}).toString();
-
 export const BundleOnboardingView = (props: Props) => {
   const l10n = props.l10n;
-  const monitorLink = `/?${promptNoneAuthParams}&${bundleQueryParamsMonitor}`;
+  const monitorLink = `/?${bundleQueryParamsMonitor}`;
   const relayLink =
     process.env.FIREFOX_RELAY_LANDING_URL +
     "/accounts/fxa/login?process=login&" +
-    bundleQueryParamsExternalProducts +
-    "&auth_params=" +
-    encodeURIComponent(promptNoneAuthParams);
+    bundleQueryParamsExternalProducts;
   const vpnLink =
     process.env.MOZILLA_VPN_LANDING_URL +
     "/download?" +
--- a/src/app/(proper_react)/(redesign)/(public)/LandingViewRedesign/LandingViewRedesign.test.tsx
+++ b/src/app/(proper_react)/(redesign)/(public)/LandingViewRedesign/LandingViewRedesign.test.tsx
@@ -780,6 +780,20 @@ describe("Privacy product bundle banner", () => {
       }),
     );
   });
+
+  it("hides the banner when the feature flag `DisableOneRepScans` is enabled", async () => {
+    const ComposedDashboard = composeStory(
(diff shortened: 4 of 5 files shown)


src/app/functions/server/user.ts
```diff
@@ -7,45 +7,47 @@ import { getBillingAndSubscriptions } from "../../../utils/fxa";
...
-    return false;
+    return;
   }
 
-  if (!process.env.PREMIUM_PLAN_ID_MONTHLY_US) {
-    console.error("Monthly Plan ID not set");
-    return false;
+  const subscriber = await getSubscriberByFxaUid(fxaUid);
+  if (!subscriber?.fxa_access_token) {
+    console.error("FXA access token not set");
```
flozia: Suggestion: Let’s use our `logger` from `/functions/server/logging` here.

src/app/functions/server/user.ts
```diff
@@ -7,45 +7,47 @@ import { getBillingAndSubscriptions } from "../../../utils/fxa";
...
   }
 
   const billingAndSubscriptionInfo = await getBillingAndSubscriptions(
     subscriber.fxa_access_token,
   );
-
-  if (billingAndSubscriptionInfo === null) {
-    return false;
+  if (!billingAndSubscriptionInfo) {
+    console.error("Billing and subscription info is null");
```
flozia: Suggestion: We should use the `logger` here as well.

src/app/functions/server/user.ts
```diff
@@ -7,45 +7,47 @@ import { getBillingAndSubscriptions } from "../../../utils/fxa";
...
 
-  const monthlyPlanId = process.env.PREMIUM_PLAN_ID_MONTHLY_US;
+  const monthlyPlanId = process.env.PREMIUM_PLAN_ID_MONTHLY_US ?? "";
   const yearlyPlanId = process.env.PREMIUM_PLAN_ID_YEARLY_US ?? "";
+  const bundlePlanId = process.env.SUBPLAT_BUNDLE_PRICE_ID ?? "";
 
-  const subscriptions = billingAndSubscriptionInfo.subscriptions;
+  if (!monthlyPlanId || !yearlyPlanId || !bundlePlanId) {
+    console.error("One or more plan IDs are not set in environment variables");
+  }
```
flozia: We would likely notice this happening in other places as well, but only logging here, it would also be worthwhile to throw a Sentry error.

codemist: Fixed in https://github.com/mozilla/blurts-server/pull/5961/commits/737ae5751d4fbef5629365a98f065ed21e989dd5

src/app/functions/server/user.ts
codemist: Fixed in https://github.com/mozilla/blurts-server/pull/5961/commits/737ae5751d4fbef5629365a98f065ed21e989dd5

src/app/functions/server/user.ts
codemist: Fixed in https://github.com/mozilla/blurts-server/pull/5961/commits/737ae5751d4fbef5629365a98f065ed21e989dd5

src/app/functions/server/user.ts
```diff
@@ -5,47 +5,50 @@
...
+    (sub) => sub.plan_id,
+  );
 
-  if (planIds.includes(yearlyPlanId)) {
-    console.error("User has yearly plan set");
-    return false;
-  }
+  if (planIds.includes(yearlyPlanId)) return "yearly";
+  if (planIds.includes(bundlePlanId)) return "bundle";
+  if (planIds.includes(monthlyPlanId)) return "monthly";
```
rhelmer, line 51: Is it possible that users could be enrolled in multiple plans? If so, and if `bundle` should be the default, that one should be first so it's the default.

src/app/functions/server/user.ts
```diff
@@ -5,47 +5,50 @@
...
 import { getBillingAndSubscriptions } from "../../../utils/fxa";
 import { getSubscriberByFxaUid } from "../../../db/tables/subscribers";
+import { logger } from "./logging";
 
 /* c8 ignore start */
-export async function checkUserHasMonthlySubscription(user: Session["user"]) {
-  if (!user.subscriber?.fxa_uid) {
+export type SubscriptionType = "yearly" | "monthly" | "bundle";
+
+export async function getUserSubscriptionType(
```
rhelmer, line 13: This used to return a boolean but may now return `undefined`, will callers handle that OK? Does that just mean the user has no subscription? I think that's fine if so, if you want to be explicit you could have a `SubscriptionType` of `none` but I don't think that's critical (it's pretty clear that `undefined` would mean no subscription IMHO)

src/app/functions/server/user.ts
```diff
@@ -5,47 +5,50 @@
...
   }
 
-  const monthlyPlanId = process.env.PREMIUM_PLAN_ID_MONTHLY_US;
+  const monthlyPlanId = process.env.PREMIUM_PLAN_ID_MONTHLY_US ?? "";
   const yearlyPlanId = process.env.PREMIUM_PLAN_ID_YEARLY_US ?? "";
+  const bundlePlanId = process.env.SUBPLAT_BUNDLE_PRICE_ID ?? "";
 
-  const subscriptions = billingAndSubscriptionInfo.subscriptions;
+  if (!monthlyPlanId || !yearlyPlanId || !bundlePlanId) {
+    logger.error("One or more plan IDs are not set in environment variables");
```
rhelmer, line 41: This is being logged, but the code continues on. Should there be a `return` or ` throw`?

rhelmer (changes requested):
lgtm overall, I noticed at least one place that might need a `return` or `throw` (it's just logging )

src/app/functions/server/user.ts
codemist: No, [that's not possible according to subplat](https://mozilla.slack.com/archives/CG0AJ6E77/p1748007960824179?thread_ts=1747938518.517689&cid=CG0AJ6E77). It would only ever be one plan at a time, so if a user upgrades from yearly to bundle, it'll be replaced with the bundle ID.

src/app/functions/server/user.ts
codemist: Yep — I went through all the places in the codebase where the old `checkUserHasMonthlySubscription()` was used and updated them to use the new `getUserSubscriptionType()` instead. In each case, I replaced the logic with g`etUserSubscriptionType() === "monthly"` to maintain the same behavior.

src/app/functions/server/user.ts
codemist: Great question and good catch. I updated `getUserSubscriptionType` to include an early return if any of the required plan ID env vars (monthly, yearly, or bundle) are missing. Previously, it was just logging the error and continuing, which could lead to invalid comparisons later in the function. Now, if those values aren’t set, the function exits immediately and returns `undefined`, which aligns with the other early-exit cases in the function.

codemist: https://github.com/mozilla/blurts-server/pull/5961/commits/71324f2d3a1df77d674fea58f66e73aad6b1233c

src/app/functions/server/user.ts
```diff
@@ -5,47 +5,51 @@
...
-  if (planIds.includes(yearlyPlanId)) {
-    console.error("User has yearly plan set");
-    return false;
-  }
+  if (planIds.includes(yearlyPlanId)) return "yearly";
+  if (planIds.includes(bundlePlanId)) return "bundle";
+  if (planIds.includes(monthlyPlanId)) return "monthly";
 
-  return planIds.includes(monthlyPlanId);
+  return;
```
codemist, line 53: I removed the `?? ""` fallback on the plan ID env vars because we’re now explicitly checking for falsy values (undefined or empty string) and returning early if any are missing. With the fallback in place, missing env vars would default to an empty string, which is still falsy, but it made the intent less clear and could potentially hide misconfiguration issues.

rhelmer (approved):

PR #5970, opened 2025-06-12
Add link forwarding for SubPlat subscriptions

flozia:
# References:

Jira: [MNTOR-4263](https://mozilla-hub.atlassian.net/browse/MNTOR-4263)
Figma:

# Description

Adds a route for link forwarding to SubPlat for external links, such as emails and announcements. 

# How to test

Depending on the feature flag `SubPlat3` the following links should redirect to the respective SubPlat plans:
- monthly: `/link/subscribe/monthly`
- yearly: `/link/subscribe/yearly`
- bundle: `/link/subscribe/bundle`


src/app/link/subscribe/[slug]/route.ts
```diff
@@ -0,0 +1,53 @@
...
+  logger.info("link_subscribe_start", { subscriptionPlanId });
+
+  if (!subscriptionPlanIds.includes(subscriptionPlanId as SubscriptionPeriod)) {
+    logger.info("link_subscribe_fallback", { subscriptionPlanId });
+    return NextResponse.redirect("/subscription-plans", 302);
+  }
+
+  const enabledFeatureFlags = await getEnabledFeatureFlags({
+    isSignedOut: true,
+  });
```
Comment (codemist) [inline src/app/link/subscribe/[slug]/route.ts:34]: This wouldn't throw any errors for authenticated users, right?

Comment (flozia) [reply src/app/link/subscribe/[slug]/route.ts:34]: No, that does not cause any errors for them.

src/app/link/subscribe/[slug]/route.ts
```diff
@@ -0,0 +1,53 @@
...
+};
+
+// The intent of this route is to forward external links to a matching SubPlat
+// subscription plan: If the slug is invalid we fallback to Monitors
+// publicly available subscription plans page.
+export async function GET(req: NextRequest, { params }: RequestParams) {
+  const { slug: subscriptionPlanId } = await params;
+  logger.info("link_subscribe_start", { subscriptionPlanId });
+
+  if (!subscriptionPlanIds.includes(subscriptionPlanId as SubscriptionPeriod)) {
```
Comment (codemist) [inline src/app/link/subscribe/[slug]/route.ts:27]: qn, is there a deeper reason for casting `subscriptionPlanId` here?

Comment (flozia) [reply src/app/link/subscribe/[slug]/route.ts:27]: Yes: The slug can’t have the type `SubscriptionPeriod` directly. Next.js insists it has to be a `string` (which is correct).

PR #5972, opened 2025-06-13
Respect allow list for the feature flag SubscriptionPlansPage

flozia:
# Description

Respect the allow list for accessing `/subscription-plans`.

# How to test

- Add email to allow list of `SubscriptionPlansPage`
- Visit `/subscription-plans`


src/app/(proper_react)/(redesign)/(public)/subscription-plans/page.tsx
```diff
@@ -37,9 +37,14 @@ export default async function Page() {
 
   const l10n = getL10n(await getAcceptLangHeaderInServerComponents());
   const eligibleForPremium = isEligibleForPremium(countryCode);
-  const enabledFeatureFlags = await getEnabledFeatureFlags({
-    isSignedOut: true,
-  });
+  const enabledFeatureFlags = await getEnabledFeatureFlags(
+    typeof session?.user.subscriber?.fxa_uid === "string"
```
rhelmer, line 41: I know we use this pattern elsewhere but it always makes me a little nervous... I think we should move this to a function and do more checks than this that it looks valid, and look for other ways to confirm that the session is signed in.

Would you mind filing a followup ticket for that

rhelmer (approved):

src/app/(proper_react)/(redesign)/(public)/subscription-plans/page.tsx
flozia: We would indeed benefit from moving our session checks to a shared function, I filed [MNTOR-4587](https://mozilla-hub.atlassian.net/browse/MNTOR-4587).

PR #5994, opened 2025-06-30
Disable more flaky unit tests

Vinnl:
This is a followup to https://github.com/mozilla/blurts-server/pull/5988 - I had searched for `applyCouponPromise`, but there's another variable with the same problem (`applyCouponErrorPromise`).


src/app/(proper_react)/(redesign)/(authenticated)/user/plus-expiration/View.tsx
Vinnl, line 1: The diff makes it look like a lot has changed, but I have only added the `c8 ignore` comments here.

src/app/(proper_react)/(redesign)/(authenticated)/user/plus-expiration/PlusExpiration.test.tsx
```diff
@@ -127,7 +130,10 @@ it("hides the error after dismissing it", async () => {
   ).not.toBeInTheDocument();
 });
 
-it("allows retrying after an error", async () => {
+// Wrapping `applyCouponErrorPromise` in an `act` still doesn't appear to be enough
+// to avoid intermittent test failures, so skipping this test for now.
+// eslint-disable-next-line jest/no-disabled-tests
+it.skip("allows retrying after an error", async () => {
```
codemist, line 140: Qn, can we just delete this test entirely instead of skipping it?

Vinnl: The reason I went with skipping is that ideally, we'd still have these tests - but we'd need to figure out why they are flaky. But this way, if someone sees the ignore comments and wants to dive into it, they can at least find the existing tests and work from those, rather than setting them up from scratch before finding out they're running into just the same issues.

Probably nobody's ever going to do that, but who knows... And meanwhile it doesn't really do any harm - unless I missed any?

codemist: That's totally valid, did we file a ticket for that?

Vinnl: That's an excellent callout and I should have filed that first. I created https://github.com/mozilla/blurts-server/issues/6004, and referenced it in the comments: 784c3520c

PR #6010, opened 2025-07-08
Simplified E2E test suite starter

flozia:
# References:

Jira: [MNTOR-4631](https://mozilla-hub.atlassian.net/browse/MNTOR-4631)

# Description

This adds a new E2E playwright test suite `/e2e-playwright` we can build up on. What it does for now:
- Minimal test example: Verify some general setup and load the landing page
- It runs for a combination of multiple devices and geolocations
- Tests all environments: `local`, `stage` & `production`
- Keeps the feature flags and experiment data consistent
- Only 2 Github Actions: PR & cronjob, deprecating the ones for the test suite `/e2e`

# How to test

`npm run e2e-playwright`

[MNTOR-4631]: https://mozilla-hub.atlassian.net/browse/MNTOR-4631?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


flozia:
> @flozia Shall I close #5923?

@Vinnl If that's all right with you. The plan is to have a fresh start and have any flakiness related to E2E test failures addressed from this PR on going forward.

Vinnl:
Yeah sure. And one quick thing I noticed as I opened the diff here: the CI jobs really shouldn't have the `--update-snapshots` argument - there's no point in having snapshot tests if they never fail :sweat_smile:

src/app/functions/server/getCountryCode.test.ts
```diff
@@ -6,6 +6,10 @@ import { it, expect, jest } from "@jest/globals";
 import { getCountryCode } from "./getCountryCode";
 import { ReadonlyHeaders } from "next/dist/server/web/spec-extension/adapters/headers";
 
+afterEach(() => {
+  delete process.env.app_env;
+});
```
codemist, line 11: @flozia Just started looking through this PR: what does this do?

flozia: It deletes `process.env.app_env` to reset in case it gets set like it is in [L85](https://github.com/mozilla/blurts-server/pull/6010/files/cf52b598085e4f6c42768fa6e116533adff16096#diff-c59ebd07950acf9262b978d793ac54c72381aa1d9c7620ac461966c25184019cR85).

e2e-playwright/fixtures/baseTest.ts
```diff
@@ -0,0 +1,80 @@
...
+  "LandingPageRedesign",
+  "CirrusV2",
+  "DataBrokerRemovalAttempts",
+  "SidebarNavigationRedesign",
+  "EditScanProfileDetails",
+  "SubPlat3",
+  "Announcements",
+  "PrivacyProductsBundle",
+  "SubscriptionPlansPage",
+];
```
codemist, line 25: This is _**so**_ nice to have - is it possible to export this from `featureFlags.ts` so we don't have to update it in multiple places, or is it intentionally meant to be different feature flags enabled here?

flozia: It would be possible to import them from `featureFlags.ts`, but those are meant to be different. The ones here are enabled for local runs and the ones in `featureFlags.ts` represent the full list of available flags.

codemist: Would there be any issues importing all of them?

e2e-playwright/fixtures/baseTest.ts
```diff
@@ -0,0 +1,80 @@
...
+    const enabledFeatureFlags: FeatureFlagName[] =
+      testInfo.project.use?.enabledFeatureFlags ?? [];
+    const mergedFlags: FeatureFlagName[] = Array.from(
+      new Set([...enabledFeatureFlags, ...localForcedFeatureFlags]),
+    );
+    await use(mergedFlags);
+  },
+  // See https://github.com/microsoft/playwright/issues/9468#issuecomment-943707670
+  // Sets the `x-forced-feature-flags` and `x-nimbus-preview-mode` on every
+  // request to Monitor.
```
codemist, line 59: Kudos, these links to relevant documentation is really helpful! It would be pretty gnarly to go through the code otherwise.

e2e-playwright/playwright.config.ts
```diff
@@ -0,0 +1,167 @@
...
+dotenvFlow.config();
+/**
+ * @see https://playwright.dev/docs/test-configuration
+ */
+
+const port = parseInt(process.env.PORT || "6060", 10);
+const __dirname = path.dirname(fileURLToPath(import.meta.url));
+
+// Helpful for local debugging: Don’t build application if `false`
+const forceLocalRebuild = true;
```
codemist, line 24: We were running tests by default locally before?

e2e-playwright/tests/general.spec.ts
```diff
@@ -0,0 +1,67 @@
...
+  defaultLocalForcedFeatureFlags,
+} from "../fixtures/baseTest";
+import { getEnabledFeatureFlags } from "../playwright.config";
+import { FeatureFlagName } from "../../src/db/tables/featureFlags";
+import { getBaseTestEnvUrl } from "../utils/helpers";
+
+const isLocal = process.env.E2E_TEST_ENV === "local";
+
+const extraFeatureFlags: FeatureFlagName[] = ["GA4SubscriptionEvents"];
+test.use({ extraLocalForcedFeatureFlags: extraFeatureFlags });
```
codemist, line 18: What are these extra feature flags for, and why is it just `GA4SubscriptionEvents`? A comment might be helpful here.

e2e-playwright/tests/general.spec.ts
```diff
@@ -0,0 +1,67 @@
...
+    );
+    expect(geolocation.longitude).toBe(
+      testInfo.project.use.geolocation?.longitude,
+    );
+
+    const language = await page.evaluate(() => navigator.language);
+    expect(language).toBe(testInfo.project.use.locale);
+  });
+
+  test("uses expected enabled feature flags", async ({
```
codemist, line 46: About this test and the one below it: curious what the significance of separating the “expected enabled feature flags” and “extending local feature flags” into two distinct tests?

src/app/functions/server/getCountryCode.test.ts
```diff
@@ -76,3 +80,18 @@ it("defaults to US", () => {
...
   expect(getCountryCode(headers as any)).toBe("us");
 });
+
+it("falls back to language when region is missing and `app_env` is local", () => {
+  process.env.app_env = "local";
+
+  const headers: Partial<jest.Mocked<ReadonlyHeaders>> = {
+    get: jest.fn((header: string) => {
+      if (header === "Accept-Language") {
+        return "fr";
```
codemist, line 90: Why is this returning `fr`?

e2e-playwright/fixtures/baseTest.ts
flozia: Yes, we don’t want to have all of the enabled by default when we add new ones.

e2e-playwright/playwright.config.ts
flozia: They can run locally or in CI against `local`, `stage` and `production` environments. Before, we had them run locally as well.

e2e-playwright/tests/general.spec.ts
flozia: This just tests that adding and enforcing additional feature flags works in the E2E tests. Added a comment in [457b448](https://github.com/mozilla/blurts-server/pull/6010/commits/457b448d7282b9243ea7eb969109112862f9d6c6).

e2e-playwright/fixtures/baseTest.ts
codemist: My concern is that we're now maintaining feature flags in multiple places. As we start deprecating old flags (or adding new ones), this scattered setup could become hard to track.. especially since it looks like a random mix, the result of tech debt from not cleaning them up earlier. It might be confusing in the future if someone adds a new flag and doesn't realize it also needs to be updated here.

Would it make sense to do something similar to how we handle `adminOnlyFlags`?

e2e-playwright/tests/general.spec.ts
flozia: “expected enabled feature flags” tests the enabled feature flags for the environment we are in and “extending local feature flags” confirms that extending feature flags in the tests works as expected.

src/app/functions/server/getCountryCode.test.ts
flozia: This is just a mock to return an accepted language and then test the return value of `getCountryCode`.

flozia:
> @flozia Are you looking to merge this in first, and then add the additional tests after just so it's decoupled (and we can start seeing tests pass again?)

@codemist Yes, we would want to land this first so we can test running them in the cronjob as well against all environments. I’m working on adding and extending the suite with the additional test cases we identified.

e2e-playwright/tests/general.spec.ts
codemist: So extending the feature flags means enabling all of them in `featureFlags.ts`?

flozia: No, it only means enabling them for the current test run in `enabledFeatureFlags` to have control over them in case we want to test features that involve toggling a flag for example.

e2e-playwright/fixtures/baseTest.ts
flozia: The idea behind this list of feature flags is that everyone has a consistent set of enabled feature flags locally. Primarily, I’ve added the ability to have full control of testing the new setup locally and in CI.

It is a good point that it gets hard to maintain in the future. I’m already loading the enabled feature flags from `stage` so we could use them here. The issue that I was trying to prevent with having this separate list is that toggling flags on `stage` will change local setups, which could make for hard to understand bugs as well.

e2e-playwright/playwright.config.ts
codemist: I see - should this value be set to `false` for local?

flozia: It’s better to have it set to `true` so the tests always run with a fresh build. This is only meant to help and speed up local dev and testing.

e2e-playwright/fixtures/baseTest.ts
flozia: @codemist  I like the idea of handling the default feature flags similar to how we do it with `adminOnlyFlags`. I’ll add this with my follow-up!

PR #6017, opened 2025-07-15
Fixes for upsell CTAs

flozia:
# References:

Jira:
- [MNTOR-4582](https://mozilla-hub.atlassian.net/browse/MNTOR-4582)
- [MNTOR-4592](https://mozilla-hub.atlassian.net/browse/MNTOR-4592)
- [MNTOR-4707](https://mozilla-hub.atlassian.net/browse/MNTOR-4707)

# Description

# How to test

- Navigate to the dashboard
- Click the upsell badge in the navbar with the feature flag `DisableOneRepScans`:
  - enabled: `/subscription-plans` should open in a new tab
  - disabled: The subscription waitlist dialog should be shown


src/app/(proper_react)/(redesign)/(authenticated)/user/(dashboard)/dashboard/fix/data-broker-profiles/manual-remove/ManualRemove.test.tsx
```diff
@@ -36,6 +37,15 @@ jest.mock(
   },
 );
 
+// We need to override the types of `useTelemetry` here, because otherwise
+// Jest infers incorrect types in `toHaveBeenCalledWith`, and throws an error.
+// See https://github.com/jestjs/jest/issues/15703
```
rhelmer, line 42: Thanks for linking to the issue... unfortunate, hopefully they'll fix upstream.

rhelmer:
Lots of nice cleanup in here thanks @flozia

src/app/components/client/dialog/Dialog.module.scss
```diff
@@ -48,36 +48,6 @@
     }
   }
 
-  &.horizontal {
```
Vinnl, line 51: Praise: I know unused CSS doesn't automatically get called out by our static analysis tools, so nice job being so thorough manually!

src/app/components/client/toolbar/UpsellBadge.module.scss
```diff
@@ -55,6 +53,15 @@
...
 
+// Override the background of the base button component
+a.upsellBadge,
+button.upsellBadge {
+  /* stylelint-disable-next-line declaration-no-important */
+  background: none !important;
+  /* stylelint-disable-next-line declaration-no-important */
+  color: tokens.$color-black !important;
+}
+
```
Vinnl: issue (blocking): This makes it risky to update the base button later, because whoever makes that edit likely won't know about these changes and whether they work well with it. (e.g. imagine a change that sets the border to transparent.) So I'd suggest either pulling these styles into that button component as another variant, or not using the base button.

`!important` also is too blunt an instrument to override styles, and should not be used. Instead, you could increase the specificity by e.g. adding the parent element to the selector, which would also emphasise that the change only applies in that context.

src/app/components/client/toolbar/UpsellBadge.tsx
```diff
@@ -155,44 +108,32 @@ function UpsellToggleButton(props: UpsellToggleButtonProps) {
...
         )}
-    </>
+    </div>
   );
 }
 
-export type UpsellBadgeProps = UpsellButtonProps & {
+export type UpsellBadgeProps = UpsellLinkButtonProps & {
   lastScanDate: Date | null;
   enabledFeatureFlags: FeatureFlagName[];
```
Vinnl: Maybe a little bit out of scope of this PR, but since it's such a small change... `enabledFeatureFlags` is already in `UpsellLinkButtonProps` :)

src/app/components/client/toolbar/UpsellBadge.tsx
```diff
@@ -27,19 +26,14 @@ import { ExperimentData } from "../../../../telemetry/generated/nimbus/experimen
...
 };
 
-export function UpsellButton(
-  props: UpsellButtonProps & {
-    label: string;
+export function UpsellLinkButton(
+  props: UpsellLinkButtonProps & {
+    children: ReactNode;
+    className?: string;
```
Vinnl, line 36: Question: why not add these directly to `UpsellLinkButtonProps`?

I'm also not a fan of exposing `className` to parent components: it requires the parent to be aware of implementation details of the child, and in turn the child has a harder time safely making styling changes, because it will have to check every call site to see how it will effect those styles. And I think in this case it's not being used anywhere?

src/app/(proper_react)/(redesign)/(authenticated)/user/(dashboard)/dashboard/Dashboard.test.tsx
```diff
@@ -61,12 +61,8 @@ import {
   DashboardUsPremiumScanInProgressUnresolvedBreaches,
 } from "./DashboardPlusUsers.stories";
 
-const routerPushMock = jest.fn();
-
 jest.mock("next/navigation", () => ({
-  useRouter: () => ({
-    push: routerPushMock,
-  }),
+  useRouter: jest.fn,
```
Vinnl: Possibly it's not used if this didn't cause problems, but I think you forgot something here :)

```suggestion
  useRouter: jest.fn(),
```

Vinnl (approved):
Always happy to see old feature flags getting removed, thanks!

Unfortunately, I have one blocking issue, though I think it should be fairly easy to resolve, and probably doesn't need another review when you do - so I'm pre-approving, but that should be fixed before merging :)

src/app/(proper_react)/(redesign)/(authenticated)/user/(dashboard)/dashboard/Dashboard.test.tsx
flozia: Added here: [91441dd](https://github.com/mozilla/blurts-server/pull/6017/commits/91441ddb31ef1b936877288a62deb814ca223529).

src/app/components/client/toolbar/UpsellBadge.tsx
flozia: Good catch! [5aadf62](https://github.com/mozilla/blurts-server/pull/6017/commits/5aadf62fb75dd79026a50dd8a3bad4b432f4cc0d)

src/app/components/client/toolbar/UpsellBadge.module.scss
flozia: Fair to call the use of `!important` out. Usually, I refrain from using it and we should keep it that way. This is such a one-off that I don’t think it makes sense to add another variant or theme to the button for this. I thought about not using the `<Button>`, but it is to perfect here for switching between an actual `button` and `link button`. I increased specificity in [fc1dffa](https://github.com/mozilla/blurts-server/pull/6017/commits/fc1dffa66484a089de9c01b9a2b06cdfea550338).

src/app/components/client/toolbar/UpsellBadge.tsx
flozia: `<UpsellLinkButton>` is used by `<UpsellToggleLinkButton>` and ultimately `<UpsellBadge>`, but `<UpsellLinkButton>` is also used standalone in other places. `UpsellLinkButtonProps` is reused for `<UpsellBadge>` that should not supposed to accept these props.

PR #6018, opened 2025-07-16
Simplified E2E test suite (Part 2)

flozia:
# References:

Jira: [MNTOR-4627](https://mozilla-hub.atlassian.net/browse/MNTOR-4627)

# Description

This PR builds up on [Simplified E2E test suite starter #6010](https://github.com/mozilla/blurts-server/pull/6010):
- Extends `global-setup` to create test accounts dynamically
- Adds `global-teardown` to clean up after E2E test runs
- Adds additional test specs for basic user flows:
  - Authentication
  - Initiate subscription flows
- Updates `.github/workflows/e2e_playwright_cron.yml` to run against `stage` & `production` while keeping the ability to run manually against one of these environments

# How to test

`npm run e2e-playwright`


e2e-playwright/fixtures/authenticatedTest.ts
```diff
@@ -0,0 +1,32 @@
...
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import fs from "fs";
+import { test as baseTest, expect } from "./baseTest";
+import { getTestUserSession } from "../utils/user";
+
+// The `use` function is no React hook and not linted correctly.
+// For more info see issue: https://github.com/facebook/react/issues/31237
+/* eslint-disable react-hooks/rules-of-hooks */
```
nshirley: I'm not sure if this linting rule will be hit anywhere else in these tests, but you could consider adding this as an override to your `eslint.config` just so you don't have to remember which rules to disable here and there. Not super necessary though! I _think_ it would be something like this.

```js
const config = [
  ...compat.config({
    extends: ["next"],
  }),
  {
  overrides: [
      {
        files: ['e2e-playwright/**/*.{js,ts,jsx,tsx}'],
        rules: {
          // disable all react rules for e2e files
          'react-hooks/rules-of-hooks': 'off',
          // any other rules to disable: `off`
      },
    ],
  // ...
  ]
```

e2e-playwright/global-teardown.ts
```diff
@@ -0,0 +1,54 @@
...
+
+  for (const userCountryCode in userEmails) {
+    const storageState = getTestUserSession(userCountryCode);
+    const context = await browser.newContext({ storageState });
+    const page = await context.newPage();
+
+    await page.goto(`${getBaseTestEnvUrl()}/user/settings/manage-account`);
+    await page.getByRole("button", { name: "Delete account" }).click();
+    const deleteDialog = page.getByRole("dialog", {
+      name: "Your ⁨Monitor⁩ account will be permanently deleted",
```
nshirley, line 35: GH is complaining about this line having hidden unicode characters. Just wanted to flag it if that's something that concerns you!
image

e2e-playwright/tests/auth.spec.ts
```diff
@@ -0,0 +1,63 @@
...
+    "user is signed in via session storage",
+    async ({ page }) => {
+      await page.waitForURL("**/user/**");
+    },
+  );
+
+  authenticatedTest(
+    "new users are redirected to the expected page",
+    async ({ page }, testInfo) => {
+      if (testInfo.project.use.countryCode === "nl") {
```
nshirley: **question: non-blocking**: I've noticed this conditional in another test. I'm curious how often this pattern will be used? It feels a bit cumbersome to manage, and if you wanted to include a new location it wouldn't automatically be handled in existing tests. 

fwiw, I don't have a good alternative to suggest right now, and not anything that needs to change. I'm mostly curious if you think it'll be a problem down the road?

e2e-playwright/utils/fxa.ts
nshirley, line 1: I know we've been talking about POM and using it over Slack. I think with what you have here it might actually be worth wrapping in a class to make it a typical page object. You could inject the `page` and other things like the `testInfo` so it doesn't need to be managed in the test. Granted, that may be a personal, stylistic choice, and I can see the value in having things very explicitly defined at the test level! 😅 

FxA uses a lot of [page objects](https://github.com/mozilla/fxa/blob/main/packages/functional-tests/pages/index.ts) and we [attach them](https://github.com/mozilla/fxa/blob/main/packages/functional-tests/lib/fixtures/standard.ts#L48) to the test so they're always available, just grab the pages you need in the [test](https://github.com/mozilla/fxa/blob/main/packages/functional-tests/tests/oauth/signin.spec.ts#L50).

nshirley (commented):
Thanks for letting me take a look at this! I just added a few general comments/questions, but nothing that I would suggest changes _right now_.

e2e-playwright/fixtures/authenticatedTest.ts
flozia: Yes, I think it is worth it to add the override rule since I’ve already added the inline comment at least two times. Added in https://github.com/mozilla/blurts-server/pull/6018/commits/bd8e1f90a2d602b5cd7cf646fb409ef67e82b678.

e2e-playwright/global-teardown.ts
flozia: These are being added by fluent in strings to isolate brand names and they were added here consciously. Thanks for flagging, though!

e2e-playwright/tests/auth.spec.ts
flozia: I’ve thought about this for a while while setting the tests up: We have very different experiences for US and non-US users the intend is to enforce covering both paths. Right now, I can’t see the need for us to add new locales anytime soon.

e2e-playwright/global-teardown.ts
nshirley: Thanks for the reminder that it's from Fluent! I _just_ ran into this myself in FxA for the first time (I've somehow avoided it for a few months) - great timing!

e2e-playwright/global-setup.ts
```diff
@@ -2,15 +2,20 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
-import { request } from "@playwright/test";
+import { chromium, request } from "@playwright/test";
```
rhelmer, line 5: Are we only testing in Chromium? We should document this somewhere if so. Given the proportion of Firefox desktop/mobile users we have, we should prioritize adding support for more browsers, but I think having anything working on all envs would be an improvement so that can come later.

flozia: This part here is only for the global setup before running the test suite. Specifically, for setting up user accounts dynamically and this is where chromium is used, not for the actual test cases. The rest of the suite is running in Chrome and Firefox and also for desktop and mobile viewports. The devices are setup in `playwright.config.ts` [here](https://github.com/mozilla/blurts-server/blob/e2e-simplified-tests/e2e-playwright/playwright.config.ts#L48-L75).

flozia:
> lgtm overall, thanks for working on this! I haven't tried running it yet, do we need changes to the `README`?

When it comes to running the tests nothing changed. I think it would be helpful to elaborate a bit more on the details of what the purpose of the new test suite. I’ve made some updates in the follow-up [Simplified E2E test suite (Part 3): Clean up #6021](https://github.com/mozilla/blurts-server/pull/6021/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5) — I’ll add more context there.

.env
```diff
@@ -110,6 +110,8 @@ E2E_TEST_ENV=
 E2E_TEST_BASE_URL=
 E2E_TEST_ACCOUNT_EMAIL=
 E2E_TEST_ACCOUNT_PASSWORD=
```
Vinnl, line 113: Can these and other old `E2E_TEST_ACCOUNT_*` variables now be deleted? (Or will that happen in the followup PR?) Would be really neat to only have the two.

.gitignore
```diff
@@ -23,7 +23,7 @@ src/e2e/specs/*-snapshots/**
 **/traces/
 
 # Playwright
-/e2e-playwright/*.json
+**/storage/
```
Vinnl: If it's not too much of a hassle, could we name this something like `e2e-cache/`, i.e. clarifying both that it's used by the e2e tests, and that it's safe to delete because it will be recreated.

e2e-playwright/utils/fxa.ts
```diff
@@ -0,0 +1,73 @@
...
+import { Page } from "@playwright/test";
+import { getBaseTestEnvUrl } from "./environment";
+import { fetchRestmailVerificationCode } from "./user";
+
+async function waitForFxa(page: Page) {
+  // Loading FxA can take a while
+  await page.waitForURL("**/oauth**", { timeout: 60_000 });
+}
+
+async function goToFxA(page: Page, isMobile?: boolean, countryCode?: string) {
```
Vinnl: Adding a note here: when I encountered this function, I did not understand what the second `false` parameter was for. Could we consolidate the second and third parameter into an `options` parameter, so that callers have to specify the property names as well?

e2e-playwright/utils/user.ts
```diff
@@ -0,0 +1,98 @@
...
+  const emails = getTestUserEmails();
+  const email = emails[countryCode ?? "us"];
+  if (!email) {
+    throw new Error(`No email found for [${countryCode}]`);
+  }
+
+  return email;
+}
+
+export function getTestUserSession(countryCode: string) {
```
Vinnl: From the function name I did not expect this to return a file path. Would a more appropriate name be something like `getTestUserSessionFilePath` or something like that?

.env
flozia: Yes, they can and are already part of the [follow-up](https://github.com/mozilla/blurts-server/pull/6021/files#diff-e9cbb0224c4a3d23a6019ba557e0cd568c1ad5e1582ff1e335fb7d99b7a1055d).

.gitignore
flozia: Sure, that’s easy to change: [f3b0b72](https://github.com/mozilla/blurts-server/pull/6018/commits/f3b0b729e51b3e600828a40300fdeb545b1b0946).

e2e-playwright/utils/fxa.ts
flozia: [Updated](https://github.com/mozilla/blurts-server/pull/6018/commits/d1f3c8ec7b2d8cb83b1a77ece1aea265a3d2555a) to take and `options` param.

e2e-playwright/utils/user.ts
flozia: Changed the function name in [0a218e5](https://github.com/mozilla/blurts-server/pull/6018/commits/0a218e58b8d46faedcd5591b3a508ad4a22fdf4a).

PR #6022, opened 2025-07-18
Address SCSS deprecation warnings

flozia:
# Description

While working on the new functional test setup I noticed some repeated SCSS deprecation warnings spamming the build logs:

```
Deprecation Warning […]:
color.hue() is deprecated. Suggestion:

color.channel($color, "hue", $space: hsl)
```

```
Deprecation Warning […]:
Sass's behavior for declarations that appear after nested
rules will be changing to match the behavior specified by CSS in an upcoming
version. To keep the existing behavior, move the declaration above the nested
rule. To opt into the new behavior, wrap the declaration in `& {}`.
```

# How to test

1. `npm run build`
2. There should be no deprecation warnings in the build logs related to SCSS


Vinnl (approved):
What? This was the only change needed to fix that nested declarations warning? Damn, I'd been so annoyed by it, even took a quick look myself already, but figured it would've been too much effort to address every instance across the codebase. If I had known... Thanks for doing this!

PR #6029, opened 2025-07-31
Fix padding for button variant link

flozia:
# References:

Jira: [MNTOR-4782](https://mozilla-hub.atlassian.net/browse/MNTOR-4782)

# Description

Fixes the padding for buttons with `variant="link"`.

# Screenshot

| Before | After |
| --- | --- |
| image | image |

[MNTOR-4782]: https://mozilla-hub.atlassian.net/browse/MNTOR-4782?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/app/components/client/Button.module.scss
+++ b/src/app/components/client/Button.module.scss
@@ -216,6 +216,10 @@
     line-height: 1.5;
     padding: tokens.$spacing-sm tokens.$spacing-md;
 
+    &.link {
+      padding: 0;
+    }
+
     &.icon {
       padding: tokens.$spacing-sm;
     }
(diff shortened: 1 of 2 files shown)


src/app/components/client/Button.module.scss
```diff
@@ -216,6 +216,10 @@
     line-height: 1.5;
     padding: tokens.$spacing-sm tokens.$spacing-md;
 
+    &.link {
+      padding: 0;
+    }
```
flozia, line 221: Technically, this is not needed as the combination of props `variant="link"` and `small` is not supposed to exist.

src/app/components/client/Button.tsx
```diff
@@ -35,6 +35,10 @@ export interface Props {
   theme?: ButtonColorThemes;
 }
 
+type Props =
+  | (BaseProps & { variant: "link"; small?: never })
+  | (BaseProps & { variant: Exclude<ButtonVariants, "link">; small?: boolean });
```
flozia, line 40: Buttons of variant `link` are only supposed to be used inline and don’t have a small alternate.

PR #6036, opened 2025-08-04
Upgrade to Next.js 15.4

Vinnl:
This upgrade was performed by running

    npx @next/codemod@canary upgrade latest

No further changes were needed. See [the release notes](https://nextjs.org/blog/next-15-4).

I upgrade manually (i.e. not through Dependabot) because Next.js sets overrides for React (I think to ensure that all transitive dependencies use the same version of React, because multiple versions can conflict), which in turn [confused Dependabot](https://github.com/mozilla/blurts-server/network/updates/1067797664):

image


EMMLynch (approved):
Approved

PR #6037, opened 2025-08-05
MNTOR-4818 — next auth error and logging

rhelmer:
# References:

Jira: MNTOR-4181

# Description

Next auth handles OAuth callback exceptions internally, which results in HTTP 500 response and multi-line errors that are difficult to filter in GCP logging console.

This PR adds:
* override for logging so it uses Winston
* catch and return HTTP 400 (client error) instead of letting exception bubble up to become a server error

# How to test

Deleting `next-auth.*` cookies during auth flow is the simplest way to trigger this locally.


src/app/api/auth/[...nextauth]/route.ts
```diff
@@ -7,15 +7,36 @@ import { NextRequest, NextResponse } from "next/server";
...
+}
+
 const handler = async (req: NextRequest, res: unknown) => {
-  // There is currently no support for handling OAuth provider callback errors:
-  // https://github.com/nextauthjs/next-auth/discussions/8209
   if (req.method === "GET") {
-    const requestSearchParams = req.nextUrl.searchParams;
-    const requestErrorQuery = requestSearchParams.get("error");
-    // Check if login is required: If the callback URL is available redirect to
-    // the authentication flow and otherwise fallback to the base URL.
```
Comment (Vinnl) [inline src/app/api/auth/[...nextauth]/route.ts:17]: I'm not entirely sure if I follow this comment (I think I miss the "why"), but it feels like maybe it should be left in?

src/app/api/auth/[...nextauth]/route.ts
```diff
@@ -25,6 +46,24 @@ const handler = async (req: NextRequest, res: unknown) => {
 
       return NextResponse.redirect(redirectUrl);
     }
+
+    if (
+      pathname.includes("/callback") &&
+      error &&
+      errorDescription &&
+      isSafeClientError(error)
+    ) {
```
Comment (Vinnl) [inline src/app/api/auth/[...nextauth]/route.ts:63]:
```suggestion
    ) {
    // By default, Next-Auth handles OAuth callback exceptions internally,
    // which results in HTTP 500 response and multi-line errors that are difficult to filter for.
    // This turns it into a 400 response that can more easily be distinguished from actual internal server errors.
```

src/app/api/auth/[...nextauth]/route.ts
Comment (rhelmer) [reply src/app/api/auth/[...nextauth]/route.ts:17]: Good point that was added in https://github.com/mozilla/blurts-server/commit/4eee9f19cce3f8bceeb136c907d7de5173620bb8 I'll look more closely at that

Comment (rhelmer) [reply src/app/api/auth/[...nextauth]/route.ts:17]: Also see https://github.com/mozilla/blurts-server/pull/5920

src/app/api/auth/[...nextauth]/route.ts
```diff
@@ -7,16 +7,45 @@ import { NextRequest, NextResponse } from "next/server";
...
+    const error = searchParams.get("error");
+    const errorDescription = searchParams.get("error_description");
+
+    // Handle the "login_required" error from the FxA OAuth callback.
+    // This error typically occurs during silent authentication (e.g. prompt=none) when no user session is present.
+    // In cases like clicking through the Relay Bento menu with no FxA login,
+    // the callback results in a failed silent sign-in. Instead of showing an error page,
+    // we redirect the user to the originally intended callback URL (if present) or to the base URL.
+    // This ensures the user sees the correct landing page and is guided to explicitly sign in if needed.
+    // See: FXA-11730 for upstream fix discussion.
```
Comment (Vinnl) [inline src/app/api/auth/[...nextauth]/route.ts:45]: This comment is great @rhelmer, very clear!

PR #6038, opened 2025-08-05
Add some input validation to admin endpoints

Vinnl:
# References:

Jira: MNTOR-4830

# Description

GitHub Code Scanning raised an alert; it's not a risk, since this is on endpoints only accessible by admins, but adding some input validation is still a good idea to better be safe than sorry.


rhelmer (approved):
This check is pretty repetitive, should we create a utility function similar to https://github.com/mozilla/blurts-server/blob/main/src/app/functions/server/checkSession.ts (that one just checks for valid session but not admin check specifically)

Vinnl:
@rhelmer I'm not sure - we could widen the `isAdmin` function to allow getting passed `undefined` and then returning `false`, which would save one leg, but the other conditions aren't admin checks. These specific functions can not be used in prod (this isn't true for all admin-gated ones), and the input validation is specific to these APIs, of course.

PR #6040, opened 2025-08-07
Adopt Turbopack for local development

Vinnl:
Next.js has supported using Turbopack for local development [since 15.0](https://nextjs.org/blog/next-15#turbopack-dev), and I tried it out then. Unfortunately, it included [a bug](https://github.com/vercel/next.js/issues/78019) that prevented it from loading FTL files, but that [was fixed](https://github.com/vercel/next.js/commit/84edd37d0a58d7d5a7d6ebb5f4ae424750011be1#commitcomment-161800505) with [15.4](https://nextjs.org/blog/next-15-4), and now it works!

This shouldn't affect the release on stage/prod, but locally you could see a speedup. It'll probably also help us catch potential issues that might arise [once](https://nextjs.org/blog/next-15-4#turbopack-builds) production builds start using it too.

To test, it would be good to see if your `npm run dev` still works properly locally.

### DIFF
--- a/next.config.js
+++ b/next.config.js
@@ -169,11 +169,23 @@ const nextConfig = {
 
     return config;
   },
+  turbopack: {
+    rules: {
+      "*.ftl": {
+        loaders: ["raw-loader"],
+        as: "*.js",
+      },
+    },
+  },
   // Without adding MJML here, Next.js has Webpack trying and failing to load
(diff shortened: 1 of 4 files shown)


flozia (approved):
Nice! Running `npm run dev` works as expected for me.

PR #6065, opened 2025-08-27
Resolve Storybook build failures by not importing server actions

Vinnl:
Reading https://github.com/mozilla/blurts-server/pull/6063 made me realise that a bunch of server code was getting imported in Storybook, which it doesn't/can't support. While mocking out the specific server-specific modules that happened to be imported at this point works for now, we'll run into similar errors as soon as we import other such modules - and that can happen anywhere in the transitive imports from the story.

Thus, I think it's probably safer to avoid server actions in the story in general, just like we're doing in e.g. `SetttingsStoryWrapper`. This PR does that.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -20,7 +20,7 @@
         "@leeoniya/ufuzzy": "^1.0.18",
         "@mozilla/glean": "^5.0.4",
         "@next/third-parties": "15.4.5",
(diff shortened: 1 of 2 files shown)


flozia (approved):
Thanks for fixing the build failure!

PR #6066, opened 2025-08-27
Upgrade to Storybook 9.1

Vinnl:
We were still on Storybook 8, but 9 has been released for a while - I think Dependabot doesn't upgrade major versions, sometimes? Unfortunately that's not entirely clear to me.

Anyway, there are a lot of changes, but they're mostly the same ones, implemented by Storybook's codemods. One new feature is an ESLint plugin, so I've also resolved the things that that pointed out.

(Another new feature is [supposedly-easier mocking](https://storybook.js.org/blog/next-generation-module-mocking/), which I was hoping to use for https://github.com/mozilla/blurts-server/pull/6065. Alas, I couldn't get that to work - but since I had already done the upgrade, I figured I'd submit that as well.)

### DIFF
--- a/src/app/(proper_react)/(redesign)/(authenticated)/user/(dashboard)/dashboard/Dashboard.stories.tsx
+++ b/src/app/(proper_react)/(redesign)/(authenticated)/user/(dashboard)/dashboard/Dashboard.stories.tsx
@@ -28,6 +28,9 @@ import {
   ScanData,
 } from "../../../../../../functions/server/moscary";
 
+// Exported values should be stories, but this was already there when
+// Storybook added this lint rule, so I guess it works?
+// eslint-disable-next-line storybook/prefer-pascal-case
 export const brokerOptions = {
   "no-scan": "No scan started",
   empty: "No scan results",
@@ -36,6 +39,9 @@ export const brokerOptions = {
   "scan-in-progress": "Scan is in progress",
   "manually-resolved": "Manually resolved",
 };
+// Exported values should be stories, but this was already there when
+// Storybook added this lint rule, so I guess it works?
+// eslint-disable-next-line storybook/prefer-pascal-case
 export const breachOptions = {
   empty: "No data breaches",
   unresolved: "With unresolved data breaches",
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -124,14 +124,10 @@
...
     "@google-cloud/bigquery": "^7.9.4",
     "@playwright/test": "^1.52.0",
-    "@storybook/addon-a11y": "^8.6.14",
-    "@storybook/addon-actions": "^8.6.14",
-    "@storybook/addon-essentials": "^8.6.14",
-    "@storybook/addon-interactions": "^8.6.14",
-    "@storybook/addon-links": "^8.6.14",
-    "@storybook/nextjs": "^8.6.14",
-    "@storybook/react": "^8.6.14",
-    "@storybook/test": "^8.5.2",
```
codemist, line 134: Is this getting removed? I noticed it was being imported still in this PR: https://github.com/mozilla/blurts-server/pull/6065/files

Perhaps this should go in first? What do you think?

src/app/(proper_react)/(redesign)/(authenticated)/user/(dashboard)/dashboard/Dashboard.stories.tsx
```diff
@@ -28,6 +28,9 @@ import {
   ScanData,
 } from "../../../../../../functions/server/moscary";
 
+// Exported values should be stories, but this was already there when
+// Storybook added this lint rule, so I guess it works?
+// eslint-disable-next-line storybook/prefer-pascal-case
```
codemist, line 33: Is the comment about exported values related to a rule in Storybook 9.1?

package.json
Vinnl: Yeah, that's why I layered this PR on top of that PR (i.e. it merges into that PR's branch, rather than into `main`) - the dependency got merged into the core package, so the imports are now changed from `@storybook/test` to `storybook/test`.

src/app/(proper_react)/(redesign)/(authenticated)/user/(dashboard)/dashboard/Dashboard.stories.tsx
Vinnl: Yeah, they added ESLint rules that demanded that stories are PascalCased, which these are not - but they're not stories anyway.

PR #6069, opened 2025-08-29
MNTOR-4875: sync moscary id

mansaj:
# References:

Jira: MNTOR-4875
Figma:

# Description
Add syncMoscaryProfiles cron job with batching

- new cron job command for syncing Moscary profiles in package.json.
- `syncMoscaryProfiles.ts` script to handle the synchronization of subscribers' Moscary IDs based on their OneRep profile IDs, including logging for success and failure cases.
- configurable batch size for processing subscribers, improving efficiency.
- pagination in subscriber retrieval to handle large datasets.
- error handling to adjust batch size dynamically on failure.

# Screenshot (if applicable)

Not applicable.

# How to test


src/scripts/cronjobs/syncMoscaryProfiles.ts
```diff
@@ -0,0 +1,288 @@
...
+async function getSubscribersNeedingSync(
+  offset: number = 0,
+  limit: number = BATCH_SIZE,
+): Promise<SubscriberRow[]> {
+  const subscribers = await knexSubscribers("subscribers")
+    .select("*")
+    .whereNotNull("onerep_profile_id")
+    .whereNull("moscary_id")
+    .orderBy("id") // Consistent ordering for pagination
+    .offset(offset)
```
rhelmer: optimization(nonblocking): One problem with using offset pagination is that if subscribers table is updated while it's running you might skip or reprocess rows. If this is safe to reprocess and this approach is simpler then no need to change, but you might consider keyset/seek pagination. It's more efficient (doesn't scan) and won't skip rows, it's also pretty easy to implement.

src/scripts/cronjobs/syncMoscaryProfiles.ts
```diff
@@ -0,0 +1,288 @@
...
+  const result = await syncMoscaryProfiles();
+
+  logger.info("moscary_profile_sync_script_completed", {
+    totalSubscribers: result.totalSubscribers,
+    successfulSyncs: result.successfulSyncs,
+    failedSyncs: result.failedSyncs,
+    batchesProcessed: result.batchesProcessed,
+  });
+
+  process.exit(0);
```
rhelmer: cleanup(blocking): Should the DB connection be destroyed here?

src/scripts/cronjobs/syncMoscaryProfiles.ts
```diff
@@ -0,0 +1,288 @@
...
+}
+
+/**
+ * Update subscriber with Moscary ID
+ */
+async function updateSubscriberMoscaryId(
+  subscriber: SubscriberRow,
+  moscaryId: string,
+): Promise<void> {
+  await setMoscaryId(subscriber, moscaryId as UUID);
```
rhelmer, line 130: validation(blocking): let's make sure it's a UUID, the `uuid` package has a validator or just checking with regex is fine

src/scripts/cronjobs/syncMoscaryProfiles.ts
```diff
@@ -0,0 +1,288 @@
...
+): Promise<{ successful: number; failed: number }> {
+  let successful = 0;
+  let failed = 0;
+
+  logger.info("processing_batch", {
+    batchNumber,
+    batchSize: subscribers.length,
+  });
+
+  for (const subscriber of subscribers) {
```
rhelmer: optimization(non-blocking): you could consider parallelizing this, `p-limit` makes it pretty easy to spin up a bunch of promises and then `await Promise.allSettled`. Not important since I think it will run quickly after the first, but this problem seems "embarrassingly parallel".

src/scripts/cronjobs/syncMoscaryProfiles.ts
mansaj: Thanks! https://github.com/mozilla/blurts-server/pull/6069/commits/b8ff8c675a855ed76ba8dd8f9df31ef2d4a785a2

src/scripts/cronjobs/syncMoscaryProfiles.ts
mansaj: Good idea: https://github.com/mozilla/blurts-server/pull/6069/commits/749d0fed2de0c92a37eeb2e532f312527666c200

src/scripts/cronjobs/syncMoscaryProfiles.ts
mansaj: https://github.com/mozilla/blurts-server/pull/6069/commits/202aa172ce9b89b227e1a7fe43d415e0b8015a6e

src/scripts/cronjobs/syncMoscaryProfiles.ts
mansaj: https://github.com/mozilla/blurts-server/pull/6069/commits/53119bc9216a711be232acfd88c1826f925e04f4

src/scripts/cronjobs/syncMoscaryProfiles.ts
```diff
@@ -0,0 +1,364 @@
...
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import { logger } from "../../app/functions/server/logging";
+import { getProfile } from "../../app/functions/server/moscary";
+import { knexSubscribers, setMoscaryId } from "../../db/tables/subscribers";
+import type { SubscriberRow } from "knex/types/tables";
+import { UUID } from "crypto";
+import { validate as validateUuid } from "uuid";
+import pLimit from "p-limit";
```
Vinnl, line 11: issue (blocking): If we're explicitly depending on this, we should probably add it to the `package.json`.

src/scripts/cronjobs/syncMoscaryProfiles.ts
```diff
@@ -0,0 +1,364 @@
...
+  batchesProcessed: number;
+}
+
+// Configurable batch size with environment variable support
+let BATCH_SIZE = parseInt(process.env.MOSCARY_SYNC_BATCH_SIZE ?? "100", 10);
+
+if (Number.isNaN(BATCH_SIZE) || BATCH_SIZE <= 0) {
+  BATCH_SIZE = 100;
+  logger.warn(
+    `Invalid MOSCARY_SYNC_BATCH_SIZE environment variable, using default: ${BATCH_SIZE}`,
```
Vinnl, line 26: Praise: good thinking not letting this slip by quietly :)

src/scripts/cronjobs/syncMoscaryProfiles.ts
```diff
@@ -0,0 +1,364 @@
...
+}
+
+// Configurable concurrency limit with environment variable support
+const CONCURRENCY_LIMIT = parseInt(
+  process.env.MOSCARY_SYNC_CONCURRENCY_LIMIT ?? "10",
+  10,
+);
+
+if (Number.isNaN(CONCURRENCY_LIMIT) || CONCURRENCY_LIMIT <= 0) {
+  logger.warn(
```
Vinnl, line 38: I think you forgot this?

```suggestion
  CONCURRENCY_LIMIT = 10;
  logger.warn(
```

src/scripts/cronjobs/syncMoscaryProfiles.ts
Vinnl, line 1: issue (blocking): I believe we need to handle `SIGINT` and `SIGTERM` interrupts for when GCP has issues, [like in this cronjob](https://github.com/mozilla/blurts-server/blob/aff520885426e40080df6fb67a056a15fdffe30d/src/scripts/cronjobs/monthlyActivityFree.tsx#L31-L38) for example.

src/scripts/cronjobs/syncMoscaryProfiles.ts
```diff
@@ -0,0 +1,364 @@
...
+    logger.error("error_closing_database_connection", {
+      error: error instanceof Error ? error.message : String(error),
+    });
+  }
+
+  // Graceful shutdown following pattern from other cronjob scripts
+  setTimeout(() => {
+    logger.info("script_shutdown_complete");
+    process.exit(0);
+  }, 1000);
```
Vinnl: suggestion (non-blocking): I don't see this in e.g. `monthlyActivityFree` - possibly because that does handle the `SIGINT`s and `SIGTERM`s? So maybe safest to remove this to avoid adding more confusion to future contributors?

src/scripts/cronjobs/syncMoscaryProfiles.ts
```diff
@@ -0,0 +1,364 @@
...
+
+/**
+ * Get subscribers that have onerep_profile_id but no moscary_id
+ * Uses keyset pagination to handle large datasets efficiently
+ */
+async function getSubscribersNeedingSync(
+  lastProcessedId: number = 0,
+  limit: number = BATCH_SIZE,
+): Promise<SubscriberRow[]> {
+  const subscribers = await knexSubscribers("subscribers")
```
Vinnl, line 51: question: Why aren't these queries in `tables/subscribers.ts`?

src/scripts/cronjobs/syncMoscaryProfiles.ts
```diff
@@ -0,0 +1,364 @@
...
+ */
+async function getTotalSubscribersNeedingSync(): Promise<number> {
+  const result = await knexSubscribers("subscribers")
+    .count("id as count")
+    .whereNotNull("onerep_profile_id")
+    .whereNull("moscary_id")
+    .first();
+
+  // @ts-ignore: count type
+  return parseInt(String(result?.count || 0), 10);
```
Vinnl: suggestion: Assuming the count will never need a BigInt, [the Knex docs recommend](https://knexjs.org/guide/query-builder.html#usage-with-typescript-3) using the Generic:

```suggestion
    .count<Record<string, number>>("id as count")
    .whereNotNull("onerep_profile_id")
    .whereNull("moscary_id")
    .first();

  return result?.count ?? 0;
```

src/scripts/cronjobs/syncMoscaryProfiles.ts
```diff
@@ -0,0 +1,364 @@
...
+
+interface SyncResult {
+  totalSubscribers: number;
+  successfulSyncs: number;
+  failedSyncs: number;
+  batchesProcessed: number;
+}
+
+// Configurable batch size with environment variable support
+let BATCH_SIZE = parseInt(process.env.MOSCARY_SYNC_BATCH_SIZE ?? "100", 10);
```
Vinnl: I'm assuming 100 consecutive requests to the Moscary API isn't going to cause it any trouble?

src/scripts/cronjobs/syncMoscaryProfiles.ts
```diff
@@ -0,0 +1,364 @@
...
+    // Use the OneRep Profile ID as the path parameter for legacy lookup
+    const profile = await getProfile(onerepProfileId);
+
+    // If we get here, the profile exists and we can return its UUID
+    logger.info("found_moscary_profile", {
+      onerepProfileId,
+      moscaryId: profile.id,
+    });
+
+    return profile.id;
```
Vinnl: suggestion: If you do the type assertion here, you don't have to do that everywhere you use it. (And I think we can safely assume that this will be a UUID, so no need to validate that again elsewhere. But if not, we probably already want to do that validation here.)

```suggestion
    return profile.id as UUID;
```

src/scripts/cronjobs/syncMoscaryProfiles.ts
```diff
@@ -0,0 +1,364 @@
...
+}
+
+/**
+ * Process a single subscriber
+ */
+async function processSubscriber(
+  subscriber: SubscriberRow,
+): Promise<{ success: boolean; error?: string }> {
+  try {
+    if (!subscriber.onerep_profile_id) {
```
Vinnl: potential issue: not sure if it's possible for someone to have a OneRep ID of `0`, but if so, we should check for a type of `undefined` rather than converting to a boolean.

src/scripts/cronjobs/syncMoscaryProfiles.ts
Vinnl: Ah lol, just checking on previously-resolved issues, I just suggested the validation is probably not needed since it'd be really weird if Moscary starts returning non-UUIDs for its IDs :P

src/scripts/cronjobs/syncMoscaryProfiles.ts
mansaj: I can drop it down to 20 first

src/scripts/cronjobs/syncMoscaryProfiles.ts
mansaj: Fixed https://github.com/mozilla/blurts-server/pull/6069/commits/870dea0cd1a4d781ef199efc5cb82af4b8d2e643

src/scripts/cronjobs/syncMoscaryProfiles.ts
mansaj: Fixed https://github.com/mozilla/blurts-server/pull/6069/commits/870dea0cd1a4d781ef199efc5cb82af4b8d2e643

src/scripts/cronjobs/syncMoscaryProfiles.ts
mansaj: Fixed https://github.com/mozilla/blurts-server/pull/6069/commits/870dea0cd1a4d781ef199efc5cb82af4b8d2e643

src/scripts/cronjobs/syncMoscaryProfiles.ts
mansaj: Fixed https://github.com/mozilla/blurts-server/pull/6069/commits/870dea0cd1a4d781ef199efc5cb82af4b8d2e643

src/scripts/cronjobs/syncMoscaryProfiles.ts
mansaj: Fixed https://github.com/mozilla/blurts-server/pull/6069/commits/870dea0cd1a4d781ef199efc5cb82af4b8d2e643

src/scripts/cronjobs/syncMoscaryProfiles.ts
mansaj: Fixed https://github.com/mozilla/blurts-server/pull/6069/commits/870dea0cd1a4d781ef199efc5cb82af4b8d2e643

src/scripts/cronjobs/syncMoscaryProfiles.ts
mansaj: Fixed https://github.com/mozilla/blurts-server/pull/6069/commits/870dea0cd1a4d781ef199efc5cb82af4b8d2e643

src/scripts/cronjobs/syncMoscaryProfiles.ts
mansaj: Good suggestion! but wouldn't it better to be applied to the entire chain?
```
const result = await knexSubscribers("subscribers")
  .count<Record<string, number>>("id as count")
  .whereNotNull("onerep_profile_id")
  .whereNull("moscary_id")
  .first<Record<string, number>>();
```

src/scripts/cronjobs/syncMoscaryProfiles.ts
mansaj: Don't think this is categorized as "reused".. It's just for this script so we don't need to shove more things there.

Also this script is meant to be temporary

rhelmer (approved):

PR #6083, opened 2025-09-04
MNTOR-4764: Deduplicate scan results

Vinnl:
# References:

Jira: MNTOR-4764
Figma:

# Description

Deduplication is relatively primitive, but it currently works as follows:
For every broker, sort the scan results by source (where `moscary` comes before `onerep`), then by `updated_at`. Remove scan results that point to the same URL. Then only keep the first three.

Note that for some brokers, different scan results will have the same URL, so we might want to revisit that in the future.

# How to test

Use the updated sandbox to return multiple results per broker, then see that only one is shown. For extra thoroughness, you can add some logging to the `deduplicateScanResults` function.

### DIFF
--- a/src/apiMocks/mockData.ts
+++ b/src/apiMocks/mockData.ts
@@ -44,6 +44,10 @@ export type RandomScanResultOptions = Partial<{
   status: RemovalStatus;
   manually_resolved: boolean;
   broker_status: DataBrokerRemovalStatus;
+  data_broker: string;
+  source: MoscaryData["ScanResult"]["source"];
+  url: MoscaryData["ScanResult"]["url"];
+  updated_at: Date;
 }>;
 
 /**
@@ -60,7 +64,7 @@ export function createRandomOnerepScanResult(
     options.status === "waiting_for_verification"
       ? faker.number.int({ min: 1, max: 42 })
       : undefined;
-  const url = faker.internet.url();
+  const url = options.url ?? faker.internet.url();
   return {
     id: faker.number.int(),
     onerep_scan_result_id: faker.number.int(),
@@ -85,10 +89,10 @@ export function createRandomOnerepScanResult(
     emails: [faker.internet.exampleEmail()],
     relatives: Array.from({ length: 3 }, () => faker.person.fullName()),
     link: url,
-    data_broker: new URL(url).hostname,
+    data_broker: options.data_broker ?? new URL(url).hostname,
     data_broker_id: faker.number.int(),
     created_at: options.createdDate ?? faker.date.recent({ days: 2 }),
-    updated_at: faker.date.recent({ days: 1 }),
+    updated_at: options.updated_at ?? faker.date.recent({ days: 1 }),
     optout_attempts,
     last_optout_at:
       typeof optout_attempts === "number" && optout_attempts > 0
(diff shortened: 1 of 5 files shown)


src/apiMocks/mockData.ts
```diff
@@ -140,18 +144,20 @@ export function createRandomMoscaryScanResult(
...
+    updated_at:
+      options.updated_at?.toISOString() ??
+      faker.date.recent({ days: 1 }).toISOString(),
     optout_attempts,
     last_optout_at:
       typeof optout_attempts === "number" && optout_attempts > 0
         ? faker.date.recent({ days: 3 }).toISOString()
         : undefined,
-    url: url,
+    source: options.source ?? faker.helpers.arrayElement(["moscary", "onerep"]),
```
rhelmer: question(blocker): is `moscary` the expected source? In `sortByRelevance` it prioritized source `monitor`.

Could we use the type here from the API docs?

src/app/functions/universal/dedupeScanResults.ts
```diff
@@ -0,0 +1,72 @@
...
+    const scanResultAUpdated = parseIso8601Datetime(scanResultA.updated_at);
+    const scanResultBUpdated = parseIso8601Datetime(scanResultB.updated_at);
+
+    // Newest one first:
+    return scanResultBUpdated.getTime() - scanResultAUpdated.getTime();
+  });
+}
+
+function removeIrrelevantResults(scanResults: MoscaryData["ScanResult"][]) {
+  return scanResults.filter((scanResult, currentIndex) => {
```
rhelmer: optimization(non-blocking): I'm not sure how large `n` could get here, but I think this is O(n^2), could probably reduce the complexity here a bit (untested):

```
function removeIrrelevantResults(scanResults: MoscaryData["ScanResult"][]) {
  const seenLinks = new Set<string>();
  return scanResults.filter((scanResult) => {
    const link = scanResult.link?.trim();
    if (!link) return true;
    if (seenLinks.has(link)) return false;
    seenLinks.add(link);
    return true;
  });
}
```

rhelmer (changes requested):
lgtm overall, I spotted one inconsistency with the `source`, could we use the type there to prevent that? Also suggestion about complexity on the `removeIrrelevantResults` function, but I did not spend the time to thoroughly test it so will leave up to you if there will be enough results for it to matter. I'd go with whatever is more readable/maintainable if there isn't a practical difference.

src/app/functions/universal/dedupeScanResults.ts
Vinnl: Yes, it is O(n^2), but I figured with a couple of hundred scan results worst case, it shouldn't be too much of an issue. But might as well fix it, and I think it's actually more readable your way too: 368fa6161 I went a bit too far in defaulting to pure functions :sweat_smile: 

(I'd hope the trimming isn't necessary, but left it in just in case.)

src/apiMocks/mockData.ts
Vinnl: It actually already did use that type, but the API docs type was just set to `string`. I had updated that to an enum in my other PR (after indeed I noticed that it used `monitor`), but I didn't want to regenerate the types here until that was actually merged. Now that I've merged it, I regenerated the types (c585c073b), and it flagged this as an error. Fixed in 5a482c73d.

PR #6095, opened 2025-09-08
MNTOR-4907 - Fix manual data broker removal view bugs

codemist:
# References:

Jira: MNTOR-4907
Figma:

# Description

It seems after the Moscary integration, when resolving data broker exposures manually in `ManualRemoveView`, cards behaved inconsistently: sometimes a resolved card would jump to the bottom of the list / a card would appear to disappear / a card would revert back to “Action needed” even after resolution.

# Screenshot (if applicable)

Not applicable.

# How to test
With a free account with moscary scan results that are unresolved, go to http://localhost:6060/user/dashboard/fix/data-broker-profiles/manual-remove and try to resolve it within the view.

### DIFF
--- a/src/app/(proper_react)/(redesign)/(authenticated)/user/(dashboard)/dashboard/[[...slug]]/page.tsx
+++ b/src/app/(proper_react)/(redesign)/(authenticated)/user/(dashboard)/dashboard/[[...slug]]/page.tsx
@@ -2,6 +2,9 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
+export const dynamic = "force-dynamic";
+export const revalidate = 0;
+
 import { headers } from "next/headers";
 import { redirect } from "next/navigation";
 import { getServerSession } from "../../../../../../../functions/server/getServerSession";
@@ -55,6 +58,7 @@ import {
   getScansCountForProfile,
   isEligibleForFreeScan,
 } from "../../../../../../../functions/server/moscary";
+import { connection } from "next/server";
 
 const dashboardTabSlugs = ["action-needed", "fixed"];
 
@@ -68,6 +72,7 @@ type Props = {
 };
 
 export default async function DashboardPage(props: Props) {
+  await connection();
   const searchParams = await props.searchParams;
   if (searchParams.dialog === "subscriptions") {
     return redirect("/subscription-plans");


src/app/(proper_react)/(redesign)/(authenticated)/user/(dashboard)/dashboard/fix/data-broker-profiles/manual-remove/RemovalCard.tsx
```diff
@@ -35,6 +35,11 @@ export const RemovalCard = (props: Props) => {
     props.scanResult.manually_resolved,
   );
 
+  // Sync local state with props when the server updates
+  useEffect(() => {
+    setIsResolved(props.scanResult.manually_resolved);
+  }, [props.scanResult.manually_resolved]);
+
```
codemist, line 42: In `async function resolve()` below this, right after calling `router.refresh()`, the backend should be returning the updated scan result with `manually_resolved: true` (I did check that this was working with the new backend, it's still the right result. That said, it seems like the UI could sometimes still show the old unresolved state, possibly because `router.refresh()` was interfering with the optimistic UI update. I can’t say for certain that this is the root cause, but adding a `useEffect` to sync local state with the latest prop feels like a reasonable safeguard.

src/app/(proper_react)/(redesign)/(authenticated)/user/(dashboard)/dashboard/fix/data-broker-profiles/manual-remove/RemovalCard.tsx
```diff
@@ -87,7 +92,9 @@ export const RemovalCard = (props: Props) => {
               });
             }}
           >
-            {l10n.getString("exposure-card-resolve-exposures-cta")}
+            {l10n.getString(
+              "fix-flow-data-broker-profiles-manual-remove-button-mark-fixed",
+            )}
```
codemist, line 97: Just added this here. I noticed `fix-flow-data-broker-profiles-manual-remove-button-mark-fixed` wasn't being used, and it doesn't make sense for this string to say "resolve exposures" in the manual removal flow. `<RemovalCard>` is only being used in the `ManualRemoveView` component so other exposure cards should still have the original cta label in tact.

src/app/(proper_react)/(redesign)/(authenticated)/user/(dashboard)/dashboard/fix/data-broker-profiles/manual-remove/ManualRemoveView.tsx
```diff
@@ -70,6 +70,9 @@ export function ManualRemoveView(props: Props) {
     "Scan",
   );
 
+  // freeze results at mount, so order never changes
+  const [initialResults] = useState(() => props.scanData.results);
```
codemist: I was only able to test this locally with sandboxed data, but in those tests I wasn’t able to reproduce the bug anymore. Resolved cards consistently stayed resolved, and the order of cards also remained the same.

codemist: https://github.com/mozilla/blurts-server/blob/main/src/db/tables/onerep_scans.ts#L172-L183

I believe why this wasn't happening with the original onerep scan results was because `getLatestOnerepScanResults` had a natural, stable order built into the query: it grouped results by broker `link` and then ordered by `onerep_scan_result_id`, so the list always appeared in the same order, regardless of status changes.

codemist: https://github.com/mozilla/blurts-server/blob/main/src/app/functions/server/moscary.ts#L347-L355 We can add similar ordering logic in `getScanAndResults` so the backend enforces an order, or as this PR is doing, simply freeze/order the results on the frontend. I’m okay with the latter because it keeps the backend fetch simple and predictable, and handling it in the UI, we ensure both OneRep and Moscary results behave consistently

flozia: Nit (non-blocking): You can also use `useMemo` here to “freeze” the results.

```suggestion
  const initialResults = useMemo(() => props.scanData.results, []);
```

codemist: Cool! Added it here https://github.com/mozilla/blurts-server/pull/6095/commits/16c827335dda449cffbaa2c2224361d172ec89da
Not super familiar with this hook, how do you discern when to use this over `useState`? I’ve noticed it being used a few times in the codebase, so I’m trying to get a sense of the conventions here.

EMMLynch (approved):
Approved.

PR #6062, opened 2025-08-26
Add Nimbus feature to feature flag

flozia:
# References:

Jira: [MNTOR-4184](https://mozilla-hub.atlassian.net/browse/MNTOR-4184)

# Description

Add Nimbus feature to the feature flag.

# How to test

Locally, you can test by changing the experiment default values in `nimbus.yaml` and rerunning `npm run build-nimbus`.

# Checklist (Definition of Done)

- [x] Commits in this PR are minimal and [have descriptive commit messages](https://chris.beams.io/posts/git-commit/).
- [x] I've added a unit test to test for potential regressions of this bug.
- [x] If this PR implements a feature flag or experimentation, I've checked that it still works with the flag both on, and with the flag off.
- [x] If this PR implements a feature flag or experimentation, the Ship Behind Feature Flag status in Jira has been set
- [x] Product Owner accepted the User Story (demo of functionality completed) or waived the privilege.
- [x] All acceptance criteria are met.
- [ ] Jira ticket has been updated (if needed) to match changes made during the development process.
- [ ] Jira ticket has been updated (if needed) with suggestions for QA when this PR is deployed to stage.

[MNTOR-4184]: https://mozilla-hub.atlassian.net/browse/MNTOR-4184?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #6097, opened 2025-09-09
Fix: Dynamically import next/headers in getExperiments

flozia:
# References:

Jira: [MNTOR-4908](https://mozilla-hub.atlassian.net/browse/MNTOR-4908)

# Description

PR https://github.com/mozilla/blurts-server/pull/6062 added imports to some of the cronjobs that are importing `next/headers`. This PR updates `getExperiments` to import `next/headers` dynamically so that it is safe to use in our cronjobs.

[MNTOR-4908]: https://mozilla-hub.atlassian.net/browse/MNTOR-4908?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/app/functions/server/getExperiments.ts
+++ b/src/app/functions/server/getExperiments.ts
@@ -2,7 +2,6 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
-import { headers } from "next/headers";
 import { captureException } from "@sentry/node";
 import { logger } from "./logging";
 import {
@@ -12,6 +11,7 @@ import {
 } from "../../../telemetry/generated/nimbus/experiments";
 import { ExperimentationId } from "./getExperimentationId";
 import { getEnabledFeatureFlags } from "../../../db/tables/featureFlags";
+import { loadNextHeaders } from "./loadNextHeaders";
 
 /**
  * After we removing the `CirrusV2` flag, we can return the full `ExperimentData`/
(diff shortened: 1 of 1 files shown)


src/app/functions/server/getExperiments.ts
```diff
@@ -59,10 +59,16 @@ export async function getExperiments(params: {
...
+  const nextHeaders = await loadNextHeaders();
+  let previewMode = false;
+  if (nextHeaders) {
+    const headersList = await nextHeaders.headers();
+    // Check if the Nimbus preview mode has been set by the middleware.
+    const nimbusPreviewMode = headersList.get("x-nimbus-preview-mode");
+    if (nimbusPreviewMode === "true") {
+      previewMode = true;
+    }
+  }
```
Vinnl, line 71: Suggestion (non-blocking): might be a little easier to read like this? Or maybe that's just me.

```suggestion
  const nextHeaders = await loadNextHeaders();
  const previewMode = (await nextHeaders?.headers())?.get("x-nimbus-preview-mode") === "true";
```

Vinnl (approved):
I think this is also the file I was wondering about last time if we didn't need one specifically for cronjob use? If so, I think this is a further point in favour of that, but still no blocker :)

src/app/functions/server/getExperiments.ts
flozia: I like that you version is more condensed, but I’ve opted for the current version so it is immediately clear that `previewMode` is `false` by default and that everything else is only noteworthy when `next/headers` is available.

flozia: … and good point regarding possibly needing to use a separate function for `getExperiments`. I’d like to change as little as possible for now to get `stage` back to a healthy state. I’ve filed MNTOR-4938 to revisit and introduce a way for us to catch these issues in CI.

PR #6122, opened 2025-09-17
MNTOR-4523: Re-apply React-Aria upgrade, fix broken location autocomplete

Vinnl:
# References:

Jira: MNTOR-4523
Figma:

# Description

This re-applies the update from https://github.com/mozilla/blurts-server/pull/6077, which caused the location autocomplete input to break: the content would clear after every character.

That stopped happening after adding `isNonModal` to the popover, but after that the suggestion dropdown would disappear as soon as you pushed Down. I'm not sure exactly why that was, but I could resolve it by removing the width calculation based on the input element - instead, the dropdown now is just as wide as its contents. This seemed fine to me.

I also added some tests that didn't end up catching this specific issue, but might still be useful if it occurs later for some other reason. And I enabled the `locationSuggestions` mock for Storybook ([which we can do now!](https://storybook.js.org/blog/next-generation-module-mocking/)), so that you can actually test the dropdown now in Storybook.

# How to test

Open the settings page or welcome flow in Storybook.

- [ ] I've added a unit test to test for potential regressions of this bug. - Not possible, depended on browser behaviour


src/app/components/client/ComboBox.test.tsx
```diff
@@ -45,6 +45,20 @@ it("shows suggestions when typing", async () => {
...
+  render(<ComposedTextComboBox />);
+
+  const comboBox = screen.getByRole("combobox");
+  await user.click(comboBox);
+  expect(comboBox).toHaveFocus();
+  await user.keyboard("T");
+  expect(comboBox).toHaveFocus();
+  await user.keyboard("u");
+  expect(comboBox).toHaveFocus();
+});
```
flozia, line 60: Praise: Nice addition to the test suite!

src/app/components/client/ComboBox.tsx
```diff
@@ -91,13 +91,14 @@ function ComboBox(
           popoverRef={popoverRef}
           state={state}
           triggerRef={inputRef}
+          isNonModal={true}
```
flozia, line 94: Nit: Just the prop would do.
```suggestion
          isNonModal
```

PR #6124, opened 2025-09-17
MNTOR-4708 - Update FAQ string in redesigned landing page

codemist:
# References:

Jira: MNTOR-4708

# Description
I missed this string because it was located in the redesigned landing page folder. I had only added the logic in the original version.
# Screenshot (if applicable)

Not applicable.

# How to test

### DIFF
--- a/src/app/(proper_react)/(redesign)/(authenticated)/user/(dashboard)/dashboard/[[...slug]]/page.tsx
+++ b/src/app/(proper_react)/(redesign)/(authenticated)/user/(dashboard)/dashboard/[[...slug]]/page.tsx
@@ -69,6 +69,10 @@ type Props = {
 };
 
 export default async function DashboardPage(props: Props) {
+  // Calling `connection` forces everything below it to be excluded from
+  // prerendering. Technically the page was already dynamic,
+  // but interestingly enough this change seemed to fix the cache-like
+  // behavior we saw in MNTOR-4908.
   await connection();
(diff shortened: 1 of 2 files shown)


src/app/(proper_react)/(redesign)/(public)/LandingViewRedesign/components/Faq.tsx
```diff
@@ -91,20 +96,37 @@ export const Faq = () => {
...
+              learn_more_link: (
+                <Button
+                  variant="link"
+                  href={CONST_URL_SUMO_MONITOR_PLUS}
+                  target="_blank"
+                />
+              ),
+            },
+          }),
+      /* c8 ignore stop */
```
flozia, line 129: Nit: We could also just toggle the string ID instead of duplicating the entire fragment if nothing else changes for those two versions.

flozia (approved):

src/app/(proper_react)/(redesign)/(public)/LandingViewRedesign/components/Faq.tsx
codemist: I think keeping the duplicated call here is actually clearer, since the original string requires a `vars` block. If we tried to toggle just the string ID, we'd also need to conditionally include or omit `vars`, which adds complexity. Duplicating the fragment call keeps each case self-contained and easier to read.

PR #6149, opened 2025-09-25
MNTOR-4981: Apply the chainsaw

Vinnl:
# References:

Jira: MNTOR-4981


src/apiMocks/mockData.ts
```diff
@@ -45,8 +44,7 @@ export type RandomScanResultOptions = Partial<{
   manually_resolved: boolean;
   broker_status: DataBrokerRemovalStatus;
   data_broker: string;
-  source: MoscaryData["ScanResult"]["source"];
```
codemist, line 48: Do we need to run the schema change first before this?

Vinnl: The [migration to change the schema](https://github.com/mozilla/blurts-server/pull/6149/files#diff-c7b3c535fff8b369098c95d5df98e761f378bfdf4e05ea7df6d3a46c794e331e) is part of this PR (and of the same commit that modifies `knex-tables`, so that it's easy to find).

src/app/(proper_react)/(redesign)/(authenticated)/user/(dashboard)/dashboard/View.tsx
```diff
@@ -190,12 +160,7 @@ export const View = (props: Props) => {
 
   const breachesDataArray = props.userBreaches.flat();
   const initialScanInProgress =
-    // MNTOR-4531: OneRep code paths will be phased out:
-    /* c8 ignore next 2 */
-    (isOneRepScan(adjustedScanData.scan)
-      ? adjustedScanData.scan.onerep_scan_status === "in_progress"
-      : adjustedScanData.scan?.status === "in_progress") &&
-    props.scanCount === 1;
+    adjustedScanData.scan?.onerep_scan_status === "in_progress";
```
rhelmer: Do we no longer need `props.scanCount === 1`?

rhelmer (approved):

src/app/(proper_react)/(redesign)/(authenticated)/user/(dashboard)/dashboard/View.tsx
Vinnl: Oh wow, very good find! I must have missed the closing bracket for the ternary, and thought the `&&` was part of the `:` branch. This would've been annoying to debug later. Fixed in c2ee90f8a.

PR #6150, opened 2025-09-26
fix: add input validation for production deployment workflow

mansaj:
# References:

Jira: MNTOR-5023

# Description
For the prod deploy workflow, add input validation for the original image tag and environment The validation checks ensure the image tag is a short SHA (7-12 alphanumeric characters) and that the environment is one of the allowed values (stage, prod, dev). Sanitized values are set as environment variables for safe usage in subsequent steps.

# Screenshot (if applicable)

Not applicable.

# How to test

# Checklist (Definition of Done)

- [ ] Localization strings (if needed) have been added.
- [ ] Commits in this PR are minimal and [have descriptive commit messages](https://chris.beams.io/posts/git-commit/).
- [ ] I've added or updated the relevant sections in readme and/or code comments
- [ ] I've added a unit test to test for potential regressions of this bug.
- [ ] If this PR implements a feature flag or experimentation, I've checked that it still works with the flag both on, and with the flag off.
- [ ] If this PR implements a feature flag or experimentation, the Ship Behind Feature Flag status in Jira has been set
- [ ] Product Owner accepted the User Story (demo of functionality completed) or waived the privilege.
- [ ] All acceptance criteria are met.
- [ ] Jira ticket has been updated (if needed) to match changes made during the development process.
- [ ] Jira ticket has been updated (if needed) with suggestions for QA when this PR is deployed to stage.

PR #6156, opened 2025-09-30
refactor: use SAFE_ENVIRONMENT variable for tagging Docker images

mansaj:
# References:

Jira: MNTOR-5023

# Description
Revert this change based on Frida's request: https://github.com/mozilla/blurts-server/pull/6150#discussion_r2387209655

# Screenshot (if applicable)

Not applicable.

# How to test

### DIFF
--- a/.github/workflows/production_deploy.yml
+++ b/.github/workflows/production_deploy.yml
@@ -33,6 +33,7 @@ jobs:
       GAR_REGISTRY: us-docker.pkg.dev # Define GAR registry hostname
       DOCKERHUB_IMAGE: mozilla/blurts-server # Define Docker Hub image name
       SAFE_IMAGE_TAG: ${{ inputs.originalImageTag }}
+      SAFE_ENVIRONMENT: ${{ inputs.environment }}
     steps:
       - name: Checkout Repository
         uses: actions/checkout@v5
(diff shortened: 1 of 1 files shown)


.github/workflows/production_deploy.yml
```diff
@@ -65,16 +66,16 @@ jobs:
         run: docker pull "${{ env.DOCKERHUB_IMAGE }}:${{ env.SAFE_IMAGE_TAG }}"
 
       - name: Retag Docker Hub image
-        run: docker tag "${{ env.DOCKERHUB_IMAGE }}:${{ env.SAFE_IMAGE_TAG }}" "${{ env.DOCKERHUB_IMAGE }}:${{ inputs.environment }}-${{ env.SAFE_IMAGE_TAG }}"
+        run: docker tag "${{ env.DOCKERHUB_IMAGE }}:${{ env.SAFE_IMAGE_TAG }}" "${{ env.DOCKERHUB_IMAGE }}:${{ env.SAFE_ENVIRONMENT }}-${{ env.SAFE_IMAGE_TAG }}"
```
fkiriakos07, line 69: is it necessary to use the {{variable}} when using environment variables, I think you can do `$SAFE_ENVIRONMENT`, I'm concerned by using the interpolation, the injected commands would be executed and not treated as strings

rhelmer (approved):

.github/workflows/production_deploy.yml
mansaj: @fkiriakos07 I think the `${{ }}` syntax is GitHub Action's expression syntax that gets evaluated before the shell command runs, so it's safe and is the standard way to reference environment variables in GitHub Actions workflows. If we try to use just `$SAFE_ENVIRONMENT` in the YAML file, I don't think GitHub Actions would recognize it as a variable reference - it would be treated as a literal string.

PR #6162, opened 2025-10-02
feat(local): add docker compose for easier local setup

kschelonka:
# References:

A quick docker-compose alternative to setting up postgres and pubsub manually.


.docker/pubsub/setup_pubsub.sh
```diff
@@ -0,0 +1,53 @@
+#!/bin/bash
```
rhelmer, line 1: lgtm, one suggestion from shellcheck (we should consider adding this to our linter setup):
```
% shellcheck .docker/pubsub/setup_pubsub.sh 

In .docker/pubsub/setup_pubsub.sh line 11:
gcloud beta emulators pubsub start --host-port=0.0.0.0:${PUBSUB_PORT} --project=${PROJECT_ID} &
                                                       ^------------^ SC2086 (info): Double quote to prevent globbing and word splitting.
                                                                                ^-----------^ SC2086 (info): Double quote to prevent globbing and word splitting.

Did you mean: 
gcloud beta emulators pubsub start --host-port=0.0.0.0:"${PUBSUB_PORT}" --project="${PROJECT_ID}" &

For more information:
  https://www.shellcheck.net/wiki/SC2086 -- Double quote to prevent globbing ...
```

.env.local.example
```diff
@@ -94,4 +94,6 @@ DATA_BROKER_REMOVAL_ESTIMATES_DATA=[]
 GCP_PUBSUB_PROJECT_ID=your-project-name
 GCP_PUBSUB_TOPIC_NAME=hibp-breaches
 GCP_PUBSUB_SUBSCRIPTION_NAME=hibp-cron
-PUBSUB_EMULATOR_HOST=localhost:8085
+PUBSUB_HOST=localhost
+PUBSBUB_PORT=8085
```
rhelmer:
```suggestion
PUBSUB_PORT=8085
```

.env.local.example
```diff
@@ -94,4 +94,6 @@ DATA_BROKER_REMOVAL_ESTIMATES_DATA=[]
...
+PUBSUB_EMULATOR_HOST="${PUBSUB_HOST}:${PUBSUB_PORT}"
```
rhelmer, line 99: Are you sure this gets expanded correctly? IIRC we're using `dotenv-flow` right now.

kschelonka: It worked on docker-compose when I ran it, but I couldn't verify that blurts was working fully

.env.local.example
kschelonka: ah ty I fixed this typo but I guess forgot to update commit

.env.local.example
kschelonka: I added a console log to print out the env vars and confirmed it was set correctly

PR #6183, opened 2025-10-14
chore(flags): remove EnableRemovalUnderMaintenanceStep flag

kschelonka:
# References:

Jira: MNTOR-5029
Jira: MNTOR-3886

# Description

Remove EnableRemovealUnderMaintenanceStep flag

I chose not to make breaking changes to function apis that seemed more general, although args did become unused (e.g. removing the enabledFeatureFlags argument). I can do that if preferred.

~I kept the RemovalUnderMaintenance view to avoid refactoring, but it has no logic other than to redirect to either `/` or `/user/dashboard` depending on login state.~ _edit: deleted this_

### DIFF
--- a/src/app/functions/server/getRelevantGuidedSteps.ts
+++ b/src/app/functions/server/getRelevantGuidedSteps.ts
@@ -141,7 +141,7 @@ function getStepWithStatus(
 export function isEligibleForStep(
   data: StepDeterminationData,
   stepId: StepLink["id"],
-  _enabledFeatureFlags?: FeatureFlagName[],
+  enabledFeatureFlags?: FeatureFlagName[],
 ): boolean {
   // Only premium users can see the manual data broker removal flow, once they have run a scan
   if (stepId === "Scan") {


src/app/(proper_react)/(redesign)/(authenticated)/user/(dashboard)/dashboard/fix/data-broker-profiles/removal-under-maintenance/page.tsx
```diff
@@ -4,89 +4,13 @@
 
 import { redirect } from "next/navigation";
```
codemist: I believe this page can be removed entirely?

codemist: Everything in `dashboard/fix/data-broker-profiles/removal-under-maintenance`, including assets/styling.

kschelonka: will do, ty!

src/app/(proper_react)/(redesign)/(authenticated)/user/(dashboard)/dashboard/fix/data-broker-profiles/removal-under-maintenance/RemovalUnderMaintenanceView.tsx
```diff
@@ -1,431 +0,0 @@
...
-          {l10n.getString("data-broker-removal-guide-button-back-to-exposures")}
-        </TelemetryButton>
-      </div>
-    </div>
-  );
-
-  return (
-    <FixView
-      subscriberEmails={props.subscriberEmails}
-      nextStep={() => moveToNextAvailableStep()}
```
codemist, line 420: I’m fairly confident that the `<FixView>` implementation here, along with removing all tests for `RemovalUnderMaintenance`, is what’s triggering the <100% test coverage regression.

I [opened a branch](https://github.com/mozilla/blurts-server/commit/cb2c0b62ed9d0e507b9734a33df12e2f8e3292cf) with the changes here for reference which shows 100% coverage.

Previously, we had a type guard on the `nextStep` prop to account for the navigation logic in `RemovalUnderMaintenance`, since that step wasn’t route-managed like the rest of the guided resolution flow. If you check our Storybook instances, you’ll notice the nav button is rendered using the first condition (Link) rather than Button (from TelemetryButton), which is unique to this step.

Wanted to flag in case this context helps explain the coverage drop. We definitely should've left more documentation around this type guard.

kschelonka: Thanks Kaitlyn, I saw the fix here is to remove this behavior then because it's no longer necessary. If that's an incorrect interpretation please let me know.

codemist (approved):

PR #6187, opened 2025-10-15
MNTOR-5071

mansaj:
# References:

Jira: MNTOR-5071
Figma:

# Description
We are getting this error when running a cronjob:
```
Error [ERR_MODULE_NOT_FOUND]: Cannot find module '/app/node_modules/next/headers' imported from /app/dist/scripts/cronjobs/emailBreachAlerts.js"
```

The issue seems to be introduced recently via an import:
```
emailBreachAlerts.tsx calls getExperiments()
getExperiments() calls getEnabledFeatureFlags()
getEnabledFeatureFlags() has an unprotected import("next/headers") (line 96)
```

getEnabledFeatureFlags() had await import("next/headers") with NO error handling. Here we wrap it so that it's "protected"

# Screenshot (if applicable)

Not applicable.

# How to test

### DIFF
--- a/src/db/tables/featureFlags.ts
+++ b/src/db/tables/featureFlags.ts
@@ -109,7 +109,7 @@ export async function getEnabledFeatureFlags(
       }
     } catch (error) {
       // next/headers not available (e.g., running in cronjob context)
-      console.debug(`next/headers not available: ${JSON.stringify(error)}`);
+      console.info(`next/headers not available: ${JSON.stringify(error)}`);
     }
   }
 


kschelonka (approved):

Vinnl (changes requested):
Florian added the `loadNextHeaders` function for exactly this purpose. See e.g. here: https://github.com/mozilla/blurts-server/blob/eed2d7fcddb3a54c83e0d160e17872f8ed23ef46/src/app/functions/server/getExperiments.ts#L62

We don't need to log an error - it is expected that we can't read request headers in cronjobs, so that would add noise to every log.

That said, it's also not clear to me that this would cause the error - [as far as I can see](https://github.com/mozilla/blurts-server/blob/eed2d7fcddb3a54c83e0d160e17872f8ed23ef46/src/db/tables/featureFlags.ts#L95-L96), the `import` is only called if `$E2E_TEST_ENV === "local"`, which I assume is not the case when the cron job is running? I also can't reproduce it when running the cronjob locally.

Edit: actually, it looks like that var [_is_ set](https://github.com/mozilla/blurts-server/blob/eed2d7fcddb3a54c83e0d160e17872f8ed23ef46/.env#L109) - that seems problematic. 

As an aside, while I can't reproduce the exact error, I get this:

```
Failed to add email notification to table:  Error: `headers` was called outside a request scope. Read more: https://nextjs.org/docs/messages/next-dynamic-api-wrong-context
```

Which is how I'd expect it to go wrong. The error we're seeing in the actual logs seems to imply that `next` isn't even installed, even though it's one of the dependencies? Or perhaps we're running an old version of Node there that can't do subpath imports?

PR #6190, opened 2025-10-16
Don't inspect request headers in cron jobs

Vinnl:
# References:

Jira: MNTOR-5071
Figma:

# Description

I could not reproduce the _exact_ issue that we're seeing in our logs, but I could trigger the code path that imports `next/headers` locally, and that shouldn't be possible.

Possibly the error in the logs are because `next` isn't installed for some reason, or we're using a version of Node that doesn't support importing from `next/headers` rather than just `next`?

# How to test

Follow [the readme instructions on pubsub](https://github.com/mozilla/blurts-server?tab=readme-ov-file#pubsub), but for the `curl` request, find a user's `primary_sha1`, take the first six characters and use that as the `hashPrefix`, and the rest as a `hashSuffix`.

Also, after you do this once, be sure to `DELETE FROM "email_notifications";` if you want to try again.

### DIFF
--- a/.env
+++ b/.env
@@ -106,7 +106,7 @@ ADMINS=
 MONTHLY_CRON_ENABLED=
 
 # Functional tests
-E2E_TEST_ENV=local
+E2E_TEST_ENV="not running e2e tests"
 E2E_TEST_SECRET=test-secret
 E2E_TEST_ACCOUNT_BASE_EMAIL=test-account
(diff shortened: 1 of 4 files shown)


Vinnl:
@kschelonka As you know, I agree :) That said, I think with that in mind too it can be helpful to merge this - when we get around to untangling it, the checks for `NEXT_RUNTIME` will help us identify which code paths go into the Next.js codebase and can be left out of the cron job/PubSub codebase. WDYT?

PR #6194, opened 2025-10-16
fix(breach-alerts): update cron logic for failure cases

kschelonka:
Move all notification logic into one try branch, and stop execution on the notification if any component fails;

Handle partial successes by keeping track of any failures in the message processing, and do not acknowledge unless message is fully successful; this will allow attempting to retry sends until the message goes to DLT

### DIFF
--- a/src/scripts/cronjobs/emailBreachAlerts.tsx
+++ b/src/scripts/cronjobs/emailBreachAlerts.tsx
@@ -320,6 +320,16 @@ export async function poll(
             // in between sending the email and marking as notified, otherwise
             // there is an unlikely edge case of sending multiple emails
             errorRecipients.push(breachedEmail);
+            Sentry.captureException(e, {
+              tags: {
+                breach_name: breachAlert.Name,
+                notification_stage: "email_send",
+              },
+              extra: {
+                subscriber_id: subscriberId,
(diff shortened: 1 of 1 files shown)


src/scripts/cronjobs/emailBreachAlerts.tsx
```diff
@@ -317,37 +312,32 @@ export async function poll(
...
-          // if this call ever fails, stop stop the script with an error
-          try {
             await markEmailAsNotified(subscriberId, breachId, recipientEmail);
-            // eslint-disable-next-line @typescript-eslint/no-explicit-any
-          } catch (e: any) {
-            console.error("Failed to mark email as notified: ", e);
-            throw new Error(e);
+            notifiedRecipients.push(breachedEmail);
+          } catch (e) {
+            logger.error("Failed to notify user of breach: ", e);
```
mansaj, line 318: nit(nonblocking): sentry err might also be helpful here.. Something like
```
Sentry.captureException(e, {
    tags: {
      breach_name: breachAlert.Name,
      notification_stage: 'email_send',
    },
    extra: {
      subscriber_id: subscriberId,
      breach_id: breachId,
    },
  });
```

PR #6201, opened 2025-10-20
MNTOR-5030: Remove unused `GA4SubscriptionEvents`

Vinnl:
# References:

Jira: MNTOR-5030
Figma:

# Description

This essentially reverts https://github.com/mozilla/blurts-server/pull/5101

Except it was done manually to avoid conflicts, leaves the migration in place and adds a new one to drop that table.

# How to test

Nothing should change. Migration should apply and rollback cleanly.

- [x] I've added a unit test to test for potential regressions of this bug. (None needed to be removed.)


src/db/migrations/20251020135047_drop_google_analytics_clients.js
```diff
@@ -0,0 +1,31 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+/**
+ * @param { import("knex").Knex } knex
+ * @returns { Promise<void> }
+ */
+export async function up(knex) {
+  await knex.schema.dropTable("google_analytics_clients");
```
Vinnl, line 10: Just calling out that this is irreversible data loss. I think in this case that's good and what we want (we'd rather not store and risk leaking data we're not using), but just in case I missed something.

rhelmer: We have backups just in case, but yes we've decided not to join this data so there's no point in collecting it.

Vinnl:
@rhelmer Not sure what you mean about `sendPingToGA` - I removed that too, because indeed it was no longer used anywhere?

PR #6209, opened 2025-10-22
feat(tests): infrastructure for integration tests

kschelonka:
Basic integration test scaffold for db and pubsub.

I updated the .env file to match important configs from .env.ci. I'm a little concerned about this somehow accidentally getting loaded into prod environment even though it seems just filled with dummy values. I didn't want to make too many changes to how configs are loaded but I think we would benefit from making it more explicit and centralized. I took out the PUBSUB_EMULATOR_HOST which is what i was concerned about and put it into jest.setup.ts for now.

# References:

Jira: Supports  MNTOR-5075
Figma:

### DIFF
--- a/jest.setup.ts
+++ b/jest.setup.ts
@@ -58,4 +58,5 @@ jest.mock("./src/envVars", () => {
 
 // Avoiding putting in the env file in case this gets loaded into prod
 // TODO: Centralize and streamline configuration for environments
-process.env.PUBSUB_EMULATOR_HOST = "localhost:8085";
+// mozilla-hub.atlassian.net/browse/MNTOR-5089
+https: process.env.PUBSUB_EMULATOR_HOST = "localhost:8085";
--- a/src/scripts/cronjobs/emailBreachAlerts.tsx
+++ b/src/scripts/cronjobs/emailBreachAlerts.tsx
@@ -384,6 +384,7 @@ async function getDataSummary(
 function createPubSubClient() {
   let options = {};
   // TODO - Consolidate configuration logic for this and other clients
+  // https://mozilla-hub.atlassian.net/browse/MNTOR-5089
   if (
     process.env.NODE_ENV === "development" ||
     process.env.NODE_ENV === "test"


src/db/knexfile.js
```diff
@@ -43,7 +43,7 @@ const TESTS_CONFIG = {
   connection: testConnectionObj,
 };
 
-let exportConfig = NODE_ENV === "tests" ? TESTS_CONFIG : RUNTIME_CONFIG;
+let exportConfig = NODE_ENV === "test" ? TESTS_CONFIG : RUNTIME_CONFIG;
```
kschelonka, line 46: fixed this typo

.env.ci
```diff
@@ -0,0 +1,180 @@
...
+PORT=6060
+NEXTAUTH_URL=http://localhost:6060
+NODE_ENV=development
+
+# 1: disables the dockerflow endpoints
+# see: https://github.com/mozilla-services/Dockerflow#containerized-app-requirements
+DISABLE_DOCKERFLOW=
+
+# Database server
+DATABASE_URL=postgres://blurts:blurts@localhost:5432/test-blurts
```
kschelonka, line 13: this connection string gets dynamically generated when setting up knex depending on the NODE_ENV (I think we should not, but that's how it is right now).  but it needs to be explicit for docker compose setup

.env
```diff
@@ -145,9 +145,12 @@ MONTHLY_SCANS_QUOTA=
...
 # GCP PubSub Project ID and subscription name
-GCP_PUBSUB_PROJECT_ID=
-GCP_PUBSUB_TOPIC_NAME=
-GCP_PUBSUB_SUBSCRIPTION_NAME=
+GCP_PUBSUB_PROJECT_ID=your-project-name
+GCP_PUBSUB_TOPIC_NAME=hibp-breaches
+GCP_PUBSUB_SUBSCRIPTION_NAME=hibp-cron
+PUBSUB_HOST=localhost
+PUBSUB_PORT=8085
+PUBSUB_EMULATOR_HOST="${PUBSUB_HOST}:${PUBSUB_PORT}"
```
kschelonka: this is what I'm afraid might get loaded into prod somehow, because the .env file is bundled in even though it's mostly placeholders

kschelonka: Moving this variable out and putting it into jest setup for now

jest.setup.ts
```diff
@@ -55,3 +55,7 @@ jest.mock("./src/envVars", () => {
     getEnvVarsOrThrow: () => process.env,
   };
 });
+
+// Avoiding putting in the env file in case this gets loaded into prod
+// TODO: Centralize and streamline configuration for environments
```
rhelmer, line 60: Please file a tech debt backlog ticket and link to it here

.env.ci
```diff
@@ -0,0 +1,180 @@
...
+AWS_ACCESS_KEY_ID=
+AWS_SECRET_ACCESS_KEY=
+AWS_REGION=
+S3_BUCKET=
+
+# Firefox Accounts OAuth
+FXA_SETTINGS_URL=https://accounts.stage.mozaws.net/settings
+
+OAUTH_CLIENT_ID=edd29a80019d61a1
+OAUTH_CLIENT_SECRET=get-this-from-groovecoder-or-fxmonitor-engineering
```
rhelmer, line 39: Could probably make this a more generic string, a little outdated 😆

PR #6214, opened 2025-10-23
MNTOR-5088 - Deprecate old settings page 

codemist:
# References:

Jira: MNTOR-5088

# Description
This PR removes the legacy `SettingsPage.test.tsx` file (and other old components like the old `EmailAddressAdder.tsx` and `SettingsPanelEditInfo.tsx`) and consolidates its test coverage into the redesigned settings test suite. All functionality previously validated is now tested through the new Storybook-based tests for each settings tab.

# Screenshot (if applicable)

Not applicable.

# How to test


src/app/(proper_react)/(redesign)/(authenticated)/user/(dashboard)/settings/DeleteAccountButton.tsx
```diff
@@ -36,8 +36,8 @@ export const DeleteAccountButton = (
         // It's currently unclear if and how we should mock our server action:
         /* c8 ignore next 8 */
         void props
-          .onDeleteAccount()
-          .then(() => {
+          ?.onDeleteAccount?.()
+          ?.then?.(() => {
```
codemist: I was running into several issues getting the tests for this component to pass, for whatever reason the linter was still running through the ignored section, even with the `/* c8 ignore */ `comment in place. To fix this safely, I added optional chaining to the onDeleteAccount call. I don't think it should change the logic, since When `onDeleteAccount` exists and returns a `Promise`, the `.then()` executes as before.

Vinnl: Linting is about ESLint finding potential syntactic issues, which doesn't really relate to coverage - which refers to code that is being executed while the tests are run.

I haven't looked at the changes in this PR in detail yet, but checking out the branch locally and removing the optional chaining, the error was about `.then` not existing [here](https://github.com/mozilla/blurts-server/blob/823778670cc3cb3ff62d52b0f6d0266835da2b71/src/app/(proper_react)/(redesign)/(authenticated)/user/(dashboard)/settings/DeleteAccountButton.tsx#L40), which implies that the `onDeleteAccount` callback is not returning a Promise (because `.then` is a method on a Promise).

Tracing where `onDeleteAccount` is defined from the test, I ended up [here](https://github.com/mozilla/blurts-server/blob/823778670cc3cb3ff62d52b0f6d0266835da2b71/src/app/(proper_react)/(redesign)/(authenticated)/user/(dashboard)/settings/stories/SettingsStoryWrapper.tsx#L46), which indeed defines a mock function that does not return any value, so not a Promise either. I could fix it as follows, no longer needing the optional chaining:

```diff
-  onDeleteAccount: fn().mockName("onDeleteAccount"),
+  onDeleteAccount: fn(() => new Promise(() => undefined)).mockName("onDeleteAccount"),
```

codemist: Ah yeah, this was tripping me up a bit until I realized there was a [mockedActions](https://github.com/mozilla/blurts-server/blob/823778670cc3cb3ff62d52b0f6d0266835da2b71/src/app/(proper_react)/(redesign)/(authenticated)/user/(dashboard)/settings/SettingsPage.test.tsx#L197-L202) in the old settings test file and also [mockedActions](https://github.com/mozilla/blurts-server/blob/823778670cc3cb3ff62d52b0f6d0266835da2b71/src/app/(proper_react)/(redesign)/(authenticated)/user/(dashboard)/settings/stories/SettingsStoryWrapper.tsx#L43-L48) in `SettingsStoryWrapper` for the redesigned settings page and new tests.

Some parts of the Settings page were still covered by the old suite, so I was mixing expectations without realizing it. I tried the suggested mocking approach and it ran fine, but TypeScript failed because `onDeleteAccount` was inferred as `Promise<unknown>`, which was causing the `actions` prop in the `SettingsView` to complain. In this [commit](https://github.com/mozilla/blurts-server/pull/6214/commits/e366bc3b3861e631cf7c0402e9bc68ae75570d84) I specified the `onDeleteAccount` type and that seemed to work. I'm confused why I had to do this in the first place though, shouldn't the `c8 ignore` in the `DeleteAccountButton` exclude that section from test coverage altogether?

Vinnl: Oh yeah, I think the tests for the old settings page rendered the React component directly, and for the new one we figured out how to get it in Storybook and use that to render the component. And so the mocks were specified in the tests or in the storybook, respectively. Obviously confusing :sweat_smile: (This is classic tech debt: we have to pay interest on our debt as time goes on, in the sense that removing the page now takes more work, because the context from the time got lost.)

I'm not quite sure exactly what error you got, but coverage and type errors are orthogonal things. Essentially, the coverage report wraps every function in another function that keeps track of how often it gets called, and then after the tests have run, the tallies of all those functions are collected to see what code has executed. TypeScript instead runs _before_ the code itself even runs, instead checking that the types that we specify, or that it can infer from the code, are consistently defined.

src/app/(proper_react)/(redesign)/(authenticated)/user/(dashboard)/settings/panels/SettingsPanelNotifications.tsx
```diff
@@ -56,9 +56,7 @@ export const NotificationsSettings = (props: NotificationSettingsProps) => {
 
   // Extract monthly report preference from the right column
   const monitorReportAllowed =
-    (hasPremium(props.user)
-      ? props.data?.monthly_monitor_report
-      : props.data?.monthly_monitor_report_free) ?? true;
```
codemist, line 61: I removed the `monthly_monitor_report_free` handling and tests because we no longer display monthly report settings for free users (MNTOR-4970). The feature was already disabled in the UI, so these props/tests were stale and could lead to confusion going forward.

Vinnl: Whoops, should've removed those when I removed the settings. Thanks for that!

src/app/(proper_react)/(redesign)/(authenticated)/user/(dashboard)/settings/stories/SettingsRedesign.stories.tsx
Vinnl, line 11:
```suggestion
  title: "Pages/Logged in/Settings",
```

To save a click in Storybook :)

Image

Edit: oh, GitHub appears confused, but I meant removing the `/Redesign` bit.

Vinnl, line 1: Can we drop the `-Redesign` bit from the filename now?

src/app/(proper_react)/(redesign)/(authenticated)/user/(dashboard)/settings/SettingsPageRedesign.test.tsx
Vinnl, line 1: Can we drop the `-Redesign` bit from the filename now?

src/app/(proper_react)/(redesign)/(authenticated)/user/(dashboard)/settings/panels/index.tsx
Vinnl, line 8: Can we drop the `-Redesign` bits from the file, component and type names?

src/app/(proper_react)/(redesign)/(authenticated)/user/(dashboard)/settings/stories/SettingsRedesign.stories.tsx
```diff
@@ -32,6 +44,20 @@ export const SettingsEditManageAccount: Story = {
   },
 };
 
+export const SettingsEditManageAccountPlus: Story = {
```
Vinnl, line 47: Praise: thanks for adding these additional permutations, they seem pretty relevant 😅

src/app/(proper_react)/(redesign)/(authenticated)/user/(dashboard)/settings/stories/SettingsRedesign.stories.tsx
```diff
@@ -20,6 +20,18 @@ const meta: Meta<typeof SettingsWrapper> = {
 export default meta;
 type Story = StoryObj<typeof SettingsWrapper>;
 
+export const SettingsNoDefaultTab: Story = {
```
Vinnl, line 23: Question: A real user should never encounter this permutation, am I understanding that correctly?

src/app/(proper_react)/(redesign)/(authenticated)/user/(dashboard)/settings/SettingsContent.tsx
```diff
@@ -79,6 +79,7 @@ function SettingsContent(props: SettingsProps) {
             selectedKey={activeTab}
             tabs={tabsData}
             orientation="vertical"
+            /*c8 ignore next 3 */
```
Vinnl, line 82: Suggestion/question: is this correct?


```suggestion
            // The `SidebarNavigationRedesign` is set to be deleted,
            // so this is not worth adding tests for:
            /*c8 ignore next 3 */
```

src/app/(proper_react)/(redesign)/(authenticated)/user/(dashboard)/settings/SettingsPageRedesign.test.tsx
```diff
@@ -44,10 +66,135 @@ jest.mock("../../../../../../hooks/locationSuggestions");
...
+
+    const tabListItemInitial = screen.getByRole("tab", {
+      name: "Edit your info",
+    });
+    expect(tabListItemInitial.getAttribute("aria-selected")).toBe("true");
+
+    const tabListItemNext = screen.getByRole("tab", {
+      name: "Set notifications",
+    });
+    // Wrap the click (state update) in act()
```
Vinnl, line 159: It's still weird to me that this is necessary, as `user.click` is supposed to handle this for you. But I suppose not a huge issue.

src/app/(proper_react)/(redesign)/(authenticated)/user/(dashboard)/settings/SettingsPageRedesign.test.tsx
```diff
@@ -1117,6 +1273,294 @@ describe("Settings page redesign", () => {
       const { container } = render(<ComposedStory />);
       expect(await axe(container)).toHaveNoViolations();
     }, 10_000);
+    it("renders the cancellation section for Plus users", () => {
```
Vinnl, line 1276: Nit 😅 

```suggestion

    it("renders the cancellation section for Plus users", () => {
```

src/app/(proper_react)/(redesign)/(authenticated)/user/(dashboard)/settings/SettingsPageRedesign.test.tsx
```diff
@@ -1128,5 +1572,385 @@ describe("Settings page redesign", () => {
...
+      const ComposedStory = composeStory(
+        SettingsEditNotifications,
+        SettingsMeta,
+      );
+      render(<ComposedStory />);
+      const monthlyMonitorReportBtn = screen.queryByLabelText(
+        "Monthly ⁨Monitor⁩ report",
+        { exact: false },
+      );
+      // The monthly email for free users is currently disabled; see MNTOR-4970.
```
Vinnl, line 1768: Praise: 👏

src/app/(proper_react)/(redesign)/(authenticated)/user/(dashboard)/settings/SettingsPageRedesign.test.tsx
Vinnl, line 1: Question: Am I correct in assuming that most of these tests were carried over from elsewhere, or should I review them more closely?

codemist: They were all carried over from the old settings page. I may have added one or two extra test cases actually, so generally more additive than reductive.

PR #6228, opened 2025-10-30
feat(fx breaches): fix breach alerts cron job in fx settings

kschelonka:
# References:

Jira: MNTOR-5024

# Description
Fixes the cron job which updates password breaches in firefox remote settings. My goal was to refactor so that it could be completely testable, but running the cron tests was blocked by jsdom environment issues (but this runs in node). I was able to implement mocked workarounds for logging and fetch in the remote settings client. But effectively mocking the logger in the cron methods required access to `setImmediate`, which is the whole reason the actual logger can't be used. So I just gave up on getting the tests to run there until [MNTOR-1880].

# How to test
- Unit tests
- Manual testing in dev environment (fill in appropriate environment variables with dev configuration and run locally).

Manual tests:
- [x] Single new breach uploads and sends review request
- [x] Multiple new breaches upload and send review request
- [x] No new breaches exits successfully without review request
- [x] Errors are forwarded to Sentry

[MNTOR-1880]: https://mozilla-hub.atlassian.net/browse/MNTOR-1880?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/scripts/cronjobs/updateBreachesInRemoteSettings/updateBreachesInRemoteSettings.ts
+++ b/src/scripts/cronjobs/updateBreachesInRemoteSettings/updateBreachesInRemoteSettings.ts
@@ -59,15 +59,13 @@ export async function main(parentLogger: Logger) {
 
   // Add handlers for ensuring all logs are sent before shutdown
   async function shutdown() {
-    return new Promise<void>(async (resolve) => {
-      logger.on("finish", () => {
-        resolve();
-      });
+    await new Promise<void>((resolve) => {
+      logger.on("finish", () => resolve());
       logger.end();
-      if (Sentry.isInitialized()) {
-        await Sentry.flush();
-      }
     });
+    if (Sentry.isInitialized()) {
+      await Sentry.flush();
+    }
   }
 
   process.on("SIGTERM", async () => {
--- a/src/test/helpers/mockLogger.ts
+++ b/src/test/helpers/mockLogger.ts
@@ -10,7 +10,7 @@ export function mockLogger() {
   return {
     info: jest.fn(),
     error: jest.fn(),
-    wran: jest.fn(),
+    warn: jest.fn(),
     debug: jest.fn(),
   } as unknown as Logger;
 }


src/app/functions/server/logging.ts
```diff
@@ -24,3 +26,22 @@ export const logger = createLogger({
     ? [getLoggingWinston()]
     : [new transports.Console()],
 });
+
+// Automatically capture logger error, warn and forward to Sentry
+// Avoids double-logging
```
kschelonka, line 31: make this a non-breaking change by keeping it additive. this requires node, should it be elsewhere? the import doesn't exist on `@sentry/next`

eslint.config.js
```diff
@@ -41,6 +41,7 @@ const config = [
       "dist",
       "coverage",
       "!.storybook",
+      "playwright-report/**",
```
kschelonka, line 44: was causing eslint local errors for me

src/test/helpers/mockLogger.ts
```diff
@@ -0,0 +1,16 @@
...
+
+import { Logger } from "winston";
+
+// Extremely minimal logger mock, works for most needs
+// After [MNTOR-1880] probably unnecessary
+export function mockLogger() {
+  return {
+    info: jest.fn(),
+    error: jest.fn(),
+    wran: jest.fn(),
```
mansaj: typo: warn

src/scripts/cronjobs/updateBreachesInRemoteSettings/updateBreachesInRemoteSettings.ts
```diff
@@ -0,0 +1,139 @@
...
+export async function main(parentLogger: Logger) {
+  const logger = parentLogger.child({ job: "updateBreachesInRemoteSettings" });
+
+  // Add handlers for ensuring all logs are sent before shutdown
+  async function shutdown() {
+    return new Promise<void>(async (resolve) => {
+      logger.on("finish", () => {
+        resolve();
+      });
+      logger.end();
```
mansaj, line 64: Event handler may not fire if end() completes synchronously? Also without timeout, could this hang (race condition)?

Something like:

```
async function shutdown() {
  return new Promise<void>((resolve) => {
    logger.on("finish", () => {
      resolve();
    });
    logger.end();
  }).then(async () => {
    if (Sentry.isInitialized()) {
      await Sentry.flush();
    }
  });
}
```

src/scripts/cronjobs/updateBreachesInRemoteSettings/updateBreachesInRemoteSettings.ts
```diff
@@ -0,0 +1,139 @@
...
+  });
+
+  logger.info("Breach data", {
+    existing: remoteBreaches.size,
+    new: newBreaches.length,
+  });
+
+  let failed = 0;
+
+  for (const breach of newBreaches) {
```
mansaj, line 104: This might not be an issue.. Processing breaches sequentially is safe, but what's the scale we are expecting here? Should we consider batching?

kschelonka: I did find a bulk API in the kinto docs. Personally I think the juice isn't worth the squeeze because we'd have to do more finicky processing for figuring out which requests were successful or not. There's a total of < 600 breaches in the lifetime of this app. The scale is based on how many breaches are discovered and reported by HIBP. Most of the time this job is run I expect there will be 0, and I'd be surprised if there was more than 1.

src/scripts/cronjobs/updateBreachesInRemoteSettings/updateBreachesInRemoteSettings.ts
kschelonka: when the 'finish' event is fired, that means that all logs have been flushed. this is adapted from the documentation: https://github.com/winstonjs/winston?tab=readme-ov-file#awaiting-logs-to-be-written-in-winston. the `end` method is what essentially schedules the finish event because it starts the flush.

But I think yes good call maybe Sentry could race, let me simplify this with await

ieasternyamessisuns1king-dotcom:
{"*.¡–ÍÈaster*¡*passÊE–¡.*"}@allinone.co-get/me?ÍÈaster‽

PR #6238, opened 2025-11-04
chore(dependabot): ignore faker major update (v10+)

kschelonka:
The newest major version of faker-js is ESM-only which is difficult to support with jest.

Since faker is a zero-dependency library, ignore major version upgrades for now (can consider the work to update if/when we need any functionality from the new version). Not having dependencies means this is pretty safe from a security perspective.

See https://github.com/faker-js/faker/issues/3606\#issuecomment-3233612736

I tried using the SWC config suggested in the discussion above, but couldn't get it to work properly. For us I think the juice isn't worth the squeeze at this time.


kschelonka:
Yeah I agree, it's definitely a bummer. Hopefully jest will keep up, or can eventually consider an alternative

PR #6205, opened 2025-10-21
MNTOR-5062 - Prevent bento items from opening a second redirect as a pop

codemist:
# References:

Jira: MNTOR-5062

# Description
This refactors the `AppPicker` component to use native `<a>` elements instead of triggering links via `ref.current?.click()` or `window.open()` inside the previously selected `onSelect`. This fixes Firefox’s popup blocking issue, since I believe why this popup error is occurring is due to synthetic click events are not treated as trusted user gestures.

I’m not entirely certain why the old pattern was used instead of placing links directly in anchor tags, my guess is it was likely for accessibility or input-handling reasons. This version seems to works as intended and removes the redirect popup issue (verified on Storybook), though additional QA testing on stage would be optimal.

# Screenshot (if applicable)

Not applicable.

# How to test

# Checklist (Definition of Done)

- [ ] Localization strings (if needed) have been added.
- [ ] Commits in this PR are minimal and [have descriptive commit messages](https://chris.beams.io/posts/git-commit/).
- [ ] I've added or updated the relevant sections in readme and/or code comments
- [ ] I've added a unit test to test for potential regressions of this bug.
- [ ] If this PR implements a feature flag or experimentation, I've checked that it still works with the flag both on, and with the flag off.
- [ ] If this PR implements a feature flag or experimentation, the Ship Behind Feature Flag status in Jira has been set
- [ ] Product Owner accepted the User Story (demo of functionality completed) or waived the privilege.
- [ ] All acceptance criteria are met.
- [ ] Jira ticket has been updated (if needed) to match changes made during the development process.
- [ ] Jira ticket has been updated (if needed) with suggestions for QA when this PR is deployed to stage.

PR #6239, opened 2025-11-05
MNTOR-5062 - Prevent redirect link in UserMenu to redirect as a popup

codemist:
# References:

Jira: MNTOR-5062
Figma:

# Description

# Screenshot (if applicable)

Not applicable.

# How to test


src/app/components/client/toolbar/UserMenu.tsx
```diff
@@ -61,43 +61,8 @@ export const UserMenu = (props: UserMenuProps) => {
     signout: "signout",
   };
 
-  const handleOnAction = (menuItemKey: Key) => {
-    switch (menuItemKey) {
-      case itemKeys.fxa:
-        recordTelemetry("ctaButton", "click", {
-          button_id: "manage_account_user_menu",
-        });
-        fxaItemRef.current?.click();
```
codemist, line 70: Similar to the fix in [this PR](https://github.com/mozilla/blurts-server/pull/6205), something about the synthetic `.click()` event here is triggering redirects as pop-ups. I removed the manual `handleOnAction` handler here and moved the telem events directly into the links.

Vinnl: Hmm, but this is not the same fix we ended up with in that PR? In that PR, the fix was to keep the `recordTelemetry` in the `handleOnAction`, but to add an `href` to the `<Item>`s, as per [the react-aria docs](https://react-spectrum.adobe.com/react-aria/useMenu.html#links).

codemist: Ah yeah, looking at it now this isn’t exactly the same fix as in that earlier PR. After digging into that link more closely, I realized that to fully adhere to the recommended pattern I needed to decouple a few things (like removing the redundant refs and shifting to `onAction` as the single source of truth).

That ended up being a bit more of a lift than I initially thought, since it required restructuring how props flow through `Item` to `MenuItem` and cleaning up nested interactive elements. The current implementation should now follow the correct React Aria approach.

So you’re right that my first pass only fixed the synthetic click, but the latest commits aligns with the React Aria pattern now.

src/app/components/client/toolbar/UserMenu.tsx
```diff
@@ -61,125 +54,107 @@ export const UserMenu = (props: UserMenuProps) => {
...
         >
-          <a
-            className={styles.menuItemCta}
-            href={CONST_URL_PLUS_CONTACT_SUPPORT}
-            ref={contactItemRef}
-            rel="noopener noreferrer"
-            target="_blank"
-            title={l10n.getString("user-menu-contact-tooltip")}
-          >
+          <span title={l10n.getString("user-menu-contact-tooltip")}>
```
codemist, line 126: While debugging this, I found that I couldn’t pass title directly to `<Item>`. React Aria treats it as part of the node’s metadata, which overrides the rendered children. To fix this, I wrapped the text and icon in a `<span>` instead, so the label remains visible while still supporting an accessible tooltip on hover.

src/app/components/client/toolbar/UserMenu.tsx
```diff
@@ -268,12 +243,42 @@ type MenuItemProps = {
...
+          aria-describedby={item.props.title}
+        >
+          {item.rendered}
+        </a>
+      ) : (
+        <button
+          {...menuItemProps}
+          className={styles.menuItemCta}
+          ref={ref as React.RefObject<HTMLButtonElement>}
+          aria-describedby={item.props.title}
```
Vinnl: As I understand it you shouldn't need to add `role="none"`, and the `href` and `aria-describedby` properties (and probably `rel` and `target` either) yourself, i.e. react-aria should be setting those? I'm not confident that setting them manually will set them to the right things. Could you elaborate what made you add them?

src/app/components/client/toolbar/UserMenu.test.tsx
```diff
@@ -41,6 +42,8 @@ it("opens and closes the user menu", async () => {
 });
 
 it("checks if the user menu items are interactive", async () => {
+  jest.spyOn(console, "error").mockImplementationOnce(() => undefined);
```
Vinnl, line 48: Could you add a comment describing why this is needed? It look suspicious, i.e. errors could have good reasons?

src/app/components/client/toolbar/UserMenu.test.tsx
```diff
@@ -53,40 +56,33 @@ it("checks if the user menu items are interactive", async () => {
   expect(menuItems.length).toBe(4);
 
   // FxA link
-  const fxaItem = screen.getByText("Manage your ⁨Mozilla account⁩");
+  const fxaItem = screen.getByRole("menuitem", {
```
Vinnl, line 62: I assume some new characters got introduced somehow? Was it just whitespace, or is something else going on?

src/app/components/client/toolbar/UserMenu.tsx
codemist: You're right - I was meant to remove `aria-describedby` (I was using it to test for tooltips) but that's now handled by the `title` attr in `span`.

src/app/components/client/toolbar/UserMenu.test.tsx
codemist: I usually get this warning whenever I add unicode control characters

src/app/components/client/toolbar/UserMenu.test.tsx
codemist: Added a comment here: https://github.com/mozilla/blurts-server/pull/6239/commits/29a10121efcd5a2ccd05ca74e7846e2f51504a1f

src/app/components/client/toolbar/UserMenu.tsx
```diff
@@ -268,12 +243,34 @@ type MenuItemProps = {
...
+          target={item.props.target}
+          className={styles.menuItemCta}
+        >
+          {item.rendered}
+        </a>
+      ) : (
+        <button {...menuItemProps} className={styles.menuItemCta}>
+          {item.rendered}
+        </button>
+      )}
```
codemist, line 278: @Vinnl Removed `rel="none"` and `rel` and `ref` in the rendered child element, since it's being handled by react aria. Thanks for the heads up on that note.

Vinnl: I think the `ref` _is_ necessary - currently I can't navigate the menu with the keyboard anymore.

Vinnl (approved):
**WARNING:** While I've approved it, there is still one blocker - keyboard navigation of the menu is broken now. I think you do need the `ref`s on the child `<a>` and `<button>`. But since that's an easy change, I've already approved it - if you could fix that and verify that keyboard navigation works, feel free to merge.

PR #6251, opened 2025-11-06
MNTOR-5040: Add banner

Vinnl:
# References:

Jira: MNTOR-5040
Figma: https://www.figma.com/design/lNoKMiYeMjhY7PaHj7m6yq/Content-design?node-id=2422-1257&m=dev

# Description

Note: there's two more things I need to check.

1. ~~Where this should appear - currently I've just assumed the dashboard. ([Slack](https://mozilla.slack.com/archives/C09HZQ5LM16/p1762347659011809?thread_ts=1761843058.417699&cid=C09HZQ5LM16))~~ [Confirmed.](https://mozilla.slack.com/archives/C09HZQ5LM16/p1762448595345659?thread_ts=1761843058.417699&cid=C09HZQ5LM16)
2. ~~Whether fetching the product data works - I haven't been able to subscribe locally yet, so I haven't been able to test it. ([Slack](https://mozilla.slack.com/archives/C04BLQVN151/p1762431482796329).)~~ Realised the specific subscription didn't matter, so changed to just `hasPremium`.

# How to test

Adjust the `_DATE` env var to put yourself in either the 29-day runup or afterwards. Alternatively, you might be able to adjust your system date. (This only works locally, since it needs to be the server date.)

- [ ] All acceptance criteria are met. - Still need a feature flag

### DIFF
--- a/.env
+++ b/.env
@@ -139,6 +139,7 @@ MOZILLA_VPN_LANDING_URL=https://www-dev.allizom.org/products/vpn
 # have seen data breach results before, will be able to see their known breaches
 # first:
 BROKER_SCAN_RELEASE_DATE=2024-02-06
+BROKER_SCAN_SHUTDOWN_DATE=2025-12-17
 
 MONTHLY_SUBSCRIBERS_QUOTA=
 MONTHLY_SCANS_QUOTA=
--- a/locales-pending/plus-shutdown.ftl
+++ b/locales-pending/plus-shutdown.ftl
@@ -1,0 +1,14 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+plus-shutdown-banner-button-learn-more = Learn more
+plus-shutdown-banner-button-close = Close
+plus-shutdown-banner-free-runup-heading = Update: { -brand-monitor } data broker scan results will be deleted in 30 days
+plus-shutdown-banner-free-runup-subheading = View your results ahead of the the { -brand-monitor-plus } shutdown.
+plus-shutdown-banner-free-shutdown-heading = Data broker scan results have been deleted, but { -brand-monitor } lives on
+plus-shutdown-banner-free-shutdown-subheading = You can still continue to access data breach monitoring to help keep your info safe online.
+plus-shutdown-banner-plus-runup-heading = Data broker scan results and in-progress removals will be deleted
+plus-shutdown-banner-plus-runup-subheading = View your dashboard before we say goodbye to { -brand-monitor-plus } for good on December 17, 2025.
+plus-shutdown-banner-plus-shutdown-heading = Data broker scan results have been deleted, but { -brand-monitor } lives on
+plus-shutdown-banner-plus-shutdown-subheading = Even though scan results and in-progress removals are gone, you can still access data breach monitoring, for free.
--- a/src/app/components/client/PlusShutdownBanner.tsx
+++ b/src/app/components/client/PlusShutdownBanner.tsx
@@ -1,0 +1,101 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+"use client";
+
+import styles from "./PlusShutdownBanner.module.scss";
+import { Button } from "./Button";
+import { CloseBtn } from "../server/Icons";
+import { useL10n } from "../../hooks/l10n";
+import { useLocalDismissal } from "../../hooks/useLocalDismissal";
+import { useHasRenderedClientSide } from "../../hooks/useHasRenderedClientSide";
+import { type ShutdownState } from "../../functions/server/getPlusShutdownState";
+import { FeatureFlagName } from "../../../db/tables/featureFlags";
+
+type Props = {
+  shutdownState: ShutdownState;
+  countryCode: string;
+  enabledFeatureFlags: FeatureFlagName[];
+};
+
+export const PlusShutdownBanner = (props: Props) => {
+  const l10n = useL10n();
+
+  const hasRenderedClientSide = useHasRenderedClientSide();
+  const localDismissal = useLocalDismissal(
+    `shutdown-banner-${props.shutdownState.hasPremium ? "plus" : "free"}-${props.shutdownState.currentMoment}`,
+  );
+
+  if (
+    !hasRenderedClientSide ||
+    !props.enabledFeatureFlags.includes("ShutdownBanner") ||
+    // Free users who haven't run a scan have no data that would be lost:
+    (!props.shutdownState.ranScan && !props.shutdownState.hasPremium) ||
+    props.shutdownState.currentMoment === "ye-olden-days" ||
+    localDismissal.isDismissed
+  ) {
+    return null;
+  }
+
+  let heading, subheading;
+  if (!props.shutdownState.hasPremium) {
+    if (props.shutdownState.currentMoment === "runup") {
+      heading = l10n.getString("plus-shutdown-banner-free-runup-heading");
+      subheading = l10n.getString("plus-shutdown-banner-free-runup-subheading");
+    } else {
+      heading = l10n.getString("plus-shutdown-banner-free-shutdown-heading");
+      subheading = l10n.getString(
+        "plus-shutdown-banner-free-shutdown-subheading",
+      );
+    }
+  } else {
+    if (props.shutdownState.currentMoment === "runup") {
+      heading = l10n.getString("plus-shutdown-banner-plus-runup-heading");
+      subheading = l10n.getString("plus-shutdown-banner-plus-runup-subheading");
+    } else {
+      heading = l10n.getString("plus-shutdown-banner-plus-shutdown-heading");
+      subheading = l10n.getString(
+        "plus-shutdown-banner-plus-shutdown-subheading",
+      );
+    }
+  }
+
+  return (
+    <aside className={styles.wrapper}>
+      <div className={styles.text}>
+        <h2>{heading}</h2>
+        <p>{subheading}</p>
+      </div>
+      <div className={`${styles.buttons} `}>
+        <Button
+          variant="primary"
+          href="https://support.mozilla.org/en-US/kb/monitor-plus-shutting-down"
+        >
+          {l10n.getString("plus-shutdown-banner-button-learn-more")}
+        </Button>
+        <Button
+          className={styles.button}
+          variant="tertiary"
+          onClick={() => {
+            localDismissal.dismiss();
+          }}
+        >
+          {l10n.getString("plus-shutdown-banner-button-close")}
+        </Button>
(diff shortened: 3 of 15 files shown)


src/app/functions/server/getCountryCode.ts
```diff
@@ -8,6 +8,9 @@ import { getTestClientRegionFromToken } from "./testCountryCodeToken";
 export function getCountryCode(
   headers: Awaited<ReturnType<typeof headersGetter>>,
 ): string {
+  // if (Math.random() > 0) {
+  //   return "us";
+  // }
```
kschelonka:
```suggestion
```

src/app/functions/server/getPlusShutdownState.ts
```diff
@@ -0,0 +1,36 @@
...
+
+const [shutdownYearString, shutdownMonthString, shutdownDayString] =
+  process.env.BROKER_SCAN_SHUTDOWN_DATE!.split("-");
+const shutdownTimestamp = Date.UTC(
+  Number.parseInt(shutdownYearString, 10),
+  Number.parseInt(shutdownMonthString, 10) - 1,
+  Number.parseInt(shutdownDayString, 10),
+);
+
+const parsedDaysRunUp = Number.parseInt(
```
kschelonka, line 23: please add a comment explaining what this means (I believe it's how many days before shutdown we start showing the banner)

src/app/functions/server/getPlusShutdownState.ts
```diff
@@ -0,0 +1,36 @@
...
+  10,
+);
+const daysRunUp = Number.isNaN(parsedDaysRunUp) ? 29 : parsedDaysRunUp;
+
+export function getPlusShutdownState(user: SubscriberRow): ShutdownState {
+  return {
+    daysRunUp: daysRunUp,
+    daysOut: Math.floor((shutdownTimestamp - Date.now()) / 1000 / 60 / 60 / 24),
+    hasPremium: hasPremium(user),
+    ranScan: typeof user.onerep_profile_id !== "undefined",
```
kschelonka, line 44: comment(blocking): this appears unused, please remove

src/app/components/client/PlusShutdownBanner.tsx
```diff
@@ -0,0 +1,107 @@
...
+type Props = {
+  shutdownState: ShutdownState;
+  countryCode: string;
+  enabledFeatureFlags: FeatureFlagName[];
+};
+
+export const PlusShutdownBanner = (props: Props) => {
+  const l10n = useL10n();
+
+  const currentMoment =
```
kschelonka: comment(non-blocking): personally I think it would be simpler if this banner type logic was moved to `getShutdownState`. So `getShutdownState` would return the banner type e.g.  `"runup" | "shutdown" | "none" // or null`, given the current date and shutdown config. Then it's easy to unit-test the core banner choice logic. `isPremium` could be passed into the Props instead

src/app/functions/server/getPlusShutdownState.ts
```diff
@@ -0,0 +1,36 @@
...
+const [shutdownYearString, shutdownMonthString, shutdownDayString] =
+  process.env.BROKER_SCAN_SHUTDOWN_DATE!.split("-");
+const shutdownTimestamp = Date.UTC(
+  Number.parseInt(shutdownYearString, 10),
+  Number.parseInt(shutdownMonthString, 10) - 1,
+  Number.parseInt(shutdownDayString, 10),
+);
+
+const parsedDaysRunUp = Number.parseInt(
+  process.env.BROKER_SCAN_SHUTDOWN_RUNUP_DAYS!,
```
kschelonka: comment(blocking): remove `!` as this value is not guaranteed to be defined

locales-pending/plus-shutdown.ftl
```diff
@@ -0,0 +1,14 @@
...
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+plus-shutdown-banner-button-learn-more = Learn more
+plus-shutdown-banner-button-close = Close
+plus-shutdown-banner-free-runup-heading = Update: { -brand-monitor } data broker scan results will be deleted in 30 days
+plus-shutdown-banner-free-runup-subheading = View your results ahead of the the { -brand-monitor-plus } shutdown.
+plus-shutdown-banner-free-shutdown-heading = Data broker scan results have been deleted, but { -brand-monitor } lives on
+plus-shutdown-banner-free-shutdown-subheading = You can still continue to access data breach monitoring to help keep your info safe online.
+plus-shutdown-banner-plus-runup-heading = Data broker scan results and in-progress removals will be deleted
+plus-shutdown-banner-plus-runup-subheading = View your dashboard before we say goodbye to { -brand-monitor-plus } for good on December 17, 2025.
```
codemist, line 12: Could we add a var here instead?

src/app/functions/server/getCountryCode.ts
Vinnl: Aargh. I'm switching to jj and am still getting used to the lack of staging area... Fixed.

(Force push is for the same reason too, going to find a better workflow for that too, sorry.)

src/app/functions/server/getPlusShutdownState.ts
Vinnl: Good call, added.

src/app/functions/server/getPlusShutdownState.ts
Vinnl: Ah good call, I added that based on mentions in the spec, but indeed none of the banners actually change what they look like based on it. Removed.

src/app/components/client/PlusShutdownBanner.tsx
Vinnl: I suppose I agree. I generally default to passing the raw data/source of truth around, and calculating derived data there where it's needed, but there's really no other use case for the raw data here, and I don't think there will be. Updated now.

src/app/functions/server/getPlusShutdownState.ts
Vinnl: Oh crap, really looking forward to picking up MNTOR-5089. Nice catch.

locales-pending/plus-shutdown.ftl
Vinnl: I debated that, but then I was worried about messing up the formatter, or doing something wrong with the date object, so I opted to hope for it either not changing, or us remembering to update this.

.env
```diff
@@ -122,6 +122,7 @@ FXA_SUBSCRIPTIONS_URL=https://accounts.stage.mozaws.net/subscriptions
 PREMIUM_PRODUCT_ID=prod_NErZh679W62lai
 PREMIUM_PLAN_ID_MONTHLY_US=price_1MUNq0Kb9q6OnNsL4BoJgepf
 PREMIUM_PLAN_ID_YEARLY_US=price_1NvqawKb9q6OnNsLRTnYrtrV
+PPP_PRODUCT_ID=prod_SFb8iVuZIOPREe
```
codemist: A lil lost here, why do we have a new product ID?

locales-pending/plus-shutdown.ftl
codemist: Ah okay. I figured it had something to do with the formatter - I passed a var in mine but that's manually released so it's different.

src/app/components/client/PlusShutdownBanner.tsx
```diff
@@ -0,0 +1,100 @@
...
+    if (props.shutdownState.currentMoment === "runup") {
+      heading = l10n.getString("plus-shutdown-banner-free-runup-heading");
+      subheading = l10n.getString("plus-shutdown-banner-free-runup-subheading");
+    } else {
+      heading = l10n.getString("plus-shutdown-banner-free-shutdown-heading");
+      subheading = l10n.getString(
+        "plus-shutdown-banner-free-shutdown-subheading",
+      );
+    }
+  } else {
```
codemist, line 52: Shouldn't this be only if they've run a scan? I'm seeing `ranScan` as a prop in `ShutdownState`

src/app/functions/server/getPlusShutdownState.ts
codemist: I've been trying to mess around with the run up/days out vars in the view where we're importing `<PlusShutdownBanner>`
```
shutdownState={{
          daysOut: 20,
          daysRunUp: 30,
          hasPremium: true,
        } as ShutdownState
}
```

I think I've got a better idea now of how the logic works by playing around with it. How are we planning to test this with QA? I'm not sure it's easy for them to just go in and mess around with the date set in our envs?

Vinnl: Yeah it's a good question, and I don't have a good answer. [In the ticket](https://mozilla-hub.atlassian.net/browse/MNTOR-5040?focusedCommentId=1171265) I suggested:

> Not sure what the best way is for QA to test. We could adjust the shutdown date on stage? Then QA could test just before and after that.

Still a bit of a hassle, required some coordination, and if they find issues it could impact the deadline, but I don't have any better ideas :/

.env
Vinnl: Oh, yes! Sorry, the requirements only gradually became clear while working on this ticket, and I added this when I thought I had to adjust the messaging to this specific plan, but that ended up not happening. Removed now, thanks for catching that. 255839c926c33e8767d605099022d03b1853a5f5

src/app/components/client/PlusShutdownBanner.tsx
Vinnl: Holy crap I can't remember the last time I made so many mistakes in a single PR :facepalm: I should've realised when Kat [pointed out](https://github.com/mozilla/blurts-server/pull/6251#discussion_r2501215758) that I never used that property. Added it back, and now free users who never ran a scan won't see the banner: 2b15d7a151b66ee53c91329a717735dc91ecc913

codemist (approved):
LGTM to me, circling back [to this comment](https://github.com/mozilla/blurts-server/pull/6251#discussion_r2513587651) we should coordinate with QA (and setting the date in the env) when they should be able to test this, and document that process to ensure that the automation works on the day of. Another thing I'm thinking is it might be nice to have an emergency PR queued (that we keep in do not merge state) that strips the timed logic and simply displays the banner as-is, in case anything goes wrong on the day.

src/app/functions/server/getPlusShutdownState.ts
```diff
@@ -0,0 +1,46 @@
...
+
+const today = new Date(Date.now());
+const todayTimestamp = Date.UTC(
+  today.getFullYear(),
+  today.getMonth(),
+  today.getDate(),
+);
+const daysOut = (shutdownTimestamp - todayTimestamp) / 1000 / 60 / 60 / 24;
+
+const currentMoment =
```
codemist, line 37: I plugged this into an LLM to sanity-check the behavior, and the recommendation was that anything that depends on today’s date must be recalculated per request. It seems to suggest that we can trying moving time-based logic into a `getCurrentMoment()` function that runs inside `getPlusShutdownState()`.

Vinnl: Oh nice, just seeing this now, but [that was indeed the issue](https://github.com/mozilla/blurts-server/pull/6284)! Looks like your LLM is better than mine, because my just suggested unrelated things.

PR #6255, opened 2025-11-08
fix(breach-alerts): address undefined fxa_uid in breach alerts email

kschelonka:
Ensure fxa_uid is defined before attempting to access session-specific methods that require it

[MNTOR-5097]

[MNTOR-5097]: https://mozilla-hub.atlassian.net/browse/MNTOR-5097?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/scripts/cronjobs/emailBreachAlerts.tsx
```diff
@@ -282,25 +282,33 @@ export async function poll(
...
-            // The unit tests are currently too complex for me to write
-            // a proper test for this, and I need to understand the code
-            // better to be able to refactor it to make it more amenable
-            // to simple tests. Hence, I don't have a test for this yet:
-            /* c8 ignore next 3 */
-            if (typeof recipient.onerep_profile_id === "number") {
-              await refreshStoredScanResults(recipient.onerep_profile_id);
+            const upSell =
+              // double eq checks both null and undefined
+              // eslint-disable-next-line eqeqeq
```
rhelmer: nit(non-blocking): I think it'd be cleaner to leave the eslint rule and explicitly check for both `null` and `undefined` if that's what you want, or maybe consider a utility function if you think you'll need it a lot that does this.

You might consider https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing while not a drop-in replacement, it does check both `null` and `undefined` (and falls back to a default so you'd have to rewrite this section)

src/scripts/cronjobs/emailBreachAlerts.tsx
```diff
@@ -282,25 +282,33 @@ export async function poll(
...
+              recipient.fxa_uid != null &&
+              isEligibleForPremium(assumedCountryCode) &&
+              !hasPremium(recipient);
+            let dataSummary: DashboardSummary | undefined = undefined;
+            if (upSell) {
+              // The unit tests are currently too complex for me to write
+              // a proper test for this, and I need to understand the code
+              // better to be able to refactor it to make it more amenable
+              // to simple tests. Hence, I don't have a test for this yet:
+              /* c8 ignore next 3 */
```
rhelmer: nit(non-blocking): I'd avoid first-person in comments, it's not immediately clear who wrote it.

src/scripts/cronjobs/emailBreachAlerts.tsx
```diff
@@ -282,25 +282,33 @@ export async function poll(
...
-              await refreshStoredScanResults(recipient.onerep_profile_id);
+            const upSell =
+              // double eq checks both null and undefined
+              // eslint-disable-next-line eqeqeq
+              recipient.fxa_uid != null &&
+              isEligibleForPremium(assumedCountryCode) &&
+              !hasPremium(recipient);
+            let dataSummary: DashboardSummary | undefined = undefined;
+            if (upSell) {
+              // The unit tests are currently too complex for me to write
```
rhelmer: Please file a ticket for this and a TODO comment with the link so we know when this is safe to remove.

rhelmer (changes requested):
Only blocking request is to add a link to ticket for the tests with a TODO, others are non-blocking suggestions

src/scripts/cronjobs/emailBreachAlerts.tsx
kschelonka: I'll just use truthiness here, but I've always configured the eslint rule to allow `== null` only and disallow any other `==` comparisons. My experience has been that `===` for nullish values is a great source of runtime bugs (when forgetting to check for null or undefined), and it's very annoying to have to check both when `==` is the same thing. 

What do you think about that?

src/scripts/cronjobs/emailBreachAlerts.tsx
kschelonka: I didn't write this comment, it's just indented. Removing in favor of ticket.

src/scripts/cronjobs/emailBreachAlerts.tsx
rhelmer: Thanks yeah tbh I am mostly just worried about the eslint comment giving readers in the future the wrong impression that it's some kind of hack/workaround etc.

kschelonka: agreed, better to standardize our rules and avoid exceptions. thank you!

PR #6263, opened 2025-11-11
MNTOR-5107: Fix invalid state with existing scans under new flag

Vinnl:
# References:

Jira: MNTOR-5107
Figma:

# Description

If I'm understanding the code correctly, we never really pointed users to the manual resolution page from the dashboard. I think you'd go through the "Upgrade" page, but that's gone with the flag on.

So this PR makes the manual remove view one that we directly point users to if they have unresolved scan results and the flag is on, and adds new conditions to do so when the user has scan results but is not eligible for Plus, which is something new that can happen with the flag on.

I also removed the main CTA at the bottom of the manual remove view, if the flag is on.

# How to test

The main one I can think of is to get a US user, make them do a scan, and then turn the flag on. You can try this with both a free and a Plus user.

But other than that, if you encounter any other weird states I'd like to hear of them.

- [ ] I've added a unit test to test for potential regressions of this bug. - I'd like to add some for the new state, but I'm running out of time. Maybe tomorrow.

### DIFF
--- a/src/app/components/client/exposure_card/ScanResultCard.tsx
+++ b/src/app/components/client/exposure_card/ScanResultCard.tsx
@@ -148,7 +148,10 @@ export const ScanResultCard = (props: ScanResultCardProps) => {
           },
         );
       case "new":
-        if (props.isOnManualRemovePage) {
+        if (
+          props.isOnManualRemovePage &&
+          !props.enabledFeatureFlags.includes("FreeOnly")
+        ) {
           return l10n.getFragment(
             "exposure-card-description-info-for-sale-action-needed-manual-fix-page",
             {


src/app/functions/server/getRelevantGuidedSteps.test.ts
```diff
@@ -674,7 +674,7 @@ describe("getNextGuidedStep", () => {
       ).toBe("Done");
     });
 
-    it("returns false when data brokers that are removal under maintenance are resolved", () => {
+    it("returns true when data brokers that are removal under maintenance are resolved", () => {
```
Vinnl, line 677: This looks suspicious, but I think this was an error introduced [here](https://github.com/mozilla/blurts-server/commit/27e3719a2766770790e28c759f8e6f791529724f#diff-464988eb813b76bdd54af40c7ef336467573f995af98ec46e458d3b18aadee97L695). This test description originally was "returns false when data brokers that are removal under maintenance are resolved, but the flag is off". The flag was removed, but that shouldn't mean that if a user has manually resolved a scan result, that this step shouldn't be marked as completed?

kschelonka: I'm a bit confused about that one. My interpretation of this is that removing the flag effectively "turned the flag off", so the test should remain false. Is that correct?

src/app/(proper_react)/(redesign)/(authenticated)/user/(dashboard)/dashboard/getUserDashboardState.tsx
```diff
@@ -321,6 +323,38 @@ export const getUserDashboardState = (
...
+     */
+    isMatchingContent(contentProps, {
+      isRelevantGuidedStep: relevantGuidedStep.id === "DataBrokerManualRemoval",
+      hasExposures: true,
+      hasUnresolvedBreaches: false,
+      hasUnresolvedBrokers: true,
+      isEligibleForFreeScan: false,
+      isEligibleForPremium: false,
+      isPremiumUser: false,
+      scanInProgress: false,
```
kschelonka, line 356: This does look like it covers the cases I was seeing (isPremiumUser=false and isEligibleForPremium=false). 

The original error I encountered came from [DashboardTopBannerContent](https://github.com/mozilla/blurts-server/blob/4cfdd65337b34a78f0443c24e474df8311b69fab/src/app/(proper_react)/(redesign)/(authenticated)/user/(dashboard)/dashboard/DashboardTopBanner/DashboardTopBannerContent.tsx#L700). Can you please add a test or fixture to ensure that the dashboard banner content doesn't fall to the default statement?

src/app/functions/server/getRelevantGuidedSteps.test.ts
Vinnl: OK, so I'm not too familiar with the manual resolution stuff, but [apparently](https://mozilla.slack.com/archives/C09HZQ5LM16/p1762966050678159?thread_ts=1762876613.917779&cid=C09HZQ5LM16) there were two manual resolution flows? And so even with the flag off, there should still remain another flow (that after this PR gets another entrypoint) that allows users to manually resolve scan results. In which case they should still be considered resolved even if the broker itself is under maintenance.

(That said, "removal under maintenance" is _also_ disabled and I think the flag will be removed, so in that sense this shouldn't have any practical effects anywhere.)

src/app/(proper_react)/(redesign)/(authenticated)/user/(dashboard)/dashboard/getUserDashboardState.tsx
Vinnl: Yes, I'll do that in a separate PR. I'll merge this for now so QA can test in the meantime.

Vinnl: https://github.com/mozilla/blurts-server/pull/6266

PR #6275, opened 2025-11-17
Update e2e tests for new flag

Vinnl:
This will fail against prod until the flag is flipped there as well. The PR runs also fail (and so will local runs), because the flag is off by default.

Edit: ah right, I can align the e2e test flags with stage. Updated the tests to remove the subscription tests and updating some selectors, but they seem flaky running locally, so I just pushed it and am hoping that it succeeds reliably in CI now, for PRs at least.

Edit 2: ah alright, figured out the flakiness issues. Will add some inline comments to clarify. (Netlify is failing because of the Cloudflare outage.)

### DIFF
--- a/src/app/components/client/SignInButton.tsx
+++ b/src/app/components/client/SignInButton.tsx
@@ -8,13 +8,15 @@ import { signIn, useSession } from "next-auth/react";
 import { useL10n } from "../../hooks/l10n";
 import { Button, ButtonProps } from "./Button";
 import { useTelemetry } from "../../hooks/useTelemetry";
+import { useHasRenderedClientSide } from "../../hooks/useHasRenderedClientSide";
 
 export const SignInButton = (props: ButtonProps) => {
   const l10n = useL10n();
   const session = useSession();
   const recordTelemetry = useTelemetry();
+  const hasRenderedClientSide = useHasRenderedClientSide();
 
-  if (typeof session.data?.user.email === "string") {
+  if (typeof session.data?.user.email === "string" || !hasRenderedClientSide) {
     return null;
   }
 


functional-tests/utils/fxa.ts
```diff
@@ -15,24 +15,12 @@ async function goToFxA(
...
   {
     countryCode,
-    isMobile,
   }: {
     countryCode?: string;
     isMobile?: boolean;
   },
 ) {
   await page.goto(`${getBaseTestEnvUrl()}/`);
-  if (
```
Vinnl, line 25: The US-specific landing page had a menu on mobile, the hitherto-ROW version doesn't.

src/app/components/client/SignInButton.tsx
```diff
@@ -8,13 +8,15 @@ import { signIn, useSession } from "next-auth/react";
...
+import { useHasRenderedClientSide } from "../../hooks/__mocks__/useHasRenderedClientSide";
 
 export const SignInButton = (props: ButtonProps) => {
   const l10n = useL10n();
   const session = useSession();
   const recordTelemetry = useTelemetry();
+  const hasRenderedClientSide = useHasRenderedClientSide();
 
-  if (typeof session.data?.user.email === "string") {
+  if (typeof session.data?.user.email === "string" || !hasRenderedClientSide) {
```
Vinnl, line 19: This is actually a change in behaviour, but when the page loads, client-side JS needs to execute to activate React and attach the `signIn` event handler. Clicking the button beforehand triggered the Next-Auth error page, which happened often locally.

(Another "fix" was for Playwright to wait before `networkidle` to click the Sign In button, but that is discouraged by the Playwright docs, and actual users can run into this error case as well if they click the button too fast. Thus, I'm now just hiding the button until it's functional.)

src/app/components/client/SignInButton.tsx
```diff
@@ -8,13 +8,15 @@ import { signIn, useSession } from "next-auth/react";
 import { useL10n } from "../../hooks/l10n";
 import { Button, ButtonProps } from "./Button";
 import { useTelemetry } from "../../hooks/useTelemetry";
+import { useHasRenderedClientSide } from "../../hooks/__mocks__/useHasRenderedClientSide";
```
kschelonka: comment(blocking): I think this import is incorrect (shouldn't use `__mocks__`)

Vinnl: 🤦 I was already surprised by how fast the rendering happened during my testing. I remember when I first started out as a programmer I was annoyed at our Java codebase that needs tens of imports everywhere, and thus couldn't really do without auto-imports, but also where I had no way to tell whether the auto-imported modules actually were the right ones. And now this is my life 🫠 

Long story short, fixed in 5013721eaba7b6c02b14fa5b28f51f68d43fee8f.

PR #6276, opened 2025-11-17
MNTOR-4586/recent authentication before deletion

rhelmer:
# References:

Jira: MNTOR-4586

# Description

Require recent authentication before deleting account via Settings menu.

# How to test

There is a unit test, if you want to test manually you can modify `RECENT_AUTH_WINDOW_MS` in `src/app/(proper_react)/(redesign)/(authenticated)/user/(dashboard)/settings/actions.ts` to something very short (e.g. 1 ms) and you should see a message "For your safety, please sign in again (within the last hour) before deleting your account" when trying to delete your account.

### DIFF
--- a/locales/en/settings.ftl
+++ b/locales/en/settings.ftl
@@ -58,6 +58,8 @@ settings-delete-monitor-free-account-dialog-cta-label = Delete account
 settings-delete-monitor-free-account-dialog-cancel-button-label = Never mind, take me back
 settings-delete-monitor-account-confirmation-toast-label-2 = Your { -brand-monitor } account is now deleted.
 settings-delete-monitor-account-confirmation-toast-dismiss-label = Dismiss
+settings-delete-account-recent-auth-required = For your safety, please sign in again (within the last hour) before deleting your account.
+settings-delete-account-error-generic = Something went wrong while trying to delete your account. Try again.
 
 ## Monthly Monitor Report
 
--- a/src/app/(proper_react)/(redesign)/(authenticated)/user/(dashboard)/settings/DeleteAccountButton.tsx
+++ b/src/app/(proper_react)/(redesign)/(authenticated)/user/(dashboard)/settings/DeleteAccountButton.tsx
@@ -9,6 +9,7 @@ import { signOut } from "next-auth/react";
 import { Button } from "../../../../../../components/client/Button";
 import { type onDeleteAccount } from "./actions";
 import { useTelemetry } from "../../../../../../hooks/useTelemetry";
+import { useL10n } from "../../../../../../hooks/l10n";
 
 /**
  * Since <Button> is a client component, the `onDeleteAccount` handler can only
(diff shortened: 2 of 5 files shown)


locales/en/settings.ftl
```diff
@@ -58,6 +58,8 @@ settings-delete-monitor-free-account-dialog-cta-label = Delete account
 settings-delete-monitor-free-account-dialog-cancel-button-label = Never mind, take me back
 settings-delete-monitor-account-confirmation-toast-label-2 = Your { -brand-monitor } account is now deleted.
 settings-delete-monitor-account-confirmation-toast-dismiss-label = Dismiss
+settings-delete-account-recent-auth-required = For your safety, please sign in again (within the last hour) before deleting your account.
+settings-delete-account-error-generic = Something went wrong while trying to delete your account. Try again.
```
Vinnl, line 62: Wordsmithing isn't my expertise, but "within the last hour" is confusing to me for something the user still has to do.

```suggestion
settings-delete-account-recent-auth-required = For your safety, please sign in again before deleting your account.
settings-delete-account-error-generic = Something went wrong while trying to delete your account. Please try again.
```

src/app/api/utils/auth.tsx
```diff
@@ -298,6 +301,9 @@ export const authOptions: AuthOptions = {
           }
         }
       }
+      if (token.authenticatedAt) {
+        session.authenticatedAt = new Date(token.authenticatedAt).toISOString();
```
Vinnl, line 305: Isn't it safer to just keep the UNIX timestamp here? I don't think we need to preserve the timezone for this use case, so it would be good to be able to be able to avoid date string parsing.

src/next-auth.d.ts
```diff
@@ -46,6 +46,7 @@ declare module "next-auth" {
   /** Session data available after deserialising the JWT */
   interface Session {
     error?: "RefreshAccessTokenError";
+    authenticatedAt?: ISO8601DateString;
```
Vinnl, line 49: As above, wouldn't it make sense to use a UNIX timestamp here as well?

src/app/(proper_react)/(redesign)/(authenticated)/user/(dashboard)/settings/DeleteAccountButton.tsx
```diff
@@ -21,29 +22,57 @@ export const DeleteAccountButton = (
...
+                l10n.getString("settings-delete-account-error-generic"),
+              );
+            });
+        }}
+      />
+      {errorMessage && (
+        <p
+          role="status"
+          style={{
+            marginTop: "0.5rem",
```
Vinnl, line 70: Probably doesn't matter for such a small and rarely visible surface, but we tend to avoid `margin`s in CSS because they hurt the composability of individual elements. Instead, we try to be consistent in having layout "flow down", e.g. having a flexbox parent container with a `gap`.

Again, probably not a biggie for this specific code, but good to know.

Vinnl (approved):
I guess this works, but UX-wise this leaves quite a burden on the user, where they have to know to manually sign out and back in again. Isn't it possible to provide them with a link that calls `signOut`, then redirects to the settings page, for which they'll be asked to log in again first? And then only show that link if their last signing was more than, say, 55 minutes ago?

locales/en/settings.ftl
flodolo: Does it mean "log in again within the next hour", or you "must have done the log in within the last hour"?

kschelonka: I don't think that the timing is particularly helpful to the user here. Why not just say, "For your safety, please sign in again before deleting your account." ?

PR #6284, opened 2025-11-19
MNTOR-5112: Control banner state using feature flags

Vinnl:
# References:

Jira: MNTOR-5112
Figma:

# Description

The previous code had a bug: the current date was determined in the module scope of `getPlusShutdownState.ts`, which meant that it only ran the first time the function was imported.

In other words, if this runs on a long-running pod, and at runtime we hit one of the transition dates (to runup or shutdown), then that code will have a current date of yesterday in memory, and thus not properly transition. And because we're running multiple pods that can be restarted, you could be redirected to a different pod on every refresh and thus inconsistently hit different banner states.

I could've fixed that by just pulling the date calculation code into the `getPlusShutdownState` function, but I figured it's best to minimise risk and just accept that we'll have to do some manual work on the specific dates, by flipping feature flags instead.

# How to test

Flip the relevant flags and view your dashboard with a US user (free and Plus).

### DIFF
--- a/src/app/functions/server/getPlusShutdownState.test.ts
+++ b/src/app/functions/server/getPlusShutdownState.test.ts
@@ -4,6 +4,8 @@
 
 import { it, expect } from "@jest/globals";
 import { SubscriberRow } from "knex/types/tables";
+import { getPlusShutdownState } from "./getPlusShutdownState";
+import { FeatureFlagName } from "../../../db/tables/featureFlags";
 
 const mockedNormalSubscriber = {
   fxa_profile_json: {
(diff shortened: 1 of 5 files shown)


src/app/functions/server/getPlusShutdownState.test.ts
```diff
@@ -16,51 +18,36 @@ const mockedPlusSubscriber = {
...
+  ]);
+
+  expect(state.currentMoment).toBe("shutdown");
+});
 
-  const state = getPlusShutdownState(mockedNormalSubscriber);
+it("marks today as a post-shutdown day if both the PostShutdownBanner *and* the ShutdownBanner feature flags are set", async () => {
+  const state = getPlusShutdownState(mockedNormalSubscriber, [
+    "PostShutdownBanner",
+    "ShutdownBanner",
```
codemist, line 49: In this scenario, are we simply enabling `PostShutdownBanner` when the time comes, and it's not a requirement to disable the `ShutdownBanner` flag? Or is `ShutdownBanner` the overall general flag for the banner?

src/app/functions/server/getPlusShutdownState.test.ts
```diff
@@ -16,51 +18,36 @@ const mockedPlusSubscriber = {
...
-  process.env.BROKER_SCAN_SHUTDOWN_RUNUP_DAYS = "42";
-  // Note: we import this function dynamically so it picks up the new env vars:
-  const getPlusShutdownState = (await import("./getPlusShutdownState"))
-    .getPlusShutdownState;
-
-  const state = getPlusShutdownState(mockedNormalSubscriber);
+it("does not mark today as a runup day if both shutdown feature flags are disabled", async () => {
+  const state = getPlusShutdownState(mockedNormalSubscriber, [
+    "NotShutdownBanner" as FeatureFlagName,
+    "NotPostShutdownBanner" as FeatureFlagName,
```
codemist, line 24: Curious why these mock flags need to be passed, can it just be left empty?

src/app/functions/server/getPlusShutdownState.test.ts
```diff
@@ -70,7 +57,7 @@ it("recognises a free user", async () => {
   const getPlusShutdownState = (await import("./getPlusShutdownState"))
     .getPlusShutdownState;
```
codemist: ^ Just read the comment above this about picking up the new env vars, since we're retiring the timed logic around `BROKER_SCAN_SHUTDOWN_DATE` is this still necessary?

src/app/functions/server/getPlusShutdownState.test.ts
Vinnl: Exactly, enabling `PostShutdownBanner` is sufficient. `ShutdownBanner` is already enabled though, and doesn't need to be disabled. (Though it can safely be disabled afterwards.)

src/app/functions/server/getPlusShutdownState.test.ts
Vinnl: Yeah I figured I'd be explicit that I'm not passing in these feature flags - that didn't feel as clear if I had just passed in an empty array. `[]` would have had the same effect though.

src/app/functions/server/getPlusShutdownState.test.ts
Vinnl: Oops no, forgot to update that, will fix. Thanks for catching.

PR #6287, opened 2025-11-20
fix(breachAlerts): update cron job to autoscaling service

kschelonka:
# References:

Jira: 
* MNTOR-5070
* MNTOR-5081
* MNTOR-5108

Required for: https://mozilla-hub.atlassian.net/browse/SVCSE-2984

# Description
Updates the breach alerts email so that it can be compatible with autoscaling up and down based on queue size rather than run as a cron job. This required major pattern changes to how the job was invoked.

I refactored the job so that the pubsub subscription handling logic was separated from the business logic/processing for each message. I also updated the job to use DI pattern for easier testing. Much of the core message processing logic was lifted, with some streamlining around db checks. Since this was such a big overhaul I included some low-hanging fruit (removing premium upsell [5108] and removing the nimbus integration [5081]). 

All the methods are unit-tested except for the main `job` entrypoint, which I had originally tried to cover with an integration test but couldn't run in `jsdom` environment because pubsub client requires `setImmediate`, etc. I tested this locally successfully and will do a dev deployment. 

**Other Changes**
* Narrow type requirements on object args where appropriate (makes fake data generation easier and the data contract explicit)
* Random assortment of unit tests to pass the coverage requirement (removing imports must have uncovered the test gaps)
* Added integration test coverage for DB methods, but these won't be factored into the test coverage until 5117.


src/db/redis/client.ts
```diff
@@ -2,18 +2,18 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
-import { logger } from "@sentry/utils";
+import { logger } from "@sentry/core";
```
kschelonka, line 5: this was deprecated, recommended to use `@sentry/core`

src/db/redis/client.ts
```diff
@@ -2,18 +2,18 @@
...
+import MockRedis from "ioredis-mock";
 
 export const REDIS_ALL_BREACHES_KEY = "breaches";
 export const REDIS_ALL_DATA_BROKERS_KEY = "dataBrokers";
 
 let singleton: Redis;
 export const redisClient = () => {
   if (process.env.REDIS_URL?.includes("redis.mock")) {
-    singleton = createRedisMockInstance();
+    singleton = new MockRedis();
```
kschelonka, line 16: this was a one-liner in `utils-mock` file, which I moved here for clarity and to eliminate needing another test for that file

src/db/redis/client.test.ts
```diff
@@ -0,0 +1,59 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import { createRedisInstance } from "./util";
```
kschelonka, line 5: added due to coverage complaint (probably from removing this import in the test files?)

src/app/functions/universal/mock.test.ts
```diff
@@ -0,0 +1,37 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import { isUsingMockHIBPEndpoint, isUsingMockONEREPEndpoint } from "./mock";
```
kschelonka, line 5: added due to coverage complaint (no changes to the files)

src/app/functions/l10n/parseMarkup.test.ts
```diff
@@ -0,0 +1,13 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import { parseMarkup } from "./parseMarkup";
+
+describe("parseMarkup", () => {
```
kschelonka, line 7: added due to coverage complaint (no changes to method)

src/app/functions/universal/mock.ts
```diff
@@ -3,11 +3,11 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
 export function isUsingMockHIBPEndpoint() {
-  return process.env.HIBP_KANON_API_ROOT?.includes("api/mock") as boolean;
+  return !!process.env.HIBP_KANON_API_ROOT?.includes("api/mock");
```
kschelonka, line 6: this was returning undefined if HIBP_KANON_API_ROOT was unset; `!!` is idiomatic boolean conversion

src/apiMocks/mockData.ts
```diff
@@ -233,7 +233,7 @@ export function createRandomHibpListing(
     >,
     Description: faker.lorem.sentence(),
     Domain: faker.internet.domainName(),
-    Id: faker.number.int(),
+    Id: faker.number.int({ max: 2147483647 }),
```
Vinnl, line 236: Did you encounter a problem with >32bit numbers?

src/db/models/BreachNotificationSubscriber.ts
```diff
@@ -0,0 +1,57 @@
...
+
+// Covered by integration tests but report is not combined
+// TODO: MNTOR-5117
+/* c8 ignore start */
+export type BreachNotificationSubscriber = {
+  subscriber_id: SubscriberRow["id"];
+  all_emails_to_primary: SubscriberRow["all_emails_to_primary"];
+  primary_email: SubscriberRow["primary_email"];
+  breached_email: string;
+  notification_email: string;
```
Vinnl: Nit, if we want to be really explicit:


```suggestion
  breached_email: SubscriberRow["primary_email"] | EmailAddressRow["email"];
  notification_email: SubscriberRow["primary_email"] | EmailAddressRow["email"];
```

src/db/redis/client.ts
Vinnl: Heh nice, always in favour of fewer "utils" :)

src/db/models/BreachNotificationSubscriber.ts
```diff
@@ -0,0 +1,57 @@
...
+  signup_language: SubscriberRow["signup_language"];
+  fxa_profile_json: SubscriberRow["fxa_profile_json"];
+  fxa_uid: SubscriberRow["fxa_uid"];
+  onerep_profile_id: SubscriberRow["onerep_profile_id"];
+};
+
+export async function breachNotificationSubscribersByHashes(
```
Vinnl: Nit: In the rest of the codebase function names generally start with a verb, so could change to be consistent.


```suggestion
export async function getBreachNotificationSubscribersByHashes(
```

src/db/tables/email_notifications.ts
```diff
@@ -2,27 +2,25 @@
...
-  const res = await knex.transaction((trx) => {
-    return trx("email_notifications")
-      .forUpdate()
-      .select("subscriber_id")
-      .where("notified", true)
-      .andWhere("breach_id", breachId);
-  });
-
-  return res.map((row) => row.subscriber_id);
+export async function subscriberNotifiedForBreach(
```
Vinnl: Nit: as elsewhere, function name could start with a verb for consistency.


```suggestion
export async function isSubscriberNotifiedForBreach(
```

src/db/tables/email_notifications.ts
```diff
@@ -46,40 +46,29 @@ async function addEmailNotification(newNotification: NewNotification) {
     notification_type: newNotification.notificationType,
   };
 
-  const res = await knex.transaction((trx) => {
```
Vinnl, line 49: Thought: I don't have the context and relevant expertise, but I'll note that Rob explicitly added this transaction: https://github.com/mozilla/blurts-server/commit/241b4cf30c46c494524bf48606d4de49ca23e4ab#diff-614d0ef3a3274ac35a230ed9df0ce0a5d3d7e4d2c9fcec9ec084e9e796627968R112

(The accompanying ticket does make it sound like that was to solve scaling issues, possibly when this all still ran in the same process as the website, so very possible that it indeed is no longer necessary in the new setup.)

src/constants.ts
```diff
@@ -65,3 +65,4 @@ export const CONST_DATA_BROKER_PROFILE_DETAIL_LIMITS = {
 } as const;
 export const CONST_MAX_SCAN_RESULTS_PER_BROKER = 3 as const;
 export const CONST_CIRRUS_V2_PATHNAME = "v2/features";
+export const BREACH_ALERT_UTM_CAMPAIGN_ID = "breach-alert";
```
Vinnl: Nit: if we're going to move all our campaign IDs (and other UTM parameters?) to constants, then maybe we want to sort them together lexicographically?


```suggestion
export const UTM_CAMPAIGN_ID_BREACH_ALERT = "breach-alert";
```

src/utils/hibp.ts
```diff
@@ -130,6 +130,9 @@ export type HibpGetBreachesResponse = Array<{
   PwnCount: number;
 }>;
 
+// This is purely a wrapper to avoid passing
+// an argument and does not require unit tests itself
```
Vinnl, line 134: Of course it would automatically be covered if any of the functions calling it were :)

src/scripts/cronjobs/emailBreachAlerts/index.ts
```diff
@@ -0,0 +1,66 @@
...
+import { breachNotificationSubscribersByHashes } from "../../../db/models/BreachNotificationSubscriber";
+import * as NotificationsRepo from "../../../db/tables/email_notifications";
+import { job } from "./emailBreachAlerts";
+
+start();
+
+async function start() {
+  const projectId = process.env.GCP_PUBSUB_PROJECT_ID;
+  const subscription = process.env.GCP_PUBSUB_SUBSCRIPTION_NAME;
+  sentryLogger.info("Info", {
```
Vinnl: Question: I'm not super familiar with Sentry logging, but does it add more details on what this log is? Just cause `"Info"` isn't very specific 😅

src/scripts/cronjobs/emailBreachAlerts/index.ts
```diff
@@ -0,0 +1,66 @@
...
+    messageFnOpts: [
+      sentryLogger,
+      getAllBreachesFromDb,
+      { findByHashes: breachNotificationSubscribersByHashes },
+      NotificationsRepo,
+      sendEmail,
+      Sentry,
+    ],
+    jobLogger: sentryLogger,
+    Sentry,
```
Vinnl, line 59: Thought: I'm not necessarily opposed to this, but especially things like `Sentry` and `sendEmail` feel like they're only there for tests, in which case they'd be somewhat redundant with `jest.mock`, and possibly confusion - when do we use which method of mocking?

src/scripts/cronjobs/emailBreachAlerts/emailBreachAlerts.tsx
```diff
@@ -0,0 +1,296 @@
...
+  Sentry: typeof Sentry;
+};
+
+// Main entry for breach alerts email consumer
+// Initializes required pubsub client and kicks off subscription handler
+// This can be covered by integration tests but only in node
+// (not jsdom) environment, due to pubsub requiring setImmediate etc.
+// TODO: [MNTOR-1880]
+/* c8 ignore start */
+export function job(config: EmailBreachAlertsJobConfig) {
```
Vinnl: Nit: could also be `runJob` for consistency.

src/scripts/cronjobs/emailBreachAlerts/emailBreachAlerts.tsx
```diff
@@ -0,0 +1,296 @@
...
+    ...localConfig,
+  });
+  const subscription = pubsub.subscription(config.gcp.subscription, {
+    flowControl: {
+      // Maximum messages that may be in progress at once
+      maxMessages: 10,
+    },
+  });
+  const messageFn = async (message: Message) => {
+    return await Sentry.withIsolationScope(async () => {
```
Vinnl, line 285: TIL about Sentry's isolation scopes!

src/scripts/cronjobs/emailBreachAlerts/subscriptionHandler.ts
```diff
@@ -0,0 +1,101 @@
...
+import type { Logger } from "winston";
+
+export type MessageSummary = {
+  success: boolean;
+  skipped: number;
+  notified: number;
+  errors: number;
+};
+
+type messageFn = (message: Message) => Promise<MessageSummary>;
```
Vinnl: Nit: the rest of the codebase capitalises the first letters of types, so could change for consistency. (But it's a nit; if it was important we should've added a lint rule.)

src/scripts/cronjobs/emailBreachAlerts/emailBreachAlerts.tsx
```diff
@@ -0,0 +1,296 @@
...
+ * @param subs DB Repository for subscribers
+ * @param notifications DB Repository for email_notifications
+ * @param sendEmail send email method
+ * @param sentry optional initialized Sentry object
+ * @returns Promise<MessageS
+ */
+export async function breachMessageHandler(
+  message: Message,
+  logger: Logger,
+  breachProvider: () => Promise<HibpLikeDbBreach[]>,
```
Vinnl, line 114: Suggestion: one downside of DI is that I can't just Cmd+Click to get an idea of what the (typical) implementation looks like. In this case, I looked up two callsites above to find that in practice this gets fulfilled by `getAllBreachesFromDb`. For someone somewhat familiar with the codebase, it could at least be helpful to use that name here too? The fact that it's theoretically more generic seems less helpful to know than the runtime implementation.

src/scripts/cronjobs/emailBreachAlerts/emailBreachAlerts.tsx
```diff
@@ -0,0 +1,296 @@
...
+  if (shouldNotify) {
+    return { shouldNotify };
+  }
+  const reasons = Object.entries(doNotNotifyRules)
+    .filter(([, condition]) => condition === true)
+    .map(([reason]) => reason);
+  return {
+    shouldNotify,
+    reasons,
+  };
```
Vinnl, line 92: Praise: holy guacamole this is so much more readable.

src/scripts/cronjobs/emailBreachAlerts/emailBreachAlerts.tsx
```diff
@@ -0,0 +1,296 @@
...
+
+  // Build list of hashes and pull any matching email
+  // addresses from our user database
+  const reqHashPrefix = data.hashPrefix.toLowerCase();
+  const hashes = data.hashSuffixes.map(
+    (suffix) => reqHashPrefix + suffix.toLowerCase(),
+  );
+  const recipients = await subs.findByHashes(hashes);
+
+  logger.info("notification", {
```
Vinnl, line 159: Question: same as elsewhere, is `"notification"` descriptive enough?

src/scripts/cronjobs/emailBreachAlerts/emailBreachAlerts.tsx
```diff
@@ -0,0 +1,296 @@
...
+    breachAlertName: breachAlert.Name,
+    length: recipients.length,
+  });
+
+  let errorRecipients = 0;
+  let skipped = 0;
+  let notified = 0;
+  for (const recipient of recipients) {
+    // Skip if user does not want to be notified
+    if (recipient.all_emails_to_primary === null) {
```
Vinnl, line 170: Note for other reviews: this is confusing (already the case in the existing code), but apparently `true` and `false` doesn't just determine what email address the user wants to receive notifications on, but also that they want to receive a notification in the first place by virtue of not being `null`.

src/scripts/cronjobs/emailBreachAlerts/emailBreachAlerts.tsx
```diff
@@ -0,0 +1,296 @@
...
+      logger.info("Instant breach alerts disabled, skipping subscriber", {
+        subscriber_id: recipient.subscriber_id,
+      });
+      skipped += 1;
+      continue;
+    }
+    // Skip if user has been notified already for this breach
+    // (regardless of which email address specifically)
+    // TODO: MNTOR-5807
```
Vinnl, line 179: Question: This ticket doesn't load for me, is it the right one? Maybe the comment could include a short description of the todo anyway?

src/scripts/cronjobs/emailBreachAlerts/subscriptionHandler.spec.ts
```diff
@@ -0,0 +1,150 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import type { Logger } from "winston";
+import { MockProcess } from "../../../test/helpers/mockProcess";
+import { mockMessage, MockSubscription } from "../../../test/helpers/pubsub";
+import { mockLogger } from "../../../test/helpers/mockLogger";
+import { MessageSummary, SubscriptionHandler } from "./subscriptionHandler"; // <-- adjust the path
```
Vinnl: Question: is this comment still relevant? (If so, not clear to me what "adjust" means.)

src/scripts/cronjobs/emailBreachAlerts/subscriptionHandler.spec.ts
```diff
@@ -0,0 +1,150 @@
...
+  let sub: MockSubscription;
+  let proc: MockProcess;
+  let subCloseSpy: jest.SpyInstance<void, []>;
+
+  beforeEach(() => {
+    logger = mockLogger();
+    subCloseSpy = jest.spyOn(MockSubscription.prototype, "close");
+    sub = new MockSubscription();
+    proc = new MockProcess();
+    jest.clearAllMocks();
```
Vinnl: This should theoretically be redundant: https://github.com/mozilla/blurts-server/blob/7521902ee1e28880ff4707714fb72974df660b65/jest.config.cjs#L29

Likewise in the other tests.


(If it's not, that's something to look into I suppose.)

```suggestion
```

src/scripts/cronjobs/emailBreachAlerts/subscriptionHandler.test.ts
```diff
@@ -0,0 +1,150 @@
...
+      hashPrefix: "AA",
+      hashSuffixes: ["11"],
+    });
+
+    sub.emitMessage(msg);
+
+    // allow event loop to flush
+    // no setImmediate in jsdom, so just wait
+    // a short while instead
+    await setTimeout(50);
```
Vinnl: Question: Would this also work? I imagine these might start to add up, and timeouts also reek of flakiness.


```suggestion
    await Promise.resolve();
```

src/scripts/cronjobs/emailBreachAlerts/subscriptionHandler.test.ts
Vinnl, line 1: Nit: All our other unit tests are named `.test.ts`. (I don't have a preference, but it's neat to be able to e.g. grep for all of them easily.)

src/scripts/cronjobs/emailBreachAlerts/subscriptionHandler.test.ts
```diff
@@ -0,0 +1,150 @@
...
+  it("logs error thrown by subscription object", async () => {
+    startHandler(async () => ({
+      success: false,
+      errors: 1,
+      notified: 0,
+      skipped: 0,
+    }));
+    sub.emitError(new Error("nope"));
+    await setTimeout(50);
+    expect(logger.error).toHaveBeenCalled();
```
Vinnl: Suggestion: Don't know all the params off the top of my head, but:

```suggestion
    expect(logger.error).toHaveBeenCalledWith("nope");
```

src/scripts/cronjobs/emailBreachAlerts/subscriptionHandler.test.ts
```diff
@@ -0,0 +1,150 @@
...
+      hashPrefix: "AA",
+      hashSuffixes: ["11"],
+    });
+
+    sub.emitMessage(msg);
+    await setTimeout(50);
+
+    expect(msg.nack).toHaveBeenCalled();
+  });
+  it.each(["SIGTERM" as const, "SIGINT" as const])(
```
Vinnl, line 95: Praise: these tests are neat.

src/test/helpers/pubsub.ts
```diff
@@ -0,0 +1,47 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import { EventEmitter } from "events";
+import type { Message } from "@google-cloud/pubsub";
+
+export function mockMessage<T extends object>(payload: T) {
```
Vinnl: Nit: I don't think this generic does anything, since it's not included in the return type :)

src/scripts/cronjobs/emailBreachAlerts/emailBreachAlerts.spec.ts
```diff
@@ -0,0 +1,254 @@
...
+
+    expect(res).toEqual({ success: true, errors: 0, notified: 2, skipped: 1 });
+    expect(subs.findByHashes).toHaveBeenCalledTimes(1);
+    expect(sendEmail).toHaveBeenCalledTimes(2);
+    expect(notifications.addEmailNotification).toHaveBeenCalledTimes(2);
+    expect(notifications.markEmailAsNotified).toHaveBeenCalledTimes(2);
+  });
+
+  it("skips if breach is not notifiable", async () => {
+    const breach = mockBreach({ Name: defaultBreachName, IsVerified: false }); // not notifiable
```
Vinnl, line 78: Suggestion: to be thorough, also check on other reasons for not notifying? (empty domains, fabricated, spam)

src/scripts/cronjobs/emailBreachAlerts/emailBreachAlerts.spec.ts
```diff
@@ -0,0 +1,254 @@
...
+      findByHashes: jest
+        .fn()
+        .mockResolvedValue([mockSubscriber({ all_emails_to_primary: true })]),
+    };
+    const notifications = {
+      subscriberNotifiedForBreach: jest.fn().mockResolvedValue(false),
+      addEmailNotification: jest.fn().mockResolvedValue(undefined),
+      markEmailAsNotified: jest.fn(),
+    };
+    const sendEmail = jest.fn().mockRejectedValue(new Error("I'm sorry Dave"));
```
Vinnl, line 185: Image

src/db/models/BreachNotificationSubscriber.integration.ts
```diff
@@ -0,0 +1,168 @@
...
+    const chaffEmails = insertedChaff.flatMap((subscriber) =>
+      Array.from(Array(faker.number.int({ max: 20 })).keys()).map((_) =>
+        seeds.emails(subscriber.id),
+      ),
+    );
+    await conn("email_addresses").insert(chaffEmails);
+  });
+  afterEach(async () => {
+    await conn.raw(`TRUNCATE TABLE subscribers CASCADE`);
+    await conn.raw(`TRUNCATE TABLE email_addresses CASCADE`);
```
Vinnl, line 36: Question: I'm not familiar with the integration test setup, but does this affect my local database, i.e. the one that I use when running the app locally, or is there a test-specific DB? Otherwise this could be somewhat annoying.

src/db/models/BreachNotificationSubscriber.integration.ts
```diff
@@ -0,0 +1,168 @@
...
+        }),
+        expect.objectContaining({
+          subscriber_id: insertedSubscriber.id,
+          breached_email: insertedEmails[0].email,
+          primary_email: insertedSubscriber.primary_email,
+        }),
+      ]),
+    );
+  });
+  it("returns primary email only if no secondary emails match", async () => {
```
Vinnl, line 66: Question: From this and the next test description, it's not clear to me what's supposed to happen when both match...

src/db/tables/email_notifications.integration.ts
```diff
@@ -0,0 +1,151 @@
...
+      const actual = await addEmailNotification(data);
+      expect(actual).toMatchObject({
+        subscriber_id: data.subscriberId,
+        breach_id: data.breachId,
+        notified: data.notified,
+        email: data.email,
+        notification_type: data.notificationType,
+      });
+    });
+    it("does nothing and does not throw if row already exists", async () => {
```
Vinnl: Suggestion: I suppose it's kind of implicit already, but given the test description I'd expect a `not.toThrows()` in here?

src/db/tables/email_notifications.integration.ts
```diff
@@ -0,0 +1,151 @@
...
+      );
+      await markEmailAsNotified(subscriber.id, insertedBreach.id, email);
+      const roundTrip = await conn("email_notifications")
+        .where("subscriber_id", subscriber.id)
+        .andWhere("breach_id", insertedBreach.id)
+        .andWhere("email", email)
+        .first();
+      expect(roundTrip).toMatchObject({ notified: true });
+    });
+    it("does not throw error if row does not exist", async () => {
```
Vinnl: As above re: `not.toThrows()`.

src/db/models/BreachNotificationSubscriber.integration.ts
kschelonka: There's a test-specific db (it's not obvious, but the connection string gets modified depending on your NODE_ENV when the knex module is loaded)

src/db/tables/email_notifications.integration.ts
kschelonka: I removed "does not throw" from the description and just described return statement instead

src/db/tables/email_notifications.integration.ts
kschelonka: It's tested that it doesn't throw by asserting the return value; I updated the description to reference return vs. throw

src/db/models/BreachNotificationSubscriber.integration.ts
kschelonka: When there are multiple matches it returns them all, which is covered in the test above. I just wanted to cover "edge" cases of only matching one table or the other

src/scripts/cronjobs/emailBreachAlerts/emailBreachAlerts.tsx
kschelonka: Sure, I can add some implementation guidance along with the interface

src/scripts/cronjobs/emailBreachAlerts/emailBreachAlerts.tsx
kschelonka: Yes, I can add a comment to make that more clear

src/db/tables/email_notifications.ts
kschelonka: All statements in postgres are executed in a transaction, which is why I removed it to be more succinct (this is a single insert statement). See more here: https://www.postgresql.org/docs/18/tutorial-transactions.html

forUpdate is used incorrectly here, since this is an insert statement and not a select statement. Postgres should ignore it, but better to reduce code noise. See https://www.postgresql.org/docs/current/sql-select.html

src/scripts/cronjobs/emailBreachAlerts/emailBreachAlerts.tsx
kschelonka: It's a tradeoff, this is the logging statement that existed before so I was keeping it consistent. Do you want me to make it more specific?

src/scripts/cronjobs/emailBreachAlerts/emailBreachAlerts.tsx
kschelonka: Can you see this? https://mozilla-hub.atlassian.net/browse/MNTOR-5087

kschelonka: the comment describes the bug, if you have feedback on making it clearer I appreciate it

src/scripts/cronjobs/emailBreachAlerts/index.ts
kschelonka: ah I forgot to remove this

src/scripts/cronjobs/emailBreachAlerts/index.ts
kschelonka: Let me know if I'm understanding correctly -- the question is why to use DI when it seems like module mocks solve the same problem?

DI does make testing easier, but the main benefits are the explicit dependencies making it easier to understand and reuse. If we didn't include Sentry or sendEmail, you'd have to read through all the files to know that they were dependencies.

As a side-effect it's easier to stub or monkeypatch the dependencies in tests. jest.mock can still be used with DI, it's just brittle to things like module name/path changes so I tend to avoid it

src/scripts/cronjobs/emailBreachAlerts/subscriptionHandler.test.ts
kschelonka: I will fix, ty

kschelonka: the functional tests are called .spec.ts, is it ok to change those?

src/scripts/cronjobs/emailBreachAlerts/subscriptionHandler.spec.ts
kschelonka: nope, fixing

src/constants.ts
kschelonka: Updating but since it's imported throughout multiple files I'm doing it locally (vs. apply suggestion)

src/utils/hibp.ts
kschelonka: haha yeah, just trying to follow c8 ignore guidelines

src/scripts/cronjobs/emailBreachAlerts/subscriptionHandler.test.ts
kschelonka: Yeah appreciate this pushback. I don't know the pubsub internals very well, but I'm more confident that all the tasks are flushed if waiting on a macrotask like setTimeout/setImmediate vs. Promise.resolve. I'll do setTimeout(0) to mimic setImmediate better and not pile up waits

src/apiMocks/mockData.ts
Vinnl: Haha, I guess I'm mostly interested in learning what the problem was, if it's not too much work to explain (but fine if it is) :)

src/scripts/cronjobs/emailBreachAlerts/index.ts
Vinnl: Right, I guess it's not clear to me when not knowing on the outside that Sentry and sendEmail are dependencies is a problem, other than in tests? Since on the flip side, outside of tests it makes call sites noisier.

And I guess maybe the larger question is: is this a pattern we should apply more widely? And if not, or if not immediately, does the difference in patterns across the codebase make it harder for future contributors to navigate, i.e. do we get the worst of both worlds?

(I don't necessarily think this is a blocker (which is why I already approved the PR), but I would lean towards applying it consistently in one go or not at all.)

src/scripts/cronjobs/emailBreachAlerts/emailBreachAlerts.tsx
Vinnl: Ah I wasn't aware. Well, you have just as much context as any reader, so only if you're confident the additional specificity is correct makes it easier to understand for people perusing the logs :)

src/scripts/cronjobs/emailBreachAlerts/emailBreachAlerts.tsx
Vinnl: Ah yes, then my feedback would be this :P 

```suggestion
    // TODO: MNTOR-5087
```

src/scripts/cronjobs/emailBreachAlerts/subscriptionHandler.test.ts
Vinnl: Oh hmm funny. I guess it's helpful to be able to distinguish between those and unit tests, though it's a bit of an obscure distinguishing feature. Which is to say, I don't have a preference :P

src/db/models/BreachNotificationSubscriber.integration.ts
Vinnl: Ah I'd missed [that addition](https://github.com/mozilla/blurts-server/blob/6f1dc900a41a96cb2e8a240dd85e9ec17226bedf/src/db/knexfile.js#L37-L46), thanks!

PR #6307, opened 2025-11-24
Describe how to test `sharp` dependency updates

Vinnl:
I didn't add the dependency, but I think this is about right.


Vinnl:
WTF, why didn't you let GitHub repoint it to `main` Dependabot?

PR #6313, opened 2025-11-25
feat(breaches): handle when breach has not yet been ingested

kschelonka:
# References:

Jira: MNTOR-5100

# Description
    feat(breaches): handle when breach has not yet been ingested
    
    Breach notifications can be sent before the breach
    data has been updated in our system (previously only
    via cron job), potentially resulting in not sending
    breach alert emails until the next breach sync.
    
    This adds new services for syncing breaches if
    the emailBreachAlerts worker encounters a breach
    that it does not have data for. The sync service
    uses a locking mechanism to ensure that we don't
    overload HIBP with rapid concurrent requests
    (our workers scale horizontally and process messages
    rapidly as they are sent).


eslint.config.js
```diff
@@ -87,7 +87,7 @@ const config = [
       ],
       // For some reason `eqeqeq` is not in the recommended set, but we try to
       // avoid implicit type casting, cause that’s where bugs lurk:
-      eqeqeq: "error",
+      eqeqeq: ["error", "always", { null: "ignore" }],
```
kschelonka, line 90: allow `== null` only, otherwise `===`

src/services/BreachDataService.ts
```diff
@@ -0,0 +1,184 @@
...
+      DataClasses: breach.data_classes,
+      IsVerified: breach.is_verified,
+      IsFabricated: breach.is_fabricated,
+      IsSensitive: breach.is_sensitive,
+      IsRetired: breach.is_retired,
+      IsSpamList: breach.is_spam_list,
+      IsMalware: breach.is_malware,
+      FaviconUrl: breach.favicon_url,
+    } as HibpLikeDbBreach;
+  }
```
Vinnl: We might want to extract this out into a separate function, and re-use it elsewhere, e.g. here: https://github.com/mozilla/blurts-server/blob/a659d778ba96a6807d8728b9272fb07a9dfd972c/src/utils/hibp.ts#L265-L288

src/services/BreachDataService.ts
```diff
@@ -0,0 +1,184 @@
...
+import { BreachRow } from "knex/types/tables";
+import { Logger } from "winston";
+
+type BreachDataServiceOpts = {
+  negTtlSec: number;
+};
+
+interface BreachRepo {
+  getBreaches(): Promise<BreachRow[]>;
+}
```
Vinnl: Question: any reason to not just pass the function in directly, rather than wrapping it in an object with a single property?


```suggestion
type AllBreachesFetcher = () => Promise<BreachRow[]>;
```

src/services/BreachDataService.ts
```diff
@@ -0,0 +1,184 @@
...
+
+const DefaultBreachServiceOpts = {
```
Vinnl: Nit: not a type, so we don't usually uppercase these. And adding a type annotation helps autocomplete for future contributors.

```suggestion
const defaultBreachServiceOpts: BreachDataServiceOpts = {
```

src/services/BreachDataService.ts
```diff
@@ -0,0 +1,184 @@
...
+  negTtlSec: 60 * 5,
+};
+
+/**
+ * Data service for handling breach lookups.
+ * Automatically checks cache, updates using read-through
+ * strategy on cache miss, and refreshes stale data (with
+ * lock to prevent multiple concurrent refreshes).
+ */
+export class BreachDataService {
```
Vinnl: Question: any reason this is wrapped in a class? Feels like maybe some heavy ceremony around what is essentially a single function `getBreach`, with a couple of functions and variables that could also just stay in its closure (the module scope)?

That's a pattern that's also common elsewhere in this codebase (see `/src/app/functions`, though I'm not sure why that's under `/app/`) - hence there not being a `services` directory yet.

src/services/BreachSyncService.ts
```diff
@@ -0,0 +1,146 @@
...
+
+const DefaultSyncOptions: SyncOptions = {
+  minFreshMs: 5 * 60 * 1000, // 5 min
+  lockTtlMs: 10 * 60 * 1000, // 10 min
+  waiterTimeoutSec: 10,
+};
+
+export interface IBreachSyncService {
+  syncBreaches: () => Promise<void>;
+}
```
Vinnl: Question: Like `BreachDataService`, this class+interface feels like a lot of ceremony around a single function, that doesn't really match existing patterns already present in this codebase - is there a good reason to adopt a different convention here?

src/services/BreachSyncService.ts
```diff
@@ -0,0 +1,146 @@
...
+  getBreaches(): Promise<BreachRow[]>;
+}
+
+type SyncOptions = {
+  minFreshMs: number;
+  lockTtlMs: number;
+  waiterTimeoutSec: number;
+};
+
+const DefaultSyncOptions: SyncOptions = {
```
Vinnl: Nit: not a type or class, so usually lowercased.


```suggestion
const defaultSyncOptions: SyncOptions = {
```

src/services/BreachSyncService.ts
```diff
@@ -0,0 +1,146 @@
...
+    private readonly redis: Redis,
+    private readonly fetchBreaches: () => Promise<HibpGetBreachesResponse>,
+    private readonly repo: BreachRepo,
+    opts?: Partial<SyncOptions>,
+  ) {
+    const optsWithDefault = {
+      ...DefaultSyncOptions,
+      ...opts,
+    };
+    if (BREACHES_EXPIRY_SECONDS * 1000 < optsWithDefault.minFreshMs) {
```
Vinnl: Praise: thanks for including the time units in these variable names, that's a bunch of bugs saved right there 😅

src/services/BreachSyncService.ts
```diff
@@ -0,0 +1,146 @@
...
+    }
+    this.opts = optsWithDefault;
+  }
+
+  private waitKey(syncId: string) {
+    return `${REDIS_ALL_BREACHES_KEY}:refresh:wait:${syncId}`;
+  }
+
+  private async isFresh(): Promise<boolean> {
+    const lastSynced = Number((await this.redis.get(this.lastSyncKey)) || 0);
```
Vinnl: Nit: not expecting a boolean. (I don't really mind individual harmless cases like this that much, but I like to encourage avoiding the implicit type coercion by default, just for that one case where it _does_ matter.)


```suggestion
    const lastSynced = Number((await this.redis.get(this.lastSyncKey)) ?? 0);
```

src/services/BreachSyncService.ts
```diff
@@ -0,0 +1,146 @@
...
+          this.opts.waiterTimeoutSec,
+        );
+      }
+      return;
+    }
+    // Stale and unblocked, ok to refresh
+    try {
+      // Fetch breaches, save to DB, and update cache
+      const breaches = await this.fetchBreaches();
+      validateBreaches(breaches);
```
Vinnl: Suggestion: Shouldn't we move `validateBreaches` into `hibpApiFetch`, replacing the type assertion that it has now? We're kind of throwing away the advantages of TypeScript now.

src/scripts/cronjobs/emailBreachAlerts/emailBreachAlerts.test.ts
```diff
@@ -74,13 +76,9 @@ describe("breachMessageHandler", () => {
     expect(notifications.markEmailAsNotified).toHaveBeenCalledTimes(2);
   });
 
-  it.each([
-    mockBreach({ Name: defaultBreachName, IsVerified: false }),
-    mockBreach({ Name: defaultBreachName, IsFabricated: true }),
-    mockBreach({ Name: defaultBreachName, IsSpamList: true }),
-    mockBreach({ Name: defaultBreachName, Domain: "" }),
-  ])("skips if breach is not notifiable", async (breach) => {
```
Vinnl, line 82: Question: what's the reason for removing the other test cases? They still seem relevant and like they would continue to pass?

src/services/BreachDataService.test.ts
```diff
@@ -0,0 +1,163 @@
...
+
+describe("BreachService", () => {
+  const redis = new MockRedis() as unknown as Redis;
+  const logger = mockLogger();
+
+  beforeAll(async () => {
+    await redis.flushall();
+  });
+  afterEach(async () => {
+    jest.clearAllMocks();
```
Vinnl: Nit: should be unnecessary.


```suggestion
```

src/services/BreachSyncService.test.ts
```diff
@@ -0,0 +1,179 @@
...
+
+describe("BreachSyncService", () => {
+  const redis = new MockRedis() as unknown as Redis;
+  let fetchBreaches: jest.Mock<Promise<HibpGetBreachesResponse>, []>;
+
+  beforeEach(() => {
+    fetchBreaches = jest.fn().mockResolvedValue(breachData);
+  });
+  afterEach(async () => {
+    jest.clearAllMocks();
```
Vinnl: Nit:


```suggestion
```

src/scripts/cronjobs/emailBreachAlerts/emailBreachAlerts.test.ts
kschelonka: I think this got lost in the merge/rebase shuffle, thank you for noticing

src/services/BreachSyncService.ts
kschelonka: I don't understand the "throwing away advantages of typescript" part, can you say more?

Why it's here: to avoid potentially breaking changes in behavior. The validation was only done in the breach alerts job, not on the fetch, which is why I didn't put it there. I agree it makes sense to validate in the fetch function, but let's make sure it doesn't break anything

src/services/BreachSyncService.ts
kschelonka: My history as a programmer 😅

I can move to a factory function pattern

src/services/BreachDataService.ts
kschelonka: It was to keep the interface the same as the other repository injectors. I'll just make it a function since it's unlikely to be extended

src/services/BreachDataService.ts
kschelonka: It's my default and was idiomatic on other teams. I'll move it to a factory function to be more idiomatic with this repository, since the internal state isn't very complicated

src/services/BreachSyncService.ts
Vinnl: (Note that the "throwing away the advantages of TypeScript" bit is in the existing code, not in your code.)

I meant that ideally, if a value has been determined to be of type `HibpGetBreachesResponse`, the rest of the code can assume that to actually be the case. The assertion (in the existing code) asserts that it is of that type without actually having validated that the actual value matches it - your `validateBreaches` function fixes that, but at this time, only at the call site.

(But yeah, my assumption that it wouldn't break anything might be a bit strong - presumably, existing code also assumes that it matches the type `HibpGetBreachesResponse`, but it might not at this time be using it in a way that would throw. A safer alternative might be to make it return `null | HibpGetBreachesResponse` instead of throwing, and then TS will tell us where we need to update... Assuming there are no other such assertions :melting_face: )

PR #6338, opened 2025-12-10
chore(hibp): handle multiple keys in webhook endpoint

kschelonka:
Handle multiple key versions for notify endpoint


src/utils/hibp.ts
```diff
@@ -37,6 +37,17 @@ const RENAMED_BREACHES_MAP = {
...
 };
 
+/**
+ * Returns true if token matches our secret tokens
+ *
+ * @param token token extracted from request
+ * @param validTokens token string, comma-separated for old vs. new
+ */
+export function isValidBearer(token: string, validTokens: string): boolean {
+  const tokenArray = validTokens.split(",");
```
mansaj: nit(nonblocking): `validTokens.split(",").map(t => t.trim())` for whitespace?

src/utils/hibp.ts
```diff
@@ -37,6 +37,17 @@ const RENAMED_BREACHES_MAP = {
...
+  return tokenArray.indexOf(token) > -1 ? true : false;
```
mansaj: question(nonblocking): is it simpler:  `return tokenArray.includes(token);`

mansaj (approved):

PR #6347, opened 2025-12-12
feat(o11y): send manually logged server errors to Sentry

kschelonka:
# References:

[MNTOR-5173]

# Description

If Sentry is initialized, add a Sentry transport to winston server logger. This will automatically forward any error-level logs to Sentry. There are currently application errors that are sent to GCP logging but not Sentry, so we aren't always alerted.

Follow-up work will be to audit and review errors as they come up, determining if the errors are informative, actionable, and appropriately leveled.

### DIFF
--- a/src/app/functions/server/logging.ts
+++ b/src/app/functions/server/logging.ts
@@ -27,23 +27,20 @@ const getLoggingWinston = () =>
 const SentryWinstonTransport = Sentry.createSentryWinstonTransport(Transport, {
   levels: ["error", "warn"],
 });
-const sentryTransports: LoggerOptions["transports"] = [
+const logTransports: LoggerOptions["transports"] = [
   ["gcpdev"].includes(process.env.APP_ENV ?? "local")
     ? getLoggingWinston()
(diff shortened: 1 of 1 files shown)


kschelonka:
It's not in the diff, but the sentryTransports includes transports.Console(). I realized there was a redundant if statement from the old pattern so I just updated it, which should also make it clearer

src/app/functions/server/logging.ts
```diff
@@ -35,17 +27,20 @@ export const logger = createLogger({
...
-  sentryTransports.push(new SentryWinstonTransport());
+  logTransports.push(new SentryWinstonTransport());
 }
 
-export const sentryLogger = createLogger({
+export const logger = createLogger({
   level: "info",
-  transports: sentryTransports,
+  // In GCP environments, use cloud logging instead of stdout.
+  // FIXME https://mozilla-hub.atlassian.net/browse/MNTOR-2401 - enable for stage and production
```
kschelonka, line 42: this is from before but it's showing up as new in the diff

PR #6350, opened 2025-12-15
MNTOR-5089: Consolidate settings

Vinnl:
# References:

Jira: MNTOR-5089
Figma:

# Description

This builds on the setup proposed in https://github.com/mozilla/blurts-server/pull/6324 (but with https://github.com/mozilla/blurts-server/pull/6343 as the target to merge into). It replaces most uses of `process.env` with a centralised `config.ts` that verifies that requisite env vars are set at app startup (and thus aren't `| undefined` when accessed), can also convert integers and enums (and error when they don't match the expected type), and provide a fallback value if so desired.

There are still uses of `process.env`, e.g. when accessing `NEXT_PUBLIC_` env vars (these are part of the client-side bundle and thus can't/shouldn't access secrets, and get compiled in at build time by Next.js), in functional/e2e tests (to avoid having to initialise all env vars just to run e2e tests), and things like middleware where the full environment might not be properly initialised. Hopefully it's still a step forward though.

# How to test

Everything should work as before.

- [ ] I've added a unit test to test for potential regressions of this bug. - Updated as needed, otherwise existing tests should cover this.


src/app/(proper_react)/(redesign)/(authenticated)/user/(dashboard)/dashboard/[[...slug]]/page.tsx
```diff
@@ -72,14 +72,12 @@ export default async function DashboardPage(props: Props) {
     email: session.user.email,
   });
 
-  const isNewUser = !isPrePlusUser(session.user);
```
Comment (Vinnl) [inline src/app/(proper_react)/(redesign)/(authenticated)/user/(dashboard)/dashboard/[[...slug]]/page.tsx:74]: This bit is no longer relevant when Plus is disabled, so I removed it (cc @codemist).

src/app/(proper_react)/(redesign)/(authenticated)/user/(dashboard)/dashboard/fix/leaked-passwords/leakedPasswordsData.tsx
```diff
@@ -210,16 +210,19 @@ function getLeakedPasswords(props: LeakedPasswordLayout) {
...
-                        eventData={{
-                          link_id: "changed_password",
-                        }}
-                        showIcon
-                      />
-                    ),
+                    link_to_breach_site:
+                      typeof breachSite === "string" ? (
+                        <TelemetryLink
+                          href={breachSite}
```
Vinnl, line 216: `breachSite` came from `getBeachInfo` (see line 179), and would be an empty string if not defined - resulting in an `href` to an empty string here, and thus an invalid link. That's why I replaced the link with a plain `<span>` here. (Maybe even should've used a React fragment?)

src/db/knexfile.js
Vinnl, line 1: Migrated to TS below, since `config.ts` is also TypeScript.

src/config.ts
```diff
@@ -0,0 +1,186 @@
...
+ * Environment-specific values
+ *
+ * Use this object to access values that differ per environment (e.g. running
+ * locally, on stage, or in production).
+ *
+ * Prefer using this object over `process.env`, as this object validates values
+ * at app startup, ensuring that their values are something we expect, optionally
+ * falling back to sensible values (for running locally) if not.
+ */
+export const config = {
```
Vinnl, line 19: Not sure if `config` is actually a good name for this - there are a number of other variables with that name, making auto-import a pain, and resulting in the need to alias the import in a couple of places. Open to suggestions!

.github/workflows/build.yaml
```diff
@@ -25,3 +25,14 @@ jobs:
...
+          SENTRY_DSN: "Unused, like the build result — we're just verifying that a build can succeed."
+          EMAIL_FROM: "Unused, like the build result — we're just verifying that a build can succeed."
+          HIBP_KANON_API_TOKEN: "Unused, like the build result — we're just verifying that a build can succeed."
+          ONEREP_API_KEY: "Unused, like the build result — we're just verifying that a build can succeed."
+          S3_BUCKET: "Unused, like the build result — we're just verifying that a build can succeed."
+          FX_REMOTE_SETTINGS_WRITER_SERVER: "Unused, like the build result — we're just verifying that a build can succeed."
+          FX_REMOTE_SETTINGS_WRITER_USER: "Unused, like the build result — we're just verifying that a build can succeed."
+          FX_REMOTE_SETTINGS_WRITER_PASS: "Unused, like the build result — we're just verifying that a build can succeed."
+          NIMBUS_UUID_NAMESPACE: "Unused, like the build result — we're just verifying that a build can succeed."
+          NIMBUS_SIDECAR_URL: "Unused, like the build result — we're just verifying that a build can succeed."
```
codemist, line 37: Should some of these be deprecated later? Like `ONEREP_API_KEY`?

src/app/(proper_react)/(redesign)/(authenticated)/admin/fxa/page.tsx
```diff
@@ -6,14 +6,15 @@ import { getServerSession } from "../../../../../functions/server/getServerSessi
...
 import { AttachedClients } from "./AttachedClients";
+import { config } from "../../../../../../config";
 
 export default async function DevPage() {
   const session = await getServerSession();
 
   if (
     !session?.user?.email ||
     !isAdmin(session.user.email) ||
-    process.env.APP_ENV === "productions"
```
codemist, line 16: Was this a typo before?

src/app/(proper_react)/(redesign)/(authenticated)/user/(dashboard)/dashboard/fix/leaked-passwords/leakedPasswordsData.tsx
```diff
@@ -260,21 +263,24 @@ function getLeakedPasswords(props: LeakedPasswordLayout) {
...
-                          name: "click",
-                          data: {
-                            link_id: "changed_security_question",
-                            // TODO: Enable after the parameter has been added to metrics.yaml.
-                            // link_name: `changed_security_question_${breachName}`,
-                          },
-                        }}
-                      />
-                    ),
+                    link_to_breach_site:
```
codemist, line 266: Curious why a conditional was added here?

src/app/api/utils/auth.tsx
```diff
@@ -217,7 +209,7 @@ export const authOptions: AuthOptions = {
             },
           });
 
-          await initEmail(process.env.SMTP_URL);
+          await initEmail();
```
codemist: Do we no longer need to pass in the config here? I think the function still has it as a param?

src/app/api/v1/premium/route.ts
```diff
@@ -14,10 +15,7 @@ export function GET(req: NextRequest) {
     const { searchParams } = new URL(req.url);
     const _email = searchParams.get("email");
 
-    return NextResponse.redirect(
-      `${process.env.SERVER_URL ?? ""}/user/breaches`,
-      302,
-    );
+    return NextResponse.redirect(`${config.serverUrl}/user/breaches`, 302);
```
codemist, line 18: nit: Should this keep the nullish operator?

src/app/api/v1/user/email/route.ts
```diff
@@ -110,6 +111,6 @@ export async function POST(req: NextRequest) {
     }
   } else {
     // Not Signed in, redirect to home
-    return NextResponse.redirect(process.env.SERVER_URL ?? "/", 301);
+    return NextResponse.redirect(config.serverUrl, 301);
```
codemist, line 114: nit: Same qn as above about omitting the nullish operator. Then again... if the server url wasn't defined we'd be running into bigger issues 😅

src/app/functions/universal/mock.test.ts
```diff
@@ -22,16 +22,4 @@ describe("mock detectors", () => {
       expect(isUsingMockHIBPEndpoint()).toEqual(expected);
     });
   });
-  describe("isUsingMockONEREPEndpoint", () => {
```
codemist, line 25: Thanks for pruning!

src/config.ts
codemist: I think it's fine to leave it as `config`, other variables should be worked around to accommodate the naming convention here, not the other way around imo.

src/emails/templates/breachAlert/BreachAlertEmail.tsx
```diff
@@ -34,7 +35,7 @@ export const BreachAlertEmail = (props: BreachAlertEmailProps) => {
   const utmCampaignId = "breach-alert-global";
 
   const premiumSubscriptionUrlObject = new URL(
-    `${process.env.SERVER_URL}/link/subscribe/monthly`,
+    `${config.serverUrl}/link/subscribe/monthly`,
```
codemist: I think this needs to be removed, since we shouldn't have `premiumSubscriptionUrlObject` anymore.

src/initializeEnvVars.js
```diff
@@ -0,0 +1,23 @@
...
+
+// Doesn't run in tests by design, so ignore for coverage:
+/* c8 ignore start */
+if (
+  typeof process.env.NEXT_RUNTIME !== "string" &&
+  process.env.NODE_ENV !== "test"
+) {
+  // If we're in Next.js, our `.env` files are already set up to be loaded.
+  // Outside of Next.js (e.g. in cron jobs), however, we need to explicitly load them.
+  // (In unit tests, `next/jest` takes care of this, so no need to run this there.)
```
codemist, line 21: kudos for the explanation!

src/scripts/cronjobs/updateBreachesInRemoteSettings/updateBreachesInRemoteSettings.test.ts
```diff
@@ -45,23 +50,27 @@ describe("updateBreachesInRemoteSettings job", () => {
...
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    (configModule.config as any).fxRemoteSettingsWriterUser = "user";
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    (configModule.config as any).fxRemoteSettingsWriterPass = "pass";
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    (configModule.config as any).fxRemoteSettingsWriterServer =
+      "https://example.com/v1";
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    (configModule.config as any).appEnv = "local";
   });
```
codemist, line 68: Could you explain why these need to be typed as `any`?

src/utils/fxa.ts
```diff
@@ -191,7 +184,7 @@ type FxaGetOauthSubscribptionsActiveResponseSuccess = Array<{
 async function getSubscriptions(
   bearerToken: string,
 ): Promise<FxaGetOauthSubscribptionsActiveResponseSuccess | null> {
-  const subscriptionIdUrl = `${envVars.OAUTH_ACCOUNT_URI}/oauth/subscriptions/active`;
+  const subscriptionIdUrl = `${config.oauthAccountUri}/oauth/subscriptions/active`;
```
codemist, line 187: This seems like it should've been part of the sunset.

src/app/(proper_react)/(redesign)/(authenticated)/user/(dashboard)/dashboard/fix/leaked-passwords/leakedPasswordsData.tsx
codemist: Nvm, I think it was explained in a previous comment of yours.

src/config.ts
Vinnl: Sometimes it can, but e.g. `next.config.ts` expects an exported `config`, and there's a couple of other places like that. We could also name this `envVars` or something, for example.

src/app/(proper_react)/(redesign)/(authenticated)/admin/fxa/page.tsx
Vinnl: Yes! I was trying to look for this but couldn't find it in the diff - wanted to highlight that the type checking now highlighted this typo, so it also prevents bugs.

src/app/api/v1/premium/route.ts
Vinnl: It no longer needs to! One of the benefits of the new setup is that we verify that the relevant env var is defined at app startup, and thus the rest of the code can assume that it is defined. (And thus the TS type is now also `string` rather than `string | undefined`.)

src/app/api/v1/user/email/route.ts
Vinnl: See above :)

src/utils/fxa.ts
Vinnl: Good point, it should. I created a new task in your epic: MNTOR-5152.

.github/workflows/build.yaml
Vinnl: Oh yes, good point - I copied the list from another CI job, to avoid having to re-push after every added variable to find which else was missing :P 495beb4d

src/app/api/utils/auth.tsx
Vinnl: Ah yeah, I removed this because it was already initialised to this value when it's left out, but I suppose being explicit isn't necessarily a bad idea, and removing it isn't related to the scope of this PR, so restored in 6580dfe1.

src/emails/templates/breachAlert/BreachAlertEmail.tsx
Vinnl: Good point, removed in 9f02913e.

src/scripts/cronjobs/updateBreachesInRemoteSettings/updateBreachesInRemoteSettings.test.ts
Vinnl: Ah, yes, good one: c92e843e

src/scripts/cronjobs/updateBreachesInRemoteSettings/updateBreachesInRemoteSettings.test.ts
```diff
@@ -53,6 +53,9 @@ describe("updateBreachesInRemoteSettings job", () => {
   beforeEach(async () => {
     jest.resetAllMocks();
 
+    // The `config` object is typed as readonly, because in normal code,
```
codemist, line 56: Thanks for adding this comment!

src/app/functions/l10n/index.ts
```diff
@@ -68,7 +69,7 @@ export function createGetL10nBundles<
   : GetL10nBundlesForLocales {
   return (acceptLangHeader = options.getAcceptLangHeader?.()) => {
     const languages = acceptedLanguages(acceptLangHeader);
-    const supportedLocales = process.env.SUPPORTED_LOCALES?.split(",");
+    const supportedLocales = config.supportedLocales.split(",");
```
kschelonka, line 72: optional: consider moving this logic to the config loader

Vinnl: Yeah, I considered that, but there was at least one other similar setup, and I didn't want to track down all their usages in this scope, so I left that as a future job. I just filed MNTOR-5159 for that.

PR #6365, opened 2025-12-22
feat(o11y): add hibp request count and metrics endpoint

kschelonka:
# References:

Jira: MNTOR-5123
Figma:

# Description
Add HIBP request counters (total, errors) and metrics endpoint for google-managed prometheus to scrape

# Screenshot (if applicable)

Not applicable.

# How to test
* Start server with `npm run dev`
* View localhost:6060/api/v1/metrics -- see no requests to hibp endpoint or errors
* Make requests to localhost:6060/api/v1/notify/hibp to see request count and appropriate errors go up (e.g. make with invalid key)

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -6,6 +6,8 @@
 import "./app/functions/server/notInClientComponent";
 import "./initializeEnvVars";
 
+// Don't need to have coverage on config object
+/* c8 ignore start */
 const isLocalOrTest =
   process.env.NODE_ENV === "test" || process.env.APP_ENV === "local";
 
@@ -81,16 +83,17 @@ export const config = {
   fxRemoteSettingsWriterServer: process.env.FX_REMOTE_SETTINGS_WRITER_SERVER,
 
   gcp: {
-    projectId: getEnvString("GCP_PROJECT_ID", {
+    projectId: getEnvString("GCP_PUBSUB_PROJECT_ID", {
       fallbackValue: isLocalOrTest ? "your-project-name" : undefined,
     }),
     pubsub: {
-      hibpTopic: getEnvString("GCP_PROJECT_ID", {
+      hibpTopic: getEnvString("GCP_PUBSUB_TOPIC_NAME", {
         fallbackValue: isLocalOrTest ? "hibp-breaches" : undefined,
       }),
     },
   },
 } as const;
+/* c8 ignore end */
 
 /**
  * Like {@link getEnvString}, but also ensures the value is a valid integer


src/instrumentation.node.ts
```diff
@@ -0,0 +1,86 @@
...
+
+declare global {
+  var metrics: Readonly<MetricsState>;
+}
+
+function getOrInitMetrics(): MetricsState {
+  // Return cached state
+  if (globalThis.metrics !== undefined) return globalThis.metrics;
+  const registry = new client.Registry();
+  client.collectDefaultMetrics({ register: registry });
```
Vinnl, line 23: Just wondering if this shouldn't be done [in `instrumentation.ts`](https://nextjs.org/docs/app/guides/instrumentation). which I think is intended exactly for this use case?

(That would only work in the context of Next.js, but I imagine the setup is going to be different in different environments anyway.)

Vinnl (approved):
I didn't know how to test this myself, but the code changes look sensible.

One thing that might be relevant is that Next.js also has [support for OpenTelemetry](https://nextjs.org/docs/app/guides/open-telemetry) that can be enabled and automatically tags e.g. spans for endpoints. (This is me just linking concepts that I have vaguely heard about in my head, but I think I also heard something about not being able to use OTel yet?)

src/instrumentation.node.ts
kschelonka: [Replied here](https://github.com/mozilla/blurts-server/pull/6365#issuecomment-3697093173)

src/config.ts
```diff
@@ -76,7 +81,19 @@ export const config = {
   fxRemoteSettingsWriterUser: process.env.FX_REMOTE_SETTINGS_WRITER_USER,
   fxRemoteSettingsWriterPass: process.env.FX_REMOTE_SETTINGS_WRITER_PASS,
   fxRemoteSettingsWriterServer: process.env.FX_REMOTE_SETTINGS_WRITER_SERVER,
+
+  gcp: {
+    projectId: getEnvString("GCP_PUBSUB_PROJECT_ID", {
+      fallbackValue: isLocalOrTest ? "your-project-name" : undefined,
```
Vinnl, line 87: Thought (unrelated to this PR): Hmm, maybe this approach is also something we could use to deal with env vars that we only need to be set at runtime, but not buildtime - only provide a fallback value if `process.env.CI` is set.

Vinnl (approved):
This time I did follow the testing instructions, and I saw the error counter increment at `/metrics`. I did see the following errors in my console after the first call to `/notify`, and the `MetadataLookupWarning` every time — not sure if that's expected:

```
(node:36490) Warning: Accessing non-existent property 'Error' of module exports inside circular dependency
(Use `node --trace-warnings ...` to show where the warning was created)
(node:36490) MetadataLookupWarning: received unexpected error = All promises were rejected code = UNKNOWN
{"level":"error","message":"error_queuing_hibp_breach:","topicName":"hibp-breaches"}
 POST /api/v1/hibp/notify 429 in 127508ms
```

PR #6369, opened 2025-12-23
Apply codemods for Next.js 16

Vinnl:
See https://nextjs.org/blog/next-16 and https://nextjs.org/blog/next-16-1

Apart from the codemode, I had to fix our ESLint config, which now had some kind of circular references but had gotten fairly messy anyway. All the same config should still be there, although some of it moved into Next's own packages. However, some of them (most notably react-hooks) apparently didn't cover all our code, so I had to fix a bunch of errors, or at least add exceptions for them for now - I left a comment when I did.

I split the changes from the codemod and the changes I had to do for ESLint into separate commits, so that might help during reviewing.

~~Note that this merges into https://github.com/mozilla/blurts-server/pull/6366, though maybe it should just target `main` :thinking:~~ Forgot that I had done that and just rebased on `main` - so I just retargeted this there.

### DIFF
--- a/next.config.js
+++ b/next.config.js
@@ -186,19 +186,23 @@ const nextConfig = {
   //    Module not found: Can't resolve './ROOT' <dynamic>
   //    server relative imports are not implemented yet. Please try an import relative to the file you are importing from.
   serverExternalPackages: ["mjml", "commonjs", "knex", "ioredis-mock"],
-  // transpilePackages is required for Jest (via next/jest) to correctly
-  // transform ESM-only dependencies in node_modules. Without this,
-  // tests importing Storybook logic or components will fail with:
-  // "SyntaxError: Cannot use import statement outside a module"
-  transpilePackages: [
+};
+
+if (process.env.NODE_ENV === "test") {
+  nextConfig.transpilePackages = [
+    ...(nextConfig.transpilePackages ?? []),
+    // transpilePackages is required for Jest (via next/jest) to correctly
+    // transform ESM-only dependencies in node_modules. Without this,
+    // tests importing Storybook logic or components will fail with:
+    // "SyntaxError: Cannot use import statement outside a module"
     "@storybook/react",
     "@storybook/nextjs",
     "@storybook/addon-links",
     "@storybook/addon-essentials",
     "@storybook/addon-interactions",
     "storybook",
-  ],
-};
+  ];
+}
 
 const sentryOptions = {
   // Additional config options for the Sentry Webpack plugin. Keep in mind that


src/app/(proper_react)/(redesign)/(authenticated)/user/(dashboard)/dashboard/View.tsx
```diff
@@ -170,66 +178,6 @@ export const View = (props: Props) => {
   const hasExposures = breachesDataArray.length > 0;
   const hasUnresolvedBreaches = tabSpecificExposures.length > 0;
 
-  const TabContentActionNeeded = () => {
```
Vinnl, line 173: This file defined three components (`TabContentActionNeeded`, `TabContentFixed`, and `getZeroStateIndicator` - the latter for some reason not a component) inside another component, leading to them getting recreated on every render. The linter pointed this out.

eslint.config.js
Vinnl, line 1: I had an LLM figure out the "circular reference" error that ESLint threw, and it actually organised it quite neatly, grouping a few things together that belonged together but were separate. What I did was inspect the diff between the before and after, and check that everything that was removed from the before had a place in the after, and that all the additions also made sense.

eslint.config.js
```diff
@@ -2,108 +2,34 @@
...
-      "dist",
-      "coverage",
-      "!.storybook",
-      "playwright-report/**",
-    ],
-  },
-  ...compat.config({
-    extends: ["next"],
-  }),
+const estlingConfig = defineConfig([
```
kschelonka: typo here, `eslintConfig` is probably intended

src/app/(proper_react)/(redesign)/(authenticated)/user/(dashboard)/dashboard/View.tsx
```diff
@@ -63,6 +66,11 @@ export const View = (props: Props) => {
     UserAnnouncementWithDetails[] | null
   >(props.userAnnouncements);
   useEffect(() => {
+    // TODO The `react-hooks` ESLint rules didn't apply correctly
```
kschelonka, line 69: Please write a ticket for this TODO so it doesn't get lost

kschelonka (commented):
I've read up on next based on your comments and I think I understand why the workarounds are done. Someone with next experience should sign off though.

eslint.config.js
Vinnl: Whoops! d6a8c536

src/app/(proper_react)/(redesign)/(authenticated)/user/(dashboard)/dashboard/View.tsx
Vinnl: Good call: 59718ba2 (and [the ticket](https://mozilla-hub.atlassian.net/browse/MNTOR-5167)).

PR #6668, opened 2026-05-21
MNTOR-5282 - Strip down image switching logic and use linear gradient instead

codemist:
# References:

Jira: MNTOR-5282
Figma:

# Description

This PR replaces the previous round of dark-mode swap logic with a simpler, mostly-light-mode-only approach. The original work tried to detect dark mode via `@media (prefers-color-scheme: dark)` and `[data-ogsc]/[data-ogsb]` attributes, then swap to dark-variant logos and re-color sections.... but QA found Gmail mobile ignored every targeting mechanism we tried (prefers-color-scheme, color-scheme: light only, supported-color-schemes), force-inverted the email anyway, and produced inconsistent results.

For Gmail specifically, there's a small workaround on the hero and footer surfaces: their backgrounds are set via the `linear-gradient` property, because Gmail's force-invert leaves CSS gradients untouched. Kudos to this blogpost for unblocking me here, because I tried every other method to keep both the image + light purple background: https://www.hteumeuleu.com/2021/fixing-gmail-dark-mode-css-blend-modes/

The result isn't perfect... in clients that force their own dark mode regardless of trying to enforce light mode, the email will render in those clients' inverted scheme rather than the original light styling (at least that's what I found when I tested witht the Gmail app on my iPhone). The important thing is that the email stays accessible: logos are still visible (single light variant rendered everywhere), text contrast is preserved by the client's own inversion, and there are no half-applied states where a dark logo lands on a light section or vice versa.

# Screens
IMG_3753
IMG_3754
hot (if applicable)

# How to test


Vinnl (approved):
Ahhaha, what a workaround, excellent. The trade-off you make seems to be the right one to me, and I love a good trade-off.

(It's ironic that forced-dark-mode is now preventing us from supporting dark mode. Maybe one day email clients will get their shit together.)

src/emails/components/EmailFooter.tsx
```diff
@@ -177,15 +176,7 @@ export const RedesignedEmailFooter = (props: Props) => {
             href={`https://www.mozilla.org/?utm_source=monitor-product&utm_medium=product-email&utm_campaign=${props.utm_campaign}&utm_content=header-logo`}
             width="150px"
             align="center"
-            css-class="dm-img-light"
-          />
-          <mj-image
-            alt={l10n.getString("email-footer-logo-mozilla-alt")}
-            src={`${config.serverUrl}/images/email/mozilla-logo-dark-mode.png`}
```
codemist, line 184: @Vinnl These images were newly added and never made it to prod, so they're good to remove.

src/emails/templates/breachAlert/BreachAlertEmailVer2.tsx
```diff
@@ -48,13 +48,12 @@ export const BreachAlertEmail = (props: BreachAlertEmailProps) => {
       `}
         </mj-style>
       </mj-head>
-      <mj-body background-color="#fff" css-class="dm-body">
+      <mj-body background-color="#fff">
```
codemist, line 51: These variants can be mostly ignored since they're not going to be user-facing (for now). They were mainly experimental.

PR #6702, opened 2026-06-02
MNTOR-5289 - Support per env FXA CI bypass tokens

codemist:
# References:

Jira: MNTOR-5289

# Description

FxA's `fxa-ci` bypass header requires different token values for stage and prod. Previously `FXA_CI_SECRET` could only hold one value, so functional tests passed on whichever env matched the current secret and failed on the other.

# Screenshot (if applicable)

Not applicable.

# How to test

### DIFF
--- a/.github/workflows/functional_tests_cron.yml
+++ b/.github/workflows/functional_tests_cron.yml
@@ -68,6 +68,8 @@ jobs:
           E2E_TEST_ACCOUNT_BASE_EMAIL: ${{ secrets.E2E_TEST_ACCOUNT_BASE_EMAIL }}
           E2E_TEST_ACCOUNT_BASE_PASSWORD: ${{ secrets.E2E_TEST_ACCOUNT_BASE_PASSWORD }}
           FXA_CI_SECRET: ${{ secrets.FXA_CI_SECRET }}
+          FXA_CI_SECRET_STAGE: ${{ secrets.FXA_CI_SECRET_STAGE }}
+          FXA_CI_SECRET_PROD: ${{ secrets.FXA_CI_SECRET_PROD }}
       - name: Send GitHub Action trigger data to Slack workflow
         id: slack
         uses: slackapi/slack-github-action@45a88b9581bfab2566dc881e2cd66d334e621e2c # v3.0.3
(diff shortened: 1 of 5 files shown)


.github/workflows/functional_tests_cron.yml
```diff
@@ -68,6 +68,8 @@ jobs:
           E2E_TEST_ACCOUNT_BASE_EMAIL: ${{ secrets.E2E_TEST_ACCOUNT_BASE_EMAIL }}
           E2E_TEST_ACCOUNT_BASE_PASSWORD: ${{ secrets.E2E_TEST_ACCOUNT_BASE_PASSWORD }}
           FXA_CI_SECRET: ${{ secrets.FXA_CI_SECRET }}
```
Vinnl: Since the tests are failing anyway, can't we just remove the existing `FXA_CI_SECRET` straight away?

.github/workflows/functional_tests_cron.yml
```diff
@@ -68,6 +68,8 @@ jobs:
...
+          FXA_CI_SECRET_STAGE: ${{ secrets.FXA_CI_SECRET_STAGE }}
+          FXA_CI_SECRET_PROD: ${{ secrets.FXA_CI_SECRET_PROD }}
```
Vinnl: Actually, maybe we can just avoid the need for code changes, and just adjust this in the workflow with something like this?


```suggestion
FXA_CI_SECRET: ${{ (github.event_name == 'workflow_dispatch' && inputs.test_env || matrix.test_env) == 'production' && secrets.FXA_CI_SECRET_PROD || secrets.FXA_CI_SECRET_STAGE }}
```

Vinnl (approved):

.github/workflows/functional_tests_cron.yml
codemist: I'm going to keep the env-aware selection in the TS helper rather than collapse into one line for readability.

codemist:
True... going to remove the legacy `FXA_CI_SECRET` once I've confirmed the per-env secrets work end-to-end. Since I always set the legacy value to the prod token, the fallback only ever quietly masks a misconfigured `FXA_CI_SECRET_PROD`, keeping it around is more of a risk than a safety net. Cleaner to delete it now and let any future misconfig surface immediately.

PR #6738, opened 2026-06-24
MNTOR-4448: merge duplicate fxa_uid subscribers and enforce uniqueness

mansaj:
# References:

Jira: MNTOR-4448

# Description
A FxA account is meant to map to exactly one `subscribers` row, but nothing at the DB level enforced it and an incident left ~45k accounts with 2-12 duplicate rows each. 

What's changed:
- **`src/db/mergeDuplicateSubscribers.ts`** – a one-off merge routine that folds each duplicate `fxa_uid` set down to a single surviving ("winner") row, aggregates user-meaningful data onto it, re-parents every child table, then deletes the losers. Each group runs in its own transaction and the routine is idempotent, so it can resume after an interruption and is safe to re-run.
- **`src/scripts/one-offs/mergeDuplicateSubscribers.ts`** – the runnable entry point (supports `--dry-run` and `--batch-size=`), with pre-flight and post-run duplicate-count audits.

### Merge strategy (single free population, post-incident)
- **Winner**: most recent `updated_at`, then newest `created_at`, then highest `id`.
- **Aggregation (nothing silently lost)**: SUM `sign_in_count`, MAX activity timestamps, OR the boolean flags.
- **`all_emails_to_primary` is treated as tri-state, not a plain OR**: `null` = instant breach alerts off, while both `true` (deliver to primary) and `false`(deliver to affected address) are subscribed modes. Merge keeps the user subscribed if any row was subscribed (`true` if any row `true`, else `false` if any row `false`, else `null`).
- **Child re-parenting**: every table with an FK to `subscribers` is handled explicitly, including the non-cascading FKs (`email_notifications`, `feature_flag_events`).

### Data-correctness handling on de-dupe
When a child row would violate a UNIQUE constraint it is dropped rather than re-parented, but never at the cost of losing meaningful state:
- **Conflicts among multiple losers** (not just winner-vs-loser) are detected, so groups where two losers share a unique key but the winner has no matching row don't trip constraints on the final bulk update.
- **`email_subscriptions.subscribed`** and **`email_notifications.notified` / `appeared`** are OR-merged into the surviving row, so a dropped duplicate can't silently unsubscribe a user or cause a breach alert to be re-sent.
- **Email verification** is preserved: since breach scanning only matches verified emails, a dropped verified duplicate promotes the surviving copy (`subscribers.primary_verified` or `email_addresses.verified`) to `true`.

## Rollout sequence (must be ordered)
1. Land this PR
2. Run the one-off with `--dry-run`, review the logged per-group actions and summary, then run for real in stage then prod
4. Confirm the post-run audit count is `0` in the both environment, then apply the unique-index migration in the following PR.

# How to test
- [x] `vitest run --config vitest.integration.config.ts src/db/mergeDuplicateSubscribers.integration.ts` (12 tests passing)
- [x] Winner selection + aggregation (sign-in SUM, timestamp MAX, flag OR)
- [x] `all_emails_to_primary` tri-state: winner `null` + loser `false` → `false`; `false` + `true` → `true`; all `null` → `null`
- [x] Multi-loser unique conflict with no matching winner row
- [x] `email_subscriptions.subscribed` and `email_notifications.notified`/`appeared` preserved on collision
- [x] Email verification promoted when a verified duplicate is dropped (primary + secondary)
- [x] Dry-run writes nothing; second run is a no-op (idempotency)

[MNTOR-4448]: https://mozilla-hub.atlassian.net/browse/MNTOR-4448?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/db/mergeDuplicateSubscribers.ts
```diff
@@ -0,0 +1,724 @@
...
+    Object.fromEntries(
+      orColumns.map((column) => [column, row[column] === true]),
+    );
+
+  const winnerRows = await trx(child.table)
+    .where(child.fkColumn, winnerId)
+    .select(...selectColumns);
+  for (const row of winnerRows as Array<Record<string, unknown>>) {
+    const key = conflictKey(row, conflictColumns);
+    if (key !== null && !survivors.has(key)) {
```
joeherm, line 348: question: does the `!survivors.has(key)` mean that we could lose information for a duplicate row if the key already exists?

joeherm (approved):
@mansaj I think this looks pretty good. I read the comments on the individual functions and worked through the logic which seems solid. It _is_ hard to grasp how all the pieces will come together when it comes to edge cases since I'm not super familiar with the incident.

I think getting this to staging and doing a dry-run is probably the next best step, so I'm going to approve. Two questions though:

1. Are there enough conflicts in staging to be representative of prod?
2. If not, is it worth writing a small SQL test case with before/after (i.e. a test that inserts to the DB, runs the function, and looks at all the tables' expected states)? The integration test is nice but seems like it might not capture enough of the complexity of the actual DB situation.

My other 2c: I think part of me wishes this code was less generic and was written more as a set of per-table functions (as opposed to using a `ChildTable` abstraction). Even if duplicative, because this is hopefully one-off, it would have made me a little more confident in reading the changes.

src/db/mergeDuplicateSubscribers.ts
mansaj: Good question, but no, I don't think anything is lost here. This loop only iterates the winner's own rows, and every table we de-dupe already has a UNIQUE constraint covering the FK + conflictColumns (e.g. UNIQUE(subscriber_id, list_id) on `email_subscriptions`, UNIQUE(subscriber_id, breach_id, email) on email_notifications). So a single subscriber can't already hold two rows with the same key. `survivors.has(key)` is never actually true while walking the winner's rows. The guard is just defensive.

Where duplicates genuinely show up is across the losers, and those are handled in the loop below: when a loser row collides with an existing survivor, we don't silently drop it, we first OR its orColumns flags (notified/appeared, subscribed) into the survivor via flagBumps, then delete the now-redundant row. The only thing not carried over from a dropped row is its non-OR columns, which is intentional... that row is a duplicate by the table's own UNIQUE key, so those columns are redundant with the survivor's.

The `key === null` branch (a NULL in any conflict column) is also deliberately skipped so it's treated as distinct, matching default NULLS DISTINCT index semantics. 

I know.. Monitor code is pretty crazy

mansaj:
> @mansaj I think this looks pretty good. I read the comments on the individual functions and worked through the logic which seems solid. It _is_ hard to grasp how all the pieces will come together when it comes to edge cases since I'm not super familiar with the incident.
> 
> I think getting this to staging and doing a dry-run is probably the next best step, so I'm going to approve. Two questions though:
> 
>     1. Are there enough conflicts in staging to be representative of prod?
> 
>     2. If not, is it worth writing a small SQL test case with before/after (i.e. a test that inserts to the DB, runs the function, and looks at all the tables' expected states)? The integration test is nice but seems like it might not capture enough of the complexity of the actual DB situation.
> 
> 
> My other 2c: I think part of me wishes this code was less generic and was written more as a set of per-table functions (as opposed to using a `ChildTable` abstraction). Even if duplicative, because this is hopefully one-off, it would have made me a little more confident in reading the changes.

Some great points here..
1. Almost certainly not fully representative. I vaguely remeber running the numbers on stage and it's like 1% of what i saw in prod. 

We can quantify it before relying on it. The script logs a pre-flight count, and the same check standalone is:
```
SELECT count(*) FROM (
  SELECT fxa_uid FROM subscribers
  WHERE fxa_uid IS NOT NULL
  GROUP BY fxa_uid HAVING count(*) > 1
)
```
The plan is always: run `--dry-run` in `stage` first and review the logged per-group actions + summary regardless of volume (it exercises the code path end-to-end), then treat the prod dry-run as the real representativeness check before committing.

2. Agreed, this is worth adding. The current integration tests are focused (one concern per test). They don't exercise a single large group touching every child table at once. I'll add a "kitchen-sink" integration test: seed one `fxa_uid` group of maybe 12 rows that populates `email_addresses`, `subscriber_email_preferences`, `attributions`, `user_announcements`... etc with deliberate `winner-vs-loser` and `loser-vs-loser` collisions, mixed `verified/notified/subscribed` states.. then assert the full post-merge state of every table. That should capture more of the real-world complexity than the per-scenario tests do.

> I think part of me wishes this code was less generic and was written more as a set of per-table functions 

I went back and forth on it. The reason I landed on the generic `ChildTable` config is that I think the six child tables all need the same operation: 
* re-parent
* drop UNIQUE collisions (winner- and loser-side)
* OR-preserve a couple of boolean flags

and writing that six times risks the bug we'd least want.. ie the same edge case fixed in five places and missed in the sixth. Centralizing it means the collision/OR logic is reviewed and tested once.

joeherm (approved):
@mansaj Great extra test case!

PR #6740, opened 2026-06-26
feat: add unique index on fxa_uid in subscribers table to prevent dup…

mansaj:
# References:

Jira: MNTOR-4448

# Description
Follow-up to #6738. This adds the partial `UNIQUE` index on `subscribers.fxa_uid` that prevents duplicate `fxa_uid` rows from recurring once the one-off merge has run.
- Partial on `fxa_uid IS NOT NULL` so the many legitimately NULL rows pre-FxA / unverified signups) are unaffected.
- Index creation is left transactional on purpose: if any duplicate `fxa_uid` values still remain it fails and rolls back, acting as the final safety gate rather than corrupting data. This is intentionally split out from the prev PR so the schema change lands only after the merge one-off has completed in each environment.

## Rollout ordering
1. Merge & run the one-off
2. Confirm the post-run audit count is `0`:
   ```sql
   SELECT count(*) FROM (
     SELECT fxa_uid FROM subscribers
     WHERE fxa_uid IS NOT NULL
     GROUP BY fxa_uid HAVING count(*) > 1
   ) 
3. Then apply this migration.


joeherm (approved):
Deployment question: Presumably migrations are run as part of a `preSync` hook in monitor? What's your plan for running the one off script? If it was to `exec` into a pod to run, I think we will have to cut two separate releases. The first one with the one-off script, the second with this change. Otherwise I'm pretty sure that this migration failing will cause the code containing your one-off script to never be released.

The alternative would be to make an infra change that creates an earlier `preSync` hook with the data migration, followed by running the normal migrations.

mansaj:
> Deployment question: Presumably migrations are run as part of a `preSync` hook in monitor? What's your plan for running the one off script? If it was to `exec` into a pod to run, I think we will have to cut two separate releases. The first one with the one-off script, the second with this change. Otherwise I'm pretty sure that this migration failing will cause the code containing your one-off script to never be released.
> 
> The alternative would be to make an infra change that creates an earlier `preSync` hook with the data migration, followed by running the normal migrations.

Right, that migrations run as a preSync hook, so a failing migration blocks that release's rollout. We're handling it exactly the way you said: two separate releases, and the first is already complete.

Release 1: #6738 (merged & deployed)
* ships the one-off script + merge logic, with no migration. So there was nothing for the preSync hook to fail on, and it released normally.
* run the one-off: I exec'd into a `monitor-www prod` pod and ran `mergeDuplicateSubscribers.ts` - dry-run first (63,265 groups), then the real run. It merged 63,264 groups / deleted 65,987 loser rows with 0 failures, and the post-run audit
```
SELECT count(*) FROM (
  SELECT fxa_uid FROM subscribers WHERE fxa_uid IS NOT NULL
  GROUP BY fxa_uid HAVING count(*) > 1
)
```
returned 0. I also verified no orphaned child rows and no constraint-violating duplicates.

Release 2: this PR
* contains only the unique-index migration. Because the merge already ran in a prior release, by the time this migration's preSync hook runs the column is clean, so the transactional CREATE UNIQUE INDEX passes its safety gate.
* Before I merge this into prod, i'll rerun the one-off script again just to be extra sure there's no drift from live data (shound't be)

PR #6753, opened 2026-07-07
Mntor 5299/ Fix + Add Sentry captureException Function Calls

GregHilstonMozilla:
# References:

Jira: [MNTOR-5299](https://mozilla-hub.atlassian.net/browse/MNTOR-5299)

# Description

## `syncBreaches` Bug Fix

Fix an existing `captureException` call in the breach-sync cron so the fields we attach (`breachName`, `breachDomain`, `logoFilename`) actually reach Sentry.

We were accidentally passing those fields under a `data` property. The [Sentry docs for `captureException`](https://docs.sentry.io/platforms/javascript/guides/node/configuration/apis/#captureException) only show a `CaptureContext`, whose `contexts` property is what renders on an issue:

​```ts
contexts?: Record<string, Record<string, unknown>>,
​```

But the second argument actually accepts either a `CaptureContext` or an [`EventHint`](https://github.com/getsentry/sentry-javascript/blob/10.53.1/packages/core/src/utils/prepareEvent.ts#L18-L22), and `data` belongs to `EventHint`. Which Sentry never transmits...

Since `{ data }` is a valid `EventHint`, the types matched and nothing errored, so it was dropped silently. The fix moves the fields under `contexts`.

## `emailBreachAlerts` Enhanced Error Capturing

Add a Sentry `captureException` invocation in our recurring cron job that is responsible for emailing individuals of breach alert.

Add a catch at the message handler boundary that reports whole message failures to Sentry and re throws to preserve nack on error (redelivery). This is distinct from the per recipient catch, which reports a single failed send and continues.

# How I Tested

I updated and added a new unit test to verify the shape of the data that we pass to the Sentry library.

However, since our previous call of `captureException` was unit tested in this way, yet Sentry was silently dropping data that we sent, I went a step further and wrote a throwaway script that invokes the three `captureException` function calls. These were all capture in Sentry and linked below:

1. [reproduction of the original broken syncBreaches captureException invocation](https://mozilla.sentry.io/issues/7598129016/). Notice there is an error displaying the data in this issue, that's demonstrating the issue. 
2. [now fixed syncBreaches captureException invocation](https://mozilla.sentry.io/issues/7598128998/). Notice that we now see `breach` in the context? That's demonstrating the issue being fixed.
3. [new emailBreachAlerts exception](https://mozilla.sentry.io/issues/7598129011/). Notice we also have `breachAlert` in the context here as well. This is showing that the data is being ingested. (Note this shows `subscriber_id` as a data point, but I removed that in a later commit.)

[MNTOR-5299]: https://mozilla-hub.atlassian.net/browse/MNTOR-5299?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/scripts/cronjobs/emailBreachAlerts/emailBreachAlerts.tsx
```diff
@@ -225,13 +225,14 @@ export async function breachMessageHandler(
       );
       notified += 1;
     } catch (error) {
-      sentry?.addBreadcrumb({
-        data: {
-          subscriber_id: recipient.subscriber_id,
-          breach_id: breachAlert.Id,
+      logger.error("Failed to notify user of breach: ", error);
+      sentry?.captureException(error, {
+        contexts: {
```
GregHilstonMozilla, line 230: Converting from a breadcrumb to an actual `captureException`, as it is much more appropriate in a `catch` block.

Notice we use the `contexts` property here, so Sentry does not filter out our adhoc data.

src/scripts/cronjobs/emailBreachAlerts/emailBreachAlerts.tsx
```diff
@@ -286,7 +287,12 @@ export function runJob(config: EmailBreachAlertsJobConfig) {
   });
   const messageFn = async (message: Message) => {
     return await Sentry.withIsolationScope(async () => {
-      return breachMessageHandler(message, ...config.messageFnOpts);
+      try {
+        return await breachMessageHandler(message, ...config.messageFnOpts);
+      } catch (error) {
+        Sentry.captureException(error);
```
GregHilstonMozilla, line 293: just adding another spot to capture an exception

src/scripts/cronjobs/syncBreaches/syncBreaches.ts
```diff
@@ -97,7 +97,9 @@ export async function getBreachIcons(breaches: HibpGetBreachesResponse) {
         logoFilename,
       });
       captureException(error, {
-        data: { breachName, breachDomain, logoFilename },
+        contexts: {
```
GregHilstonMozilla, line 100: This additional of the contexts property is the fix required for Sentry to ingest our adhoc data

src/scripts/cronjobs/emailBreachAlerts/index.ts
GregHilstonMozilla, line 1: Adds a status update for this cron job running, which will toggle between `in progress`, `ok`, and `done`.

joeherm (approved):
Looks good- thanks for the links to Sentry exceptions you generated.

PR #6766, opened 2026-07-13
feat(notify): log full Pub/Sub publish error

GregHilstonMozilla:
# References:

Jira: [MNTOR-5311](https://mozilla-hub.atlassian.net/browse/MNTOR-5311)

# Description

Diagnostic-only. 

Breach-alert emails stopped ~2026-05-24 because the HIBP notify webhook's Pub/Sub `publishMessage()` fails ~100% of the time, but it was logged with too little detail to diagnose. The logging library Winston drops the gRPC error's `message`/`stack`, and Sentry titled it the useless `undefined undefined: undefined`.

One file changes (`route.ts`). On the first publish failure per pod, the `catch` now:

- Dumps the real error — `message` / `stack` / a full `util.inspect` (`cause` / `statusDetails` /
  gRPC metadata).
- **Probes the failing client's own credentials** to test the workload-identity suspect: project id, universe domain, service-account email, and whether a token can be obtained. Never logs the access token or private key.
- Attaches the above to the Sentry issue and fingerprints it so it stops hiding under `undefined undefined: undefined`.

Every failure still logs the cheap `code`/`details`. No behavior change, still returns 429 and increments the `pubsub-error` metric.

# How to test

- `npm run lint` passes.
- Stage: smoke only (app boots, notify route responds). Stage gets no organic HIBP notify traffic,   so the bug surfaces in prod, not stage.
- Prod: on the next breach-load burst, Cloud Logging shows the new fields. Filter: `resource.labels.namespace_name="monitor-prod" jsonPayload.message="error_queuing_hibp_breach:"`

- [x] Unit test — n/a: logging/diagnostics only; the failure can't be reproduced off-cluster.
- [x] Localization strings — n/a (no UI).

[MNTOR-5311]: https://mozilla-hub.atlassian.net/browse/MNTOR-5311?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/app/api/v1/hibp/notify/route.ts
+++ b/src/app/api/v1/hibp/notify/route.ts
@@ -24,18 +24,18 @@ import { inspect } from "node:util";
 // (Google API Extensions) errors
 // carry their real signal in non-enumerable Error props (message/stack) and nested fields
 // (cause / statusDetails / the gRPC metadata map) that winston's JSON format drops, so a full
-// util.inspect is the only faithful dump; it also runs the one-time credential probe below.
+// util.inspect is the only faithful dump; it also runs the one-time credential check below.
 // Every failure still logs the cheap top-level code/details in case the failure mode changes.
 let pubsubErrorInspected = false;
 
 /**
- * One-time diagnostic probe of the failing Pub/Sub client's own credentials, to test whether the
+ * One-time diagnostic check of the failing Pub/Sub client's own credentials, to test whether the
  * workload-identity path is what's breaking publish (the leading suspect for the all-undefined
  * gRPC status). Never throws, and never emits the access token or private key — only non-secret
  * identity/infra values (project id, universe domain, service-account email) plus whether a token
  * could be obtained and its length. If a call throws, its inspected error is the likely root cause.
  */
-async function probePubSubCredentials(
+async function tryPubSubCredentials(
   pubsub: ReturnType<typeof getPubSub>,
 ): Promise<Record<string, unknown>> {
   const auth = pubsub.auth;
@@ -120,7 +120,7 @@ export async function POST(req: NextRequest) {
     // gRPC/gax errors hide their real signal in non-enumerable/nested fields that winston's JSON
     // format drops (the log used to collapse to an opaque {note, metadata:{}}). Log the cheap
     // top-level code/details every time; dump everything verbose (message/stack/full inspect)
-    // plus the one-time credential probe once per process (see pubsubErrorInspected above).
+    // plus the one-time credential check once per process (see pubsubErrorInspected above).
     const err = ex as {
       code?: unknown;
       details?: unknown;
@@ -139,7 +139,7 @@ export async function POST(req: NextRequest) {
       fields.message = err?.message;
       fields.stack = err?.stack;
       fields.inspect = inspect(ex, { depth: 8, breakLength: Infinity });
-      fields.credentialProbe = await probePubSubCredentials(pubsub);
+      fields.credentialCheck = await tryPubSubCredentials(pubsub);
     }
     logger.error("error_queuing_hibp_breach:", fields);
     incHibpNotifyFailure("pubsub-error");
--- a/src/app/api/v1/hibp/notify/route.ts
+++ b/src/app/api/v1/hibp/notify/route.ts
@@ -20,11 +20,49 @@ import * as Sentry from "@sentry/nextjs";
 import { getPubSub } from "../../../../gcp/clients";
 import { inspect } from "node:util";
 
-// Dump a full util.inspect of the publish error once per process — it surfaces nested /
-// non-enumerable gax (Google API Extensions) fields (cause / statusDetails / the metadata map)
-// the explicit field
-// extraction can't. The concise fields (code / details / message / stack) log every failure.
+// The publish failure is logged verbosely only once per process (this guard). gRPC/gax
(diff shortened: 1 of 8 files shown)


src/app/api/v1/hibp/notify/route.ts
```diff
@@ -18,6 +18,12 @@ import {
 
 import * as Sentry from "@sentry/nextjs";
 import { getPubSub } from "../../../../gcp/clients";
+import { inspect } from "node:util";
```
groovecoder, line 21: question (non-blocking): is this the only `import` statement like this? I don't recognize this `import` ... `node:util` syntax.

src/app/api/v1/hibp/notify/route.ts
```diff
@@ -18,6 +18,12 @@ import {
...
+
+// Dump a full util.inspect of the publish error once per process — it surfaces nested /
+// non-enumerable gax fields (cause / statusDetails / the metadata map) the explicit field
```
groovecoder: question (non-blocking): what's a "gax" field?

src/app/api/v1/hibp/notify/route.ts
```diff
@@ -72,7 +78,29 @@ export async function POST(req: NextRequest) {
     });
     return NextResponse.json({ success: true }, { status: 200 });
   } catch (ex) {
-    logger.error("error_queuing_hibp_breach:", { topicName, exception: ex });
+    // gRPC/gax errors don't JSON-serialize their useful fields (winston's default json format
```
groovecoder: question (non-blocking): what's "winston"? sounds like a logging library?

src/app/gcp/clients.ts
```diff
@@ -4,15 +4,19 @@
...
 import { config } from "../../config";
 import { PubSub } from "@google-cloud/pubsub";
+import { logger } from "../functions/server/logging";
 
 let _pubSub: PubSub | undefined = undefined;
 
 export function getPubSub(): PubSub {
   if (_pubSub !== undefined) return _pubSub;
 
+  const enableOpenTelemetryTracing = config.gcp.pubsub.enableOtelTracing;
```
groovecoder: question (non-blocking): will this need a corresponding change in webservices-infra repo to set it to `true` now that it's not hard-coded anymore?

groovecoder (approved):

src/app/api/v1/hibp/notify/route.ts
groovecoder: suggestion (non-blocking): spell this out as "Google API Extensions".

src/app/gcp/clients.ts
groovecoder: question (non-blocking): @GregHilstonMozilla - do we want this to be `true` for all environments?

GregHilstonMozilla: Yeah, I think we'll want this to be `true` for all environments, which should be what the situation is today. 

Then, we can temporarily set it to `false` to further debug if the errors are not immediately useful to us

mansaj (approved):

src/app/api/v1/hibp/notify/route.ts
```diff
@@ -20,12 +20,50 @@ import * as Sentry from "@sentry/nextjs";
...
 let pubsubErrorInspected = false;
 
+/**
+ * One-time diagnostic probe of the failing Pub/Sub client's own credentials, to test whether the
+ * workload-identity path is what's breaking publish (the leading suspect for the all-undefined
+ * gRPC status). Never throws, and never emits the access token or private key — only non-secret
+ * identity/infra values (project id, universe domain, service-account email) plus whether a token
+ * could be obtained and its length. If a call throws, its inspected error is the likely root cause.
+ */
+async function probePubSubCredentials(
```
groovecoder: quibble (non-blocking): re-name to `tryPubSubCredentials` so this doesn't seem like it's doing a credential-stuffing attack. `checkPubSubCredentials`, `inspectPubSubCredentials`?

src/app/api/v1/hibp/notify/route.ts
```diff
@@ -20,12 +20,50 @@ import * as Sentry from "@sentry/nextjs";
...
+  };
+  await capture("projectId", () => auth.getProjectId());
+  await capture("universeDomain", () => auth.getUniverseDomain());
+  await capture(
+    "clientEmail",
+    async () => (await auth.getCredentials()).client_email,
+  );
+  await capture("accessToken", async () => {
+    // NEVER log the token itself — only whether one was obtained and its length.
+    const token = await auth.getAccessToken();
```
groovecoder, line 62: There is a small residual risk in the error dumps. Each `capture` failure does `inspect(e, { depth: 4 })`, and the publish `catch` does `inspect(ex, { depth: 8 })`. Both go to winston and to Sentry via `extra`. A gRPC publish error can carry the outgoing request metadata, which includes the `authorization` bearer token. A deep inspect could surface it. Low probability, not zero, and Sentry is a third-party sink.

suggestion(non-blocking): redact keys matching `authorization`/`token`/`key`/`secret` in the inspected strings before logging. Cheap, keeps the diagnostic value, and honors the "never emits the access token" promise the code comments already make.

groovecoder (approved):

PR #6771, opened 2026-07-20
fix: rehydrate breach dates so alert emails render

GregHilstonMozilla:
# References:

Jira: [MNTOR-5317](https://mozilla-hub.atlassian.net/browse/MNTOR-5317)

_This is a follow up to [MNTOR-5314](https://mozilla-hub.atlassian.net/browse/MNTOR-5314) / [#6767 (the Pub/Sub publish fix)](https://github.com/mozilla/blurts-server/pull/6767)._

TLDR: A few properties could either be strings or dates, depending on if a cache was hit or missed. The former causes an exception to bubble up, causing HIBP breach messages to get stuck in our dead letter queue. This was introduced in a PR 7 weeks ago.

# Description

## Problem

The MNTOR-5314 publish fix ([#6767](https://github.com/mozilla/blurts-server/pull/6767)) worked. `POST /api/v1/hibp/notify` publishes again and the `monitor-breach-alerts` consumer pulls messages. But the first breach to reach the consumer since that fix, [Paidwork (added 2026-07-19)](https://monitor.mozilla.org/en/breach-details/Paidwork), crashes on email render instead of sending:

```
TypeError: props.breach.BreachDate.toLocaleDateString is not a function
  job: emailBreachAlerts   breachName: Paidwork
```

The consumer writes an `email_notifications` incident row with `notified=false` *before* rendering (`emailBreachAlerts.tsx`), then throws. The message is nacked, redelivered, and after max delivery attempts routed to the dead letter queue. This fired a "HIBP Breach Alert Deadletter Delivery Rate" alert on 2026-07-19 19:32.

*Impact:* Every notifiable breach ingested since [#6767](https://github.com/mozilla/blurts-server/pull/6767) deployed is un-alerted. Paidwork has 2,000+ incident rows, all `notified=false`, 0 sent, and the count is still climbing as messages redeliver.

## Root cause

`HibpLikeDbBreach.BreachDate` is typed (and documented) as a `Date`, but on the Redis cache hit path it is a `string` at runtime. `BreachDataService.readFromCache` does `JSON.parse` on the cached breaches, which turns every `Date` column into an ISO string, and `dbToHibp` copied `breach_date` straight through without rehydrating it. The breach alert email template (`BreachAlertEmail.tsx`, and `Ver2/3/4`) calls `BreachDate.toLocaleDateString()`, a `Date`only method, so it throws on a string. 

Cache miss returns a real `Date` (node-postgres), so it only fails on cache hits, which is the hot path for the long lived consumer, making the crash deterministic.

The fix wraps each date column in `new Date(...)` in `dbToHibp`, which rehydrates the ISO string on the cache hit path and is a harmless clone of the already Date value on the cache miss path.

## Evidence

- Sentry (prod): the `TypeError` above, tagged `job: emailBreachAlerts` / `breachName: Paidwork`.
- Prod DB (BigQuery over `email_notifications` + `breaches`): Paidwork has incident rows but 0 `notified=true` (consumed, then render crashed), vs the 22 earlier outage breaches at 0 incident rows (never published). Two distinct failure modes.

## Timeline (all UTC)

- **2026-05-07**: the `BreachDate.toLocaleDateString()` line is added to the breach-alert email template ([MNTOR-5269](https://mozilla-hub.atlassian.net/browse/MNTOR-5269), `b06106334`); before this the template did not render the breach date. Dormant until it deploys.
- **2026-05-26**: `prod-2026.05.26` deploys, carrying *both* the template change and the protobufjs bump that broke publishing ([MNTOR-5314](https://mozilla-hub.atlassian.net/browse/MNTOR-5314)). Publishing dies, so the consumer never runs, this bug is masked.
- **2026-07-15**:  [#6767](https://github.com/mozilla/blurts-server/pull/6767) ([MNTOR-5314](https://mozilla-hub.atlassian.net/browse/MNTOR-5314)) merges and later deploys; publishing is restored and the consumer starts pulling again.
- **2026-07-19**: Paidwork is ingested, reaches the consumer, and crashes on render; incident rows pile up at `notified=false` and messages exhaust into the DLQ; the deadletter alert fires 19:32.

## Why it wasn't caught

Two reasons it stayed hidden: the publish outage meant the render path never executed in prod for ~7 weeks, so the latent template bug never fired and the template unit tests / Storybook stories pass a `Date` mock, while dev/CI use the Pub/Sub emulator, neither exercises a real Redis cache hit where `BreachDate` is a string.

## Recovery / backfill (follow-up, not in this PR)

Since the consumer got far enough to write `email_notifications` rows before crashing, Paidwork's data is preserved. Per the [breach alerts runbook](https://mozilla-hub.atlassian.net/wiki/x/UIAwd), once this fix is deployed we can either replay the dead letter messages back to the topic (they still hold the original `{breachName, hashPrefix, hashSuffixes}` payloads, within DLQ retention), or, past retention, regenerate HIBP-like payloads from the `notified=false` rows via the runbook's BigQuery query and publish them. The pipeline dedupes on `notified=true`, so a redrive won't double send. 

I have no preference on which path we take here, ideally whichever is simpler. I'm thinking that will be the first one.

# How to test

1. Regression test: `npx vitest run src/services/BreachDataService.test.ts` -> passes.
2. Reproduce the bug: revert the `dbToHibp` change -> the new test fails with `expected '<iso string>' to be an instance of Date`, and `() => result.BreachDate.toLocaleDateString()` throws.
3. Confirm no collateral: `npx vitest run src/emails/templates/breachAlert/ src/scripts/cronjobs/emailBreachAlerts/` -> green.

- [x] Commits are minimal and descriptive (one transform fix + a regression test).
- [x] I've added or updated the relevant code comments (why the date columns must be rehydrated).
- [x] I've added a unit test to test for potential regressions of this bug (`BreachDataService.test.ts`, cache-hit Date rehydration).

[MNTOR-5317]: https://mozilla-hub.atlassian.net/browse/MNTOR-5317?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[MNTOR-5314]: https://mozilla-hub.atlassian.net/browse/MNTOR-5314?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[MNTOR-5269]: https://mozilla-hub.atlassian.net/browse/MNTOR-5269?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


mansaj (approved):

src/services/BreachDataService.test.ts
```diff
@@ -76,6 +76,44 @@ describe("BreachDataService factory", () => {
...
+      // Regression guard. On a cache hit the breach is round tripped through
+      // Redis, and JSON.parse turns the Date columns into ISO strings. The
+      // breach alert email template calls `breach.BreachDate.toLocaleDateString()`,
+      // which throws "toLocaleDateString is not a function" on a string. dbToHibp
+      // must rehydrate the date columns so the runtime matches the declared type.
+      const fakeBreach = seeds.breaches();
+      // Mirror the real cache write path: JSON.stringify serialises Dates to strings.
+      await redis.set(REDIS_ALL_BREACHES_KEY, JSON.stringify([fakeBreach]));
+      const mockSync: BreachSyncService = {
+        syncBreaches: vi.fn().mockResolvedValue(undefined),
```
groovecoder, line 89: question (non-blocking): What is this `syncBreaches` function? Why mock it instead of running the real function?

src/services/BreachDataService.test.ts
```diff
@@ -76,6 +76,44 @@ describe("BreachDataService factory", () => {
...
+      };
+      const getBreachesFromDb = vi.fn().mockResolvedValue([]);
```
groovecoder, line 91: question (non-blocking): why mock this getBreachesFromDb instead of running the real function?

src/services/BreachDataService.test.ts
```diff
@@ -76,6 +76,44 @@ describe("BreachDataService factory", () => {
...
+        sync: mockSync,
+        getBreachesFromDb,
+        logger,
+      });
+      const result = await service.getBreach(fakeBreach.name);
+      expect(result).not.toBeUndefined();
+      expect(result!.BreachDate).toBeInstanceOf(Date);
+      expect(result!.AddedDate).toBeInstanceOf(Date);
+      expect(result!.ModifiedDate).toBeInstanceOf(Date);
+      // ...and the rehydrated instants match the source, so a value-shifting
```
groovecoder: quibble (non-blocking): "instants" ? is that just the proper word for "instances"?

src/utils/hibp.ts
```diff
@@ -249,9 +249,16 @@ function dbToHibp(breach: BreachRow): HibpLikeDbBreach {
...
-    BreachDate: breach.breach_date,
-    AddedDate: breach.added_date,
-    ModifiedDate: breach.modified_date,
+    // Rehydrate the date columns into Date objects. `breach` arrives here either
+    // straight from Postgres (already a Date) OR from the Redis breaches cache,
+    // where a JSON.stringify/JSON.parse round trip has turned every Date into an
+    // ISO string. Consumers (e.g. the breach alert email template's
+    // `BreachDate.toLocaleDateString()`) rely on the declared Date type, so
+    // normalise both cases to a Date here. `new Date(existingDate)` is a safe
+    // clone. See MNTOR-5317.
```
groovecoder: change (blocking): let's put this verbose explanation in a single place like the git commit message, the pull request, or the jira ticket ... and then use a small comment like "See ___" to send readers to the full details.

I'm thinking in particular that LLM tools reading this code will over-rotate and over-fixate on this content & context in sessions that are not related to this code or change.

groovecoder (changes requested):
comment (non-blocking): leaving inline questions that I’ll then feed back into my own Claude Code to get as many answers for myself as I can. so don’t worry about answering anything that’s non-blocking.

src/services/BreachDataService.test.ts
groovecoder: suggestion (non-blocking): maybe use `timestamps` here to avoid the confusion?

groovecoder:
Comment (non-blocking): Answering my own question so @GregHilstonMozilla doesn't have to: mocking `syncBreaches` and `getBreachesFromDb` here is the right call. This is the cache-hit path, so `getBreach` returns before it reaches either one. They're just injected seams.

The deeper reason is structural: `BreachDataService` depends on Redis, and this repo has no real-Redis test path. `docker-compose.yml` has Postgres but no Redis, and every Redis test uses `ioredis-mock` (a deliberate choice — [MNTOR-4154](https://mozilla-hub.atlassian.net/browse/MNTOR-4154)). So this service can only live in the unit-tier of tests, where its DB and sync deps get mocked. The integration-tier tests (`*.integration.ts`) run against real Postgres, but Redis has no equivalent.

For this specific bug it doesn't matter — the `Date`→string coercion happens at `JSON.stringify` in app code, and the test reproduces it faithfully. But as noted in the PR: "the template unit tests / Storybook stories pass a Date mock".

This is a common problem with unit-tier test suites that mock too much - these kinds of real-system bugs slip thru the test suite. The stronger regression guard would round-trip a real Postgres row thru a real Redis cache.

@GregHilstonMozilla : could you file a follow-up ticket to add a Redis service to the integration tier tests (`docker-compose` + `.env.ci`) so `BreachDataService`/`BreachSyncService` can be tested end-to-end? And in the same ticket we should see if we can actually remove `ioredis-mock` completely. It makes sense to mock out external 3rd-party systems like HIBP, but Redis is a 1st-party system the Monitor, so we should test with it - not without it.

Not a blocker for this PR.

[MNTOR-4154]: https://mozilla-hub.atlassian.net/browse/MNTOR-4154?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

GregHilstonMozilla:
@groovecoder As requested, I captured the follow up ticket in [MNTOR-5318](https://mozilla-hub.atlassian.net/browse/MNTOR-5318)

[MNTOR-5318]: https://mozilla-hub.atlassian.net/browse/MNTOR-5318?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

groovecoder (approved):
quibble (non-blocking): I slightly prefer re-basing PR branches on main rather than merging main onto the branch, but I don't think we have a rule for this across our repos. So just a personal preference.

PR #6817, opened 2026-08-25
Mntor 5342

GregHilstonMozilla:
# References:

Jira: [MNTOR-5342](https://mozilla-hub.atlassian.net/browse/MNTOR-5342)

# Description

Rewrites `openapi.yml`, the contract between Monitor and the Firefox privacy panel. This is to support the Firefox Breach Integration Phase 2

Must not be merged until #6818 is merged, and deployed

The substantive changes:

- **Thin response.** `{ id, breachedAccounts }` only. Firefox already has breach metadata from Remote Settings and joins on the HIBP name.
- **Scoped server-side**, no data-class filters.
- **Serves email-only breaches**, not just password breaches. (This is what depends on #6818 )
- **Errors have a stable `code`** to branch on, a documented body, and examples.
- **A failed HIBP lookup is a `503`.** It used to be an empty `200`, which would have told the panel to clear a user's breach state during an outage.

Also adds `redocly.yaml` and a README section on validating the contract.

# How to test

`npx @redocly/cli@2 lint` passes. No code changes, nothing to run.

Render the docs to read it as Firefox will: `npx @redocly/cli@2 preview-docs openapi.yml`

OR use something like [editor.swagger.io](https://editor.swagger.io/)

[MNTOR-5342]: https://mozilla-hub.atlassian.net/browse/MNTOR-5342?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/openapi.yml
+++ b/openapi.yml
@@ -48,6 +48,7 @@ components:
             - invalid-request
             - not-found
             - rate-limited
+            - internal-error
             - unavailable
         message:
           type: string
@@ -135,6 +136,15 @@ components:
           example:
             code: rate-limited
             message: Too many requests. Retry after 60 seconds.
+    InternalError:
+      description: Something failed on our side. The request may not have been applied. Retry, and keep any state you already hold.
+      content:
+        application/json:
+          schema:
+            $ref: "#/components/schemas/Error"
+          example:
+            code: internal-error
+            message: Monitor could not complete this request.
     ServiceUnavailable:
       description: Monitor is unavailable. Safe to retry, after the interval in `Retry-After` when it is present.
       headers:
@@ -223,6 +233,8 @@ paths:
           $ref: "#/components/responses/Forbidden"
         "429":
           $ref: "#/components/responses/TooManyRequests"
+        "500":
+          $ref: "#/components/responses/InternalError"
         "503":
           $ref: "#/components/responses/ServiceUnavailable"
           description: We could not check this user against HIBP, so we do not know what they are affected by. Keep the breaches you already hold, do not clear them, and retry.
@@ -330,5 +342,7 @@ paths:
                 message: One or more of the referenced breaches or email addresses was not found.
         "429":
           $ref: "#/components/responses/TooManyRequests"
+        "500":
+          $ref: "#/components/responses/InternalError"
         "503":
           $ref: "#/components/responses/ServiceUnavailable"
--- a/openapi.yml
+++ b/openapi.yml
@@ -2,11 +2,16 @@ openapi: "3.1.1"
 info:
   title: Monitor Breach Alerts API
   description: Monitor's external API for Firefox clients
-  version: "1.0"
+  version: "1.0.0"
+# The web app already owns /api/v1.
 servers:
-  - url: https://monitor.mozilla.org/api/v1
+  - url: https://monitor.mozilla.org/api/firefox/v1
     description: Production endpoint
-  - url: http://localhost:6060/api/v1
+  - url: https://monitor-stage.allizom.org/api/firefox/v1
+    description: Stage endpoint
+  - url: https://dev.monitor.nonprod.webservices.mozgcp.net/api/firefox/v1
+    description: Dev endpoint
+  - url: http://localhost:6060/api/firefox/v1
     description: Local development server
 security:
   - bearerAuth: []
--- a/openapi.yml
+++ b/openapi.yml
@@ -161,6 +161,8 @@ paths:
       description: |
         Breaches that leaked passwords or email addresses, with resolution state for each monitored email the breach affected.
 
+        Sensitive breaches, as HIBP marks them, are never returned. The privacy panel must not surface one, so there is no parameter to ask for them. Remote Settings does carry them, flagged, because the credential manager is allowed to.
+
         **Freshness.** When HIBP is unavailable we serve the last known breach data as a normal `200`, with no staleness signal.
 
         **Cadence.** Poll no more than hourly, and once a day is enough. Breaches reach HIBP weeks or months after the incident, so polling harder gains nothing, and the `429` threshold is set with this range in mind.
@@ -177,14 +179,6 @@ paths:
             example:
               summary: Example domain
               value: example.com
-        - name: includeSensitive
-          in: query
-          required: false
-          description: Include breaches marked sensitive.
-          schema:
-            type: boolean
-            default: false
-
       responses:
         "200":
           description: |
--- a/openapi.yml
+++ b/openapi.yml
@@ -53,7 +53,7 @@ components:
       description: A data class we report as resolved. `passwords` and `email-addresses` today, but this is an open set, so accept a value you do not recognise instead of rejecting the response.
       type: string
     BreachResolution:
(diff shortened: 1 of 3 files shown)


openapi.yml
GregHilstonMozilla, line 1: I highly recommend you do not try to read this file manually, or the diff. But instead use a tool like redocly (see the README.md changes for documentation) or https://editor.swagger.io/

openapi.yml
```diff
@@ -204,121 +233,129 @@ paths:
...
       description: |
-        Bulk endpoint for breach resolutions.
+        Mark data classes resolved for one or more breaches. Resolutions only move one way, this API cannot unresolve a data class.
 
-        Upsert resolution state for one or more breaches across one or more monitored email addresses.
+        Idempotent per `(email, dataClass, breachId)`. Repeated `breachId` entries merge, and order does not matter.
 
-        This operation is idempotent per `(email, dataClass, breachId)`. Repeating the same
-        request results in the same stored state and does not produce an error.
+        Atomic: every targeted email and data class updates, or none does.
```
mansaj, line 278: Atomicity could be a problem based on the code today..

specifically, resolution state lives in a single JSONB column `subscribers.breach_resolution` - sadly this is architectural failure/pain from legacy codebase. 

There is no transaction and no optimistic concurrency control. Two concurrent Firefox writes, or one Firefox write racing the website's PUT /api/v1/user/breaches, will silently lose one of them

NOTE: this might entail implementing the write inside a transaction with SELECT ... FOR UPDATE on the subscriber row (smallest fix to preserves the JSONB shape) .. or to do more db schema change and data migration.. which will be painful but better in the long run

openapi.yml
```diff
@@ -15,185 +28,201 @@ components:
...
-          type: string
-          format: date-time
-        dataClasses:
-          description: >
-            This attribute describes the nature of the data compromised in the breach and contains an alphabetically
-            ordered string array of impacted data classes. See https://haveibeenpwned.com/api/v3/dataclasses for list of dataclasses.
+    DataClass:
+      description: A kind of data leaked in a breach. Remote Settings serves the same values in HIBP's Title Case, `Passwords` and `Email addresses`.
+      type: string
+      enum:
```
mansaj, line 49: two-value closed enum here. 
I think it's good for the request. On the response it means that the moment Monitor serves a third data class, every strictly-validating generated client rejects the entire payload.. I fear this won't be backwards compatible (as extendable as we'd like)

btw, same issue with `Error.code`

rec: split the type. Keep the closed enum for input, use an open string documented with the current values for output.

openapi.yml
```diff
@@ -15,185 +28,201 @@ components:
...
-            By default, sensitive breaches are excluded.
-          schema:
-            type: boolean
-          examples:
-            include:
-              summary: Include sensitive breaches
-              value: true
-
       responses:
         "200":
```
mansaj, line 215: rec:
```
"200":
  headers:
    Cache-Control:
      description: Always `private, no-store`. This response is scoped to one user and must never be cached by a shared intermediary.
      schema:
        type: string
```

openapi.yml
```diff
@@ -204,121 +233,129 @@ paths:
...
+              example:
+                code: not-found
+                message: One or more of the referenced breaches or email addresses was not found.
+        "429":
+          $ref: "#/components/responses/TooManyRequests"
+        "500":
+          $ref: "#/components/responses/InternalError"
+          description: Something failed on our side. Treat the batch as not applied and retry it, repeats are safe.
+        "503":
+          $ref: "#/components/responses/ServiceUnavailable"
```
mansaj, line 361: I'm reading this and line 346.. seems confusing. Just double checking and make sure they aren't contradictory

mansaj (changes requested):

openapi.yml
mansaj: If I'm reading this correctly.. consolidated, it's trying to say:
```
Every address in `emails` must be monitored by the authenticated user, and affected by the referenced breach.
Omitting `emails` targets every monitored address the breach affected. If that set is empty, the entry resolves nothing and the request is a `404`, the same as naming an unaffected address.
```

openapi.yml
GregHilstonMozilla: Great call out! I don't think this warrants any change on this `openapi.yml`, unless you do.

I have added the following to the acceptance on ticket [MNTOR-5345](https://mozilla-hub.atlassian.net/browse/MNTOR-5345)

> Recall that breach resolutions live in one JSONB column, subscribers.breach_resolution, which we overwrite wholesale, so two concurrent writes (ours, or ours racing the website) silently lose one. A simple solution to this is to do the read and the write in a transaction with SELECT ... FOR UPDATE on the subscriber row. This came out of this discussion https://github.com/mozilla/blurts-server/pull/6817#discussion_r3917918821

openapi.yml
GregHilstonMozilla: Ah good idea, taking this advice

openapi.yml
GregHilstonMozilla: Added, thanks!

openapi.yml
GregHilstonMozilla: Yeah, you're right, my writing was not the best. I took another pass at simplifying it, to make it more clear

mansaj (approved):

openapi.yml
```diff
@@ -15,185 +28,200 @@ components:
...
+
+        **Cadence.** Poll no more than hourly, and once a day is enough. Breaches reach HIBP weeks or months after the incident, so polling harder gains nothing.
 
       parameters:
         - name: domain
           in: query
           required: false
-          description: >
-            Filter breaches by the breached site's primary domain (exact match), e.g. "example.com".
+          description: Filter by the breached site's primary domain, matched exactly. A breach filed under `twitter.com` does not match `x.com`.
```
Vinnl, line 199: Since (I think?) we're not planning to use this, would it maybe be safer to leave it out? So that future contributors can be sure that it's unused.

openapi.yml
```diff
@@ -15,185 +28,200 @@ components:
...
-        isSensitive:
-          description: Indicates if the breach is considered [sensitive](https://haveibeenpwned.com/FAQs#SensitiveBreach)
-          type: boolean
+            format: email
+        dataClasses:
+          type: array
+          minItems: 1
+          description: Data classes to mark resolved for the selected emails. Each must be one the referenced breach actually leaked, so a breach that leaked only email addresses takes `email-addresses`, not `passwords`. Sending one it did not leak is a `400`, so refresh your Remote Settings copy before retrying.
+          items:
+            $ref: "#/components/schemas/DataClass"
```
Vinnl: Maybe we should leave those out? I think in the designs I've seen, the user can only resolve everything at once (or at least, email and password leaks for all affected emails), so it might not be worth taking up the maintenance burden.

Vinnl (approved):
Not an expert in API design, but the general approach seems right to me and workable from the frontend.

PR #3879, opened 2023-12-14
Add missing breach data classes

pdehaan:
# References: 
Jira: MNTOR-
Figma: 

# Description

```sh
npx pdehaan/blurts-server-data-classes

Missing "telecommunications-carrier" data class from ApexSMS breach
Missing "loyalty-program-details" data class from MalindoAir breach
```

Not visible on the details page, but noticed them when browsing on https://monitor.firefox.com/breaches (search for "Apex" or "Mali").

# Screenshot (if applicable)

Not applicable.

# How to test

# Checklist (Definition of Done)
- [ ] Localization strings (if needed) have been added.
- [ ] Commits in this PR are minimal and [have descriptive commit messages](https://chris.beams.io/posts/git-commit/).
- [ ] I've added or updated the relevant sections in readme and/or code comments
- [ ] I've added a unit test to test for potential regressions of this bug.
- [ ] Product Owner accepted the User Story (demo of functionality completed) or waived the privilege.
- [ ] All acceptance criteria are met.
- [ ] Jira ticket has been updated (if needed) to match changes made during the development process.
- [ ] Jira ticket has been updated (if needed) with suggestions for QA when this PR is deployed to stage.

PR #6818, opened 2026-08-27
Mntor 5363

GregHilstonMozilla:
# References:

Jira: [MNTOR-5363](https://mozilla-hub.atlassian.net/browse/MNTOR-5363)

# Description

We are going to be iterating on our [openapi.yml](https://github.com/mozilla/blurts-server/blob/main/openapi.yml) contract, which is defined between the frontend and the backend for a breach integration. To achieve that with a thin and simple API, we'll be relying on Remote Settings data that is already being pushed. That dataset was missing records.

The sync previously only carried breaches that leaked passwords, so a breach that leaked a user's email but not their password would be invisible to Firefox and the privacy panel could never render it. This PR widens the periodic cron that pushes this data.

The rest of the PR is the types and doc strings that came with that change. HIBP spells data classes differently to us, and nothing in the codebase recorded that.

## Changed since the first review: no `IsSensitive`

This PR originally also wrote an `IsSensitive` field onto each new record. [Vincent and I agreed it is not needed](https://mozilla-hub.atlassian.net/browse/MNTOR-5363?focusedCommentId=1715164): phase 2 only shows breach alerts for breaches the API returns as affecting the user, and the API does not return sensitive breaches, so Firefox never sees one to hide.

# How to test

```sh
npx vitest run src/scripts/cronjobs/updateBreachesInRemoteSettings/ src/utils/hibp.test.ts
```

- [x] I've added or updated the relevant sections in readme and/or code comments. Block comments added to `BreachDataTypes`, `ResolutionRelevantBreachDataTypes` and `HibpLabelByDataType`, which sit together and were easy to confuse.
- [x] If this PR implements a feature flag or experimentation, I've checked that it still works with the flag both on, and with the flag off. Not applicable.
- [ ] All acceptance criteria are met. The remaining AC is a post-deploy check on the live collection.
- [x] Jira ticket has been updated (if needed) to match changes made during the development process. Summary, description and AC updated to drop `IsSensitive`.

[MNTOR-5363]: https://mozilla-hub.atlassian.net/browse/MNTOR-5363?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/app/functions/universal/breach.ts
+++ b/src/app/functions/universal/breach.ts
@@ -6,6 +6,11 @@ import { DataClassEffected } from "../../../utils/subscriberBreaches";
 
 // TODO: Move pure functions that operate on breaches to this file
 
+/**
+ * Every kind of data a breach can leak, in Monitor's own kebab-case spelling.
+ * This is what we store, what a HibpLikeDbBreach carries, and what our APIs
+ * speak. See HibpLabelByDataType for how HIBP spells the same things.
+ */
 export const BreachDataTypes = {
   Passwords: "passwords",
   Email: "email-addresses",
@@ -22,6 +27,39 @@ export const BreachDataTypes = {
   General: "general",
 } as const;
 
+/**
+ * The same data types, spelled the way HIBP spells them. Use these only against
+ * a raw HibpGetBreachesResponse. Anything that came from our database already
+ * speaks BreachDataTypes, because formatDataClass converts it at ingestion.
+ *
+ * Keys match BreachDataTypes, so one data type's two spellings line up on one
+ * key. This cannot be a transform, the difference is not mechanical: `pins` is
+ * `PINs` and `ip-addresses` is `IP addresses`. `General` is ours alone with no
+ * HIBP counterpart, hence the Omit.
+ */
+export const HibpLabelByDataType = {
+  Passwords: "Passwords",
+  Email: "Email addresses",
+  SSN: "Social security numbers",
+  CreditCard: "Partial credit card data",
+  BankAccount: "Bank account numbers",
+  PIN: "PINs",
+  IP: "IP addresses",
+  Address: "Physical addresses",
+  DoB: "Dates of birth",
+  Phone: "Phone numbers",
+  SecurityQuestions: "Security questions and answers",
+  HistoricalPasswords: "Historical passwords",
+} as const satisfies Omit<
+  Record<keyof typeof BreachDataTypes, string>,
+  "General"
+>;
+
+/**
+ * The subset of BreachDataTypes that Monitor asks a user to act on, and so the
+ * only ones that appear in the guided resolution flow and in breach alert
+ * emails. The rest are shown as context only.
+ */
 export const ResolutionRelevantBreachDataTypes = {
   Passwords: BreachDataTypes.Passwords,
   Email: BreachDataTypes.Email,
--- a/src/app/functions/universal/breach.ts
+++ b/src/app/functions/universal/breach.ts
@@ -7,9 +7,14 @@ import { DataClassEffected } from "../../../utils/subscriberBreaches";
 // TODO: Move pure functions that operate on breaches to this file
 
 /**
- * Every kind of data a breach can leak, in Monitor's own kebab-case spelling.
- * This is what we store, what a HibpLikeDbBreach carries, and what our APIs
- * speak. See HibpLabelByDataType for how HIBP spells the same things.
+ * The data types we name, in our own kebab-case spelling. This is what we
+ * store and what our APIs speak. See HibpLabelByDataType for HIBP's spelling.
+ *
+ * Not the full list. HIBP adds a data type whenever it sees a new kind of
+ * data, and locales/en/data-classes.ftl lists around 170.
+ *
+ * These are roughly the data types the public breach detail page gives advice
+ * for, see getAllPriorityDataClasses in src/utils/recommendations.ts.
  */
 export const BreachDataTypes = {
   Passwords: "passwords",
--- a/src/scripts/cronjobs/updateBreachesInRemoteSettings/updateBreachesInRemoteSettings.test.ts
+++ b/src/scripts/cronjobs/updateBreachesInRemoteSettings/updateBreachesInRemoteSettings.test.ts
@@ -134,9 +134,16 @@ describe("updateBreachesInRemoteSettings job", () => {
 
       await main(mockLog);
 
-      expect(addBreachSpy).toHaveBeenCalledWith(
-        expect.objectContaining({ Name: emailOnly.Name }),
-      );
+      // Pins the whole record
+      expect(addBreachSpy).toHaveBeenCalledWith({
+        Name: emailOnly.Name,
+        Domain: emailOnly.Domain,
+        BreachDate: emailOnly.BreachDate,
+        PwnCount: emailOnly.PwnCount,
+        AddedDate: emailOnly.AddedDate,
+        DataClasses: emailOnly.DataClasses,
+        IsSensitive: false,
+      });
     });
     it("skips a breach that leaked neither passwords nor emails", async () => {
       const neither = {
--- a/src/scripts/cronjobs/updateBreachesInRemoteSettings/updateBreachesInRemoteSettings.test.ts
+++ b/src/scripts/cronjobs/updateBreachesInRemoteSettings/updateBreachesInRemoteSettings.test.ts
@@ -124,6 +124,48 @@ describe("updateBreachesInRemoteSettings job", () => {
       await main(mockLog);
       expect(reviewSpy).not.toHaveBeenCalled();
     });
+    it("syncs a breach that leaked emails but no passwords", async () => {
+      const emailOnly = HibpData[0] as HIBP.HibpGetBreachesResponse[number];
+      expect(emailOnly.DataClasses).not.toContain("Passwords");
+      vi.mocked(HIBP.fetchHibpBreaches).mockResolvedValue([emailOnly]);
+      fetchBreachesSpy.mockResolvedValueOnce(new Set([]));
+      addBreachSpy.mockResolvedValue(undefined);
+      reviewSpy.mockResolvedValue(undefined);
+
+      await main(mockLog);
+
+      expect(addBreachSpy).toHaveBeenCalledWith(
+        expect.objectContaining({ Name: emailOnly.Name }),
+      );
+    });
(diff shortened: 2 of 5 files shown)


src/scripts/cronjobs/updateBreachesInRemoteSettings/updateBreachesInRemoteSettings.ts
```diff
@@ -103,6 +111,8 @@ export async function main(parentLogger: Logger) {
       PwnCount: breach.PwnCount,
       AddedDate: breach.AddedDate,
       DataClasses: breach.DataClasses,
+      // The panel hides these, the credential manager may not.
```
GregHilstonMozilla, line 106: this is where I'm adding the new property to the remote settings

src/scripts/cronjobs/updateBreachesInRemoteSettings/updateBreachesInRemoteSettings.ts
```diff
@@ -75,15 +82,16 @@ export async function main(parentLogger: Logger) {
...
   const newBreaches = allHibpBreaches.filter((breach) => {
     return (
       !breach.IsRetired &&
       !breach.IsSpamList &&
       !breach.IsFabricated &&
       breach.IsVerified &&
       breach.Domain !== "" &&
-      breach.DataClasses.includes("Passwords") &&
+      (breach.DataClasses.includes(HibpLabelByDataType.Passwords) ||
+        breach.DataClasses.includes(HibpLabelByDataType.Email)) &&
```
GregHilstonMozilla, line 86: this is where we widen what we send to remote settings to include email data

src/scripts/cronjobs/updateBreachesInRemoteSettings/updateBreachesInRemoteSettings.ts
```diff
@@ -12,10 +12,17 @@ import * as HIBP from "../../../utils/hibp";
...
 type RemoteSettingsBreach = Pick<
   HIBP.HibpGetBreachesResponse[number],
-  "Name" | "Domain" | "BreachDate" | "PwnCount" | "AddedDate" | "DataClasses"
+  | "Name"
+  | "Domain"
+  | "BreachDate"
+  | "PwnCount"
+  | "AddedDate"
+  | "DataClasses"
+  | "IsSensitive"
```
GregHilstonMozilla: linting/formatting changed this line to look like this. the important note is we add `IsSensitive`

src/scripts/cronjobs/updateBreachesInRemoteSettings/updateBreachesInRemoteSettings.ts
joeherm: question: What will happen if an older FF client receives data w/ this data class? Presumably just ignores the email ones? Or do older clients consume the OpenAPI spec?

joeherm (commented):
Code change LGTM- but left a question about the impact on older FF clients.

src/scripts/cronjobs/updateBreachesInRemoteSettings/updateBreachesInRemoteSettings.ts
mansaj: @joeherm Older Firefox password-manager clients will ignore email-only records.

You can see it here: https://searchfox.org/firefox-main/source/browser/components/aboutlogins/LoginBreaches.sys.mjs

```
_breachInvolvedPasswords(breach) {
  return (
    breach.hasOwnProperty("DataClasses") &&
    breach.DataClasses.includes("Passwords")
  );
}
```
Extra fields like `IsSensitive` are ignored. DataClasses is already an unconstrained string array in the collection schema, so "Email addresses" is a new value, not a new shape.

src/scripts/cronjobs/updateBreachesInRemoteSettings/updateBreachesInRemoteSettings.ts
```diff
@@ -75,15 +82,16 @@ export async function main(parentLogger: Logger) {
...
     return (
       !breach.IsRetired &&
       !breach.IsSpamList &&
       !breach.IsFabricated &&
       breach.IsVerified &&
       breach.Domain !== "" &&
-      breach.DataClasses.includes("Passwords") &&
+      (breach.DataClasses.includes(HibpLabelByDataType.Passwords) ||
+        breach.DataClasses.includes(HibpLabelByDataType.Email)) &&
       !remoteBreaches.has(breach.Name)
```
mansaj, line 87: This cron only POSTs names it has never seen, so IsSensitive will never land on existing records.

A client that hides with `if (record.IsSensitive)` will treat missing as false and surface those. HIBP marks them sensitive so visiting the site does not out the user. 

Might be something to consider.. should patch existing records?

src/scripts/cronjobs/updateBreachesInRemoteSettings/updateBreachesInRemoteSettings.ts
mansaj: This Pick gained IsSensitive, but `src/utils/remoteSettingsClient.ts:8-11` still omits it, right? Runtime still POSTs the extra field via JSON.stringify. The next person who maps through the client type will drop it. Maybe update the client type and delete this duplicate?

src/scripts/cronjobs/updateBreachesInRemoteSettings/updateBreachesInRemoteSettings.test.ts
```diff
@@ -124,6 +124,48 @@ describe("updateBreachesInRemoteSettings job", () => {
...
+      const emailOnly = HibpData[0] as HIBP.HibpGetBreachesResponse[number];
+      expect(emailOnly.DataClasses).not.toContain("Passwords");
+      vi.mocked(HIBP.fetchHibpBreaches).mockResolvedValue([emailOnly]);
+      fetchBreachesSpy.mockResolvedValueOnce(new Set([]));
+      addBreachSpy.mockResolvedValue(undefined);
+      reviewSpy.mockResolvedValue(undefined);
+
+      await main(mockLog);
+
+      expect(addBreachSpy).toHaveBeenCalledWith(
```
mansaj: This would still pass if DataClasses or IsSensitive: false are gone right? Can we assert those on the payload?

src/scripts/cronjobs/updateBreachesInRemoteSettings/updateBreachesInRemoteSettings.ts
GregHilstonMozilla: Great call out, documenting here what we discussed in stand up for posterity. 

I am going to ask Vincent and Kaitlyn if there's a way to backfill the data right into Remote Settings, so I can avoid having to write new code, and test it, for this one off operation!

src/scripts/cronjobs/updateBreachesInRemoteSettings/updateBreachesInRemoteSettings.ts
GregHilstonMozilla: Good catch, fixed in a commit!

src/app/functions/universal/breach.ts
```diff
@@ -6,6 +6,11 @@ import { DataClassEffected } from "../../../utils/subscriberBreaches";
 
 // TODO: Move pure functions that operate on breaches to this file
 
+/**
+ * Every kind of data a breach can leak, in Monitor's own kebab-case spelling.
```
Vinnl: Maybe good to know: this list is far from exhaustive. A fuller list is in [this file](https://github.com/mozilla/blurts-server/blob/main/locales/en/data-classes.ftl); however, HIBP can extend it with new entries whenever it finds a new breach with a data type that it hadn't seen before. (I think we're supposed to regularly [update that list](https://github.com/mozilla/blurts-server/pull/3879) so that they're localised when they show up on https://monitor.mozilla.org/breaches, but I don't think anyone was ever responsible for that.)

I *think* that the data types listed here are the ones we have recommendations for on [a public breach listing page](https://fx-monitor-storybook.netlify.app/?path=/story/pages-public-breach-listing--breach-detail-view-story), but it predates my time on Monitor, so I'm not sure.

src/app/functions/universal/breach.ts
```diff
@@ -22,6 +27,39 @@ export const BreachDataTypes = {
...
+  IP: "IP addresses",
+  Address: "Physical addresses",
+  DoB: "Dates of birth",
+  Phone: "Phone numbers",
+  SecurityQuestions: "Security questions and answers",
+  HistoricalPasswords: "Historical passwords",
+} as const satisfies Omit<
+  Record<keyof typeof BreachDataTypes, string>,
+  "General"
+>;
```
Vinnl, line 61: Nice work to ensure these two stay aligned.

src/scripts/cronjobs/updateBreachesInRemoteSettings/updateBreachesInRemoteSettings.ts
Vinnl: It looks like the Remote Settings UI has individual deletion buttons for all entries (currently ~700); I didn't dare click on one yet, but possibly we could click on all of them with a `document.querySelectorAll`, and then this script would recreate them? And there is a stage env, so while we'd need to figure out how that works, we could test it there.

That said, I'm not sure if it's less work, or safer, than updating the script.

(In fact, after updating the script, if it makes a mistake, we can just not approve it for Prod, so I think that might be safer.)

Vinnl (approved):

src/app/functions/universal/breach.ts
GregHilstonMozilla: Okay thanks for this call out, I will roll this into the docstring, TY

src/scripts/cronjobs/updateBreachesInRemoteSettings/updateBreachesInRemoteSettings.ts
GregHilstonMozilla: @Vinnl and I discussed this, we are going to remove the addition of `IsSensitive` from the Remote Settings posting, as it is not needed for this effort

Vinnl (approved):
Changes look good to me! As discussed, no more need for `IsSensitive`, considering that we're not even going to be telling the trust panel about sensitive breaches.

PR #6826, opened 2026-09-04
Mntor 5319

GregHilstonMozilla:
# References:

Jira: [MNTOR-5319](https://mozilla-hub.atlassian.net/browse/MNTOR-5319)

Incident doc: [2026-09-03 Monitor Website Unresponsive](https://docs.google.com/document/d/1arUsmpbZwPQZtPT1zu0RsOKNXcePvE0pRMbVil877zA/edit)

# Description

[`redisClient()`](https://github.com/mozilla/blurts-server/blob/783560d3aee8f583c62ac0cab6cacb52d0c99e4b/src/db/redis/client.ts#L14-L23) declared a module-level `singleton` but never read it, so every call opened a new ioredis connection and abandoned the previous one. [`getAllBreachesFromDb()`](https://github.com/mozilla/blurts-server/blob/783560d3aee8f583c62ac0cab6cacb52d0c99e4b/src/utils/hibp.ts#L278-L280) calls it per request, so the web tier leaked roughly one socket per request until Memorystore hit its 64,998 `maxclients` ceiling.

Deploys and pod restarts have been silently protecting prod from this, as the connections would be cleared out.

Scope note: this PR was originally larger. The breach read-path hardening moved out to [MNTOR-5378](https://mozilla-hub.atlassian.net/browse/MNTOR-5378) with its own PR, and the per-request catalogue parse to [MNTOR-5380](https://mozilla-hub.atlassian.net/browse/MNTOR-5380), so this one touches `src/db/redis/` only.

# Testing Before and After

Built for production twice, once on `main` and once on this branch, pointed `REDIS_URL` at a throwaway `docker run -d --name redis-demo -p 6379:6379 redis:7-alpine`, then fired 300 requests at `/en/breaches` with concurrency 10 while sampling `redis-cli info clients` once a second.

Redis connections:

| sec | main | this branch |
| --- | ---- | ----------- |
| 1 | 0 | 1 |
| 5 | 60 | 1 |
| 10 | 134 | 1 |
| 15 | 210 | 1 |
| 20 | 283 | 1 |
| 22, load ends | 300 | 1 |
| 30 | 300 | 1 |

One socket per request, and they are never released. That last row is the incident in miniature: nothing reclaims them until the pod dies.

Response times, same 300 requests. Sharing one connection instead of 300 was the thing I was most worried about, and it turns out to be slightly faster, because removing a TCP handshake per request more than pays for the serialised reply.

| | p50 | p95 | p99 | max |
| --- | --- | --- | --- | --- |
| main | 688ms | 807ms | 977ms | 996ms |
| this branch | 683ms | 763ms | 831ms | 966ms |

Separately, stopping Redis for 8 seconds under a live client: `main` dies at t≈1.2s with `Redis Could not connect after 4 attempts`.

# How to test

```sh
npx vitest run src/db/redis
```

Each regression case fails on `main`, which reports:

- [`client.test.ts`](https://github.com/mozilla/blurts-server/blob/783560d3aee8f583c62ac0cab6cacb52d0c99e4b/src/db/redis/client.test.ts): 100 connections for 100 calls
- `util.test.ts`: throws at `retryStrategy(4)`

- [x] Commits in this PR are minimal and have descriptive commit messages. Four commits, each with its own tests.
- [x] I've added or updated the relevant sections in readme and/or code comments. `retryStrategy` carries a doc comment explaining why it must never throw, and why the cap doubles as the worst-case per-request stall.
- [x] I've added a unit test to test for potential regressions of this bug. One per commit, each verified to fail on `main`.
- [x] If this PR implements a feature flag or experimentation, I've checked that it still works with the flag both on, and with the flag off. Not applicable, see above for why there is no flag.
- [ ] All acceptance criteria are met. Two of three in code. The third, Sentry FIREFOX-MONITOR-D5H and 6HV no longer recurring, is confirmed post-deploy.

# Follow Up Action

Watch `redis.googleapis.com/clients/connected` for 24 to 48h after deploy. It should stay near its ~450 floor instead of climbing.

[MNTOR-5319]: https://mozilla-hub.atlassian.net/browse/MNTOR-5319?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[MNTOR-5378]: https://mozilla-hub.atlassian.net/browse/MNTOR-5378?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/db/redis/client.ts
+++ b/src/db/redis/client.ts
@@ -2,8 +2,8 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
-import { logger } from "@sentry/core";
 import { createRedisInstance } from "./util";
+import { logger } from "../../app/functions/server/logging";
 import type { Redis } from "ioredis";
 import MockRedis from "ioredis-mock";
 
@@ -18,12 +18,9 @@ export const redisClient = () => {
     return singleton;
   }
 
-  if (process.env.REDIS_URL?.includes("redis.mock")) {
-    singleton = new MockRedis();
-    logger.debug("redis_mock_client_created_success");
-  } else {
-    singleton = createRedisInstance();
-    logger.debug("redis_client_created_success");
-  }
+  const useMock = process.env.REDIS_URL?.includes("redis.mock") ?? false;
+  singleton = useMock ? new MockRedis() : createRedisInstance();
+  // One line per process, so prod logs prove the reuse.
+  logger.info("redis_client_created", { mock: useMock });
   return singleton;
 };
--- a/src/db/redis/client.ts
+++ b/src/db/redis/client.ts
@@ -11,8 +11,13 @@ export const REDIS_ALL_BREACHES_KEY = "breaches";
 export const REDIS_ALL_DATA_BROKERS_KEY = "dataBrokers";
 export const BREACHES_EXPIRY_SECONDS = 3600 * 12; // 12 hour
 
-let singleton: Redis;
+let singleton: Redis | undefined;
 export const redisClient = () => {
+  // Reuse the connection; a new one would cause a leak
+  if (singleton) {
+    return singleton;
+  }
+
   if (process.env.REDIS_URL?.includes("redis.mock")) {
     singleton = new MockRedis();
     logger.debug("redis_mock_client_created_success");
--- a/src/db/redis/util.ts
+++ b/src/db/redis/util.ts
@@ -6,7 +6,24 @@ import { Redis, RedisOptions } from "ioredis";
 import { redisConfiguration } from "./configuration";
 import { logger } from "../../app/functions/server/logging";
 
-// TODO: Testing
+// Long enough to stop hammering, short enough to recover.
+const MAX_RETRY_DELAY_MS = 5000;
+
+/**
+ * How long ioredis waits before retrying a lost connection.
+ *
+ * This must never throw. ioredis calls it outside any try/catch, so a throw
+ * here escapes as an uncaught exception and takes the process down instead
+ * of letting the connection heal. Returning a delay every time means a
+ * refused connection keeps retrying until Redis comes back.
+ *
+ * @param times how many reconnect attempts have already been made
+ * @returns milliseconds to wait before the next attempt
+ */
+export function retryStrategy(times: number): number {
+  return Math.min(times * 200, MAX_RETRY_DELAY_MS);
+}
+
 /* c8 ignore start */
 function getRedisConfiguration(): {
   port: number;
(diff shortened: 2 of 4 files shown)


docs/architecture/flows/breach-read-path.md
GregHilstonMozilla: Just some simple update to our documentation now that we made a change to the flow of data when Redis is down.

src/app/functions/server/getBreaches.test.ts
GregHilstonMozilla: This file is being added as the original `src/app/functions/server/getBreaches.ts` had no tests

src/db/redis/client.ts
```diff
@@ -2,23 +2,25 @@
...
+let singleton: Redis | undefined;
 export const redisClient = () => {
-  if (process.env.REDIS_URL?.includes("redis.mock")) {
-    singleton = new MockRedis();
-    logger.debug("redis_mock_client_created_success");
-  } else {
-    singleton = createRedisInstance();
-    logger.debug("redis_client_created_success");
+  // Reuse the connection; a new one would cause a leak
+  if (singleton) {
```
GregHilstonMozilla, line 17: This is the big fix in this PR, allowing our singleton to actually be leveraged.

src/db/redis/client.ts
```diff
@@ -2,23 +2,25 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
-import { logger } from "@sentry/core";
 import { createRedisInstance } from "./util";
+import { logger } from "../../app/functions/server/logging";
```
GregHilstonMozilla, line 6: this fixes the capturing of our logs for cron executions, as previously they were going into the void

src/db/redis/util.ts
```diff
@@ -24,13 +45,7 @@ export function createRedisInstance(config = getRedisConfiguration()) {
       showFriendlyErrorStack: true,
       enableAutoPipelining: true,
       maxRetriesPerRequest: 0,
-      retryStrategy: (times: number) => {
-        if (times > 3) {
-          throw new Error(`Redis Could not connect after ${times} attempts`);
```
GregHilstonMozilla, line 29: see the doc comment above for why this change was important

src/utils/hibp.ts
GregHilstonMozilla: big take away here is we've separated out the three failures into three separate logs we can more easily trace. we can make these metrics in a future effort

joeherm (approved):
Looks great- good catch!

